PR #358, opened 2014-02-13
Added support for SPM. Still needs work.

georg-rath:
A problem I ran into with this easyblock is that the zip-file when extracted has a name like 'spm8', which I think is quite ugly, so I tried to flatten it (equivalent of --strip-components=1 in tar). I am not sure why deleting the original directory after moving breaks the prepare step. Would be nice if someone could comment on this.

Also I am not sure if there is a better way to check if the MATLAB module is loaded.

Stuff that needs to be done before a merge:
- Clean out the compiled MATLAB files before building (there are precompiled versions included).
- Check if the MATLAB compiler takes some kind of compiler optimization flags.
- Probably update the license.

### DIFF
--- a/easybuild/easyblocks/s/spm.py
+++ b/easybuild/easyblocks/s/spm.py
@@ -13,19 +13,13 @@ EasyBuild support for SPM, implemented as an easyblock
 import os
 import shutil
 from easybuild.easyblocks.generic.configuremake import ConfigureMake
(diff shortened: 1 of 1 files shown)


boegel:
This was opened over 10 years ago, with most recent update from June 2014, so at the very least it needs a thorough revision before it can be merged.

I'll close this, if someone still has interest in SPM, they can use it as a starting point.

I'm sure @georg-rath will forgive us for not picking up on it....

PR #394, opened 2014-04-17
SNPhylo support

ehiggs:
SNPhylo is a glue script to pull in some R, Python, MUSCLE, and PHYLIP. Installation is a script which asks yes/no questions and then generates a single bash script.

This block just says yes to everything and then copies the script into the appropriate location.


boegel:
merged via #865, thanks @ehiggs!

PR #463, opened 2014-08-28
made configure and make run async

JensTimmerman:
- you can now see the output of the configure and make commands scroll by if you're in debug mode
- only in easyblocks derived from configuremake and cmakemake

depends on https://github.com/hpcugent/easybuild-framework/pull/1026


JensTimmerman:
@boegel is this still relevant or should I close?

JensTimmerman:
https://github.com/easybuilders/easybuild-framework/pull/1026 was never fixed and this has conflicts, so not going to look at this, someone else can have a go if still wanted.

PR #470, opened 2014-09-04
Gamess

pescobar:
easyblock for gamess. Written by @boegel , I just did some dirty hacks to make it somehow work


boegel:
@pescobar: PR #558 is an enhanced version of this easyblock (after seeing #544 coming in), which is cleaner (no hardocoding) and does more: it also runs the tests included, and runs the provided verification script to check whether the tests worked correctly, etc.

PR #475, opened 2014-09-12
verify installation of a dummy Python package before actually installing a Python package

boegel:
After spending about 2h on trying to figure out why a particular `setuptools` installation was blatently ignoring `--prefix`, I was sufficiently frustrated to come up with this: a verification of whether whichever `setuptools` is active honors `--prefix` before actually installing a Python package with it.

Up until now, I've run into two different ways of making `setuptools` blatently ignore `--prefix`:
- via the `distutils.cfg` or `$HOME/.pydistutils.cfg` configuration files which effectively override any specified `--prefix` comand line option (cfr. https://docs.python.org/2/install/#distutils-configuration-files)
- by using a `setuptools` installation that was installed with `--user` into `$HOME/.local`, causing it to install everything to `$HOME/.local` regardless of what `--prefix` specifies (yes, you read that right)

Since there's little we can do to prevent the use of a (basically broken) `setuptools` installation (there's no way I'm aware of in which we can install our own and be 100% sure it'll be used), we better have a way of verifying whether the `active` setuptools does the right thing.

related: https://github.com/hpcugent/easybuild/issues/31, https://github.com/hpcugent/easybuild-easyblocks/issues/373 and last nights discussion with @kcgthb on the `#easybuild` IRC channel

@stdweird, @JensTimmerman: thoughts on this?

cc: @kcgthb, @chwilk, @DimitriSteyaert, @pforai, @ocaisa

### DIFF
--- a/easybuild/easyblocks/generic/pythonpackage.py
+++ b/easybuild/easyblocks/generic/pythonpackage.py
@@ -32,6 +32,8 @@ EasyBuild support for Python packages, implemented as an easyblock
 @author: Jens Timmerman (Ghent University)
 """
 import os
+import re
+import shutil
 import tempfile
 from os.path import expanduser
 from vsc import fancylogger
@@ -40,8 +42,35 @@ import easybuild.tools.environment as env
 from easybuild.easyblocks.python import EXTS_FILTER_PYTHON_PACKAGES
 from easybuild.framework.easyconfig import CUSTOM
 from easybuild.framework.extensioneasyblock import ExtensionEasyBlock
-from easybuild.tools.filetools import mkdir, rmtree2, run_cmd
+from easybuild.tools.filetools import mkdir, rmtree2, run_cmd, write_file
 from easybuild.tools.modules import get_software_version
+
+
+# test setup.py script for PythonPackage.python_safe_install
+TEST_SETUP_PY = """#!/usr/bin/env python
+try:
+    from setuptools import setup
+except ImportError:
+    from distutils.core import setup
+
+setup(
+    name='%(pkg)s',
+    version='1.0',
+    scripts=['%(pkg)s.py'],
+    packages=['%(pkg)s'],
+    data_files=['%(pkg)s.txt'],
+    provides=['%(pkg)s.py', '%(pkg)s'],
+    zip_safe=False,
+)
+"""
+TEST_SCRIPT_PY = """#!/usr/bin/env python
+import os, sys
+sys.stdout.write(os.path.dirname(os.path.abspath(__file__)))
+"""
+TEST_INIT_PY = """import os, sys
+def where():
+   sys.stdout.write(os.path.dirname(os.path.abspath(__file__)))
+"""
 
 
 def det_pylibdir():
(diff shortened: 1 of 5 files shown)


easybuild/easyblocks/generic/pythonpackage.py
```diff
@@ -153,6 +182,124 @@ def build_step(self):
...
+        if os.path.exists(os.path.join(datainstalldir, test_data_file)):
+            self.log.debug("Found file %s in %s during verification of dummy Python installation" % tup)
+        else:
+            self.log.warning("Failed to find file %s in %s during verification of dummy Python installation" % tup)
+            verified = False
+
+        if verified:
+            self.log.debug("Verification of dummy Python installation OK.")
+        else:
+            self.log.error("Verification of dummy Python installation failed, setuptools not honoring --prefix?")
```
boegel: no, setuptools may (apparently) blatently ignore `--prefix`

this seems to happen when setuptools itself was installed using `--user`

easybuild/easyblocks/generic/pythonpackage.py
```diff
@@ -153,6 +182,124 @@ def build_step(self):
...
+        write_file(os.path.join(tmpdir, pkg, '__init__.py'), TEST_INIT_PY)
+
+        # install dummy Python package
+        try:
+            os.chdir(tmpdir)
+            testinstalldir = tempfile.mkdtemp()
+            self.python_install(prefix=testinstalldir)
+            os.chdir(cwd)
+        except OSError, err:
+            self.log.error("Failed to move to %s: %s" % (tmpdir, err))
```
boegel: or the latter

easybuild/easyblocks/generic/pythonpackage.py
```diff
@@ -40,10 +42,37 @@
...
 from easybuild.framework.easyconfig import CUSTOM
 from easybuild.framework.extensioneasyblock import ExtensionEasyBlock
-from easybuild.tools.filetools import mkdir, rmtree2, run_cmd
+from easybuild.tools.filetools import mkdir, rmtree2, run_cmd, write_file
 from easybuild.tools.modules import get_software_version
 
 
+# test setup.py script for PythonPackage.python_safe_install
+TEST_SETUP_PY = """#!/usr/bin/env python
+try:
```
boegel: not necessarily, e.g. when installing EasyBuild with EasyBuild, we're typically using the system Python

and even if we think we're using a EB-installed Python with an EB-installed setuptools, another setuptools may 'win' (e.g. one installed with `--user`)

defining `$PYTHONNOUSERSITE` would already help there, though...

easybuild/easyblocks/generic/pythonpackage.py
```diff
@@ -153,6 +182,124 @@ def build_step(self):
...
+                            txt = shebang_re.sub(new_shebang, txt)
+                            open(script, 'w').write(txt)
+                    except IOError, err:
+                        self.log.error("Failed to patch shebang header line in %s: %s" % (script, err))
+
+        # restore PYTHONPATH if it was set
+        if pythonpath is not None:
+            env.setvar('PYTHONPATH', pythonpath)
+
+    def python_safe_install(self, **kwargs):
```
boegel: +1

PR #486, opened 2014-10-09
enhance WIEN2k easyblock for recent versions + cleanup

boegel:
This builds on top of #479 by @MichaelSluydts


boegel:
Tested with easyconfigs in https://github.com/hpcugent/easybuild-easyconfigs/pull/1130 (see , looking great now.

Thanks for initiating this @MichaelSluydts!

PR #491, opened 2014-10-21
Amber12

valtandor:
Easyblock for Amber 12, prepared according to Kenneth's suggestions.


boegel:
Jenkins: ok to test

easybuild/easyblocks/a/amber.py
```diff
@@ -0,0 +1,157 @@
+from easybuild.framework.easyconfig import CUSTOM, MANDATORY, BUILD
```
wpoely86: add a header (look in another easyblock)

easybuild/easyblocks/a/amber.py
```diff
@@ -0,0 +1,157 @@
...
+from easybuild.tools.filetools import run_cmd
+from easybuild.easyblocks.generic.configuremake import ConfigureMake
+import easybuild.tools.environment as env
+from easybuild.tools.modules import get_software_root
+import easybuild.tools.toolchain as toolchain
+import os
```
wpoely86: python imports first, then EB imports (in alphabetical order)

easybuild/easyblocks/a/amber.py
```diff
@@ -0,0 +1,157 @@
...
+        extra_vars.update({
+            # 'Amber': [True, "Build Amber in addition to AmberTools", CUSTOM],
+            'patchlevels': ["latest", "(AmberTools, Amber) updates to be applied", CUSTOM],
+        })
+        return ConfigureMake.extra_options(extra_vars)
+
+    def __init__(self, *args, **kwargs):
+        super(EB_Amber, self).__init__(*args, **kwargs)
+        self.already_extracted = False
+        #self.amberhome = os.path.join(self.installdir, 'amber%s' % self.version)
```
wpoely86: remove?

easybuild/easyblocks/a/amber.py
```diff
@@ -0,0 +1,157 @@
...
+        self.build_in_installdir = True
+
+    def extract_step(self):
+        """Only extract from the tarball if this has not already been done."""
+        if (self.already_extracted == True):
+            pass
+        else:
+            self.cfg['unpack_options'] = "--strip-components=1"
```
wpoely86: use append

easybuild/easyblocks/a/amber.py
```diff
@@ -0,0 +1,157 @@
...
+        else:
+            for (tree, patch_level) in zip(['AmberTools', 'Amber'], self.cfg['patchlevels']):
+                if patch_level == 0: continue
+                cmd = "./update_amber --update-to %s/%s" % (tree, patch_level)
+                # This needs to be run multiple times, more's the pity.
+                run_cmd(cmd, log_all=True)
+                run_cmd(cmd, log_all=True)
+        return super(EB_Amber, self).patch_step(**kw)
+
+    def configure_step(self):
```
wpoely86: why can the default `configure_step` from the `ConfigureMake` block not be used?

valtandor:
@boegel: The software builds successfully on our cluster, but this easyblock is failing a test - apparently when testing it throws up an error log entry and causes an assertion failure. Could you by any chance let me know what the problem is and how I might fix it? I can't get into the machine to figure out the details, unless I've missed an informative link.

easybuild/easyblocks/a/amber.py
```diff
@@ -0,0 +1,187 @@
...
+        return out
+
+    def build_step(self):
+
+        # Set the AMBERHOME environment variable
+        env.setvar('AMBERHOME', self.installdir)
+        try:
+            os.chdir(self.installdir)
+        except OSError, err:
+            self.log.error("Could not chdir to %(amberhome)s: %(error)s" % { 'amberhome': self.installdir, 'error': err })
```
wpoely86: This is the problem. We don't use this anymore. Replace it by:
```python
raise EasyBuildError("Could not chdir to %s: %s",  self.installdir, err ) 
```
And also import it:
```python
```

easybuild/easyblocks/a/amber.py
valtandor: Is this satisfactory?

wpoely86: yes, except that you can claim the copyright entirely on your name :wink:

easybuild/easyblocks/a/amber.py
```diff
@@ -0,0 +1,189 @@
...
+            # unpack_options is a string, not an array; can't use append
+            self.cfg['unpack_options'] += " --strip-components=1"
+            super(EB_Amber, self).extract_step()
+            self.already_extracted = True
+
+    def patch_step(self, **kw):
+        env.setvar('AMBERHOME', self.installdir)
+        if self.cfg['patchlevels'] == "latest":
+            cmd = "./update_amber --update"
+            # This needs to be run multiple times, more's the pity.
```
wpoely86: Is twice gonna suffice? Else I would suggest that you print a warning about this.

easybuild/easyblocks/a/amber.py
```diff
@@ -0,0 +1,189 @@
...
+    def __init__(self, *args, **kwargs):
+        super(EB_Amber, self).__init__(*args, **kwargs)
+        self.already_extracted = False
+        self.build_in_installdir = True
+
+    def extract_step(self):
+        """Only extract from the tarball if this has not already been done."""
+        if (self.already_extracted == True):
+            pass
+        else:
```
wpoely86: a bit of nitpicking but why not simpely:
```python
if not self.already_extracted:
            # unpack_options is a string, not an array; can't use append
            self.cfg['unpack_options'] += " --strip-components=1"
            super(EB_Amber, self).extract_step()
            self.already_extracted = True
```

easybuild/easyblocks/a/amber.py
```diff
@@ -0,0 +1,189 @@
...
+    def patch_step(self, **kw):
+        env.setvar('AMBERHOME', self.installdir)
+        if self.cfg['patchlevels'] == "latest":
+            cmd = "./update_amber --update"
+            # This needs to be run multiple times, more's the pity.
+            run_cmd(cmd, log_all=True)
+            run_cmd(cmd, log_all=True)
+        else:
+            for (tree, patch_level) in zip(['AmberTools', 'Amber'], self.cfg['patchlevels']):
+                if patch_level == 0: continue
```
wpoely86: please split this in two lines.

easybuild/easyblocks/a/amber.py
```diff
@@ -0,0 +1,189 @@
...
+        # Set the AMBERHOME environment variable
+        env.setvar('AMBERHOME', self.installdir)
+        try:
+            os.chdir(self.installdir)
+        except OSError, err:
+            raise EasyBuildError("Could not chdir to {0}: {1}".format(self.installdir, err))
+
+        # Kenneth Hoste recommends making sure the LIBS env var is unset
+        if 'LIBS' in os.environ:
+            del os.environ['LIBS']
```
wpoely86: Use `env.unset_env_vars(['LIBS'])` for this. It will print out a message about the unsetting.

Even better would be to do this in the easyconfig with:
```
unwanted_env_vars = ['LIBS']
```

boegel: Why would doing it in the easyconfig be better? If there's an easyblock, you should do as much as possible in the easyblock, and keep the easyconfig clean?

wpoely86: I have to opposite view: why complicate matters in the easyblock when it can be done in the easyconfig. Much handier to change from version to version. The long term maintenance is much more practical with easyconfigs than with easyblocks. And it is done in a uniform way. It is also immediately clear what is going to happen (no need for `-x`). 

In my view: if you can do it in the easyconfig, do it there. Keep the easyblock as simple as possible.

boegel: Well, I disagree (and we've discussed this before).

The easyblock is 'do-once-and-forget'. In the easyconfig, people may be tempted to try kicking out the `unwanted_env_vars` ("why is this (still) here?"), which may work on their system but not on others.

Having it in the easyconfig results in duplicate 'code'. Maybe in the future another env vars will need to be unset, which will result in having to adjust multiple easyconfigs. Also remember: we don't control all easyconfigs out there. If we can fix it in the easyblock, everybody benefits, also the people who have an old copy of an easyconfig available somewhere, and decide to reuse it in the future without checking what may have changed.

I thing that's a good argument to do it in the easyblock, and it doesn't really complicate things at all imho.

If you want to know what's going to happen, you'll need to look at the easyblock anyway (but `-x` will be a great help there).

easybuild/easyblocks/a/amber.py
valtandor: Because the default `configure_step` insists on using the --prefix option, which Amber's configure script (being an ad hoc script) doesn't implement or recognise.

easybuild/easyblocks/a/amber.py
```diff
@@ -0,0 +1,195 @@
...
+            self.cfg['buildopts'] = 'clean'
+            super(EB_Amber, self).build_step()
+
+    def test_step(self):
+        pass
+
+    def install_step(self):
+        """In Amber, installation is conflated with building,
+        so that 'make install' is done during the build step."""
+        pass
```
boegel: wouldn't it be better to pass the whole `build_step` body in here then?

i.e., collapse build&install into the install step, rather than the build step?

valtandor: Unclear. The build is done in situ so there is no install step as such. In my opinion, it boils down to a philosophical question, and I think the answer is likely to be a matter of opinion and inconsequential.

easybuild/easyblocks/a/amber.py
valtandor: I dare say this is now resolved.

wpoely86:
@boegel lgtm

valtandor:
@boegel: About three months ago, a question was asked about whether the build process should be incorporated into the install step, rather than vice versa (as it currently is). Could you give me an update as to whether any work is desired on this easyblock before it is adopted.

boegel:
@valtandor: it's OK to leave as is w.r.t. build vs install step, I guess I missed that this is a build-in-installdir procedure

one small remark: please add docstrings in the `configure_step` and `build_step` methods

valtandor:
@boegel: Docstrings have been added.

boegel:
updated version merged via #958, so this can be closed @valtandor?

valtandor:
I think that should be OK. Closing.

PR #494, opened 2014-10-27
METIS easyblock taking into account configopts

hajgato:


boegel:
Jenkins: test this please

easybuild/easyblocks/m/metis.py
```diff
@@ -47,7 +47,7 @@ def configure_step(self, *args, **kwargs):
 
         if LooseVersion(self.version) >= LooseVersion("5"):
 
-            cmd = "make config prefix=%s" % self.installdir
+            cmd = "make %s config prefix=%s" % (self.cfg['configopts'],self.installdir)
```
boegel: minor style issue: space after `,`

easybuild/easyblocks/m/metis.py
```diff
@@ -125,7 +125,7 @@ def sanity_check_step(self):
 
         custom_paths = {
                         'files': ['bin/%s' % x for x in binfiles] + ['include/%s' % x for x in incfiles] +
-                                 ['lib/libmetis.a'],
+                                 [('lib/libmetis.a','lib/libmetis.so')],
```
boegel: minor style issue: space after `,`

boegel: None of the current `METIS` installations include `libmetis.so`... So, this requires a particular configure option to be used?

If so, either include that in the easyblock, or don't include `libmetis.so` in the sanity check in the easyblock (you can easily override `sanity_check_paths` in an easyconfig that should produce it).

hajgato:
@boegel metis easyblock corrected

easybuild/easyblocks/m/metis.py
```diff
@@ -123,9 +123,22 @@ def sanity_check_step(self):
         if LooseVersion(self.version) < LooseVersion("5"):
             dirs += ["Lib"]
 
+        libext = []
+        if isinstance(self.cfg['configopts'], list):
+            for elements in self.cfg['configopts']:
+                if 'shared=1' in elements:
```
boegel: this is a very involved bit of code, why not simply enable building of the `.so` by always including `shared=1`?

```python
# in configure_step
self.cfg.update('configopts', 'shared=1')
```

Alternatively, if you check for `shared=1` in `self.cfg['configopts']` in `configure_step`, you don't need to worry about checking whether `self.cfg['configopts']` is a list or not (you can assume it'll be a string when in `configure_step`, in fact, we already do when constructing `cmd`).

hajgato: METIS makes shared object _or_ archive. Never both. If I use the above mentioned line, I might break old builds, if they would not work with the shared object file. 
I had to give a list in my METIS easyconfig to build both archive and shared object.
```
configopts = [' ', 'shared=1 ']
```

boegel: OK, thanks for clarifying that.

But, we should still move the check which libraries should get installed to `configure_step`, to avoid dealing with the list/string status of `configopts`.

I'll send a PR your way for that.

boegel:
Works fine, all existing METIS easyconfigs still work with this in place, see https://gist.github.com/6373a7707218240528a3.

The METIS easyconfig included in https://github.com/hpcugent/easybuild-easyconfigs/pull/1166, for which both `libmetis.so` and `libmetis.a` are built works fine too, so going in.

Thanks @hajgato!

PR #496, opened 2014-10-28
Easyblock for XERCES library

greg-hernandez:
Easyblock contribution for XERCES library support


boegel:
better late than never, I guess... :(

Jenkins: ok to test

branfosj:
We now have easyconfigs for Xerces-C++ using CMake, so I'll close this as no longer required.

PR #497, opened 2014-10-28
OpenFOAM with external metis, cgal, and paraview

hajgato:


boegel:
Jenkins: test this please

easybuild/easyblocks/o/openfoam.py
```diff
@@ -139,9 +139,13 @@ def configure_step(self):
...
-            scotch = get_software_root('SCOTCH')
-            if scotch:
-                self.cfg.update('prebuildopts', "SCOTCH_ROOT=$EBROOTSCOTCH")
+            for depend in ['SCOTCH', 'METIS', 'CGAL', 'PARAVIEW']:
+                dependloc = None
+                dependloc = get_software_root(depend)
+                if dependloc:
+                    self.cfg.update('prebuildopts', "{0}_ROOT=$EBROOT{0}".format(depend))
+                    if depend == 'CGAL':
+                        self.cfg.update('prebuildopts', "BOOST_ROOT=$EBROOTBOOST")
```
boegel: also use `get_software_root` here, to make sure `Boost` is actually available?

easybuild/easyblocks/o/openfoam.py
```diff
@@ -139,9 +139,13 @@ def configure_step(self):
...
                 self.cfg.update('prebuildopts', "%(name)s_INCLUDE_DIR=$EBROOT%(name)s/include" % {'name': dep['name'].upper()})
         else:
-            scotch = get_software_root('SCOTCH')
-            if scotch:
-                self.cfg.update('prebuildopts', "SCOTCH_ROOT=$EBROOTSCOTCH")
+            for depend in ['SCOTCH', 'METIS', 'CGAL', 'PARAVIEW']:
+                dependloc = None
+                dependloc = get_software_root(depend)
+                if dependloc:
+                    self.cfg.update('prebuildopts', "{0}_ROOT=$EBROOT{0}".format(depend))
```
boegel: this looks better than the approach with `%(name)s` used above; are you up for changing the other instances of this pattern throughout this easyblock?

hajgato:
@boegel openfoam easyblock corrected

easybuild/easyblocks/o/openfoam.py
```diff
@@ -136,12 +136,18 @@ def configure_step(self):
...
+                self.cfg.update('prebuildopts', "{0}_INCLUDE_DIR=$EBROOT{0}/include".format(dep['name'].upper()))
         else:
-            scotch = get_software_root('SCOTCH')
-            if scotch:
-                self.cfg.update('prebuildopts', "SCOTCH_ROOT=$EBROOTSCOTCH")
+            for depend in ['SCOTCH', 'METIS', 'CGAL', 'PARAVIEW']:
+                dependloc = None
+                dependloc = get_software_root(depend)
+                if dependloc:
+                    if depend == 'CGAL' and get_software_root('BOOST'):
```
boegel: what if CGAL *is* a dep, but Boost is not available?

Then `CGAL_ROOT` will not be specified. Is that the intention here, or is the combination of having CGAL without Boost a problem?

hajgato: @boegel `CGAL` have a `Boost` dependency. When we use `CGAL`, we have to give to `OpenFOAM` where is `Boost`. The original version did not check whether we have `Boost`, because I assumed that is available in `CGAL` (since it depends on it). Now I am checking it (as you requested previously). All in all, we need the extra boost `configopt` only if we have `CGAL`. And If we have `CGAL` then withourt giving the `Boost` location, the compilation is unsuccessful, I assume it needs some `Boost` header files. The current `CGAL` cannot be compiled without `Boost` but I do not know about past and future versions. (all of our `CGAL` easyconfigs have `Boost` dependency)

boegel:
OK, this looks good, I'll give this a test with an existing OpenFOAM easyconfig, but should be fine.

easybuild/easyblocks/o/openfoam.py
```diff
@@ -198,12 +198,18 @@ def configure_step(self):
...
                 self.cfg.update('prebuildopts', "%s_SYSTEM=1" % dep['name'].upper())
-                self.cfg.update('prebuildopts', "%(name)s_LIB_DIR=$EBROOT%(name)s/lib" % {'name': dep['name'].upper()})
-                self.cfg.update('prebuildopts', "%(name)s_INCLUDE_DIR=$EBROOT%(name)s/include" % {'name': dep['name'].upper()})
+                self.cfg.update('prebuildopts', "{0}_LIB_DIR=$EBROOT{0}/lib".format(dep['name'].upper()))
+                self.cfg.update('prebuildopts', "{0}_INCLUDE_DIR=$EBROOT{0}/include".format(dep['name'].upper()))
         else:
-            scotch = get_software_root('SCOTCH')
-            if scotch:
-                self.cfg.update('prebuildopts', "SCOTCH_ROOT=$EBROOTSCOTCH")
+            for depend in ['SCOTCH', 'METIS', 'CGAL', 'PARAVIEW']:
```
boegel: @hajgato: can you replace `PARAVIEW` with `Paraview`, and `'BOOST'` with `'Boost'` below?

Right now, it matters little, since `get_software_root` will use `.upper()` on the name it is passed, but this may change in the future

easybuild/easyblocks/o/openfoam.py
```diff
@@ -198,12 +198,18 @@ def configure_step(self):
...
+                dependloc = None
+                dependloc = get_software_root(depend)
+                if dependloc:
+                    if depend == 'CGAL' and get_software_root('BOOST'):
+                        self.cfg.update('prebuildopts', "{0}_ROOT=$EBROOT{0}".format(depend))
+                        self.cfg.update('prebuildopts', "BOOST_ROOT=$EBROOTBOOST")
```
boegel: use `get_software_root('Boost')` rather than hardcoding `$EBROOTBOOST`

easybuild/easyblocks/o/openfoam.py
```diff
@@ -198,12 +198,18 @@ def configure_step(self):
...
         else:
-            scotch = get_software_root('SCOTCH')
-            if scotch:
-                self.cfg.update('prebuildopts', "SCOTCH_ROOT=$EBROOTSCOTCH")
+            for depend in ['SCOTCH', 'METIS', 'CGAL', 'PARAVIEW']:
+                dependloc = None
+                dependloc = get_software_root(depend)
+                if dependloc:
+                    if depend == 'CGAL' and get_software_root('BOOST'):
+                        self.cfg.update('prebuildopts', "{0}_ROOT=$EBROOT{0}".format(depend))
```
boegel: hmm, same here actually, better not use `$EBROOT`, but use the value of `dependloc`

boegel:
Looks good, retested with existing OpenFOAM easyconfigs, working great, so going in, thanks @hajgato!

PR #499, opened 2014-10-29
Julia

pm67nz:
An easyblock for Julia, which has not configure script.  prebuildopts and buildopts didn't work for me, so this easy block does what the Juila documentation says to do and writes a Make.user file.  I use it with an easyconfig which includes:

configopts = """
USE_SYSTEM_BLAS=1
USE_SYSTEM_LAPACK=1
USE_SYSTEM_FFTW=1
"""


boegel:
@pm67nz: several open remarks here?

boegel:
Jenkins: ok to test

pm67nz:
I'm not sure this works with more recent Julia versions.

PR #502, opened 2014-10-31
move support for staged_install from CPLEX easyblock to generic Binary easyblock

boegel:
that way, the staged install option is also available for other easyblocks that derive from `Binary` (e.g. for ABAQUS, which triggered this PR), and for easyconfigs that use `Binary` directly

### DIFF
--- a/easybuild/easyblocks/c/cplex.py
+++ b/easybuild/easyblocks/c/cplex.py
@@ -93,6 +93,7 @@ class EB_CPLEX(Binary):
 
     def post_install_step(self):
         """Determine bin directory for this CPLEX installation."""
+        # handle staged install via Binary parent class
         super(EB_CPLEX, self).post_install_step()
 
         # determine bin dir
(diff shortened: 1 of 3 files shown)


easybuild/easyblocks/c/cplex.py
```diff
@@ -98,20 +86,15 @@ def install_step(self):
...
-                self.log.error("Failed to move staged install from %s to %s: %s" % (stagedir, self.installdir, err))
-
         try:
             os.chmod(self.installdir, stat.S_IRWXU | stat.S_IXOTH | stat.S_IXGRP | stat.S_IROTH | stat.S_IRGRP)
         except OSError, err:
             self.log.exception("Can't set permissions on %s: %s" % (self.installdir, err))
 
+    def post_install_step(self):
+        """Determine bin directory for this CPLEX installation."""
+        super(EB_CPLEX, self).post_install_step()
```
boegel: yes, since `Binary` (i.e., the parent class), now takes care of the `staged_install` stuff

boegel: I'll add a comment to clarify that

boegel:
reviving this since I have a need for it (again)


fixed remarks and synced with `develop`, good to go in now, thanks @wpoely86!

PR #510, opened 2014-11-14
also copy hamcrest jar to lib/optional for building ant, if available

boegel:
required for ant v1.9.4, it seems


boegel:
hmm, apparently this was never followed up on...

@wpoely86: please review?

wpoely86:
:+1:

PR #511, opened 2014-11-19
adding a new bcftools block for the new SAMtools separated bcftools

rjeschmi:
This is the block to install the new separate bcftools. It is a copy of the samtools block, with the filelist changed to covery the files included with bcftools.

There is a corresponding easyconfig which I'll merge.

### DIFF
--- a/easybuild/easyblocks/b/bcftools.py
+++ b/easybuild/easyblocks/b/bcftools.py
@@ -1,0 +1,76 @@
+##
+# This file is an EasyBuild reciPY as per https://github.com/hpcugent/easybuild
+#
+# Copyright:: Copyright 2012-2013 University of Luxembourg/Luxembourg Centre for Systems Biomedicine
+# Authors::   Robert Schmidt <roschmidt@ohri.ca>,Cedric Laczny <cedric.laczny@uni.lu>, Fotis Georgatos <fotis.georgatos@uni.lu>, Kenneth Hoste
+# License::   MIT/GPL
+# $Id$
+#
+##
+"""
+EasyBuild support for building BCFtools (BCF - Binary VCF - Variant Call Format), implemented as an easyblock
+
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/b/bcftools.py
```diff
@@ -0,0 +1,52 @@
...
+        """
+        No configure
+        """
+        pass
+
+    def sanity_check_step(self):
+        """Custom sanity check for BCFtools."""
+
+        custom_paths = {
+                        'files': ['bin/%s' % x for x in [f.split('/')[-1] for f in self.bin_files]] ,
```
boegel: This can never work, where does self.bin_files come from?

also, as is, this easyblock is pretty useless, since it only skips `configure` compared to `ConfigureMake` (other than a broken custom `sanity_check_step`)

rjeschmi:
I can't remember what happened here. Perhaps it isn't important (I haven't really had any local requests for this part of samtools).

I'll look at this tomorrow.

easybuild/easyblocks/b/bcftools.py
fgeorgatos: Hi folks,

just for the record, this easyblock was originally written by Cedric, long before any of the likes of `skipsteps` or `makecp` was around (ie. before 1st hackathon). I think that explains much, react as you deem fit.

IMHO, this EasyBlock should rather be called obsolete and move functionality of easyconfigs side.

btw.
This is a good example on why the pile of easyconfigs & corresponding easyblock should be seen as one unit
(ie. belong to same repo and be treatable with a single developer commit)

fgeorgatos: and, it also shows why, the proposition of enforced sanitychecks within easyblocks and 1-1 correspondence of easyblocks to software components should rather be dropped as unsustainable, ie. it does not promote good reusability of python objects, where otherwise it should be possible (and clean to do so).

PR #513, opened 2014-12-02
GROMACS easyblock updated for version 5

wpoely86:
Small fix in sanity check but it doesn't work everywhere: `get_platform_name()` doesn't always return the correct string...

### DIFF
--- a/easybuild/easyblocks/g/gromacs.py
+++ b/easybuild/easyblocks/g/gromacs.py
@@ -26,6 +26,7 @@
 EasyBuild support for building and installing GROMACS, implemented as an easyblock
 
 @author: Kenneth Hoste (Ghent University)
+@author: Ward Poelmans (Ghent University)
 """
 import os
 import re
@@ -35,6 +36,7 @@ from vsc.utils.missing import any
 import easybuild.tools.environment as env
 from easybuild.easyblocks.generic.cmakemake import CMakeMake
 from easybuild.tools.modules import get_software_root
+from easybuild.tools.systemtools import get_platform_name
 
 
 class EB_GROMACS(CMakeMake):
@@ -90,17 +92,17 @@ class EB_GROMACS(CMakeMake):
                     libs = os.getenv('LIB%s' % libname)
                     self.cfg.update('configopts', '-DGMX_%s_USER="-L%s %s"' % (libname, lib_dir, libs))
 
+            # set regression test path
+            prefix = 'regressiontests'
+            if any([src['name'].startswith(prefix) for src in self.src]):
+                self.cfg.update('configopts', "-DREGRESSIONTEST_PATH='%%(builddir)s/%s-%%(version)s' " % prefix)
+
+        if LooseVersion(self.version) < LooseVersion('5.0'):
             # enable GSL when it's provided
             if get_software_root('GSL'):
                 self.cfg.update('configopts', "-DGMX_GSL=ON")
             else:
                 self.cfg.update('configopts', "-DGMX_GSL=OFF")
-
-            # set regression test path
-            prefix = 'regressiontests'
-            if any([src['name'].startswith(prefix) for src in self.src]):
-                self.cfg.update('configopts', "-DREGRESSIONTEST_PATH='%%(builddir)s/%s-%%(version)s' " % prefix)
-            
 
         # complete configuration with configure_method of parent
         out = super(EB_GROMACS, self).configure_step()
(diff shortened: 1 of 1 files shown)


sbecuwe:
This gromacs.py did not succeed in installing GROMACS 5.0.4 on SL 6.6.
Error: Sanity check failed: no file of ('lib/x86_64-unknown-linux/libgromacs_mpi.a', 'lib64/x86_64-unknown-linux/libgromacs_mpi.a')
FYI: our installation has no x86_64-unknown-linux/ subdirectory of lib64

wpoely86:
@sbecuwe do you mean that those files are directly under `lib64`? Or is the platform name incorrectly detected on SL 6.6?

sbecuwe:
It is installed directly under lib64. (In our 4.6.1 tree, files were directly installed under lib.)
The platform name returned is x86_64-unknown-linux indeed.

wpoely86:
Okay, it seems to do the same thing on one of my test machines. It's bizar that the install dir changed on my system on some point. Might have something to do with plumed.

I'm gonna drop this PR for the time being.

wpoely86:
OK, closed a bit too soon. The situation is:
- On RHEL & derivatives the libs end up under `lib`.
- On Debian based systems, they end up under `lib/x86_64-linux-gnu`.

A check on Linux distro could fix this. Or figuring out why autoconf/libtool want to use a platform name subdir. The latter is probably the preferred but hardest option. If I find the time, I will have a look. 

Don't merge this for the moment.

pescobar:
I tried this one and is working great for gromacs 5.0.4. Thanks @wpoely86 !!

wpoely86:
@pescobar on what distro did you try it?

easybuild/easyblocks/g/gromacs.py
```diff
@@ -139,10 +141,14 @@ def sanity_check_step(self):
         libnames = ['gromacs']
         if LooseVersion(self.version) < LooseVersion('5.0'):
             libnames = ['gmxana', 'gmx', 'gmxpreprocess', 'md']
+
+        libprefix = ''
+        if LooseVersion(self.version) >= LooseVersion('5.0'):
+            libprefix = get_platform_name()
         libs = ['lib%s%s.a' % (libname, suff) for libname in libnames]
         custom_paths = {
             'files': ['bin/%s%s' % (binary, suff) for binary in ['editconf', 'g_lie', 'genbox', 'genconf', 'mdrun']] +
```
boegel: GROMACS 5.x includes a `gmx` binary that should be used as main program; in 5.0 these are symlinks, in 5.1 the symlinks are gone, so just check for `gmx` in 5.x

boegel: feedback provided by @rapostolo, he had more suggestions for improvements, more info later...

rapostolo:
Regarding the library installation directory have a look at Gromacs commit d5c5abd9aa. There we started using the cmake GNUInstallDirs to set paths on multiarch systems.

easybuild/easyblocks/g/gromacs.py
```diff
@@ -139,10 +141,14 @@ def sanity_check_step(self):
...
-                     [(os.path.join('lib', lib), os.path.join('lib64', lib)) for lib in libs],
-            'dirs': ['include/gromacs', ('lib/pkgconfig', 'lib64/pkgconfig')],
+                     [(os.path.join('lib', libprefix, lib), os.path.join('lib64', libprefix, lib)) for lib in libs],
+            'dirs': ['include/gromacs', (os.path.join('lib', libprefix, 'pkgconfig'), os.path.join('lib64', libprefix, 'pkgconfig'))],
```
boegel: line too long

easybuild/easyblocks/g/gromacs.py
```diff
@@ -90,18 +92,18 @@ def configure_step(self):
                     libs = os.getenv('LIB%s' % libname)
                     self.cfg.update('configopts', '-DGMX_%s_USER="-L%s %s"' % (libname, lib_dir, libs))
 
+            # set regression test path
+            prefix = 'regressiontests'
+            if any([src['name'].startswith(prefix) for src in self.src]):
+                self.cfg.update('configopts', "-DREGRESSIONTEST_PATH='%%(builddir)s/%s-%%(version)s' " % prefix)
+
+        if LooseVersion(self.version) < LooseVersion('5.0'):
```
boegel: so, no more GSL support in 5.x?

wpoely86: as far as I can make out, yes, it's not needed anymore

boegel: @rapostolo: any feedback on this? No more support for using GSL in GROMACS 5.x?

boegel:
@rapostolo: with respect to GROMACS 5.x support, we're seeing something weird...

For me, on Scientific Linux 6, the GROMACS installation is producing files like `$prefix/lib64/libgromacs.a` and `$prefix/bin/gmx`.

For @wpoely86 on Debian on the other hand, the library is going to `$prefix/lib/x86_64-linux-gnu/libgromacs.a`.

Do you have any idea why the library is one level deeper on Debian? This makes sanity checking, and including the right `prepend-path` statements in module files quite tricky...

easybuild/easyblocks/g/gromacs.py
rapostolo: right, no more GSL in 5.x series: http://www.gromacs.org/About_Gromacs/Release_Notes/Versions_5.0.x

rapostolo:
@boegel : the different paths are likely due to the GNUInstallDirs cmake module for settings paths on multiarch systems. Probably Debian reports x86_64-linux-.., while SL6 reports lib64

boegel:
@rapostolo: thanks for the feedback; another reason to despise CMake... ;)

rapostolo:
There might be something in python that works the same way as GNUInstallDirs??

boegel:
@rapostolo: I'll let the easyblock just figure out where `libgromacs.a` is located, and use that path. I'm not keen on figuring out what kind of magic CMake pulls to decide one the library subdirectory. I don't care really, as long as the library is there, and EasyBuild can figure out where it went to (so `$LIBRARY_PATH` and co can be set correctly in the module).

boegel:
well-tested with https://github.com/hpcugent/easybuild-easyconfigs/pull/1221 and https://github.com/hpcugent/easybuild-easyconfigs/pull/1599, both on Debian & RedHat systems (which matters a lot for this easyblock)

all existing GROMACS easyconfigs were retested, see https://gist.github.com/4db020f79ec7ef3f1651

going in, thanks @wpoely86 for kicking this off!

PR #515, opened 2014-12-03
Easyblock for wxPython

hajgato:


easybuild/easyblocks/w/wxpython.py
```diff
@@ -0,0 +1,64 @@
+##
+# Copyright 2009-2013 Ghent University
```
fgeorgatos: 2015 ;-)

easybuild/easyblocks/w/wxpython.py
```diff
@@ -0,0 +1,64 @@
...
+class EB_wxPython(PythonPackage):
+    """Support for installing the wxPython Python package."""
+
+    def build_step(self):
+        """Perform the actual Python package build procedure"""
+
+        #wxPython configure, build, and install with one script
+        cmd = 'python '
+        cmd += os.path.join('wxPython', 'build-wxpython.py')
+        cmd += ' --prefix={0} --wxpy_installdir={0} --install '.format(self.installdir)
```
boegel: well, no, because `PythonPackage` performs `python setup.py build`, while there's no `setup.py` at all here...

boegel:
merged via #883, thanks @hajgato!

PR #522, opened 2014-12-11
latest picard-1.119 doesn't include sam-119.jar. Fixed sanity check

pescobar:


boegel:
tested with https://github.com/hpcugent/easybuild-easyconfigs/pull/1269, working great, thanks @pescobar!

PR #535, opened 2015-01-07
GCC LTO cleanup

geimer:
GCC's LTO support is controlled via the custom `withlto` easyconfig parameter -- and so should be the corresponding sanity check.

### DIFF
--- a/easybuild/easyblocks/g/gcc.py
+++ b/easybuild/easyblocks/g/gcc.py
@@ -525,7 +525,7 @@ class EB_GCC(ConfigureMake):
             bin_files.append('gfortran')
             lib_files.extend(['libgfortran.%s' % sharedlib_ext, 'libgfortran.a'])
 
-        if 'lto' in self.cfg['languages']:
+        if self.cfg['withlto']:
             libexec_files.extend(['lto1', 'lto-wrapper'])
             if os_type in ['Linux']:
                 libexec_files.append('liblto_plugin.%s' % sharedlib_ext)


wpoely86:
ok to test

easybuild/easyblocks/g/gcc.py
```diff
@@ -525,7 +525,7 @@ def sanity_check_step(self):
             bin_files.append('gfortran')
             lib_files.extend(['libgfortran.%s' % sharedlib_ext, 'libgfortran.a'])
 
-        if 'lto' in self.cfg['languages']:
+        if self.cfg['withlto']:
```
boegel: since listed `lto` in `languages` is another way of enabling LTO support, shouldn't this be:

```python
if self.cfg['withlto'] or 'lto' in self.cfg['languages']:
```

geimer: Short answer: No.

Long answer: Your proposition is already wrong. As far as I could figure out, passing `lto` via `--enable-languages` was never a documented way of enabling LTO support. The only documented option was, and still is, `--enable-lto` (which is controlled by the `withlto` custom easyconfig parameter). Relying on such undocumented behavior is dangerous at best. Looking a bit more closely at it, passing `lto` to `--enable-languages` only does part of the job, as can be seen in this excerpt from the `configure.ac` file of GCC 4.6.3 (identical code can be found in later versions as well):
```
  # If LTO is enabled, add the LTO front end.
  extra_host_libiberty_configure_flags=
  if test "$enable_lto" = "yes" ; then
    case ,${enable_languages}, in
      *,lto,*) ;;
      *) enable_languages="${enable_languages},lto" ;;
    esac
    if test "${build_lto_plugin}" = "yes" ; then
      configdirs="$configdirs lto-plugin"
      extra_host_libiberty_configure_flags=--enable-shared
    fi
  fi
  AC_SUBST(extra_host_libiberty_configure_flags)
```
That is, if you only pass `lto` as language but disable LTO support using `--disable-lto`, `extra_host_libiberty_configure_flags` will be unset.

boegel: OK, you've clearly figured this out really well, my apologies for questioning your patch!

Thanks a lot for the very detailed explanation, excellent stuff! o/

boegel:
I've also tested this together with https://github.com/hpcugent/easybuild-easyconfigs/pull/1286, working great, so going in.

Thanks @geimer!

PR #536, opened 2015-01-09
revamp EB_libint2 easyblock into EB_Libint that works for both Libint v1x and v2.x

boegel:
@wpoely86: please review

### DIFF
--- a/easybuild/easyblocks/l/libint.py
+++ b/easybuild/easyblocks/l/libint.py
@@ -23,47 +23,60 @@
 # along with EasyBuild.  If not, see <http://www.gnu.org/licenses/>.
 ##
 """
-EasyBuild support for building and installing libint2, implemented as an easyblock
+EasyBuild support for building and installing Libint, implemented as an easyblock
 
 @author: Toon Verstraelen (Ghent University)
 @author: Ward Poelmans (Ghent University)
 """
 
 import os.path
+from distutils.version import LooseVersion
 
 from easybuild.easyblocks.generic.configuremake import ConfigureMake
 
 
-class EB_libint2(ConfigureMake):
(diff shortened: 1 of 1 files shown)


boegel:
cfr. https://github.com/hpcugent/easybuild-easyconfigs/pull/1287

easybuild/easyblocks/l/libint.py
```diff
@@ -23,47 +23,60 @@
...
+        """Custom sanity check for Libint."""
+        if LooseVersion(self.version) >= LooseVersion('2.0'):
+            custom_paths = {
+                'files': ['lib/libint2.a', 'lib/libint2.so', 'include/libint2/libint2.h'],
+                'dirs': [],
+            }
+        else:
+            custom_paths = {
+                'files': ["include/libint/libint.h", "include/libint/hrr_header.h", "include/libint/vrr_header.h",
+                          "lib/libint.a"],
```
wpoely86: no shared lib with 1.x?

easybuild/easyblocks/l/libint.py
```diff
@@ -23,47 +23,60 @@
...
     def make_module_req_guess(self):
         """Specify correct CPATH for this installation."""
-        guesses = super(EB_libint2, self).make_module_req_guess()
-        guesses.update({
-            'CPATH': ["include", os.path.join("include", "libint2")],
-        })
+        guesses = super(EB_Libint, self).make_module_req_guess()
+        if LooseVersion(self.version) >= LooseVersion('2.0'):
+            guesses.update({
+                'CPATH': ["include", os.path.join("include", "libint2")],
```
wpoely86: wouldn't it be nicer to set a string `includepath` in this if and to the actual `guesses.update` below?

PR #537, opened 2015-01-09
remove fes from CP2K sanity check

boegel:
remove `fes` from sanity check, since it's not really part of CP2K (and is no longer there in CP2K v2.6.0)


wpoely86:
good to go

PR #538, opened 2015-01-16
relocate FDTD RPM to fix installation on SL6

boegel:
This fixes https://github.com/hpcugent/easybuild-easyconfigs/issues/640

@wpoely86: please review?

### DIFF
--- a/easybuild/easyblocks/f/fdtd_solutions.py
+++ b/easybuild/easyblocks/f/fdtd_solutions.py
@@ -45,16 +45,16 @@ class EB_FDTD_underscore_Solutions(EasyBlock):
     def build_step(self):
         """No build step for FDTD Solutions."""
         # locate RPM and rebuild it to make it relocatable
-        rpms = glob.glob(os.path.join(self.builddir, 'rpm_install_files', 'FDTD-%s*.rpm' % self.version))
+        rpms = glob.glob(os.path.join(self.cfg['start_dir'], 'rpm_install_files', 'FDTD-%s*.rpm' % self.version))
         if len(rpms) != 1:
-            self.log.debug("Incorrect number of RPMs found, was expecting exactly one: %s" % rpms)
-        rebuilt_dir = os.path.join(self.builddir, 'rebuilt')
+            self.log.error("Incorrect number of RPMs found, was expecting exactly one: %s" % rpms)
+        rebuilt_dir = os.path.join(self.cfg['start_dir'], 'rebuilt')
         rebuild_rpm(rpms[0], rebuilt_dir)
 
         # replace original RPM with relocatable RPM
         rebuilt_rpms = glob.glob(os.path.join(rebuilt_dir, '*', '*.rpm'))
         if len(rebuilt_rpms) != 1:
-            self.log.debug("Incorrect number of rebuilt RPMs found, was expecting exactly one: %s" % rebuilt_rpms)
+            self.log.error("Incorrect number of rebuilt RPMs found, was expecting exactly one: %s" % rebuilt_rpms)
 
         try:
             os.rename(rpms[0], '%s.bk' % rpms[0])
--- a/easybuild/easyblocks/f/fdtd_solutions.py
+++ b/easybuild/easyblocks/f/fdtd_solutions.py
@@ -27,6 +27,10 @@ EasyBuild support for building and installing FDTD Solutions, implemented as an 
 
 @author: Kenneth Hoste (Ghent University)
 """
+import glob
+import os
+import shutil
+from easybuild.easyblocks.generic.rpm import rebuild_rpm
 from easybuild.framework.easyblock import EasyBlock
 from easybuild.tools.run import run_cmd_qa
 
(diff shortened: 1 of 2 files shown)


easybuild/easyblocks/f/fdtd_solutions.py
```diff
@@ -40,7 +44,23 @@ def configure_step(self):
 
     def build_step(self):
         """No build step for FDTD Solutions."""
-        pass
+        # locate RPM and rebuild it to make it relocatable
+        rpms = glob.glob(os.path.join(self.cfg['start_dir'], 'rpm_install_files', 'FDTD-%s*.rpm' % self.version))
```
wpoely86: any particular reason why you need `glob` here? There is a 'random' part in the filename?

JensTimmerman:
ok

easybuild/easyblocks/f/fdtd_solutions.py
```diff
@@ -40,7 +44,23 @@ def configure_step(self):
...
+        rebuilt_dir = os.path.join(self.cfg['start_dir'], 'rebuilt')
+        rebuild_rpm(rpms[0], rebuilt_dir)
+
+        # replace original RPM with relocatable RPM
+        rebuilt_rpms = glob.glob(os.path.join(rebuilt_dir, '*', '*.rpm'))
+        if len(rebuilt_rpms) != 1:
+            self.log.error("Incorrect number of rebuilt RPMs found, was expecting exactly one: %s" % rebuilt_rpms)
+
+        try:
+            os.rename(rpms[0], '%s.bk' % rpms[0])
```
wpoely86: why renamed? We're not going to use the original one anymore?

easybuild/easyblocks/f/fdtd_solutions.py
boegel: I just don't want to make any assumptions here. In the versions I've tested have 'el5' in them, but I don't think that's a safe assumption to make for future versions.

easybuild/easyblocks/f/fdtd_solutions.py
boegel: mostly for debugging purposes, this will be cleaned up once the installation succeeds anyway

easybuild/easyblocks/generic/rpm.py
```diff
@@ -46,6 +48,41 @@
...
+            os.makedirs(targetdir)
+            _log.debug("Created target directory for rebuilt RPMs %s" % targetdir)
+    except OSError, err:
+        _log.error("Failed to create directories for rebuilding RPM: %s" % err)
+
+    _log.debug("Rebuilding %s in %s to make it relocatable" % (rpm_path, targetdir))
+    cmd = ' '.join([
+        "rpmrebuild -v",
+        r"""--change-spec-whole='sed -e "s/^Prefix:.*/Prefix:    \//"'""",
+        r"""--change-spec-whole='sed -e "s/^\(.*:[ ]\+\..*\)/#ERROR \1/"'""",
```
wpoely86: It's not entirely clear what happens here to me?
Your replace `Prefix` by an empty string and then use `#ERROR` on it?

easybuild/easyblocks/generic/rpm.py
```diff
@@ -91,58 +128,29 @@ def configure_step(self):
             ver = res.groupdict()['version']
 
             if LooseVersion(ver) >= LooseVersion('4.8.0'):
```
wpoely86: we have to rebuild if rpm version is higher than `4.8`?

easybuild/easyblocks/generic/rpm.py
boegel: I replace whatever prefix is set with `/`

not sure what the 2nd bit is about anymore... looks like commenting out any other specifications, or some such

whatever, it works ;)

easybuild/easyblocks/generic/rpm.py
```diff
@@ -91,58 +128,29 @@ def configure_step(self):
...
     # --relocate doesn't seem to work (error: Unable to change root directory: Operation not permitted)
-    def rebuildRPMs(self):
+    def rebuild_rpms(self):
         """Rebuild RPMs to make relocation work."""
-
-        # make sure that rpm is available
+        # make sure that rpmrebuild command is available
         if not 'rpmrebuild' in self.cfg['osdependencies']:
             self.cfg['osdependencies'].append('rpmrebuild')
             self.cfg.validate_os_deps()
```
wpoely86: why do this like this instead of directly checking the `rpmrebuild` dep?

easybuild/easyblocks/generic/rpm.py
wpoely86: why not `Prefix: \/` instead of `Prefix:      \/`?

I really don't like it if we don't know what the second part does. As far as I can see, it replaces everything that start with a `.`?

boegel: the 2nd part comments out all lines that refer to relative paths...

this whole trickery is about making RPMs relocatable

wpoely86: relative paths are good no? It's the absolute paths you want to get rid of?

easybuild/easyblocks/generic/rpm.py
boegel: yes, because that version has become stricter... installing RPM under a prefix was easier with SL5 (older version of `rpm`)

easybuild/easyblocks/generic/rpm.py
boegel: I'm not sure... I can fix that.

easybuild/easyblocks/generic/rpm.py
wpoely86: add a comment that this corresponds to SL5 ?

easybuild/easyblocks/generic/rpm.py
boegel: hmm, depends... we're *hard* rewriting whatever `Prefix` is set to, which may be some standard dir like `/usr/lib` or something...

easybuild/easyblocks/generic/rpm.py
boegel: actually, I only know this is needed starting with the version of `rpm` that was on SL6 at the time

boegel: but I'll add a comment to clarify

easybuild/easyblocks/generic/rpm.py
wpoely86: well, if it works...

boegel: indeed... added a comment to mention what we think it's doing ;)

boegel:
fixes the FDTD build, see https://github.com/hpcugent/easybuild-easyconfigs/pull/1307

also verified that this doesn't break the installation of QLogicMPI, which relies on the `Rpm` easyblock

so, going in, thanks for the thorough review @wpoely86!

PR #541, opened 2015-01-23
Fix/suitesparse

gjbex:
Fixed two problems with this easyblock:
1) it assumes variable definitions in the .mk file will not have leading whitespace (assumption doesn't hold for SuiteSparse 4.4.3),
2) the mechanism for ensuring that all variables not replaced in the .mk file are appended works only if the variable name does not occur in any line (e.g., comment) prior to its definition (definitely not the case in SuiteSparse 4.4.3).

Although the current code works, it is not future proof.  In Python 2, the items() method returns a list of key/value tuples, hence it is save to apply the pop() method to the cfgvars dictionary.  In Python 3, the items() method will return an iterator, and removing items from a data structure one is iterating over is not a good idea.
I've only added a comment, since I assume that this pattern will be quite pervasive given the list of authors of this block, so it should probably be addressed more systematically.

### DIFF
--- a/easybuild/easyblocks/s/suitesparse.py
+++ b/easybuild/easyblocks/s/suitesparse.py
@@ -94,11 +94,10 @@ class EB_SuiteSparse(ConfigureMake):
         try:
             for line in fileinput.input(fp, inplace=1, backup='.orig'):
                 for (var, val) in cfgvars.items():
-                    # varaibles in cfgvars that have not been substituted
-                    # here will have to be appended to the end of the file,
-                    # so those substituted should be removed from cfgvars
-                    # if a line was modified
                     orig_line = line
+                    # for variables in cfgvars, substiture lines assignment 
+                    # in the file, whatever they are, by assignments to the
+                    # values in cfgvars
                     line = re.sub(r"^\s*(%s\s*=\s*).*$" % var,
                                   r"\1 %s # patched by EasyBuild" % val,
                                   line)
--- a/easybuild/easyblocks/s/suitesparse.py
+++ b/easybuild/easyblocks/s/suitesparse.py
@@ -59,8 +59,6 @@ class EB_SuiteSparse(ConfigureMake):
         else:
             self.config_name = 'SuiteSparse_config'
 
-        fp = os.path.join(self.cfg['start_dir'], self.config_name, '%s.mk' % self.config_name)
-
         cfgvars = {
             'CC': os.getenv('MPICC'),
             'CFLAGS': os.getenv('CFLAGS'),
@@ -91,16 +89,21 @@ class EB_SuiteSparse(ConfigureMake):
         })
 
         # patch file
+        fp = os.path.join(self.cfg['start_dir'], self.config_name, '%s.mk' % self.config_name)
+
         try:
             for line in fileinput.input(fp, inplace=1, backup='.orig'):
-                for (k, v) in cfgvars.items():
+                for (var, val) in cfgvars.items():
+                    # varaibles in cfgvars that have not been substituted
+                    # here will have to be appended to the end of the file,
+                    # so those substituted should be removed from cfgvars
+                    # if a line was modified
                     orig_line = line
-                    line = re.sub(r"^\s*(%s\s*=\s*).*$" % k,
-                                  r"\1 %s # patched by EasyBuild" % v,
+                    line = re.sub(r"^\s*(%s\s*=\s*).*$" % var,
+                                  r"\1 %s # patched by EasyBuild" % val,
                                   line)
                     if line != orig_line:
-# Note: this will crash & burn in Python 3
-                        cfgvars.pop(k)
+                        cfgvars.pop(var)
                 sys.stdout.write(line)
         except IOError, err:
             self.log.error("Failed to patch %s in: %s" % (fp, err))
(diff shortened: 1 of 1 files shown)


boegel:
Jenkins: ok to test

easybuild/easyblocks/s/suitesparse.py
```diff
@@ -94,8 +94,12 @@ def configure_step(self):
...
             for line in fileinput.input(fp, inplace=1, backup='.orig'):
                 for (k, v) in cfgvars.items():
-                    line = re.sub(r"^(%s\s*=\s*).*$" % k, r"\1 %s # patched by EasyBuild" % v, line)
-                    if k in line:
+                    orig_line = line
+                    line = re.sub(r"^\s*(%s\s*=\s*).*$" % k,
+                                  r"\1 %s # patched by EasyBuild" % v,
+                                  line)
+                    if line != orig_line:
+# Note: this will crash & burn in Python 3
```
boegel: there's lots of stuff all throughout EasyBuild right now which is not Python 3 compatible, so this comment has little value I think

also: the comment should be indented with the code it applies to

easybuild/easyblocks/s/suitesparse.py
```diff
@@ -94,8 +94,12 @@ def configure_step(self):
         try:
             for line in fileinput.input(fp, inplace=1, backup='.orig'):
```
boegel: let's move down the definition of `fp` closer to where it's actually being used, i.e. right above the `try`...

easybuild/easyblocks/s/suitesparse.py
```diff
@@ -94,8 +94,12 @@ def configure_step(self):
...
                 for (k, v) in cfgvars.items():
-                    line = re.sub(r"^(%s\s*=\s*).*$" % k, r"\1 %s # patched by EasyBuild" % v, line)
-                    if k in line:
+                    orig_line = line
+                    line = re.sub(r"^\s*(%s\s*=\s*).*$" % k,
+                                  r"\1 %s # patched by EasyBuild" % v,
+                                  line)
```
boegel: no need to split this across three lines?

easybuild/easyblocks/s/suitesparse.py
```diff
@@ -94,8 +94,12 @@ def configure_step(self):
         try:
             for line in fileinput.input(fp, inplace=1, backup='.orig'):
                 for (k, v) in cfgvars.items():
-                    line = re.sub(r"^(%s\s*=\s*).*$" % k, r"\1 %s # patched by EasyBuild" % v, line)
-                    if k in line:
+                    orig_line = line
+                    line = re.sub(r"^\s*(%s\s*=\s*).*$" % k,
```
boegel: include a comment explaining what's going on here?

easybuild/easyblocks/s/suitesparse.py
```diff
@@ -94,8 +94,12 @@ def configure_step(self):
         try:
             for line in fileinput.input(fp, inplace=1, backup='.orig'):
                 for (k, v) in cfgvars.items():
```
boegel: replace `k` and `v` by non-single-letter variable names (`key` and `val` will do, but more meaningful (yet still short) would be even better)

boegel:
@gjbex: can you tweak the PR title a bit to try and summarise the patch in a single line? that will be a tremendous help in the `git log` that will contain `Merge pull request #541 from hpcleuven:fix/suitesparse: <good title goes here>`

easybuild/easyblocks/s/suitesparse.py
gjbex: When doing a substitution, I find it useful to be able to visually compare the two expressions, aligning them vertically helps avoiding bugs.

gjbex:
Yes, I can tweak it, but allow me to make a few things clear.

1) Code reviews are definitely a good thing, sometimes even for the reviewers.
2) I don't mind issues with my code being raised, and  I certainly don't mind fixing them, they are my responsibility.
3) I do mind getting suggestions to fix someone else's code though, since it is not my responsibility.  I would suggest to take that up with the original authors since they may have had some good reason to do what they did.

easybuild/easyblocks/s/suitesparse.py
gjbex: Please be consistent in your code review, there's another one like it on on 109.

easybuild/easyblocks/s/suitesparse.py
gjbex: For a simple for-loop?

boegel: No, for the regex substitution, maybe just make the existing `patch file` comment a bit clearer, to explain what the regex in the for loop does?

gjbex:
Fixed all issues test report coming with PR 1328 for easyconfigs.

boegel:
@gjbex: thanks a lot for tackling the remarks, despite your reservations, much appreciated

As discussed off-PR, I realise our way of giving feedback can be quite aggressive, and daunting for people new to the process.

In this particular case, you were unlucky to (having to) touch an easyblock that hasn't been getting love for quite some time (Feb'14), and hence is behind on the code style standard we're currently trying to maintain. The code style is constantly evolving (not necessarily changing much though), so it's important to try and resolve any issues we run into when they're being put in the spotlight.

That being said, I understand that you feel like you shouldn't be cleaning up other people's mess. Question is, if we don't tackle it when it's being thrown in our face, when will we ever...

I hope this doesn't put you off w.r.t. to future contributions, since you're clearly "on a roll" the last couple of weeks.

boegel:
Tested this together with https://github.com/hpcugent/easybuild-easyconfigs/pull/1328, working great, looking good, so going in. Thanks again @gjbex!

PR #542, opened 2015-01-26
Deprecate ncurses easyblock

wpoely86:
For hpcugent/easybuild-easyconfigs#1337


easybuild/easyblocks/n/ncurses.py
```diff
@@ -34,18 +36,21 @@ def configure_step(self):
...
-                                  ['lib/lib%s.a' % x for x in ["form", "form", "menu", "menu_g", "ncurses", "ncurses++",
-                                                               "ncurses_g", "panel", "panel_g"]],
-                        'dirs' : ['include']
-                       }
+            'files': ['bin/%s' % x for x in ["captoinfo", "clear", "infocmp", "infotocap", "ncurses5-config",
+                                             "reset", "tabs", "tic", "toe", "tput", "tset"]] +
+                     ['lib/lib%s%s.a' % (x, y) for x in libs for y in ['', '_g', 'w', 'w_g']] +
+                     ['lib/lib%s%s.so' % (x, y) for x in libs for y in ['', 'w']] +
+                     ['lib/libncurses++%s.a' % x for x in ['', 'w']],
+            'dirs': ['include', 'include/ncursesw']
```
boegel: it makes no sense to include this here unless we also include the configure option that makes sure this gets installed in the easyblock itself...

Maybe we should trash this easyblock all together, it has very little reason to exist.

The only major problem is that removing the easyblock will break existing `ncurses` easyconfigs. Which may be acceptable in the light of EasyBuild v2.0...

wpoely86: +1 for trashing it!

boegel: so, can you tackle what needs to be done in the `ncurses` easyconfigs PR you're working on then?

i) adding `easyblock = 'ConfigureMake'` in all of them
ii) add `sanity_check_paths` in all of them
iii) adding the `CPLUS` thing in all of them

Once that is covered, we can re-evaluate dropping this easyblock like a stone, and having a proper way of doing error reporting on this one (specifically).

wpoely86: I would even dare to drop point (iii). I can find no reference about that var anywhere. Even google know nothing.
If a specific easyconfigs depends on it, it should happen there. But g++ and icpc don't use it.

boegel: Maybe this grew historically, I can try and figure out why it's there. Not sure if I'll find something though...

wpoely86:
The `ncurses` easyblock is no longer needed. Everything it does is handled by the easyconfigs in hpcugent/easybuild-easyconfigs#1337

boegel:
@wpoely86: I'm still reluctant to drop this like a stone, at least without a proper error message being thrown if people use old `ncurses` easyconfig files...

Should we deprecate this easyblock for EasyBuild v3.0, and spit out a proper error message pointing to `https://github.com/hpcugent/easybuild-easyconfigs/tree/develop/easybuild/easyconfigs/n/ncurses`?

wpoely86:
@boegel I've put in a deprecated warning for 3.0

boegel:
@wpoely86: remind me about this one in a bit ;)

wpoely86:
We don't need it. There was an env vars being added to the module but it had no purpose. So we don't need the entire easyblock.

boegel:
I know, but I just want to recheck the deprecation before merging this. And updates the docs to flag this easyblock as deprecated.

branfosj:
replaced by removing this for EB5 in #3265

PR #543, opened 2015-01-28
fix easyblock unit tests after dropping support for deprecated behaviour in framework

boegel:


boegel:
Jenkins was still running the tests using Python 2.4, which doesn't make sense since the framework tests for Py2.4 are disabled, and thus the Py2.4-specific installation of framework develop isn't being updated anymore

fixed that now, kicked out the Py2.4 unit test suite run in both easyblocks and easyconfigs PR testers

boegel:
Jenkins: test this please

boegel:
going in, only touched unit tests, and Jenkins is happy (on Py2.6 and Py2.7)

PR #544, opened 2015-01-29
GAMESS-US

scintimandrion:
New easyblock for GAMESS-US.

### DIFF
--- a/easybuild/easyblocks/c/cplex.py
+++ b/easybuild/easyblocks/c/cplex.py
@@ -53,14 +53,6 @@ class EB_CPLEX(Binary):
         super(EB_CPLEX, self).__init__(*args, **kwargs)
         self.bindir = None
 
-    @staticmethod
-    def extra_options():
-        extra_vars = {
-            # staged install via a tmp dir can help with the hard (potentially faulty) check on available disk space
-            'staged_install': [False, "Should the installation should be staged via a temporary dir?", CUSTOM],
-        }
-        return Binary.extra_options(extra_vars)
-
     def install_step(self):
         """CPLEX has an interactive installer, so use Q&A"""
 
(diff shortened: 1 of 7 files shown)


boegel:
Jenkins: ok to test

boegel:
@scintimandrion: have you seen the GAMESS-US easyblock in #470?

That's a ported version of an easyblock we had working with EasyBuild v0.x (different API).

I'll be looking into GAMESS-US in the next couple of days, I'd like to merge your effort and the one in #470 together...

Are you available for interactive discussions (IM, IRC, conf call) in the coming days?

scintimandrion:
I wish I were, but I'm actually on holiday overseas right now. Could it, perchance, wait until I'm back? I get back home on the 23rd of February.

boegel:
@scintimandrion: I've joint this PR together with #470, see #558 resulting in a superior easyblock imho.

To be more specific: I'm using `run_cmd_qa` for running the `config` script, I'm running the tests individually (since using `runall` doesn't work with the `mpi` type build), and I'm verifying the result of the tests too.

I used a lot of the ideas from your easyblock, so thank you very much for contributing it.

Please take a look at #558, and tell me what you think. There's also a PR for an easyconfig for the most recent version of GAMESS-US, alongside a part of your `rungms` patch to add SLURM support.

For making a `sockets` build, you should set `ddi_comm` to `sockets` in your easyconfig file. I've opted to make `mpi` the default though, since that makes sense I think (EasyBuild toolchain typically provide MPI support).

For the `$SCRATCH_DIR` stuff you had in your patch, the easyblock in #558 support setting the `scratch_dir` easyconfig parameter, together with a couple of others custom parameters for `GAMESS-US`.

As soon as my PR is merged in, this PR will close automagically, so no further action required from you here, but I'd love to get some feedback from you since you probably spent quite a bit of time studying the GAMESS-US build procedure...

PR #545, opened 2015-02-03
PSI: rework for future version

wpoely86:
They switched to a CMake based build system

The plugin stuff doesn't yet work but that will probably need patching in the easyconfig.

### DIFF
--- a/easybuild/easyblocks/p/psi.py
+++ b/easybuild/easyblocks/p/psi.py
@@ -32,6 +32,7 @@ EasyBuild support for building and installing PSI, implemented as an easyblock
 from distutils.version import LooseVersion
 import os
 import shutil
+import tempfile
 
 import easybuild.tools.environment as env
 from easybuild.easyblocks.generic.cmakemake import CMakeMake
@@ -160,8 +161,14 @@ class EB_PSI(CMakeMake):
         """
         Run the testsuite of PSI4
         """
-        env.setvar('PSI_SCRATCH', os.getenv('TMPDIR'))
+        testdir = tempfile.mkdtemp()
+        env.setvar('PSI_SCRATCH', testdir)
         super(EB_PSI, self).test_step()
+
+        try:
+            shutil.rmtree(testdir)
+        except OSError, err:
+            self.log.error("Failed to remove test directory %s: %s" % (testdir, err))
 
     def sanity_check_step(self):
         """Custom sanity check for PSI."""
--- a/easybuild/easyblocks/p/psi.py
+++ b/easybuild/easyblocks/p/psi.py
@@ -159,8 +159,8 @@ class EB_PSI(CMakeMake):
     def sanity_check_step(self):
         """Custom sanity check for PSI."""
         custom_paths = {
-            'files': ['bin/psi%s' % self.version.split('.')[0], 'lib/libplugin.a', 'include/libplugin/plugin.h'],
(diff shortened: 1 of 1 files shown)


boegel:
@wpoely86: I'd like to see a test report of reinstalling the existing PSI easyconfigs on top of this modified easyblock...

easybuild/easyblocks/p/psi.py
```diff
@@ -138,7 +156,7 @@ def sanity_check_step(self):
         """Custom sanity check for PSI."""
         custom_paths = {
             'files': ['bin/psi%s' % self.version.split('.')[0]],
-            'dirs': ['include', 'share/psi'],
+            'dirs': ['share/psi'],
```
boegel: why this change?

easybuild/easyblocks/p/psi.py
```diff
@@ -121,7 +87,59 @@ def configure_step(self):
...
 
-        super(EB_PSI, self).configure_step(cmd_prefix=self.cfg['start_dir'])
+        # explicitely specify Python binary to use
+        pythonroot = get_software_root('Python')
+        if not pythonroot:
+            self.log.error("Python module not loaded.")
+        env.setvar('PYTHON', os.path.join(pythonroot, 'bin', 'python'))
+
+        # They have a non-standard variable to define the install path
+        env.setvar('PREFIX', self.installdir)
```
boegel: this wasn't there before, so specific to the recent versions using `CMake`?

boegel:
Looks pretty good overall.

What do you mean with your comment w.r.t. the plugin support exactly? Which easyconfigs will need patching? Or are we talking a patch file that will be required in order to enable plugin support?

easybuild/easyblocks/p/psi.py
wpoely86: They no longer install the headers like this. No idea if that's intentional or a bug in their new build system

easybuild/easyblocks/p/psi.py
wpoely86: Older version also had an CMake based build system, but it was optional.

I don't see the harm in doing this in all cases?

wpoely86:
@boegel the `psi4 --new-plugin` will generate a makefile that doesn't work. That still needs fixing but there are issues with their build system.

boegel:
Existing PSI easyconfigs are broken because of the changes made here:

```
configure: error: unrecognized option: -DENABLE_MPI=ON
```

wpoely86:
@boegel fixed

boegel:
@wpoely86: still issues, for example with the existing `PSI-4.0b4-ictce-4.1.13-mt.eb`:

```
== 2015-03-01 16:02:39,063 main.EB_PSI WARNING Sanity check: no file of ('lib/libplugin.a',) in ...
...
== 2015-03-01 16:02:39,064 main.EB_PSI WARNING Sanity check: no file of ('include/libplugin/plugin.h',) in ...
...
== 2015-03-01 16:02:39,065 main.EB_PSI WARNING Sanity check: no dir of ('lib',) in ...
```

wpoely86:
@boegel okay, I restored the original sanity check. Should be fine now.

boegel:
Tested with https://github.com/hpcugent/easybuild-easyconfigs/pull/1431 and all existing PSI easyconfigs, works like a charm (now).

Going in, thanks @wpoely86!

PR #546, opened 2015-02-04
In the generic CMake class: always be verbose

wpoely86:


easybuild/easyblocks/generic/cmakemake.py
```diff
@@ -96,6 +96,9 @@ def configure_step(self, srcdir=None, builddir=None):
             if value is not None:
                 options.append("-D%s='%s'" % (option, value))
 
+        # show what CMake is doing by default
+        options.append("-DCMAKE_VERBOSE_MAKEFILE=ON")
```
boegel: maybe only set this if debug logging is enabled?

wpoely86: Why? The logging options already decide over that `run_cmd` should display. And I cannot think of a single reason why we wouldn't want the full output?

boegel: well, bigger logs ;)

but you're right, I see no reason not to include this

others have been doing this for ages, cfr. https://bugzilla.redhat.com/show_bug.cgi?id=474053

boegel:
going in, thanks @wpoely86!

PR #547, opened 2015-02-04
Drop deprecated stuff

wpoely86:
The little brother of hpcugent/easybuild-framework#1143

### DIFF
--- a/easybuild/easyblocks/generic/intelbase.py
+++ b/easybuild/easyblocks/generic/intelbase.py
@@ -94,7 +94,7 @@ class IntelBase(EasyBlock):
         self.setup_local_home_subdir()
         self.clean_home_subdir()
 
-        if self.cfg._config['license'][0] is not None:
+        if 'license' in self.cfg._config:
             _log.nosupport('No old style license parameter, use license_file', '2.0')
 
     @staticmethod
--- a/easybuild/easyblocks/generic/intelbase.py
+++ b/easybuild/easyblocks/generic/intelbase.py
@@ -94,9 +94,6 @@ class IntelBase(EasyBlock):
         self.setup_local_home_subdir()
         self.clean_home_subdir()
 
-        if 'license' in self.cfg._config:
-            _log.nosupport('No old style license parameter, use license_file', '2.0')
-
     @staticmethod
     def extra_options(extra_vars=None):
         extra_vars = EasyBlock.extra_options(extra_vars)
@@ -177,11 +174,7 @@ class IntelBase(EasyBlock):
         if not license_specs:
             self.log.debug("No env var from %s set, trying 'license_file' easyconfig parameter..." % lic_env_vars)
             # obtain license path
-            try:
-                self.license_file = self.cfg['license_file']
-            except:
-                # the default should exist
-                self.log.nosupport('No new style license_file parameter, license_file is now mandatory', '2.0')
+            self.license_file = self.cfg['license_file']
 
             if self.license_file:
                 self.log.info("Using license file %s" % self.license_file)
--- a/easybuild/easyblocks/generic/intelbase.py
+++ b/easybuild/easyblocks/generic/intelbase.py
@@ -42,11 +42,8 @@ import glob
 import easybuild.tools.environment as env
 from easybuild.framework.easyblock import EasyBlock
 from easybuild.framework.easyconfig import CUSTOM
-from easybuild.tools.filetools import rmtree2
 from easybuild.tools.run import run_cmd
 
-# required for deprecated log in static function (ie no self)
-from easybuild.framework.easyconfig.licenses import License
 from vsc import fancylogger
 _log = fancylogger.getLogger('generic.intelbase')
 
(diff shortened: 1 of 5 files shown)


wpoely86:
@boegel will you give it a good look?

easybuild/easyblocks/generic/intelbase.py
```diff
@@ -97,8 +94,8 @@ def __init__(self, *args, **kwargs):
         self.setup_local_home_subdir()
         self.clean_home_subdir()
 
-        if self.cfg._config['license'][0] is not None:
-            _log.deprecated('No old style license parameter, use license_file', '2.0')
+        if 'license' in self.cfg._config:
```
boegel: just `if 'license' in self.cfg` is sufficient now that https://github.com/hpcugent/easybuild-framework/pull/1155 is merged in

easybuild/easyblocks/generic/intelbase.py
```diff
@@ -97,8 +94,8 @@ def __init__(self, *args, **kwargs):
...
+            _log.nosupport('No old style license parameter, use license_file', '2.0')
```
boegel: add `instead` at the end of the error message, use `"` quotes for sentences (just style, no technical reason

easybuild/easyblocks/generic/intelbase.py
boegel: even better: just drop this, since this check is already done in framework (@ https://github.com/hpcugent/easybuild-framework/blob/develop/easybuild/framework/easyconfig/easyconfig.py#L264)

if `license` is defined in the easyconfig, the `EasyConfig` instance will never be created since it'll die in `parse()`, so you'll never even get here

easybuild/easyblocks/generic/intelbase.py
```diff
@@ -187,18 +181,7 @@ def configure_step(self):
...
-            if self.license_file is None:
-                self.log.deprecated('Checking for old style license', '2.0')
-                self.cfg.enable_templating = False
-                lic = self.cfg['license']
-                # old style license is a path (type string)
-                if isinstance(lic, License) and isinstance(lic, str):
-                    self.log.deprecated('No old style license parameter, license has to be pure License subclass', '2.0')
-                    self.license_file = lic
-                self.cfg.enable_templating = True
+                self.log.nosupport('No new style license_file parameter, license_file is now mandatory', '2.0')
```
boegel: change this to:

```python
self.license_file = self.cfg['license_file']
if not self.license_file:
    self.log.error("No license file specified, yet it must be")
```

don't use it in a `try`, since `license_file` will always be available (if not set in the easyconfig, it'll be the default value)

no reason to still use `nosupport`, just log an error if no license file is specified

easybuild/easyblocks/p/perl.py
```diff
@@ -113,5 +113,5 @@ def get_site_suffix(tag):
 def get_sitearch_suffix():
     """Deprecated more specific version of get_site_suffix. Only here for backward compatibility."""
     _log = fancylogger.getLogger('Perl.get_sitearch_suffix', fname=False)
-    _log.deprecated("Use get_site_suffix('sitearch') instead of get_sitearch_suffix()", "2.0")
-    return get_site_suffix('sitearch')
+    _log.nosupport("Use get_site_suffix('sitearch') instead of get_sitearch_suffix()", "2.0")
+    return None
```
boegel: `log.nosupport` is an enhanced `log.error`, so drop the `return`

boegel:
Thank you very much for looking into this @wpoely86!

easybuild/easyblocks/p/perl.py
wpoely86: well, that's C++ habit, the function must have the same signature :wink:

wpoely86:
@boegel done

boegel:
Hmm, where's the "Test PASSed" comment by Jenkins? I hope I didn't kill that by upgrading the plugin just now, that would suck...

boegel:
Jenkins: test this please

wpoely86:
I wouldn't miss those comments :grinning:

boegel:
They're very useful to me, since it's a heads-up that extra commits were included...

wpoely86:
There should be easier and nicer ways of tracking such things as maintainer?

boegel:
Jenkins: test this please

boegel:
@wpoely86: GitHub should notify on pushed commits imho, but fine

boegel:
Jenkins: test this please

boegel:
Looks fine, but I'd like to see at least some of the affected builds tested on top of this to make sure nothing breaks. Are you up for that?

Building & installing Perl/MrBayes and something using CMakeMake on top of `ictce` or `intel` from scratch (e.g. with `--force --installdir=/tmp/$USER/easyblocks-pr-547`) should do it...

You can use `--upload-test-report` which will create a gist that you can then include here.

wpoely86:
@boegel https://gist.github.com/wpoely86/abd0eba3745750165bc7
If I find the time, will try a intel toolchain.

boegel:
Hmm, we need to get this documented at http://easybuild.readthedocs.org/en/latest/Deprecated-functionality.html too...

I'll look into that.

boegel:
Looking great on my end: https://gist.github.com/04a8075a1a5992a46199

I'll look into updating the docs first, since none of this is covered in there yet (doh).

boegel:
PR to update the docs accordingly at https://github.com/hpcugent/easybuild/pull/79

boegel:
Documentation has been updated, and this has been well-tested, so should be good to go in.

Thanks a lot for tackling this @wpoely86!

PR #548, opened 2015-02-04
ant needs Java Hamcrest to build

gjbex:
Ant seems to need Java Hamcrest library in order to build the command line tool.  Coded similarly to JUnit dependency.

### DIFF
--- a/easybuild/easyblocks/a/ant.py
+++ b/easybuild/easyblocks/a/ant.py
@@ -53,15 +53,25 @@ class EB_ant(EasyBlock):
         junit_root = get_software_root('JUnit')
         if not junit_root:
             self.log.error("JUnit module not loaded!")
-
         junit_ver = get_software_version('JUnit')
 
-        # copy JUnit jar to where it's expected
+        hamcrest_root = get_software_root('Java-Hamcrest')
+        if not hamcrest_root:
+            self.log.error("Java-Hamcrest module not loaded!")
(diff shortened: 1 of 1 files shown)


boegel:
Jenkins: ok to test

boegel:
@gjbex: I assume this is a requirement for one of your easyconfig PRs? please indicate which (refer to this PR in the easyconfigs PR)

boegel:
@gjbex: this looks perfect, but it will break existing `ant` easyconfig given the hard requirement

So, we can't merge this unless the existing `ant` easyconfigs are updated to include the `Java-Hamcrest` (build?) dependency

boegel:
@gjbex: let us know whether you're up for adjusting the existing `ant` easyconfigs as well (preferably in a dedicated easyconfigs PR)

gjbex:
Alternative: include version guard for Hamcrest stuff.  My assumption is that this has been build succesfully with previous versions of ant, and that the Hamcrest dependency was either introduced recently, or the Hamcrest code separated out of the ant distribution.

boegel:
@gjbex: that makes a lot of sense... This rang a bell, I've dealt with this before.

See #510, which was left unhandled.

Can you take a look at that, and see whether this helps figuring out the correct approach to take?

boegel:
see also https://github.com/hpcugent/easybuild-easyconfigs/pull/1193

PR #549, opened 2015-02-04
don't use hardcoded path in sanity check, it changes between releases

JensTimmerman:


easybuild/easyblocks/t/trinity.py
```diff
@@ -294,7 +294,7 @@ def install_step(self):
     def sanity_check_step(self):
         """Custom sanity check for Trinity."""
 
-        path = 'trinityrnaseq_r%s' % self.version
+        path = os.path.basename(self.cfg['start_dir'].strip('/'))
```
JensTimmerman: this parsing of the path is copy pasted from below, where it is also used to guess PATH. so it's know to work correctly ;-)

boegel: seems like you stole this from `make_module_req_guess`?

avoid code duplication, introduce a `self.path`, preferably with a more sensible name (and initialise it to `None` first in `__init__`)

easybuild/easyblocks/t/trinity.py
```diff
@@ -54,6 +54,7 @@ def __init__(self, *args, **kwargs):
         EasyBlock.__init__(self, *args, **kwargs)
 
         self.build_in_installdir = True
+        self.path = ''
```
boegel: initialise to `None` (so it's not prematurely used by accident without causing errors), rename to `self.trinity_subdir`

JensTimmerman: well, you either want it initialised with a value that's not going to give errors later on, or you can just skip initialising it, then it will also give errors later on. If you're going to initialise with None you're just making pylint happy, and nothing else, because you will end up getting errors since not is not what you expect to be in there.

boegel: ok, fine, keep it `''` then, but change the name to something less generic as suggested above

boegel:
fix implemented in #802 is already merged, this is obsolete now

PR #550, opened 2015-02-05
add new generic easyblock for installing a bundle of modules

boegel:
this has been talked over a lot, but never got looked into

`Bundle` should be used for anything that's just a (drumroll) bundle of modules, without any additional meaning; examples include the HPCBIOS bundles (e.g., https://github.com/hpcugent/easybuild-easyconfigs/tree/master/easybuild/easyconfigs/h/HPCBIOS_LifeSciences); @fgeorgatos: up for tackling that once this gets merged?

toolchains should still use the `Toolchain` which is now reworked on top of `Bundle`, since it does have additional meaning and I expect the `Toolchain` generic easyblock to do more than just bundle things together in the feature (e.g. potentially defining things like `$CC`, etc., probably with `pushenv`)

### DIFF
--- a/easybuild/easyblocks/generic/bundle.py
+++ b/easybuild/easyblocks/generic/bundle.py
@@ -54,6 +54,9 @@ class Bundle(EasyBlock):
 
     def sanity_check_step(self):
         """
-        Nothing is being installed, so nothing to sanity check on
+        Nothing is being installed, so just being able to load the (fake) module is sufficient
         """
-        pass
+        self.log.info("Testing loading of module '%s' by means of sanity check" % self.full_mod_name)
+        fake_mod_data = self.load_fake_module(purge=True)
+        self.log.debug("Cleaning up after testing loading of module")
+        self.clean_up_fake_module(fake_mod_data)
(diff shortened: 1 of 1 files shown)


boegel:
@wpoely86: please review?

wpoely86:
looks great

boegel:

fgeorgatos:
Just for the record, here are 3 different cases that may look as apparent "bundles" yet different each:

* HPCBIOS_LifeSciences: these should rely on a single toolchain; $CC and co unneeded though
* HPCBIOS_Debuggers: these should NOT rely on any toolchain, ideally should be static builds
* PRACE: this actually may remain a genuine toolchain; it is a bit "fat" stack, yet there is meaning to it.

I recommended to use the hackathon to discuss the different use-cases and understand the merits of possible choices. IMHO, a bundle that does no builds is a whole new case than the above ;-)

boegel:
It doesn't really matter for a bundle whether it requires a particular toolchain (dummy or otherwise).

The point is that this is just a group of modules, nothing more. A toolchain is a very particular sort of group of modules, but as far as just installing the toolchain module itself is concerned, there's nothing really special about it.

PR #552, opened 2015-02-11
Make Score-P easyblock usable for subtoolchains w/o MPI

geimer:
Depends on https://github.com/hpcugent/easybuild-framework/pull/1164 to be usable with toolchains that don't provide MPI.


boegel:
Jenkins: ok to test

boegel:
Using `--with-qt` doesn't seem to work as expected (in terms of replacing `QMAKE=$EBROOTQT/bin/qmake MOC=... UIC=...`), see the failed test report in https://github.com/hpcugent/easybuild-easyconfigs/pull/1574 ...

boegel:
The issues I encountered are fixed with the patch included in https://github.com/hpcugent/easybuild-easyconfigs/pull/1574.

I've retested all builds that use this easyblocks, all is well, so going in.

Thanks @geimer!

PR #558, opened 2015-02-14
one GAMESS-US easyblock to rule them all (combines #470 and #544) (WIP)

boegel:
This easyblock combines the ones in PR #470 by @pescobar and #544 by @scintimandrion, picking the best of both and combining it in one GAMESS-US easyblock to rule them all.

Need a bit more test before I consider this final.

### DIFF
--- a/easybuild/easyblocks/g/gamess_us.py
+++ b/easybuild/easyblocks/g/gamess_us.py
@@ -45,7 +45,7 @@ import tempfile
 import easybuild.tools.toolchain as toolchain
 from easybuild.framework.easyblock import EasyBlock
 from easybuild.framework.easyconfig import CUSTOM, MANDATORY
-from easybuild.tools.filetools import read_file, write_file
+from easybuild.tools.filetools import mkdir, read_file, write_file
 from easybuild.tools.modules import get_software_root, get_software_version
 from easybuild.tools.run import run_cmd, run_cmd_qa
 from easybuild.tools.systemtools import get_platform_name
@@ -219,7 +219,9 @@ class EB_GAMESS_minus_US(EasyBlock):
         """Run GAMESS-US tests (if 'runtest' easyconfig parameter is set to True)."""
         # don't use provided 'runall' script for tests, since that only runs the tests single-core
         if self.cfg['runtest']:
-            testdir = tempfile.mkdtemp()
+            # make sure test dir doesn't contain '[' or ']', csh doesn't handle that well ("set: No match")
+            testdir = re.sub(r'[\[\]]', '-', tempfile.mkdtemp())
+            mkdir(testdir, parents=True)
             try:
                 cwd = os.getcwd()
                 os.chdir(testdir)
@@ -236,7 +238,10 @@ class EB_GAMESS_minus_US(EasyBlock):
             rungms = os.path.join(self.installdir, 'rungms')
             test_env_vars = ['TMPDIR=%s' % testdir]
             if self.toolchain.mpi_family() == toolchain.INTELMPI:
-                test_env_vars.append('I_MPI_FALLBACK=enable')
+                test_env_vars.extend([
+                    'I_MPI_FALLBACK=enable',  # enable fallback in case first fabric fails (see $I_MPI_FABRICS_LIST)
+                    'I_MPI_HYDRA_BOOTSTRAP=fork',  # tests are only run locally (2 processes), so no SSH required
+                ])
 
             # run all exam<id> tests, dump output to exam<id>.log
             n_tests = 47
--- a/easybuild/easyblocks/g/gamess_us.py
+++ b/easybuild/easyblocks/g/gamess_us.py
@@ -35,13 +35,16 @@ EasyBuild support for building and installing GAMESS-US, implemented as an easyb
 @author: Benjamin Roberts (The University of Auckland)
 """
 import fileinput
+import glob
 import os
 import re
+import shutil
 import sys
+import tempfile
 
 from easybuild.framework.easyblock import EasyBlock
 from easybuild.framework.easyconfig import CUSTOM, MANDATORY
-from easybuild.tools.filetools import read_file
+from easybuild.tools.filetools import read_file, write_file
 from easybuild.tools.modules import get_software_root, get_software_version
 from easybuild.tools.run import run_cmd, run_cmd_qa
 from easybuild.tools.systemtools import get_platform_name
@@ -170,10 +173,11 @@ class EB_GAMESS_minus_US(EasyBlock):
 
         # patch hardcoded settings in rungms to use values specified in easyconfig file
         rungms = os.path.join(self.builddir, 'rungms')
+        extra_gmspath_lines = "set ERICFMT=$GMSPATH/auxdata/ericfmt.dat\nset MCPPATH=$GMSPATH/auxdata/MCP\n"
         try:
             for line in fileinput.input(rungms, inplace=1, backup='.orig'):
                 line = re.sub(r"^(\s*set\s*TARGET)=.*", r"\1=%s" % self.cfg['ddi_comm'], line)
-                line = re.sub(r"^(\s*set\s*GMSPATH)=.*", r"\1=%s" % self.installdir, line)
+                line = re.sub(r"^(\s*set\s*GMSPATH)=.*", r"\1=%s\n%s" % (self.installdir, extra_gmspath_lines), line)
                 line = re.sub(r"(null\) set VERNO)=.*", r"\1=%s" % self.version, line)
                 line = re.sub(r"^(\s*set DDI_MPI_CHOICE)=.*", r"\1=%s" % mpilib, line)
                 line = re.sub(r"^(\s*set DDI_MPI_ROOT)=.*%s.*" % mpilib.lower(), r"\1=%s" % mpilib_path, line)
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/g/gamess_us.py
```diff
@@ -0,0 +1,232 @@
...
+        run_cmd_qa(cmd, qa=qa, std_qa=stdqa, log_all=True, simple=True)
+
+        self.log.debug("Contents of install.info:\n%s" % read_file(os.path.join(self.builddir, 'install.info')))
+
+        # patch hardcoded settings in rungms to use values specified in easyconfig file
+        rungms = os.path.join(self.builddir, 'rungms')
+        try:
+            for line in fileinput.input(rungms, inplace=1, backup='.orig'):
+                line = re.sub(r"^(\s*set\s*TARGET)=.*", r"\1=%s" % self.cfg['ddi_comm'], line)
+                line = re.sub(r"^(\s*set\s*GMSPATH)=.*", r"\1=%s" % self.installdir, line)
```
boegel: TODO: also define `ERICFMT ` and `MCPPATH `, see https://github.com/hpcugent/easybuild-easyconfigs/pull/1352/files#diff-e899eb0decac3e9b1edfaf407ba28764R29

easybuild/easyblocks/g/gamess_us.py
```diff
@@ -0,0 +1,232 @@
...
+        run_cmd(compall, log_all=True, simple=True)
+
+        cmd = "%s gamess %s" % (os.path.join(self.cfg['start_dir'], 'lked'), self.version)
+        run_cmd(cmd, log_all=True, simple=True)
+
+    def test_step(self):
+        """Run GAMESS-US tests (if 'runtest' easyconfig parameter is set to True)."""
+        if self.cfg['runtest']:
+            runall_cmd = "%s %s" % (os.path.join(self.builddir, 'runall'), self.version)
+            run_cmd(runall_cmd, path=self.builddir, log_all=True)
```
boegel: TODO: run tests in a `tests` subdir, and check results from tests (check `exam<id>.log` for `EXECUTION OF GAMESS TERMINATED NORMALLY`)

boegel:
@scintimandrion: ping w.r.t. to this PR and how it relates to yours?

scintimandrion:
I'll have a look at it, now that I'm back in the office.

scintimandrion:
Is there any chance of me having push access to this PR?

boegel:
@scintimandrion: if you have any suggestions, just make an inline remarks (via https://github.com/hpcugent/easybuild-easyblocks/pull/558/files), or issue a PR to my branch `boegel:one_gamess_us_to_rule_them_all`

easybuild/easyblocks/g/gamess_us.py
```diff
@@ -0,0 +1,271 @@
...
+
+        # math library config
+        known_mathlibs = ['imkl', 'OpenBLAS', 'ATLAS', 'ACML']
+        mathlib, mathlib_root = None, None
+        for mathlib in known_mathlibs:
+            mathlib_root = get_software_root(mathlib)
+            if mathlib_root is not None:
+                break
+        if mathlib_root is None:
+            self.log.error("None of the know math libraries (%s) available, giving up." % known_mathlibs)
```
scintimandrion: Should read "None of the known math libraries..."

scintimandrion:
Apart from my suggestion for a cosmetic change, this looks good (I think). I'd have to try it to see though.

boegel:
@wpoely86: care to give this a review, maybe even a test? I really want to get this in for EasyBuild v2.0

easyconfigs available at https://github.com/hpcugent/easybuild-easyconfigs/pull/1153 and https://github.com/hpcugent/easybuild-easyconfigs/pull/1406

easybuild/easyblocks/g/gamess_us.py
```diff
@@ -0,0 +1,275 @@
+##
+# Copyright 2009-2014 Ghent University
```
wpoely86: 2015

easybuild/easyblocks/g/gamess_us.py
```diff
@@ -0,0 +1,275 @@
...
+        if mathlib_root is None:
+            self.log.error("None of the known math libraries (%s) available, giving up." % known_mathlibs)
+        if mathlib == 'imkl':
+            mathlib = 'mkl'
+            mathlib_root = os.path.join(mathlib_root, 'mkl')
+        else:
+            mathlib = mathlib.lower()
+
+        # verify selected DDI communication layer
+        known_ddi_comms = ['mpi', 'mixed', 'shmem', 'sockets']
```
wpoely86: make this a constant?

boegel: it's only used here, so there's little point?

boegel:
works like a charm for old/new GAMESS-US versions, when built with `intel/2015a` (see https://github.com/hpcugent/easybuild-easyconfigs/pull/1406)

some care has been taken to also make it work on top of GCC/OpenMPI, but a patch to `rungms` is required to make `rungms` (used in the `test_step`) work on top of OpenMPI... (see also https://github.com/hpcugent/easybuild-easyconfigs/pull/1153)

@pescobar: can you test https://github.com/hpcugent/easybuild-easyconfigs/pull/1406?

pescobar:
when trying to compile this with toolchain intel/2015.02 still uses gcc (this is the debug log)

```
Compiling DDI object: ddi_scatter_acc.o
gcc -DLINUX -m64 -O3 -fstrict-aliasing -I./ -I/opt/soft/apps/impi/5.0.3.048-iccifort-2015.2.164-GCC-4.9.2/include64 -DDDI_MPI -DUSE_SYSV -DMAX_SMP_PROCS=16 -DMAX_NODES=256 -o ddi_scatter_acc.o -c ddi_scatter_acc.c
endif
unset echo
Finished compiling: ddi_scatter_acc.o
 
Compiling FORTRAN wrappers for DDI
gcc -DLINUX -m64 -O3 -fstrict-aliasing -I./ -I/opt/soft/apps/impi/5.0.3.048-iccifort-2015.2.164-GCC-4.9.2/include64 -DDDI_MPI -DUSE_SYSV -DMAX_SMP_PROCS=16 -DMAX_NODES=256 -DINT_SIZE=long -D_UNDERSCORES=1 -o ddi_fortran.o -c ddi_fortran.c
unset echo
Finished compiling: ddi_fortran.o
```

Also during the testing phase it asks for my ssh key?????  This is what I get in the log file
```
== 2015-03-05 13:44:41,422 main.run DEBUG run_cmd: running cmd TMPDIR=/tmp/easybuild-JVe_h_/tmpQkJo7T I_MPI_FALLBACK=enable /home/urz/pub/.local/easybuild/software/GAMESS-US/20130501-R1-intel-2015.02/rungms exam01 20130501-R1 1 2 (in /tmp/easybuild-JVe_h_/tmpQkJo7T)
pub@maia password:
```

This is the content of script rungms which tries to run some mpi code and asks for the ssh key
https://gist.github.com/78be40a7620fce7ebaf0

boegel:
@pescobar: it's normal that's it's using `gcc` for some things, e.g. compiling the DDI part:

```
$ sed -n "685,686p" ddi/compddi 
#       Note that in all cases, we use the gcc compiler from GNU for
#       the compilation of DDI, even if a non-GNU FORTRAN is used.
```

I'm not sure how we can get around the issue that `rungms` is assuming that your SSH setup is such that you can SSH to the host you're on... You can always skip the tests using `runtest = False` in the easyconfig file...

boegel:
@pescobar: the issue you reported with the SSH password being prompted for should be fixed, I'm now setting `$I_MPI_HYDRA_BOOTSTRAP=fork` for the tests (the default is `ssh`)

boegel:
Tests with two versions of GAMESS-US and `intel/2015a` work on two different platforms (Debian and SL6), see https://github.com/hpcugent/easybuild-easyconfigs/pull/1406, so good to go.

Building with `goolf` toolchain also works, but tests need to be skipped since `rungms` doesn't support OpenMPI (yet), see https://github.com/hpcugent/easybuild-easyconfigs/pull/1153.

Going in, thanks for all the effort @scintimandrion and @pescobar, thanks for reviewing/testing @wpoely86!

scintimandrion:
Trying to build GAMESS-US 20130501-R1 with the ictce-5.4.0 toolchain gives this error:

```
== 2015-03-30 14:56:12,927 main.easyblock WARNING build failed (first 300 chars): EasyBuild crashed with an error (at easybuild/tools/run.py:332 in run_cmd_qa): run_cmd_qa: cmd  ./config  : Max nohits 50 reached: end of output tains one or more version number,
enter the specific version you prefer, giving all decimals.
b) If the above output contains paths like 'bin' and 'lib',
e
```

I guess this is because the output of GAMESS-US' `config` script is very verbose between questions?

boegel:
That may indeed be the problem (e.g., things taking longer than on our end due to a slow filesystem), or for some reason a question is being generated that `eb` can't figure out how to solve...

Can you provide a full debug log for this? The error message you've pasted doesn't provide all information to really tell what's going on.

scintimandrion:
I was able to resolve it by changing maxHitCount:

```
--- a/easybuild/tools/run.py
+++ b/easybuild/tools/run.py
@@ -251,7 +251,7 @@ def run_cmd_qa(cmd, qa, no_qa=None, log_ok=True, log_all=False, simple=False, re
     else:
         runLog = None
 
-    maxHitCount = 50
+    maxHitCount = 500
 
     try:
         p = Popen(cmd, shell=True, stdout=PIPE, stderr=STDOUT, stdin=PIPE, close_fds=True, executable="/bin/bash")
```

The other error I'm grappling with is a whole pile of segmentation faults. Still trying to chase that one down.

scintimandrion:
Further to my previous comment, it seems that if I run a test through Easybuild, I get a segmentation fault, while if I run the same command on the command line on the same host, no segfault. I'm still trying to get to the bottom of the problem, but if you have any experience and tips to share where such things have happened before, I'd greatly appreciate some advice.

scintimandrion:
Apparently, this is related to the interplay between EasyBuild and our scheduler (SLURM). There may be workarounds at this end.

PR #563, opened 2015-02-23
add easyblock for Hadoop that supports building native libs

boegel:
@ehiggs: please review?

### DIFF
--- a/easybuild/easyblocks/generic/tarball.py
+++ b/easybuild/easyblocks/generic/tarball.py
@@ -68,4 +68,4 @@ class Tarball(EasyBlock):
             # self.cfg['keepsymlinks'] is False by default except when explicitly put to True in .eb file
             shutil.copytree(src, self.installdir, symlinks=self.cfg['keepsymlinks'])
         except OSError, err:
-            self.log.exception("Copying %s to installation dir %s failed: %s" % (src, self.installdir, err))
+            self.log.error("Copying %s to installation dir %s failed: %s" % (src, self.installdir, err))
--- a/easybuild/easyblocks/generic/tarball.py
+++ b/easybuild/easyblocks/generic/tarball.py
@@ -55,15 +55,17 @@ class Tarball(EasyBlock):
         """
         pass
 
-    def install_step(self):
+    def install_step(self, src=None):
+        """Install by copying from specified source directory (or 'start_dir' if not specified)."""
+        if src is None:
+            src = self.cfg['start_dir']
 
-        src = self.cfg['start_dir']
         # shutil.copytree cannot handle destination dirs that exist already.
         # On the other hand, Python2.4 cannot create entire paths during copytree.
         # Therefore, only the final directory is deleted.
         rmtree2(self.installdir)
         try:
             # self.cfg['keepsymlinks'] is False by default except when explicitly put to True in .eb file
-            shutil.copytree(src,self.installdir, symlinks=self.cfg['keepsymlinks'])
-        except:
-            self.log.exception("Copying %s to installation dir %s failed" % (src,self.installdir))
+            shutil.copytree(src, self.installdir, symlinks=self.cfg['keepsymlinks'])
+        except OSError, err:
+            self.log.exception("Copying %s to installation dir %s failed: %s" % (src, self.installdir, err))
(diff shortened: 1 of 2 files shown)


easybuild/easyblocks/h/hadoop.py
```diff
@@ -0,0 +1,85 @@
...
+
+    def build_step(self):
+        """Custom build procedure for Hadoop: build native libraries, if requested."""
+        if self.cfg['build_native_libs']:
+            cmd = "mvn package -DskipTests -Dmaven.javadoc.skip -Dtar -Pdist,native"
+            run_cmd(cmd, log_all=True, simple=True, log_ok=True)
+
+    def install_step(self):
+        """Custom install procedure for Hadoop: install-by-copy."""
+        if self.cfg['build_native_libs']:
```
wpoely86: doesn't this do actually the same as the `install_step()` from `Tarball`?

boegel: no, `Tarball` copies from a different location

boegel: hmm, I guess I could support specifying the location to copy from in `Tarball.install_sep`

easybuild/easyblocks/h/hadoop.py
```diff
@@ -0,0 +1,85 @@
...
+            'dirs': ['etc', 'libexec'],
+        }
+        super(EB_Hadoop, self).sanity_check_step(custom_paths=custom_paths)
+
+    def make_module_extra(self):
+        """Custom extra module file entries for Hadoop."""
+        txt = super(EB_Hadoop, self).make_module_extra()
+        mapreduce_subdir = os.path.join('share', 'hadoop', 'mapreduce')
+        txt += self.module_generator.prepend_paths('HADOOP_HOME', mapreduce_subdir)
+        txt += self.module_generator.prepend_paths('HADOOP_CONF_DIR', os.path.join(mapreduce_subdir, 'conf'))
```
ehiggs: `HADOOP_CONF_DIR` needs to point to the directory where configs are. AFAICS, for Hadoop installs this is usually in the install path but for installs on HPC systems I can't see how this could work correctly since the configs would need to be configured -and EB doesn't really handle that. So they would either not define it and have users provide it or, if they have a site wide install, install it with a config mgmnt tool.

So we can remove the HADOOP_CONF_DIR part.

easybuild/easyblocks/h/hadoop.py
wpoely86: seems like a good idea?

ehiggs: +1

easybuild/easyblocks/h/hadoop.py
boegel: ok, will remove, thanks

easybuild/easyblocks/h/hadoop.py
boegel: done

easybuild/easyblocks/generic/tarball.py
```diff
@@ -55,15 +55,17 @@ def build_step(self):
         """
         pass
 
-    def install_step(self):
+    def install_step(self, src=None):
```
wpoely86: why not an easyblock option?

easybuild/easyblocks/generic/tarball.py
```diff
@@ -55,15 +55,17 @@ def build_step(self):
...
         # Therefore, only the final directory is deleted.
         rmtree2(self.installdir)
         try:
             # self.cfg['keepsymlinks'] is False by default except when explicitly put to True in .eb file
-            shutil.copytree(src,self.installdir, symlinks=self.cfg['keepsymlinks'])
-        except:
-            self.log.exception("Copying %s to installation dir %s failed" % (src,self.installdir))
+            shutil.copytree(src, self.installdir, symlinks=self.cfg['keepsymlinks'])
+        except OSError, err:
+            self.log.exception("Copying %s to installation dir %s failed: %s" % (src, self.installdir, err))
```
wpoely86: I lost track of the discussion but is this now fatal or not?

boegel: ah, this should be `log.error`, for now

after the 2.0 release, we will deprecate `log.error` and switch to `raise EasyBuildError`

easybuild/easyblocks/generic/tarball.py
boegel: I don't think that makes a lot of sense? usually, no 'source' is involved, and it's just a command to run

wpoely86: I meant an easyconfig parameter but it's fine as it is for me.

boegel:
OK, going in, thanks for the review @wpoely86!

PR #565, opened 2015-03-03
{lang} Preliminary versions of an easyblock for ruby and its gems (REVIEW)

rjeschmi:


easybuild/easyblocks/r/ruby.py
```diff
@@ -0,0 +1,41 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with EasyBuild.  If not, see <http://www.gnu.org/licenses/>.
+##
+"""
+EasyBuild support for Ruby, implemented as an easyblock
+
+@author: Robert Schmidt (OHRI)
```
boegel: better use fully written out affiliation?

easybuild/easyblocks/r/ruby.py
```diff
@@ -0,0 +1,41 @@
...
+"""
+
+from easybuild.easyblocks.generic.configuremake import ConfigureMake
+
+#Seems like the quickest test for whether a gem is installed
+EXTS_FILTER_GEMS = ("gem list %(ext_name)s -i", "")
+
+class EB_Ruby(ConfigureMake):
+    
+    def prepare_for_extensions(self):
```
boegel: docstring pls

easybuild/easyblocks/generic/rubygem.py
```diff
@@ -0,0 +1,89 @@
...
+"""
+
+from easybuild.framework.extensioneasyblock import ExtensionEasyBlock
+
+from easybuild.tools.run import run_cmd
+
+class RubyGem(ExtensionEasyBlock):
+    """ Builds and installs Ruby Gems """
+
+    def __init__(self, *args, **kwargs):
```
boegel: docstring please

easybuild/easyblocks/generic/rubygem.py
```diff
@@ -0,0 +1,89 @@
...
+"""
+EasyBuild support for Ruby Gems, implemented as an easyblock
+
+@author: Robert Schmidt (OHRI)
+"""
+
+from easybuild.framework.extensioneasyblock import ExtensionEasyBlock
+
+from easybuild.tools.run import run_cmd
+
```
boegel: nitpicking: two empty lines above `class`

easybuild/easyblocks/generic/rubygem.py
```diff
@@ -0,0 +1,89 @@
...
+    def __init__(self, *args, **kwargs):
+        super(RubyGem, self).__init__(*args, **kwargs)
+
+    def install_ruby_gem(self):
+        """install ruby gems using gem package manager"""
+        run_cmd('gem install %s' % (self.ext_src))
+
+    def run(self):
+
```
boegel: missing docstring

easybuild/easyblocks/generic/rubygem.py
```diff
@@ -0,0 +1,89 @@
...
+        """install ruby gems using gem package manager"""
+        run_cmd('gem install %s' % (self.ext_src))
+
+    def run(self):
+        
+        if not self.src:
+            self.log.error("No source found for Ruby Gem %s, required for installation. (src: %s)" %
+                           (self.name, self.src))
+        super(RubyGem, self).run()
+        if self.src:
```
boegel: this can never be `False`, giving the condition above?

easybuild/easyblocks/generic/rubygem.py
```diff
@@ -0,0 +1,89 @@
...
+        else:
+            self.log.error("Do we need to install the newest version?")
+
+        self.install_ruby_gem()
+
+    def extract_step(self):
+        """Skip extraction, gemfiles will be installed as downloaded"""
+        pass
+        if len(self.src) > 1:
+            self.log.error("Don't know how to handle R packages with multiple sources.'")
```
boegel: fix log message, this is Ruby, not R

easybuild/easyblocks/generic/rubygem.py
```diff
@@ -0,0 +1,89 @@
...
+        pass
+
+    def test_step(self):
+        """No separate (standard) test procedure for Ruby Gems."""
+        pass
+
+    def install_step(self):
+        """Run install for Ruby Gems"""
+        self.install_ruby_gem()
+
```
boegel: kill empty line at the end

easybuild/easyblocks/r/ruby.py
```diff
@@ -0,0 +1,41 @@
...
+        self.cfg['exts_defaultclass'] = "RubyGem"
+        self.cfg['exts_filter'] = EXTS_FILTER_GEMS
+
```
boegel: kill empty line at the end

easybuild/easyblocks/r/ruby.py
```diff
@@ -0,0 +1,41 @@
...
+##
+"""
+EasyBuild support for Ruby, implemented as an easyblock
+
+@author: Robert Schmidt (OHRI)
+"""
+
+from easybuild.easyblocks.generic.configuremake import ConfigureMake
+
+#Seems like the quickest test for whether a gem is installed
```
boegel: nitpicking: space after `#`

boegel:
Jenkins: ok to test

easybuild/easyblocks/generic/rubygem.py
rjeschmi: I was basing this from R and Perl and there are some slight differences in the interface. It seems like self.src is true, but self.ext_src is empty and needs to be replace. I'll pick at it some more.

easybuild/easyblocks/generic/rubygem.py
rjeschmi: I don't really know if this is supposed to be defined since it is only a call to the super, but it is possible we will need it in the future.

boegel: Ah, no, no need to have it here to only call `super`, so drop it (it can re-added if needed later).

boegel:
well-tested with https://github.com/hpcugent/easybuild-easyconfigs/pull/1440, so good to go in, thanks @rjeschmi!

PR #569, opened 2015-03-05
GROMACS-3.3.3

valtandor:
Updates to the existing "gromacs.py" to support GROMACS 3.3.3. Imperfect and untested support for other pre-4.6 versions is also included.


scintimandrion:
Yes, if there's a way to test an EasyBlock.

boegel:
@wpoely86: please review?

valtandor:
@boegel: Please let me know if you're not the right person to be handling this PR. I've put in some updates to reflect changes to the GROMACS installation procedure in version 5.1. The problem is that I don't know how backwards compatible they are. In most cases I've assumed that they will work with all versions 4.6 and later (when CMake was introduced) but the PR may need to be substantially modified to ensure compatibility with versions between 4.6 and 5.0.

boegel:
@valtandor I've merged this PR with two others for the GROMACS easyblock, cfr. #960 

I've already cleaned up a couple of things, but need to do some further testing to make sure it's ready to go without breaking existing easyconfigs & any GROMACS easyconfig PRs...

valtandor:
@boegel No problem. Would you like me to close this PR, or wait until you've finished your testing?

boegel:
@valtandor This PR will be closed automagically once I merge in #960, since the commits in here are included there too, I did a 'proper' merge of all three PRs (& resolved conflicts that occurred).

easybuild/easyblocks/g/gromacs.py
```diff
@@ -161,10 +234,41 @@ def install_step(self):
...
+                cmake_objdir = "build-mdrun-only"
+                os.mkdir(cmake_objdir)
+                os.chdir(cmake_objdir)
+                self.cfg['configopts'] = re.sub(r'-DGMX_MPI=OFF', r'', self.cfg['configopts'])
+                if self.cfg['mpi_numprocs'] == 0:
+                    self.log.info("No specific number of test MPI tasks requested -- using parallelism ({0})".format(self.cfg['parallel']))
+                    self.cfg['mpi_numprocs'] = self.cfg['parallel']
+                elif self.cfg['mpi_numprocs'] > self.cfg['parallel']:
+                    self.log.warning("Number of test MPI tasks ({0}) is greater than parallelism ({1})".format(self.cfg['mpi_numprocs'], self.cfg['parallel']))
+                self.cfg.update('configopts', "-DGMX_MPI=ON -DGMX_THREAD_MPI=OFF -DMPIEXEC={0} -DMPIEXEC_NUMPROC_FLAG={1} -DNUMPROC={2} -DGMX_BUILD_MDRUN_ONLY=ON".format(self.cfg['mpiexec'], self.cfg['mpiexec_numproc_flag'], self.cfg['mpi_numprocs']))
```
boegel: @valtandor why did you include `-DGMX_BUILD_MDRUN_ONLY=ON` here (and also add `mdrun` to `buildopts` above for older GROMACS versions)?

I had to get rid of this in #960 because I was resulting in an incomplete installation w.r.t. MPI, e.g. `libgromacs_mpi.a` was missing, etc.

It seems fine to just drop this, since the MPI-able binaries & libraries are suffixed with `_mpi`, so nothing is overwritten?

valtandor: @boegel Well, that's silly. I thought there was some reason, but I guess if it works without that option and doesn't overwrite anything serial, then there's no argument from me about not having the option.

boegel:
merged via #960, thanks @valtandor!

PR #570, opened 2015-03-06
Update disabling sanitizer tests for clang 3.6

wpoely86:


easybuild/easyblocks/c/clang.py
```diff
@@ -197,35 +198,47 @@ def configure_step(self):
...
-            "projects/compiler-rt/lib/dfsan/CMakeLists.txt",
-            "projects/compiler-rt/lib/lsan/CMakeLists.txt",
-            "projects/compiler-rt/lib/msan/CMakeLists.txt",
-            "projects/compiler-rt/lib/tsan/CMakeLists.txt",
-            "projects/compiler-rt/lib/ubsan/CMakeLists.txt",
-        ]
-
-        for patchfile in patchfiles:
+        if LooseVersion(self.version) < LooseVersion('3.6'):
+            patchfiles = [
```
boegel: please add some comments do describe what is being patched here, exactly

easybuild/easyblocks/c/clang.py
```diff
@@ -197,35 +198,47 @@ def configure_step(self):
...
+                "projects/compiler-rt/lib/asan/CMakeLists.txt",
+                "projects/compiler-rt/lib/dfsan/CMakeLists.txt",
+                "projects/compiler-rt/lib/lsan/CMakeLists.txt",
+                "projects/compiler-rt/lib/msan/CMakeLists.txt",
+                "projects/compiler-rt/lib/tsan/CMakeLists.txt",
+                "projects/compiler-rt/lib/ubsan/CMakeLists.txt",
```
boegel: flesh out the common part `projects/compiler-rt`, throw that in the `os.path.join` below`

boegel: and `CMakeLists.txt` too, so only list dirs like `lib/asan` in `cmakelists_to_patch`

easybuild/easyblocks/c/clang.py
```diff
@@ -197,35 +198,47 @@ def configure_step(self):
...
+                    try:
+                        for line in fileinput.input(patchfile_fp, inplace=1, backup='.orig'):
+                            if "add_subdirectory(lit_tests)" not in line:
+                                sys.stdout.write(line)
+                    except (IOError, OSError), err:
+                        self.log.error("Failed to patch %s: %s" % (patchfile_fp, err))
+                else:
+                    self.log.debug("Not patching non-existent %s in %s" % (patchfile, self.llvm_src_dir))
+
+            patchfile = "projects/compiler-rt/lib/sanitizer_common/CMakeLists.txt"
```
boegel: pls add a comment explaining what's being patched here

easybuild/easyblocks/c/clang.py
```diff
@@ -197,35 +198,47 @@ def configure_step(self):
...
+            try:
+                for line in fileinput.input("%s/%s" % (self.llvm_src_dir, patchfile), inplace=1, backup='.orig'):
+                    if "add_subdirectory(tests)" not in line:
+                        sys.stdout.write(line)
+            except IOError, err:
+                self.log.error("Failed to patch %s/%s: %s" % (self.llvm_src_dir, patchfile, err))
+        else:
+            patchfile = "projects/compiler-rt/test/CMakeLists.txt"
```
boegel: add a comment what is being patched exactly

easybuild/easyblocks/c/clang.py
```diff
@@ -196,36 +197,54 @@ def configure_step(self):
...
-            "projects/compiler-rt/lib/msan/CMakeLists.txt",
-            "projects/compiler-rt/lib/tsan/CMakeLists.txt",
-            "projects/compiler-rt/lib/ubsan/CMakeLists.txt",
-        ]
-
-        for patchfile in patchfiles:
+        """Disable the tests of all the sanitizers by removing the test directories from the build system"""
+        if LooseVersion(self.version) < LooseVersion('3.6'):
+        # for Clang 3.5 and lower, the tests are scattered over several CMakeLists.
+        # We loop over them, and patch out the rule that adds the sanitizers tests to the testsuite
```
boegel: fix indent pls

boegel:
well-tested with https://github.com/hpcugent/easybuild-easyconfigs/pull/1389, and existing Clang easyconfigs

There is some weird issue w.r.t. the condition for calling `disable_sanitizer_tests` tests though. Having `ulimit -v` set to `unlimited` doesn't mean the tests will pass...

@wpoely86: can you open an issue on that, to track it down further?

I don't consider that a blocker for this PR, just something we need to look into to make things more robust.

So, going in! Thanks @wpoely86!

PR #573, opened 2015-03-06
fix easyblock for EasyBuild to support upgrade path to EasyBuild v2.0

boegel:
The easyblock used for installing EasyBuild included in EasyBuild v1.16.1 is unaware of the extra `vsc-base` component in the upcoming EasyBuild v2.0.0. This PR fixes that issue, supporting an intermediate release EasyBuild v1.16.2 which does support upgrading to EasyBuild v2.0.0.

### DIFF
--- a/easybuild/easyblocks/e/easybuildmeta.py
+++ b/easybuild/easyblocks/e/easybuildmeta.py
@@ -69,14 +69,16 @@ class EB_EasyBuildMeta(PythonPackage):
 
         try:
             subdirs = os.listdir(self.builddir)
-            for pkg in ['easyconfigs', 'easyblocks', 'framework']:
-                seldirs = [x for x in subdirs if x.startswith('easybuild-%s-' % pkg)]
+            for pkg in ['vsc-base', 'easybuild-framework', 'easybuild-easyblocks', 'easybuild-easyconfigs']:
+                seldirs = [x for x in subdirs if x.startswith(pkg)]
                 if not len(seldirs) == 1:
-                    self.log.error("Failed to find EasyBuild %s package (subdirs: %s, seldirs: %s)" % (pkg, subdirs, seldirs))
-
-                self.log.debug("Installing EasyBuild package %s" % pkg)
-                os.chdir(os.path.join(self.builddir, seldirs[0]))
-                super(EB_EasyBuildMeta, self).install_step()
+                    if pkg != 'vsc-base':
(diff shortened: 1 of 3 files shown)


boegel:
Jenkins: test this please

rjeschmi:
looks fine, I'll keep looking at it though..

boegel:
Jenkins: test this please

easybuild/easyblocks/e/easybuildmeta.py
```diff
@@ -69,14 +69,16 @@ def install_step(self):
...
-                seldirs = [x for x in subdirs if x.startswith('easybuild-%s-' % pkg)]
+            for pkg in ['vsc-base', 'easybuild-framework', 'easybuild-easyblocks', 'easybuild-easyconfigs']:
+                seldirs = [x for x in subdirs if x.startswith(pkg)]
                 if not len(seldirs) == 1:
-                    self.log.error("Failed to find EasyBuild %s package (subdirs: %s, seldirs: %s)" % (pkg, subdirs, seldirs))
-
-                self.log.debug("Installing EasyBuild package %s" % pkg)
-                os.chdir(os.path.join(self.builddir, seldirs[0]))
-                super(EB_EasyBuildMeta, self).install_step()
+                    if pkg != 'vsc-base':
```
wpoely86: I don't understand this one? If there is only 1 element and it's not `vsc-base`, we give an error and if it's `vsc-base` we do nothing?

boegel: Yes, we basically don't care if `vsc-base` is not found. If it's needed (EB v2.x), it will be picked up when installing framework.

boegel: Note that this is handled better in the current version of the `EasyBuildMeta` easyblock.

This patch is just to release EB v1.16.2, which is required to install EasyBuild v2.0.0 'offline' using EasyBuild (EB v1.16.1 doesn't work).

wpoely86: maybe add a comment about that?

boegel: Well, this version of the EasyBuildMeta easyblock is dead. The currently active version in `develop` is the one last tweaked in https://github.com/hpcugent/easybuild-easyblocks/pull/567

PR #576, opened 2015-03-10
fix definition of $MKLROOT, should be set to 'mkl' subdir of install dir

boegel:
bug mentioned by @klust

in practice, this doesn't make too much of a difference for use of `eb` itself (w.r.t. use of `$MKLROOT` in easyblocks/easyconfigs), but the path is incorrect w.r.t. MKL Link Advisor


boegel:
@wpoely86: thoughts on this?

wpoely86:
Well, I hope it doesn't break anything. It shouldn't but we've seen stranger shit.

boegel:
breaking stuff early in `dev` is fine ;)

I'll take a look where `$MKLROOT` is being (obviously) used, and retest those builds on top of this.

boegel:
merged via #590

PR #578, opened 2015-03-12
add easyblock for building and installing TINKER

boegel:


JensTimmerman:
looks ok to me

boegel:
broken test reported via https://github.com/jayponder/tinker/issues/7

easybuild/easyblocks/t/tinker.py
```diff
@@ -0,0 +1,154 @@
...
+        if comp_dir is None:
+            self.log.error("Failed to determine compiler directory for %s (known: %s)" % (comp_fam, comp_dirs))
+
+        self.build_subdir = os.path.join(os_dir, comp_dir)
+        self.log.info("Using build scripts from %s subdirectory" % self.build_subdir)
+
+        # patch 'link.make' script to use FFTW provided via EasyBuild
+        link_make_fp = os.path.join(self.cfg['start_dir'], self.build_subdir, 'link.make')
+        for line in fileinput.input(link_make_fp, inplace=1, backup='.orig'):
+            line = re.sub(r"libfftw3_threads.a libfftw3.a", r"-L$EBROOTFFTW/lib -lfftw3_threads -lfftw3", line)
```
wpoely86: it's unclear to me, but if this a makefile you're patching, you need `$(EBROOTFFTW)`

easybuild/easyblocks/t/tinker.py
```diff
@@ -0,0 +1,154 @@
...
+                os.chdir(testdir)
+            except OSError, err:
+                self.log.error("Failed to move to %s to run tests: %s" % (testdir, err))
+
+            # run all tests via the provided 'run' scripts
+            tests = glob.glob(os.path.join(testdir, '*.run'))
+            # gpcr takes too logn (~1h), ifabp fails due to input issues (?)
+            tests = [t for t in tests if not (t.endswith('gpcr.run') or t.endswith('ifabp.run'))]
+            for test in tests:
+                run_cmd(test)
```
wpoely86: `simple=True`?

easybuild/easyblocks/t/tinker.py
```diff
@@ -0,0 +1,154 @@
...
+            'files': ['tinker/source/libtinker.a'],
+            'dirs': ['tinker/bin'],
+        }
+        super(EB_TINKER, self).sanity_check_step(custom_paths=custom_paths)
+
+    def make_module_req_guess(self):
+        """Custom guesses for module file prepend-path statements."""
+        guesses = super(EB_TINKER, self).make_module_req_guess()
+        guesses['PATH'].append(os.path.join('tinker', 'bin'))
+        guesses['LIBRARY_PATH'].append(os.path.join('tinker', 'source'))
```
wpoely86: no `LD_LIBRARY_PATH`?

easybuild/easyblocks/t/tinker.py
boegel: it's not a `Makefile`, it's a script, basically

see https://github.com/jayponder/tinker/blob/master/linux/intel/link.make

easybuild/easyblocks/t/tinker.py
boegel: doesn't really matter, does it?

it checks the exit code, and nicely logs the output, so the defaults are fine

easybuild/easyblocks/t/tinker.py
boegel: only a static library, so setting `$LD_LIBRARY_PATH` is useless

boegel:
Thanks for the review @JensTimmerman & @wpoely86!

boegel:
Going in, thanks again for reviewing!

boegel:
I was putting this on hold until I had a working test report, which is now there, see https://github.com/hpcugent/easybuild-easyconfigs/pull/1465#issuecomment-78518164.

PR #579, opened 2015-03-12
Allow support for MPICH/MPICH2

ocaisa:


easybuild/easyblocks/c/cp2k.py
```diff
@@ -272,7 +272,7 @@ def configure_common(self):
             regflags = 'NOOPT'
 
         # make sure a MPI-2 able MPI lib is used
-        mpi2libs = ['impi', 'MVAPICH2', 'OpenMPI']
+        mpi2libs = ['impi', 'MVAPICH2', 'OpenMPI', 'MPICH2', 'MPICH']
```
boegel: looks trivial, but.. .:-)

We should probably be using the `mpi_family()` function here, rather than relying on `get_software_root`.

@ocaisa: up for looking into that?

Also: do you have a matching easyconfig to test this change? It's pretty clear that this shouldn't break anything that works already.

boegel:
Jenkins: ok to test

ocaisa:
Annoyingly I don't have a verified matching easyconfig, I can't get it build at all on JuropaTest because during the regtest it tries to use mpirun (which is not supported, only srun) so it just hangs. I tried getting it to skip tests but it refuses to

easybuild/easyblocks/c/cp2k.py
```diff
@@ -272,14 +272,30 @@ def configure_common(self):
...
+            mpi_fam = self.toolchain.mpi_family()
+            mpi_spec = mpi_spec_by_fam.get(mpi_fam)
+            if mpi_spec is not None:
                 mpi2 = True
-            else:
-                self.log.debug("MPI-2 supporting MPI library %s not loaded.")
-
+            self.log.debug("Determined MPI specification based on MPI toolchain component: %s" % mpi_spec)
+        else:
+            # can't use toolchain.mpi_family, because of dummy toolchain
```
boegel: hmm, does it make sense to have that in here?

boegel:
@ocaisa: just including 'runtest = False' in the CP2K easyconfig should take care of skipping the `test_step`?

boegel:
Ah, no, the CP2K easyblock includes this:

```python
# always enable testing for CP2K
self.cfg['runtest'] = True
```

That's just wrong... The `runtest` should be enabled by default, like we do for WRF:

```python
def extra_options():
            ...
            'runtest': [True, "Build and run WRF tests", CUSTOM],
            ...
```

Feel free to tackle that too in here, we shouldn't be hard enabling the tests like that.

ocaisa:
I'll give it a whirl now and see how it goes

easybuild/easyblocks/c/cp2k.py
```diff
@@ -272,14 +270,32 @@ def configure_common(self):
...
+        if hasattr(self.toolchain, 'MPI_FAMILY') and self.toolchain.MPI_FAMILY is not None:
+            mpi_spec_by_fam = {
+                toolchain.MPICH: 'mpi2',  # MPICH is MPICH v3.x, which is MPI2 compatible
+                toolchain.MPICH2: 'mpi2',
+                toolchain.MVAPICH2: 'mpi2',
+                toolchain.OPENMPI: 'mpi2',
+                toolchain.IMPI: 'mpi2',
+            }
+            mpi_fam = self.toolchain.mpi_family()
+            mpi_spec = mpi_spec_by_fam.get(mpi_fam)
```
boegel: since the value for `mpi_spec` is the same regardless of which toolchain it is, this can be cleaned up a bit to:

```python
known_mpi2_fams = [toolchain.MPICH, toolchain.MPICH2, toolchain.MVAPICH2, toolchan.OPENMPI, toolchain.IMPI]
if self.toolchain.mpi_family() in known_mpi2_fams:
    mpi2 = True
```

boegel:
Tested via https://github.com/hpcugent/easybuild-easyconfigs/pull/1471, looking good, so going in.

Thanks @ocaisa!

PR #580, opened 2015-03-17
Add the ability to make parallel builds.

ehiggs:


boegel:
Looks great, let's get this tested along with https://github.com/hpcugent/easybuild-easyconfigs/pull/1488 (maybe clean up the remarks there first?).

easybuild/easyblocks/h/hadoop.py
```diff
@@ -43,6 +43,7 @@ def extra_options():
         """Custom easyconfig parameters for Hadoop."""
         extra_vars = {
             'build_native_libs': [False, "Build native libraries", CUSTOM],
+            'parallel': [1, "Build in parallel with more threads", CUSTOM],
```
boegel: Why fix this to `1` here? `parallel` can be set in all easyconfigs, just picking it up in the easyblock is sufficient

boegel:
tested with https://github.com/hpcugent/easybuild-easyconfigs/pull/1488, looks fine, so going in, thanks @ehiggs

PR #581, opened 2015-03-17
{viz} Xmipp (REVIEW)

JensTimmerman:
requires https://github.com/hpcugent/easybuild-framework/pull/1229

### DIFF
--- a/easybuild/easyblocks/x/xmipp.py
+++ b/easybuild/easyblocks/x/xmipp.py
@@ -76,7 +76,6 @@ class EB_Xmipp(EasyBlock):
             external_path = os.path.join(os.getcwd(), 'external')
             extract_file(os.path.join(external_path, src), external_path)
 
-        # TODO: add support for java and gtest (and others?)
         cmd = ('{preconfigopts} PATH=$PWD:$PATH xmipp_python external/scons/scons.py mode=configure -j {parallel}'
                ' --config=force  profile=no fast=yes warn=no release=yes gtest=no cuda=no debug=no matlab=no'
                ' java=no LINKERFORPROGRAMS="$CXX" MPI_BINDIR="$EBROOTIMPI" JAVA_HOME="$JAVA_HOME" JAVAC=javac CC="$CC"'
@@ -94,14 +93,6 @@ class EB_Xmipp(EasyBlock):
               'mode=compile -j {parallel}'.format(parallel=self.cfg['parallel'])
 
         run_cmd(cmd, log_all=True, simple=True)
-
-    def test_step(self):
-        """
-        Run Xmipp tests (if 'runtest' easyconfig parameter is set to True).
-        TODO: patch for gtest includes first
-        """
-        # TODO: implement
-        super(self.__class__, self).test_step(self)
 
     def install_step(self):
         """install step for xmipp, this builds a local database and seems to do some tests?"""
--- a/easybuild/easyblocks/x/xmipp.py
+++ b/easybuild/easyblocks/x/xmipp.py
@@ -139,9 +139,9 @@ class EB_Xmipp(EasyBlock):
             python_dynlib_dir,
         ]
         cmd = ' '.join([
-            'XMIPP_HOME={cwd}',
+            'XMIPP_HOME=%s' % self.cfg['start_dir'],
             'PATH=%s:$PATH' % os.path.join(self.cfg['start_dir'], 'bin'),
-            'PYTHONPATH="%s"' % os.pathsep.join(['$PYTHONPATH'] + [extra_pythonpaths]),
+            'PYTHONPATH="%s"' % os.pathsep.join(['$PYTHONPATH'] + extra_pythonpaths),
             'python setup.py install',
         ])
         run_cmd(cmd, log_all=True, simple=True)
--- a/easybuild/easyblocks/x/xmipp.py
+++ b/easybuild/easyblocks/x/xmipp.py
@@ -1,5 +1,5 @@
 ##
-# Copyright 2009-2015 Ghent University
+# Copyright 2015-2015 Ghent University
 #
 # This file is part of EasyBuild,
 # originally created by the HPC team of Ghent University (http://ugent.be/hpc/en),
@@ -35,82 +35,118 @@ from easybuild.tools.filetools import mkdir, extract_file
 from easybuild.tools.modules import get_software_root, get_software_version
 from easybuild.tools.run import run_cmd
 
+from easybuild.easyblocks.generic.pythonpackage import det_pylibdir
+
 
 class EB_Xmipp(EasyBlock):
     """Support for building/installing Xmipp."""
 
     def __init__(self, *args, **kwargs):
         """Easyblock constructor, enable building in installation directory."""
-        super(self.__class__, self).__init__(*args, **kwargs)
+        super(EB_Xmipp, self).__init__(*args, **kwargs)
         self.build_in_installdir = True
 
     def extract_step(self):
         """Extract sources."""
         # strip off 'xmipp' part to avoid having everything in a 'xmipp' subdirectory
-        self.cfg['unpack_options'] = "--strip-components=1"
-        super(self.__class__, self).extract_step()
-
-    def configure_step(self):
-        """Configure xmipp build via a provided wrapper around sconss."""
-        # install step expects these to exist
-        mkdir('lib')
-        mkdir('bin')
-        # run interactive 'config' script to generate install.info file
-        try:
-            # ln -s `which python` xmipp_python
-            orig = os.path.join(get_software_root('Python'), 'bin', 'python')
-            dest = 'xmipp_python'
-            os.symlink(orig, dest)
-            # ln -s $EBROOTPYTHON/lib/python2.7 lib/python2.7
-            self.python_short_ver = ".".join(get_software_version('Python').split(".")[0:2])
-            orig = os.path.join(get_software_root('Python'), 'lib', 'python%s' % self.python_short_ver)
-            dest = os.path.join('lib', 'python%s' % self.python_short_ver)
-            os.symlink(orig, dest)
-        except OSError as err:
-            self.log.error("Failed to symlink %s to %s: %s" % (orig, dest, err))
+        self.cfg.update('unpack_options', '--strip-components=1')
+        super(EB_Xmipp, self).extract_step()
 
         # extract some dependencies that we really need and can't find anywhere else
         # alglib tarball has version in name, so lets find it with a glob
         alglib_tar = glob.glob(os.path.join(os.getcwd(), 'external', 'alglib*.tgz'))[0]
         for src in ['bilib.tgz', 'bilib.tgz', 'condor.tgz', alglib_tar]:
-            external_path = os.path.join(os.getcwd(), 'external')
+            external_path = os.path.join(self.cfg['start_dir'], 'external')
             extract_file(os.path.join(external_path, src), external_path)
 
-        cmd = ('{preconfigopts} PATH=$PWD:$PATH xmipp_python external/scons/scons.py mode=configure -j {parallel}'
-               ' --config=force  profile=no fast=yes warn=no release=yes gtest=no cuda=no debug=no matlab=no'
-               ' java=no LINKERFORPROGRAMS="$CXX" MPI_BINDIR="$EBROOTIMPI" JAVA_HOME="$JAVA_HOME" JAVAC=javac CC="$CC"'
-               ' CXXFLAGS="$CXXFLAGS -DMPICH_IGNORE_CXX_SEEK -I$EBROOTPYTHON/include/python{short_python_ver}"'
-               ' CXX="$CXX" MPI_CC="$MPICC" MPI_CXX="$MPICXX" MPI_INCLUDE="$MPI_INC_DIR" MPI_LIBDIR="$MPI_LIB_DIR"'
-               ' MPI_LINKERFORPROGRAMS="$MPICC" LIBPATH="$LD_LIBRARY_PATH" {configopts}').format(
-                   parallel=self.cfg['parallel'], short_python_ver=self.python_short_ver,
-                   preconfigopts=self.cfg['preconfigopts'], configopts=self.cfg['configopts']
-                   )
+    def configure_step(self):
+        """Configure xmipp build via a provided wrapper around sconss."""
+        # check if all our dependencies are in place
+        python_root = get_software_root('Python')
+        if not python_root:
+            self.log.error("Python not loaded as a dependency, which is required for %s" % self.name)
+        python_libdir = det_pylibdir()
+        self.python_short_ver = ".".join(get_software_version('Python').split(".")[0:2])
+        java_root = get_software_root('Python')
+        if not java_root:
+            self.log.error("Java not loaded as a dependency, which is required for %s" % self.name)
+
+        # build step expects these to exist
+        mkdir(os.path.join(self.cfg['start_dir'], 'bin'))
+        mkdir(os.path.join(self.cfg['start_dir'], 'lib'))
+
+        cmd = ' '.join([
+            '{preconfigopts}',
+            'python external/scons/scons.py mode=configure -j {parallel}',
+            '--config=force',
+            'profile=no',
+            'fast=yes',
+            'warn=no',
+            'release=yes',
+            'gtest=no',
+            'cuda=no',
+            'debug=no',
+            'matlab=no',
+            'java=no',
+            'LINKERFORPROGRAMS="$CXX"',
+            'MPI_BINDIR="{mpi_bindir}"',
+            'JAVA_HOME="{java_home}"',
+            'JAVAC=javac',
+            'CC="$CC"',
+            'CXXFLAGS="$CXXFLAGS -DMPICH_IGNORE_CXX_SEEK -I$EBROOTPYTHON/include/python{short_python_ver}" \
+                      "-I{python_libdir}/numpy/core/include/"',
+            'CXX="$CXX"',
+            'MPI_CC="$MPICC"',
+            'MPI_CXX="$MPICXX"',
+            'MPI_INCLUDE="$MPI_INC_DIR"',
+            'MPI_LIBDIR="$MPI_LIB_DIR"',
+            'MPI_LINKERFORPROGRAMS="$MPICC"',
+            'LIBPATH="$LD_LIBRARY_PATH"',
+            '{configopts}',
+        ])
+        cmd.format(
+            parallel=self.cfg['parallel'],
+            short_python_ver=self.python_short_ver,
+            python_libdir=python_libdir,
+            preconfigopts=self.cfg['preconfigopts'],
+            configopts=self.cfg['configopts'],
+            mpi_bindir=os.path.join(get_software_root(self.toolchain.MPI_MODULE_NAME[0]), 'bin'),
+            java_home=java_root
+        )
         run_cmd(cmd, log_all=True, simple=True)
 
     def build_step(self):
         """Custom build procedure for Xmipp: call the scons wrapper with compile argument"""
-        cmd = 'LD_LIBRARY_PATH=$LD_LIBRARY_PATH:$PWD/lib PATH=$PWD:$PATH xmipp_python external/scons/scons.py' \
-              'mode=compile -j {parallel}'.format(parallel=self.cfg['parallel'])
-
+        cmd = ' '.join([
+            'LD_LIBRARY_PATH=$LD_LIBRARY_PATH:$PWD/lib',
+            'python external/scons/scons.py',
+            'mode=compile -j {parallel}',
+        ])
+        cmd.format(parallel=self.cfg['parallel'])
         run_cmd(cmd, log_all=True, simple=True)
 
     def install_step(self):
         """install step for xmipp, this builds a local database and seems to do some tests?"""
-        cmd = 'XMIPP_HOME=$PWD PATH=$PWD:$PWD/bin:$PATH PYTHONPATH=$PYTHONPATH:$PWD/protocols:' \
-              '$PWD/libraries/bindings/python/:$EBROOTPYTHON/lib/python{short_python_ver}/lib-dynload/' \
-              ' python setup.py install'.format(self.python_short_ver)
+        cmd = ' '.join([
+            'XMIPP_HOME=$PWD',
+            'PATH=$PWD/bin:$PATH',
+            'PYTHONPATH="$PYTHONPATH:$PWD/protocols:$PWD/libraries/bindings/python/:' \
+            '$EBROOTPYTHON/lib/python{short_python_ver}/lib-dynload/"',
+            'python setup.py install'
+        ])
+        cmd.format(short_python_ver=self.python_short_ver)
         run_cmd(cmd, log_all=True, simple=True)
 
     def sanity_check_step(self):
-        """Custom sanity check for GAMESS-US."""
+        """Custom sanity check for Xmipp."""
         custom_paths = {
             'files': ['xmipp_%s' % x for x in ['imagej', 'mpi_run', 'phantom_create', 'tomo_project', 'volume_align']],
-            'dirs': ['lib/'],
+            'dirs': ['lib'],
         }
-        super(self.__class__, self).sanity_check_step(custom_paths=custom_paths)
+        super(EB_Xmipp, self).sanity_check_step(custom_paths=custom_paths)
 
     def make_module_extra(self):
-        """Define GAMESS-US specific variables in generated module file, i.e. $GAMESSUSROOT."""
-        txt = super(self.__class__, self).make_module_extra()
+        """Define Xmipp specific variables in generated module file, i.e. XMIPP_HOME."""
+        txt = super(EB_Xmipp, self).make_module_extra()
         txt += self.module_generator.set_environment('XMIPP_HOME', '$root')
         return txt
--- a/easybuild/easyblocks/x/xmipp.py
+++ b/easybuild/easyblocks/x/xmipp.py
@@ -82,7 +82,9 @@ class EB_Xmipp(EasyBlock):
 
         cmd = ' '.join([
             '{preconfigopts}',
-            'python external/scons/scons.py mode=configure -j {parallel}',
+            'python external/scons/scons.py',
+            'mode=configure',
+            '-j {parallel}',
             '--config=force',
             'profile=no',
             'fast=yes',
@@ -98,8 +100,12 @@ class EB_Xmipp(EasyBlock):
             'JAVA_HOME="{java_home}"',
             'JAVAC=javac',
             'CC="$CC"',
-            'CXXFLAGS="$CXXFLAGS -DMPICH_IGNORE_CXX_SEEK -I$EBROOTPYTHON/include/python{short_python_ver} \
-                      -I$EBROOTPYTHON/{python_libdir}/numpy/core/include/"',
+            'CXXFLAGS="%s"' % ' '.join(
+                '$CXXFLAGS',
+                '-DMPICH_IGNORE_CXX_SEEK',
+                '-I$EBROOTPYTHON/include/python{short_python_ver}',
+                '-I$EBROOTPYTHON/{python_libdir}/numpy/core/include/',
+            ),
             'CXX="$CXX"',
             'MPI_CC="$MPICC"',
             'MPI_CXX="$MPICXX"',
@@ -108,8 +114,7 @@ class EB_Xmipp(EasyBlock):
             'MPI_LINKERFORPROGRAMS="$MPICC"',
             'LIBPATH="$LD_LIBRARY_PATH"',
             '{configopts}',
-        ])
-        cmd = cmd.format(
+        ]).format(
             parallel=self.cfg['parallel'],
             short_python_ver=self.python_short_ver,
             python_libdir=python_libdir,
@@ -125,21 +130,22 @@ class EB_Xmipp(EasyBlock):
         cmd = ' '.join([
             'LD_LIBRARY_PATH=$LD_LIBRARY_PATH:$PWD/lib',
             'python external/scons/scons.py',
-            'mode=compile -j {parallel}',
+            'mode=compile',
+            '-j %s' % self.cfg['parallel'],
         ])
-        cmd = cmd.format(parallel=self.cfg['parallel'])
         run_cmd(cmd, log_all=True, simple=True)
 
     def install_step(self):
         """install step for xmipp, this builds a local database and seems to do some tests?"""
+        python_dynlib_dir = '$EBROOTPYTHON/lib/python{short_python_ver}/lib-dynload/"'
+
         cmd = ' '.join([
             'XMIPP_HOME=$PWD',
             'PATH=$PWD/bin:$PATH',
-            'PYTHONPATH="$PYTHONPATH:$PWD/protocols:$PWD/libraries/bindings/python/:' \
-            '$EBROOTPYTHON/lib/python{short_python_ver}/lib-dynload/"',
+            'PYTHONPATH="$PYTHONPATH:$PWD/protocols:$PWD/libraries/bindings/python/:{python_dynlib_dir}'
             'python setup.py install'
         ])
-        cmd = cmd.format(short_python_ver=self.python_short_ver)
+        cmd = cmd.format(short_python_ver=self.python_short_ver, python_dynlib_dir=python_dynlib_dir)
         run_cmd(cmd, log_all=True, simple=True)
 
     def sanity_check_step(self):
--- a/easybuild/easyblocks/x/xmipp.py
+++ b/easybuild/easyblocks/x/xmipp.py
@@ -52,13 +52,6 @@ class EB_Xmipp(EasyBlock):
         self.cfg.update('unpack_options', '--strip-components=1')
         super(EB_Xmipp, self).extract_step()
 
-        # extract some dependencies that we really need and can't find anywhere else
-        # alglib tarball has version in name, so lets find it with a glob
-        alglib_tar = glob.glob(os.path.join(os.getcwd(), 'external', 'alglib*.tgz'))[0]
-        for src in ['bilib.tgz', 'bilib.tgz', 'condor.tgz', alglib_tar]:
-            external_path = os.path.join(self.cfg['start_dir'], 'external')
-            extract_file(os.path.join(external_path, src), external_path)
-
     def configure_step(self):
         """Configure xmipp build via a provided wrapper around sconss."""
         # check if all our dependencies are in place
@@ -70,6 +63,16 @@ class EB_Xmipp(EasyBlock):
         java_root = get_software_root('Python')
         if not java_root:
             self.log.error("Java not loaded as a dependency, which is required for %s" % self.name)
+
+        # extract some dependencies that we really need and can't find anywhere else
+        # alglib tarball has version in name, so lets find it with a glob
+        # we can't do this in extract step before these are in the original sources tarball, so we need to know
+        # startdir first
+        alglib_tar = glob.glob(os.path.join(os.getcwd(), 'external', 'alglib*.tgz'))[0]
+        for src in ['bilib.tgz', 'bilib.tgz', 'condor.tgz', alglib_tar]:
+            external_path = os.path.join(self.cfg['start_dir'], 'external')
+            extract_file(os.path.join(external_path, src), external_path)
+
 
         # build step expects these to exist
         mkdir(os.path.join(self.cfg['start_dir'], 'bin'))
--- a/easybuild/easyblocks/x/xmipp.py
+++ b/easybuild/easyblocks/x/xmipp.py
@@ -68,11 +68,13 @@ class EB_Xmipp(EasyBlock):
         # alglib tarball has version in name, so lets find it with a glob
         # we can't do this in extract step before these are in the original sources tarball, so we need to know
         # startdir first
-        alglib_tar = glob.glob(os.path.join(os.getcwd(), 'external', 'alglib*.tgz'))[0]
-        for src in ['bilib.tgz', 'bilib.tgz', 'condor.tgz', alglib_tar]:
-            external_path = os.path.join(self.cfg['start_dir'], 'external')
+        external_path = os.path.join(self.cfg['start_dir'], 'external')
+        alglib_tar = glob.glob(external_path, 'alglib*.tgz'))[0]
+        for src in ['bilib.tgz', 'bilib.tgz', 'condor.tgz', alglib_tar, 'scons.tgz']:
             extract_file(os.path.join(external_path, src), external_path)
 
+        # make sure we start in the start dir
+        os.chdir(self.cfg['start_dird'])
 
         # build step expects these to exist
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/x/xmipp.py
```diff
@@ -0,0 +1,116 @@
...
+from easybuild.tools.modules import get_software_root, get_software_version
+from easybuild.tools.run import run_cmd
+
+
+class EB_Xmipp(EasyBlock):
+    """Support for building/installing Xmipp."""
+
+    def __init__(self, *args, **kwargs):
+        """Easyblock constructor, enable building in installation directory."""
+        super(self.__class__, self).__init__(*args, **kwargs)
```
boegel: don't use `self.__class__`, since this will cause havoc for easyblocks deriving from `EB_Xmipp`

easybuild/easyblocks/x/xmipp.py
```diff
@@ -0,0 +1,116 @@
...
+        """Extract sources."""
+        # strip off 'xmipp' part to avoid having everything in a 'xmipp' subdirectory
+        self.cfg['unpack_options'] = "--strip-components=1"
+        super(self.__class__, self).extract_step()
+
+    def configure_step(self):
+        """Configure xmipp build via a provided wrapper around sconss."""
+        # install step expects these to exist
+        mkdir('lib')
+        mkdir('bin')
```
boegel: use these instead, to make it more clear what's going on

```python
mkdir(os.path.join(self.installdir, 'bin'))
mkdir(os.path.join(self.installdir, 'lib'))
```

easybuild/easyblocks/x/xmipp.py
```diff
@@ -0,0 +1,116 @@
...
+        # run interactive 'config' script to generate install.info file
+        try:
+            # ln -s `which python` xmipp_python
```
boegel: `# symlink xmipp_python to python binary`

the below is *not* equivalent to `which python`, rather you're relying on `Python` being specified as a dep

if it's not, this will go down hard btw, you want something like:

```python
if python_root:
    os.symlink(os.path.join(python_root, 'bin', 'python'), 'xmipp_python')
else:
    self.log.error("Python not loaded as a dependency, which is require for %s" % self.name)
```

easybuild/easyblocks/x/xmipp.py
```diff
@@ -0,0 +1,116 @@
...
+            orig = os.path.join(get_software_root('Python'), 'bin', 'python')
+            dest = 'xmipp_python'
+            os.symlink(orig, dest)
+            # ln -s $EBROOTPYTHON/lib/python2.7 lib/python2.7
+            self.python_short_ver = ".".join(get_software_version('Python').split(".")[0:2])
```
boegel: nitpicking: use `'.'` for single-char stuff, use `[:2]` (drop the `0`)

easybuild/easyblocks/x/xmipp.py
```diff
@@ -0,0 +1,116 @@
...
+        # install step expects these to exist
+        mkdir('lib')
+        mkdir('bin')
+        # run interactive 'config' script to generate install.info file
+        try:
+            # ln -s `which python` xmipp_python
+            orig = os.path.join(get_software_root('Python'), 'bin', 'python')
+            dest = 'xmipp_python'
+            os.symlink(orig, dest)
+            # ln -s $EBROOTPYTHON/lib/python2.7 lib/python2.7
```
boegel: change to: `# symlink to version-specific python lib subdir` (or somesuch)

easybuild/easyblocks/x/xmipp.py
```diff
@@ -0,0 +1,116 @@
...
+            self.python_short_ver = ".".join(get_software_version('Python').split(".")[0:2])
+            orig = os.path.join(get_software_root('Python'), 'lib', 'python%s' % self.python_short_ver)
+            dest = os.path.join('lib', 'python%s' % self.python_short_ver)
+            os.symlink(orig, dest)
+        except OSError as err:
+            self.log.error("Failed to symlink %s to %s: %s" % (orig, dest, err))
```
boegel: will this pick up the `orig` and `dest` from the `try` block?

I'd rather have *only* the call to `os.symlink` in a `try`-`except` block, with an error like:

```python
self.log.error("Failed to create symlinks: %s" % err)
```

easybuild/easyblocks/x/xmipp.py
```diff
@@ -0,0 +1,116 @@
...
+
+        # extract some dependencies that we really need and can't find anywhere else
+        # alglib tarball has version in name, so lets find it with a glob
+        alglib_tar = glob.glob(os.path.join(os.getcwd(), 'external', 'alglib*.tgz'))[0]
+        for src in ['bilib.tgz', 'bilib.tgz', 'condor.tgz', alglib_tar]:
+            external_path = os.path.join(os.getcwd(), 'external')
+            extract_file(os.path.join(external_path, src), external_path)
```
boegel: do this in `extract_step`?

easybuild/easyblocks/x/xmipp.py
```diff
@@ -0,0 +1,116 @@
...
+            dest = os.path.join('lib', 'python%s' % self.python_short_ver)
+            os.symlink(orig, dest)
+        except OSError as err:
+            self.log.error("Failed to symlink %s to %s: %s" % (orig, dest, err))
+
+        # extract some dependencies that we really need and can't find anywhere else
+        # alglib tarball has version in name, so lets find it with a glob
+        alglib_tar = glob.glob(os.path.join(os.getcwd(), 'external', 'alglib*.tgz'))[0]
+        for src in ['bilib.tgz', 'bilib.tgz', 'condor.tgz', alglib_tar]:
+            external_path = os.path.join(os.getcwd(), 'external')
```
boegel: use `self.cfg['start_dir']` rather than `os.getcwd()`

easybuild/easyblocks/x/xmipp.py
boegel: is creating this symlink absolutely needed btw by the Xmipp install procedure?

if not, just use `os.path.join(python_root, 'bin', 'python')`, no need to put symlink magic in place unless it's absolutely necessary

easybuild/easyblocks/x/xmipp.py
```diff
@@ -0,0 +1,116 @@
...
+            extract_file(os.path.join(external_path, src), external_path)
+
+        cmd = ('{preconfigopts} PATH=$PWD:$PATH xmipp_python external/scons/scons.py mode=configure -j {parallel}'
```
boegel: haha. no. ;-)

this is totally unreadable, and a f*cking nightmare to debug; don't assume you'll never have to get back to this

should be something like:

```python
cmd = ' '.join([
    self.cfg['preconfigopts'],
    "PATH=$PWD:$PATH",
    os.path.join(python_root, 'bin', 'python'),  # or xmipp_python
    ...
    '-j %s' % self.cfg['parallel'],
   ...
    self.cfg['configopts'],
])

File: easybuild/easyblocks/x/xmipp.py [hunk 10]
```diff
@@ -0,0 +1,116 @@
...
+               ' --config=force  profile=no fast=yes warn=no release=yes gtest=no cuda=no debug=no matlab=no'
+               ' java=no LINKERFORPROGRAMS="$CXX" MPI_BINDIR="$EBROOTIMPI" JAVA_HOME="$JAVA_HOME" JAVAC=javac CC="$CC"'
```
boegel: only works if `impi` is loaded, so add a check to make sure it is?

isn't this supposed to point to the `bin` subdir btw?

or just use something like `os.path.join(get_software_root(self.toolchain.MPI_MODULE_NAME[0]), 'bin')`

boegel: same for `$JAVA_HOME`, make sure it's set, or better: make sure `Java` is loaded as a dep, and use `get_software_root('Java')`, don't rely on any `$JAVA_HOME` that just happens to be defined

easybuild/easyblocks/x/xmipp.py
```diff
@@ -0,0 +1,116 @@
...
+               ' CXXFLAGS="$CXXFLAGS -DMPICH_IGNORE_CXX_SEEK -I$EBROOTPYTHON/include/python{short_python_ver}"'
+               ' CXX="$CXX" MPI_CC="$MPICC" MPI_CXX="$MPICXX" MPI_INCLUDE="$MPI_INC_DIR" MPI_LIBDIR="$MPI_LIB_DIR"'
```
boegel: open a PR to framework to get `$MPI_BIN_DIR` defined via `easybuild/tools/toolchain/mpi.py` as well, so you can use it?

easybuild/easyblocks/x/xmipp.py
```diff
@@ -0,0 +1,116 @@
...
+               ' MPI_LINKERFORPROGRAMS="$MPICC" LIBPATH="$LD_LIBRARY_PATH" {configopts}').format(
+                   parallel=self.cfg['parallel'], short_python_ver=self.python_short_ver,
+                   preconfigopts=self.cfg['preconfigopts'], configopts=self.cfg['configopts']
+                   )
+        run_cmd(cmd, log_all=True, simple=True)
+
+    def build_step(self):
+        """Custom build procedure for Xmipp: call the scons wrapper with compile argument"""
+        cmd = 'LD_LIBRARY_PATH=$LD_LIBRARY_PATH:$PWD/lib PATH=$PWD:$PATH xmipp_python external/scons/scons.py' \
```
boegel: same here, split up and use `' '.join([...])`

easybuild/easyblocks/x/xmipp.py
```diff
@@ -0,0 +1,116 @@
...
+              'mode=compile -j {parallel}'.format(parallel=self.cfg['parallel'])
+
+        run_cmd(cmd, log_all=True, simple=True)
+
+    def install_step(self):
+        """install step for xmipp, this builds a local database and seems to do some tests?"""
+        cmd = 'XMIPP_HOME=$PWD PATH=$PWD:$PWD/bin:$PATH PYTHONPATH=$PYTHONPATH:$PWD/protocols:' \
```
boegel: same here, split up and use `' '.join([...])`

easybuild/easyblocks/x/xmipp.py
```diff
@@ -0,0 +1,116 @@
...
+              '$PWD/libraries/bindings/python/:$EBROOTPYTHON/lib/python{short_python_ver}/lib-dynload/' \
+              ' python setup.py install'.format(self.python_short_ver)
+        run_cmd(cmd, log_all=True, simple=True)
+
+    def sanity_check_step(self):
+        """Custom sanity check for GAMESS-US."""
```
boegel: fix docstring

easybuild/easyblocks/x/xmipp.py
```diff
@@ -0,0 +1,116 @@
...
+        custom_paths = {
+            'files': ['xmipp_%s' % x for x in ['imagej', 'mpi_run', 'phantom_create', 'tomo_project', 'volume_align']],
+            'dirs': ['lib/'],
```
boegel: no need for the trailing `/`

easybuild/easyblocks/x/xmipp.py
```diff
@@ -0,0 +1,116 @@
...
+        }
+        super(self.__class__, self).sanity_check_step(custom_paths=custom_paths)
```
boegel: no `__class__`

easybuild/easyblocks/x/xmipp.py
```diff
@@ -0,0 +1,116 @@
...
+    def __init__(self, *args, **kwargs):
+        """Easyblock constructor, enable building in installation directory."""
+        super(self.__class__, self).__init__(*args, **kwargs)
+        self.build_in_installdir = True
+
+    def extract_step(self):
+        """Extract sources."""
+        # strip off 'xmipp' part to avoid having everything in a 'xmipp' subdirectory
+        self.cfg['unpack_options'] = "--strip-components=1"
+        super(self.__class__, self).extract_step()
```
boegel: no `__class__`

easybuild/easyblocks/x/xmipp.py
```diff
@@ -0,0 +1,116 @@
...
+        """Custom sanity check for GAMESS-US."""
+        custom_paths = {
+            'files': ['xmipp_%s' % x for x in ['imagej', 'mpi_run', 'phantom_create', 'tomo_project', 'volume_align']],
+            'dirs': ['lib/'],
+        }
+        super(self.__class__, self).sanity_check_step(custom_paths=custom_paths)
+
+    def make_module_extra(self):
+        """Define GAMESS-US specific variables in generated module file, i.e. $GAMESSUSROOT."""
+        txt = super(self.__class__, self).make_module_extra()
```
boegel: `__class__`

easybuild/easyblocks/x/xmipp.py
```diff
@@ -0,0 +1,116 @@
...
+    def sanity_check_step(self):
+        """Custom sanity check for GAMESS-US."""
+        custom_paths = {
+            'files': ['xmipp_%s' % x for x in ['imagej', 'mpi_run', 'phantom_create', 'tomo_project', 'volume_align']],
+            'dirs': ['lib/'],
+        }
+        super(self.__class__, self).sanity_check_step(custom_paths=custom_paths)
+
+    def make_module_extra(self):
+        """Define GAMESS-US specific variables in generated module file, i.e. $GAMESSUSROOT."""
```
boegel: fix docstring

easybuild/easyblocks/x/xmipp.py
```diff
@@ -0,0 +1,116 @@
...
+        txt = super(self.__class__, self).make_module_extra()
+        txt += self.module_generator.set_environment('XMIPP_HOME', '$root')
```
boegel: this will not work with module files in Lua syntax

@pforai: do we have an alternative in the works for this in your PR https://github.com/hpcugent/easybuild-framework/pull/1060?

easybuild/easyblocks/x/xmipp.py
```diff
@@ -0,0 +1,116 @@
...
+        try:
+            # ln -s `which python` xmipp_python
+            orig = os.path.join(get_software_root('Python'), 'bin', 'python')
+            dest = 'xmipp_python'
+            os.symlink(orig, dest)
+            # ln -s $EBROOTPYTHON/lib/python2.7 lib/python2.7
+            self.python_short_ver = ".".join(get_software_version('Python').split(".")[0:2])
+            orig = os.path.join(get_software_root('Python'), 'lib', 'python%s' % self.python_short_ver)
+            dest = os.path.join('lib', 'python%s' % self.python_short_ver)
+            os.symlink(orig, dest)
```
boegel: same here, is this symlinking magic strictly required by the Xmipp install procedure, or is it just something you rely on further down?

easybuild/easyblocks/x/xmipp.py
```diff
@@ -0,0 +1,116 @@
...
+
+    def __init__(self, *args, **kwargs):
+        """Easyblock constructor, enable building in installation directory."""
+        super(self.__class__, self).__init__(*args, **kwargs)
+        self.build_in_installdir = True
+
+    def extract_step(self):
+        """Extract sources."""
+        # strip off 'xmipp' part to avoid having everything in a 'xmipp' subdirectory
+        self.cfg['unpack_options'] = "--strip-components=1"
```
boegel: use `self.cfg.update`, don't blindly overwrite `unpack_options` in order to not destroy any (potentially useful) unpack options set in the easyconfig file

easybuild/easyblocks/x/xmipp.py
```diff
@@ -0,0 +1,116 @@
+##
+# Copyright 2009-2015 Ghent University
```
boegel: if this is an easyblock from scratch, use `2015-2015`

easybuild/easyblocks/x/xmipp.py
```diff
@@ -0,0 +1,116 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with EasyBuild.  If not, see <http://www.gnu.org/licenses/>.
+##
+"""
+EasyBuild support for building and installing Xmipp, implemented as an easyblock
+
+@author: Jens Timmerman (Ghent University)
```
boegel: if this is a cleaned up version of a ported easyblock, include the classic full list of names (including Toon, Pieter, etc.)

boegel:
@JensTimmerman: fully reviewed, looks pretty clean (after fixing the remarks)

major remark is to not blindly assume deps you require the install root for are loaded, make sure they are by checking that `get_software_root` is not `None`

easybuild/easyblocks/x/xmipp.py
JensTimmerman: they use hardcoded ./xmipp_python in a few places in some internal scripts, it was easier to create a symlink here then to patch the files and change them all to call python.

boegel: ok, fair enough, maybe mention that in a comment

don't use `xmipp_python` yourself deeper down though, we know where to find the `python` binary

easybuild/easyblocks/x/xmipp.py
JensTimmerman: copy pasted this from another easyblock :p
I think the correct way here would be to define this in some tools package as afunction (or a python easyblock??) and import it.

easybuild/easyblocks/x/xmipp.py
JensTimmerman: needed by the installation scripts

easybuild/easyblocks/x/xmipp.py
JensTimmerman: yes it will

easybuild/easyblocks/x/xmipp.py
boegel: ok, clarify with a comment then

easybuild/easyblocks/x/xmipp.py
boegel: that's not an excuse

`det_pylibdir` available from the generic `PythonPackage` easyblock does close to what you want/need, `os.path.dirname(det_pylibdir())` probably does what you want (but you'll need to verify that)

easybuild/easyblocks/x/xmipp.py
boegel: still, I'd prefer only having the `symlink` bits in a `try`/`except`, to make it clear what may be throwing an `OSError`

easybuild/easyblocks/x/xmipp.py
JensTimmerman: right, it breaks inheratance, I forgot why I didn't recommend people to use this :p

easybuild/easyblocks/x/xmipp.py
JensTimmerman: this should actually be in the build dir, your version will work because we're building in install dir, but maybe it could be done better in the future, so I should go for `self.cfg['startdir']` I guess.

boegel: ok, fair enough

easybuild/easyblocks/x/xmipp.py
JensTimmerman: ok, you're right, after some more testing and setting a few environment variables we can get rid of the symlinks all together.

easybuild/easyblocks/x/xmipp.py
JensTimmerman: what do I do here then?

easybuild/easyblocks/x/xmipp.py
JensTimmerman: `Can't update configuration value for unpack_options, because it's not a string or list.")`

easybuild/easyblocks/x/xmipp.py
```diff
@@ -0,0 +1,152 @@
...
+        custom_paths = {
+            'files': ['xmipp_%s' % x for x in ['imagej', 'mpi_run', 'phantom_create', 'tomo_project', 'volume_align']],
+            'dirs': ['lib'],
+        }
+        super(EB_Xmipp, self).sanity_check_step(custom_paths=custom_paths)
+
+    def make_module_extra(self):
+        """Define Xmipp specific variables in generated module file, i.e. XMIPP_HOME."""
+        txt = super(EB_Xmipp, self).make_module_extra()
+        txt += self.module_generator.set_environment('XMIPP_HOME', '$root')
```
boegel: I don't think we have a generic alternative for this yet (as in, something that will work for both Tcl and Lua module files), @pforai may correct me here

so stick with this `$root` approach, for now

easybuild/easyblocks/x/xmipp.py
JensTimmerman: fixed in https://github.com/hpcugent/easybuild-framework/pull/1229

easybuild/easyblocks/x/xmipp.py
pforai: In this case I would go with installdir and not $root.

easybuild/easyblocks/x/xmipp.py
JensTimmerman: I actually need to know the start_dir, (or the getcwd for the config step) before I can extract these, and guess_start_dir is only called before  configure_step

JensTimmerman:
@boegel ready for new review

easybuild/easyblocks/x/xmipp.py
boegel: Good point.

easybuild/easyblocks/x/xmipp.py
```diff
@@ -0,0 +1,157 @@
...
+        self.build_in_installdir = True
+
+    def extract_step(self):
+        """Extract sources."""
+        # strip off 'xmipp' part to avoid having everything in a 'xmipp' subdirectory
+        self.cfg.update('unpack_options', '--strip-components=1')
+        super(EB_Xmipp, self).extract_step()
+
+    def configure_step(self):
+        """Configure xmipp build via a provided wrapper around sconss."""
```
boegel: fix typo: `sconss`

easybuild/easyblocks/x/xmipp.py
```diff
@@ -0,0 +1,157 @@
...
+        # check if all our dependencies are in place
+        python_root = get_software_root('Python')
+        if not python_root:
+            self.log.error("Python not loaded as a dependency, which is required for %s" % self.name)
+        python_libdir = det_pylibdir()
+        self.python_short_ver = ".".join(get_software_version('Python').split(".")[0:2])
+        java_root = get_software_root('Python')
```
boegel: wahahahaha, nice one

easybuild/easyblocks/x/xmipp.py
```diff
@@ -0,0 +1,157 @@
...
+        # alglib tarball has version in name, so lets find it with a glob
+        # we can't do this in extract step before these are in the original sources tarball, so we need to know
+        # startdir first
+        external_path = os.path.join(self.cfg['start_dir'], 'external')
+        alglib_tar = glob.glob(os.path.join(external_path, 'alglib*.tgz'))[0]
+        for src in ['bilib.tgz', 'bilib.tgz', 'condor.tgz', alglib_tar, 'scons.tgz']:
+            extract_file(os.path.join(external_path, src), external_path)
+
+        # make sure we start in the start dir
+        os.chdir(self.cfg['start_dird'])
```
boegel: this is clearly not retested... fix typo

boegel: also, wrap this in `try`-`except OSError`

easybuild/easyblocks/x/xmipp.py
```diff
@@ -0,0 +1,157 @@
...
+
+        # build step expects these to exist
+        mkdir(os.path.join(self.cfg['start_dir'], 'bin'))
+        mkdir(os.path.join(self.cfg['start_dir'], 'lib'))
+
+        cmd = ' '.join([
+            '{preconfigopts}',
+            'python external/scons/scons.py mode=configure -j {parallel}',
```
boegel: these have nothing to do with each other, so split up in separate entries/lines?

easybuild/easyblocks/x/xmipp.py
```diff
@@ -0,0 +1,157 @@
...
+            'cuda=no',
+            'debug=no',
+            'matlab=no',
+            'java=no',
+            'LINKERFORPROGRAMS="$CXX"',
+            'MPI_BINDIR="{mpi_bindir}"',
+            'JAVA_HOME="{java_home}"',
+            'JAVAC=javac',
+            'CC="$CC"',
+            'CXXFLAGS="$CXXFLAGS -DMPICH_IGNORE_CXX_SEEK -I$EBROOTPYTHON/include/python{short_python_ver} \
```
boegel: better construct `cxxflags` before, rather than having this long line

easybuild/easyblocks/x/xmipp.py
```diff
@@ -0,0 +1,157 @@
...
+            'CXX="$CXX"',
+            'MPI_CC="$MPICC"',
+            'MPI_CXX="$MPICXX"',
+            'MPI_INCLUDE="$MPI_INC_DIR"',
+            'MPI_LIBDIR="$MPI_LIB_DIR"',
+            'MPI_LINKERFORPROGRAMS="$MPICC"',
+            'LIBPATH="$LD_LIBRARY_PATH"',
+            '{configopts}',
+        ])
+        cmd = cmd.format(
```
boegel: this looks fancy, but it's not needed, just template the strings directly in the list above, it'll be clearer

easybuild/easyblocks/x/xmipp.py
```diff
@@ -0,0 +1,157 @@
...
+            java_home=java_root
+        )
+        run_cmd(cmd, log_all=True, simple=True)
+
+    def build_step(self):
+        """Custom build procedure for Xmipp: call the scons wrapper with compile argument"""
+        cmd = ' '.join([
+            'LD_LIBRARY_PATH=$LD_LIBRARY_PATH:$PWD/lib',
+            'python external/scons/scons.py',
+            'mode=compile -j {parallel}',
```
boegel: split up

easybuild/easyblocks/x/xmipp.py
```diff
@@ -0,0 +1,157 @@
...
+        ])
+        cmd = cmd.format(parallel=self.cfg['parallel'])
```
boegel: same here, no need for format, use this above:

```python
    '-j %s' % self.cfg['parallel'],
```

easybuild/easyblocks/x/xmipp.py
```diff
@@ -0,0 +1,157 @@
...
+        run_cmd(cmd, log_all=True, simple=True)
+
+    def install_step(self):
+        """install step for xmipp, this builds a local database and seems to do some tests?"""
+        cmd = ' '.join([
+            'XMIPP_HOME=$PWD',
+            'PATH=$PWD/bin:$PATH',
+            'PYTHONPATH="$PYTHONPATH:$PWD/protocols:$PWD/libraries/bindings/python/:' \
+            '$EBROOTPYTHON/lib/python{short_python_ver}/lib-dynload/"',
```
boegel: same here, template this line

easybuild/easyblocks/x/xmipp.py
```diff
@@ -0,0 +1,157 @@
...
+        ])
+        cmd = cmd.format(parallel=self.cfg['parallel'])
+        run_cmd(cmd, log_all=True, simple=True)
+
+    def install_step(self):
+        """install step for xmipp, this builds a local database and seems to do some tests?"""
+        cmd = ' '.join([
+            'XMIPP_HOME=$PWD',
+            'PATH=$PWD/bin:$PATH',
+            'PYTHONPATH="$PYTHONPATH:$PWD/protocols:$PWD/libraries/bindings/python/:' \
```
boegel: avoid long line (especially without indenting...)

easybuild/easyblocks/x/xmipp.py
```diff
@@ -0,0 +1,157 @@
...
+    def sanity_check_step(self):
+        """Custom sanity check for Xmipp."""
+        custom_paths = {
+            'files': ['xmipp_%s' % x for x in ['imagej', 'mpi_run', 'phantom_create', 'tomo_project', 'volume_align']],
+            'dirs': ['lib'],
+        }
+        super(EB_Xmipp, self).sanity_check_step(custom_paths=custom_paths)
+
+    def make_module_extra(self):
+        """Define Xmipp specific variables in generated module file, i.e. XMIPP_HOME."""
```
boegel: nitpicking: mention `$XMIPP_HOME` in the docstring

JensTimmerman: it is

easybuild/easyblocks/x/xmipp.py
JensTimmerman: but then the line where I construct `cxxflags` will have the same issue? being a long line...

easybuild/easyblocks/x/xmipp.py
```diff
@@ -0,0 +1,164 @@
...
+
+        # make sure we start in the start dir
+        os.chdir(self.cfg['start_dir'])
+
+        # build step expects these to exist
+        mkdir(os.path.join(self.cfg['start_dir'], 'bin'))
+        mkdir(os.path.join(self.cfg['start_dir'], 'lib'))
+
+        cmd = ' '.join([
+            '{preconfigopts}',
```
boegel: use `self.cfg['preconfigopts']` directly here, no need for templating

JensTimmerman:
should be good to go (for another round of remarks)

easybuild/easyblocks/x/xmipp.py
```diff
@@ -0,0 +1,161 @@
...
+    def __init__(self, *args, **kwargs):
+        """Easyblock constructor, enable building in installation directory."""
+        super(EB_Xmipp, self).__init__(*args, **kwargs)
+        self.build_in_installdir = True
+
+    def extract_step(self):
+        """Extract sources."""
+        # strip off 'xmipp' part to avoid having everything in a 'xmipp' subdirectory
+        if not self.cfg['unpack_options']:
+            self.cfg['unpack_options'] = '--strip-components=1'
```
boegel: no pls, you really need the `--strip-components=1` bit below (and you're not sure it'll be there if `unpack_options` is defined by the easyconfig), so use this:

```python
self.cfg.update('unpack_options', '--strip-components=1')
```

I'll merge https://github.com/hpcugent/easybuild-framework/pull/1229 as soon as Jenkins gives it a green light, no worries.

easybuild/easyblocks/x/xmipp.py
```diff
@@ -0,0 +1,161 @@
...
+        super(EB_Xmipp, self).extract_step()
+
+    def configure_step(self):
+        """Configure xmipp build via a provided wrapper around scons."""
+        # check if all our dependencies are in place
+        python_root = get_software_root('Python')
+        if not python_root:
```
boegel: `python_root` is never used, so change to

```python
if get_software_root('Python') is None:
```

easybuild/easyblocks/x/xmipp.py
```diff
@@ -0,0 +1,161 @@
...
+        # alglib tarball has version in name, so lets find it with a glob
+        # we can't do this in extract step before these are in the original sources tarball, so we need to know
+        # startdir first
+        external_path = os.path.join(self.cfg['start_dir'], 'external')
+        alglib_tar = glob.glob(os.path.join(external_path, 'alglib*.tgz'))[0]
+        for src in ['bilib.tgz', 'bilib.tgz', 'condor.tgz', alglib_tar, 'scons.tgz']:
+            extract_file(os.path.join(external_path, src), external_path)
+
+        # make sure we start in the start dir
+        os.chdir(self.cfg['start_dir'])
```
boegel:
```python
try:
except OSError as err:
    self.log.error("Failed to change back to %s: %s" % (self.cfg['start_dir'], err))
```

easybuild/easyblocks/x/xmipp.py
```diff
@@ -0,0 +1,161 @@
...
+
+        # build step expects these to exist
+        mkdir(os.path.join(self.cfg['start_dir'], 'bin'))
+        mkdir(os.path.join(self.cfg['start_dir'], 'lib'))
+
+        cmd = ' '.join([
+            self.cfg['preconfigopts'],
+            'python external/scons/scons.py',
+            'mode=configure',
+            '-j {parallel}',
```
boegel:
```python
    '-j %d' % self.cfg['parallel'],
```

or

```python
    '-j ' + self.cfg['parallel'],
```

easybuild/easyblocks/x/xmipp.py
```diff
@@ -0,0 +1,161 @@
...
+            'fast=yes',
+            'warn=no',
+            'release=yes',
+            'gtest=no',
+            'cuda=no',
+            'debug=no',
+            'matlab=no',
+            'java=no',
+            'LINKERFORPROGRAMS="$CXX"',
+            'MPI_BINDIR="{mpi_bindir}"',
```
boegel:
```python
    'MPI_BINDIR="%s"' % os.path.join(get_software_root(self.toolchain.MPI_MODULE_NAME[0]), 'bin'),
```

easybuild/easyblocks/x/xmipp.py
```diff
@@ -0,0 +1,161 @@
...
+            'JAVA_HOME="{java_home}"',
```
boegel:
```python
    'JAVA_HOME=%s' % java_root,
```

easybuild/easyblocks/x/xmipp.py
```diff
@@ -0,0 +1,161 @@
...
+            'JAVAC=javac',
+            'CC="$CC"',
+            'CXXFLAGS="%s"' % ' '.join([
+                '$CXXFLAGS',
+                '-DMPICH_IGNORE_CXX_SEEK',
+                '-I$EBROOTPYTHON/include/python{short_python_ver}',
```
boegel:
```python
        '-I%s' % os.path.join(python_root, 'include', 'python%s' % self.python_short_ver),
```

easybuild/easyblocks/x/xmipp.py
boegel: hold on, it is being used by one of my remarks below, so leave as is

easybuild/easyblocks/x/xmipp.py
```diff
@@ -0,0 +1,161 @@
...
+                '-I$EBROOTPYTHON/{python_libdir}/numpy/core/include/',
```
boegel:
```python
        '-I%s' % os.path.join(python_root, python_libdir, 'numpy', 'core', 'include'),
```

And actually, you should make sure this path is there, and fail hard if it's not.

easybuild/easyblocks/x/xmipp.py
```diff
@@ -0,0 +1,161 @@
...
+            ]),
+            'CXX="$CXX"',
+            'MPI_CC="$MPICC"',
+            'MPI_CXX="$MPICXX"',
+            'MPI_INCLUDE="$MPI_INC_DIR"',
+            'MPI_LIBDIR="$MPI_LIB_DIR"',
+            'MPI_LINKERFORPROGRAMS="$MPICC"',
+            'LIBPATH="$LD_LIBRARY_PATH"',
+            self.cfg['configopts'],
+        ]).format(
```
boegel: drop all of this, see remarks above

easybuild/easyblocks/x/xmipp.py
```diff
@@ -0,0 +1,161 @@
...
+            'LD_LIBRARY_PATH=$LD_LIBRARY_PATH:$PWD/lib',
+            'python external/scons/scons.py',
+            'mode=compile',
+            '-j %s' % self.cfg['parallel'],
+        ])
+        run_cmd(cmd, log_all=True, simple=True)
+
+    def install_step(self):
+        """install step for xmipp, this builds a local database and seems to do some tests?"""
+        python_dynlib_dir = '$EBROOTPYTHON/lib/python%s/lib-dynload/' % self.python_short_ver
```
boegel:
```python
python_dynlib_dir = os.path.join(self.python_root, 'lib', 'python%s' % self.python_short_ver, 'lib-dynload')
```

and maybe also make very sure this exists?

easybuild/easyblocks/x/xmipp.py
```diff
@@ -0,0 +1,161 @@
...
+
+        cmd = ' '.join([
+            'XMIPP_HOME=$PWD',
+            'PATH=$PWD/bin:$PATH',
```
boegel: don't use `$PWD`, too fragile

```python
    'PATH=%s:$PATH' % os.path.join(self.cfg['start_dir'], 'bin'),
```

easybuild/easyblocks/x/xmipp.py
```diff
@@ -0,0 +1,161 @@
...
+            'PYTHONPATH="$PYTHONPATH:$PWD/protocols:$PWD/libraries/bindings/python/:{python_dynlib_dir}"',
```
boegel:
```python
extra_pythonpaths = [
    os.path.join(self.cfg['start_dir'], 'protocols'),
    os.path.join(self.cfg['start_dir'], 'libraries', 'bindings', 'python'),
    python_dynlib_dir,
]
...
    'PYTHONPATH=%s' % os.pathsep.join(['$PYTHONPATH'] + [extra_pythonpaths])
```

and do you really want the original `$PYTHONPATH` *first*?

easybuild/easyblocks/x/xmipp.py
boegel: ok, whateva

easybuild/easyblocks/x/xmipp.py
JensTimmerman: what is wrong with the current way?

boegel: different style than in other easyblocks, using one thing consistently helps

there's no technical reason to switch to using `.format` (it's not faster or something like that, afaik), so let's stick to `%` style formatting

there may be cases where `.format` is the way to go (we use it in framework in a couple of places), but this is not one of them

easybuild/easyblocks/x/xmipp.py
JensTimmerman: well, it works like this, so I'll leave it for now. 
combined with the step below this can be a nice example for people to see the 2 ways you can do string templating in python, and how the new .format function works, so they can get ready to start using it.

JensTimmerman:
further cleaned up, more remarks?

JensTimmerman:
Test report: #### Test result
Build succeeded for 1 out of 1

 * **SUCCESS** _xmipp.eb_

#### Time info
 * start: Wed, 18 Mar 2015 16:41:06 +0000 (UTC)
 * end: Wed, 18 Mar 2015 16:43:49 +0000 (UTC)

JensTimmerman:
@boegel any more remarks?

boegel:
tested with https://github.com/hpcugent/easybuild-easyconfigs/pull/1489, works like a charm, looking good, so going in, thanks @JensTimmerman!

PR #582, opened 2015-03-19
drop -noroot for recent FLUENT versions, honor installopts, enable -debug

boegel:
@wpoely86: please review, well (re)test all existing FLUENT easyconfigs with this in place

### DIFF
--- a/easybuild/easyblocks/f/fluent.py
+++ b/easybuild/easyblocks/f/fluent.py
@@ -27,9 +27,9 @@ EasyBuild support for installing FLUENT, implemented as an easyblock
 
 @author: Kenneth Hoste (Ghent University)
 """
-
 import os
 import stat
+from distutils.version import LooseVersion
 
 from easybuild.framework.easyblock import EasyBlock
 from easybuild.tools.filetools import adjust_permissions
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/f/fluent.py
```diff
@@ -55,7 +55,12 @@ def build_step(self):
     def install_step(self):
         """Custom install procedure for FLUENT."""
 
-        cmd = "./INSTALL -noroot -silent -install_dir %s" % self.installdir
+        extra_args =''
+        # only include -noroot flag for older versions
+        if LooseVersion(self.version) < LooseVersion('15.0'):
+            extra_args += '-noroot'
+
+        cmd = "./INSTALL %s -debug -silent -install_dir %s %s" % (extra_args, self.installdir, self.cfg['installopts'])
```
wpoely86: we want debug enabled always?

boegel: doesn't hurt, I guess

the real solution for large (debug) logs is supporting automatically zipping them at the end of the installation ;)

wpoely86: debug refers too the install process or the actual application being installed?

boegel: the install process

wpoely86: forget the remark than :wink:

boegel:
retested with all existing FLUENT easyconfigs, nothing was broken, so going in

PR #584, opened 2015-03-26
Add libmpichf90 to the libmpi variables

ocaisa:
For NWChem to build with the latest version of psmpi (which includes proper MPI 3.0 support) the additional lib is required to build. We've chosen libmpichf90 (which is just a soft link to libmpifort, that is the library that appears in ```mpifort -show```) to ensure backwards compatibility


boegel:
Jenkins: ok to test

boegel:
It would be nice to have an NWChem easyconfig that triggers this code patch. @ebgregory?

I won't hold back this PR though, going in, thanks @ocaisa!

PR #586, opened 2015-03-27
Rosetta fix

pescobar:
one of my users complained because he was missing some scripts in my rosetta installation. I added this line to the rosetta easyblock to fix it

### DIFF
--- a/easybuild/easyblocks/r/rosetta.py
+++ b/easybuild/easyblocks/r/rosetta.py
@@ -235,7 +235,6 @@ class EB_Rosetta(EasyBlock):
         extract_and_copy('rosetta_database%s')
         extract_and_copy('BioTools%s', optional=True)
         extract_and_copy('rosetta_tools%s')
-        extract_and_copy('rosetta_demos%s')
 
 
     def sanity_check_step(self):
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/r/rosetta.py
```diff
@@ -234,6 +234,8 @@ def extract_and_copy(dirname_tmpl, optional=False):
         # (extract and) copy database and biotools (if it's there)
         extract_and_copy('rosetta_database%s')
         extract_and_copy('BioTools%s', optional=True)
+        extract_and_copy('rosetta_tools%s')
+
```
boegel: why more blank lines?

easybuild/easyblocks/r/rosetta.py
```diff
@@ -234,6 +234,8 @@ def extract_and_copy(dirname_tmpl, optional=False):
         # (extract and) copy database and biotools (if it's there)
         extract_and_copy('rosetta_database%s')
         extract_and_copy('BioTools%s', optional=True)
+        extract_and_copy('rosetta_tools%s')
```
boegel: are you sure this is always there? should we have `optional=True` included to just ignore it if it's not there?

boegel:
retested Rosette easyconfigs on top of this, works fine

boegel:
Thanks @pescobar!

PR #590, opened 2015-04-05
stop using $root to make easyblock compatible with module files in Lua syntax

boegel:
cfr. https://github.com/hpcugent/easybuild-framework/pull/1060

### DIFF
--- a/easybuild/easyblocks/o/openbabel.py
+++ b/easybuild/easyblocks/o/openbabel.py
@@ -27,7 +27,7 @@ EasyBuild support for OpenBabel, implemented as an easyblock
 
 @author: Ward Poelmans (Ghent University)
 """
-
+import os
 from easybuild.easyblocks.generic.cmakemake import CMakeMake
 from easybuild.tools.modules import get_software_root, get_software_version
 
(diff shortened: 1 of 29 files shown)


pforai:
:+1: here as well!

boegel:
Jenkins: test this please

boegel:
@wpoely86: can you give this a thorough review? should be fine to go in, but it's good to have someone else double-check

easybuild/easyblocks/o/openbabel.py
```diff
@@ -76,6 +76,8 @@ def make_module_extra(self):
         """Custom variables for OpenBabel module."""
         txt = super(EB_OpenBabel, self).make_module_extra()
         txt += self.module_generator.prepend_paths('PYTHONPATH', ['lib'])
-        txt += self.module_generator.set_environment('BABEL_LIBDIR', '$root/lib/openbabel/%s/' % self.version)
-        txt += self.module_generator.set_environment('BABEL_DATADIR', '$root/share/openbabel/%s/' % self.version)
+        babel_libdir = os.path.join(self.installdir, 'lib', 'openbabel', self.version)
+        txt += self.module_generator.set_environment('BABEL_LIBDIR', babel_libdir)
+        babel_datadir = os.path.join('share', 'openbabel', self.version)
```
wpoely86: `self.installdir,`

easybuild/easyblocks/p/python_meep.py
```diff
@@ -136,10 +136,10 @@ def make_module_extra(self):
 
         meep = get_software_root('Meep')
```
wpoely86: why is this here? Why not `self.installdir`?

wpoely86:
looks good

easybuild/easyblocks/p/python_meep.py
boegel: Meep != python-meep

easybuild/easyblocks/o/openbabel.py
boegel: wow, nice catch

boegel:
I've tested this thoroughly by rebuilding all existing easyconfigs with the following toolchains: `goolf/1.4.10`, `ictce/5.3.0`, `intel/*` and `foss/*`, + NWChem with `ictce/4.1.13`

`eb` can now generate Lua module files for all of those builds, hooray!

PR #592, opened 2015-04-17
GAMESS-US

valtandor:
Support for RDMA and exclusion of tests that don't run in parallel


scintimandrion:
Yes.

boegel:
Jenkins: ok to test

boegel:
hmm, this never got any attention at all... :(

@wpoely86: please review?

easybuild/easyblocks/g/gamess_us.py
```diff
@@ -248,28 +250,43 @@ def test_step(self):
...
             # run all exam<id> tests, dump output to exam<id>.log
-            n_tests = 47
-            for i in range(1, n_tests+1):
-                test_cmd = ' '.join(test_env_vars + [rungms, 'exam%02d' % i, self.version, '1', '2'])
-                (out, _) = run_cmd(test_cmd, log_all=True, simple=False)
-                write_file('exam%02d.log' % i, out)
+            total_num_tests = 47
+            badtests = []
+            if self.cfg['ddi_comm'] == 'mpi':
+                badtests = [ 5, 32, 39, 42, 45, 46, 47 ]
```
wpoely86: drop the spaces before and after the first and last element (or run `pep8`).

wpoely86:
looks good

easybuild/easyblocks/g/gamess_us.py
```diff
@@ -248,28 +250,43 @@ def test_step(self):
...
+            else:
+                success_regex = re.compile("^%d job\(s\) got incorrect numerical results\. Please examine why\." % len(badtests), re.M)
             if success_regex.search(out):
                 self.log.info("All tests ran successfully!")
+                # cleanup
+                os.chdir(cwd)
+                try:
+                    shutil.rmtree(self.testdir)
+                except OSError, err:
+                    self.log.error("Failed to remove test directory %s: %s" % (self.testdir, err))
```
wpoely86: change the `self.log.error` to `raise EasyBuildError`

easybuild/easyblocks/g/gamess_us.py
```diff
@@ -248,28 +250,43 @@ def test_step(self):
...
+            total_num_tests = 47
+            badtests = []
+            if self.cfg['ddi_comm'] == 'mpi':
+                badtests = [5, 32, 39, 42, 45, 46, 47]
+            for i in range(1, total_num_tests+1):
+                runthistest = True
+                if self.cfg['ddi_comm'] == 'mpi':
+                    # Not all tests can run in parallel. Skip those that can't.
+                    if i in badtests:
+                        runthistest = False
```
boegel: you can use `continue` instead of playing with a boolean variable like `runthistest`, which will just jump to the next iteration of the for loop (https://docs.python.org/2/reference/simple_stmts.html#the-continue-statement)

For the tests that can't be run in parallel, can we run them serially instead?


```python
skip_tests = []
if self.cfg['ddi_comm'] == 'mpi':
    # FIXME: reference to which tests are serial/parallel
    skip_tests = [5, 32, 39, 42, 45, 46, 47]
for i in range(range(1, total_num_tests+1):
    ncpus = 1
    if self.cfg['ddi_comm'] == 'mpi':
        ncpus = 2
        if i in skip_tests:
            # skip this tests & move to the next one
            
            continue
```

easybuild/easyblocks/g/gamess_us.py
```diff
@@ -248,28 +250,43 @@ def test_step(self):
...
+                    write_file('exam%02d.log' % i, out)
 
             # verify output of tests
             check_cmd = os.path.join(self.installdir, 'tests', 'standard', 'checktst')
             (out, _) = run_cmd(check_cmd, log_all=True, simple=False)
-            success_regex = re.compile("^All %d test results are correct" % n_tests, re.M)
+            if len(badtests) == 0:
+                success_regex = re.compile("^All %d test results are correct" % total_num_tests, re.M)
+            else:
+                success_regex = re.compile("^%d job\(s\) got incorrect numerical results\. Please examine why\." % len(badtests), re.M)
```
boegel: hmm, is this because they're being skipped?

would just running them with `ncpus = 1` fix this?

easybuild/easyblocks/g/gamess_us.py
```diff
@@ -248,28 +250,43 @@ def test_step(self):
...
             if success_regex.search(out):
```
boegel: add a comment above this line: `# only clean up if all tests were successful`

easybuild/easyblocks/g/gamess_us.py
```diff
@@ -185,11 +186,12 @@ def configure_step(self):
...
         # patch hardcoded settings in rungms to use values specified in easyconfig file
         rungms = os.path.join(self.builddir, 'rungms')
-        extra_gmspath_lines = "set ERICFMT=$GMSPATH/auxdata/ericfmt.dat\nset MCPPATH=$GMSPATH/auxdata/MCP\n"
+        extra_gmspath_lines = "set ERICFMT=$GMSPATH/auxdata/ericfmt.dat\nset MCPPATH=$GMSPATH/auxdata/MCP"
+        extra_rdma_line = "set USE_RDMA"
         try:
             for line in fileinput.input(rungms, inplace=1, backup='.orig'):
                 line = re.sub(r"^(\s*set\s*TARGET)=.*", r"\1=%s" % self.cfg['ddi_comm'], line)
-                line = re.sub(r"^(\s*set\s*GMSPATH)=.*", r"\1=%s\n%s" % (self.installdir, extra_gmspath_lines), line)
+                line = re.sub(r"^(\s*set\s*GMSPATH)=.*", r"\1=%s\n%s\n%s=%s\n" % (self.installdir, extra_gmspath_lines, extra_rdma_line, self.cfg['userdma']), line)
```
boegel: this line is way too long, please split it up:

```python
repl_with = r"\1=%s\n%s\n%s=%s\n" % (self.installdir, extra_gmspath_lines, extra_rdma_line, self.cfg['userdma'])
line = re.sub(r"^(\s*set\s*GMSPATH)=.*", repl_with, line)
```

easybuild/easyblocks/g/gamess_us.py
```diff
@@ -185,11 +186,12 @@ def configure_step(self):
 
         # patch hardcoded settings in rungms to use values specified in easyconfig file
         rungms = os.path.join(self.builddir, 'rungms')
-        extra_gmspath_lines = "set ERICFMT=$GMSPATH/auxdata/ericfmt.dat\nset MCPPATH=$GMSPATH/auxdata/MCP\n"
+        extra_gmspath_lines = "set ERICFMT=$GMSPATH/auxdata/ericfmt.dat\nset MCPPATH=$GMSPATH/auxdata/MCP"
+        extra_rdma_line = "set USE_RDMA"
         try:
             for line in fileinput.input(rungms, inplace=1, backup='.orig'):
```
boegel: ideally, use of `fileinput` should be replace with using `apply_regex_substitutions`, see http://easybuild.readthedocs.org/en/latest/Extended_dry_run.html#runtime-patching-of-files-apply-regex-substitutions

easybuild/easyblocks/g/gamess_us.py
```diff
@@ -61,6 +61,7 @@ def extra_options():
         """Define custom easyconfig parameters for GAMESS-US."""
         extra_vars = {
             'ddi_comm': ['mpi', "DDI communication layer to use", CUSTOM],
+            'userdma': [False, "Use RDMA rather than RSH/SSH between nodes", CUSTOM],
```
boegel: make it clear that this only applies to the tests?

boegel: urgh, no it doesn't, sorry, please ignore

lexming:
Closing as superseded by https://github.com/easybuilders/easybuild-easyblocks/pull/3047

PR #594, opened 2015-04-21
easyblock for LAMMPS (REVIEW)

boegel:
cc @ocaisa


easybuild/easyblocks/l/lammps.py
```diff
@@ -0,0 +1,148 @@
...
+        """Custom easyconfig parameters for LAMMPS."""
+        extra_vars = {
+            'packages': [[], "LAMMPS packages to install", CUSTOM],
+        }
+        return EasyBlock.extra_options(extra_vars)
+
+    def extract_step(self):
+        """Extract LAMMPS sources."""
+        # strip off top-level subdirectory
+        self.cfg.update('unpack_options', '--strip-components=1')
```
ocaisa: Syntax for this line is wrong somehow (at least with EB 2.0.0), it could be:
self.cfg['unpack_options'] = "--strip-components=1"

boegel: works fine with upcoming 2.1 and current develop thanks to https://github.com/hpcugent/easybuild-framework/pull/1229

for now, you can change this to what you mention, but that will override any unpack options listed in the easyconfig file; using `update`, you'd append to unpack options specified in the easyconfig instead

easybuild/easyblocks/l/lammps.py
```diff
@@ -0,0 +1,148 @@
...
+EasyBuild support for building and installing LAMMPS, implemented as an easyblock
+
+@author: Kenneth Hoste (Ghent University)
+@author: Alexander Schnurpfeil (Juelich Supercomputer Centre)
+"""
+import os
+
+import easybuild.tools.toolchain as toolchain
+from easybuild.framework.easyblock import EasyBlock
+from easybuild.framework.easyconfig import CUSTOM
```
ocaisa: Some of the triggers for lammps use ```make yes-XXX``` but the user ones need ```make yes-user-XXX```, we figured the easiest way to set this was to make the packages variable mandatory and force the user to list the packages they want.

easybuild/easyblocks/l/lammps.py
```diff
@@ -0,0 +1,148 @@
...
+            'lammps',
+        ]
+
+        # build all packages
+        # skipped:
+        # * cuda
+        # * gpu
+        # * kim (requires external dep, see https://openkim.org)
+        # * 
+        for pkg in [p for p in os.listdir(libdir) if p not in ['README', 'cuda', 'gpu', 'kim', 'kokkos']]:
```
ocaisa: We removed other troublesome packages here: ```'molfile', 'voronoi', 'linalg'```

easybuild/easyblocks/l/lammps.py
```diff
@@ -0,0 +1,148 @@
...
+                # make sure active compiler is used
+                run_cmd('make -f %s CC="%s" CXX="%s" FC="%s"' % (makefile, cc, cxx, f90), log_output=True)
+
+            try:
+                os.chdir(srcdir)
+            except OSError as err:
+                raise EasyBuildError("Failed to change to %s: %s", srcdir, err)
+
+            self.log.info("Building %s package", pkg)
+            run_cmd("make yes-%s" % pkg, log_output=True)
```
ocaisa: With a mandatory package list you can move this step outside the loop:
```
for p in self.cfg['packages']:
   self.log.info("Building %s package", p)
   run_cmd("make yes-%s" % p, log_output=True)
```

ocaisa:
Working version of easyblock can be found at https://gist.github.com/ocaisa/bc46068cd74cea05f64f
Associated easyconfig can be found at https://gist.github.com/ocaisa/4289cdc747d752f74ae9

easybuild/easyblocks/l/lammps.py
```diff
@@ -0,0 +1,323 @@
...
+#
+# You should have received a copy of the GNU General Public License
+# along with EasyBuild.  If not, see <http://www.gnu.org/licenses/>.
+##
+"""
+EasyBuild support for building and installing LAMMPS, implemented as an easyblock
+
+@author: Kenneth Hoste (Ghent University)
+@author: Alexander Schnurpfeil (Juelich Supercomputer Centre)
+@author: Maxime Boissonneault (Compute Canada)
```
mboisson: @author: Ali Kerrache (Compute Canada)

easybuild/easyblocks/l/lammps.py
```diff
@@ -0,0 +1,323 @@
...
+        if self.cfg['build_shared_libs']:
+            shlib_ext = get_shared_lib_ext()
+            files.extend(['lib/liblammps.%s' % shlib_ext, 'lib/liblammps_%s.%s' % (build_type, shlib_ext)])
+
+        if self.cfg['build_static_libs']:
+            files.extend(['lib/liblammps.a', 'lib/liblammps_%s.a' % build_type])
+
+        if self.lammps_pylibdir:
+            files.append(os.path.join(self.lammps_pylibdir, 'lammps.py'))
+            custom_commands.append("python -c 'from lammps import lammps; lmp = lammps()'")
```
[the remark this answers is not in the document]
mboisson: For what it's worth, if I add `sanity_check_commands = ['']` to the easyconfig, it installs fine, and then if I load the module, the sanity check command runs just fine,
```[mboisson@build-node easyconfigs]$ python -c "from lammps import lammps; lmp = lammps()"
LAMMPS (11 Aug 2017)
OMP_NUM_THREADS environment is not set. Defaulting to 1 thread. (../comm.cpp:90)
  using 1 OpenMP thread(s) per MPI task
Total wall time: 0:00:00```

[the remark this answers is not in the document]
Comment (mboisson) [reply easybuild/easyblocks/l/lammps.py]: Ok, this is partly linked to https://github.com/easybuilders/easybuild-framework/issues/2499

However, this will still fail if python is a build dependency rather than a runtime dependency.

**Comment by ocaisa** [comment]
@mstud Looks like this could be related to https://github.com/easybuilders/easybuild-easyblocks/issues/1450

**Comment by mstud** [comment]
When the LAMMPS package "PYTHON" is enabled, this also fails if no Python-headers from the OS are installed, because of #1446 :
```
../pair_python.cpp:18:20: fatal error: Python.h: No such file or directory
 #include <Python.h>
                    ^
compilation terminated.
```

junghans:
This might simplify the build receipt:
https://sourceforge.net/p/lammps/mailman/message/36358633/

JackPerdue:
Well... I'm no big fan of easyblocks other than the toolchains and generic.  I think they hide too much away.  I think most things can be installed using ConfigureMake and CMakeMake (two really good easyblocks).

But I guess I get really irritated when easyblock authors don't provide the hooks to try to fix problems.  In this case its a prebuildopts that doesn't get run until the very end of the build (so I can't experiment with fixing the no Python.h problem there) and there is no preconfigopts.

If I were king of the forest I'd requre a preconfigopts/prebuildopts/preinstallopts before accepting a new easyblock.  

Anyway... attached is my suggested addtion for preconfigopts (which allows me to replace python3-config --include with pkg-config python3 --cflags)
[lammps_preconfigopts.diff.txt](https://github.com/easybuilders/easybuild-easyblocks/files/2299959/lammps_preconfigopts.diff.txt)

ocaisa:
Since this never got merged and there is now a pretty comprehensive [CMake procedure for LAMMPS](https://github.com/lammps/lammps/blob/master/cmake/README.md), maybe we should just take that route.

ocaisa:
Working example of LAMMPS easyconfig using `CmakeMake` available at [this Gist](https://gist.github.com/ocaisa/965ee383b4ec156b7e9fc0251411d8ec)

boegel:
So with @ocaisa's proof-of-concept in https://github.com/easybuilders/easybuild-easyconfigs/pull/6917, I guess the question becomes how relevant older LAMMPS versions are.

Can we just trash what we have here and only support sufficiently recent LAMMPS versions that are CMake-based?

I think @vanzod once mentioned he has extensive experience with LAMMPS, maybe he can pitch in here...

alikerr:
Hi @ocaisa ... In general, the new versions of LAMMPS still have all the packages from the older versions. There might be some exceptions and some of them are not supported anymore. There are also changes in the command lines (that are used in the input script). So, some users may need to use the same older version to reproduce data in rare cases. I think the CMake from the new versions should work for the old versions if the commands related to the recent packages were not included in the compilation. If you follow the changes from one version to another, you will see that there are new packages added. I would say, the EasyBlock lammps.py will be changing as the new versions appear: since the last stable version of 22Aug18, there are already two additional features (see Recent LAMMPS News ): https://lammps.sandia.gov/

vanzod:
I agree with @alikerr about the fact that the need of older versions is a rare event. `LAMMPS` tends to have additive releases. I think we can go ahead and only support `CMake` builds.

akesandgren:
@boegel Status of this?

boegel:
closing, superseded by https://github.com/easybuilders/easybuild-easyblocks/pull/1964

PR #596, opened 2015-04-22
tweak CP2K easyblock w.r.t. LAPACK/FFTW support

boegel:
required to get CP2K to build on Cray

### DIFF
--- a/easybuild/easyblocks/c/cp2k.py
+++ b/easybuild/easyblocks/c/cp2k.py
@@ -170,17 +170,19 @@ class EB_CP2K(EasyBlock):
         else:
             raise EasyBuildError("Don't know how to tweak configuration for compiler family %s" % comp_fam)
 
-        # BLAS related
+        # BLAS/FFTW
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/c/cp2k.py
```diff
@@ -314,7 +318,7 @@ def configure_common(self):
                                                        self.cfg['extracflags']),
             'DFLAGS': ' -D__parallel -D__BLACS -D__SCALAPACK -D__FFTSG %s' % self.cfg['extradflags'],
 
-            'LIBS': os.getenv('LIBS'),
+            'LIBS': os.getenv('LIBS', ''),
```
wpoely86: why have a default value here but not for `FFT_INC_DIR` etc?

boegel: LIBS may not be set, FFT_INC_DIR should always be set

but I can add a proper default there too, thanks for the suggestion

boegel:

PR #597, opened 2015-04-22
unset buildopts/installopts before installing Python extensions

boegel:


pforai:
Looks good, but could be generalized in future for other stuff where we have extensions like R, Perl, etc.

boegel:
I wouldn't generalise it unless it's really needed, and even then I would make it optional by introducing an extra easyconfig parameter to enable/disable it.

For it, only doing it here (where we have a use case for it) should suffice.

boegel:
@wpoely86: thoughts on this one?

rjeschmi:
It looks ok, I just wonder whether it indicates there should be a stronger separation between the Lang EB and it's extensions. Probably something to think about, but this looks fine for now.

easybuild/easyblocks/p/python.py
```diff
@@ -31,7 +31,7 @@
 @author: Pieter De Baets (Ghent University)
 @author: Jens Timmerman (Ghent University)
 """
-
+import copy
```
wpoely86: keep an empty line here?

boegel: there's no reason imho

wpoely86: maybe we really should have a style guide. Some easyblocks have a empty line, some don't.

boegel: if this is the main thing we're going to worry about, we should consider EB finished ;)

wpoely86:
Looks fine IMHO.

boegel:
Thanks for the review everyone!

easybuild/easyblocks/p/python.py
wpoely86: but but but but, we need something to bitch about! How can we consider this to be a true OSS project if we can't bitch about style?

boegel: ;-)

PR #598, opened 2015-04-23
Update GCC easyblock for GCC 5

wpoely86:


boegel:
Tested with https://github.com/hpcugent/easybuild-easyconfigs/pull/1553, looking great, going in.

Thanks @wpoely86!

easybuild/easyblocks/g/gcc.py
```diff
@@ -79,8 +79,8 @@ def __init__(self, *args, **kwargs):
             raise EasyBuildError("Using ISL bundled with CLooG is unsupported in >=GCC-4.8.0. "
                                  "Use a seperate ISL: set withisl=True")
 
-        # I think ISL without CLooG has no purpose in GCC...
-        if self.cfg['withisl'] and not self.cfg['withcloog']:
+        # I think ISL without CLooG has no purpose in GCC < 5.0.0 ...
+        if self.version < LooseVersion("5.0.0") and self.cfg['withisl'] and not self.cfg['withcloog']:
```
boegel: hmm, hold on...

why isn't this:

```python
if LooseVersion(self.version) < LooseVersion('5.0') ...
```

We should be using `LooseVersion` *everywhere* where we check versions.

It may work now, but that's just by accident, I fear...

boegel: Seems to work as expected without `LooseVersion` on both sides, but I don't trust it...

```
>>> from distutils.version import LooseVersion
>>> '4.9' < LooseVersion('5.0')
True
>>> '5.0' == LooseVersion('5.0')
True
>>> '5.1' < LooseVersion('5.0')
False
>>> '5.0' < LooseVersion('5.0')
False
>>> 
```

@wpoely86: please fix consistently throughout the entire easyblock before we merge this is?

boegel:
Going in, thanks @wpoely86!

PR #599, opened 2015-04-24
update Boost easyblock to support building with Cray toolchains (WIP)

boegel:
work by @pforai, but needs some cleanup


easybuild/easyblocks/b/boost.py
```diff
@@ -121,8 +121,33 @@ def configure_step(self):
             # http://www.boost.org/doc/libs/1_47_0/doc/html/mpi/getting_started.html
             # let Boost.Build know to look here for the config file
             f = open('user-config.jam', 'a')
-            f.write("using mpi : %s ;" % os.getenv("MPICXX"))
-            f.close()
+	    self.log.info("TC_CONSTANT_CRAYPE is %s " % (self.toolchain.TC_CONSTANT_CRAYPE))
```
pforai: This needs to be changed to the new cray code.

PR #600, opened 2015-04-25
vsc-base sources are optional, use self.initial_environ rather than self.orig_environ

boegel:
vsc-base sources are optional, so bootstrapping can be tested without providing vsc-base source tarball (see http://easybuild.readthedocs.org/en/latest/Installation.html#bootstrapping-using-supplied-source-tarballs)

`self.orig_environ` was replaced by `self.initial_environ` in framework

### DIFF
--- a/easybuild/easyblocks/e/easybuildmeta.py
+++ b/easybuild/easyblocks/e/easybuildmeta.py
@@ -49,7 +49,7 @@ class EB_EasyBuildMeta(PythonPackage):
     def __init__(self, *args, **kwargs):
         """Initialize custom class variables."""
         super(EB_EasyBuildMeta, self).__init__(*args, **kwargs)
-        self.orig_orig_environ = None
+        self.real_initial_environ = None
 
         self.easybuild_pkgs = ['easybuild-framework', 'easybuild-easyblocks', 'easybuild-easyconfigs']
         if LooseVersion(self.version) >= LooseVersion('2.0'):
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/e/easybuildmeta.py
```diff
@@ -180,10 +183,10 @@ def sanity_check_step(self):
         ]
 
         # (temporary) cleanse copy of original environment to avoid conflict with (potentially) loaded EasyBuild module
-        self.orig_orig_environ = copy.deepcopy(self.orig_environ)
+        self.orig_orig_environ = copy.deepcopy(self.initial_environ)
```
wpoely86: great name :smile:

easybuild/easyblocks/e/easybuildmeta.py
```diff
@@ -81,13 +81,16 @@ def install_step(self):
...
+                    # vsc-base sources are optional, can be pulled in from PyPi when installing easybuild-framework too
+                    if pkg != 'vsc-base':
+                        raise EasyBuildError("Failed to find EasyBuild %s package (subdirs: %s, seldirs: %s)",
+                                             pkg, subdirs, seldirs)
 
-                self.log.debug("Installing EasyBuild package %s" % pkg)
-                os.chdir(os.path.join(self.builddir, seldirs[0]))
-                super(EB_EasyBuildMeta, self).install_step()
+                else:
+                    self.log.debug("Installing EasyBuild package %s" % pkg)
```
wpoely86: shouldn't we output this to info?

easybuild/easyblocks/e/easybuildmeta.py
boegel: ok,ok, changed to `self.real_initial_environ` :)

easybuild/easyblocks/e/easybuildmeta.py
boegel: sure, why not

boegel:
going in, thanks for the review @wpoely86!

PR #602, opened 2015-04-26
update sanity check in R easyblock for v3.2

boegel:


boegel:
merging in as trivial, tested with https://github.com/hpcugent/easybuild-easyconfigs/pull/1570

PR #606, opened 2015-05-07
No need to check for compiler if the interfaces are not going to be b…

akesandgren:
There is no need to check for compiler(s) if the interfaces are not being built.
The patch should be small enough to be easily understood.


boegel:
Jenkins: ok to test

boegel:
For future reference: this fixes an issue that pops up when (re)build the FFTW interfaces of MKL is disabled *and* a dummy toolchain is used. According to what @akesandgren has told me, (re)building the FFTW interfaces is no longer need with recent version of MKL, they're readily provided... To be figured out in detail (cfr. #607).

In any case, this change makes perfect sense, thanks for tackling this @akesandgren!

Welcome to the wonderful world of EasyBuild, and congrats on your first contribution (one of many, hopefully). This may be the quickest contribution yet after getting to know the project (< 24h)...

PR #610, opened 2015-05-08
fix compatibility of easyblocks with --module-only + dedicated unit test module

boegel:
@wpoely86: can you take a look at this?


wpoely86:
looks fine

boegel:
retested builds that use modified easyblocks, and applied a couple of fixes, good to go now

Thanks @wpoely86!

PR #612, opened 2015-05-15
Easyblock for pbdMPI R package

pm67nz:


boegel:
@pm67nz: duplicate of #620, which seems to take a better approach (i.e. via `configure_step`, recognize more MPI types, only specify required `--with-mpi-type` since rest is derived correctly from enviroment)?

@ehiggs: thoughts?

ehiggs:
This is a good version. My PR also has pbdSLAP. Maybe merge mine and then push this on top?

boegel:
I collapsed this into #620, so this PR will be included once #620 is merged into develop.

boegel:
merged via #620

PR #615, opened 2015-05-28
make out-of-source build with CMake truly out-of-source

boegel:
setting `separate_build_dir = True` in an easyconfig using the `CMakeMake` easyblock results in executing the build procedure in a subdirectory of the unpacked sources

However, some packages (like CLHEP 2.2.x) require a true out-of-source build directory.

This fix should not break other easyconfigs that are already using `separate_build_dir = True` (I'll verify that before merging).

### DIFF
--- a/easybuild/easyblocks/generic/cmakemake.py
+++ b/easybuild/easyblocks/generic/cmakemake.py
@@ -68,13 +68,13 @@ class CMakeMake(ConfigureMake):
 
         default_srcdir = '.'
         if self.cfg.get('separate_build_dir', False):
-            objdir = 'easybuild_obj'
+            objdir = os.path.join(self.builddir, 'easybuild_obj')
             try:
                 os.mkdir(objdir)
                 os.chdir(objdir)
             except OSError, err:
                 raise EasyBuildError("Failed to create separate build dir %s in %s: %s", objdir, os.getcwd(), err)
-            default_srcdir = '..'
+            default_srcdir = self.cfg['start_dir']
 
         if srcdir is None:
             if self.cfg.get('srcdir', None) is not None:


wpoely86:
yeah should be good

wpoely86:
why again didn't we make this the default in EB 2.0?

boegel:
I don't know. Maybe because it's hard to deprecate the current behaviour?

wpoely86:
Maybe, but it really should be the default to do it out of tree.

boegel:
Sure, but we need to test carefully before we flip the default.

If the builds that now do not use an out-of-source build still work, I guess we can switch (we probably don't even have to wait until EB v3.0, then).

Currently, 12 easyblocks derive from `CMakeMake` (others may derive from those), and 32 easyconfigs specify they're using the generic `CMakeMake` easyblock directly...

If we carefully compile a list of things to test after switching the default, it's doable.

How about starting with a PR that changes the default (after this one is merged), and compiling a list of things to test there?

wpoely86:
good idea. lets have a look

boegel:
Retested on top of this, since they enable `separate_build_dir = True`:

* [ ] ANTs-2.1.0rc3-goolf-1.5.14.eb
* [ ] eXpress-1.5.1-goolf-1.4.10.eb
* [ ] GMT-5.1.0-goolf-1.4.10.eb
* [ ] getdp-2.5.0-intel-2015a-Python-2.7.9.eb
* [ ] gmsh-2.9.1-intel-2015a.eb
* [ ] Paraview-4.0.1-gmpolf-1.4.8.eb
* [ ] Paraview-4.1.0-gmpolf-1.4.8.eb
* [ ] Paraview-4.1.0-goolf-1.5.14.eb
* [ ] SIBELia-3.0.4-goolf-1.4.10.eb
* [ ] SPAdes-3.0.0-goolf-1.4.10.eb
* [ ] SPAdes-3.1.0-goolf-1.4.10.eb

(via `EB_OpenBabel`, which sets `self.cfg['separate_build_dir'] = True`)
* [ ] OpenBabel-2.3.2-goalf-1.5.12-no-OFED-Python-2.7.5.eb
* [ ] OpenBabel-2.3.2-ictce-4.1.13-Python-2.7.3.eb
* [ ] OpenBabel-2.3.2-ictce-5.5.0-Python-2.7.5.eb
* [ ] OpenBabel-2.3.2-ictce-7.1.2-Python-2.7.8.eb
* [ ] OpenBabel-2.3.2-intel-2014b-Python-2.7.8.eb

boegel:
No existing builds were harmed (see above), as expected, so good to go in.

PR #617, opened 2015-05-28
tweak easybuild.easyblocks package declaration to better support custom easyblocks repo

boegel:
Using `extend_path` to 'flatten' the easyblocks namespace results in better support for custom easyblocks repos, i.e. easyblocks that override existing easyblocks can be put in place easily.

This does require adding an (empty) `__init__.py` file in each letter-subdir.

There's one special case: because the R easyblock clashes with the `easybuild.easyblocks.r`, an icky `from ... import *` needs to be put in place in the `r/__init__.py` file...

For custom easyblock repos, including the following in the `easybuild/easyblocks/__init__.py` file should suffice with this in place:

```python
subdirs = [chr(l) for l in range(ord('a'), ord('z') + 1)] + ['0']
for subdir in subdirs:
    __path__ = extend_path(__path__, '%s.%s' % (__name__, subdir))

 __path__ = extend_path(__path__, __name__)
```

Although a custom easyblocks repo may not include letter-subdirs, the `extend_path` loop still needs to be there, because only the first `__init__.py` there appears in the Python search path is actually executed.

Preliminary testing shows this works as intended (see below). A unit test will be added to confirm this (and make sure we don't break it).

```
$ find /tmp/easybuild -type f              
/tmp/easybuild/__init__.py
/tmp/easybuild/easyblocks/__init__.py
/tmp/easybuild/easyblocks/icc.py

$ PYTHONPATH=/tmp:$PYTHONPATH python -c "from easybuild.easyblocks import VERSION; print VERSION"
2.2.0dev

$ PYTHONPATH=/tmp:$PYTHONPATH python -c "import easybuild.easyblocks; import easybuild.easyblocks.icc; print easybuild.easyblocks.icc; from easybuild.easyblocks.icc import EB_icc; print EB_icc.__module__; import easybuild.easyblocks.ifort; print easybuild.easyblocks.ifort"
<module 'easybuild.easyblocks.icc' from '/tmp/easybuild/easyblocks/icc.py'>
easybuild.easyblocks.icc
<module 'easybuild.easyblocks.ifort' from '/Users/kehoste/work/easybuild-easyblocks/easybuild/easyblocks/i/ifort.pyc'>

```

The reason this works is because `extend_path` basically scans `sys.path`, and thus considers *all* `easybuild/easyblocks` package directories that are included in the Python search path...

@JensTimmerman: please review?

### DIFF
--- a/easybuild/easyblocks/__init__.py
+++ b/easybuild/easyblocks/__init__.py
@@ -1,10 +1,5 @@
-# #
+##
 # Copyright 2009-2015 Ghent University
-# Copyright 2009-2015 Stijn De Weirdt
-# Copyright 2010 Dries Verdegem
-# Copyright 2010-2015 Kenneth Hoste
-# Copyright 2011 Pieter De Baets
-# Copyright 2011-2015 Jens Timmerman
 #
 # This file is part of EasyBuild,
 # originally created by the HPC team of Ghent University (http://ugent.be/hpc/en),
@@ -26,7 +21,16 @@
 #
 # You should have received a copy of the GNU General Public License
 # along with EasyBuild.  If not, see <http://www.gnu.org/licenses/>.
-# #
+##
+"""
+easybuild.easyblocks package declaration
+
+@author: Stijn De Weirdt (Ghent University)
+@author: Dries Verdegem (Ghent University)
+@author: Kenneth Hoste (Ghent University)
+@author: Pieter De Baets (Ghent University)
+@author: Jens Timmerman (Ghent University)
+"""
 import os
 from distutils.version import LooseVersion
 from pkgutil import extend_path
@@ -64,7 +68,9 @@ else:
 
 # Extend path so python finds our easyblocks in the subdirectories where they are located
 subdirs = [chr(l) for l in range(ord('a'), ord('z') + 1)] + ['0']
-__path__.extend([os.path.join(__path__[0], subdir) for subdir in subdirs])
+for subdir in subdirs:
+    __path__ = extend_path(__path__, '%s.%s' % (__name__, subdir))
+
 # And let python know this is not the only place to look for them, so we can have multiple
 # easybuild/easyblock paths in your python search path, next to the official easyblocks distribution
 __path__ = extend_path(__path__, __name__)  # @ReservedAssignment
--- a/easybuild/easyblocks/r/__init__.py
+++ b/easybuild/easyblocks/r/__init__.py
@@ -1,0 +1,30 @@
+##
+# Copyright 2009-2015 Ghent University
+#
+# This file is part of EasyBuild,
+# originally created by the HPC team of Ghent University (http://ugent.be/hpc/en),
+# with support of Ghent University (http://ugent.be/hpc),
+# the Flemish Supercomputer Centre (VSC) (https://vscentrum.be/nl/en),
+# the Hercules foundation (http://www.herculesstichting.be/in_English)
+# and the Department of Economy, Science and Innovation (EWI) (http://www.ewi-vlaanderen.be/en).
+#
+# http://github.com/hpcugent/easybuild
+#
+# EasyBuild is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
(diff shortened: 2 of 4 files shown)


JensTimmerman:
Does this also deal with the case where a custom easyblock repository has a custom `r.py` package?

boegel:
In that case, there's no problem, since the path to the custom easyblock repo will be first in the Python search path anyway:

```
$ find /tmp/easybuild -name '*.py'
/tmp/easybuild/__init__.py
/tmp/easybuild/easyblocks/__init__.py
/tmp/easybuild/easyblocks/icc.py
/tmp/easybuild/easyblocks/r.py

$ PYTHONPATH=/tmp:$PYTHONPATH python -c "import easybuild.easyblocks.r; print easybuild.easyblocks.r.__file__"
/tmp/easybuild/easyblocks/r.pyc
```

JensTimmerman:
ok, yes, this works because you use
`from easybuild.easyblocks.r.r import *` and not `.r.r import *`

JensTimmerman:
please add some kind of integration test for this.

boegel:
Current implementation of the test doesn't work when an `easybuild-easyblocks` repo that was installed with `easy_install`, and hence which is included in an `easy-install.pth` available through `$PYTHONPATH`, is around (as is the case on Jenkins).

The problem is that that repo gets precedence over whichever repo is listed first through `$PYTHONPATH`...

boegel:
Worked around possible complications from `easy-install.pth` by reconstructing only the required parts of the Python search path (i.e. easybuild-framework and vsc-base) into `$PYTHONPATH`.

Since the `easy-install.pth` trickery may make it very hard for people not intimately familiar with Python to set up a custom easyblocks repo such that theur easyblocs are preferred over the ones part of the EasyBuild installation, I'll look into adding support for a dedicated configuration option in the EasyBuild framework to specify the location of own easyblocks (and also toolchains, toolchain components and module naming schemes, while I'm at it). Cfr. https://github.com/hpcugent/easybuild-framework/issues/1010.

JensTimmerman:
You've only fixed this in the tests, so, in real world situations this issue will still pop up?

boegel:
In real world situations, we need something like `eb --my-easyblocks=/foo/*.py`, which just injects a location for a custom easyblocks repo straight into `sys.path`. We can also take care of the magic that needs to be included in `easybuild/easyblocks/__init__.py`, in that case.

Setting this up via `$PYTHONPATH` is clearly not going to work out in practice, there are too many possible complications.

The changes here are orthogonal to that though, and are still required to make things work as expected.

JensTimmerman:
So the first 'untweaked' testcase is also broken in master?

boegel:
@JensTimmerman: not sure what you mean here... If you use `$PYTHONPATH` now to make `eb` consider extra easyblocks, it'll work, the trouble with `easy-install.pth` is only relevant w.r.t. giving own easyblocks priority over existing ones (which doesn't work with current master due to the way we flatten the easyblocks namespace in `__init__.py`).

JensTimmerman:
My question is: if you take the unittest before you tweaked it to actually work with a workaround, does that also fail on the current master branch of easybuild?

boegel:
The unit test tests whether custom easyblocks can overrule an existing easyblock, so yes, it would fail on master. Making that possible is exactly the point of this PR.

boegel:
BTW, the workaround for the failing test is in the test itself (I basically avoid running into an `easy-install.pth` for the `easybuild.framework` and `vsc` namespaces by composing `$PYTHONPATH` myself.

JensTimmerman:
Can you then please add a test that also tests the current behaviour? Where you can just add an easyblock to the path, and see that it's picked up?

boegel:
@JensTimmerman: it's doing that (too) already, see https://github.com/hpcugent/easybuild-easyblocks/pull/617/files#diff-6c1ff9f6cc8975b3ffa9e414bddf0e10R148

JensTimmerman:
I guess it's fine then

boegel:
documentation will be updated along with https://github.com/hpcugent/easybuild-framework/pull/1301

PR #618, opened 2015-06-01
implement custom implementation of obtain_file method in QuantumESPRESSO easyblock

boegel:


boegel:
@wpoely86, @JensTimmerman: thoughts on this?

I'm not sure how happy I am with this, but it does fix the issue at hand.

boegel:
See also https://github.com/hpcugent/easybuild-easyconfigs/pull/1644

easybuild/easyblocks/q/quantumespresso.py
```diff
@@ -62,6 +63,59 @@ def __init__(self, *args, **kwargs):
...
+            path = super(EB_QuantumESPRESSO, self).obtain_file(*args, **kwargs)
+        except EasyBuildError, err:
+            self.log.debug("Ignoring expected download error: %s", err)
+
+        # restore source_urls value
+        self.cfg['source_urls'] = source_urls
+
+        if path:
+            self.log.info("Source file %s already there at %s, assuming it's OK.", filename, path)
+
```
wpoely86: remove empty line

wpoely86:
so what is the problem here? It returns a html page instead of a 404 ?

boegel:
@wpoely86: yes, for example: http://www.qe-forge.org/gf/download/frsrelease/185/753/GWW-5.1.2.tar.gz

boegel:
@wpoely86: yes, for example: http://www.qe-forge.org/gf/download/frsrelease/185/753/GWW-5.1.2.tar.gz (vs http://www.qe-forge.org/gf/download/frsrelease/185/754/GWW-5.1.2.tar.gz, which is the valid URL)

wpoely86:
Is this not worth adding the framework then?

boegel:
@wpoely86: maybe, I'm not sure, this is the first time we run into an issue like this, afaik

easybuild/easyblocks/q/quantumespresso.py
```diff
@@ -62,6 +63,59 @@ def __init__(self, *args, **kwargs):
...
+        filename = args[0]
+        source_urls = self.cfg['source_urls']
+
+        # check whether file is already there, by running obtain_file without providing any source_urls
+        path = None
+        self.cfg['source_urls'] = []
+        try:
+            path = super(EB_QuantumESPRESSO, self).obtain_file(*args, **kwargs)
+        except EasyBuildError, err:
+            self.log.debug("Ignoring expected download error: %s", err)
```
JensTimmerman: you're ignoring every EasyBuildError, not just download errors?

JensTimmerman:
This should probably be fixed somewhere in the framework, instead of just this EasyBlock?

easybuild/easyblocks/q/quantumespresso.py
```diff
@@ -62,6 +63,59 @@ def __init__(self, *args, **kwargs):
...
+            # put copy of provided source_urls in place to play around with
+            kwargs.update({'urls': source_urls[:]})
+
+            # if source is not available yet, try to download it
+            path = super(EB_QuantumESPRESSO, self).obtain_file(*args, **kwargs)
+
+            # check whether downloaded tarball can be unpacked
+            _, ec = run_cmd("tar tfz %s" % path, simple=False, log_ok=False, log_all=False)
+
+            # download may result in an HTML page when wrong source URL was used (rather than a proper 404 HTML code)
```
JensTimmerman: woudn't the md5sum pick this up?

JensTimmerman:
Why is this special workaround needed? Can't the md5sum pick this failure to download up?

easybuild/easyblocks/q/quantumespresso.py
wpoely86: only if it is set, which is still a minority of the easyconfigs.

wpoely86:
I agree with @JensTimmerman here. Should be handled in the framework. Does a simple check on the return mimetype not suffice to check whether we get html or something else?

easybuild/easyblocks/q/quantumespresso.py
JensTimmerman: sure, but it's a lot faster and easier to set the md5sum for the sources here then to try to fix it this way?

JensTimmerman:
The problem is you then have to specify in the easyconfig what mime type you are expecting, the extension doesn't really mean anything anyway, but of course, it could be used as an indicator, I have some code in a branch that checks magic numbers, so we could spit out warnings if magic numbers don't match up with the extension...

We could do that, but since we have the checksum infrastructure in place already I suggest we use that?

wpoely86:
Why not do both? Checksums are the best option but it's nice to have a fallback.

Do we actually need the magic numbers stuff? Do we have any cases where the content-type is `text/html` of a valid download link?

easybuild/easyblocks/q/quantumespresso.py
boegel: If there's an MD5sum available, it will detect this issue, yes, but simply complain that things don't match.

Checking the MD5sum if you know that the download is faulty is pretty useless.

boegel:
The problem here is that without having this in place, the download for QuantumESPRESSO can never work.

The non-generic URLs used in https://github.com/hpcugent/easybuild-easyconfigs/pull/1644 don't yield a 404 if a download is attempted that is not available at that URL; the generic URL does, but not everything can be downloaded that way.

So, without this workaround, auto-download for QuantumESPRESSO is simply broken, and there's no way around it by reordering the `source_urls`.

I agree that this fix can be implemented more generally, and maybe it should (and be disabled by default).

But we've never run into an issue like this, i.e. source URLs that do not properly yield a 404 when attempting to download something that isn't there. So this problem is specific to QuantumESPRESSO, for now.

easybuild/easyblocks/q/quantumespresso.py
JensTimmerman: the sum only get's checked after the download stage, not after each individual download?

JensTimmerman:
My point about the md5sum is that this is the way to figure out if a download is correct or not, So can the general framework code be changed to check the md5sum after a download, and if that's one wrong, move on to the next source url?

boegel:
What if checksums are not available? (as they are now, in QuantumESPRESSO easyconfigs)?
Or when `--try-software-version` is used (checksums are cleared then, since they're guaranteed to be broken)?

JensTimmerman:
then you have to add the checksums to the easyconfig, or fix it in the framework, instead of trying to patch it in in the easyblock?

ocaisa:
With #3257, we're moving to a whole new easyblock for QE, so closing this

PR #622, opened 2015-06-05
update clang to handle unlimited stack size

wpoely86:
Causes the Thread sanitizer to fail

### DIFF
--- a/easybuild/easyblocks/c/clang.py
+++ b/easybuild/easyblocks/c/clang.py
@@ -166,13 +166,24 @@ class EB_Clang(CMakeMake):
             self.llvm_obj_dir_stage3 = os.path.join(self.builddir, 'llvm.obj.3')
 
         if LooseVersion(self.version) >= LooseVersion('3.3'):
+            disable_san_tests = False
             # all sanitizer tests will fail when there's a limit on the vmem
             # this is ugly but I haven't found a cleaner way so far
             (vmemlim, ec) = run_cmd("ulimit -v", regexp=False)
             if not vmemlim.startswith("unlimited"):
+                disable_san_tests = True
                 self.log.warn("There is a virtual memory limit set of %s KB. The tests of the "
                               "sanitizers will be disabled as they need unlimited virtual "
                               "memory." % vmemlim.strip())
+
+            # the same goes for unlimited stacksize
+            (stacklim, ec) = run_cmd("ulimit -s", regexp=False)
+            if stacklim.startswith("unlimited"):
+                disable_san_tests = True
+                self.log.warn("The stacksize limit is set to unlimited. This causes the ThreadSanitizer "
+                              "to fail. The sanitizers tests will be disabled.")
+
+            if disable_san_tests:
                 self.disable_sanitizer_tests()
 
         # Create and enter build directory.
--- a/easybuild/easyblocks/c/clang.py
+++ b/easybuild/easyblocks/c/clang.py
@@ -42,7 +42,9 @@ from distutils.version import LooseVersion
 
 from easybuild.easyblocks.generic.cmakemake import CMakeMake
 from easybuild.framework.easyconfig import CUSTOM
+from easybuild.tools import run
 from easybuild.tools.build_log import EasyBuildError
+from easybuild.tools.config import build_option
 from easybuild.tools.filetools import mkdir
 from easybuild.tools.modules import get_software_root
 from easybuild.tools.run import run_cmd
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/c/clang.py
```diff
@@ -166,13 +166,24 @@ def configure_step(self):
...
                               "memory." % vmemlim.strip())
+
+            # the same goes for unlimited stacksize
+            (stacklim, ec) = run_cmd("ulimit -s", regexp=False)
+            if stacklim.startswith("unlimited"):
+                disable_san_tests = True
+                self.log.warn("The stacksize limit is set to unlimited. This causes the ThreadSanitizer "
+                              "to fail. The sanitizers tests will be disabled.")
+
+            if disable_san_tests:
```
fgeorgatos: @wpoely86: I think the above code should be conditional upon a "strict_checking" boolean tunable (configfile?).
The meaning of this remark is, to avoid hiding an issue under the carpet, since knowing about it may trigger a policy change elsewhere!

fgeorgatos: thanks for the PR, in any case.

wpoely86: I'm afraid I don't quite follow what you mean @fgeorgatos ? If we don't do this, the test suite will fail for sure.

fgeorgatos: in terms of systems administration, we frown upon the concept that two runs of a tool across two sites may yield silently different results. So, by making it a tunable you restore responsibility on the fellow invoking the tool to make his choices and, hence, trust of the process. It will no longer have -in your words- "ugliness"!

wpoely86: Well, this only concerns the sanitizers, not any of the core parts. In EB terms, the sanitizer should never be used. There not meant for production use.

And have a limited virtual memory and unlimited stack size are not the default on any distro that I know. Admins have to manually change this to get that. And if you do, that for whatever reason, there are consequences. We spit out a warning about it?

boegel: I wouldn't mind having a custom easyconfig parameter like `disable_san_tests_based_on_ulimit` in place to be able to disable this 'magic', but it should be `True` by default imho.

Shouldn't be too hard @wpoely86?

wpoely86: Fine by me but I think @fgeorgatos was thinking about a global config option?

boegel: You mean `--strict` (which we have already)?

boegel: So, no need for a custom option then! ;-)

boegel: We'll need to make `strict` a build option though...

wpoely86: OK, you take care of that?

fgeorgatos: @wpoely86, @boegel: yes, I originally meant something in the lines of `--strict`; 
I'm following your discussion and I let it up to you to decide how it should be like!

boegel: @wpoely86: https://github.com/hpcugent/easybuild-framework/pull/1295

easybuild/easyblocks/c/clang.py
```diff
@@ -166,13 +166,24 @@ def configure_step(self):
             self.llvm_obj_dir_stage3 = os.path.join(self.builddir, 'llvm.obj.3')
 
         if LooseVersion(self.version) >= LooseVersion('3.3'):
```
boegel: add `and not build_option('strict') == run.ERROR` (requires https://github.com/hpcugent/easybuild-framework/pull/1295)

wpoely86: on this line? I would the let it spit out the warning but just do nothing?

boegel: You're going to warn that you'll disable the tests, and then don't do it? ;-)

wpoely86: Yeah, great idea no? :laughing: 

I would change the warning a bit (add: unless strict or something).

boegel: unless `--strict=error` is used

easybuild/easyblocks/c/clang.py
```diff
@@ -166,13 +168,24 @@ def configure_step(self):
...
+                              "memory unless --strict=error is used." % vmemlim.strip())
+
+            # the same goes for unlimited stacksize
+            (stacklim, ec) = run_cmd("ulimit -s", regexp=False)
+            if stacklim.startswith("unlimited"):
+                disable_san_tests = True
+                self.log.warn("The stacksize limit is set to unlimited. This causes the ThreadSanitizer "
+                              "to fail. The sanitizers tests will be disabled unless --strict=error is used.")
+
+            if disable_san_tests and not build_option('strict') == run.ERROR:
```
boegel: maybe use `!=` rather than `not ... == ...`?

boegel:
Tested with a Clang build that used to fail because of `ulimit -s`, now works fine.

Thanks @wpoely86!

PR #623, opened 2015-06-05
Easyblock for MCR (REVIEW)

hajgato:


boegel:
@hajgato: is this still WIP? If so, please mark it in the subject.

hajgato:
@boegel :)

easybuild/easyblocks/m/mcr.py
```diff
@@ -0,0 +1,146 @@
...
+                       }
+
+        super(EB_MCR, self).sanity_check_step(custom_paths=custom_paths)
+
+    def make_module_extra(self):
+        """Extend PATH and set proper _JAVA_OPTIONS (e.g., -Xmx)."""
+
+        txt = super(EB_MCR, self).make_module_extra()
+
+        txt += self.module_generator.set_environment('XAPPLRESDIR', os.path.join(self.installdir, self.extradir[self.cfg['version']], 'X11', 'app-defaults'))
```
boegel: @hajgato: please introduce a variable for the extradir, and fetch the value in a way that it's 'safe' in case the version is not known yet; also, use `self.version` rather than `self.cfg['version']`

```python
extradir = self.extradir.get(self.version, 'UNKNOWN')
```

easybuild/easyblocks/m/mcr.py
```diff
@@ -0,0 +1,146 @@
...
+
+class EB_MCR(EasyBlock):
+    """Support for installing MCR."""
+
+    def __init__(self, *args, **kwargs):
+        """Add extra config options specific to MCR."""
+        super(EB_MCR, self).__init__(*args, **kwargs)
+        self.comp_fam = None
+        self.configfilename = "my_installer_input.txt"
+        self.extradir =  {'R2014a': 'v83', 'R2014b': 'v84', 'R2015a': 'v85'}
```
boegel: at this point, you already know the version, so just use this:

```python
```

easybuild/easyblocks/m/mcr.py
```diff
@@ -0,0 +1,146 @@
...
+                config = regagree.sub("agreeToLicense=Yes", config)
+                config = regmode.sub("mode=silent", config)
+            else:
+                config = "destinationFolder=%s\n" % self.installdir
+                config += "agreeToLicense=Yes\n"
+                config += "mode=silent\n"
+
+            f = open(configfile, 'w')
+            f.write(config)
+            f.close()
```
boegel: use `write_file` function from `easybuild.tools.filetools`

easybuild/easyblocks/m/mcr.py
```diff
@@ -0,0 +1,146 @@
...
+##
+"""
+EasyBuild support for installing MCR, implemented as an easyblock
+
+@author: Stijn De Weirdt (Ghent University)
+@author: Dries Verdegem (Ghent University)
+@author: Kenneth Hoste (Ghent University)
+@author: Pieter De Baets (Ghent University)
+@author: Jens Timmerman (Ghent University)
+@author: Fotis Georgatos (Uni.Lu, NTUA)
```
boegel: is this a new easyblock? if so, you're the only author

easybuild/easyblocks/m/mcr.py
```diff
@@ -0,0 +1,146 @@
+##
+# Copyright 2009-2015 Ghent University
```
boegel: 2015-2015?

easybuild/easyblocks/m/mcr.py
```diff
@@ -0,0 +1,146 @@
...
+        """Add extra config options specific to MCR."""
+        super(EB_MCR, self).__init__(*args, **kwargs)
+        self.comp_fam = None
+        self.configfilename = "my_installer_input.txt"
+        self.extradir =  {'R2014a': 'v83', 'R2014b': 'v84', 'R2015a': 'v85'}
+
+    @staticmethod
+    def extra_options():
+        extra_vars = {
+            'java_options': ['-Xmx256m', "$_JAVA_OPTIONS value set for install and in module file.", CUSTOM],
```
boegel: the `256m` value looks arbitrary, at least add a comment why this default makes sense?

easybuild/easyblocks/m/mcr.py
```diff
@@ -0,0 +1,146 @@
...
+        return EasyBlock.extra_options(extra_vars)
+
+    def configure_step(self):
+        """Configure MCR installation: create license file."""
+
+        configfile = os.path.join(self.builddir, self.configfilename)
+        try:
+            if LooseVersion(self.version) < LooseVersion('2015a'):
+                shutil.copyfile("%s/installer_input.txt" % self.builddir, configfile)
+                config = file(configfile).read()
```
boegel: use `read_file` function

easybuild/easyblocks/m/mcr.py
```diff
@@ -0,0 +1,146 @@
...
+                regdest = re.compile(r"^# destinationFolder=.*", re.M)
+                regagree = re.compile(r"^# agreeToLicense=.*", re.M)
+                regmode = re.compile(r"^# mode=.*", re.M)
```
boegel: no need to compile these separately, since they're used only once:

```python
config = re.sub(r"^# destinationFolder=.*", "destinationFolder=%s" % self.installdir, config, re.M)
...
```

easybuild/easyblocks/m/mcr.py
```diff
@@ -0,0 +1,146 @@
...
+        configfile = "%s/%s" % (self.builddir, self.configfilename)
+        cmd = "%s ./install -v -inputFile %s %s" % (self.cfg['preinstallopts'], configfile, self.cfg['installopts'])
+        run_cmd(cmd, log_all=True, simple=True)
+
+    def sanity_check_step(self):
+        """Custom sanity check for MCR."""
+
+        custom_paths = {
+                        'files': [''],
+                        'dirs': ['%s/%s/glnxa64' % (self.extradir[self.cfg['version']], x) for x in ['runtime' , 'bin', 'sys/os']],
```
boegel: fix indet

easybuild/easyblocks/m/mcr.py
```diff
@@ -0,0 +1,146 @@
...
+    def make_module_extra(self):
+        """Extend PATH and set proper _JAVA_OPTIONS (e.g., -Xmx)."""
+
+        txt = super(EB_MCR, self).make_module_extra()
+
+        txt += self.module_generator.set_environment('XAPPLRESDIR', os.path.join(self.installdir, self.extradir[self.cfg['version']], 'X11', 'app-defaults'))
+        for ldlibdir in ['runtime', 'bin', os.path.join('sys', 'os')]:
+            txt += self.module_generator.prepend_paths('LD_LIBRARY_PATH', os.path.join(self.extradir[self.cfg['version']], ldlibdir, 'glnxa64'))
+
+
```
boegel: avoid multiple empty line

easybuild/easyblocks/m/mcr.py
```diff
@@ -0,0 +1,146 @@
...
+        txt += self.module_generator.set_environment('XAPPLRESDIR', os.path.join(self.installdir, self.extradir[self.cfg['version']], 'X11', 'app-defaults'))
+        for ldlibdir in ['runtime', 'bin', os.path.join('sys', 'os')]:
+            txt += self.module_generator.prepend_paths('LD_LIBRARY_PATH', os.path.join(self.extradir[self.cfg['version']], ldlibdir, 'glnxa64'))
+
+        
+        txt += self.module_generator.set_environment('_JAVA_OPTIONS', self.cfg['java_options'])
+
+        return txt
+
```
boegel: drop trailing empty line

easybuild/easyblocks/m/mcr.py
hajgato: hmm, I do not understand it. The `self.version` would be something like R2015b,and the extradir should be `v86` or something. Should we directly define `extradir` in the easyconfig file?

boegel: It's OK to make the mapping from version to subdir in the easyblock, but to make this easyblock compatible with `--module-only` and avoid hard crashes if a new version is attempted, use `self.extradir.get` rather than `self.extradir[..]`

easybuild/easyblocks/m/mcr.py
hajgato: Well, it is taken from matlab easyblock, and heavily simplified, so I am not really the author.....

boegel: Yeah, I somehow missed that.

You're not the original author, sure, but since you have modified it to make it work for MCR, you should add yourself as an author here.

easybuild/easyblocks/m/mcr.py
hajgato: Taken form Matlab easyblock, dunno why that was there chosen. I though that might be useful here as well...

boegel:
Tested with https://github.com/hpcugent/easybuild-easyconfigs/pull/1677, working as intended.

Going in, thanks @hajgato!

PR #625, opened 2015-06-12
Add run_cmd_qa as run_cmd function when 'qa' parameter is used in .eb…

npklein:
… file.

Part of adding question/answer  functionallity during command line installation to install Anaconda

### DIFF
--- a/easybuild/easyblocks/generic/binary.py
+++ b/easybuild/easyblocks/generic/binary.py
@@ -41,6 +41,7 @@ from easybuild.framework.easyconfig import CUSTOM
 from easybuild.tools.build_log import EasyBuildError
 from easybuild.tools.filetools import mkdir, rmtree2
 from easybuild.tools.run import run_cmd
+from easybuild.tools.run import run_cmd_qa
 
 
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/generic/binary.py
```diff
@@ -106,8 +107,10 @@ def install_step(self):
         else:
             cmd = ' '.join([self.cfg['preinstallopts'], self.cfg['install_cmd'], self.cfg['installopts']])
             self.log.info("Installing %s using command '%s'..." % (self.name, cmd))
-            run_cmd(cmd, log_all=True, simple=True)
-
+            if self.cfg['qa'] is None:
+                run_cmd(cmd, log_all=True, simple=True)
+            else:
+                run_cmd_qa(cmd, self.cfg['qa'], log_all=True, simple=True)
```
boegel: you can define `qa` as a custom easyconfig parameter specific to `Binary`, via the `extra_options` method in this easyblock

boegel:
Jenkins: ok to test

PR #631, opened 2015-06-17
don't restrict det_pylibdir to only EB-provided Python, there's no reason to

boegel:
@wpoely86: please review

### DIFF
--- a/easybuild/easyblocks/generic/pythonpackage.py
+++ b/easybuild/easyblocks/generic/pythonpackage.py
@@ -42,7 +42,7 @@ from easybuild.easyblocks.python import EXTS_FILTER_PYTHON_PACKAGES
 from easybuild.framework.easyconfig import CUSTOM
 from easybuild.framework.extensioneasyblock import ExtensionEasyBlock
 from easybuild.tools.build_log import EasyBuildError
-from easybuild.tools.filetools import mkdir, rmtree2
+from easybuild.tools.filetools import mkdir, rmtree2, which
 from easybuild.tools.modules import get_software_root, get_software_version
 from easybuild.tools.run import run_cmd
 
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/generic/pythonpackage.py
```diff
@@ -53,27 +53,24 @@
...
-            raise EasyBuildError("Last line of output of %s does not start with specified prefix %s: %s (exit code %s)",
-                                 cmd, prefix, out, ec)
-
-        pylibdir = txt[len(prefix):]
-        log.debug("Determined pylibdir using '%s': %s", cmd, pylibdir)
-        return pylibdir
+
+    # determine Python lib dir via distutils
+    # use run_cmd, we can to talk to the active Python, not the system Python running EasyBuild
+    prefix = '/tmp/'
```
wpoely86: don't hard code this, use the EB provided tmp path.

easybuild/easyblocks/generic/pythonpackage.py
```diff
@@ -53,27 +53,24 @@
...
-        if not txt.startswith(prefix):
-            raise EasyBuildError("Last line of output of %s does not start with specified prefix %s: %s (exit code %s)",
-                                 cmd, prefix, out, ec)
-
-        pylibdir = txt[len(prefix):]
-        log.debug("Determined pylibdir using '%s': %s", cmd, pylibdir)
-        return pylibdir
+
+    # determine Python lib dir via distutils
+    # use run_cmd, we can to talk to the active Python, not the system Python running EasyBuild
```
wpoely86: shouldn't we at least print out a debugging message about which python is being used?

easybuild/easyblocks/generic/pythonpackage.py
boegel: this doesn't create any files, it just queries what the path would look like with this prefix

easybuild/easyblocks/generic/pythonpackage.py
boegel: I guess it doesn't hurt to include it, will do

boegel:

PR #632, opened 2015-06-29
Support snappy and other optional native libs in Hadoop.

ehiggs:


easybuild/easyblocks/h/hadoop.py
```diff
@@ -28,10 +28,12 @@
 @author: Kenneth Hoste (Ghent University)
 """
 import os
+import os.path
```
boegel: this is not needed if `os` was already imported?

easybuild/easyblocks/h/hadoop.py
```diff
@@ -43,13 +45,19 @@ def extra_options():
...
         return Tarball.extra_options(extra_vars)
 
     def build_step(self):
         """Custom build procedure for Hadoop: build native libraries, if requested."""
         if self.cfg['build_native_libs']:
-            cmd = "mvn package -DskipTests -Dmaven.javadoc.skip -Dtar -Pdist,native"
+            # Building snappy, bzip2 jars w/ native libs requires -Drequire.snappy -Drequire.bzip2, etc.
+            native_lib_flags = ""
+            for native_lib, lib_path in self.cfg['extra_native_libs']:
+                native_lib_flags += '-Drequire.%s=true ' % native_lib
```
boegel: why not leave `cmd` like it was, and append to it instead here?

```python
    cmd += ' -Drequire.%s=true' % native_lib
```

easybuild/easyblocks/h/hadoop.py
```diff
@@ -62,6 +70,17 @@ def install_step(self):
         else:
             super(EB_Hadoop, self).install_step()
 
+    def post_install_step(self):
+        """After the install, copy libsnappy into place."""
+        for native_library, lib_path in self.cfg['extra_native_libs']:
+            lib_root = get_software_root(native_library)
+            if not lib_root:
+                self.log.warn("%s not found. Skipping install" % native_library)
+                continue
```
boegel: no, fail if you were asked to include extra native libraries with the required dep missing

```python
    raise EasyBuildError("Dependency %s is required for specified extra native libs, but not found" % native_library)
```

easybuild/easyblocks/h/hadoop.py
```diff
@@ -62,6 +70,17 @@ def install_step(self):
...
+            lib_src = os.path.join(lib_root, lib_path)
+            lib_dest = os.path.join(self.installdir, 'lib')
+            shutil.copytree(lib_src, lib_dest)
```
boegel: wrap this in `try, except OSerror`, and raise `EasyBuildError` in the `except` bit (we actually need a proper function for that in `easybuild.tools.filetools`)

easybuild/easyblocks/h/hadoop.py
```diff
@@ -62,6 +70,16 @@ def install_step(self):
         else:
             super(EB_Hadoop, self).install_step()
 
+    def post_install_step(self):
+        """After the install, copy libsnappy into place."""
+        for native_library, lib_path in self.cfg['extra_native_libs']:
+            lib_root = get_software_root(native_library)
+            if not lib_root:
+                raise EasyBuildError("%s not found. Skipping install" % native_library)
```
boegel: fix error message + move the check up to `build_step`

boegel:
tested with https://github.com/hpcugent/easybuild-easyconfigs/pull/1758, working as intended now, thanks @ehiggs!

PR #633, opened 2015-06-30
Add easyblock to wrap system compilers

geimer:
This easyblock can, in conjunction with some fairly minimal easyconfigs (separate PR will follow), create EasyBuild-compatible modulefiles for GCC and Intel compilers that are already installed on the system by some other means (distribution package or by running the installer manually).

### DIFF
--- a/easybuild/easyblocks/generic/systemcompiler.py
+++ b/easybuild/easyblocks/generic/systemcompiler.py
@@ -1,5 +1,5 @@
 ##
-# Copyright 2009-2015 Ghent University
+# Copyright 2015-2015 Ghent University
 #
 # This file is part of EasyBuild,
 # originally created by the HPC team of Ghent University (http://ugent.be/hpc/en),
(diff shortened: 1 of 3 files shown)


boegel:
Jenkins: ok to test

easybuild/easyblocks/generic/systemcompiler.py
```diff
@@ -0,0 +1,145 @@
+##
+# Copyright 2009-2015 Ghent University
```
boegel: 2015-2015

geimer: 2015-2015 or just 2015?

boegel: 2015-2015, easier to update in the future ;-)

but let me send a PR your way for this PR, it'll include this too

boegel:
Tested with https://github.com/hpcugent/easybuild-easyconfigs/issues/1778, passes unit tests (now that an exception is in place for this SystemCompiler easyblocks), looking good.

Going in, thanks @berndmohr and @geimer!

PR #638, opened 2015-07-03
.so files need to be in $EBROOTHADOOP/lib/native.

ehiggs:
native libraries in hadoop need to go in `$EBROOTHADOOP/lib/native`.

Tested.

```
$ ls $EBROOTHADOOP/Hadoop/2.5.0-cdh5.3.1-native/lib/native/
libhadoop.a	  libhadoop.so	      libhadooputils.a	libhdfs.so	  libnativetask.a   libnativetask.so.1.0.0
libhadooppipes.a  libhadoop.so.1.0.0  libhdfs.a		libhdfs.so.0.0.0  libnativetask.so  libsnappy.so
```

### DIFF
--- a/easybuild/easyblocks/h/hadoop.py
+++ b/easybuild/easyblocks/h/hadoop.py
@@ -78,7 +78,7 @@ class EB_Hadoop(Tarball):
         for native_library, lib_path in self.cfg['extra_native_libs']:
             lib_root = get_software_root(native_library)
             lib_src = os.path.join(lib_root, lib_path)
-            lib_dest = os.path.join(self.installdir, 'lib')
+            lib_dest = os.path.join(self.installdir, 'lib', 'native')
             shutil.copy2(lib_src, lib_dest)
 
     def sanity_check_step(self):
--- a/easybuild/easyblocks/h/hadoop.py
+++ b/easybuild/easyblocks/h/hadoop.py
@@ -106,7 +106,7 @@ class EB_Hadoop(Tarball):
         not_found = []
         installdir = os.path.realpath(self.installdir)
         lib_src = os.path.join([installdir, 'lib', 'native'])
-        for native_lib, lib_path  in self.cfg['extra_native_libs']:
+        for native_lib, _ in self.cfg['extra_native_libs']:
             if not re.search(r'%s: *true *%s' % (native_lib, lib_src), out):
                 not_found.append(native_lib)
         if not_found:
--- a/easybuild/easyblocks/h/hadoop.py
+++ b/easybuild/easyblocks/h/hadoop.py
@@ -27,7 +27,9 @@ EasyBuild support for building and installing Hadoop, implemented as an easybloc
 
 @author: Kenneth Hoste (Ghent University)
 """
+import glob
 import os
+import re
 import shutil
 
 from easybuild.easyblocks.generic.tarball import Tarball
@@ -79,7 +81,8 @@ class EB_Hadoop(Tarball):
             lib_root = get_software_root(native_library)
             lib_src = os.path.join(lib_root, lib_path)
             lib_dest = os.path.join(self.installdir, 'lib', 'native')
-            shutil.copy2(lib_src, lib_dest)
+            for lib in glob.glob(os.path.join(lib_src, '*.so*')):
+                shutil.copy2(lib, lib_dest)
 
     def sanity_check_step(self):
         """Custom sanity check for Hadoop."""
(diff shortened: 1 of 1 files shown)


boegel:
@ehiggs: is there a way to check which native libraries Hadoop is aware of?

Maybe we can include some kind of self-check in `sanity_check_step`?

boegel:
We should add a check on the output of:
```
hadoop checknative -a
```

easybuild/easyblocks/h/hadoop.py
```diff
@@ -93,6 +96,19 @@ def sanity_check_step(self):
...
+        out, _ = run_cmd(cmd, log_all=True, simple=False, log_ok=True)
+        self.clean_up_fake_module(fake_mod_data)
+
+        not_found = []
+        for lib, lib_path  in self.cfg['extra_native_libs']:
+            if not re.search(r'%s: *true' % lib, out):
+                not_found.append(lib)
+        if not_found:
+            raise EasyBuildError("%s not found by 'hadoop checknative -a'" %
+                    ', '.join(not_found))
```
boegel: no need to spread this across two lines? also, delay string interpolation until error actually happens:

```python
raise EasyBuildError("%s not found by 'hadoop checknative -a'", ', '.join(not_found))
```

easybuild/easyblocks/h/hadoop.py
```diff
@@ -93,6 +96,19 @@ def sanity_check_step(self):
...
         super(EB_Hadoop, self).sanity_check_step(custom_paths=custom_paths)
 
+        fake_mod_data = self.load_fake_module(purge=True)
+        cmd = "hadoop checknative -a"
+        out, _ = run_cmd(cmd, log_all=True, simple=False, log_ok=True)
+        self.clean_up_fake_module(fake_mod_data)
+
+        not_found = []
+        for lib, lib_path  in self.cfg['extra_native_libs']:
+            if not re.search(r'%s: *true' % lib, out):
```
boegel: shouldn't we also check for the paths from where it's being picked up?

ehiggs: We can check the path. We can run nm on the file and check for required symbols. How long is a piece of string?

easybuild/easyblocks/h/hadoop.py
```diff
@@ -93,6 +98,20 @@ def sanity_check_step(self):
...
 
+        fake_mod_data = self.load_fake_module(purge=True)
+        cmd = "hadoop checknative -a"
+        out, _ = run_cmd(cmd, log_all=True, simple=False, log_ok=True)
+        self.clean_up_fake_module(fake_mod_data)
+
+        not_found = []
+        installdir = os.path.realpath(self.installdir)
+        for native_lib, lib_path  in self.cfg['extra_native_libs']:
+            lib_src = os.path.join([installdir, 'lib', 'native'])
```
boegel: this doesn't change, so hoist this out of the loop:

```python
native_lib_installdir = os.path.join([os.path.realpath(self.installdir), 'lib', 'native'])
for ...
```

easybuild/easyblocks/h/hadoop.py
```diff
@@ -93,6 +98,20 @@ def sanity_check_step(self):
...
         super(EB_Hadoop, self).sanity_check_step(custom_paths=custom_paths)
 
+        fake_mod_data = self.load_fake_module(purge=True)
+        cmd = "hadoop checknative -a"
+        out, _ = run_cmd(cmd, log_all=True, simple=False, log_ok=True)
+        self.clean_up_fake_module(fake_mod_data)
+
+        not_found = []
+        installdir = os.path.realpath(self.installdir)
+        for native_lib, lib_path  in self.cfg['extra_native_libs']:
```
boegel: `lib_path` is not used here, so use `_` instead

easybuild/easyblocks/h/hadoop.py
```diff
@@ -93,6 +98,20 @@ def sanity_check_step(self):
...
 
+        fake_mod_data = self.load_fake_module(purge=True)
+        cmd = "hadoop checknative -a"
+        out, _ = run_cmd(cmd, log_all=True, simple=False, log_ok=True)
+        self.clean_up_fake_module(fake_mod_data)
+
+        not_found = []
+        installdir = os.path.realpath(self.installdir)
+        lib_src = os.path.join([installdir, 'lib', 'native'])
+        for native_lib, lib_path  in self.cfg['extra_native_libs']:
```
boegel: `lib_path` is not used here, so use `_` instead

boegel:
Retested with https://github.com/hpcugent/easybuild-easyconfigs/pull/1773, looking good, so going in.

Thanks @ehiggs!

PR #640, opened 2015-07-09
Fix a bug wrt os.path.join

ehiggs:
I had a call to `os.path.join([...])` and it should have been `os.path.join(...)`. I don't know how it worked before, but it happened to work on one out of three of the clusters I tried to install this on.

### DIFF
--- a/easybuild/easyblocks/h/hadoop.py
+++ b/easybuild/easyblocks/h/hadoop.py
@@ -105,7 +105,7 @@ class EB_Hadoop(Tarball):
 
         not_found = []
         installdir = os.path.realpath(self.installdir)
-        lib_src = os.path.join([installdir, 'lib', 'native'])
+        lib_src = os.path.join(installdir, 'lib', 'native')
         for native_lib, _ in self.cfg['extra_native_libs']:
             if not re.search(r'%s: *true *%s' % (native_lib, lib_src), out):
                 not_found.append(native_lib)


boegel:
I'd love to figure out how this managed to get through undetected...

boegel:
I think I know why this passed undetected...

```
>>> os.path.join([installdir, 'lib', 'native'])
['/some/prefix', 'lib', 'native']
```

This is then used in a regular expression, which after string templates becomes something like `"snappy: *true ['/some/prefix', 'lib', 'native']" `.

This is a perfectly valid regular expression, which will most likely pass the check, since you're basically checking whether `'snappy: *true'` is there, followed by any character in the *character range* `['/some/prefix', 'lib', 'native']`...

So basically, the check was utterly broken...

How did you run into this issue exactly? The sanity check failed on you?

ehiggs:
On one cluster it worked. On another it failed with a bad character error. Perhaps related to different default Python installs that easybuild was using on the two clusters.

boegel:
Ah, yes, that makes sense, on our phanpy cluster we have a `-` in the installdir:

```
>>> installdir = '/vscmnt/gent_vulpix_apps/_apps_gent/CO7/haswell-ib/software/'
>>> re.compile(r'%s' % os.path.join([installdir, 'foo', 'bar']))
  File "/usr/lib64/python2.7/re.py", line 190, in compile
    return _compile(pattern, flags)
  File "/usr/lib64/python2.7/re.py", line 242, in _compile
    raise error, v # invalid expression
sre_constants.error: bad character range
```

Thanks for the info @ehiggs!

Merging this fix.

ehiggs:
>Ah, yes, that makes sense, on our phanpy cluster we have a - in the installdir:

One for the test suite!

boegel:
Well, this is a rather weird collapse of issues imho...

But, this may explain why some builds mysteriously fail on the systems that have a `-` in the install prefix... :)

PR #652, opened 2015-07-17
Update gcc.py

JensTimmerman:
a recent llvm breaks the comparisonprocess between stage 2 and 3 for some reason. This should fix it, see https://github.com/Homebrew/homebrew/issues/38501

### DIFF
--- a/easybuild/easyblocks/g/gcc.py
+++ b/easybuild/easyblocks/g/gcc.py
@@ -410,6 +410,10 @@ class EB_GCC(ConfigureMake):
 
             # enable bootstrapping for self-containment
             configopts += " --enable-bootstrap "
+            # Use 'bootstrap-debug' build configuration to force stripping of object
+            # files prior to comparison during bootstrap (broken by Xcode 6.3).
+            # see https://github.com/Homebrew/homebrew/issues/38501
+            configopts += "--with-build-config=bootstrap-debug",
 
             # PPL config options
             if self.cfg['withppl']:
--- a/easybuild/easyblocks/g/gcc.py
+++ b/easybuild/easyblocks/g/gcc.py
@@ -18,7 +18,7 @@
 #
 # EasyBuild is free software: you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/g/gcc.py
```diff
@@ -18,7 +18,7 @@
 #
 # EasyBuild is free software: you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
-# the Free Software Foundation v2.
+# the Free Software Foundation v2.bo
```
boegel: hmm?

easybuild/easyblocks/g/gcc.py
```diff
@@ -410,6 +410,10 @@ def build_step(self):
 
             # enable bootstrapping for self-containment
             configopts += " --enable-bootstrap "
+            # Use 'bootstrap-debug' build configuration to force stripping of object
+            # files prior to comparison during bootstrap (broken by Xcode 6.3).
+            # see https://github.com/Homebrew/homebrew/issues/38501
+            configopts += "--with-build-config=bootstrap-debug "
```
boegel: at best, we should only do this when the OS is detected as being OS X (via `get_os_type()` from `easybuild.tools.systemtools`)

it's not clear to me what other implications this extra configopts would have

boegel:
@JensTimmerman: care to mention why you're closing this?

easybuild/easyblocks/g/gcc.py
```diff
@@ -410,6 +410,10 @@ def build_step(self):
 
             # enable bootstrapping for self-containment
             configopts += " --enable-bootstrap "
+            # Use 'bootstrap-debug' build configuration to force stripping of object
+            # files prior to comparison during bootstrap (broken by Xcode 6.3).
+            # see https://github.com/Homebrew/homebrew/issues/38501
```
boegel: should also mention https://gcc.gnu.org/bugzilla/show_bug.cgi?id=65733

JensTimmerman:
didn't completely fix it after all.

PR #653, opened 2015-07-21
sort glob result to choose latest version

kthust:
Without the sort the installation took 4.1 instead of 4.4 on our system since the returned list of glob is not necessarily sorted by version.


boegel:
Jenkins: ok to test

boegel:
Looks fine, I'll test this together with #654 as soon as the small remark there is tackled.

boegel:
Tested with existing `tbb` easyconfigs and https://github.com/hpcugent/easybuild-easyconfigs/pull/1806, works like a charm, thanks @kthust!

PR #654, opened 2015-07-22
added env vars to module

kthust:
These environment variables are usually available after a standard installation.

### DIFF
--- a/easybuild/easyblocks/t/tbb.py
+++ b/easybuild/easyblocks/t/tbb.py
@@ -96,4 +96,7 @@ class EB_tbb(IntelBase):
         """Add correct path to lib to LD_LIBRARY_PATH. and intel license file"""
         txt = super(EB_tbb, self).make_module_extra()
         txt += self.module_generator.prepend_paths('LD_LIBRARY_PATH', [self.libpath])
+        txt += self.module_generator.prepend_paths('LIBRARY_PATH', [self.libpath])
+        txt += self.module_generator.prepend_paths('CPATH', [os.path.join('tbb', 'include')])
+        txt += self.module_generator.set_environment('TBBROOT', os.path.join(self.installdir, 'tbb'))
         return txt


boegel:
Jenkins: ok to test

easybuild/easyblocks/t/tbb.py
```diff
@@ -96,4 +96,7 @@ def make_module_extra(self):
         """Add correct path to lib to LD_LIBRARY_PATH. and intel license file"""
         txt = super(EB_tbb, self).make_module_extra()
         txt += self.module_generator.prepend_paths('LD_LIBRARY_PATH', [self.libpath])
+        txt += self.module_generator.prepend_paths('LIBRARY_PATH', [self.libpath])
+        txt += self.module_generator.prepend_paths('CPATH', [os.path.join('tbb', 'include')])
```
boegel: looks like `self.installdir` is missing here in the `os.path.join`?

kthust:
prepend_paths seems to be defined to add installdir to relative paths as long as prepend_paths evaluates true, which is its default value if not changed. This is also the behavior I see in my tests.

easybuild-framework/easybuild/tools/module_generator.py:240

says

# prepend $root (= installdir) for (non-empty) relative paths

Should I prefer doing this manually and prepend installdir and set allow_abs=True?

boegel:
@kthust: I'm sorry, you're right. I guess I got confused by seeing it in the lines around, but not on that particular line.

Not using `allow_abs` is better, so let's leave it as is. Sorry for the confusion.

I'll test this along with #653 and get them merged (hopefully for EB v2.2.1), thanks for the contribution!

kthust:
Thanks. If there is anything more to change just let me know.

boegel:
@kthust: should be fine, I just need to test this in combination with #653 and get it merged in

boegel:
Tested with existing `tbb` easyconfigs and https://github.com/hpcugent/easybuild-easyconfigs/pull/1806, works like a charm, thanks @kthust!

PR #658, opened 2015-08-07
Easyblock for PGI compilers

bartoldeman:
This is a working easyblock to install PGI compilers. It is my first contribution so it may not be the most elegant yet. Please let me know if I missed something.

it works by untarring followed by running a sed-patched install script (to use perhaps use GCC module compilers and libraries instead of always the system compilers)

license_file is set in a matching easyconfig to which is a symlink is produced (we use the same license file for multiple PGI's, and using $PGI/license.dat looks more specific than using $LM_LICENSE_FILE).

### DIFF
--- a/easybuild/easyblocks/p/pgi.py
+++ b/easybuild/easyblocks/p/pgi.py
@@ -1,0 +1,94 @@
+##
+# Copyright 2015 Bart Oldeman
+#
+# This file is part of EasyBuild,
+# originally created by the HPC team of Ghent University (http://ugent.be/hpc/en),
+# with support of Ghent University (http://ugent.be/hpc),
+# the Flemish Supercomputer Centre (VSC) (https://vscentrum.be/nl/en),
+# the Hercules foundation (http://www.herculesstichting.be/in_English)
+# and the Department of Economy, Science and Innovation (EWI) (http://www.ewi-vlaanderen.be/en).
+#
+# http://github.com/hpcugent/easybuild
+#
+# EasyBuild is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation v2.
+#
+# EasyBuild is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with EasyBuild.  If not, see <http://www.gnu.org/licenses/>.
+##
+"""
+EasyBuild support for installing PGI compilers, implemented as an easyblock
+
+@author: Bart Oldeman (McGill University, Calcul Quebec, Compute Canada)
+"""
+import os
+
+from easybuild.easyblocks.generic.tarball import Tarball
+from easybuild.tools.build_log import EasyBuildError
+from easybuild.tools.run import run_cmd
+
+class EB_pgi(Tarball):
+    """
+    Support for installing the PGI compilers
+    """
+
+    def make_module_req_guess(self):
+        """
+        A dictionary of possible directories to look for.
+        """
+        dirs = super(EB_pgi, self).make_module_req_guess()
+        prefix = os.path.join('linux86-64', self.version)
+        for key in dirs:
+            dirs[key] = [os.path.join(prefix, d) for d in dirs[key]]
+        return dirs
+
+    def sanity_check_step(self):
+        """Custom sanity check for PGI"""
+
+        prefix = os.path.join('linux86-64', self.version)
+        custom_paths = {
+                        'files': [os.path.join(prefix, "bin", "pgcc")],
+                        'dirs': [os.path.join(prefix, "bin"), os.path.join(prefix, "lib"),
+                                 os.path.join(prefix, "include"), os.path.join(prefix, "man")]
+                       }
+
+        super(EB_pgi, self).sanity_check_step(custom_paths=custom_paths)
+
+    def make_module_extra(self):
+        """Overwritten from Application to add extra txt"""
+        txt = super(EB_pgi, self).make_module_extra()
+        txt += self.module_generator.set_environment('PGI', self.installdir)
+        return txt
+
+    def install_step(self):
+        """Install by running install command."""
+
+        # make sure localrc uses GCC in PATH, not always the system GCC
+        cmd = "sed -i 's/^PATH/#PATH/' install"
+        run_cmd(cmd, log_all=True, simple=True)
+        cmd = "sed -i 's/^PATH/#PATH/' %s" % os.path.join("linux86-64", self.version,
+                                                          "bin", "makelocalrc")
+        run_cmd(cmd, log_all=True, simple=True)
+
+        cmd = 'PGI_SILENT=true PGI_ACCEPT_EULA=accept PGI_INSTALL_DIR=%s ' % self.installdir
+        cmd += 'PGI_INSTALL_NVIDIA=true PGI_INSTALL_AMD=true PGI_INSTALL_JAVA=true '
+        cmd += 'PGI_INSTALL_MANAGED=true '
+        cmd += './install'
+
+        run_cmd(cmd, log_all=True, simple=True)
+        if self.cfg['license_file']:
+            license_file = self.cfg['license_file']
+            try:
+                licfile = os.path.join(self.installdir, "license.dat")
+                os.symlink(license_file, licfile)
+            except OSError, err:
+                raise EasyBuildError("Failed to symlink %s to %s: %s", licfile, license_file, err)
+        else:
+            raise EasyBuildError("Please specify a license_file location in your easyconfig")
+
--- a/easybuild/easyblocks/p/pgi.py
+++ b/easybuild/easyblocks/p/pgi.py
@@ -1,5 +1,8 @@
 ##
-# Copyright 2015 Bart Oldeman
+# Copyright 2015-2015 Bart Oldeman
+#
+# This file is triple-licensed under GPLv2 (see below), MIT, and
+# BSD three-clause licenses.
 #
 # This file is part of EasyBuild,
 # originally created by the HPC team of Ghent University (http://ugent.be/hpc/en),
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/p/pgi.py
```diff
@@ -0,0 +1,94 @@
+##
+# Copyright 2015 Bart Oldeman
```
boegel: usually, copyright on contributions is transferred to Ghent University

if you'd like to retain copyright on this yourself, please read https://github.com/hpcugent/easybuild-easyblocks/blob/master/CONTRIBUTING.md and modify this bit accordingly

easybuild/easyblocks/p/pgi.py
```diff
@@ -0,0 +1,94 @@
...
+
+@author: Bart Oldeman (McGill University, Calcul Quebec, Compute Canada)
+"""
+import os
+
+from easybuild.easyblocks.generic.tarball import Tarball
+from easybuild.tools.build_log import EasyBuildError
+from easybuild.tools.run import run_cmd
+
+class EB_pgi(Tarball):
```
boegel: should be `EB_PGI`, and I don't see the point of deriving from `Tarball`, since you're not using any of its functionality (other than `configure_step` and `build_step` being skipped)

easybuild/easyblocks/p/pgi.py
```diff
@@ -0,0 +1,94 @@
...
+        prefix = os.path.join('linux86-64', self.version)
+        custom_paths = {
+                        'files': [os.path.join(prefix, "bin", "pgcc")],
+                        'dirs': [os.path.join(prefix, "bin"), os.path.join(prefix, "lib"),
+                                 os.path.join(prefix, "include"), os.path.join(prefix, "man")]
+                       }
+
+        super(EB_pgi, self).sanity_check_step(custom_paths=custom_paths)
+
+    def make_module_extra(self):
```
boegel: keep the `make_module` method definitions together

easybuild/easyblocks/p/pgi.py
```diff
@@ -0,0 +1,94 @@
...
+from easybuild.easyblocks.generic.tarball import Tarball
+from easybuild.tools.build_log import EasyBuildError
+from easybuild.tools.run import run_cmd
+
+class EB_pgi(Tarball):
+    """
+    Support for installing the PGI compilers
+    """
+
+    def make_module_req_guess(self):
```
boegel: move this down below `sanity_check_step`

easybuild/easyblocks/p/pgi.py
```diff
@@ -0,0 +1,94 @@
...
+
+        super(EB_pgi, self).sanity_check_step(custom_paths=custom_paths)
+
+    def make_module_extra(self):
+        """Overwritten from Application to add extra txt"""
+        txt = super(EB_pgi, self).make_module_extra()
+        txt += self.module_generator.set_environment('PGI', self.installdir)
+        return txt
+
+    def install_step(self):
```
boegel: move this up, above `sanity_check_step`

easybuild/easyblocks/p/pgi.py
```diff
@@ -0,0 +1,94 @@
...
+        """Install by running install command."""
+
+        # make sure localrc uses GCC in PATH, not always the system GCC
+        cmd = "sed -i 's/^PATH/#PATH/' install"
+        run_cmd(cmd, log_all=True, simple=True)
+        cmd = "sed -i 's/^PATH/#PATH/' %s" % os.path.join("linux86-64", self.version,
+                                                          "bin", "makelocalrc")
```
boegel: no reason to wrap this line?

easybuild/easyblocks/p/pgi.py
```diff
@@ -0,0 +1,94 @@
...
+        """Overwritten from Application to add extra txt"""
+        txt = super(EB_pgi, self).make_module_extra()
+        txt += self.module_generator.set_environment('PGI', self.installdir)
+        return txt
+
+    def install_step(self):
+        """Install by running install command."""
+
+        # make sure localrc uses GCC in PATH, not always the system GCC
+        cmd = "sed -i 's/^PATH/#PATH/' install"
```
boegel: calling out to `sed` is not the 'Pythonic' way

we usually do this via the `fileinput` module available in Python

that being said, we really need a `apply_regex_subs` function (probably in `easybuild.tools.filetools`) to make this easier...

easybuild/easyblocks/p/pgi.py
```diff
@@ -0,0 +1,94 @@
...
+        run_cmd(cmd, log_all=True, simple=True)
+        cmd = "sed -i 's/^PATH/#PATH/' %s" % os.path.join("linux86-64", self.version,
+                                                          "bin", "makelocalrc")
+        run_cmd(cmd, log_all=True, simple=True)
+
+        cmd = 'PGI_SILENT=true PGI_ACCEPT_EULA=accept PGI_INSTALL_DIR=%s ' % self.installdir
```
boegel: it would be easier to get an overview of the `PGI*` environment variables being set here if they were kept in a dict, like so (note: order them alphabetically):

```python
pgi_env_vars = {
    'PGI_ACCEPT_EULA': 'true',
    ...
    'PGI_SILENT': 'true',
}
cmd = "%s ./install" % ' '.join(['%s=%s' % x for x in sorted(pgi_env_vars.items()))
```

easybuild/easyblocks/p/pgi.py
```diff
@@ -0,0 +1,94 @@
...
+        cmd += 'PGI_INSTALL_NVIDIA=true PGI_INSTALL_AMD=true PGI_INSTALL_JAVA=true '
+        cmd += 'PGI_INSTALL_MANAGED=true '
+        cmd += './install'
+
+        run_cmd(cmd, log_all=True, simple=True)
+        if self.cfg['license_file']:
+            license_file = self.cfg['license_file']
+            try:
+                licfile = os.path.join(self.installdir, "license.dat")
+                os.symlink(license_file, licfile)
```
boegel: I would opt for copying the file rather than symlinking?

easybuild/easyblocks/p/pgi.py
```diff
@@ -0,0 +1,94 @@
...
+        custom_paths = {
+                        'files': [os.path.join(prefix, "bin", "pgcc")],
+                        'dirs': [os.path.join(prefix, "bin"), os.path.join(prefix, "lib"),
+                                 os.path.join(prefix, "include"), os.path.join(prefix, "man")]
+                       }
+
+        super(EB_pgi, self).sanity_check_step(custom_paths=custom_paths)
+
+    def make_module_extra(self):
+        """Overwritten from Application to add extra txt"""
```
boegel: fix docstring, mention that `$PGI` is being defined

easybuild/easyblocks/p/pgi.py
```diff
@@ -0,0 +1,94 @@
...
+from easybuild.tools.run import run_cmd
+
+class EB_pgi(Tarball):
+    """
+    Support for installing the PGI compilers
+    """
+
+    def make_module_req_guess(self):
+        """
+        A dictionary of possible directories to look for.
```
boegel: clarify docstring, explain what's going on, something like:

```python
"""Prefix subdirectories in PGI install directory considered for environment variables defined in module file."""
```

easybuild/easyblocks/p/pgi.py
boegel: oh, and use `2015-2015` (to allow easy updating in 2016)

easybuild/easyblocks/p/pgi.py
```diff
@@ -0,0 +1,94 @@
...
+        dirs = super(EB_pgi, self).make_module_req_guess()
+        prefix = os.path.join('linux86-64', self.version)
+        for key in dirs:
+            dirs[key] = [os.path.join(prefix, d) for d in dirs[key]]
+        return dirs
+
+    def sanity_check_step(self):
+        """Custom sanity check for PGI"""
+
+        prefix = os.path.join('linux86-64', self.version)
```
boegel: this is used in a couple of places, so define this once (in the easyblock constructor) and reuse that value?

```python
    """Easyblock constructor, define custom class variables specific to PGI."""
    super(EB_PGI, self).__init__(*args, **kwargs)
    self.install_subdir = os.path.join('linux86-64', self.version)
```

easybuild/easyblocks/p/pgi.py
```diff
@@ -0,0 +1,94 @@
...
+                                                          "bin", "makelocalrc")
+        run_cmd(cmd, log_all=True, simple=True)
+
+        cmd = 'PGI_SILENT=true PGI_ACCEPT_EULA=accept PGI_INSTALL_DIR=%s ' % self.installdir
+        cmd += 'PGI_INSTALL_NVIDIA=true PGI_INSTALL_AMD=true PGI_INSTALL_JAVA=true '
+        cmd += 'PGI_INSTALL_MANAGED=true '
+        cmd += './install'
+
+        run_cmd(cmd, log_all=True, simple=True)
+        if self.cfg['license_file']:
```
boegel: with the Intel compilers, we give preference to (`$INTEL_LICENSE_FILE` or) `$LM_LICENSE_FILE` over `license_file` specified in the easyconfig file, since that is usually defined already for runtime usage of the compilers

I don't know if something similar makes any sense here; does PGI listen to a particular environment variable at runtime to find the license file, or does it only consider `license.dat` in the install dir?

boegel:
Jenkins: ok to test

boegel:
@bartoldeman: thanks a lot for this, I know several people have interest in seeing PGI support added to EasyBuild, and being able to install PGI with EasyBuild is the first step to make this happen

The easyblock is pretty OK overall, just a couple of suggestions (see other comments). Let me know if you have any questions or concerns related to them.

We currently don't have access to PGI, which is an issue w.r.t. to testing this PR (and future ones that require PGI, e.g. easyconfigs using a PGI toolchain).

We've talked to PGI in the past about getting our hands on some kind of free license with limited capabilities, and it sounded like that was a viable option. I need to look into that again...

easybuild/easyblocks/p/pgi.py
bartoldeman: is it ok to use the same header as used in easyblocks/c/clang.py? (as far as I can see in CONTRIBUTING.md this is ok)

# This file is triple-licensed under GPLv2 (see below), MIT, and
# BSD three-clause licenses.

boegel: Yes, that works too. If it's licensed under MIT/BSD, we have the freedom of either relicensing it at GPLv2, or to relicense everything 'down' to BSD if we ever wanted to (cfr. https://github.com/hpcugent/easybuild-framework/issues/335), which is the whole point of these requirements.

bartoldeman:
Thanks for all the comments. I have incorporated them except the change from EB_pgi to EB_PGI.

The compiler itself ships with a sample modulefile named "pgi/15.7", so people are used to loading modulefiles for this compiler with the non-capitalized pgi. A pgi easyconfig then looks for EB_pgi, much like icc looks for EB_icc; EB_PGI is not found (unless I overlooked something?).

easybuild/easyblocks/p/pgi.py
```diff
@@ -0,0 +1,114 @@
...
+import os
+import fileinput
+import re
+import sys
+
+from easybuild.framework.easyblock import EasyBlock
+from easybuild.tools.build_log import EasyBuildError
+from easybuild.tools.run import run_cmd
+
+class EB_pgi(EasyBlock):
```
boegel: should be `EB_PGI`?

easybuild/easyblocks/p/pgi.py
```diff
@@ -0,0 +1,114 @@
...
+    def install_step(self):
+        """Install by running install command."""
+
+        pgi_env_vars = {
+            'PGI_ACCEPT_EULA': 'accept',
+            'PGI_INSTALL_AMD': 'true',
+            'PGI_INSTALL_DIR': self.installdir,
+            'PGI_INSTALL_JAVA': 'true',
+            'PGI_INSTALL_MANAGED': 'true',
+            'PGI_INSTALL_NVIDIA': 'true',
```
geimer: Shouldn't these be better custom easyconfig parameters to allow sites to tweak their setup easily?

boegel: +1

bartoldeman: Done

boegel:
Retested with https://github.com/hpcugent/easybuild-easyconfigs/pull/1833, I see no reason to hold this back any further.

Thanks *a lot* @bartoldeman for tackling this!

PR #659, opened 2015-08-07
correct import of fancylogger

JensTimmerman:
importing directly from vsc will break in the future


boegel:
Thanks @JensTimmerman!

PR #661, opened 2015-08-12
PETSc library with Complex support

jordiblasco:


boegel:
Jenkins: ok to test

easybuild/easyblocks/p/petsc.py
```diff
@@ -312,10 +317,18 @@ def sanity_check_step(self):
         else:
             libext = "a"
 
-        custom_paths = {
-                        'files': [os.path.join(prefix2, "lib", "libpetsc.%s" % libext)],
-                        'dirs': [os.path.join(prefix1, "bin"), os.path.join(prefix2, "conf"),
+        if LooseVersion(self.version) < LooseVersion("3.6"):
+            custom_paths = {
+                            'files': [os.path.join(prefix2, "lib", "libpetsc.%s" % libext)],
+                            'dirs': [os.path.join(prefix1, "bin"), os.path.join(prefix2, "conf"),
```
boegel: fix indent (4 spaces suffices)

easybuild/easyblocks/p/petsc.py
```diff
@@ -312,10 +317,18 @@ def sanity_check_step(self):
...
+                                 os.path.join(prefix1, "include"), os.path.join(prefix2, "include")]
+                           }
+        else:
+            custom_paths = {
+                            'files': [os.path.join(prefix2, "lib", "libpetsc.%s" % libext)],
+                            'dirs': [os.path.join(prefix1, "bin"),
```
boegel: fix indent

easybuild/easyblocks/p/petsc.py
```diff
@@ -122,6 +123,10 @@ def configure_step(self):
                     raise EasyBuildError("PAPI header (%s) and/or lib (%s) not found, can not enable PAPI support?",
                                          papi_inc_file, papi_lib)
 
+            # Complex number suppport
+            if self.cfg['with_complex']:
+                    self.cfg.update('configopts', '--with-scalar-type=complex')
```
boegel: fix indent

easybuild/easyblocks/p/petsc.py
boegel: hmm, the only thing different here is that the `conf` dir is no longer there in recent versions?

I'm willing to just simply drop that from the sanity check then, since if this is not there, it's not something EB can do much about anyway? As long as the binaries/libraries/header files are there, things should be fine?

boegel:
@jordiblasco: is there any way to check that the resulting build is indeed supporting complex?

easybuild/easyblocks/p/petsc.py
jordiblasco: Yes, it seems that ```conf``` dir is not longer there. From my point of view this folder can be removed from the sanity check. I added these lines of code for legacy reasons.

boegel:
@jordiblasco: please merge https://github.com/jordiblasco/easybuild-easyblocks/pull/2 to fix the remarks

boegel:
@jordiblasco: this is still waiting for https://github.com/jordiblasco/easybuild-easyblocks/pull/2 to be included...

boegel:
Bumping the milestone (again) on this, since https://github.com/hpcugent/easybuild-easyconfigs/pull/1848 isn't ready either yet.

ocaisa:
This is really old but nevertheless I am looking at it now. There are two knobs that you can tweak with PetSC: `--with-scalar-type` and `--with-precision` and each fundamentally change the libraries you get out the other end (using complex when you only need real has a performance impact, using single precision when you need double for a solver doesn't work).

I don't think it's a good idea to handle this in the easyblock as each variant is a separate installation. This means they all should be versionsuffix-ed, and that you can only really control that in an easyconfig. You could sanity check the given values in the easyblock.

boegel:
This is over 10 years old by now, time to put an end to it...

PR #665, opened 2015-08-14
add easyblock for Molpro (REVIEW)

boegel:


boegel:
Still testing this one, but early feedback can't hurt. :-)

boegel:
@wpoely86: please review

easybuild/easyblocks/m/molpro.py
```diff
@@ -0,0 +1,196 @@
...
+                    raise EasyBuildError("Failed to create symlink for license token at %s", self.license_token)
+
+            else:
+                # no license token available, so fail early
+                raise EasyBuildError("No license token found at either %s or via 'license_file'", self.license_token)
+
+    def configure_step(self):
+        """Custom configuration procedure for Molpro: use 'configure -batch'."""
+
+        if not self.cfg['license_file'] or not os.path.exists(self.cfg['license_file']):
```
wpoely86: Isn't this already sufficiently done in the `__init__`?

boegel: Yeah, also, I had to move this to the `configure_step`, for compatibility with `--module-only`.

easybuild/easyblocks/m/molpro.py
```diff
@@ -0,0 +1,194 @@
...
+        for line in fileinput.input(cfgfile, inplace=1, backup='.orig'):
+            line = re.sub(r"^(LAUNCHER\s*=\s*).*$", r"\1 %s" % launcher, line)
+            sys.stdout.write(line)
+
+        # reread CONFIG and log contents
+        cfgtxt = read_file(cfgfile)
+        self.log.info("Contents of CONFIG file:\n%s", cfgtxt)
+
+        # make sure license token is available
+        token_path = os.path.join(os.environ['HOME'], '.molpro', 'token')
```
wpoely86: why not do this at the beginning?

wpoely86:
looks fine

easybuild/easyblocks/m/molpro.py
boegel: ah, yes, this check can be removed, since we made sure it has been put in place at the start of `configure_step`

boegel:
Tested with easyconfig in https://github.com/hpcugent/easybuild-easyconfigs/pull/1880, works as intended, so going in.

PR #668, opened 2015-08-21
call PythonPackage.configure_step in ConfigureMakePythonPackage.configure_step

boegel:
This is required to make sure that `self.all_pylibdirs` gets defined, which is used for defining `$PYTHONPATH` in the module file in `make_module_extra`.


wpoely86:
looks ok

boegel:
Thx @wpoely86!

PR #669, opened 2015-08-23
add custom sanity check in scipy easyblock

boegel:
required because of enabling fallback to default bin/lib sanity check paths due to hpcugent/easybuild-framework#1366


pescobar:
boegel it would be nice to add this in the sanity checking

```
python -c 'import scipy; scipy.test("full");
```

the same works for numpy...

boegel:
@pescobar: the tests are already being run for scipy, this is triggered by this line in the scipy easyblock:

```
self.testcmd = "cd .. && python -c 'import numpy; import scipy; scipy.test(verbose=2)'" 
```

It seems like the tests are not being run yet for numpy, however...

So, this definitely needs to be fixed in a separate PR, since this PR fixes the broken sanity check for scipy...

PR #670, opened 2015-08-23
specify to always use the bfd linker for OpenFOAM, to stay away from using ld.gold

boegel:
linking `ld.gold` with doesn't work with OpenFOAM, yields errors like:

```
.../bin/ld.gold: error: --add-needed is not supported but is required for librdmacm.so.1 in .../OpenMPI/1.8.5-GNU-4.9.2-2.25/lib/libmpi_cxx.so
```

### DIFF
--- a/easybuild/easyblocks/o/openfoam.py
+++ b/easybuild/easyblocks/o/openfoam.py
@@ -111,7 +111,7 @@ class EB_OpenFOAM(EasyBlock):
 
         # make sure non-gold version of ld is used, since OpenFOAM requires it
         # see http://www.openfoam.org/mantisbt/view.php?id=685
-        extra_flags = '-fuse-ld=bfd'
+        extra_flags = '-Wl,-fuse-ld=bfd'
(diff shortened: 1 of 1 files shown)


boegel:
cc @ocaisa: although we specified that `ld` (i.e. `ld.bfd`) should be the default linker in `binutils`, we end up using `ld.gold` anyway because GCC defaults to using `ld.gold` is it sees it's available...

This breaks the OpenFOAM build, it's simply not compatible with how `ld.gold` works (somehow that doesn't surprise me ;-)). cfr. http://www.openfoam.org/mantisbt/view.php?id=685

boegel:
To be more precise: *we* configure GCC with `--enable-gold=default` in the GCC easyblock, so that's why it's picking up `ld.gold` by default...

boegel:
@wpoely86: please review?

wpoely86:
looks fine

boegel:
Retested all OpenFOAM builds on top of this, all is well.

PR #671, opened 2015-08-26
add 'foldx3b6' as possible FoldX binary name

boegel:


easybuild/easyblocks/f/foldx.py
```diff
@@ -49,6 +49,7 @@ def install_step(self):
             'foldx_%s.linux' % self.version,  # FoldX v2.x
             'FoldX.linux64',  # FoldX 3.x
             'foldx64Linux',  # FoldX 3.0-beta6
+            'foldx3b6',  # FoldX 3.0 beta 6.1 >_<
```
boegel: deriving this from the version (`3.0-beta6.1`) is probably not worth the trouble...

wpoely86:
nice one

boegel:
@wpoely86: https://github.com/hpcugent/easybuild-easyconfigs/pull/1916 makes it even better...

boegel:
Thanks @wpoely86!

PR #672, opened 2015-08-26
enhance/cleanup MATLAB easyblock

boegel:
mostly cleanup, implementation was made a little bit less strict w.r.t. the contents of the 'source' tarball by using `self.cfg['start_dir']` where applicable


wpoely86:
looks good

boegel:
Works too. ;-)

PR #673, opened 2015-08-27
move preparing of 'intel' subdir in $HOME to configure_step

boegel:
Fiddling with `$HOME/intel` (which is required to make multiple installations of Intel tools in parallel work) is only needed when actually doing the installation, not when initialising the easyblock (e.g. in the unit tests), or when using `--module-only`.


wpoely86:
looks great

boegel:

PR #674, opened 2015-08-28
Netcdf easyblock updated for 4.3.3.1

hajgato:


easybuild/easyblocks/n/netcdf.py
```diff
@@ -71,6 +71,27 @@ def configure_step(self):
 
             CMakeMake.configure_step(self)
 
+        if LooseVersion(self.version) >= LooseVersion("4.3.3"):
```
boegel: is this version check strictly needed? this probably applies to all netCDF versions where CMake is used to configure (i.e. 4.3.x and up)?

easybuild/easyblocks/n/netcdf.py
```diff
@@ -71,6 +71,27 @@ def configure_step(self):
...
+            set.cfg.update('configopts', '-DCMAKE_BUILD_TYPE=RELEASE -DCMAKE_C_FLAGS_RELEASE="-DNDEBUG " ')
+            hdf5 = get_software_root('HDF5')
+            hdf5_libdir = get_software_libdir('HDF5')
+            szip_libdir = get_software_libdir('Szip')
+            if hdf5_libdir and szip_libdir:
```
boegel: why the `and`, rather than having `if hdf5_libdir` and `if szip_libdir` separately?

easybuild/easyblocks/n/netcdf.py
```diff
@@ -71,6 +71,27 @@ def configure_step(self):
...
+                set.cfg.update('configopts', '-DHDF5_LIB=%s/libhdf5.so ' % hdf5_libdir)
+                set.cfg.update('configopts', '-DHDF5_HL_LIB=%s/libhdf5_hl.so ' % hdf5_libdir)
+                set.cfg.update('configopts', '-DHDF5_INCLUDE_DIR=%s/include ' % hdf5)
+                set.cfg.update('configopts', '-DSZIP_LIBRARY=%s/libsz.so ' % szip_libdir)
+
+            zlib_libdir = get_software_libdir('zlib')
+            if zlib_libdir:
+                 set.cfg.update('configopts', '-DZLIB_LIBRARY=%s/libz.so ' % zlib_libdir)
```
boegel: why doesn't `ZLIB_INCLUDE_DIR` not need to be set?

easybuild/easyblocks/n/netcdf.py
```diff
@@ -71,6 +71,27 @@ def configure_step(self):
...
+        if LooseVersion(self.version) >= LooseVersion("4.3.3"):
+            set.cfg.update('configopts', '-DCMAKE_BUILD_TYPE=RELEASE -DCMAKE_C_FLAGS_RELEASE="-DNDEBUG " ')
+            hdf5 = get_software_root('HDF5')
+            hdf5_libdir = get_software_libdir('HDF5')
+            szip_libdir = get_software_libdir('Szip')
+            if hdf5_libdir and szip_libdir:
+                set.cfg.update('configopts', '-DHDF5_LIB=%s/libhdf5.so ' % hdf5_libdir)
+                set.cfg.update('configopts', '-DHDF5_HL_LIB=%s/libhdf5_hl.so ' % hdf5_libdir)
+                set.cfg.update('configopts', '-DHDF5_INCLUDE_DIR=%s/include ' % hdf5)
+                set.cfg.update('configopts', '-DSZIP_LIBRARY=%s/libsz.so ' % szip_libdir)
```
boegel: `SZIP_INCLUDE_DIR`?

easybuild/easyblocks/n/netcdf.py
```diff
@@ -71,6 +71,27 @@ def configure_step(self):
...
+
+            zlib_libdir = get_software_libdir('zlib')
+            if zlib_libdir:
+                 set.cfg.update('configopts', '-DZLIB_LIBRARY=%s/libz.so ' % zlib_libdir)
+
+            curl = get_software_root('cURL')
+            curl_libdir = get_software_libdir('cURL')
+            if curl_lindir:
```
boegel: typo :)

easybuild/easyblocks/n/netcdf.py
hajgato: I think szip is only useful if HDF5 is used

easybuild/easyblocks/n/netcdf.py
hajgato: No idea.I can set, but it might not change anything.

easybuild/easyblocks/n/netcdf.py
hajgato: See ZLIB. I think szip and zlib just deps of other libs (HDF5?) so include dirs are not needed,

easybuild/easyblocks/n/netcdf.py
```diff
@@ -71,6 +71,27 @@ def configure_step(self):
...
+                set.cfg.update('configopts', '-DCURL_LIBRARY=%s/libcurl.so ' % curl_libdir)
+                set.cfg.update('configopts', '-DCURL_INCLUDE_DIR=%s/include ' % curl)
```
boegel: you can probably tighten this up quite a bit:

```python
for (dep, libname) in [('cURL', 'curl'), ('HDF5', 'hdf5'), ('Szip', 'sz'), ('zlib', 'z')]:
    dep_root = curl = get_software_root(dep)
    dep_libdir = get_software_libdir(dep)
    if dep_root:
        set.cfg.update('configopts', '-D%s_INCLUDE_DIR=%s/include ' % (dep.upper(), dep_root))
        if dep == 'HDF5':
            set.cfg.update('configopts', '-DHDF5_LIB=%s/libhdf5.so ' % dep_libdir)
            set.cfg.update('configopts', '-DHDF5_HL_LIB=%s/libhdf5_hl.so ' % dep_libdir)
        else:
            set.cfg.update('configopts', '-D%s_LIBRARY=%s/lib%s.so ' % (dep.upper(), dep_libdir, libname))

easybuild/easyblocks/n/netcdf.py
hajgato: Good question. You are probably right. The behaviour might also affected by the CMake version. (because netcdf uses some CMake provided modules to find HDF5)

boegel:
@hajgato: please include https://github.com/hajgato/easybuild-easyblocks/pull/11 to fix style remarks

boegel:
retested with existing netCDF 4.3.* easyconfigs, and with https://github.com/hpcugent/easybuild-easyconfigs/pull/1912, good to go, thanks @hajgato!

PR #679, opened 2015-09-03
make unpacked source dir detection in easyblock for VSC-tools a little bit more flexible

boegel:
Sometimes, the name of unpacked source tarballs have a prefix, e.g. when downloaded directly from GitHub: `vsc-mympirun-vsc-mympirun-3.4.2`


boegel:
no brainer imho

PR #680, opened 2015-09-03
Various changes to make molpro work with precompiled binaries and wit…

valtandor:


valtandor:
Yes

boegel:
PR should target `develop`, not `master`

in general, I'm not a big fan of this... why not simply use the `Binary` easyblock in the `Molpro` easyconfig when installing a prebuilt version?

You would have to add a couple of things extra in the easyconfig, sure, but it would be rather minimal?

Besides, why go with a binary install if EB supports building from source?

valtandor:
@boegel: As I noted on the easyconfig PR, not all users receive a copy of the source code.

Does the binary easyblock allow for a QA format?

boegel:
@valtandor: ok, after taking a closer look, I understand that using `Binary` isn't going to work out, indeed

easybuild/easyblocks/m/molpro.py
```diff
@@ -32,15 +32,31 @@
...
 
+    @staticmethod
+    def extra_options():
+        """Define custom easyconfig parameters for Molpro."""
+        # Combine extra variables from Binary and ConfigureMake easyblocks as
+        # well as those needed for Molpro specifically
+        extra_vars = Binary.extra_options()
+        extra_vars = ConfigureMake.extra_options(extra_vars)
+        extra_vars.update({
+            'license_token_present': [True, "Can we supply a license token at install time?", CUSTOM],
```
boegel: is this custom parameter necessary? can't this be detected based on whether `self.license_token` is available (which is  required currently)

if you need support to install without a license token being available, just relax the requirements w.r.t. `self.license_token` instead (e.g. just issue a warning if it's not available), I would stay away from adding a custom parameter like this

easybuild/easyblocks/m/molpro.py
```diff
@@ -32,15 +32,31 @@
...
+from easybuild.framework.easyblock import EasyBlock
+from easybuild.framework.easyconfig import CUSTOM
 from easybuild.tools.build_log import EasyBuildError
 from easybuild.tools.filetools import mkdir, read_file
-from easybuild.tools.run import run_cmd
+from easybuild.tools.run import run_cmd, run_cmd_qa
 
 
-class EB_Molpro(ConfigureMake):
+class EB_Molpro(Binary, ConfigureMake):
```
boegel: swap this around; the first class being derived from determines what `super` calls refer to

```python
class EB_Molpro(ConfigureMake, Binary):
```

easybuild/easyblocks/m/molpro.py
```diff
@@ -139,13 +165,21 @@ def configure_step(self):
         cfgtxt = read_file(cfgfile)
         self.log.info("Contents of CONFIG file:\n%s", cfgtxt)
 
+    def build_step(self):
+        if not self.cfg['precompiled_binaries']:
+            ConfigureMake.build_step(self)
```
boegel: this is not needed if you list `ConfigureMake` in the `class` definition

easybuild/easyblocks/m/molpro.py
```diff
@@ -67,8 +89,12 @@ def configure_step(self):
...
 
             else:
-                # no license token available, so fail early
-                raise EasyBuildError("No license token found at either %s or via 'license_file'", self.license_token)
+                self.cfg['license_token_present'] = False
+                self.log.warning("No licence token found at either {0} or via 'license_file'".format(self.license_token))
+        
+        # Precompiled binaries
+        if self.cfg['precompiled_binaries']:
+            return
```
boegel: I'd rather see you indent code than `return` in the middle of a method, it makes the code harder to grasp

easybuild/easyblocks/m/molpro.py
```diff
@@ -154,15 +188,45 @@ def install_step(self):
...
-        # since the value used during installation point to temporary files
-        for line in fileinput.input(os.path.join(self.full_prefix, 'bin', 'molpro'), inplace=1):
-            line = re.sub(r"^(LAUNCHER\s*=\s*).*$", r"\1 %s" % self.orig_launcher, line)
-            sys.stdout.write(line)
+        else:
+
+            if self.cfg['license_token_present']:
+                run_cmd("make tuning")
+
+            ConfigureMake.install_step(self)
```
boegel: use `super` like it was before please

valtandor:
I will have a look at these suggested changes, and probably implement most of them tomorrow.

boegel:
:+1:

easybuild/easyblocks/m/molpro.py
valtandor: Done.

easybuild/easyblocks/m/molpro.py
valtandor: Done

easybuild/easyblocks/m/molpro.py
valtandor: Even though the method uses the value of the precompiled_binaries parameter?

easybuild/easyblocks/m/molpro.py
valtandor: Done

valtandor:
Closing and creating a new PR into develop.

PR #681, opened 2015-09-04
Plumed support for CP2K

wpoely86:
Requires testing


easybuild/easyblocks/c/cp2k.py
```diff
@@ -191,6 +192,13 @@ def configure_step(self):
         if os.getenv('LIBSCALAPACK', None) is not None:
             options = self.configure_ScaLAPACK(options)
 
+        # PLUMED
+        if self.cfg["plumed"]:
+            if not get_software_root('PLUMED'):
+                raise EasyBuildError("The PLUMED module needs to be loaded to use PLUMED")
```
boegel: ... to build CP2K with PLUMED support?

boegel:
Tested with https://github.com/hpcugent/easybuild-easyconfigs/pull/1948, working great, going in. Thanks @wpoely86!

PR #682, opened 2015-09-07
Updated go easyblock for version 1.5. This version need the variable …

cfenoy:
…GOROOT_BOOTSTRAP to point to a current Go installation

### DIFF
--- a/easybuild/easyblocks/g/go.py
+++ b/easybuild/easyblocks/g/go.py
@@ -25,6 +25,7 @@
 """
 EasyBuild support for building and installing Go, implemented as an easyblock
 
+@author: Carlos Fenoy (Roche)
 @author: Adam DeConinck (NVIDIA)
 @author: Kenneth Hoste (HPC-UGent)
 """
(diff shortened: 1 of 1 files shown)


boegel:
@cfenoy: please target the `develop` branch for making changes; `master` is the 'stable' version, and is only updated when a new EB release is made (by merging `develop` into `master` after running regression tests)

You'll need to issue a new PR since you can't change the target branch for an open PR.

boegel:
@cfenoy: see also https://github.com/hpcugent/easybuild/wiki/Contributing-back and https://github.com/hpcugent/easybuild/wiki/Policy-for-easyconfig-pull-requests (latter are just guidelines, we're not strictly enforcing those things)

cfenoy:
@boegel Sorry for that, I'll create a new PR

boegel:
@cfenoy: no problem, it's a common 'mistake' for people issuing PRs to EB repos for the first time

we want to keep `master` as the default branch since that's what people should be playing with when they clone the repo, but that implies it's also the default target for PRs...

PR #683, opened 2015-09-07
Updated go easyblock for version 1.5

cfenoy:
Go version 1.5 needs the variable GOROOT_BOOTSTRAP to point to a current Go installation

### DIFF
--- a/easybuild/easyblocks/g/go.py
+++ b/easybuild/easyblocks/g/go.py
@@ -61,7 +61,7 @@ class EB_Go(ConfigureMake):
 
         # $GOROOT_FINAL only specifies the location of the final installation, which gets baked into the binaries
         # the installation itself is *not* done by the all.bash script, that needs to be done manually
-        cmd = "GOROOT_FINAL=%s ./all.bash" % self.installdir
+        cmd = "GOROOT_BOOTSTRAP=%s GOROOT_FINAL=%s ./all.bash" % (os.environ.get('EBROOTGO'),self.installdir)
         run_cmd(cmd, log_all=True, simple=False)
 
         try:
--- a/easybuild/easyblocks/g/go.py
+++ b/easybuild/easyblocks/g/go.py
@@ -31,15 +31,20 @@ EasyBuild support for building and installing Go, implemented as an easyblock
 import os
 import shutil
 
+from distutils.version import LooseVersion
+
 from easybuild.easyblocks.generic.configuremake import ConfigureMake
+from easybuild.framework.easyconfig import CUSTOM
 from easybuild.tools.build_log import EasyBuildError
 from easybuild.tools.filetools import rmtree2
 from easybuild.tools.run import run_cmd
+from easybuild.tools.modules import get_software_root
 
 class EB_Go(ConfigureMake):
     """
     Build Go compiler
     """
+
     def configure_step(self):
         """No dedicated configure step."""
         pass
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/g/go.py
```diff
@@ -61,7 +61,7 @@ def install_step(self):
 
         # $GOROOT_FINAL only specifies the location of the final installation, which gets baked into the binaries
         # the installation itself is *not* done by the all.bash script, that needs to be done manually
-        cmd = "GOROOT_FINAL=%s ./all.bash" % self.installdir
+        cmd = "GOROOT_BOOTSTRAP=%s GOROOT_FINAL=%s ./all.bash" % (os.environ.get('EBROOTGO'),self.installdir)
```
boegel: looks good, but you should use the `get_software_root` function rather than hardcoding `$EBROOTGO`

or, actually, how can `EBROOTGO` be defined in the environment when you're installing Go?!

maybe you mean this?

```python
cmd = "GOROOT_BOOTSTRAP=%(installdir)s GOROOT_FINAL=%(installdir)s ./all.bash" % {'installdir': self.installdir}
```

boegel:
Jenkins: ok to test

easybuild/easyblocks/g/go.py
boegel: maybe it's also good to add a comment why `$GOROOT_BOOTSTRAP` must be defined now too?

should we make this specific to Go 1.5+?

```python
if LooseVersion(self.version) >= LooseVersion('1.5'):
   ...
```

cfenoy: The 1.5 version of Go needs go to be compiled. Hence it needs the 1.4.2
module loaded to be built. (I've just realised that I would need to change
the Easyconfigs to avoid having both versions loaded due to the dependency)
The 1.4.2 only needs gcc to be compiled, so the EBROOTGO should be empty

On Mon, Sep 7, 2015 at 6:37 PM, Kenneth Hoste <notifications@github.com>
wrote:

> In easybuild/easyblocks/g/go.py
> <https://github.com/hpcugent/easybuild-easyblocks/pull/683#discussion_r38873196>
> :
>
> > @@ -61,7 +61,7 @@ def install_step(self):
> >
> >          # $GOROOT_FINAL only specifies the location of the final installation, which gets baked into the binaries
> >          # the installation itself is *not* done by the all.bash script, that needs to be done manually
> > -        cmd = "GOROOT_FINAL=%s ./all.bash" % self.installdir
> > +        cmd = "GOROOT_BOOTSTRAP=%s GOROOT_FINAL=%s ./all.bash" % (os.environ.get('EBROOTGO'),self.installdir)
>
> looks good, but you should use the get_software_root function rather than
> hardcoding $EBROOTGO
>
> or, actually, how can EBROOTGO be defined in the environment when you're
> installing Go?!
>
> maybe you mean this?
>
> cmd = "GOROOT_BOOTSTRAP=%(installdir)s GOROOT_FINAL=%(installdir)s ./all.bash" % {'installdir': self.installdir}
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/hpcugent/easybuild-easyblocks/pull/683/files#r38873196>
> .
>



-- 
--
Carles Fenoy

cfenoy:
@boegel Does it make sense to add a new option to specify the goroot, and in the easyblock check if the EBROOTGO or the goroot are set and exit otherwise?

boegel:
@cfenoy: I think you should only require having Go loaded for Go 1.5 and up. And you shouldn't check for `$EBROOTGO` directly yourself, but use `get_software_root('Go')` instead; right now, this checks `$EBROOTGO` for you, but we have been thinking about changing the underlying implementation to have a more direct check rather than replying on `$EBROOTGO` (which could get defined in ways other than the Go module being loaded)

if `Go` is not loaded as a (build) dep, `get_software_root` will return `None`

something like (untested, requires extra `import`s):
```python
if LooseVersion(self.version) >= LooseVersion('1.5'):
    go_root = get_software_root('Go')
    if go_root:
        cmd = "GOROOT_BOOTSTRAP=%s GOROOT_FINAL=%s ./all.bash" % (go_root, self.installdir)
    else:
        raise EasyBuildError("Go is required as a build dependency for installing Go since version 1.5")
else:
    cmd = "GOROOT_FINAL=%s ./all.bash" % self.installdir
```

boegel:
@cfenoy: please merge https://github.com/cfenoy/easybuild-easyblocks/pull/1 to fix the style remark

cfenoy:
Merged

boegel:
Tested with https://github.com/hpcugent/easybuild-easyconfigs/pull/1951, good to go, thanks @cfenoy, and congratulations on your first contribution to EasyBuild!

One of many, I hope. :)

PR #657, opened 2015-07-31
minor change for MUMPS 5 release

jordiblasco:

PR #686, opened 2015-09-14
update MUMPS easyblock for 5.x

boegel:
wrapper PR for #657 by @jordiblasco


boegel:
(re)tested with https://github.com/hpcugent/easybuild-easyconfigs/pull/1867, working as it should

reviewed version of #657, so good to go

thanks @jordiblasco!

PR #687, opened 2015-09-15
add MCRROOT environmental variable

hajgato:


easybuild/easyblocks/m/mcr.py
```diff
@@ -136,5 +136,6 @@ def make_module_extra(self):
             txt += self.module_generator.prepend_paths('LD_LIBRARY_PATH', libdir)
 
         txt += self.module_generator.set_environment('_JAVA_OPTIONS', self.cfg['java_options'])
+        txt += self.module_generator.set_environment('MCRROOT', self.subdir)
```
boegel: @hajgato: should use `os.path.join(self.installdir, self.subdir)` instead of `self.subdir`, since the latter is a relative path?

`prepend_paths` prefixes with the install prefix, `set_environment` does not since it may be something totally different than a filepath

hajgato:
@boegel done!

boegel:
Tested with existing MCR easyconfigs, works like a charm, thanks @hajgato!

PR #688, opened 2015-09-16
implement runtime patching of $WM_* and compiler variables for OpenFOAM

boegel:
This implements the fix that was proposed by @wpoely86 as an alternative to https://github.com/hpcugent/easybuild-easyconfigs/pull/1920 .

It fixes https://github.com/hpcugent/easybuild-easyconfigs/issues/1842, and also makes one of the required patch file quite a bit smaller (cfr. https://github.com/hpcugent/easybuild-easyconfigs/pull/1968).


boegel:
cc @FokkeDijkstra

boegel:
@JensTimmerman: please review

JensTimmerman:
ok

boegel:

I'll make sure this doesn't break any of the existing OpenFOAM builds before I merge it in...

boegel:
Some more shuffling around was needed to make sure that the extra compiler flags were injected as well, all OpenFOAM and OpenFOAM-Extend builds now still work.

Going in, thanks for the review @JensTimmerman!

PR #694, opened 2015-09-18
Fix getting pylibdir in easyblock for python_meep

jas02:
python-meep package easyblock is broken. Currently, there is inherited method configure_step() that don't handle self.pylibdir variable. So during installation we have self.pylibdir 'UNKNOWN' and  numpyinc like:

```
/apps/Python/2.7.9-intel-2015b/UNKNOWN/numpy/core/include
```

instead of correct one 

```
/apps/Python/2.7.9-intel-2015b/lib/python2.7/site-packages/numpy/core/include
```

**Also, there is problem, that this easyblock count on numpy package in Python instance, not as separate module. This dependency is not mentioned in easyconfig.**

### DIFF
--- a/easybuild/easyblocks/p/python_meep.py
+++ b/easybuild/easyblocks/p/python_meep.py
@@ -36,7 +36,7 @@ import os
 import shutil
 import tempfile
 
-from easybuild.easyblocks.generic.pythonpackage import PythonPackage, det_pylibdir, UNKNOWN
+from easybuild.easyblocks.generic.pythonpackage import PythonPackage
 from easybuild.tools.build_log import EasyBuildError
 from easybuild.tools.filetools import extract_file, rmtree2
 from easybuild.tools.modules import get_software_root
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/p/python_meep.py
```diff
@@ -56,6 +56,9 @@ def configure_step(self):
         for dep in deps:
             if not get_software_root(dep):
                 raise EasyBuildError("Module for %s not loaded.", dep)
+        # prepare easyblock by determining Python site lib dir
+        if self.pylibdir == UNKNOWN:
+            self.pylibdir = det_pylibdir()
```
boegel: actually, a better fix would be to simply add the missing super call:

```python
super(EB_python_minus_meep, self).configure_step()
```

That will take care of this too, and more (debug logging w.r.t. Python version being used, for example)

boegel:
Jenkins: ok to test

boegel:
@jas02: w.r.t. numpy: maybe we should include a check command like `python -c "import numpy"` in this easyblock, and complain if that fails, to make sure that numpy is available as required?

adding a comment in the python_meep easyconfigs wouldn't hurt either?

boegel:
Going in for EB v2.4.0, thanks @jas02!

PR #698, opened 2015-09-19
add support in PythonPackage for installing with easy_install + installing zipped eggs (REVIEW)

boegel:


easybuild/easyblocks/generic/pythonpackage.py
```diff
@@ -108,6 +110,21 @@ def __init__(self, *args, **kwargs):
...
+
+        self.install_cmd = "python setup.py install"
+        if self.cfg['use_easy_install']:
+            if which('easy_install') is None:
+                raise EasyBuildError("easy_install command not found")
+
+            # mainly for debugging
+            run_cmd("easy_install --version")
+
+            self.install_cmd = "easy_install"
```
boegel: we need to add a `.` at the very end of the command...

easybuild/easyblocks/generic/pythonpackage.py
```diff
@@ -108,6 +110,21 @@ def __init__(self, *args, **kwargs):
...
+
+        self.install_cmd = "python setup.py install"
+        if self.cfg['use_easy_install']:
+            if which('easy_install') is None:
+                raise EasyBuildError("easy_install command not found")
+
+            # mainly for debugging
+            run_cmd("easy_install --version")
+
+            self.install_cmd = "easy_install --no-deps"
```
boegel: TODO: make sure `.` is added at the end of the install commands being executed

boegel:
Jenkinss: test this please

easybuild/easyblocks/generic/pythonpackage.py
```diff
@@ -193,9 +212,8 @@ def test_step(self):
                 abs_pylibdirs = [os.path.join(testinstalldir, pylibdir) for pylibdir in self.all_pylibdirs]
                 extrapath = "export PYTHONPATH=%s && " % os.pathsep.join(abs_pylibdirs + ['$PYTHONPATH'])
 
-                tup = (extrapath, self.cfg['preinstallopts'], testinstalldir, self.cfg['installopts'])
-                cmd = "%s%s python setup.py install --prefix=%s %s" % tup
-                run_cmd(cmd, log_all=True, simple=True)
+                tup = (extrapath, self.cfg['preinstallopts'], self.install_cmd, testinstalldir, self.cfg['installopts'])
+                run_cmd("%s%s %s --prefix=%s %s" % tup, log_all=True, simple=True)
```
wpoely86: why not add a space between the first two?

wpoely86:
lgtm

easybuild/easyblocks/generic/pythonpackage.py
boegel: I could, but there's already a space at the end in `extrapath`; maybe removed it there and adding it here is clearer

easybuild/easyblocks/generic/pythonpackage.py
```diff
@@ -108,6 +113,26 @@ def __init__(self, *args, **kwargs):
         if not 'modulename' in self.options:
             self.options['modulename'] = self.name.lower()
 
+        if self.cfg.get('zipped_egg', False) and not self.cfg.get('use_easy_install', False):
+            raise EasyBuildError("Installing zipped eggs requires use_easy_install")
```
wpoely86: `"Installing zipped eggs requires use_easy_install=True"`

easybuild/easyblocks/generic/pythonpackage.py
```diff
@@ -108,6 +113,26 @@ def __init__(self, *args, **kwargs):
...
+
+        if self.cfg.get('use_easy_install', False):
+            if which('easy_install') is None:
```
wpoely86: `EASY_INSTALL_CMD`

boegel:

ocaisa:
For me this is triggering
```
ERROR: Build of /usr/local/software/juropa3/Stage5/software/Core/EasyBuild/develop/easybuild-easyconfigs/easybuild/easyconfigs/g/GC3Pie/GC3Pie-2.4.2.eb failed (err: "build failed (first 300 chars): Use of unknown easyconfig parameter 'use_easy_install' when getting parameter value")
```

boegel:
@ocaisa: do you have a some more context? In which step did this error occur?

easybuild/easyblocks/generic/pythonpackage.py
```diff
@@ -165,9 +203,9 @@ def configure_step(self):
 
     def build_step(self):
         """Build Python package using setup.py"""
-
-        cmd = "%s python setup.py build %s" % (self.cfg['prebuildopts'], self.cfg['buildopts'])
-        run_cmd(cmd, log_all=True, simple=True)
+        if not self.cfg['use_easy_install']:
```
boegel: @ocaisa: this may be the culprit... If we use the below, you shouldn't run into the problem you're having.

```python
if not self.cfg.get('use_easy_install', None):
```

The problem is that `use_easy_install` is not a general easyconfig parameter, but one specific to `PythonPackage`.

Although, even with that in mind, this line should never yield the error you're seeing...

Can you provide a full (debug) log?

PR #699, opened 2015-09-22
added c-interface (lapacke) support

DirkdeDraak:


boegel:
Jenkins: ok to test

boegel:
looks fine, but we'll need to make sure this works with all existing LAPACK easyconfig we have (there should be no regressions)

I'll take a look at that.

easybuild/easyblocks/l/lapack.py
```diff
@@ -143,6 +143,10 @@ def configure_step(self):
             self.log.info('No BLAS library provided, so only building LAPACK library (no testing).')
             self.cfg.update('buildopts', 'lib')
 
+
+        if os.path.exists(os.path.join(self.cfg['start_dir'], 'lapacke')):
```
boegel: how about just checking the version? this should be there in LAPACK 3.4.0 and up

boegel: see https://github.com/DirkdeDraak/easybuild-easyblocks/pull/1, if you merge my PR to your branch, it'll be included here

boegel:
@DirkdeDraak: hmm, this doesn't work as expected...

It results in running this:

```
make -j 10  OPTS="$FFLAGS -m64" NOOPT="-fPIC -m64 -O0"  BLASLIB=""  lib  lapackelib
```

which yields:

```
make: *** No rule to make
```

because you can't specify two targets (`lib` and `lapackelib`) at the same time...

PR #700, opened 2015-09-24
Scons test step made optional 

hajgato:


hajgato:
@boegel line inserted

easybuild/easyblocks/generic/scons.py
```diff
@@ -54,7 +54,8 @@ def test_step(self):
         Test with SCons 
         """
         cmd = "scons check"
-        (out, _) = run_cmd(cmd, log_all=True)
+        if self.cfg['runtest']:
+            (out, _) = run_cmd(cmd, log_all=True)
```
wpoely86: would there be any harm in doing it exactly like in the `ConfigureMake` block?
```python
        if self.cfg['runtest']:
            cmd = "make %s" % (self.cfg['runtest'])
            (out, _) = run_cmd(cmd, log_all=True, simple=False)
```
It might come in handy later on?

boegel: I think `scons check` is different from `make test`, where `test` is a common target but it can be different, while I expect it's always `scons check`?

wpoely86: As far as I understand, `check` is just a target like in a Makefile. It can have another name, it can take arguments. So, why not add the flexibility?

wpoely86: https://bitbucket.org/scons/scons/wiki/UnitTests

hajgato:
I am not an `SCons` expert, so I am not able to decide what is best. I think we can just let this is as it is, and if we have further needs, we will modify the easyblock.

easybuild/easyblocks/generic/scons.py
boegel: In that case, it makes sense to do it as in `ConfigureMake`, indeed.

easybuild/easyblocks/generic/scons.py
```diff
@@ -54,7 +54,9 @@ def test_step(self):
         Test with SCons 
         """
         cmd = "scons check"
```
wpoely86: this line can go

easybuild/easyblocks/generic/scons.py
```diff
@@ -54,7 +54,9 @@ def test_step(self):
...
-        (out, _) = run_cmd(cmd, log_all=True)
+        if self.cfg['runtest']:
+            cmd = "scons %s" % (self.cfg['runtest'])
+            (out, _) = run_cmd(cmd, log_all=True)
 
         return out
```
wpoely86: will be undefined if the `if` is not run?

boegel: @hajgato: +1, I'd just drop the `return` and not catch the return value of `run_cmd`

boegel:
Thanks @hajgato!

PR #701, opened 2015-09-25
Boost: enable parallel builds and allow to completely deactivate mpi

torbjoernk:
Extending the Boost easyblock to read in the `parallel` config variable and make _bjam_/_b2_ do parallel building.

Also disable libboost-mpi completely if `boost_mpi==False`.

### DIFF
--- a/easybuild/easyblocks/b/boost.py
+++ b/easybuild/easyblocks/b/boost.py
@@ -89,6 +89,8 @@ class EB_Boost(EasyBlock):
     def configure_step(self):
         """Configure Boost build using custom tools"""
 
+        configopts = "--prefix=%s" % self.objdir
+
         # mpi sanity check
         if self.cfg['boost_mpi'] and not self.toolchain.options.get('usempi', None):
             raise EasyBuildError("When enabling building boost_mpi, also enable the 'usempi' toolchain option.")
@@ -111,11 +113,15 @@ class EB_Boost(EasyBlock):
             else:
                 raise EasyBuildError("Unknown compiler used, don't know what to specify to --with-toolset, aborting.")
 
-        cmd = "./bootstrap.sh --with-toolset=%s --prefix=%s %s" % (toolset, self.objdir, self.cfg['configopts'])
+        configopts += " --with-toolset=%s" % toolset
+
+        if not self.cfg['boost_mpi']:
+            configopts += " --without-libraries=mpi"
+
+        cmd = "./bootstrap.sh %s %s" % (configopts, self.cfg['configopts'])
         run_cmd(cmd, log_all=True, simple=True)
 
         if self.cfg['boost_mpi']:
-
             self.toolchain.options['usempi'] = True
             # configure the boost mpi module
             # http://www.boost.org/doc/libs/1_47_0/doc/html/mpi/getting_started.html
@@ -128,6 +134,9 @@ class EB_Boost(EasyBlock):
         """Build Boost with bjam tool."""
 
         bjamoptions = " --prefix=%s" % self.objdir
+
+        if self.cfg['parallel']:
+            bjamoptions += " -j%s" % self.cfg['parallel']
 
         # specify path for bzip2/zlib if module is loaded
         for lib in ["bzip2", "zlib"]:
--- a/easybuild/easyblocks/b/boost.py
+++ b/easybuild/easyblocks/b/boost.py
@@ -89,7 +89,7 @@ class EB_Boost(EasyBlock):
     def configure_step(self):
         """Configure Boost build using custom tools"""
 
-        configopts = "--prefix=%s" % self.objdir
+        self.cfg.update('configopts', "--prefix=%s" % self.objdir)
 
         # mpi sanity check
         if self.cfg['boost_mpi'] and not self.toolchain.options.get('usempi', None):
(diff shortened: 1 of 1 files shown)


wpoely86:
ok to test

wpoely86:
looks good to me.

boegel:
@torbjoernk: why did you close this?

torbjoernk:
*ups* didn't meant to delete the branch

boegel:
Jenkins: ok to test

easybuild/easyblocks/b/boost.py
```diff
@@ -111,11 +113,15 @@ def configure_step(self):
             else:
                 raise EasyBuildError("Unknown compiler used, don't know what to specify to --with-toolset, aborting.")
 
-        cmd = "./bootstrap.sh --with-toolset=%s --prefix=%s %s" % (toolset, self.objdir, self.cfg['configopts'])
+        configopts += " --with-toolset=%s" % toolset
+
+        if not self.cfg['boost_mpi']:
+            configopts += " --without-libraries=mpi"
+
+        cmd = "./bootstrap.sh %s %s" % (configopts, self.cfg['configopts'])
```
boegel: let's use `preconfigopts` too, while we're at it?

```python
cmd = "%s ./bootstrap.sh %s %s" % (self.cfg['preconfigopts'], configopts, self.cfg['configopts'])
```

or (whatever looks nicer to you)

```python
cmd = ' '.join([self.cfg['preconfigopts'], 'bootstrap.sh', self.cfg['configopts']])
```

easybuild/easyblocks/b/boost.py
```diff
@@ -89,6 +89,8 @@ def patch_step(self):
     def configure_step(self):
         """Configure Boost build using custom tools"""
 
+        configopts = "--prefix=%s" % self.objdir
```
boegel: use `self.cfg['configopts'].update("--prefix=%s" % self.objdir)`, and drop `configopts` (same below)

easybuild/easyblocks/b/boost.py
```diff
@@ -111,11 +113,15 @@ def configure_step(self):
             else:
                 raise EasyBuildError("Unknown compiler used, don't know what to specify to --with-toolset, aborting.")
 
-        cmd = "./bootstrap.sh --with-toolset=%s --prefix=%s %s" % (toolset, self.objdir, self.cfg['configopts'])
+        configopts += " --with-toolset=%s" % toolset
```
boegel: use `self.cfg['configopts'].update(...)`

easybuild/easyblocks/b/boost.py
boegel: oh, and do this *after* `self.objdir` is defined below

so, this wasn't tested? :)

easybuild/easyblocks/b/boost.py
```diff
@@ -129,6 +135,9 @@ def build_step(self):
 
         bjamoptions = " --prefix=%s" % self.objdir
 
+        if self.cfg['parallel']:
+            bjamoptions += " -j%s" % self.cfg['parallel']
```
boegel: same here, use `self.cfg['buildopts'].update(...)` instead of defining/updating/using `bjamoptions`?

easybuild/easyblocks/b/boost.py
torbjoernk: Are you sure I want/can use `.update()` on `self.cfg['configopts']`? EB complains on not being able to call `.update()` on a string.

wpoely86: Use `+=` will do

torbjoernk: Right, but how will this work in case multiple configurations are specified in the easyconfig?
```{.py}
configopts = [
  " --shared",  # first build
  " --static"    # second build
]
```
Though this is not an issue right now for the Boost configs.

wpoely86: @boegel can correct if I'm wrong but it will simply be run multiple times. each time with a different value for `configopts`.

boegel: @wpoely86 is right, don't worry about list values in the easyconfig file for `configopts`, the `configure_step` will run for each of the values once

@torbjoernk: my syntax was wrong, use `self.cfg.update('configopts', "...")`

easybuild/easyblocks/b/boost.py
boegel: self.cfg.update('configopts', ...)

boegel:
@torbjoernk: ping on the open remarks? let us know if you want us to take this over from you (we can create a wrapper PR to continue on top of your commits)

torbjoernk:
Sorry for my extensive silence and thanks for the ping. Was distracted by other stuff.

easybuild/easyblocks/b/boost.py
```diff
@@ -127,31 +133,34 @@ def configure_step(self):
...
+            self.cfg.update('buildopts', "-j%s" % self.cfg['parallel'])
 
         # specify path for bzip2/zlib if module is loaded
         for lib in ["bzip2", "zlib"]:
             libroot = get_software_root(lib)
             if libroot:
-                bjamoptions += " -s%s_INCLUDE=%s/include" % (lib.upper(), libroot)
-                bjamoptions += " -s%s_LIBPATH=%s/lib" % (lib.upper(), libroot)
+                self.cfg.update('buildopts', "-s%s_INCLUDE=%s/include" % (lib.upper(), libroot))
+                self.cfg.update('buildopts', "-s%s_LIBPATH=%s/lib" % (lib.upper(), libroot))
```
wpoely86: maybe also add `lib64`?

wpoely86:
lgtm

easybuild/easyblocks/b/boost.py
boegel: @torbjoernk: ping?

boegel: @torbjoernk: reping?

boegel:
Closing, see #1005 and the `boost_mpi` easyconfig parameter

PR #705, opened 2015-09-26
Flex: don't only check in `lib` but also `lib64`

torbjoernk:
On openSUSE 13.2 64bit there is only `lib64`.


boegel:
Jenkins: ok to test

boegel:
Looks great, going in, thanks @torbjoernk!

Congratulations on your first contribution to EasyBuild! :+1:
Hopefully the first of many to come...

PR #707, opened 2015-10-07
enhance Bowtie easyblock: install additional binaries, make it version-aware

boegel:
relates to https://github.com/hpcugent/easybuild-easyconfigs/pull/2035


ehiggs:
:+1: Fixes a bug in v1.1.0 where more executables not listed were added to Bowtie.  For 1.1.2 they moved to a more traditional approach that could be handled by `MakeCp` but I can see it making sense that we have an easy block for bowtie so we should use it.

boegel:
retested with all existing Bowtie easyconfigs + https://github.com/hpcugent/easybuild-easyconfigs/pull/2035, so good to go

thanks for the review @ehiggs!

PR #708, opened 2015-10-07
some fixes for cuda easyblock for CUDA SDK 7.5

molden:
This PR depends on ~~https://github.com/hpcugent/easybuild-framework/pull/1417~~ .

### DIFF
--- a/easybuild/easyblocks/c/cuda.py
+++ b/easybuild/easyblocks/c/cuda.py
@@ -74,8 +74,9 @@ class EB_CUDA(Binary):
         # this is workaround for not being able to specify --nox11 to the Perl install scripts
         if 'DISPLAY' in os.environ:
             os.environ.pop('DISPLAY')
-
-        run_cmd_qa(cmd, qanda, std_qa=stdqa, no_qa=noqanda, log_all=True, simple=True, maxhits=1500)
+    
+        #overriding maxhits default value to 300 (300s wait for nothing to change in the output without seeing a known question)
+        run_cmd_qa(cmd, qanda, std_qa=stdqa, no_qa=noqanda, log_all=True, simple=True, maxhits=300)
 
     def sanity_check_step(self):
         """Custom sanity check for CUDA."""
@@ -88,7 +89,7 @@ class EB_CUDA(Binary):
 
         extra_files = []
         if LooseVersion(self.version) < LooseVersion('7.5'):
-	    extra_files.append('open64/bin/ncopencc')
+            extra_files.append('open64/bin/ncopencc')
 
         custom_paths = {
             'files': ["bin/%s" % x for x in ["fatbinary", "nvcc", "nvlink", "ptxas"]] +
--- a/easybuild/easyblocks/c/cuda.py
+++ b/easybuild/easyblocks/c/cuda.py
@@ -75,7 +75,7 @@ class EB_CUDA(Binary):
         if 'DISPLAY' in os.environ:
             os.environ.pop('DISPLAY')
 
-        run_cmd_qa(cmd, qanda, std_qa=stdqa, no_qa=noqanda, log_all=True, simple=True)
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/c/cuda.py
```diff
@@ -89,8 +89,7 @@ def sanity_check_step(self):
         custom_paths = {
             'files': ["bin/%s" % x for x in ["fatbinary", "nvcc", "nvlink", "ptxas"]] +
                      ["%s/lib%s.so" % (x, y) for x in chk_libdir for y in ["cublas", "cudart", "cufft",
-                                                                           "curand", "cusparse"]] +
-                     ["open64/bin/nvopencc"],
```
boegel: so, this is no longer around for recent versions? we should only drop this for recent versions then?

something like this:

```python
extra_files = []
if LooseVersion(self.version) < LooseVersion('7.5'):
    extra_files.append('open64/bin/ncopencc')

custom_paths = {
    ...
```

and then use `extra_files` in here?

boegel:
Jenkins: test this please

easybuild/easyblocks/c/cuda.py
```diff
@@ -75,7 +75,7 @@ def install_step(self):
         if 'DISPLAY' in os.environ:
             os.environ.pop('DISPLAY')
 
-        run_cmd_qa(cmd, qanda, std_qa=stdqa, no_qa=noqanda, log_all=True, simple=True)
+        run_cmd_qa(cmd, qanda, std_qa=stdqa, no_qa=noqanda, log_all=True, simple=True, maxhits=1500)
```
boegel: this allows the installer to hang without producing new output for 1500 seconds before EB gives up... isn't that *very* long?

maybe it's just a matter of fleshing out the `noqanda` a bit?

how did you see it hang exactly?

easybuild/easyblocks/c/cuda.py
boegel: OK, covered now, thanks!

easybuild/easyblocks/c/cuda.py
molden: Well it produces indeed for a long time no output apparently. But I will have a look to reduce it a little a bit with your suggestion.

boegel: ping me when it's tackled, and I'll test it with existing CUDA easyconfigs + https://github.com/hpcugent/easybuild-easyconfigs/issues/2039

molden: Actually I think it was yesterday a local problem. Access to storage was very slow temporarily. I guess we can reduce it to let's say 300?

boegel: Sure, just add a comment maybe why you're overriding the default value, and what is means (300s wait for nothing to change in the output without seeing a known question)

boegel:
Jenkins: test this please

easybuild/easyblocks/c/cuda.py
```diff
@@ -86,11 +87,14 @@ def sanity_check_step(self):
         if LooseVersion(self.version) < LooseVersion("6"):
             chk_libdir += ["lib"]
 
+        extra_files = []
+        if LooseVersion(self.version) < LooseVersion('7'):
+            extra_files.append('open64/bin/ncopencc')
```
boegel: typo: should be `nvopencc`, not `ncopencc` (yaay for testing)

boegel:
Jenkins: ok to test

boegel:
good to go, retested will existing CUDA easyconfigs + the one in https://github.com/hpcugent/easybuild-easyconfigs/pull/2039, see https://github.com/hpcugent/easybuild-easyconfigs/pull/2039#issuecomment-146819673

boegel:
Thanks @molden!

PR #709, opened 2015-10-08
Eigen: also copy signature file and don't copy CMake files

torbjoernk:
The signature file is used by CMake's FindEigen module to determine the
include path of Eigen3.

The CMakeLists.txt files are only used by Eigen3's CMake based installation.


boegel:
Jenkins: ok to test

boegel:
Looks great, tested with all existing Eigen easyconfigs, works like a charm, so going in.

Thanks @torbjoernk!

PR #711, opened 2015-10-13
use less evil 'python setup.py easy_install' instead of 'easy_install'

boegel:
Apparently, `easy_install` is not equivalent with `python setup.py easy_install`...

When using `easy_install`, Python package installations listed in `easy-install.pth` files that are available through `$PYTHONPATH` (and probably also in `~/.local`) get preference over other installations in the Python search path, which can (and does) cause problems.

One example where this is an issue is installing `vsc-base-2.4.1.eb` on a system where an older version of vsc-base (<=v2.3) is available through `easy-install.pth`.

Using `python setup.py easy_install`, these problems do not occur.


wpoely86:
lgtm

boegel:

boegel:

PR #712, opened 2015-10-15
Ansys

valtandor:
Some directory names for additional entries in the module file had been misspelled. The easyblock is corrected accordingly.


boegel:
replaced by #713

PR #715, opened 2015-10-15
be more careful with checking 'use_easy_install' in PythonPackage.build_step

boegel:
fix for issue reported at https://github.com/hpcugent/easybuild-easyblocks/pull/698#issuecomment-148432391 by @ocaisa


boegel:
@ocaisa: this should solve the problem you ran into, please confirm by retesting with `develop`

Issue #716, opened 2015-10-15
Backtrace when building Python w/ numpy on wrong toolchain


thiell:
Numpy + Intel MKL requires a specific numpy patch. I found a small defect in numpy.py when trying a non-intel toolchain (I know, my bad, a mistake!):

https://gist.github.com/thiell/8acdaa4c4c5194c8b24d

An exception is already in place a few lines below to explain the user a specific patch is needed, but is not raised with the current code. Seems easy to fix, I plan to submit a patch through a pull request ASAP.

boegel:
@thiell: if the output you want to refer to is reasonable small, it's better to include it in the issue itself, for future reference (the gist may be removed, yielding loss of context)

```
$ eb --configfiles=/tmp/config.cfg SWIG-3.0.7-foss-2015a-Python-2.7.9.eb  --try-toolchain=intel,2015.5.223 --robot intel-2015.5.223/
== temporary log file in case of crash /tmp/eb-hHtY6G/easybuild-bS1o_a.log
== resolving dependencies ...
== processing EasyBuild easyconfig /tmp/eb-hHtY6G/tweaked_easyconfigs/Python-2.7.9-intel-2015.5.223.eb
== building and installing MPI/intel/2015.5.223-GNU-4.9.2-2.25/impi/5.0.3.049/Python/2.7.9...
== fetching files...
== creating build dir, resetting environment...
== unpacking...
== patching...
== preparing...
== configuring...
== building...
== testing...
== installing...
== taking care of extensions...
ERROR: Traceback (most recent call last):
  File "/global/software/Core/EasyBuild/2.3.0/lib/python2.6/site-packages/easybuild_framework-2.3.0-py2.6.egg/easybuild/main.py", line 112, in build_and_install_software
    (ec_res['success'], app_log, err) = build_and_install_one(ec, init_env)
  File "/global/software/Core/EasyBuild/2.3.0/lib/python2.6/site-packages/easybuild_framework-2.3.0-py2.6.egg/easybuild/framework/easyblock.py", line 1994, in build_and_install_one
    result = app.run_all_steps(run_test_cases=run_test_cases)
  File "/global/software/Core/EasyBuild/2.3.0/lib/python2.6/site-packages/easybuild_framework-2.3.0-py2.6.egg/easybuild/framework/easyblock.py", line 1934, in run_all_steps
    self.run_step(step_name, step_methods)
[...]

PR #717, opened 2015-10-16
numpy: test whether self.patches is None (with mkl)

thiell:
Patch proposal for issue #716.

### DIFF
--- a/easybuild/easyblocks/n/numpy.py
+++ b/easybuild/easyblocks/n/numpy.py
@@ -127,13 +127,14 @@ class EB_numpy(FortranPythonPackage):
             # not fool-proof, but better than enforcing a particular patch filename
             patch_found = False
             patch_wl_regex = re.compile(r"replace\(':',\s*','\)")
-            for patch in self.patches:
-                # patches are either strings (extension) or dicts (easyblock)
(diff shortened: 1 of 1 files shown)


boegel:
@thiell: although this will fix the problem for numpy, the problem actually lies higher up, in the EasyBuild framework, i.e. here:

```
$ sed -n "60,61p" easybuild/framework/extension.py
        self.src = self.ext.get('src', None)
        self.patches = self.ext.get('patches', None)
```

The fallback value `None` used above doesn't correspond with the default `[]` value for `self.patches` and `self.src` at:

```
$ sed -n "138,139p" easybuild/framework/easyblock.py
        self.patches = []
        self.src = []

```

So, the fix should be made in `easybuild/framework/extension.py` instead, since this problem is not specific to the numpy easyblock.

Are you up for looking into that?

thiell:
@boegel: ok thx for looking, I understand, I will look at it ASAP and submit another pull request (to easybuild-framework).

boegel:
@thiell: let's close this one since https://github.com/hpcugent/easybuild-framework/issues/1434 provides a better solution?

thiell:
@boegel: sure, let's close it, thx!

PR #719, opened 2015-10-19
implement support for 'use_pip' in PythonPackage easyblock

boegel:


easybuild/easyblocks/generic/pythonpackage.py
```diff
@@ -114,20 +117,31 @@ def __init__(self, *args, **kwargs):
         if not 'modulename' in self.options:
             self.options['modulename'] = self.name.lower()
 
-        if self.cfg.get('zipped_egg', False) and not self.cfg.get('use_easy_install', False):
-            raise EasyBuildError("Installing zipped eggs requires use_easy_install = True")
-
+        self.use_setup_py = False
         if self.cfg.get('use_easy_install', False):
```
wpoely86: add a debug output about what is chosen?

easybuild/easyblocks/generic/pythonpackage.py
```diff
@@ -114,20 +117,31 @@ def __init__(self, *args, **kwargs):
...
+
+            # don't auto-install dependencies
+            self.cfg.update('installopts', '--no-deps')
+
+            if self.cfg.get('zipped_egg', False):
+                self.cfg.update('installopts', '--egg')
 
         else:
+            self.use_setup_py = True
             self.install_cmd = "python setup.py install"
```
wpoely86: why not makes this one a constant?  It is more consistent?

wpoely86:
retest this please

dominikborkowski:
While on the subject of being able to leverage use_pip in PythonPackage, could this include support for wheel packages in addition to eggs? 

https://pypi.python.org/pypi/wheel

Our specific use case would be ability to deploy TensorFlow:

https://www.tensorflow.org/versions/0.6.0/get_started/os_setup.html#pip_install

easybuild/easyblocks/generic/pythonpackage.py
boegel: done below

boegel:
@wpoely86: please re-review? It would be great to get this in (required for https://github.com/hpcugent/easybuild-easyconfigs/pull/2461, which is required for IPython 4.x (cfr. https://github.com/hpcugent/easybuild-easyconfigs/pull/2457), which would be nice to support in HOD).

wpoely86:
lgtm

boegel:

PR #721, opened 2015-10-26
BinariesTarball

valtandor:
Removal of inconsistent execute permissions on certain EasyBlocks, and addition of functionality to BinariesTarball so that it can take a list of binaries (assuming that, sometimes, not everything in start_dir is required to be copied).


boegel:
Jenkins: ok to test

boegel:
@wpoely86: please review?

wpoely86:
lgtm

easybuild/easyblocks/generic/binariestarball.py
```diff
@@ -40,21 +41,43 @@ class BinariesTarball(Tarball):
...
+        extra_vars = Tarball.extra_options(extra_vars)
+        # Allow specification of files to copy explicitly.
+        # If files_to_copy is a zero-length list, all files (but not
+        # directories, etc.) in start_dir will be copied with their
+        # current names.
+        extra_vars['files_to_copy'] = [
+            [],
+            "List of optional (source_file, destination_file) tuples",
+            CUSTOM,
+        ]
```
boegel: keep this on a single line, since it's really one entry (not three)

easybuild/easyblocks/generic/binariestarball.py
```diff
@@ -40,21 +41,43 @@ class BinariesTarball(Tarball):
...
+        """
+        Define list of files to be copied during the installation process
+        """
+        extra_vars = Tarball.extra_options(extra_vars)
+        # Allow specification of files to copy explicitly.
+        # If files_to_copy is a zero-length list, all files (but not
+        # directories, etc.) in start_dir will be copied with their
+        # current names.
+        extra_vars['files_to_copy'] = [
+            [],
```
boegel: use `None` as default value?

easybuild/easyblocks/generic/binariestarball.py
```diff
@@ -40,21 +41,43 @@ class BinariesTarball(Tarball):
...
 
     def install_step(self):
         """Install by copying unzipped binaries to 'bin' subdir of installation dir, and fixing permissions."""
 
         bindir = os.path.join(self.installdir, 'bin')
+        items_to_copy = []
         try:
             os.makedirs(bindir)
-            for item in os.listdir(self.cfg['start_dir']):
+            if len(self.cfg['files_to_copy']) == 0:
```
boegel:
```python
if self.cfg['files_to_copy'] is None:
```

easybuild/easyblocks/generic/binariestarball.py
```diff
@@ -40,21 +41,43 @@ class BinariesTarball(Tarball):
...
+                for item in os.listdir(self.cfg['start_dir']):
+                    items_to_copy.append((os.path.join(self.cfg['start_dir'], item), item))
+            else:
+                items_to_copy = self.cfg['files_to_copy']
```
boegel: this implies that the entries should be specified full path, like `%(installpath)s/foo` rather than `foo`?

it would make more sense to also join these entries with `self.cfg['start_dir']`?

easybuild/easyblocks/generic/binariestarball.py
```diff
@@ -40,21 +41,43 @@ class BinariesTarball(Tarball):
...
                     # make sure binary has executable permissions
-                    adjust_permissions(os.path.join(bindir, item), stat.S_IXUSR|stat.S_IXGRP|stat.S_IXOTH, add=True)
+                    adjust_permissions(os.path.join(bindir, destname), stat.S_IXUSR|stat.S_IXGRP|stat.S_IXOTH, add=True)
                     self.log.debug("Copied %s to %s and fixed permissions" % (item, bindir))
                 else:
-                    self.log.warning("Skipping non-file %s in %s, not copying it." % (item, self.cfg['start_dir']))
+                    self.log.warning("%s: not a file. Skipping." % item)
         except OSError, err:
-            raise EasyBuildError("Copying binaries in %s to install dir 'bin' failed: %s", self.cfg['start_dir'], err)
+            raise EasyBuildError("Copying binaries to install dir 'bin' failed: %s" % err)
```
boegel: use `, err` please

easybuild/easyblocks/generic/binariestarball.py
```diff
@@ -40,21 +41,43 @@ class BinariesTarball(Tarball):
...
                 if os.path.isfile(item):
-                    shutil.copy2(os.path.join(self.cfg['start_dir'], item), bindir)
+                    shutil.copy2(item, os.path.join(bindir, destname))
                     # make sure binary has executable permissions
-                    adjust_permissions(os.path.join(bindir, item), stat.S_IXUSR|stat.S_IXGRP|stat.S_IXOTH, add=True)
+                    adjust_permissions(os.path.join(bindir, destname), stat.S_IXUSR|stat.S_IXGRP|stat.S_IXOTH, add=True)
                     self.log.debug("Copied %s to %s and fixed permissions" % (item, bindir))
                 else:
-                    self.log.warning("Skipping non-file %s in %s, not copying it." % (item, self.cfg['start_dir']))
+                    self.log.warning("%s: not a file. Skipping." % item)
```
boegel: use `, item` please

easybuild/easyblocks/generic/binariestarball.py
```diff
@@ -40,21 +41,43 @@ class BinariesTarball(Tarball):
...
+            else:
+                items_to_copy = self.cfg['files_to_copy']
+            for (item, destname) in items_to_copy:
                 if os.path.isfile(item):
-                    shutil.copy2(os.path.join(self.cfg['start_dir'], item), bindir)
+                    shutil.copy2(item, os.path.join(bindir, destname))
                     # make sure binary has executable permissions
-                    adjust_permissions(os.path.join(bindir, item), stat.S_IXUSR|stat.S_IXGRP|stat.S_IXOTH, add=True)
+                    adjust_permissions(os.path.join(bindir, destname), stat.S_IXUSR|stat.S_IXGRP|stat.S_IXOTH, add=True)
                     self.log.debug("Copied %s to %s and fixed permissions" % (item, bindir))
```
boegel: use `, item, bindir` please

boegel:
Some small remarks, looks OK overall.

It would be great to have a testcase for this before merging it in (https://github.com/hpcugent/easybuild-easyconfigs/pull/2084 requires this, but is still WIP).

easybuild/easyblocks/generic/binariestarball.py
valtandor: No, as far as I can tell, if `start_dir` is not explicitly set, it is assumed to be `builddir` and paths (where not explicitly absolute) are computed relative to it. However, entries can be joined with `self.cfg['start_dir']` if that would make sense. I guess it would be more understandable. I'll make that change.

easybuild/easyblocks/generic/binariestarball.py
valtandor: Precedents elsewhere seem to be in favour of an empty list (see e.g. the `makesymlinks` extra variable in `rpm.py`).

easybuild/easyblocks/generic/binariestarball.py
valtandor: Only if initialising the value to None (see above) is considered more desirable.

valtandor:
@boegel: Regarding a test case for this EasyBlock, it's a curious one, given that (so far as I can tell) it's not actually used by any EasyConfigs or other EasyBlocks. Any test case would therefore have to be specially written. I'm sure we could do that between us, but I'm not sure how good a use of our time it would be.

easybuild/easyblocks/generic/binariestarball.py
valtandor: @boegel: Further thoughts regarding `None` as the default?

easybuild/easyblocks/generic/binariestarball.py
boegel: Using `None` as a default value makes more sense anyway; using mutable values (like lists) is considered evil for defaults.

See also http://docs.python-guide.org/en/latest/writing/gotchas/#mutable-default-arguments

valtandor:
@boegel: files_to_copy is now None by default. items_to_copy (a purely internal variable) is still an empty list so it can be appended to without further ado.

valtandor:
@boegel What do you think - develop a test case for this, or drop the PR and the branch? I'm inclined to the latter if there's no particular use case for it.

PR #724, opened 2015-10-27
add an option to allow removal of the -Dusethreads flag in Perl easyblock

valtandor:
Add an option to allow removal of the -Dusethreads flag

### DIFF
--- a/easybuild/easyblocks/p/perl.py
+++ b/easybuild/easyblocks/p/perl.py
@@ -62,7 +62,10 @@ class EB_Perl(ConfigureMake):
             '-Dinc_version_list=none',
         ])
         if self.cfg['use_perl_threads']:
-            configopts.append("-Dusethreads")
+            configopts = ' '.join([
+                self.cfg['configopts'],
+                '-Dusethreads',
+            ])
 
         cmd = './Configure -de %s -Dprefix="%s" ' % (configopts, self.installdir)
         run_cmd(cmd, log_all=True, simple=True)
--- a/easybuild/easyblocks/p/perl.py
+++ b/easybuild/easyblocks/p/perl.py
@@ -55,19 +55,15 @@ class EB_Perl(ConfigureMake):
         """
         Configure Perl build: run ./Configure instead of ./configure with some different options
         """
-        configopts = ' '.join([
-            self.cfg['configopts'],
-            '-Dcc="%s"' % os.getenv('CC'),
-            '-Dccflags="%s"' % os.getenv('CFLAGS'),
-            '-Dinc_version_list=none',
-        ])
+        configopts = []
+        configopts.append(self.cfg['configopts'])
+        configopts.append('-Dcc="{0}"'.format(os.getenv('CC')))
+        configopts.append('-Dccflags="{0}"'.format(os.getenv('CFLAGS')))
+        configopts.append('-Dinc_version_list=none')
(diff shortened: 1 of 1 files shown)


valtandor:
Yes.

boegel:
Jenkins: ok to test

easybuild/easyblocks/p/perl.py
```diff
@@ -42,17 +43,27 @@
...
         """
         configopts = ' '.join([
             self.cfg['configopts'],
-            "-Dusethreads",
             '-Dcc="%s"' % os.getenv('CC'),
             '-Dccflags="%s"' % os.getenv('CFLAGS'),
             '-Dinc_version_list=none',
         ])
+        if self.cfg['use_perl_threads']:
+            configopts.append("-Dusethreads")
```
boegel: `configopts` is never a list, so `.append` will not work...

easybuild/easyblocks/p/perl.py
```diff
@@ -42,17 +43,27 @@
...
+
         cmd = './Configure -de %s -Dprefix="%s" ' % (configopts, self.installdir)
```
boegel: maybe use `' '.join(configopts)` here instead of `configopts`, and remove it above (then the `.append` will work)

valtandor:
@boegel: I've replaced the append with a join, by analogy with the preceding block. I hope this addresses the problem.

easybuild/easyblocks/p/perl.py
```diff
@@ -42,17 +43,30 @@
...
             '-Dccflags="%s"' % os.getenv('CFLAGS'),
             '-Dinc_version_list=none',
         ])
+        if self.cfg['use_perl_threads']:
+            configopts = ' '.join([
+                self.cfg['configopts'],
+                '-Dusethreads',
+            ])
+
         cmd = './Configure -de %s -Dprefix="%s" ' % (configopts, self.installdir)
```
boegel: @valtandor: I would clean this up a bit as follows, i.e. construct `configopts` as a list first, and only `join` when it's used:

```python
configopts = [
...
]
if self.cfg['use_perl_threads']:
    configopts.append('-Dusethreads')

cmd = './Configure -de %s -Dprefix="%s" ' % (' '.join(configopts), self.installdir)
```

valtandor:
@boegel: Done.

easybuild/easyblocks/p/perl.py
```diff
@@ -42,18 +43,27 @@
...
-            '-Dcc="%s"' % os.getenv('CC'),
-            '-Dccflags="%s"' % os.getenv('CFLAGS'),
-            '-Dinc_version_list=none',
-        ])
-        cmd = './Configure -de %s -Dprefix="%s" ' % (configopts, self.installdir)
+        configopts = []
+        configopts.append(self.cfg['configopts'])
+        configopts.append('-Dcc="{0}"'.format(os.getenv('CC')))
+        configopts.append('-Dccflags="{0}"'.format(os.getenv('CFLAGS')))
+        configopts.append('-Dinc_version_list=none')
```
boegel: I didn't mean to imply you should use `.append` all over the place... just define the initial `configopts` in one go, and append `-Dusethreads` if desired

see https://github.com/valtandor/easybuild-easyblocks/pull/1

boegel:
Going in, thanks @valtandor!

Issue #728, opened 2015-10-29
Tkinter links with wrong Tk in Python


boegel:
```
    fig1 = plt.figure(figsize=(32, 18), dpi=60)
  File "/user/scratchdelcatty/gent/gvo000/gvo00002/vsc40023/easybuild_REGTEST/SL6/sandybridge/software/matplotlib/1.4.3-intel-2015b-Python-2.7.10/lib/python2.7/site-packages/matplotlib-1.4.3-py2.7-linux-x86_64.egg/matplotlib/pyplot.py", line 435, in figure
    **kwargs)
  File "/user/scratchdelcatty/gent/gvo000/gvo00002/vsc40023/easybuild_REGTEST/SL6/sandybridge/software/matplotlib/1.4.3-intel-2015b-Python-2.7.10/lib/python2.7/site-packages/matplotlib-1.4.3-py2.7-linux-x86_64.egg/matplotlib/backends/backend_tkagg.py", line 81, in new_figure_manager
    return new_figure_manager_given_figure(num, figure)
  File "/user/scratchdelcatty/gent/gvo000/gvo00002/vsc40023/easybuild_REGTEST/SL6/sandybridge/software/matplotlib/1.4.3-intel-2015b-Python-2.7.10/lib/python2.7/site-packages/matplotlib-1.4.3-py2.7-linux-x86_64.egg/matplotlib/backends/backend_tkagg.py", line 89, in new_figure_manager_given_figure
    window = Tk.Tk()
  File "/user/scratch/gent/vsc400/vsc40023/easybuild_REGTEST/SL6/sandybridge/software/Python/2.7.10-intel-2015b/lib/python2.7/lib-tk/Tkinter.py", line 1816, in __init__
    self._loadtk()
  File "/user/scratch/gent/vsc400/vsc40023/easybuild_REGTEST/SL6/sandybridge/software/Python/2.7.10-intel-2015b/lib/python2.7/lib-tk/Tkinter.py", line 1832, in _loadtk
    % (_tkinter.TK_VERSION, tk_version)
RuntimeError: tk.h version (8.6) doesn't match libtk.a version (8.5)
```

from the Python install log:

```
-L/usr/local/lib -L. -ltk8.5 -ltcl8.5 -lX11 -lpython2.7 -o build/lib.linux-x86_64-2.7/_tkinter.so
```

while TK 8.6.4 is included as a dependency.

boegel:
[...]

PR #729, opened 2015-10-29
use --with-tcltk* configure options to point to ensure Tcl/Tk deps are picked up

boegel:
fix for #728

### DIFF
--- a/easybuild/easyblocks/p/python.py
+++ b/easybuild/easyblocks/p/python.py
@@ -40,7 +40,7 @@ from distutils.version import LooseVersion
 
 from easybuild.easyblocks.generic.configuremake import ConfigureMake
 from easybuild.tools.build_log import EasyBuildError
-from easybuild.tools.modules import get_software_libdir, get_software_root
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/p/python.py
```diff
@@ -104,6 +104,26 @@ def configure_step(self):
                 line = re.sub(r"^#(\s*-L\$\(SSL\)/lib )", r"\1 -L$(SSL)/lib64 ", line)
                 sys.stdout.write(line)
 
+        tcl = get_software_root('Tcl')
+        tk = get_software_root('Tk')
+        if tcl and tk:
```
wpoely86: maybe warning in the `else`?

wpoely86:
lgtm

easybuild/easyblocks/p/python.py
boegel: People are free to use the system Tcl/Tk (e.g. by suing `--filter-deps`), which is fine.

boegel:
Testing confirms that this fixes the (original) issue reported in #728.

boegel:

PR #731, opened 2015-11-01
fix OpenSSL sanity check paths: lib/engines is a directory

boegel:
went undetected because of a bug in the sanity check implementation, fixed by https://github.com/hpcugent/easybuild-framework/pull/1436


boegel:
Merging as in trivial.

PR #732, opened 2015-11-01
correctly handle having both $FC(FLAGS) and $F90(FLAGS) defined when building MVAPICH2

boegel:
caused by now also having the EasyBuild framework define `$FC` next to `$F90`, see https://github.com/hpcugent/easybuild-framework/pull/1394


boegel:
@geimer or @wpoely86: please review?

rjeschmi:
It looks fine.

boegel:
fix confirmed, thanks for the review @rjeschmi!

PR #734, opened 2015-11-04
update Doxygen easyblock for 1.10.x (CMake) (REVIEW)

hajgato:


easybuild/easyblocks/d/doxygen.py
```diff
@@ -30,30 +30,36 @@
...
+from distutils.version import LooseVersion
 from easybuild.tools.run import run_cmd
+from easybuild.easyblocks.generic.cmakemake import CMakeMake
 
-
-class EB_Doxygen(ConfigureMake):
+class EB_Doxygen(CMakeMake):
     """Support for building/installing Doxygen"""
 
+    version_switch = LooseVersion(self.version) < LooseVersion("1.8.10")
```
boegel: `self` is not defined here

do the version check in `configure_step`, always import `ConfigureMake`

hajgato: boegel: Ok, but how I will switch to CMakeMake?

hajgato: I mean where and how? Apparently 3rd works with the old (but then the EasyBlock is identical to the old), but not for the new.

hajgato:
@boegel seems to work.

easybuild/easyblocks/d/doxygen.py
```diff
@@ -30,30 +30,32 @@
...
 
-
-class EB_Doxygen(ConfigureMake):
+class EB_Doxygen(CMakeMake):
     """Support for building/installing Doxygen"""
 
     def configure_step(self):
-        """Configure build using non-standard configure script (see prefix option)"""
+        if LooseVersion(self.version) < LooseVersion("1.8.10"):
+             """Configure build using non-standard configure script (see prefix option)"""
```
boegel: @hajgato: the docstring should stay above the `if`

boegel: and it should be updated since we now have two cases...

what was so special about the `--prefix` below btw? I must be blind...

easybuild/easyblocks/d/doxygen.py
```diff
@@ -30,30 +30,32 @@
...
 
-        cmd = "%s ./configure --prefix %s %s" % (self.cfg['preconfigopts'], self.installdir,
+             cmd = "%s ./configure --prefix %s %s" % (self.cfg['preconfigopts'], self.installdir,
                                                    self.cfg['configopts'])
```
boegel: this line should be intended too

easybuild/easyblocks/d/doxygen.py
hajgato: @boegel was without `=` 
the normal is `--prefix=foo/bar` and here is `--prefix foo/bar`

easybuild/easyblocks/d/doxygen.py
hajgato: @boegel what is the recommeded one?
this:
```
cmd = "%s ./configure --prefix %s %s" % (self.cfg['preconfigopts'], self.installdir,
                                         self.cfg['configopts'])
```
or this
```
cmd = "%s ./configure --prefix %s %s" % 
    (self.cfg['preconfigopts'], self.installdir, self.cfg['configopts'])
```

boegel: @hajgato: the first one, or this:

```python
cmd = "%(preconfigopts)s ./configure --prefix %(prefix)s %(configopts)s" % {
    'prefix': self.installdir,
}
```

easybuild/easyblocks/d/doxygen.py
```diff
@@ -30,30 +30,33 @@
...
+from easybuild.easyblocks.generic.cmakemake import CMakeMake
 
-
-class EB_Doxygen(ConfigureMake):
+class EB_Doxygen(CMakeMake):
     """Support for building/installing Doxygen"""
 
     def configure_step(self):
-        """Configure build using non-standard configure script (see prefix option)"""
+        """Configure build using non-standard configure prefix option (wothout `=`)
```
boegel: please fix the `wothout` typo?

boegel:
Going in, thanks @hajgato!

PR #736, opened 2015-11-09
take into account that $LIBS may not be defined in HPL easyblock

boegel:
This is the case with Cray toolchains, for example


boegel:
fixes broken HPL build with Cray toolchains, which manifested itself as:

```
/opt/cray/cce/8.3.12/cray-binutils/x86_64-unknown-linux-gnu/bin/ld: cannot find None: No such file or directory
```

after this was used as build command:

```
make -j 1 ... LINKFLAGS="-craype-verbose -O2 -L.../lib None"
```

PR #740, opened 2015-11-12
Sanity check fitted for Qt 5

hajgato:


easybuild/easyblocks/q/qt.py
```diff
@@ -33,7 +33,7 @@
 from easybuild.easyblocks.generic.configuremake import ConfigureMake
 from easybuild.tools.build_log import EasyBuildError
 from easybuild.tools.run import run_cmd_qa
-
+from distutils.version import LooseVersion
```
boegel: this import should go at the top to keep it consistent with elsewhere:

```
import os
```

boegel:
@hajgato: looks fine, but I guess this means only stuff that is well aware of Qt 5 can use Qt 5 as a dependency?

hajgato:
@boegel: Yes. And some of them are requesting Qt 5 (Qt 4 does not work anymore), like `CGAL 4.7`

wpoely86:
lgtm

boegel:
Jenkins: test this please

boegel:
Tested with both the Qt5 easyconfig in https://github.com/hpcugent/easybuild-easyconfigs/pull/2136 and an existing easyconfig for Qt 4.x, thanks @hajgato!

PR #741, opened 2015-11-12
support for multilib build of GCC on PowerPC

JackPerdue:


easybuild/easyblocks/g/gcc.py
```diff
@@ -249,6 +249,7 @@ def configure_step(self):
         if self.cfg['multilib']:
             glibc_32bit = [
                 "glibc.i686",  # Fedora, RedHat-based
+                "glibc.ppc",   # "" on Power
```
boegel: @JackPerdue: I'm not sure what you mean with the `""` in the comment?

boegel:
Jenkins: ok to test

easybuild/easyblocks/g/gcc.py
JackPerdue: "" == 'same as above'... i.e. Fedora, Redhat-based

boegel: OK, makes sense.

boegel:
@wpoely86: do you see any reason not to merge this?

Should we add a check to only consider the `.ppc` one when we're on a Linux/POWER system, and only consider the ones we have now when we're on a Linux/x86_64 system?

wpoely86:
@boegel that's probably a bit far fetched. If you have glibc.ppc on a x86_64, you're doing weird stuff. 

Good to merge IMHO

JackPerdue:
Alas, I don't know the equivalent for Debian or SUSE (or have a way to test) so that might be good later.

boegel:
Merging this one, thanks @JackPerdue!

PR #742, opened 2015-11-13
Fix module path extension of system compiler in HMNS setup

geimer:
Running `eb GCC-system.eb` in an HMNS setup will create a module file `GCC/system`, which extends `$MODULEPATH` with `<prefix>/modules/Compiler/GCC/<gcc_version>`.  However, this may conflict with an EasyBuild-created GCC installation of the same version.  With this PR, the modulepath extension will be `<prefix>/modules/Compiler/GCC/system`.

Note: This PR also removes the custom `post_install_step` which originally was intended to skip changing permissions in system directories.  However, this was moved to a separate step.  In addition, EB only tries to modify permissions in `<prefix>`, i.e., suppressing this step is actually not needed at all.

### DIFF
--- a/easybuild/easyblocks/generic/systemcompiler.py
+++ b/easybuild/easyblocks/generic/systemcompiler.py
@@ -137,10 +137,6 @@ class SystemCompiler(Bundle):
         """
         return {}
 
-    def post_install_step(self):
-        """Custom post install step: Do not to touch system compiler directories and files."""
-        pass
-
     def make_module_step(self, fake=False):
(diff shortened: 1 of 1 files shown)


boegel:
Jenkins: ok to test

boegel:
@geimer: looks fine, but maybe the comments should make it very clear what is meant by 'real version' and 'original version', for example by including examples mentioning `'system'` or `'4.9.2'`?

boegel:
retested installing `GCC-system.eb` in a couple of places on top of this, works fine, thanks @geimer!

PR #743, opened 2015-11-16
only restore $PYTHONPATH if it was defined

boegel:


JensTimmerman:
can we have a test for this?

JensTimmerman:
wouldn't it not make sense to have some unittests for the generic EasyBlocks?

boegel:
@JensTimmerman: you mean like, creating small toy programs that are built using the generic easyblocks? it can get tricky expensive, for example if CMake is involved, or for OCamlPackage, etc...

We should do proper testing with the actual applications instead when updating easyblocks (which I typically do when an existing easyblock is touched).

boegel:
@wpoely86: please review?

wpoely86:
looks good

boegel:
Thanks @wpoely86!

PR #748, opened 2015-11-25
Add missing self.set_pylibdirs()

blappm:
eb --module-only skips the PYTHONPATH define, resulting in a corrupt PYTHONPATH in the created module file.

This happens independent of an installed EB software, so it even happens if the software was installed before and you recreate the modulefile.


boegel:
Jenkins: ok to test

boegel:
Going in, thanks @blappm!

PR #750, opened 2015-11-25
add support for IBM XL compilers

JackPerdue:
Notes:
## 1) I need a "base" install to build upon once configured with GCC.  i.e. I currently am working on:

GCC/4.8.4-CLooG-multilib
xlmass/8.1.0
xlsmp/4.1.0
xlcbase/13.01.0
xlc/13.01.0-GCC-4.8.4-CLooG-multilib
xlfbase/15.01.0
xlf/15.01.0-GCC-4.8.4-CLooG-multilib
xlcxlf/2015B-484
MPICH2/1.5-xlcxlf-2015B-484
xlmpich2/1.5-2015B-484
## zlib/1.2.8-xlmpich2-1.5-2015B-484

I can't have xlc configured with GCC load a plain xlc so I need to call it something different (e.g. xlcbase).  And I've already written the framework to use xlc/xlf.

2) The version check was eased because:
    a) I'd prefer to specify a version for our users (vs. just "system")
    b) I don't much care for the version the compiler spits out (see comment at top)

### DIFF
--- a/easybuild/easyblocks/generic/systemcompiler.py
+++ b/easybuild/easyblocks/generic/systemcompiler.py
@@ -53,6 +53,7 @@ class SystemCompiler(Bundle):
         # look for 3-4 digit version number, surrounded by spaces
         # examples:
         # gcc (GCC) 4.4.7 20120313 (Red Hat 4.4.7-11)
+        # IBM XL C/C++ for Linux, V13.1\nVersion: 13.01.0000.0000
         # Intel(R) C Intel(R) 64 Compiler XE for applications running on Intel(R) 64, Version 15.0.1.133 Build 20141023
         version_regex = re.compile(r'\s([0-9]+(?:\.[0-9]+){1,3})\s', re.M)
         res = version_regex.search(txt)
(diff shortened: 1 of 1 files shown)


boegel:
Jenkins: ok to test

boegel:
@JackPerdue: would be great to have a test report for corresponding easyconfigs that uses this

omula:
@JackPerdue I am planning to try xl compilers but I am bit lost.
I see that there's xlmass which is a math library. Does it replace BLAS, FFTW, ScalaPACK, etc?
I see that there's also ESSL and xlsmp.
I would be really happy if you can provide some guidance.

I tried to go to your website but it seems to be down.
http://www.siliconslick.com/easybuild/toolchains/compiler/ibmxl.py

Thanks in advance.

JackPerdue:
Hmmm.... I guess siliconslick.com is down at the moment..... should be back in a bit.
In the meantime, I'm  a bit rusty but:  xlsmp and xlmass are components of the xlc and xlf compilers.  I tried to break out the common libraries between xlc/xlf (much like we need to do with icc/ifort).  For linear algebra, you will want ESSL. I'd point you to another directory on the web site but...

omula:
It seems that siliconslick is back :) I've been navigate through the directories and this is quite interesting
http://www.siliconslick.com/easybuild/easyconfigs/curie/
Is there any other path I should have a look to?

Thank you!

JackPerdue:
That's the place unless you have a BlueGene in which case you'll want http://www.siliconslick.com/easybuild/easyconfigs/bgsys/

Maybe start with xlcbase/13.1.0.1 and xlfbase/15.1.0.1 since I think they were my latest "best" effort (3 years ago). Maybe not.  It is all kind of a blurry mess at this point. We tossed our BlueGene rental (nobody was using it) and dropped Power compiler support so it has just withered.   I mostly just use foss now (make life much simpler).

boegel:
I don't think this is going anywhere, so I'll go ahead and close this...

PR #755, opened 2015-11-27
avoid adding lib subdirs to $*LIBRARY_PATH if no libraries are there in PythonPackage easyblock

boegel:


boegel:
@wpoely86: please review?

easybuild/easyblocks/generic/pythonpackage.py
```diff
@@ -294,6 +294,25 @@ def sanity_check_step(self, *args, **kwargs):
...
+        for envvar in ['LD_LIBRARY_PATH', 'LIBRARY_PATH']:
+            newlist = []
+            for subdir in guesses[envvar]:
+                # only subdirectories that contain one or more files/libraries should be retained
+                fullpath = os.path.join(self.installdir, subdir)
+                if os.path.exists(fullpath) and any([os.path.isfile(x) for x in os.listdir(fullpath)]):
+                    newlist.append(subdir)
+            guesses[envvar] = newlist
+
+        return guesses
```
wpoely86: maybe add a debug print about which directories are retained?

wpoely86:
lgtm

easybuild/easyblocks/generic/pythonpackage.py
boegel: ok, will do, thanks

boegel: debug log now includes something like:

```
== 2015-11-27 10:26:36,023 runpy.PythonPackage DEBUG Only retaining [] subdirs from ['lib', 'lib64', 'lib32'] for $LIBRARY_PATH (others don't provide any libraries)
```

boegel:
Good to go after some further testing, thanks for the review @wpoely86!

PR #745, opened 2015-11-20
Add mkl 11.3.1.150 support

jas02:
Fixes the problem with wrong directory checking. There's no `mkl/bin/intel64` anymore.

PR #691, opened 2015-09-18
Added Intel compiler v2016 support

DirkdeDraak:

PR #756, opened 2015-12-02
update easyblocks for Intel tools to support 2016 versions

wpoely86:
This PR is based on top of #691 and #745. It adds support for specifying which components to install.


easybuild/easyblocks/i/icc.py
```diff
@@ -30,6 +30,7 @@
 @author: Kenneth Hoste (Ghent University)
 @author: Pieter De Baets (Ghent University)
 @author: Jens Timmerman (Ghent University)
+@author: Ward Poelmans (Ghent University)
```
wpoely86: @DirkdeDraak You didn't add yourself to the list of authors. What should I add?

easybuild/easyblocks/i/icc.py
```diff
@@ -58,6 +60,15 @@ class EB_icc(IntelBase):
         - will fail for all older versions (due to newer silent installer)
     """
 
+    def configure_step(self):
+        """have fallback for components for > 2016 versions"""
+
+        if LooseVersion(self.version) >= LooseVersion('2016') and not self.cfg['components']:
+            self.log.warning("No components specified, falling back to ALL")
+            self.cfg.update('components', ['ALL'])
```
boegel: no `DEFAULTS` as fallback?

boegel: or just not specify `components` at all if `components` easyconfig parameter is not defined

easybuild/easyblocks/i/icc.py
```diff
@@ -80,23 +91,24 @@ def sanity_check_step(self):
...
-        libprefix = "lib/intel64/lib"
+        libprefix = "lib/intel64"
         if LooseVersion(self.version) >= LooseVersion("2011"):
             if LooseVersion(self.version) <= LooseVersion("2011.3.174"):
                 binprefix = "bin"
             elif LooseVersion(self.version) >= LooseVersion("2013_sp1"):
                 binprefix = "bin"
-                libprefix = "lib/intel64/lib"
+                if LooseVersion(self.version) >= LooseVersion("2016"):
+                    libprefix = "lib/intel64_lin"
```
boegel: if `intel64` is still available (e.g. as a symlink), stick to that, because they *will* change this again later (but will have to keep `intel64` around)

easybuild/easyblocks/i/icc.py
```diff
@@ -105,53 +117,117 @@ def sanity_check_step(self):
...
-                'MANPATH': ['man', 'share/man', 'man/en_US'],
-                'IDB_HOME': ['bin/intel64']
-            }
+            # 64-bit toolkit
+
+            # using get_software_version('GCC') won't work, while the compiler toolchain is dummy:dummy, which does not
+            # load dependencies.
+            gccversion = get_software_version('GCC')
+            # manual approach to at least have the system version of gcc
+            if not gccversion:
```
boegel: also check if `GCC` is listed as a dependency? you don't want to use the system GCC if GCC is listed as a dep?

easybuild/easyblocks/i/icc.py
```diff
@@ -105,53 +117,117 @@ def sanity_check_step(self):
...
+                (out, _) = run_cmd(cmd, log_all=True, simple=False)
+                ver_re = re.compile("^gcc \(GCC\) (?P<version>[0-9.]+) [0-9]+", re.M)
+                gccversion = ver_re.search(out).group('version')
+                self.log.warning("Using system GCC version %s" % gccversion)
+
+            # TBB directory structure
+            # https://www.threadingbuildingblocks.org/docs/help/tbb_userguide/Linux_OS.htm
+            tbbgccversion = 'gcc4.4'  # gcc version 4.4 or higher that may or may not support exception_ptr
+            if gccversion and LooseVersion(gccversion) >= LooseVersion("4.1") and LooseVersion(gccversion) < LooseVersion("4.4"):
+                tbbgccversion = 'gcc4.1'  # gcc version number between 4.1 and 4.4 that do not support exception_ptr
```
boegel: please flesh out this whole block into a separate function/method (maybe even in the TBB easyblock, where it belongs)?

easybuild/easyblocks/i/icc.py
```diff
@@ -105,53 +117,117 @@ def sanity_check_step(self):
...
+                    'DAALROOT': ['daal'],
+                    'TBBROOT': ['tbb'],
+                    'IPPROOT': ['ipp'],
+                    'CLASSPATH': ['daal/lib/daal.jar']
+                }
+
+        # set debugger path
+        if debuggerpath:
+            if os.path.isdir(os.path.join(self.installdir, debuggerpath, 'python/intel64')):
+                self.cfg['modextravars'] = {'INTEL_PYTHONHOME': os.path.join('$root', debuggerpath, 'python/intel64')}
```
boegel: don't hard modify self.cfg['modextravars'], there may be something there (from the easyconfig file)

better update `dirmap` so that it also specifies `INTEL_PYTHONHOME`

but strictly speaking, this should be handled via `make_module_extra`

easybuild/easyblocks/i/icc.py
```diff
@@ -105,53 +117,117 @@ def sanity_check_step(self):
...
+                dirmap = {
+                    'PATH': [
+                        'mpi/intel64/bin',
+                        'ipp/bin/intel64',
+                        os.path.join(debuggerpath, 'gdb/intel64/bin'),
+                        'bin/intel64',
+                        'bin',
+                    ],
+                    'LD_LIBRARY_PATH': libpaths,
+                    'LIBRARY_PATH': libpaths,
```
boegel: way too much copy-pasting here, this should be refactored like:

```python
dirmap = {
    'PATH': [...],
    ...
}

if LooseVersion(self.version) < LooseVersion("2016"):
    dirmap['PATH'].extend([...])
    ...
else:
    dirmap['PATH'].extend([...])
    ...

dirmap.update({
    'LD_LIBRARY_PATH': libpaths,
    'LIBRARY_PATH': libpaths,
})
```

easybuild/easyblocks/i/ifort.py
```diff
@@ -56,6 +57,8 @@ def sanity_check_step(self):
             elif LooseVersion(self.version) >= LooseVersion("2013_sp1"):
                 binprefix = "bin"
                 libprefix = "lib/intel64/lib"
+                if LooseVersion(self.version) >= LooseVersion("2016"):
+                    libprefix = "lib/intel64_lin/lib"
```
boegel: stick to `intel64` (a symlink to `intel64_lin`), since they may change `intel64_lin` *again* in the future...

easybuild/easyblocks/i/imkl.py
```diff
@@ -357,7 +357,7 @@ def sanity_check_step(self):
             if self.cfg['m32']:
                 raise EasyBuildError("Sanity check for 32-bit not implemented yet for IMKL v%s (>= 10.3)", self.version)
             else:
-                mkldirs = ["bin", "mkl/bin", "mkl/bin/intel64", "mkl/lib/intel64", "mkl/include"]
+                mkldirs = ["bin", "mkl/bin", "mkl/lib/intel64", "mkl/include"]
```
boegel: why remove something? this should be done conditionally on the imkl version?

easybuild/easyblocks/generic/intelbase.py
```diff
@@ -102,10 +105,41 @@ def extra_options(extra_vars=None):
...
         return extra_vars
 
+    def parse_components_list(self):
+        """parse the regex in the components extra_options and select the matching components
+        from the mediaconfig.xml file in the install dir"""
+
+        mediaconfigpath = os.path.join(self.cfg['start_dir'], 'pset', 'mediaconfig.xml')
+        if not os.path.isfile(mediaconfigpath):
+            self.log.warning("Could not find pset/mediaconfig.xml in the build directory. Skipping list of components to install")
+            return
```
boegel: don't return inline please, make it harder to understand the code

easybuild/easyblocks/generic/intelbase.py
```diff
@@ -102,10 +105,41 @@ def extra_options(extra_vars=None):
...
 
         return extra_vars
 
+    def parse_components_list(self):
+        """parse the regex in the components extra_options and select the matching components
+        from the mediaconfig.xml file in the install dir"""
+
+        mediaconfigpath = os.path.join(self.cfg['start_dir'], 'pset', 'mediaconfig.xml')
+        if not os.path.isfile(mediaconfigpath):
+            self.log.warning("Could not find pset/mediaconfig.xml in the build directory. Skipping list of components to install")
```
boegel: if `parse_components_list` is only called in case `components` is specified, make this a `raise`

easybuild/easyblocks/generic/intelbase.py
```diff
@@ -102,10 +105,41 @@ def extra_options(extra_vars=None):
...
+        self.log.debug("Intel components found: %s" % available_components)
+
+        to_install = []
+        self.log.debug("Using regex list: %s" % self.cfg['components'])
+
+        if 'ALL' in self.cfg['components']:
+            to_install = ['ALL']
+        elif 'DEFAULTS' in self.cfg['components']:
+            to_install = ['DEFAULTS']
+        else:
```
boegel: only here we need to actually parse `mediaconfig.xml`, so only do it here?

easybuild/easyblocks/generic/intelbase.py
```diff
@@ -102,10 +105,41 @@ def extra_options(extra_vars=None):
...
+        available_components = re.findall("<Abbr>(?P<component>[^<]+)</Abbr>", mediaconfig, re.M)
+        self.log.debug("Intel components found: %s" % available_components)
+
+        to_install = []
+        self.log.debug("Using regex list: %s" % self.cfg['components'])
+
+        if 'ALL' in self.cfg['components']:
+            to_install = ['ALL']
+        elif 'DEFAULTS' in self.cfg['components']:
+            to_install = ['DEFAULTS']
```
boegel: change to

```python
if len(self.cfg['components']) == 1 and self.cfg['components'][0] in ['ALL', 'DEFAULTS']:
    to_install = self.cfg['components']

File: easybuild/easyblocks/generic/intelbase.py [hunk 14]
```diff
@@ -102,10 +105,41 @@ def extra_options(extra_vars=None):
...
+        else:
+            for comp in available_components:
+                for comp_regex in self.cfg['components']:
+                    if re.search(comp_regex, comp):
+                        to_install.append(comp)
+
+        if to_install:
```
boegel: if `to_install` is empty but `self.cfg['components']` is not, something is very wrong?

easybuild/easyblocks/generic/intelbase.py
```diff
@@ -102,10 +105,41 @@ def extra_options(extra_vars=None):
...
+            self.install_components = to_install
```
boegel: why not define `self.install_components` directly rather than going via an internal variable `to_install`?

easybuild/easyblocks/generic/intelbase.py
```diff
@@ -287,6 +324,13 @@ def install_step(self, silent_cfg_names_map=None, silent_cfg_extras=None):
             'install_dir': silent_cfg_names_map.get('install_dir', self.installdir),
         }
 
+        if self.install_components:
+            if len(self.install_components) == 1:
+                silent += 'COMPONENTS=%s\n' % self.install_components[0]
```
boegel: no need for this separate case?

`';'.join(['ALL']` is equivalent to `'ALL'`?

easybuild/easyblocks/generic/intelbase.py
boegel: even more: if `'ALL' or 'DEFAULTS' is specified next to other stuff, throw an error, they should be used in isolation

easybuild/easyblocks/generic/intelbase.py
wpoely86: it's currently always called but will change

easybuild/easyblocks/generic/intelbase.py
wpoely86: Now I can print the list of components available when using `ALL` in debug mode. Lets keep that.

boegel: ok, sure, but it does mean there's no way around if mediaconfig.xml is not available for some reason...

unless you allow that mediaconfig.xml is not available above, and only complain here if `available_components` is `None`

wpoely86: I'm going to `raise` if `mediaconfig.xml` is not found. It has no point specifying components if it's not there.

boegel:
@wpoely86: please drop in a comment when you're "done" with this, I'll pick it up then to finish it and get it merged

wpoely86:
@boegel all yours. I can install `icc`, `ifort` and `impi`. `imki` fails because I install too few components. Currently I use: `components = ['intel-icc-', 'intel-ccomp-', 'intel-openmp-']`. This is not sufficient as `imkl` fails with: `icc: error while loading shared libraries: libintelremotemon.so`.

Have fun :smiling_imp:

boegel:
Thanks @wpoely86!

easybuild/easyblocks/i/icc.py
```diff
@@ -80,23 +90,23 @@ def sanity_check_step(self):
...
         binfiles = ["icc", "icpc"]
         if LooseVersion(self.version) < LooseVersion("2014"):
             binfiles += ["idb"]
 
         custom_paths = {
-            'files': ["%s/%s" % (binprefix, x) for x in binfiles] +
-                     ["%s%s" % (libprefix, x) for x in ["iomp5.a", "iomp5.so"]],
+            'files': [os.path.join(binprefix, x) for x in binfiles] +
+            [os.path.join(libprefix, 'lib%s' % x) for x in ['iomp5.a', 'iomp5.so']] +
+            ['include/omp.h'],
```
boegel: this looks sensible, but there's a range of icc versions that don't have this, at least not in this location

e.g. `2011.13.367` and `2013.5.192` do *not* have this file there...

so, we should only include this conditionally (e.g. since `2015` and up)

@wpoely86: I'll fix this in a PR to your branch

easybuild/easyblocks/i/icc.py
```diff
@@ -79,87 +95,129 @@ def install_step(self):
...
+        """
+        prefix = None
+
+        # guesses per environment variables
+        # some of these paths only apply to certain versions, but that doesn't really matter
+        # existence of paths is checked by module generator before 'prepend-paths' statements are included
+        guesses = {
+            'CLASSPATH': ['daal/lib/daal.jar'],
+            # 'include' is deliberately listed last, so it gets precedence over the other subdirs
+            'CPATH': ['daal/include', 'ipp/include', 'mkl/include', 'tbb/include', 'include'],
```
boegel: `$CPATH` was not set before, and it seems like we shouldn't start setting it either...

I think this is causing the problem in discussed at https://github.com/hpcugent/easybuild-easyblocks/pull/772/files#r47450112, i.e. error like this popping up:

```
lib/types.h(107): error: qualified name is not allowed                                                                                                                         
    typedef std::complex<double> cdouble;                                                                                                                                      
...                                                                                                                    
lib/types.h(107): error: expected a ";"                                                                                                                                        
    typedef std::complex<double> cdouble;  
```

unless the `include/c++/4.9.3` subdirectory of the GCC install dir is prepended to `$CPATH`

this is also mentioned in https://software.intel.com/en-us/forums/intel-c-compiler/topic/338378, to quote:

```
The problem has been resolved by now. Our system admin, who installed the compiler using the standard Intel installer (only changed the install directory from the default, but nothing else), unnecessarily added the the Intel include paths to the environment variables CPLUS_INCLUDE_PATH, so that #include_next included the same file again (like me, he wasn't aware of the #include_next directive).
```

wpoely86: ok, drop them.

boegel: fixed with https://github.com/wpoely86/easybuild-easyblocks/issues/20

easybuild/easyblocks/i/icc.py
boegel: done, see below

boegel:
I gave this another thorough test using the new easyconfigs for:

* `intel/2015.08` (https://github.com/hpcugent/easybuild-easyconfigs/pull/2194)
* `intel/2016.00` (https://github.com/hpcugent/easybuild-easyconfigs/pull/2209)
* `intel/2016.01` (https://github.com/hpcugent/easybuild-easyconfigs/pull/2219)

and also by reinstalling the full toolchain and rebuilding HPL for `intel/2015a`, `intel/2015b`, `ictce/4.0.6` and also rebuilding `Bison` with a reinstalled `ictce/3.2.2.u3`.

I noticed a small issue with the sanity check of `tbb` failing for 4.3.x and 4.4.x, but I'll issue a follow-up PR to fix that.

Thanks a lot for contributing to this @DirkDeDraak, @jas02 and @wpoely86!

PR #758, opened 2015-12-02
Modified CUDA easyblock to support wrapper creation

damianam:
This PR modifies the CUDA easyblock to support the optional creation of nvcc wrappers. It adds two extra boolean options for easyconfigs: generate_intel_wrapper and generate_gcc_wrapper. The first one will create a wrapper invcc. The second a wrapper called gnvcc.

This allows to simply tell users "Use invcc to have icpc as a host compiler".

This is a convenience fix that doesn't modify the default behaviour or installation.

It will be followed by another PR in the easyconfigs repo, with an example for iccifort+CUDA.

UPDATE: The naming of the wrappers and how they are created has changed significantly. Take a look at the development of the PR.

### DIFF
--- a/easybuild/easyblocks/c/cuda.py
+++ b/easybuild/easyblocks/c/cuda.py
@@ -1,7 +1,7 @@
 ##
 # This file is an EasyBuild reciPY as per https://github.com/hpcugent/easybuild
 #
-# Copyright:: Copyright 2012-2015 Cyprus Institute / CaSToRC, Uni.Lu, NTUA, Ghent University, Forschungszentrum Juelich GmbH
+# Copyright:: Copyright 2012-2016 Cyprus Institute / CaSToRC, Uni.Lu, NTUA, Ghent University, Forschungszentrum Juelich GmbH
 # Authors::   George Tsouloupas <g.tsouloupas@cyi.ac.cy>, Fotis Georgatos <fotis@cern.ch>, Kenneth Hoste, Damian Alvarez
 # License::   MIT/GPL
 # $Id$
--- a/easybuild/easyblocks/c/cuda.py
+++ b/easybuild/easyblocks/c/cuda.py
@@ -1,8 +1,8 @@
 ##
 # This file is an EasyBuild reciPY as per https://github.com/hpcugent/easybuild
 #
-# Copyright:: Copyright 2012-2015 Cyprus Institute / CaSToRC, Uni.Lu, NTUA, Ghent University
-# Authors::   George Tsouloupas <g.tsouloupas@cyi.ac.cy>, Fotis Georgatos <fotis@cern.ch>, Kenneth Hoste
+# Copyright:: Copyright 2012-2015 Cyprus Institute / CaSToRC, Uni.Lu, NTUA, Ghent University, Forschungszentrum Juelich GmbH
+# Authors::   George Tsouloupas <g.tsouloupas@cyi.ac.cy>, Fotis Georgatos <fotis@cern.ch>, Kenneth Hoste, Damian Alvarez
 # License::   MIT/GPL
 # $Id$
 #
@@ -17,21 +17,42 @@ Ref: https://speakerdeck.com/ajdecon/introduction-to-the-cuda-toolkit-for-buildi
 @author: George Tsouloupas (Cyprus Institute)
 @author: Fotis Georgatos (Uni.lu)
 @author: Kenneth Hoste (Ghent University)
+@author: Damian Alvarez (Forschungszentrum Juelich)
 """
 import os
+import stat
+
 from distutils.version import LooseVersion
 
 from easybuild.easyblocks.generic.binary import Binary
-from easybuild.tools.filetools import patch_perl_script_autoflush
+from easybuild.framework.easyconfig import CUSTOM
+from easybuild.tools.filetools import adjust_permissions, patch_perl_script_autoflush, write_file
 from easybuild.tools.run import run_cmd, run_cmd_qa
 from easybuild.tools.systemtools import get_shared_lib_ext
 
+# Wrapper script definition
+wrapper = """#!/bin/sh
+echo "$@" | grep -e '-ccbin' -e '--compiler-bindir' > /dev/null
+if [ $? -eq 0 ];
+then
+        echo "ERROR: do not set -ccbin or --compiler-bindir when using the `basename $0` wrapper"
+else
+        nvcc -ccbin=%s "$@"
+        exit $?
+fi """
 
 class EB_CUDA(Binary):
     """
     Support for installing CUDA.
     """
 
+    @staticmethod
+    def extra_options():
+        extra_vars = {
+            'host_compilers': [[], "Host compilers for which a wrapper will be generated", CUSTOM]
+        }
+        return Binary.extra_options(extra_vars)
+ 
     def extract_step(self):
         """Extract installer to have more control, e.g. options, patching Perl scripts, etc."""
         execpath = self.src[0]['path']
@@ -78,6 +99,16 @@ class EB_CUDA(Binary):
     
         #overriding maxhits default value to 300 (300s wait for nothing to change in the output without seeing a known question)
         run_cmd_qa(cmd, qanda, std_qa=stdqa, no_qa=noqanda, log_all=True, simple=True, maxhits=300)
+
+    def post_install_step(self):
+        def create_wrapper(wrapper_name,wrapper_comp):
+	    wrapper_f = "%s/bin/%s" % (self.installdir, wrapper_name)
+	    write_file(wrapper_f, wrapper % wrapper_comp)
+            adjust_permissions(wrapper_f, stat.S_IXUSR|stat.S_IRUSR|stat.S_IXGRP|stat.S_IRGRP|stat.S_IXOTH|stat.S_IROTH)
+
+	# Prepare wrappers to handle a default host compiler other than g++
+        for comp in self.cfg['host_compilers']:
+            create_wrapper('nvcc_%s' % comp,'%s' % comp)
 
     def sanity_check_step(self):
         """Custom sanity check for CUDA."""
--- a/easybuild/easyblocks/c/cuda.py
+++ b/easybuild/easyblocks/c/cuda.py
@@ -31,7 +31,7 @@ from easybuild.tools.run import run_cmd, run_cmd_qa
 from easybuild.tools.systemtools import get_shared_lib_ext
 
 # Wrapper script definition
-wrapper = """#!/bin/sh
+WRAPPER_TEMPLATE = """#!/bin/sh
 echo "$@" | grep -e '-ccbin' -e '--compiler-bindir' > /dev/null
 if [ $? -eq 0 ];
 then
@@ -48,8 +48,9 @@ class EB_CUDA(Binary):
 
     @staticmethod
     def extra_options():
+        """Create a set of wrappers based on a list determined by the easyconfig file"""
         extra_vars = {
-            'host_compilers': [[], "Host compilers for which a wrapper will be generated", CUSTOM]
+            'host_compilers': [None, "Host compilers for which a wrapper will be generated", CUSTOM]
         }
         return Binary.extra_options(extra_vars)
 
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/c/cuda.py
```diff
@@ -17,8 +17,13 @@
 @author: George Tsouloupas (Cyprus Institute)
 @author: Fotis Georgatos (Uni.lu)
 @author: Kenneth Hoste (Ghent University)
+@author: Damian Alvarez (Forschungszentrum Juelich)
 """
 import os
+import stat
+import shutil
```
boegel: @damianam: we try to keep these alphabetical, please reorder?

easybuild/easyblocks/c/cuda.py
```diff
@@ -17,8 +17,13 @@
...
+
+from easybuild.framework.easyconfig import CUSTOM
```
boegel: same here, order alphabetically in between the other `easybuild` imports below please

easybuild/easyblocks/c/cuda.py
```diff
@@ -31,6 +36,14 @@ class EB_CUDA(Binary):
     Support for installing CUDA.
     """
 
+    @staticmethod
+    def extra_options():
+        extra_vars = {
+            'generate_intel_wrapper': [False, "Generate an nvcc wrapper (called invcc) to enable the usage of the Intel compiler as a host compiler, without explicitely using -ccbin.", CUSTOM],
+            'generate_gcc_wrapper': [False, "Generate an nvcc wrapper (called gnvcc) to enable the usage of the GCC compiler as a host compiler, without explicitely using -ccbin. g++ is the default host compiler. The wrapper is done just for clarity and completion (invcc for Intel and gnvcc for GCC)", CUSTOM],
```
boegel: lines longer than 120 characters should be avoided, can you refactor this to split it across multiple lines (let me know if you need help on this)

easybuild/easyblocks/c/cuda.py
```diff
@@ -78,6 +91,35 @@ def install_step(self):
         #overriding maxhits default value to 300 (300s wait for nothing to change in the output without seeing a known question)
         run_cmd_qa(cmd, qanda, std_qa=stdqa, no_qa=noqanda, log_all=True, simple=True, maxhits=300)
 
+
+    def post_install_step(self):
+        # Wrapper script definition
+        wrapper = """#!/bin/sh
```
boegel: defining the wrapper script as a constant outside of the `EB_CUDA` class would be nicer

easybuild/easyblocks/c/cuda.py
```diff
@@ -78,6 +91,35 @@ def install_step(self):
...
+        echo "ERROR: do not set -ccbin or --compiler-bindir when using the `basename $0` wrapper"
+else
+        nvcc -ccbin=%s "$@"
+        exit $?
+fi """
+
+        def create_wrapper(wrapper_name,wrapper_comp):
+	    wrapper_f = os.fdopen(os.open("%s/bin/%s" % (self.installdir, wrapper_name), os.O_WRONLY | os.O_CREAT , stat.S_IXUSR | stat.S_IRUSR | stat.S_IXGRP | stat.S_IRGRP | stat.S_IXOTH | stat.S_IROTH ),'w')
+	    wrapper_f.write(wrapper % wrapper_comp)
+	    wrapper_f.close()
```
boegel: please use `write_file` + `adjust_permissions`

easybuild/easyblocks/c/cuda.py
```diff
@@ -78,6 +91,35 @@ def install_step(self):
...
+
+	# Prepare wrappers to handle a default host compiler other than g++
+        if self.cfg['generate_intel_wrapper']:
+            wrapper_name = "invcc"
+            wrapper_comp = "icpc"
```
boegel: why define variables?

just do

```python
create_wrapper('invcc' ,'icpc')
```

easybuild/easyblocks/c/cuda.py
```diff
@@ -78,6 +91,35 @@ def install_step(self):
...
+            create_wrapper(wrapper_name,wrapper_comp)
+
+        if self.cfg['generate_gcc_wrapper']:
+            wrapper_name = "gnvcc"
+            wrapper_comp = "g++"
+            create_wrapper(wrapper_name,wrapper_comp)
```
boegel: same here, collapse to a single line

boegel:
Jenkins: ok to test

boegel:
@damianam: thanks for the contribution! Please take a look at the remarks, and push an extra commit to your branch to resolve them, let us know if you need any help (you know where to find me ;-)).

easybuild/easyblocks/c/cuda.py
ocaisa: I think this can be generalised to `generate_wrapper` using the `COMPILER_CUDA_UNIQUE_OPTION_MAP` `from easybuild.tools.toolchain.cuda`

boegel: Why not always include both wrappers (or maybe additional ones in the future, e.g. for PGI)?

So, only define `generate_wrappers` as an extra easyconfig parameter, and go:

```python
if self.cfg['generate_wrappers']:
    wrappers = [
        ('gnvcc', 'g++'),
        ('invcc', 'icpc'),
    ]
    for wrapper_name, wrapper_comp in wrappers:
        create_wrapper(wrapper_name, wrapper_comp)
```

If deemed necessary, you can only define the wrapper that make sense, by checking via the `which` function whether the wrapper compiler is available, and only then defining the matching wrapper.

damianam:
I have applied the suggested changes from this PR and https://github.com/hpcugent/easybuild-easyconfigs/pull/2177. The naming of the host compiler is now provided in the easyconfig, since taking it directly from the framework didn't work (COMPILER_CXX is None, and COMPILER_CUDA_UNIQUE_OPTION_MAP provides an unresolved string 'ccbin="%(CXX_base)s"', since CXX_base is not defined).

It is possible now to create a wrapper and give it a host compiler in the easyconfig. Multiple wrappers in a single installation are not supported in the current PR, since in our scheme we will always have CUDA+iccifort, or CUDA+gcc, or CUDA+pgi. In other words, always paired.

easybuild/easyblocks/c/cuda.py
```diff
@@ -114,3 +148,12 @@ def make_module_req_guess(self):
         })
 
         return guesses
+    
+    def make_module_extra(self):
+        """Specify CUDA_HOME """
+
+        txt = super(EB_CUDA, self).make_module_extra()
+
+        txt += self.module_generator.set_environment('CUDA_HOME', self.installdir)
```
boegel: @damianam: this is already there (see method above)?

damianam: Correct, but for some reason it is not working. We should keep the discussion here, instead of the one in https://github.com/hpcugent/easybuild-easyconfigs/pull/2177 since this is an easyblock issue. In any case, why use guesses instead of setting these PATHs manually to well known directories?

boegel: @damianam: the guesses are checked, in the sense that `eb` will check whether the directory actually exists before including the corresponding statement in the module file.

Additionally, via guesses we'll use `prepend-path` rather than `setenv`.

For `$CUDA_HOME` in particular, I'm not sure if `prepend-path` makes sense at all. Probably not...

damianam: I don't think it does. ```CUDA_PATH``` is not set either in the module. Could it be that ```prepend-path``` doesn't work when passing an empty path in the ```guesses.update```? ```PATH```, ```LD_LIBRARY_PATH``` and ```CPATH``` are defined correctly

boegel: OK, I traced not having `$CUDA_HOME` or `$CUDA_PATH` being defined down to a bug in the framework.
It has been there for a while (March 2014), but luckily only affects CUDA.

I'll get that fixed as a last-minute thing in the framework, but this PR needs a bit more love before it can go in, so it'll have to be post v2.5.0.

boegel: Bug that causes the issue with `$CUDA_HOME` is fixed in https://github.com/hpcugent/easybuild-framework/pull/1519

damianam: @boegel I've made a couple of small changes in the easyblock and easyconfig, so now you can specify a list of host compilers, and the easyblock will create the appropriate wrappers based on that.

boegel:
@damianam: you'll need to sync this with current `develop`, GitHub has detected conflicts and so this can't be merged in as is; let me know if you need help with that

damianam:
@boegel done. I think :-P

boegel:
@damianam: hmm, not sure what you did there... take a look at the diff now :-/

damianam:
@boegel I rebased my repo to match the new main develop branch, but it seems I did something wrong :-|. I'll try to fix it before the user meeting.

damianam:
@boegel I think it is fixed now. I'll do the same with https://github.com/hpcugent/easybuild-easyconfigs/pull/2177 after lunch

fgeorgatos:
I Could help this a bit during EB meeting, just ask. F.

On Tuesday, January 26, 2016, HPC Ugent build bot <notifications@github.com>
wrote:

> Easyblocks unit test suite *PASSed* (see
> https://jenkins1.ugent.be/job/easybuild-easyblocks-pr-builder/1638/console
> for more details).
>
> This pull request is now ready for review/testing.
>
> Please try and find someone who can tackle this; contact @boegel
> <https://github.com/boegel> if you're not sure what to do.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/hpcugent/easybuild-easyblocks/pull/758#issuecomment-174958776>
> .
>


-- 


echo "sysadmin know better bash than english"|sed s/min/mins/ \
  | sed 's/better bash/bash better/' # signal detected in a CERN forum

easybuild/easyblocks/c/cuda.py
```diff
@@ -79,6 +100,16 @@ def install_step(self):
         #overriding maxhits default value to 300 (300s wait for nothing to change in the output without seeing a known question)
         run_cmd_qa(cmd, qanda, std_qa=stdqa, no_qa=noqanda, log_all=True, simple=True, maxhits=300)
 
+    def post_install_step(self):
+        def create_wrapper(wrapper_name,wrapper_comp):
+	    wrapper_f = "%s/bin/%s" % (self.installdir, wrapper_name)
+	    write_file(wrapper_f, wrapper % wrapper_comp)
+            adjust_permissions(wrapper_f, stat.S_IXUSR|stat.S_IRUSR|stat.S_IXGRP|stat.S_IRGRP|stat.S_IXOTH|stat.S_IROTH)
```
boegel: wrong indentation (tabs involved)?

easybuild/easyblocks/c/cuda.py
```diff
@@ -79,6 +100,16 @@ def install_step(self):
         #overriding maxhits default value to 300 (300s wait for nothing to change in the output without seeing a known question)
         run_cmd_qa(cmd, qanda, std_qa=stdqa, no_qa=noqanda, log_all=True, simple=True, maxhits=300)
 
+    def post_install_step(self):
```
boegel: please add a docstring

easybuild/easyblocks/c/cuda.py
```diff
@@ -79,6 +100,16 @@ def install_step(self):
...
+        def create_wrapper(wrapper_name,wrapper_comp):
```
boegel: also a docstring for internal functions (an add space after `,`)

easybuild/easyblocks/c/cuda.py
```diff
@@ -79,6 +100,16 @@ def install_step(self):
...
+	    wrapper_f = "%s/bin/%s" % (self.installdir, wrapper_name)
```
boegel: using `os.path.join` is more pythonic:

```python
wrapper_f = os.path.join(self.installdir, 'bin', wrapper_name)
```

easybuild/easyblocks/c/cuda.py
```diff
@@ -17,21 +17,42 @@
...
 
 from easybuild.easyblocks.generic.binary import Binary
-from easybuild.tools.filetools import patch_perl_script_autoflush
+from easybuild.framework.easyconfig import CUSTOM
+from easybuild.tools.filetools import adjust_permissions, patch_perl_script_autoflush, write_file
 from easybuild.tools.run import run_cmd, run_cmd_qa
 from easybuild.tools.systemtools import get_shared_lib_ext
 
+# Wrapper script definition
+wrapper = """#!/bin/sh
```
boegel: constants should have uppercase names, and make it clear this is a template:

```python
WRAPPER_TEMPLATE = """#!/bin/sh
...
```

easybuild/easyblocks/c/cuda.py
```diff
@@ -17,21 +17,42 @@
...
+        exit $?
+fi """
 
 class EB_CUDA(Binary):
     """
     Support for installing CUDA.
     """
 
+    @staticmethod
+    def extra_options():
```
boegel: docstring :)

easybuild/easyblocks/c/cuda.py
```diff
@@ -17,21 +17,42 @@
...
+        extra_vars = {
+            'host_compilers': [[], "Host compilers for which a wrapper will be generated", CUSTOM]
```
boegel: it's a good idea to avoid using a list as a default value

better use `None`, but you'll have to take it into account where you're using `self.cfg['host_compilers']`, for example with

```python
for comp in (self.cfg['host_compilers'] or []):
```

easybuild/easyblocks/c/cuda.py
```diff
@@ -79,6 +100,16 @@ def install_step(self):
...
 
+    def post_install_step(self):
+        def create_wrapper(wrapper_name,wrapper_comp):
+	    wrapper_f = "%s/bin/%s" % (self.installdir, wrapper_name)
+	    write_file(wrapper_f, wrapper % wrapper_comp)
+            adjust_permissions(wrapper_f, stat.S_IXUSR|stat.S_IRUSR|stat.S_IXGRP|stat.S_IRGRP|stat.S_IXOTH|stat.S_IROTH)
+
+	# Prepare wrappers to handle a default host compiler other than g++
+        for comp in self.cfg['host_compilers']:
+            create_wrapper('nvcc_%s' % comp,'%s' % comp)
```
boegel: the `'%s' % comp` can be replaced with just `comp` (+ space after `,`)

easybuild/easyblocks/c/cuda.py
```diff
@@ -116,3 +147,12 @@ def make_module_req_guess(self):
         })
 
         return guesses
+    
+    def make_module_extra(self):
+        """Specify CUDA_HOME """
```
boegel: this is not needed anymore, `$CUDA_HOME` will now be defined as it should thanks to the bug fix in https://github.com/hpcugent/easybuild-framework/pull/1519

so, drop this whole method

damianam:
@boegel I fixed the style, lack of docstrings, etc.

damianam:
ping

boegel:
@damianam: thanks for the ping, need to fix conflict with current `develop` first, cfr. https://github.com/damianam/easybuild-easyblocks/pull/2

damianam:
@boegel done

boegel:
Tested with all existing CUDA easyconfigs and the new one from https://github.com/hpcugent/easybuild-easyconfigs/pull/2177, good to go, thanks @damianam!

PR #759, opened 2015-12-03
update IPP easyblock for v9.x (REVIEW)

jas02:
Re-written from scratch, need to be merged with current code.

### DIFF
--- a/easybuild/easyblocks/generic/intelbase.py
+++ b/easybuild/easyblocks/generic/intelbase.py
@@ -31,6 +31,7 @@ Generic EasyBuild support for installing Intel tools, implemented as an easybloc
 @author: Pieter De Baets (Ghent University)
 @author: Jens Timmerman (Ghent University)
 @author: Ward Poelmans (Ghent University)
+@author: Lumir Jasiok (IT4Innovations)
 """
 
 import os
@@ -69,12 +70,12 @@ ACTIVATION_NAME_2012 = 'ACTIVATION'  # previous activation type parameter used i
 # silent.cfg parameter name for install prefix
 INSTALL_DIR_NAME = 'PSET_INSTALL_DIR'
 # silent.cfg parameter name for install mode
-INSTALL_MODE_NAME  = 'PSET_MODE'
+INSTALL_MODE_NAME = 'PSET_MODE'
 # Older (2015 and previous) silent.cfg parameter name for install mode
-INSTALL_MODE_NAME_2015  = 'INSTALL_MODE'
+INSTALL_MODE_NAME_2015 = 'INSTALL_MODE'
 # Install mode for 2016 version
 INSTALL_MODE = 'install'
-# Install mode for 2015 and oder versions
+# Install mode for 2015 and older versions
 INSTALL_MODE_2015 = 'NONRPM'
 # silent.cfg parameter name for license file/server specification
 LICENSE_FILE_NAME = 'ACTIVATION_LICENSE_FILE'  # since icc/ifort v2013_sp1, impi v4.1.1, imkl v11.1
@@ -281,9 +282,9 @@ class IntelBase(EasyBlock):
         silent = '\n'.join([
             "%(activation_name)s=%(activation)s",
             lic_file_entry,
-            "%(install_dir_name)s=%(insstall_dir)s",
+            "%(install_dir_name)s=%(install_dir)s",
             "ACCEPT_EULA=accept",
-            "%(install_mode_name)s=%(install_mode)",
+            "%(install_mode_name)s=%(install_mode)s",
             "CONTINUE_WITH_OPTIONAL_ERROR=yes",
             ""  # Add a newline at the end, so we can easily append if needed
         ]) % {
@@ -293,8 +294,8 @@ class IntelBase(EasyBlock):
             'activation': self.cfg['license_activation'],
             'license_file': self.license_file,
             'install_dir': silent_cfg_names_map.get('install_dir', self.installdir),
-            'install_mode_name': silent_cfg_names_map('install_mode_name', INSTALL_MODE_NAME),
-            'install_mode': silent_cfg_names_map('install_mode', INSTALL_MODE),
+            'install_mode': silent_cfg_names_map.get('install_mode', INSTALL_MODE),
+            'install_mode_name': silent_cfg_names_map.get('install_mode_name', INSTALL_MODE_NAME),
         }
 
         if silent_cfg_extras is not None:
--- a/easybuild/easyblocks/generic/intelbase.py
+++ b/easybuild/easyblocks/generic/intelbase.py
@@ -68,6 +68,14 @@ ACTIVATION_NAME = 'ACTIVATION_TYPE'  # since icc/ifort v2013_sp1, impi v4.1.1, i
 ACTIVATION_NAME_2012 = 'ACTIVATION'  # previous activation type parameter used in older versions
 # silent.cfg parameter name for install prefix
 INSTALL_DIR_NAME = 'PSET_INSTALL_DIR'
+# silent.cfg parameter name for install mode
+INSTALL_MODE_NAME  = 'PSET_MODE'
+# Older (2015 and previous) silent.cfg parameter name for install mode
+INSTALL_MODE_NAME_2015  = 'INSTALL_MODE'
+# Install mode for 2016 version
+INSTALL_MODE = 'install'
+# Install mode for 2015 and oder versions
+INSTALL_MODE_2015 = 'NONRPM'
 # silent.cfg parameter name for license file/server specification
 LICENSE_FILE_NAME = 'ACTIVATION_LICENSE_FILE'  # since icc/ifort v2013_sp1, impi v4.1.1, imkl v11.1
 LICENSE_FILE_NAME_2012 = 'PSET_LICENSE_FILE'  # previous license file parameter used in older versions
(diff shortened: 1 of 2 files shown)


boegel:
Jenkins: ok to test

boegel:
@jas02: since this is basically a rewrite, it needs some cleaning up in order to avoid backwards incompatible changes

easybuild/easyblocks/i/ipp.py
```diff
@@ -1,77 +1,82 @@
...
+CONTINUE_WITH_OPTIONAL_ERROR=yes
+PSET_INSTALL_DIR=%s
+CONTINUE_WITH_INSTALLDIR_OVERWRITE=yes
+COMPONENTS=ALL
+PSET_MODE=install
+ACTIVATION_LICENSE_FILE=%s
+ACTIVATION_TYPE=license_file
+PHONEHOME_SEND_USAGE_DATA=no
+SIGNING_ENABLED=yes
+ARCH_SELECTED=%s
```
wpoely86: I don't see why this cannot be handled in the `intelbase` block? It's all generic.

boegel: We probably can, and some of it already is (and more of it will be with #756).

jas02:
Hope that with current commit 01e600d3c67492d2fdedb81a2a5bcf9f1d1e77d4 won't be break backward compatibility. Installation seems to be OK.

jas02:
In case that you will be OK with this code, I'll rewrite other part of Intel Parallel Studio in similar fashion.

easybuild/easyblocks/generic/intelbase.py
```diff
@@ -285,6 +294,8 @@ def install_step(self, silent_cfg_names_map=None, silent_cfg_extras=None):
             'activation': self.cfg['license_activation'],
             'license_file': self.license_file,
             'install_dir': silent_cfg_names_map.get('install_dir', self.installdir),
+            'install_mode': silent_cfg_names_map.get('install_mode', INSTALL_MODE),
+            'install_mode_name': silent_cfg_names_map.get('install_mode_name', INSTALL_MODE_NAME),
```
boegel: this is a backwards incompatible changes, since you're defaulting to the new name here

This implies that all other easyblocks deriving from have to be updated accordingly...

easybuild/easyblocks/i/ipp.py
```diff
@@ -30,10 +30,14 @@
...
+@author: Lumir Jasiok (IT4Innovations)
 """
-from distutils.version import LooseVersion
 
-from easybuild.easyblocks.generic.intelbase import IntelBase, ACTIVATION_NAME_2012, LICENSE_FILE_NAME_2012
+import platform
+
+
+from distutils.version import LooseVersion
+from easybuild.easyblocks.generic.intelbase import IntelBase, ACTIVATION_NAME_2012, LICENSE_FILE_NAME_2012, INSTALL_MODE_NAME_2015, INSTALL_MODE_2015
```
boegel: line too long

boegel: you're not using the `2012` constants?

easybuild/easyblocks/i/ipp.py
```diff
@@ -62,16 +84,48 @@ def sanity_check_step(self):
...
         super(EB_ipp, self).sanity_check_step(custom_paths=custom_paths)
+
+    def make_module_req_guess(self):
+        """
+        A dictionary of possible directories to look for
+        """
+        guesses = super(EB_ipp, self).make_module_req_guess()
+
+        if LooseVersion(self.version) > LooseVersion('9.0'):
+            lib_path = 'lib/%s' % self.arch
```
boegel: use `os.path.join` please

easybuild/easyblocks/i/ipp.py
```diff
@@ -62,16 +84,48 @@ def sanity_check_step(self):
...
+                            'dirs': dirs
+                           }
+        else:
+            custom_paths = {
+                            'files': ["ipp/lib/intel64/libipp%s" % y
+                                       for x in ["ac", "cc", "ch", "core", "cv", "dc", "di",
+                                                 "i", "j", "m", "r", "s", "sc", "vc", "vm"]
+                                       for y in ["%s.a" % x, "%s.so" % x]],
+                            'dirs': dirs
+               }
```
boegel: fix indent (indent by 4 spaces, and align parts of a broken line)
I also tend to use single quotes for `words' (i.e. any strings without spaces), I feel it's easier to read

```python
if LooseVersion(self.version) > LooseVersion('9.0'):
    custom_paths = {
        'files': ['ipp/lib/intel64/libipp%s' % y
                  for x in ['cc', 'ch', 'core', 'cv', 'dc', 'i', 's', 'vm']
                  for y in ['%s.a' % x, '%s.so' % x]],
        'dirs': dirs,
    }
else:
    custom_paths = {
        'files': ['ipp/lib/intel64/libipp%s' % y
                   for x in ['ac', 'cc', 'ch', 'core', 'cv', 'dc', 'di', 'i', 'j', 'm', 'r', 's', 'sc', 'vc', 'vm']
                   for y in ['%s.a' % x, '%s.so' % x]],
        'dirs': dirs,
    }
```

easybuild/easyblocks/i/ipp.py
```diff
@@ -62,16 +84,48 @@ def sanity_check_step(self):
         if LooseVersion(self.version) < LooseVersion('8.0'):
             dirs = ["compiler/lib/intel64", "ipp/bin", "ipp/include",
                     "ipp/interfaces/data-compression", "ipp/tools/intel64"]
+        elif LooseVersion(self.version) > LooseVersion('9.0'):
+            dirs = ["ipp/bin", "ipp/include",
+                    "ipp/tools/intel64"]
```
boegel: single line, it's short enough?

boegel: also, use `>=` here?

easybuild/easyblocks/generic/intelbase.py
boegel: this change doesn't seem to be needed for the 2016 versions of the compilers, which is... weird

boegel: cc @wpoely86

jas02:
@boegel @wpoely86 I figured it out. Problem with installation of IPP 2016 is that there MUST be ARCH_SELECTED=INTEL64 parameter in silent.cfg. Installation crash without this parameter.

boegel:
@jas02: so, this means that most of the changes to `IntelBase` are not required after all? can you clean that up, to only the required `ARCH_SELECTED=INTEL64` is added for IPP?

jas02:
@boegel Don't you think that changes are well prepared for this release and probably for next releases? I don't know right know about other parts of Parallel Studio. Maybe there is tool that accepts just `install` parameter. I would prefer to keep it and test other components of Studio. I am sure that there is reason for change variable name from INSTALL_* to PSET_*. Maybe INSTALL_* will be deprecated in future releases,

boegel:
@jas02: please review/merge https://github.com/jas02/easybuild-easyblocks/pull/1 to update this PR, it currently breaks older IPP versions (7.x for example)

easybuild/easyblocks/i/ipp.py
```diff
@@ -30,10 +30,16 @@
 @author: Kenneth Hoste (Ghent University)
 @author: Pieter De Baets (Ghent University)
 @author: Jens Timmerman (Ghent University)
+@author: Lumir Jasiok (IT4Innovations)
 """
+
 from distutils.version import LooseVersion
+import os
 
+from easybuild.easyblocks.generic.intelbase import INSTALL_MODE_NAME_2015, INSTALL_MODE_2015
```
boegel: @jas02: no longer used, please remove this line?

easybuild/easyblocks/i/ipp.py
```diff
@@ -46,7 +51,17 @@ def install_step(self):
         - create silent cfg file
         - execute command
         """
+
+        platform_name = get_platform_name()
+        if platform_name.startswith('x86_64'):
+            self.arch = "intel64"
+        elif platform_name.startswith('i386'):
```
wpoely86: don't forget `i686`. Most will give your `i686` these days.

boegel:
Tested with existing IPP easyconfigs + https://github.com/hpcugent/easybuild-easyconfigs/pull/2202 (see https://github.com/hpcugent/easybuild-easyconfigs/pull/2202#issuecomment-163549735), good to go, thanks @jas02!

easybuild/easyblocks/i/ipp.py
```diff
@@ -54,24 +69,52 @@ def install_step(self):
...
+            ipp_libs.extend(['ac', 'di', 'j', 'm', 'r', 'sc', 'vc'])
 
         custom_paths = {
-                        'files': ["ipp/lib/intel64/libipp%s" % y
-                                   for x in ["ac", "cc", "ch", "core", "cv", "dc", "di",
-                                             "i", "j", "m", "r", "s", "sc", "vc", "vm"]
-                                   for y in ["%s.a" % x, "%s.so" % x]],
-                        'dirs': dirs
-                       }
+            'files': ['ipp/lib/intel64/libipp%s' % y for x in ipp_libs for y in ['%s.a' % x, '%s.so' % x]],
```
wpoely86: I would swap the two `for` to increase readability?

easybuild/easyblocks/i/ipp.py
```diff
@@ -54,24 +69,52 @@ def install_step(self):
...
+        A dictionary of possible directories to look for
+        """
+        guesses = super(EB_ipp, self).make_module_req_guess()
+
+        if LooseVersion(self.version) >= LooseVersion('9.0'):
+            lib_path = os.path.join('lib', self.arch)
+            include_path = 'ipp/include'
+
+            guesses.update({
+                'LD_LIBRARY_PATH': [lib_path],
```
wpoely86: shouldn't we append instead of overwrite?

boegel: good point, but it doesn't matter here

* `lib` is not useful, `lib/intel64` is (which is `lib_path` here)
* `include` doesn't exist, `ipp/include` does

so, the paths in the original `guesses` are useless/wrong

easybuild/easyblocks/i/ipp.py
boegel: ok, PR pls? :)

PR #761, opened 2015-12-07
[bugfix] fix numpy sanity check command + cleaner output under -x

boegel:
We need to use the full path to `multiarray.so` to avoid problems with the sanity check.

### DIFF
--- a/easybuild/easyblocks/n/numpy.py
+++ b/easybuild/easyblocks/n/numpy.py
@@ -231,6 +231,8 @@ class EB_numpy(FortranPythonPackage):
             elif self.dry_run:
                 # use fake value during dry run
                 time_msec = 123
+                self.log.warning("Using fake value for time required for %dx%d matrix dot product under dry run: %s",
+                                 size, size, time_msec)
             else:
                 raise EasyBuildError("Failed to determine time for numpy.dot test run.")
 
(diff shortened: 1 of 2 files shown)


easybuild/easyblocks/n/numpy.py
```diff
@@ -228,7 +228,10 @@ def test_step(self):
             res = sec_re.search(out)
             if res:
                 time_msec = 1000 * float(res.group('time'))
-            elif not self.dry_run:
+            elif self.dry_run:
+                # use fake value during dry run
```
wpoely86: warning about fake value?

boegel: sort of useless under dry run, but sure, why not

easybuild/easyblocks/n/numpy.py
```diff
@@ -256,8 +259,8 @@ def sanity_check_step(self, *args, **kwargs):
         ]
         if LooseVersion(self.version) >= LooseVersion("1.10"):
             # if blas_dot symbol is not there, numpy isn't properly linked against a BLAS library
-            multiarray_so = os.path.join(self.pylibdir, 'numpy', 'core', 'multiarray.so')
-            custom_commands.append("nm %s | grep blas_dot" % multiarray_so)
+            multiarray_so = os.path.join(self.installdir, self.pylibdir, 'numpy', 'core', 'multiarray.so')
+            custom_commands.append("nm %s | grep blas_dot" % self.installdir(multiarray_so)
```
wpoely86: why do you still need `installdir` here?

boegel: I was testing you. ;-)

No, typo, this is actually a syntax error...

wpoely86: I was wondering what this was supposed to do. Use `synaptics` or such?

boegel: I have PyLint set up, but wasn't paying close enough attention, my apologies. I was going to test this before merging it though. :-)

boegel:

PR #757, opened 2015-12-02
fix sanity check to support numpy 1.10 (dropped dotblas support)

gppezzi:
sanity check could be improved for v >= 1.10, but with this the build should work.

PR #762, opened 2015-12-07
use generic way of checking BLAS linkage for recent numpy versions

boegel:
implement generic way of checking BLAS linkage in numpy

check for `blas_dot` symbol (cfr. #757) doesn't always work, apparently (seems like it's not there when optimising more aggressively (-O3 vs -O2), or whne `-fPIC` is used)

cfr. https://github.com/numpy/numpy/issues/6675#issuecomment-162601149

cc @gppezzi


boegel:
@wpoely86: please review?

wpoely86:
looks good

boegel:
Tested by rebuilding with numpy 1.9 and 1.10 + Python easyconfigs, good to go, thanks for the review @wpoely86!

PR #765, opened 2015-12-10
Libxml2 when not available from OS

ocaisa:
This only worked initially because setup.py found the libxml headers from the system. When these are not available the installation fails.

Also need to explicitly tell the configure not to build the python bindings itself (otherwise it will try to install them as an extension to the python it finds and you might not have permission there...and it's also not what you want).

### DIFF
--- a/easybuild/easyblocks/l/libxml2.py
+++ b/easybuild/easyblocks/l/libxml2.py
@@ -27,7 +27,7 @@ EasyBuild support for building and installing libxml2 with python bindings,
 implemented as an easyblock.
 
 @author: Jens Timmerman (Ghent University)
-@author: Alan O'Cais (Juelich Supercomputing Centre)
+@author: Alan O'Cais (Juelich Supercomputing 'entre)
 """
 import os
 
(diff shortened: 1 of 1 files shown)


ocaisa:
@JensTimmerman You might want to  review this

JensTimmerman:
jenkins: ok to test

JensTimmerman:
Looks fine to me, I have no idea why this worked for me then, since I don't have permission to install things in the system python.

ocaisa:
I ran into a problem because someone else (actually me by another name) had installed the Python dep. It tried to install the modules there by default, I only got around it by explicitly switching the bindings off.

boegel:
@ocaisa: should the sanity check be enhanced to catch this problem, i.e. to check that the Python bindings are available from the `libxml2` install prefix?

ocaisa:
Done, also checked that the expected happens when someone manually adds `--with-python` on the command line and `--without-python` added in the block takes precedence (as expected)

boegel:
Thoroughly tested with existing `libxml2` easyconfigs, works like a charm, thanks @ocaisa!

PR #744, opened 2015-11-19
OpenFOAM and scotch 64 bit easyblocks (REVIEW)

hajgato:

PR #770, opened 2015-12-11
Openfoam extend 3.2 update

wpoely86:
Add support for openfoam-extend 3.2

### DIFF
--- a/easybuild/easyblocks/o/openfoam.py
+++ b/easybuild/easyblocks/o/openfoam.py
@@ -85,7 +85,7 @@ class EB_OpenFOAM(EasyBlock):
             try:
                 contents_installdir = os.listdir(self.installdir)
                 # it's one directory but has a wrong name
-                if len(contents_installdir) == 1 and os.path.isdir(contents_installdir[0]):
+                if len(contents_installdir) == 1 and os.path.isdir(os.path.join(self.installdir, contents_installdir[0])):
                     source = os.path.join(self.installdir, contents_installdir[0])
                     target = os.path.join(self.installdir, self.openfoamdir)
                     self.log.debug("Moving %s to %s" % (source, target))
@@ -142,7 +142,7 @@ class EB_OpenFOAM(EasyBlock):
         for script in [os.path.join(self.builddir, self.openfoamdir, x) for x in ['etc/bashrc', 'etc/cshrc']]:
             self.log.debug("Patching out hardcoded $WM_* env vars in %s", script)
             # disable any third party stuff, we use EB controlled builds
-            regex_subs = [(r"^setenv|export) WM_THIRD_PARTY_USE_.*[ =].*$", r"# \g<0>")]
+            regex_subs = [(r"^(setenv|export) WM_THIRD_PARTY_USE_.*[ =].*$", r"# \g<0>")]
             for env_var in ['WM_COMPILER', 'WM_MPLIB', 'WM_THIRD_PARTY_DIR']:
                 regex_subs.append((r"^(setenv|export) (?P<var>%s)[ =](?P<val>.*)$" % env_var,
                                    r": ${\g<var>:=\g<val>}; export \g<var>"))
@@ -307,8 +307,9 @@ class EB_OpenFOAM(EasyBlock):
                                                     "refineMesh", "vorticity"]]
         # check for the Pstream and scotchDecomp libraries, there must be a dummy one and an mpi one
         if 'extend' in self.name.lower():
-            libs = [os.path.join(libsdir, x, "libPstream.so") for x in ["dummy", "mpi"]] + \
-                   [os.path.join(libsdir, "libscotchDecomp.so")]
+            libs = [os.path.join(libsdir, "libscotchDecomp.so")]
+            if LooseVersion(self.version) < LooseVersion('3.2'):
+                libs.extend([os.path.join(libsdir, x, "libPstream.so") for x in ["dummy", "mpi"]])
         else:
             libs = [os.path.join(libsdir, x, "libPstream.so") for x in ["dummy", "mpi"]] + \
                    [os.path.join(libsdir, x, "libptscotchDecomp.so") for x in ["dummy", "mpi"]] +\
--- a/easybuild/easyblocks/o/openfoam.py
+++ b/easybuild/easyblocks/o/openfoam.py
@@ -213,19 +213,17 @@ class EB_OpenFOAM(EasyBlock):
         if LooseVersion(self.version) < LooseVersion("2") or openfoam_extend_v3:
             self.log.debug("List of deps: %s" % self.cfg.dependencies())
             for dep in self.cfg.dependencies():
-                cur_dep = {
-                    'name': dep['name'].upper(),
-                    'root': get_software_root(dep['name'])
-                }
+                dep_name = dep['name'].upper(),
+                dep_root = get_software_root(dep['name'])
                 dep_vars = [
-                    ("%(name)s_SYSTEM", "1"),
-                    ("%(name)s_DIR", "%(root)s"),
-                    ("%(name)s_BIN_DIR", "%(root)s/bin"),
-                    ("%(name)s_LIB_DIR", "%(root)s/lib"),
-                    ("%(name)s_INCLUDE_DIR", "%(root)s/include"),
+                    ("%s_SYSTEM", "1"),
+                    ("%s_DIR", "%s"),
+                    ("%s_BIN_DIR", "%s/bin"),
+                    ("%s_LIB_DIR", "%s/lib"),
+                    ("%s_INCLUDE_DIR", "%s/include"),
                 ]
                 for var, val in dep_vars:
-                    env.setvar(var % cur_dep, val % cur_dep)
+                    env.setvar(var % dep_name, val % dep_root)
         else:
             for depend in ['SCOTCH', 'METIS', 'CGAL', 'Paraview']:
                 dependloc = get_software_root(depend)
--- a/easybuild/easyblocks/o/openfoam.py
+++ b/easybuild/easyblocks/o/openfoam.py
@@ -233,7 +233,7 @@ class EB_OpenFOAM(EasyBlock):
                     "%s_LIB_DIR": "%s/lib",
                     "%s_INCLUDE_DIR": "%s/include",
                 }
-                for var, val in dep_vars:
+                for var, val in dep_vars.iteritems():
                     env.setvar(var % dep_name, val % dep_root)
         else:
             for depend in ['SCOTCH', 'METIS', 'CGAL', 'Paraview']:
--- a/easybuild/easyblocks/o/openfoam.py
+++ b/easybuild/easyblocks/o/openfoam.py
@@ -31,21 +31,21 @@ EasyBuild support for building and installing OpenFOAM, implemented as an easybl
 @author: Pieter De Baets (Ghent University)
 @author: Jens Timmerman (Ghent University)
 @author: Xavier Besseron (University of Luxembourg)
+@author: Ward Poelmans (Ghent University)
 """
-import fileinput
+
 import glob
 import os
 import re
 import shutil
 import stat
-import sys
 from distutils.version import LooseVersion
 
 import easybuild.tools.environment as env
 import easybuild.tools.toolchain as toolchain
 from easybuild.framework.easyblock import EasyBlock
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/o/openfoam.py
```diff
@@ -282,15 +307,16 @@ def sanity_check_step(self):
                                                     "refineMesh", "vorticity"]]
         # check for the Pstream and scotchDecomp libraries, there must be a dummy one and an mpi one
         if 'extend' in self.name.lower():
-            libs = [os.path.join(libsdir, x, "libPstream.so") for x in ["dummy", "mpi"]] + \
-                   [os.path.join(libsdir, "libscotchDecomp.so")]
+            libs = [os.path.join(libsdir, "libscotchDecomp.so")]
+            if LooseVersion(self.version) < LooseVersion('3.2'):
+                libs.extend([os.path.join(libsdir, x, "libPstream.so") for x in ["dummy", "mpi"]])
```
boegel: hold on, did this actually disappear, or are you ignored something that went wrong during the build?

references?

easybuild/easyblocks/o/openfoam.py
```diff
@@ -197,19 +213,28 @@ def configure_step(self):
...
+                }
+                dep_vars = [
+                    ("%(name)s_SYSTEM", "1"),
+                    ("%(name)s_DIR", "%(root)s"),
+                    ("%(name)s_BIN_DIR", "%(root)s/bin"),
+                    ("%(name)s_LIB_DIR", "%(root)s/lib"),
+                    ("%(name)s_INCLUDE_DIR", "%(root)s/include"),
+                ]
+                for var, val in dep_vars:
+                    env.setvar(var % cur_dep, val % cur_dep)
```
boegel: this is a bit weird, since you never need `%(name)s` in `val`, or `%(root)s` in `var`, so don't use `cur_dep`?

easybuild/easyblocks/o/openfoam.py
wpoely86: I cannot find a reference to it anymore in any `Makefile`. Test build works.

easybuild/easyblocks/o/openfoam.py
```diff
@@ -83,24 +83,38 @@ def extract_step(self):
...
+                            shutil.move(source, target)
+                    os.chdir(openfoam_installdir)
             except OSError, err:
                 raise EasyBuildError("Failed to move all files to %s: %s", openfoam_installdir, err)
 
+    def patch_step(self):
+        """Adjust beginpath and start_dir to correct openfoam dir"""
+        beginpath = os.path.join(self.installdir, self.openfoamdir)
+        self.cfg['start_dir'] = beginpath
+        super(EB_OpenFOAM, self).patch_step(beginpath=beginpath)
```
boegel: why is this required? are you sure this is not going to break existing OpenFOAM easyconfigs?

testing will tell...

easybuild/easyblocks/o/openfoam.py
```diff
@@ -83,24 +83,38 @@ def extract_step(self):
...
+                    os.rename(source, target)
+                else:
+                    mkdir(openfoam_installdir)
+                    for fil in contents_installdir:
+                        if fil != self.openfoamdir:
+                            source = os.path.join(self.installdir, fil)
+                            target = os.path.join(openfoam_installdir, fil)
+                            self.log.debug("Moving %s to %s" % (source, target))
+                            shutil.move(source, target)
+                    os.chdir(openfoam_installdir)
```
boegel: wrap this in `try:`, `except OSError as err:`

easybuild/easyblocks/o/openfoam.py
wpoely86: because the original unpacked dir is renamed

easybuild/easyblocks/o/openfoam.py
wpoely86: see one line below?

boegel:
@wpoely86: any idea how this will blend with https://github.com/hpcugent/easybuild-easyblocks/pull/744 ?

easybuild/easyblocks/o/openfoam.py
wpoely86: if it makes you happy

easybuild/easyblocks/o/openfoam.py
boegel: does using it also work?

easybuild/easyblocks/o/openfoam.py
boegel: it's just too confusing as is imho

wpoely86:
It will clash, hence my request to merge it first. I will merge develop in here then

easybuild/easyblocks/o/openfoam.py
wpoely86: Haven't run a complete test suite but as far as I can tell, it works

boegel:
@wpoely86: there are a couple of minor style remarks there, but it works, so should go in soon

boegel:
@wpoely86: please rebase on #744?

I'd like to use this to test https://github.com/hpcugent/easybuild-easyconfigs/pull/2048, https://github.com/hpcugent/easybuild-easyconfigs/pull/2023, https://github.com/hpcugent/easybuild-easyconfigs/pull/2216, https://github.com/hpcugent/easybuild-easyconfigs/pull/2142

easybuild/easyblocks/o/openfoam.py
boegel: I can't find anything on Pstream disappearing, and since this check is relatively recent (and not having the libPstream libraries went undetected for a while), cfr. https://github.com/hpcugent/easybuild-easyblocks/pull/436/files#diff-0edc74d66adc49c07871a687be0ef167R212, I'd like to get confirmation of this somehow...

I don't see any mention of Pstream in the OpenFOAM-Extend 3.2 release notes (http://sourceforge.net/p/openfoam-extend/foam-extend-3.1/ci/v3.2/tree/ReleaseNotes.txt), so it's fishy to me that this just disappeared...

Maybe @besserox has an idea?

besserox: The Pstream library has apparently been removed in this commit:
http://sourceforge.net/p/foam-extend/foam-extend-3.2/ci/4c23c3f9e812674bf54b6a26e58f32511bb6af69/

It looks like it is now integrated in the main foam library.

What is important to check is that OpenFoam executables and libraries (eg libfoam.so) are actually linked to the MPI library of the toolchain, and not the one from the ThirdParty directory that they provide.
Running `ldd lib/linux64GccDPOpt/libfoam.so` should help to check this.

wpoely86:
@boegel It's ready for testing

easybuild/easyblocks/o/openfoam.py
boegel: This sounds like an additional sanity check command to me. @wpoely86?

easybuild/easyblocks/o/openfoam.py
```diff
@@ -209,19 +225,26 @@ def configure_step(self):
         if LooseVersion(self.version) < LooseVersion("2") or openfoam_extend_v3:
             self.log.debug("List of deps: %s" % self.cfg.dependencies())
             for dep in self.cfg.dependencies():
-                self.cfg.update('prebuildopts', "%s_SYSTEM=1" % dep['name'].upper())
-                self.cfg.update('prebuildopts', "{0}_LIB_DIR=$EBROOT{0}/lib".format(dep['name'].upper()))
-                self.cfg.update('prebuildopts', "{0}_INCLUDE_DIR=$EBROOT{0}/include".format(dep['name'].upper()))
+                dep_name = dep['name'].upper(),
+                dep_root = get_software_root(dep['name'])
+                dep_vars = [
+                    ("%s_SYSTEM", "1"),
```
wpoely86: I knew I had a reason for using dicts. This currently doesn't work due to this line. I don't see what's so weird about not using every key in every string formatting. It's what dicts are for?

easybuild/easyblocks/o/openfoam.py
wpoely86: Will see if I can manage something but as I disable all third party stuff, this is extremely unlikely to happen.

boegel: Doesn't hurt to have an extra sanity check for this imho, we've missed it before, we may introduce a bug that breaks it again...

besserox:
BTW, thanks for working on this one. I will need it very soon. 
I hope it will make it in EasyBuild 2.5  :-)

boegel:
@besserox: it's going to be tight, but still being considered

easybuild/easyblocks/o/openfoam.py
boegel: just hoist this out of the for loop below, and replace with:

```python
env.setvar('%s_SYSTEM' % dep_name, '1')
```

easybuild/easyblocks/o/openfoam.py
boegel: ok

easybuild/easyblocks/o/openfoam.py
```diff
@@ -128,16 +142,17 @@ def configure_step(self):
...
-                # OpenFOAM >= 3.0.0 can use 64 bit integers
-                if 'extend' not in self.name.lower() and LooseVersion(self.version) >= LooseVersion('3.0'):
-                    WM_env_var.append('WM_LABEL_SIZE') 
-                for env_var in WM_env_var:
-                    from_pat = r"^(setenv|export) (?P<var>%s)[ =](?P<val>.*)$" % env_var
-                    to_pat = r": ${\g<var>:=\g<val>}; export \g<var>"
-                    line = re.sub(from_pat, to_pat, line)
-                sys.stdout.write(line)
+            # disable any third party stuff, we use EB controlled builds
+            regex_subs = [(r"^(setenv|export) WM_THIRD_PARTY_USE_.*[ =].*$", r"# \g<0>")]
```
boegel: not so sure about this one... this is bound to break existing builds?

easybuild/easyblocks/o/openfoam.py
```diff
@@ -169,12 +184,13 @@ def configure_step(self):
...
-                for comp_var, newval in comp_vars.items():
-                    line = re.sub(r"^(%s\s*=\s*).*$" % re.escape(comp_var), r"\1%s" % newval, line)
-                sys.stdout.write(line)
+            regex_subs = []
+            for comp_var, newval in comp_vars.items():
+                regex_subs.append((r"^(%s\s*=\s*).*$" % re.escape(comp_var), r"\1%s" % newval))
+            apply_regex_substitutions(fullpath, regex_subs)
 
         # enable verbose build for debug purposes
+        # starting with openfoam-extend 3.2, PS1 also needs to be set
```
boegel: what's this about?

wpoely86: In their almighty wisdom, the openfoam-extend devs decide to only print stuff when both `FOAM_VERBOSE` and `PS1` are set.

boegel: so, we should define `$PS1` if it's not set then, to get verbose output when OpenFOAM is built as a job too?

not sure if that would have other fallout, so let's not do it in this PR please, since it's undergoing heavy testing currently to get it in for EB v2.5.0

wpoely86: I tried but it gets unset somewhere (probably bash?). It's not a major issue. It already prints some output but some finer details are not printed.

boegel: OK, fine, at least we have *some* trace of it. :-)

wpoely86: I might file a bug report about it :wink:

easybuild/easyblocks/o/openfoam.py
wpoely86: no, only used in openfoam-extend 3.2 so far

boegel:
Tested thoroughly with existing OpenFOAM(-Extend) easyconfigs + new OpenFOAM(-Extend) easyconfigs in https://github.com/hpcugent/easybuild-easyconfigs/pull/2023, https://github.com/hpcugent/easybuild-easyconfigs/pull/2048, https://github.com/hpcugent/easybuild-easyconfigs/pull/2142 and https://github.com/hpcugent/easybuild-easyconfigs/pull/2216.

So, good to go, thanks @wpoely86!

PR #747, opened 2015-11-24
Added GXX_ROOT to variables in module file as this impacts people who…

ocaisa:
… use locales and intel compilers

PR #771, opened 2015-12-12
improved implementation of determining value for $GXX_ROOT, moved to icc easyblock

boegel:
improved version of #747, we need to be a little bit careful, since this is `IntelBase`...


boegel:
cc @oaisa

wpoely86:
lgtm

boegel:

PR #777, opened 2015-12-14
[bugfix] fix determine lib subdir for recent tbb versions

boegel:
fix for small bug that was introduced in #756, which breaks the sanity check for tbb 4.3.x and 4.4.x


boegel:
Verified, so going in.

PR #780, opened 2015-12-16
Add cleanup in easyblock of OpenFoam

wpoely86:


boegel:
This requires retesting all OpenFOAM(-Extend) builds, so assigning 2.6.0 milestone on this...

boegel:
ping on retesting?

boegel:
@wpoely86: reping...

boegel:
@besserox: thoughts on this?

besserox:
@wpoely86 Can you elaborate on why it is necessary to do this?

Doesn't EasyBuild remove any previous install before building?
Or does the downloaded tarball come unclean?

wpoely86:
@besserox It's not clear to me. It's part of several easyconfigs for OpenFoam-Extend and @boegel asked to put it in the block.

Under most normal circumstances, this should do nothing but it doesn't hurt either.

besserox:
OK. It should not hurt.

Do you have a PR to remove these lines from the easyconfigs?

wpoely86:
There is now: hpcugent/easybuild-easyconfigs#2689

akesandgren:
wcleanAll is indeed needed under certain situations. Like adding patches that adds new include files for instance. So yes, please include this.
Although a simple wcleanAll without the "|| echo" stuff is quite sufficient.

wpoely86:
OK, lets gets is old PR merged then! 

We should clean up the easyconfigs too but it doesn't hurt to already merge this. In the worst case, the clean is done twice.

boegel:
Testing this with existing OpenFOAM easyconfigs...

boegel:
Checks out, no problems encountered when testing with existing `OpenFOAM` and `OpenFOAM-Extend` easyconfigs, so going in, thanks @wpoely86!

boegel:
So apparently my checking of this wasn't thorough enough, since `wcleanAll` is not available anymore for recent OpenFOAM versions, it has been replaced by `wcleanPlatform -all` (cfr. https://github.com/OpenFOAM/OpenFOAM-4.x/commit/119708e401328f40a6e10535987cea550fec97d9)...

It seems like `wcleanPlatform -all` should be used for OpenFOAM 4.x , but not for OpenFOAM-Extend 4.x...

Fix in https://github.com/easybuilders/easybuild-easyblocks/pull/1258

PR #785, opened 2015-12-20
removing shared objects from newer versions of bamtools

rjeschmi:
Newer versions of bamtools no longer include the shared libraries for bamtools-utils and jsoncpp

### DIFF
--- a/easybuild/easyblocks/b/bamtools.py
+++ b/easybuild/easyblocks/b/bamtools.py
@@ -77,8 +77,8 @@ class EB_BamTools(MakeCp, CMakeMake):
                 "lib/libjsoncpp.%s" % sharedlib_ext
             ])
         else:
-            custom_paths['files'] += [
+            custom_paths['files'].extend([
                 "lib/libbamtools-utils.a",
                 "lib/libjsoncpp.a"
-            ]
+            ])
         super(EB_BamTools, self).sanity_check_step(custom_paths=custom_paths)
--- a/easybuild/easyblocks/b/bamtools.py
+++ b/easybuild/easyblocks/b/bamtools.py
@@ -72,10 +72,10 @@ class EB_BamTools(MakeCp, CMakeMake):
         if LooseVersion(self.version) < LooseVersion('2.3.0'):
             # Buid environment changed:
             # https://github.com/pezmaster31/bamtools/commit/9cfa70bfe9cdf1b6adc06beb88246b45fdd6250a
-            custom_paths['files'] += [
+            custom_paths['files'].extend([
                 "lib/libbamtools-utils.%s" % sharedlib_ext,
                 "lib/libjsoncpp.%s" % sharedlib_ext
-            ]
+            ])
         else:
             custom_paths['files'] += [
                 "lib/libbamtools-utils.a",
--- a/easybuild/easyblocks/b/bamtools.py
+++ b/easybuild/easyblocks/b/bamtools.py
@@ -63,7 +63,7 @@ class EB_BamTools(MakeCp, CMakeMake):
                       "lib/libbamtools.%s" % sharedlib_ext],
             'dirs': ["include/api", "docs"]
         }
-        if LooseVersion(self.version) <= LooseVersion('2.3.0'):
+        if LooseVersion(self.version) < LooseVersion('2.3.0'):
             custom_paths['files'] += [
                 "lib/libbamtools-utils.%s" % sharedlib_ext,
                 "lib/libjsoncpp.%s" % sharedlib_ext]
--- a/easybuild/easyblocks/b/bamtools.py
+++ b/easybuild/easyblocks/b/bamtools.py
@@ -62,18 +62,23 @@ class EB_BamTools(MakeCp, CMakeMake):
                 "bin/bamtools",
                 "include/shared/bamtools_global.h",
                 "lib/libbamtools.a",
-                "lib/libbamtools.%s" % sharedlib_ext],
+                "lib/libbamtools.%s" % sharedlib_ext
+            ],
             'dirs': [
                 "include/api",
-                "docs"]
+                "docs"
+            ]
         }
         if LooseVersion(self.version) < LooseVersion('2.3.0'):
-            #Buid environment changed https://github.com/pezmaster31/bamtools/commit/9cfa70bfe9cdf1b6adc06beb88246b45fdd6250a
+            # Buid environment changed:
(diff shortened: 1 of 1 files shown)


rjeschmi:
as reported by @nathanhaigh

easybuild/easyblocks/b/bamtools.py
```diff
@@ -57,11 +57,15 @@ def sanity_check_step(self):
 
         sharedlib_ext = get_shared_lib_ext()
 
+
```
wpoely86: drop this line

wpoely86:
These libraries do longer exists or what happened to them?

rjeschmi:
They aren't built any more as shared libraries. They are built statically.

wpoely86:
Then add a check of the static version?

easybuild/easyblocks/b/bamtools.py
```diff
@@ -57,11 +57,15 @@ def sanity_check_step(self):
...
         custom_paths = {
             'files': ["bin/bamtools", "include/shared/bamtools_global.h", "lib/libbamtools.a",
-                      "lib/libbamtools.%s" % sharedlib_ext, "lib/libbamtools-utils.%s" % sharedlib_ext,
-                      "lib/libjsoncpp.%s" % sharedlib_ext],
+                      "lib/libbamtools.%s" % sharedlib_ext],
             'dirs': ["include/api", "docs"]
         }
+        if LooseVersion(self.version) < LooseVersion('2.3.0'):
```
boegel: including a reference for this in a comment would be great

boegel:
+1 on @wpoely86's remark. Thanks for sharing @nathanhaigh , very useful!

easybuild/easyblocks/b/bamtools.py
```diff
@@ -58,10 +58,22 @@ def sanity_check_step(self):
...
         custom_paths = {
-            'files': ["bin/bamtools", "include/shared/bamtools_global.h", "lib/libbamtools.a",
-                      "lib/libbamtools.%s" % sharedlib_ext, "lib/libbamtools-utils.%s" % sharedlib_ext,
-                      "lib/libjsoncpp.%s" % sharedlib_ext],
-            'dirs': ["include/api", "docs"]
+            'files': [
+                "bin/bamtools",
+                "include/shared/bamtools_global.h",
+                "lib/libbamtools.a",
+                "lib/libbamtools.%s" % sharedlib_ext],
```
wpoely86: The `]` should be on the next line.

wpoely86:
Small style nit pick but else good to go

easybuild/easyblocks/b/bamtools.py
rjeschmi: I've started using pylint and it didn't warn on this.

I'm not sure whether it is better or not really. I guess it is a question of whether you prefer array items to be aligned as close to the left as possible.

I can fix it, but I'd need to fix more than this one in the file to be consistent.

wpoely86: It's pure style and valid python. I use `pep8` for style and it will warn about this.

But I agree that we're not consistent in this :wink:

rjeschmi: I added pep8 and couldn't get it to show the error by default either. What options do you have enabled?

I just find the style fixes to be tedious and would rather know the standard and ensure the code passes a standard linter.

wpoely86: Weird, it's there: http://legacy.python.org/dev/peps/pep-0008/#indentation (at the end)

Yeah, style fixes are annoying but with a good editor, they are done properly by default?

rjeschmi:
Ok, I think this is consistent.

Ideally, we should come up with a reproducible standard so that we can reduce everyone's workload on these sort of things.

Thanks for your patience @wpoely86  :)

easybuild/easyblocks/b/bamtools.py
```diff
@@ -58,10 +58,27 @@ def sanity_check_step(self):
...
         custom_paths = {
-            'files': ["bin/bamtools", "include/shared/bamtools_global.h", "lib/libbamtools.a",
-                      "lib/libbamtools.%s" % sharedlib_ext, "lib/libbamtools-utils.%s" % sharedlib_ext,
-                      "lib/libjsoncpp.%s" % sharedlib_ext],
-            'dirs': ["include/api", "docs"]
+            'files': [
+                "bin/bamtools",
+                "include/shared/bamtools_global.h",
+                "lib/libbamtools.a",
+                "lib/libbamtools.%s" % sharedlib_ext
```
boegel: hmm, this is getting a bit long (vertically), but fine :-)

easybuild/easyblocks/b/bamtools.py
```diff
@@ -58,10 +58,27 @@ def sanity_check_step(self):
...
+            'dirs': [
+                "include/api",
+                "docs"
+            ]
         }
-
+        if LooseVersion(self.version) < LooseVersion('2.3.0'):
+            # Buid environment changed:
+            # https://github.com/pezmaster31/bamtools/commit/9cfa70bfe9cdf1b6adc06beb88246b45fdd6250a
+            custom_paths['files'] += [
```
boegel: using `.extend` here instead of `+=` is more Pythonic imho

```python
custom_paths['files'].extend([
    ...
])
```

easybuild/easyblocks/b/bamtools.py
```diff
@@ -58,10 +58,27 @@ def sanity_check_step(self):
...
-
+        if LooseVersion(self.version) < LooseVersion('2.3.0'):
+            # Buid environment changed:
+            # https://github.com/pezmaster31/bamtools/commit/9cfa70bfe9cdf1b6adc06beb88246b45fdd6250a
+            custom_paths['files'].extend([
+                "lib/libbamtools-utils.%s" % sharedlib_ext,
+                "lib/libjsoncpp.%s" % sharedlib_ext
+            ])
+        else:
+            custom_paths['files'] += [
```
boegel: pssst, `.extend` here too please, keep things consistent

easybuild/easyblocks/b/bamtools.py
boegel: I'm having second thoughts about this, it doesn't follow the style of other easyblocks...

easybuild/easyblocks/b/bamtools.py
boegel: ping

boegel:
retested all existing BamTools easyconfigs on top of this, good to go, thanks @rjeschmi!

PR #786, opened 2015-12-31
use 'python -S' when installing Python packages using system Python (WIP)

boegel:
This should fix the problem with EasyBuild picking up system Python packages when installing Python packages, cfr. https://github.com/hpcugent/easybuild-framework/issues/1409 .

Since this also impacts the installation of EasyBuild itself (which is typically installed on top of the system Python), changes to the EasyBuildMeta easyblock are also required.

### DIFF
--- a/easybuild/easyblocks/generic/pythonpackage.py
+++ b/easybuild/easyblocks/generic/pythonpackage.py
@@ -42,13 +42,12 @@ from easybuild.easyblocks.python import EXTS_FILTER_PYTHON_PACKAGES
 from easybuild.framework.easyconfig import CUSTOM
 from easybuild.framework.extensioneasyblock import ExtensionEasyBlock
 from easybuild.tools.build_log import EasyBuildError
-from easybuild.tools.config import build_option
 from easybuild.tools.filetools import mkdir, rmtree2, which
 from easybuild.tools.run import run_cmd
 
 
 # not 'easy_install' deliberately, to avoid that pkg installations listed in easy-install.pth get preference
-EASY_INSTALL_CMD = "python setup.py easy_install"
+EASY_INSTALL_CMD = "python %(python_opts)s setup.py easy_install"
 UNKNOWN = 'UNKNOWN'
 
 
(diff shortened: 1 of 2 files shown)


boegel:
Further testing is needed, since recent versions of `vsc-base` (indirectly) require `setuptools` to be readily available, so it will need to be made part of the EasyBuild install procedure...

Also, we should consider checking for a `setuptools.pth` file in `post_install_step` for EasyBuild, and remove it if it's present.

boegel:
Hmm, this seems to be breaking things..

* GC3Pie: includes `setuptools` as 1st extension, but then not available when installing the other extensions due to `python -S` being used
* JUBE is installed via the `VersionIndependentPythonPackage` easyblock; it uses `--single-version-externally-managed`, which is an option only available when `setuptools` is used (which was previously picked up from the system)

easybuild/easyblocks/generic/pythonpackage.py
```diff
@@ -127,9 +126,22 @@ def __init__(self, *args, **kwargs):
...
         else:
-            self.install_cmd = "python setup.py install"
+            self.install_cmd = "python %(python_opts)s setup.py install"
             self.install_cmd_extra = None
 
+        self.python_opts = []
+
+    def set_python_opts(self):
+        """Set options to pass to 'python'."""
+        # use python -S when using system Python, to avoid that Python packages installed system-wide are picked up
```
JensTimmerman: why would you be using the system python in the first place?shouldn't this be an error?

JensTimmerman:
to install version independed python packages, you could use a python-base module instead of the system python? This way you know it's site-packages are clean? and you don't need to use -S, so the setuptools you put in the pythonpath because it's a dependency will still be picked up for e.g. GC3P and JUBE ?

boegel:
@JensTimmerman: but then you're stuck to that Python module... How do you install another Python module in that situation?

Anyway, `python -S` isn't the answer here. In fact, it causes more problems than it solves. It breaks `setuptools` for example, so this is a no-go.

proper fix is a combo of https://github.com/hpcugent/easybuild-framework/pull/1569 and https://github.com/hpcugent/easybuild-easyblocks/pull/813

JensTimmerman:
no you're not stuck on that python module.
you're using a python which has an empty site-packages, which is what you want, you can use any other python with an empyt site-packages.

since you're installing in a prefix, and not in the python's site packages, you're not stuck with anything.

boegel:
@JensTimmerman: if this Python is provided via a module, you are stuck to it (unless you unload the module after firing EasyBuild)

anyway, I feel that installing another Python on your system just to use it with EB is weird, and shouldn't be needed at all...

JensTimmerman:
What? you don't need another python on your system just to use with EB, you already need another python because you're installing a PythonPackage, pythonpackages need python, so it makes sence to install them with the pristine and clean python module you already installed.

The fact that EasyBuild has issues when creating modules when another python module is loaded is a bug in EasyBuild and should be fixed, not a limitation of the scheme I am proposing.

PR #789, opened 2016-01-02
don't hardcode .so, use get_shared_lib_ext instead

citibob:
Repeat of .dylib fix

(edit: same changes as in #788, but to `develop`)

PR #790, opened 2016-01-03
bug fix in ACML easyblock and minor style fixes after #789 (1/4)

boegel:
follow-up for #789


boegel:
All easyblocks touched have been retested, no (other) bugs were introduced.

Going in to fix bug introduced in #789 and various style remarks.

PR #791, opened 2016-01-03
minor style fixes after #789 (2/4)

boegel:
follow-up for #789


boegel:
Going in to fix style issues that came into view in #789

PR #792, opened 2016-01-03
160103 cmakemakecp

citibeth:
CMake-based counterpart for software that lakes a "make install" counterpart.  This is useful to convert easyconfigs that previously usd MakeCp, but where the underlying project has converted to CMake.

### DIFF
--- a/easybuild/easyblocks/generic/cmakemakecp.py
+++ b/easybuild/easyblocks/generic/cmakemakecp.py
@@ -1,0 +1,122 @@
+##
+# Copyright 2013 the Cyprus Institute
+#
+# This file is part of EasyBuild,
+# originally created by the HPC team of Ghent University (http://ugent.be/hpc/en),
+# with support of Ghent University (http://ugent.be/hpc),
+# the Flemish Supercomputer Centre (VSC) (https://vscentrum.be/nl/en),
+# the Hercules foundation (http://www.herculesstichting.be/in_English)
+# and the Department of Economy, Science and Innovation (EWI) (http://www.ewi-vlaanderen.be/en).
+#
+# http://github.com/hpcugent/easybuild
+#
+# EasyBuild is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation v2.
+#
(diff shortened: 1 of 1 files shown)


citibeth:
yes

easybuild/easyblocks/generic/cmakemakecp.py
```diff
@@ -0,0 +1,122 @@
+##
+# Copyright 2013 the Cyprus Institute
```
wpoely86: We're at 2016 :wink:

easybuild/easyblocks/generic/cmakemakecp.py
```diff
@@ -0,0 +1,122 @@
...
+    Running CMake is optional (see 'with_configure' in extra_options)
+    """
+    @staticmethod
+    def extra_options(extra_vars=None):
+        """
+        Define list of files or directories to be copied after make
+        """
+        extra = {
+            'files_to_copy': [[], "List of files or dirs to copy", MANDATORY],
+            'with_configure': [False, "Run CMake before building", BUILD],
```
wpoely86: this might be confusing. We should rename it to `with_cmake`?

easybuild/easyblocks/generic/cmakemakecp.py
```diff
@@ -0,0 +1,122 @@
...
+            extra_vars = {}
+        extra.update(extra_vars)
+        return CMakeMake.extra_options(extra_vars=extra)
+
+    def configure_step(self, *args, **kwargs):
+        """
+        Configure build if required
+        """
+        if self.cfg.get('with_configure', False):
+            return super(CMakeMakeCp, self).configure_step(*args, **kwargs)
```
wpoely86: if this is the only change from the `MakeCp` block, we should just inherit from both `MakeCp` and `CMakeMake` and only change the `configure_step`.

wpoely86:
@citibeth Looks fine mostly. Could you retarget to the develop branch? Please split the `cmakemakecp` to a separate PR as it's something completely different.

boegel:
@citibeth: you'll need to reopen this PR and target `develop`...

citibeth:
OK, see PR #819

PR #797, opened 2016-01-05
Update CPLEX easyblock

sbecuwe:
- added extra PATHs
- added LD_LIBRARY_PATH
- added CPLEXDIR (as used in the Makefiles)

### DIFF
--- a/easybuild/easyblocks/c/cplex.py
+++ b/easybuild/easyblocks/c/cplex.py
@@ -110,11 +110,21 @@ class EB_CPLEX(Binary):
             raise EasyBuildError("No bins found using %s in %s", binglob, self.installdir)
 
     def make_module_extra(self):
-        """Add installdir to path and set CPLEX_HOME"""
+        """Add bin dirs and lib dirs and set CPLEX_HOME and CPLEXDIR"""
+
+        os.chdir(self.installdir)
+        binglob = '*/bin/x86-64*'
+        bins = glob.glob(binglob)
+        libglob = '*/lib/x86-64*/*pic'
+        libs = glob.glob(libglob)
 
         txt = super(EB_CPLEX, self).make_module_extra()
-        txt += self.module_generator.prepend_paths("PATH", [self.bindir])
+
+        txt += self.module_generator.prepend_paths("PATH", [path for path in bins])
+        txt += self.module_generator.prepend_paths("LD_LIBRARY_PATH", reduce(lambda x, y: x+y, [bins,libs]))
+
         txt += self.module_generator.set_environment("CPLEX_HOME", os.path.join(self.installdir, 'cplex'))
+        txt += self.module_generator.set_environment("CPLEXDIR", os.path.join(self.installdir, 'cplex'))
         self.log.debug("make_module_extra added %s" % txt)
         return txt
 
--- a/easybuild/easyblocks/c/cplex.py
+++ b/easybuild/easyblocks/c/cplex.py
@@ -113,15 +113,13 @@ class EB_CPLEX(Binary):
         """Add bin dirs and lib dirs and set CPLEX_HOME and CPLEXDIR"""
 
         os.chdir(self.installdir)
-        binglob = '*/bin/x86-64*'
-        bins = glob.glob(binglob)
-        libglob = '*/lib/x86-64*/*pic'
-        libs = glob.glob(libglob)
+        bins = glob.glob(os.path.join('*', 'bin', 'x86-64*'))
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/c/cplex.py
```diff
@@ -110,11 +110,21 @@ def post_install_step(self):
...
 
     def make_module_extra(self):
-        """Add installdir to path and set CPLEX_HOME"""
+        """Add bin dirs and lib dirs and set CPLEX_HOME and CPLEXDIR"""
+
+        os.chdir(self.installdir)
+        binglob = '*/bin/x86-64*'
+        bins = glob.glob(binglob)
+        libglob = '*/lib/x86-64*/*pic'
+        libs = glob.glob(libglob)
```
boegel: this should be wrapped in a `try:`/`except OSError`?

easybuild/easyblocks/c/cplex.py
```diff
@@ -110,11 +110,21 @@ def post_install_step(self):
             raise EasyBuildError("No bins found using %s in %s", binglob, self.installdir)
 
     def make_module_extra(self):
-        """Add installdir to path and set CPLEX_HOME"""
+        """Add bin dirs and lib dirs and set CPLEX_HOME and CPLEXDIR"""
+
+        os.chdir(self.installdir)
```
boegel: why not use `os.path.join(self.installdir, '*', 'bin', 'x86-64*')` instead of changing to this directory?

if you insist on changing there, make sure you change back to where you were?

easybuild/easyblocks/c/cplex.py
```diff
@@ -110,11 +110,21 @@ def post_install_step(self):
...
+        binglob = '*/bin/x86-64*'
+        bins = glob.glob(binglob)
+        libglob = '*/lib/x86-64*/*pic'
+        libs = glob.glob(libglob)
 
         txt = super(EB_CPLEX, self).make_module_extra()
-        txt += self.module_generator.prepend_paths("PATH", [self.bindir])
+
+        txt += self.module_generator.prepend_paths("PATH", [path for path in bins])
+        txt += self.module_generator.prepend_paths("LD_LIBRARY_PATH", reduce(lambda x, y: x+y, [bins,libs]))
```
boegel: why not use `bins + libs` instead?

easybuild/easyblocks/c/cplex.py
```diff
@@ -110,11 +110,21 @@ def post_install_step(self):
...
+
         txt += self.module_generator.set_environment("CPLEX_HOME", os.path.join(self.installdir, 'cplex'))
+        txt += self.module_generator.set_environment("CPLEXDIR", os.path.join(self.installdir, 'cplex'))
```
boegel: hmm, do we have a reference for these? I'm wondering why one is using `_`, and another is not...

boegel:
Jenkins: ok to test

easybuild/easyblocks/c/cplex.py
sbecuwe: CPLEXDIR is what the Makefiles of CPLEX expect to find. CPLEX_HOME was in the original cplex.py so I didn't touch it.

easybuild/easyblocks/c/cplex.py
sbecuwe: Correct. I will update it.

easybuild/easyblocks/c/cplex.py
sbecuwe: I started from the original cplex.py file by adding lines ... I agree this looks nicer. I will update it.

sbecuwe: Decided not to add self.installdir to the above suggestion, since this would result in absolute paths in PATH and LD_LIBRARY_PATH, which is, by default, not the case: relative paths seem to be preferred.

easybuild/easyblocks/c/cplex.py
sbecuwe: No. If it's empty, it is. It's actually tested in the post_install_step.

easybuild/easyblocks/c/cplex.py
boegel: OK, makes sense.

I did some minor cleanup in https://github.com/hpcuantwerpen/easybuild-easyblocks/pull/2, without removing the `chdir`, please take a look & merge?

boegel:
tested with existing CPLEX easyconfigs and https://github.com/hpcugent/easybuild-easyconfigs/pull/2259, good to go

Thanks @sbecuwe!

PR #798, opened 2016-01-06
support eigen 2.x in eigen easyblock

pescobar:
eigen easyblock was working fine for me to install eigen 2.x but this commit broke eigen2.x support
https://github.com/hpcugent/easybuild-easyblocks/commit/59120dabc3b3134b74f793e88c775a9328380fd0

with this modifications both 2.x and 3.x are supported.

### DIFF
--- a/easybuild/easyblocks/e/eigen.py
+++ b/easybuild/easyblocks/e/eigen.py
@@ -19,6 +19,7 @@ EasyBuild support for building and installing Eigen, implemented as an easyblock
 
 import os
 import shutil
+from distutils.version import LooseVersion
 
 from easybuild.framework.easyblock import EasyBlock
 from easybuild.tools.build_log import EasyBuildError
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/e/eigen.py
```diff
@@ -73,9 +75,13 @@ def sanity_check_step(self):
                                                        'QtAlignedMalloc', 'SVD', 'Sparse',
                                                        'SparseCholesky', 'SparseCore', 'StdDeque',
                                                        'StdList', 'StdVector', 'SuperLUSupport',
-                                                       'UmfPackSupport']] + ['include/signature_of_eigen3_matrix_library'],
+                                                       'UmfPackSupport']],
             'dirs': []
         }
+        
+        if LooseVersion(self.version) > LooseVersion('3.0'):
+            custom_paths['files'] = custom_paths['files'] + ['include/signature_of_eigen3_matrix_library']
```
boegel: please use `custom_paths['files'].append('include/signature_of_eigen3_matrix_library')`

boegel:
@pescobar: I guess the reason we missed this is because there's no Eigen v2.x easyconfig included in the easyconfigs repo... do you have one you can contribute?

pescobar:
@boegel just sent my eigen2 easyconfig 
https://github.com/hpcugent/easybuild-easyconfigs/pull/2272

pescobar:
I just noticed these are the files in my eigen 2.x install:

```
[escobar@login18 ~]$ ls /scicore/soft/apps/Eigen/2.0.17-goolf-1.4.10/include/Eigen/
Array  Cholesky  CMakeLists.txt  Core  Dense  Eigen  Geometry  LeastSquares  LU  NewStdVector  QR  QtAlignedMalloc  Sparse  src  StdVector  SVD
```

based on this the sanity check in this block should not work but it works :?    I need to review this in detail....

boegel:
How old is that install? Maybe it was done with an older version of the Eigen easyblock?

easybuild/easyblocks/e/eigen.py
```diff
@@ -73,9 +75,13 @@ def sanity_check_step(self):
                                                        'QtAlignedMalloc', 'SVD', 'Sparse',
                                                        'SparseCholesky', 'SparseCore', 'StdDeque',
                                                        'StdList', 'StdVector', 'SuperLUSupport',
-                                                       'UmfPackSupport']] + ['include/signature_of_eigen3_matrix_library'],
+                                                       'UmfPackSupport']],
             'dirs': []
         }
+        
+        if LooseVersion(self.version) > LooseVersion('3.0'):
```
boegel: ah, maybe change this to `>=` (may not matter, but it's "more correct")

boegel:
tested with https://github.com/hpcugent/easybuild-easyconfigs/pull/2272, good to go, thanks @pescobar!

PR #799, opened 2016-01-07
Fix sanity check of netcdf4-python

wpoely86:
Now works for both egg and non-egg versions

Needed for hpcugent/easybuild-easyconfigs#2221

### DIFF
--- a/easybuild/easyblocks/n/netcdf4_python.py
+++ b/easybuild/easyblocks/n/netcdf4_python.py
@@ -32,7 +32,6 @@ import os
 import easybuild.tools.environment as env
 from easybuild.easyblocks.generic.pythonpackage import PythonPackage
 from easybuild.tools.modules import get_software_root
-from easybuild.tools.systemtools import get_shared_lib_ext
 
 
 class EB_netcdf4_minus_python(PythonPackage):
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/n/netcdf4_python.py
```diff
@@ -70,11 +69,8 @@ def test_step(self):
 
     def sanity_check_step(self):
         """Custom sanity check for netcdf4-python"""
-        shlib_ext = get_shared_lib_ext()
-
         custom_paths = {
-            'files': ['bin/nc3tonc4', 'bin/nc4tonc3', 'bin/ncinfo'] +
-                     [os.path.join(self.pylibdir, x) for x in ['netCDF4.%s' % shlib_ext, 'netCDF4_utils.py', 'netcdftime.py']],
```
boegel: is the `.so` also in the `.egg`? if not, keep it separate as a file to check on

wpoely86: it's all in the egg

boegel: OK, then this is good enough I guess, especially since it'll do the `import` check too.

boegel:
Thanks @wpoely86!

Issue #725, opened 2015-10-28
use apply_regex_substitutions in easyblocks, where appropriate


boegel:
requires https://github.com/hpcugent/easybuild-framework/pull/1388

TODO:

* [ ] aladin.py
* [ ] atlas.py
* [ ] boost.py
* [ ] clang.py
* [ ] cp2k.py
* [ ] cufflinks.py
* [ ] ferret.py
* [ ] gamess_us.py
* [ ] impi.py
* [ ] molpro.py
* [ ] mummer.py
* [ ] ncl.py
* [ ] python.py
* [ ] quantumespresso.py
* [ ] rosetta.py
* [ ] scipion.py
* [ ] scotch.py
* [ ] suitesparse.py
* [ ] tinker.py
* [ ] trinity.py
* [ ] wien2k.py
* [ ] wps.py
* [x] wrf.py (cfr. #685)
* [ ] xcrysden.py
* [ ] xmipp.py

PR #685, opened 2015-09-11
use apply_regex_substitutions in WRF easyblock

boegel:
requires ~~https://github.com/hpcugent/easybuild-framework/pull/1388~~ and https://github.com/hpcugent/easybuild-framework/pull/1458

PR #802, opened 2016-01-11
Trinity easyblock bumped to work with version 2.1.1

hajgato:


easybuild/easyblocks/t/trinity.py
```diff
@@ -114,6 +115,8 @@ def inchworm(self, run=True):
         """Install procedure for Inchworm."""
 
         make_flags = 'CXXFLAGS="%s %s"' % (os.getenv('CXXFLAGS'), self.toolchain.get_flag('openmp'))
+        if LooseVersion(self.version) > "2" and LooseVersion(self.version) < "3":
```
boegel: you're comparing a `LooseVersion` instance with a string here

Python allows that, and it may *seem* to work, but it doesn't do what you think it does

please change to (note the `>=`):

```python
if LooseVersion(self.version) >= LooseVersion('2.0') and LooseVersion(self.version) < LooseVersion('3.0'):
```

easybuild/easyblocks/t/trinity.py
```diff
@@ -230,7 +233,7 @@ def build_step(self):
     def install_step(self):
         """Custom install procedure for Trinity."""
 
-        if LooseVersion(self.version) < LooseVersion('2012-10-05'):
+        if LooseVersion(self.version) < LooseVersion('2012-10-05') and LooseVersion(self.version) > "3":
```
boegel: same here, compare with `LooseVersion('3.0')`

also, add a comment above this line to make it clear that this does not apply for Trinity 2.x (which explains the somewhat weird `> 3` requirement

maybe we should even change the `> 3` to `>= 2012`?

this is kind of dead code though, we don't ship any Trinity easyconfigs older than the 2012-10-05 'version' (but keep it there)

easybuild/easyblocks/t/trinity.py
```diff
@@ -261,12 +264,16 @@ def install_step(self):
...
+                if LooseVersion(self.version) > "2" and LooseVersion(self.version) < "3":
+                    line = re.sub(r'^( INCHWORM_CONFIGURE_FLAGS\s*=\s*).*$', r'\1%s' % inchworm_flags, line)
+                    line = re.sub(r'^( CHRYSALIS_MAKE_FLAGS\s*=\s*).*$', r'\1%s' % chrysalis_flags, line)
+                else:
+                    line = re.sub(r'^(INCHWORM_CONFIGURE_FLAGS\s*=\s*).*$', r'\1%s' % inchworm_flags, line)
+                    line = re.sub(r'^(CHRYSALIS_MAKE_FLAGS\s*=\s*).*$', r'\1%s' % chrysalis_flags, line)
+                    line = re.sub(r'(/rsem && \$\(MAKE\))\s*$',
+                                  r'\1 CC=%s CXX="%s %s" CFLAGS_EXTRA="%s"\n' % (cc, cxx, lib_flags, lib_flags), line)
+                    line = re.sub(r'(/fastool && \$\(MAKE\))\s*$',
+                                  r'\1 CC="%s -std=c99" CFLAGS="%s ${CFLAGS}"\n' % (cc, lib_flags), line)
```
boegel: since we're touching this, it would be better to switch to using `apply_regex_substitutions`, are you up for that @hajgato?

cfr. #725, PR #685 is an example

easybuild/easyblocks/t/trinity.py
```diff
@@ -279,7 +286,16 @@ def install_step(self):
             else:
                 raise EasyBuildError("Don't know how to set TRINITY_COMPILER for %s compiler", comp_fam)
 
-            cmd = "make TRINITY_COMPILER=%s" % trinity_compiler
+            explicit_make_args = ''
+            if LooseVersion(self.version) > "2" and LooseVersion(self.version) < "3":
```
boegel: same here, don't compare `LooseVersion` instances with strings please

easybuild/easyblocks/t/trinity.py
```diff
@@ -279,7 +286,16 @@ def install_step(self):
...
+                explicit_make_args = 'all plugins'
+            # Delete CXXFLAGS CPPFLAGS and CFLAGS (interfere with plugins installation)
+            if LooseVersion(self.version) > "2" and LooseVersion(self.version) < "3":
```
boegel: no string comparison please

boegel: no need to check again, just drop this line and indent the comment?

easybuild/easyblocks/t/trinity.py
```diff
@@ -295,7 +311,11 @@ def install_step(self):
     def sanity_check_step(self):
         """Custom sanity check for Trinity."""
 
-        path = 'trinityrnaseq_r%s' % self.version
+        sep_r_inpath='_r'
+        if LooseVersion(self.version) > "2" and LooseVersion(self.version) < "3":
```
boegel: fix comparison

easybuild/easyblocks/t/trinity.py
```diff
@@ -295,7 +311,11 @@ def install_step(self):
     def sanity_check_step(self):
         """Custom sanity check for Trinity."""
 
-        path = 'trinityrnaseq_r%s' % self.version
+        sep_r_inpath='_r'
```
boegel: style: spaces around `=`

easybuild/easyblocks/t/trinity.py
```diff
@@ -295,7 +311,11 @@ def install_step(self):
...
+        if LooseVersion(self.version) > "2" and LooseVersion(self.version) < "3":
+            sep_r_inpath='-'
```
boegel: style: spaces around `=`

boegel:
@hajgato: remarks fixes in https://github.com/hajgato/easybuild-easyblocks/pull/12

easybuild/easyblocks/t/trinity.py
```diff
@@ -279,7 +286,16 @@ def install_step(self):
...
 
-            cmd = "make TRINITY_COMPILER=%s" % trinity_compiler
+            explicit_make_args = ''
+            if LooseVersion(self.version) > "2" and LooseVersion(self.version) < "3":
+                explicit_make_args = 'all plugins'
+            # Delete CXXFLAGS CPPFLAGS and CFLAGS (interfere with plugins installation)
+            if LooseVersion(self.version) > "2" and LooseVersion(self.version) < "3":
+                del os.environ['CXXFLAGS']
+                del os.environ['CPPFLAGS']
+                del os.environ['CFLAGS']
```
boegel: The problem I bumped into in https://github.com/hpcugent/easybuild-easyconfigs/pull/2306#issuecomment-170876129 is related to this.

The reason that this is needed seems to be that `gcc` is hardcoded for the plugins, and using Intel compiler flags with GCC isn't working out well.

You're not undefining `$LIBS` here, which explains the problem I'm running into.

However, I feel this is wrong approach: we should make sure the Intel compilers are used instead where they should be?

boegel: fixed with https://github.com/hajgato/easybuild-easyconfigs/pull/31

hajgato:
@boegel: merged

easybuild/easyblocks/t/trinity.py
hajgato: @boegel: Yes, I am a bit afraid of this. I am not 100% sure how we have to handle this, as it seems that the `Trinity` authors recommend only `Chrysalis` and `Inchworm` to be compiled with Intel, and the rest with `GCC`. I do not know what is the correct solution in this cases, as I do not know how to pull the right `GCC` compiler flags. (And I am not sure that we do not have to use the hard-coded options). Anyway, one of our user already got into a problem with `Intel` compiled `blastx`, it gave wrong results (but not `blastn`), and apparently the `foss` version was rather similar in speed .

boegel: I rarely run into problems anymore when swapping GCC with the Intel compilers, I think BLAST+ is an exception here (do you have an issue for it?).
Usually, *if* is happens, it can be fixed by making sure 'strict' precision is used; the Intel compilers are (a lot) less strict w.r.t. floating-point precision that GCC is. EB already 'corrects' for that a bit by specifying certain precision-related flags by default when using the Intel compilers, but sometimes they need to be made a little bit stricter (by using `'strict': True` in `toolchainopts` in the easyconfig).
Also, I'm not sure that explicitly using GCC for the Trinity plugins is actually the intention; it looks like a case of "GCC is the only compiler" to me...

Also, this is relevant: https://github.com/hpcugent/easybuild-easyconfigs/blob/master/easybuild/easyconfigs/t/Trinity/trinity_ictce-no-jellyfish.patch .
Older versions of jellyfish seemed to fail to build with Intel compilers, this is clearly not the case anymore.

hajgato: @boegel: Maybe we have to consider compile everything with `Intel` compiler then?

boegel: Sure, we can try and track down the places where `gcc` or `g++` is hardcoded, and see how much effort it is to get that changed (and working).

But that's for a follow-up PR imho.

boegel:
tested with https://github.com/hpcugent/easybuild-easyconfigs/pull/2306, good to go, thanks @hajgato!

PR #806, opened 2016-01-13
Add an option to skip the sanitizer tests of Clang

wpoely86:
These sanitizer tests often fail on HPC systems due to the 'weird'
evironment. It doesn't neccessarily mean that they don't work (they
usually do).

@boegel please review

### DIFF
--- a/easybuild/easyblocks/c/clang.py
+++ b/easybuild/easyblocks/c/clang.py
@@ -187,10 +187,7 @@ class EB_Clang(CMakeMake):
                 self.log.warn("The stacksize limit is set to unlimited. This causes the ThreadSanitizer "
                               "to fail. The sanitizers tests will be disabled unless --strict=error is used.")
 
-            if self.cfg['skip_sanitizer_tests']:
-                disable_san_tests = True
-
-            if disable_san_tests and build_option('strict') != run.ERROR:
+            if (disable_san_tests or self.cfg['skip_sanitizer_tests']) and build_option('strict') != run.ERROR:
                 self.log.debug("Disabling the sanitizer tests")
                 self.disable_sanitizer_tests()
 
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/c/clang.py
```diff
@@ -185,7 +187,11 @@ def configure_step(self):
                 self.log.warn("The stacksize limit is set to unlimited. This causes the ThreadSanitizer "
                               "to fail. The sanitizers tests will be disabled unless --strict=error is used.")
 
+            if self.cfg['skip_sanitizer_tests']:
```
boegel: use this below instead:

```python
if (disable_san_tests or self.cfg['skip_sanitizer_tests']) and build_option('strict') != run.ERROR:
    self.log.debug("Disabling the sanitizer tests")
    self.disable_sanitizer_tests()
```

easybuild/easyblocks/c/clang.py
```diff
@@ -67,6 +67,8 @@ def extra_options():
             'bootstrap': [True, "Bootstrap Clang using GCC", CUSTOM],
             'usepolly': [False, "Build Clang with polly", CUSTOM],
             'static_analyzer': [True, "Install the static analyser of Clang", CUSTOM],
+            # The sanitizer tests often fail on HPC systems due to the 'weird' environment.
+            'skip_sanitizer_tests': [False, "Do not run the sanitizer tests", CUSTOM],
```
boegel: enable by default, to avoid people tripping over this?

I would also mention the error you may run into in the description (take them from the EB ML post), for easy googling... ;-)

wpoely86: Why, these test should be run. It's only on specific systems that it fails.

boegel: OK, fine.

Can you copy-paste the error messages in the PR description? I'll merge this in then.

wpoely86: Which error messages? Some sanitizer test can fail, but which one is random. It will only tell you test X, Y, ... have failed

boegel: OK, fine. :-)

PR #684, opened 2015-09-08
Molpro

valtandor:
Changes to Molpro easyblock to accommodate precompiled binaries

PR #807, opened 2016-01-16
update Molpro easyblock to support binary installs and 2015 version

boegel:
builds on top of #684 by by @valtandor, fixes some style issues, adds support for installing the 2015 version of Molpro and replaces use of `fileinput` with `apply_regex_substitutions`


boegel:
Reviewed/enhanced version of #684, going in, thanks @valtandor!

PR #810, opened 2016-01-21
make ConfigureMake more robust w.r.t. custom easyconfig parameters

boegel:
This makes the `ConfigureMake` easyblock easier to derive when multiple inheritance is used and `ConfigureMake` isn't the 'main' parent, i.e. it avoids having to fiddle with `extra_options` to avoid problems like:

```
Use of unknown easyconfig parameter 'configure_cmd_prefix' when getting parameter value
```

### DIFF
--- a/easybuild/easyblocks/generic/configuremake.py
+++ b/easybuild/easyblocks/generic/configuremake.py
@@ -50,7 +50,7 @@ class ConfigureMake(EasyBlock):
         extra_vars = EasyBlock.extra_options(extra=extra_vars)
         extra_vars.update({
             'configure_cmd_prefix': ['', "Prefix to be glued before ./configure", CUSTOM],
-            'prefix_opt': ['--prefix=', "Prefix command line option for configure script", CUSTOM],
+            'prefix_opt': [None, "Prefix command line option for configure script ('--prefix=' if None)", CUSTOM],
             'tar_config_opts': [False, "Override tar settings as determined by configure.", CUSTOM],
         })
         return extra_vars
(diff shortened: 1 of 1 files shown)


boegel:
@wpoely86: please review

easybuild/easyblocks/generic/configuremake.py
```diff
@@ -50,7 +50,7 @@ def extra_options(extra_vars=None):
         extra_vars = EasyBlock.extra_options(extra=extra_vars)
         extra_vars.update({
             'configure_cmd_prefix': ['', "Prefix to be glued before ./configure", CUSTOM],
-            'prefix_opt': ['--prefix=', "Prefix command line option for configure script", CUSTOM],
+            'prefix_opt': [None, "Prefix command line option for configure script ('--prefix=' if None)", CUSTOM],
```
wpoely86: explain more. Why would you not use a default value?

easybuild/easyblocks/generic/configuremake.py
```diff
@@ -82,7 +82,7 @@ def configure_step(self, cmd_prefix=''):
         cmd = "%(preconfigopts)s %(cmd_prefix)s./configure %(prefix_opt)s%(installdir)s %(configopts)s" % {
             'preconfigopts': self.cfg['preconfigopts'],
             'cmd_prefix': cmd_prefix,
-            'prefix_opt': self.cfg['prefix_opt'],
+            'prefix_opt': self.cfg.get('prefix_opt') or '--prefix=',
```
wpoely86: what if it's an empty string?

easybuild/easyblocks/generic/configuremake.py
boegel: because then I need to use the default value in two places, here, and below

the whole point of this change is that there are situations (with derived easyblocks and multiple inheritance) where `prefix_opt` may not be available in `self.cfg`

easybuild/easyblocks/generic/configuremake.py
boegel: ok, good point, will fix

boegel:

PR #763, opened 2015-12-10
fix for vsc-base being picked from OS during EasyBuild install (WIP)

boegel:
If vsc-base is also available on the OS, it may be picked up during the EasyBuild installation, causing the (strict) sanity check to fail:

```
$ eb EasyBuild-2.4.0.eb
...
== sanity checking...
== FAILED: Installation ended unsuccessfully (build directory: /tmp/vsc40003/easybuild/EasyBuild/2.4.0/dummy-dummy): build failed (first 300 chars): Failed to find pattern '^\./vsc_base' in /gpfs/scratch/apps/develop/software/EasyBuild/2.4.0/lib/python2.6/site-packages/easy-install.pth: import sys; sys.__plen = len(sys.path)
./easybuild_framework-2.4.0-py2.6.egg
./easybuild_easyblocks-2.4.0-py2.6.egg
./easybuild_easyconfigs-2.4.0-py2.6.egg
impor
```

Although `vsc-base` *is* being installed in the target installation prefix, it is not listed in the `easy_install.pth` file, because a more recent version is found on the OS.

```
$ ls /gpfs/scratch/apps/develop/software/EasyBuild/2.4.0/lib/python2.6/site-packages/    
easy-install.pth		      easybuild_easyconfigs-2.4.0-py2.6.egg  setuptools.pth  site.pyc
easybuild_easyblocks-2.4.0-py2.6.egg  easybuild_framework-2.4.0-py2.6.egg    site.py	     vsc_base-2.2.4-py2.6.egg
```

```
$ cat /gpfs/scratch/apps/develop/software/EasyBuild/2.4.0/lib/python2.6/site-packages/easy-install.pth 
import sys; sys.__plen = len(sys.path)
./easybuild_framework-2.4.0-py2.6.egg
./easybuild_easyblocks-2.4.0-py2.6.egg
./easybuild_easyconfigs-2.4.0-py2.6.egg
import sys; new=sys.path[sys.__plen:]; del sys.path[sys.__plen:]; p=getattr(sys,'__egginsert',0); sys.path[p:p]=new; sys.__egginsert = p+len(new)
```

From the `python setup.py install` output for `easybuild-framework`:

```
Processing dependencies for easybuild-framework==2.4.0
Searching for vsc-base==2.4.10
Best match: vsc-base 2.4.10
Removing vsc-base 2.2.4 from easy-install.pth file
Adding vsc-base 2.4.10 to easy-install.pth file
```

cc @wpoely86

PR #813, opened 2016-01-22
fix for vsc-base being picked up from OS in EB easyblock + remove setuptools.pth if it includes absolute paths

boegel:
This will help significantly with https://github.com/hpcugent/easybuild-framework/issues/1409, is a proper fix compared to the hack in #763, and should make #786 mostly unnecessary (it didn't really work anyway).


riccardomurri:
LGTM

boegel:
Going in, thanks for the review @riccardomurri!

PR #819, opened 2016-01-29
CMake-based counterpart for software that lakes a "make install" coun…

citibeth:
…terpart.  This is useful to convert easyconfigs that previously usd MakeCp, but where the underlying project has converted to CMake.

Usage example:

gtest-1.7.0-GCC-4.9.3.eb
========================
easyblock = 'CMakeMakeCp'

name = 'gtest'
version = '1.7.0'

homepage = 'https://code.google.com/p/googletest/'
description = "Google's framework for writing C++ tests on a variety of platforms"

toolchain = {'name': 'GCC', 'version': '4.9.3'}

sources = [SOURCE_ZIP]
source_urls = ['https://googletest.googlecode.com/files']

# Tells CMakeMakeCp to configure and build the project...
with_configure = True

builddependencies = [
    ('CMake', '3.4.1', '', ('GCC', '4.9.3')),
]

files_to_copy = [
	(['libgtest.a', 'libgtest_main.a'], 'lib'),
	'include'
]

sanity_check_paths={
    'files': ['lib/libgtest.a', 'lib/libgtest_main.a'],
    'dirs': ['include'],
}

moduleclass = 'devel'

### DIFF
--- a/easybuild/easyblocks/generic/cmakemakecp.py
+++ b/easybuild/easyblocks/generic/cmakemakecp.py
@@ -1,0 +1,122 @@
+##
+# Copyright 2013 the Cyprus Institute
+#
+# This file is part of EasyBuild,
+# originally created by the HPC team of Ghent University (http://ugent.be/hpc/en),
+# with support of Ghent University (http://ugent.be/hpc),
+# the Flemish Supercomputer Centre (VSC) (https://vscentrum.be/nl/en),
+# the Hercules foundation (http://www.herculesstichting.be/in_English)
+# and the Department of Economy, Science and Innovation (EWI) (http://www.ewi-vlaanderen.be/en).
+#
+# http://github.com/hpcugent/easybuild
+#
+# EasyBuild is free software: you can redistribute it and/or modify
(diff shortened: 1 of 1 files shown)


citibeth:
yes

easybuild/easyblocks/generic/cmakemakecp.py
```diff
@@ -0,0 +1,122 @@
...
+import os
+import shutil
+import glob
+
+from easybuild.easyblocks.generic.cmakemake import CMakeMake
+from easybuild.framework.easyconfig import BUILD, MANDATORY
+from easybuild.tools.build_log import EasyBuildError
+
+
+class CMakeMakeCp(CMakeMake):
```
boegel: rather than copy-pasting from `MakeCp`, please use multiple inheritance here; something like:

```python
class CMakeMakeCp(CMakeMake, MakeCp):
    ...
    def install_step(self):
        MakeCp.install_step(self)
```

boegel:
Jenkins: ok to test

PR #821, opened 2016-02-04
Add option to create the install dir prior to configure in CMakeMake

wpoely86:
If CMake uses external project that also must be installed in the final
install dir, it cannot be cleaned between the build and install step.
With the `create_installdir_before_configure` option set to True, the
install dir will be cleaned and create prior to running cmake.
`keeppreviousinstall` is set to True so that it doesn't get cleaned
again.


boegel:
Looks good. Out of curiosity: what's the use case for this?

boegel:
Jenkins: test this please

PR #822, opened 2016-02-04
HPL with no FFTW as dependency 

pforai:
This is actually not used by HPL at all.


gppezzi:
:+1:  Tested with CrayGNU-2015.11-XC on Dora

boegel:
Verified, good to go, thanks @pforai!

PR #823, opened 2016-02-04
Module swap for all deps of Cray TC components (REVIEW)

pforai:
This needs some serious refactoring, but produces the correct module file for our Cray toolchains. 

- Push things into a function that does the swaps
- Add LUA syntax support 
- Refactor variable names for clarity 
- Adjust comments for the same reason
- Explain rationale (why we need to do it like this)

edit: requires ~~https://github.com/hpcugent/easybuild-framework/pull/1609~~

### DIFF
--- a/easybuild/easyblocks/generic/craytoolchain.py
+++ b/easybuild/easyblocks/generic/craytoolchain.py
@@ -46,6 +46,8 @@ class CrayToolchain(Bundle):
         """
         prgenv_name, prgenv_mod = None, None
         unload_info = {}
+        swap_mod = ''
+
 
         # build dict with info for 'module swap' statements for dependencies,
         # i.e. a mapping of full module name to the module name to unload before loading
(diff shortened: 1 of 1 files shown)


boegel:
@pforai: proper support for `swap` implemented in https://github.com/hpcugent/easybuild-framework/pull/1609, can you please clean this up accordingly?

boegel:
not out of the woods yet, mentioned by @gppezzi: https://gist.github.com/gppezzi/7cb30d5661a6c4b4ff67

boegel:
Thoroughly tested with the reworked Cray toolchain definitions from https://github.com/hpcugent/easybuild-easyconfigs/pull/2554, with the easyconfigs from https://github.com/hpcugent/easybuild-easyconfigs/pull/2434 and https://github.com/hpcugent/easybuild-easyconfigs/pull/2435, good to go.

Thanks for your efforts on this @gppezzi and @pforai!

PR #824, opened 2016-02-05
The name of the PSI4 data dir changed

wpoely86:


easybuild/easyblocks/p/psi.py
```diff
@@ -154,7 +155,8 @@ def install_step(self):
         try:
             for subdir in ['obj', self.psi_srcdir]:
                 # copy symlinks as symlinks to work around broken symlinks
-                shutil.copytree(os.path.join(self.builddir, subdir), os.path.join(self.installdir, subdir), symlinks=True)
+                shutil.copytree(os.path.join(self.builddir, subdir), os.path.join(self.installdir, subdir),
+                                symlinks=True)
```
boegel: why the wrap? looked fine?

easybuild/easyblocks/p/psi.py
```diff
@@ -175,12 +177,16 @@ def sanity_check_step(self):
...
         super(EB_PSI, self).sanity_check_step(custom_paths=custom_paths)
 
     def make_module_extra(self):
         """Custom variables for PSI module."""
         txt = super(EB_PSI, self).make_module_extra()
-        txt += self.module_generator.set_environment('PSI4DATADIR', os.path.join(self.installdir, 'share', 'psi'))
+        psi4datadir = glob.glob(os.path.join(self.installdir, 'share', 'psi*'))
+        if len(psi4datadir) != 1:
+            raise EasyBuildError("Could not determine the PSI4 data dir, there are multiple possibilities: %s"
+                                 % psi4datadir)
```
boegel: don't wrap, and use `, psi4datadir`, no need for the `% "..."`

easybuild/easyblocks/p/psi.py
```diff
@@ -175,12 +177,16 @@ def sanity_check_step(self):
...
+        txt += self.module_generator.set_environment('PSI4DATADIR', psi4datadir[0])
```
boegel: I would do this:

```python
if len(psi4datadir) == 1:
    txt += self.module_generator.set_environment('PSI4DATADIR', psi4datadir[0])
else:
    raise EasyBuildError("...", psi4datadir)
```

Also, there could be no hits, so your "multiple possibilities" could be wrong/weird :)

easybuild/easyblocks/p/psi.py
wpoely86: 120 chars

easybuild/easyblocks/p/psi.py
wpoely86: 120 chars

easybuild/easyblocks/p/psi.py
wpoely86: sanity check makes sure that there is at least one

boegel: OK, sure, we we like to keep if conditions 'positive'

and you're really doing an if-else kind of thing here, so use `if`/`else`

easybuild/easyblocks/p/psi.py
boegel: ok

easybuild/easyblocks/p/psi.py
boegel: ok, but use `, psi4datadir)`

easybuild/easyblocks/p/psi.py
wpoely86: I really don't see why this would be better...

boegel: It's mostly about readability.

wpoely86: yes, so? My is: if not 1, stop right here. I don't see how it can get any more readable?

boegel:
Going in, thanks @wpoely86!

PR #826, opened 2016-02-05
add easyblock for ADF (REVIEW)

boegel:


easybuild/easyblocks/a/adf.py
```diff
@@ -0,0 +1,95 @@
...
+
+    def build_step(self):
+        """No separate custom build procedure for ADF, see install step."""
+        pass
+
+    def install_step(self):
+        """Custom install procedure for ADF."""
+
+        # this shouldn't be needed, is init.sh shipped in the wrong place?!
+        shutil.copy2(os.path.join('Install', 'init.sh'), os.path.join('bin', 'init.sh'))
```
boegel: unclear why this is needed

easybuild/easyblocks/a/adf.py
```diff
@@ -0,0 +1,95 @@
...
+        # strip off 'adf<version>' part to avoid having everything in a subdirectory
+        self.cfg['unpack_options'] = "--strip-components=1"
+        super(EB_ADF, self).extract_step()
+
+    def configure_step(self):
+        """Custom configuration procedure for ADF."""
+
+        env.setvar('ADFHOME', self.installdir)
+        env.setvar('ADFBIN', os.path.join(self.installdir, 'bin'))
+        env.setvar('ADFRESOURCES', os.path.join(self.installdir, 'atomicdata'))
```
boegel: this seems to be wrong `atomicdata` isn't there?

easybuild/easyblocks/a/adf.py
```diff
@@ -0,0 +1,95 @@
...
+        shutil.copy2(os.path.join('Install', 'init.sh'), os.path.join('bin', 'init.sh'))
+
+        cmd = "./bin/foray -j %d" % self.cfg['parallel']
+        run_cmd(cmd, log_all=True, simple=True, log_ok=True)
+
+    def sanity_check_step(self):
+        """Custom sanity check for ADF."""
+        custom_paths = {
+            'files': ['bin/adf'],
+            'dirs': ['atomicdata', 'examples'],
```
boegel: currently fails, `atomicdata` and `example` are not there like they are in the binary version...

easybuild/easyblocks/a/adf.py
boegel: ah, seems like you need to unpack both the sources and binary version, and then recompile

http://www.scm.com/Doc/Doc2014/Install/Installation/page5.html

that explains why some stuff is missing

easybuild/easyblocks/a/adf.py
boegel: fixed in easyconfig, both sources and binary version are available to recompile from source >_<

boegel:
@wpoely86: please review?

easybuild/easyblocks/a/adf.py
```diff
@@ -0,0 +1,100 @@
...
+# EasyBuild is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with EasyBuild.  If not, see <http://www.gnu.org/licenses/>.
+##
+"""
+EasyBuild support for building and installing ADF, implemented as an easyblock
```
wpoely86: `@author ` tag

wpoely86:
lgtm

boegel:

boegel:
Tested with https://github.com/hpcugent/easybuild-easyconfigs/pull/2451, works as intended, so good to go.

Issue #83, opened 2013-01-18
Review external guidelines for HPC software building


boegel:
(originally by @fgeorgatos)

this issue does not mean that guidelines have to be followed at 100% but,
there should be some good understanding if something obvious has to be adhered to:
http://www.hpcadvisorycouncil.com/subgroups_hpc_works.php

PR #832, opened 2016-02-18
New EasyBlock for CPMD software

valtandor:


boegel:
@wpoely86: up for giving this a review?

@valtandor: please take a look at http://easybuild.readthedocs.org/en/latest/Extended_dry_run.html#runtime-patching-of-files-apply-regex-substitutions, to replace the `fileinput` stuff

boegel:
Jenkins: ok to test

valtandor:
@boegel: Will do, though it may be a few days.

easybuild/easyblocks/c/cpmd.py
```diff
@@ -0,0 +1,208 @@
+##
+# Copyright 2013 Ghent University
```
wpoely86: 2016 :wink:

wpoely86: And if it's yours, you can claim copyright.

easybuild/easyblocks/c/cpmd.py
```diff
@@ -0,0 +1,208 @@
...
+    def configure_step(self, cmd_prefix=''):
+        """
+        Configure step
+        """
+
+        config_file_candidates = []
+        config_file_prefix = os.path.join(self.builddir, "CPMD", "configure")
+
+        # Work out a starting configuration file if one is not supplied in the easyconfig
+        if self.cfg['base_configuration']:
```
wpoely86: check if this file exists

easybuild/easyblocks/c/cpmd.py
```diff
@@ -0,0 +1,208 @@
...
+            config_file_base = self.cfg['base_configuration']
+        else:
+            os_type_mappings = {
+                                "LINUX"  : "LINUX",
+                                "DARWIN" : "MACOSX",
+                                }
```
wpoely86: fix indenting

fgeorgatos: that is, 4-space indenting...

fgeorgatos:
Fished this from #83, in case it helps (say, for alternative toolchain/MPI stacks' builds):
http://www.hpcadvisorycouncil.com/pdf/CPMD_Best_Practices.pdf
We can see there somewhere a `-DPOINTER` and some more options, you might wish to inspect that!

easybuild/easyblocks/c/cpmd.py
```diff
@@ -0,0 +1,208 @@
...
+
+        (out, _) = run_cmd(cmd, log_all=True, simple=False)
+
+        return out
+
+    def build_step(self):
+
+        os.chdir(self.installdir)
+        """
+        Make some changes to files in order to make the build process more EasyBuild-friendly
```
wpoely86: put this below the function

easybuild/easyblocks/c/cpmd.py
```diff
@@ -0,0 +1,208 @@
...
+
+        # enable OpenMP support if desired
+        if self.toolchain.options.get('openmp', None):
+            options.append("-omp")
+
+        # This "option" has to come last as it's the chief argument, coming after
+        # all flags and so forth.
+        options.append(selected_base_config)
+
+        cmd = "%(preconfigopts)s %(cmd_prefix)s./configure.sh %(prefix_opt)s%(installdir)s %(configopts)s" % {
```
wpoely86: why copy this from the super class?

valtandor: Because the superclass configure_step method also does a bunch of other things I don't want.

easybuild/easyblocks/c/cpmd.py
valtandor: That is done down at line 119 and following. Unless you think that a check should also be put in place up here, with a note that the configuration file could not be found and has been overridden? I'm wary of that; in my opinion, it's better to fail loudly so that the user knows his preferences couldn't be carried out.

ocaisa:
There's something wrong with the last regex commit

easybuild/easyblocks/c/cpmd.py
```diff
@@ -0,0 +1,208 @@
...
+                ar_exe = "xiar -ruv"
+                apply_regex_substitutions(selected_full_config, [
+                    (r"^(\s*AR)=.*", r"\1='%s'" % ar_exe)
+                ]
+            apply_regex_substitutions(selected_full_config, [
+                # Better to get CC and FC from the EasyBuild environment in this instance
+                (r"^(\s*CC=.*)", r"#\1"),
+                (r"^(\s*FC=.*)", r"#\1"),
+                (r"^(\s*LD)=.*", r"\1='$(FC)'"),
+            ]
```
wpoely86: missing closing parenthese?

valtandor:
@boegel @ocaisa Apologies for the delay. I think it should be fixed now.

mboisson:
Can this be merged ? Looking to install CPMD on our server.

boegel:
@mboisson This needs to be thoroughly tested by others (using https://github.com/hpcugent/easybuild-easyconfigs/pull/2509), and reviewed, before it can go in...

damianam:
There are things like `-O2 -axAVX -openmp` leaking into the `FFLAGS` defined in the `Makefile`. These are not defined by EB and can cause issues (`-openmp` will be deprecated soon in intel compilers, `-axAVX` conflicts with `-xHosts`). Moreover, the code currently does not compile with Intel 2017, due to the lack of `-fpp`. Additionally, the flags passed to `configure.sh` (`-FFLAGS`, `-CFLAGS`, etc) are ignored (the generated `Makefile` prepends them with `#`).

It looks to me like handling this whole thing with regular expressions can become troublesome and difficult to maintain. I'd suggest to fix it with a combination of a patch, that will add a configuration file tailored to EB, and a smaller easyblock that takes care of filling the new configuration with EB parameters (`xFLAGS` and such) and allowing the user to pass its own flags, fixing the generated `Makefile`

boegel:
@pescobar Took an alternate approach that doesn't require a custom easyblock, see https://github.com/easybuilders/easybuild-easyconfigs/pull/5282.

@valtandor Thoughts on that, anything missing there that you took care of via a custom easyblock for CPMD?

mboisson:
Why a need for a separate easyblock ? We install CPMD with the MakeCp easyblock. 

https://github.com/ComputeCanada/easybuild-easyconfigs/blob/computecanada-master/easybuild/easyconfigs/c/CPMD/CPMD-3.17.1-imkl-intel-2016b-FFTW.eb

PR #833, opened 2016-02-18
Easyblock for LAMMPS molecular dynamics software

valtandor:
Note that LAMMPS includes several optional libraries that have to be compiled via their own steps. I went for broke and put in a custom rule for each such package. I'm open to suggestions for improvement if this is believed to be suboptimal.


boegel:
@valtandor: are you aware of https://github.com/hpcugent/easybuild-easyblocks/pull/594 ?

valtandor:
@boegel: I am now. Ugh - this could be an ugly merge.

easybuild/easyblocks/l/lammps.py
```diff
@@ -0,0 +1,802 @@
...
+
+        # Can't be None as needs to be appended to
+        out = ''
+        src_dir = os.path.join(self.cfg['start_dir'], 'src')
+
+        if self.toolchain.options.get('openmp', None):
+            out += self.build_user_omp(verbose, path)
+
+        if self.cfg['use_asphere']:
+            out += self.build_asphere(verbose, path)
```
boegel: this can be collapsed with some more Python magic...

```python
for comp in components:
    if self.cfg['use_%s' % comp]:
        getattr(self, 'build_%s' % comp)(verbose, path)
```

(with some more checking)

boegel:
@valtandor: well, it never got finished, and maybe this is better, I don't know...

Are you up for taking a look at it, and get back with an idea of whether it makes sense to try and merge both together?

boegel:
Jenkins: ok to test

gppezzi:
@boegel @valtandor we're planning to work on a GPU-enabled version of LAMMPS for Cray, should we use this PR? or https://github.com/hpcugent/easybuild-easyblocks/pull/594 ?

boegel:
@gppezzi to be honest, I'm not quite sure, LAMMPS is a weird beast...

The two different easyblocks take a totally different approach. This one here is very verbose and tries to iterate all possible packages, which seems like a recipe for a maintenance nightmare in the future.

Mine in #594 allows to list which packages to install, and then iterates over that list. But it's probably missing stuff (cfr. `build_gpu` and `build_meam` here).

I think in the long run #594 is a better approach, but that easyblocks definitely needs work...

The one in #594 doesn't support the `gpu` target yet (it's hard filtered out), and I'm not sure whether what this one here does for `meam` should be ported to it or not.

I'm open to suggestions/patches...

valtandor:
I have no objection to an approach other than listing every single package. Indeed, I would have preferred not to do that, but I'm by no means a Python expert yet, so I was working with the methods I knew.

valtandor:
@boegel This may be better. I don't know entirely how it will work with the generic library compilation method in https://github.com/hpcugent/easybuild-easyblocks/pull/594, but then it seems that there is often a case to tweak the library configurations anyway. I can also see that in https://github.com/hpcugent/easybuild-easyblocks/pull/594 you built all the libraries whether they would be needed or not. Perhaps you can comment on what would be most desirable.

boegel:
Closing this, superseded by https://github.com/easybuilders/easybuild-easyblocks/pull/1964.

@valtandor Sorry for not following up on this more closely...

PR #837, opened 2016-02-24
Fixed license checking and CPATH definition

damianam:
This PR does 2 things:

-Eliminates the `CPATH` definition (see https://github.com/hpcugent/easybuild-easyblocks/issues/830)
-Improves the license checking to be compatible with license servers. It does it in the same was as `IntelBase.py`. In the future, some of the functionality in these 2 places should be moved to the framework (see https://lists.ugent.be/wws/arc/easybuild/2016-02/msg00028.html)

### DIFF
--- a/easybuild/easyblocks/p/pgi.py
+++ b/easybuild/easyblocks/p/pgi.py
@@ -41,6 +41,7 @@ import easybuild.tools.environment as env
 from easybuild.framework.easyblock import EasyBlock
 from easybuild.framework.easyconfig import CUSTOM
 from easybuild.tools.build_log import EasyBuildError
+from easybuild.tools.filetools import find_flexlm_license
 from easybuild.tools.run import run_cmd
 
 class EB_PGI(EasyBlock):
--- a/easybuild/easyblocks/p/pgi.py
+++ b/easybuild/easyblocks/p/pgi.py
@@ -84,7 +84,6 @@ class EB_PGI(EasyBlock):
                 self.log.info("Using PGI license specifications from %s: %s", self.license_env_var, lic_specs)
 
             self.license_file = os.pathsep.join(lic_specs)
-            self.cfg['license_file'] = self.license_file
             env.setvar(self.license_env_var, self.license_file)
 
         else:
--- a/easybuild/easyblocks/p/pgi.py
+++ b/easybuild/easyblocks/p/pgi.py
@@ -1,5 +1,6 @@
 ##
-# Copyright 2015-2015 Bart Oldeman
+# Copyright 2015-2016 Bart Oldeman
+# Copyright 2016-2016 Forschungszentrum Juelich
 #
 # This file is triple-licensed under GPLv2 (see below), MIT, and
 # BSD three-clause licenses.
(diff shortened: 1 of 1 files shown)


boegel:
Jenkins: ok to test

boegel:
@damianam: needs a sync with `develop` (already)?

easybuild/easyblocks/p/pgi.py
```diff
@@ -58,16 +61,88 @@ def extra_options():
...
+        self.license_file = 'UNKNOWN'
+        self.license_env_var = 'UNKNOWN' # Probably not really necessary for PGI
+
         self.install_subdir = os.path.join('linux86-64', self.version)
 
     def configure_step(self):
         """
-        Dummy configure method, just a license check
+        Handle license file. It mimics intelbase.py behaviour. It should be merged into the framekwork
+        at some point
```
boegel: How much tweaking did you need to do here?

I'm not a big fan of including it like this, it's just a copy-paste, and then someone may start to fix issues here, but not in `intelbase.py`, etc. etc.

Now is the time to get this into framework...

easybuild/easyblocks/p/pgi.py
```diff
@@ -115,11 +190,12 @@ def make_module_req_guess(self):
         dirs = super(EB_PGI, self).make_module_req_guess()
         for key in dirs:
             dirs[key] = [os.path.join(self.install_subdir, d) for d in dirs[key]]
+        dirs.update({'CPATH': ''})
```
boegel: just chuck the `CPATH` entry entirely, and explain why

```python
dirs = super(EB_PGI, self).make_module_req_guess()
for key in dirs:
    ...

# $CPATH should not be defined in module for PGI, it causes problems
# cfr. https://github.com/hpcugent/easybuild-easyblocks/issues/830
if 'CPATH' in dirs:
    self.log.info("Removing $CPATH entry: %s", dirs['CPATH']
    del dirs['CPATH']
```

easybuild/easyblocks/p/pgi.py
damianam: I haven't tweak it much. Just removed the intel bits. You are absolutely right, it is basically a copy&paste. The reason for it is to make easier the transition from this to the new state with this functionality present in the framework. Meaning that all this should become a couple of lines in the future. I guess the options are:

1. Stick to a buggy easyblock until the new piece of the framework gets done and fix it then.
2. Accept a copy&paste to fix the easyblock ASAP and have a trivial transition once the new piece of the framework is done.

1 is more elegant but keeps things unsolved for an undefined amount of time. 2 is a quick and ugly hack, but that works well until we have what we need in the framework.

I'd suggest to go for 1 if the changes in the framework are done before the next release, or go for 2 otherwise.

boegel: @damianam: here you go https://github.com/hpcugent/easybuild-framework/pull/1633 :)

easybuild/easyblocks/p/pgi.py
```diff
@@ -115,11 +190,18 @@ def make_module_req_guess(self):
         dirs = super(EB_PGI, self).make_module_req_guess()
         for key in dirs:
             dirs[key] = [os.path.join(self.install_subdir, d) for d in dirs[key]]
+
+        # $CPATH should not be defined in module for PGI, it causes problems
+        # cfr. https://github.com/hpcugent/easybuild-easyblocks/issues/830
+        if 'CPATH' in dirs:
+            self.log.info("Removing $CPATH entry: %s", dirs['CPATH']
```
boegel: missing `)` at the end

Jenkins knoooooows if you don't retest ;)

boegel:
Rereviewed, tested and approved, so going in.

Thanks @damianam!

bartoldeman:
Thanks, this all looks great to me! The CPATH issue was an oversight, not deliberate.

PR #840, opened 2016-02-25
fix check for non-empty lib dirs in PythonPackage easyblock

boegel:
`os.listdir` returns only the filenames, not the full paths...

cc @wpoely86


wpoely86:
lgtm

boegel:
Going in, I'll see if we can come up with a test dir this...

boegel:
@wpoely86: test implemented in #841

PR #841, opened 2016-02-25
add unit test to check module file generated by PythonPackage easyblock

boegel:
test that checks for the bug fixed in #840

### DIFF
--- a/test/easyblocks/module.py
+++ b/test/easyblocks/module.py
@@ -46,7 +46,7 @@ from easybuild.framework.easyconfig import MANDATORY
 from easybuild.framework.easyconfig.easyconfig import EasyConfig, get_easyblock_class
 from easybuild.framework.easyconfig.tools import get_paths_for
 from easybuild.tools import config
-from easybuild.tools.filetools import write_file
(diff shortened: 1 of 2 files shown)


boegel:
@wpoely86: please review?

test/easyblocks/module.py
```diff
@@ -90,6 +90,59 @@ def setUp(self):
...
+
+        self.assertTrue(TMPDIR in app.installdir)
+        self.assertTrue(TMPDIR in app.installdir_mod)
+
+        modtxt = None
+        for cand_mod_filename in ['3.14', '3.14.lua']:
+            full_modpath = os.path.join(app.installdir_mod, 'testpypkg', cand_mod_filename)
+            if os.path.exists(full_modpath):
+                modtxt = read_file(full_modpath)
+                break
```
wpoely86: and if both tcl and lua are generated?

boegel: that's not supported yet; and we don't really care here, that's not what we care about in this test?

boegel:
Thanks for the review @wpoely86, this is a good base to have some sort of tests (beyond init) for all easyblocks...

PR #842, opened 2016-02-26
Libxml2 easyblock just checks for one of the possibly paths in the sanity check

damianam:
This fixes that issue, and now the sanity check looks into both `lib `and `lib64`. This was a problem for installing libxml2 in GCCcore, since it looked into `lib`, but the python libraries were installed in `lib64`

### DIFF
--- a/easybuild/easyblocks/l/libxml2.py
+++ b/easybuild/easyblocks/l/libxml2.py
@@ -95,10 +95,9 @@ class EB_libxml2(ConfigureMake, PythonPackage):
         """Custom sanity check for libxml2"""
         shlib_ext = get_shared_lib_ext()
         pyfiles = ['libxml2mod.%s' % shlib_ext, 'libxml2.py', 'drv_libxml2.py']
-        #[os.path.join(self.pylibdir, x) for x in pyfiles],
         custom_paths = {
             'files': ["lib/libxml2.a", "lib/libxml2.%s" % shlib_ext] +
-                     [tuple([(os.path.join(y,x)) for y in self.all_pylibdirs]) for x in pyfiles],
+                     [tuple([(os.path.join(d, f)) for d in self.all_pylibdirs]) for f in pyfiles],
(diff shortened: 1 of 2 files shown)


boegel:
Jenkins: ok to test

easybuild/easyblocks/l/libxml2.py
```diff
@@ -95,9 +95,10 @@ def sanity_check_step(self):
         """Custom sanity check for libxml2"""
         shlib_ext = get_shared_lib_ext()
         pyfiles = ['libxml2mod.%s' % shlib_ext, 'libxml2.py', 'drv_libxml2.py']
+        #[os.path.join(self.pylibdir, x) for x in pyfiles],
```
boegel: please remove this?

easybuild/easyblocks/l/libxml2.py
```diff
@@ -95,9 +95,10 @@ def sanity_check_step(self):
...
         custom_paths = {
             'files': ["lib/libxml2.a", "lib/libxml2.%s" % shlib_ext] +
-                     [os.path.join(self.pylibdir, x) for x in pyfiles],
-            'dirs': ["bin", self.pylibdir, "include/libxml2/libxml"],
+                     [tuple([(os.path.join(y,x)) for y in self.all_pylibdirs]) for x in pyfiles],
```
boegel: tiny style remark: space after `,` (in `(y,x)`)

and while you're at it, why not use more descriptive letter: `f` for file, `d` for directory; makes it easier to read imho

```python
[tuple([(os.path.join(d, f)) for d in self.all_pylibdirs]) for f in pyfiles],
```

boegel:
@damianam: looks good, but it would be nice to also include https://github.com/damianam/easybuild-easyblocks/pull/3 for usable `-x` output

damianam:
@boegel makes sense, it has been merged already.

boegel:
tested on existing libxml2 easyconfigs, (still) works like a charm, going in, thanks @damianam!

PR #818, opened 2016-01-29
MPICH derivatives easyblocks

damianam:
See https://github.com/hpcugent/easybuild-easyconfigs/issues/2334

PR #844, opened 2016-02-29
Update easyblocks for MPICH and MVAPICH2

besserox:
MVAPICH2 library names have changed since version 2.1.
These changes in the library names are inherited from MPICH 3.1.1 and above.
cf http://git.mpich.org/mpich.git/blob_plain/v3.1.1:/CHANGES

### DIFF
--- a/easybuild/easyblocks/m/mvapich2.py
+++ b/easybuild/easyblocks/m/mvapich2.py
@@ -34,6 +34,7 @@ EasyBuild support for building and installing the MVAPICH2 MPI library, implemen
 """
 
 import os
+from distutils.version import LooseVersion
 
 import easybuild.tools.environment as env
 from easybuild.easyblocks.mpich import EB_MPICH
@@ -47,6 +48,13 @@ class EB_MVAPICH2(EB_MPICH):
     Support for building the MVAPICH2 MPI library.
     - some compiler dependent configure options
     """
+
+    def __init__(self, *args, **kwargs):
+        """Custom constructor for EB_MVAPICH2 easyblock, initialize custom class parameters."""
+        super(EB_MVAPICH2, self).__init__(*args, **kwargs)        
+        # MVAPICH2 >=2.1 depends on MPICH >=3.1.1, which uses new library names
+        # cf http://git.mpich.org/mpich.git/blob_plain/v3.1.1:/CHANGES
+        self.use_new_libnames = LooseVersion(self.version) >= LooseVersion('2.1')
 
     @staticmethod
     def extra_options():
@@ -115,6 +123,6 @@ class EB_MVAPICH2(EB_MPICH):
         Custom sanity check for MVAPICH2
         """
         custom_paths = {
-            'files': ['bin/%s' % x for x in ['mpiexec.hydra']] 
+            'files': ['bin/%s' % x for x in ['mpiexec.mpirun_rsh']] 
         }
         super(EB_MVAPICH2, self).sanity_check_step(custom_paths=custom_paths)
--- a/easybuild/easyblocks/m/mvapich2.py
+++ b/easybuild/easyblocks/m/mvapich2.py
@@ -49,12 +49,11 @@ class EB_MVAPICH2(EB_MPICH):
     - some compiler dependent configure options
     """
 
-    def __init__(self, *args, **kwargs):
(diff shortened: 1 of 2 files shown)


boegel:
Jenkins: ok to test

boegel:
hmm, this is going to conflict nicely with #818...

@damianam: thoughts?

besserox:
Thanks for pointing this out.
I will test #818 and see if I can add these sanity check on top of it.

damianam:
@boegel I still have to get the list of files to be checked by MVAPICH2, MPICH and ParaStationMPI. I think it makes sense to merge this first, and then I will update my PR accordingly. Does it make sense?

easybuild/easyblocks/m/mvapich2.py
```diff
@@ -134,10 +135,14 @@ def sanity_check_step(self):
...
-            'dirs': ['include'],
-        }
+        binaries = ['bin/%s' % x for x in ['mpicc', 'mpicxx', 'mpif77', 'mpif90', 'mpiexec.hydra']]
+        directories = ['include']
+        if LooseVersion(self.version) < LooseVersion('2.1'):
+            libraries = ['lib/lib%s' % y for x in ['fmpich', 'mpichcxx', 'mpichf90', 'mpich', 'mpl', 'opa']
+                                      for y in ['%s.a' % x, '%s.%s' % (x, shlib_ext)]]
+        else:
+            libraries = ['lib/lib%s' % y for x in ['mpi', 'mpicxx', 'mpifort']
+                                      for y in ['%s.a' % x, '%s.%s' % (x, shlib_ext)]]
```
boegel: repeating the list comprehension feels a bit bloated, since only the library names are different

cleaned up version proposed in https://github.com/besserox/easybuild-easyblocks/pull/2

boegel:
@damianam: ok, makes sense, thanks for the feedback

boegel:
I tested this easyblock (with https://github.com/besserox/easybuild-easyblocks/pull/2 included) with all existing MVAPICH2 easyconfigs + the new ones available from https://github.com/hpcugent/easybuild-easyconfigs/pull/2589, works like a charm.

Let's get this merged @besserox?

besserox:
I have updated this branch:
- #818 is merged into my branch. Thanks @damianam !
- The sanity checks have been updated and moved to the MPICH easyblock
-  I have applied the changes suggested in https://github.com/besserox/easybuild-easyblocks/pull/2 manually (because now they apply to the MPICH easyblock)
- I have moved some configuration options from MVAPICH to MPICH 
- I have added a `use_new_libnames()` member function in the MPICH easyblock to tell if the new library names are used or not. This function then can be overridden in derived class, eg for MVAPICH2.

boegel:
@besserox: Thank you for the effort!

Frankly, it may have been easier to just get your original PR merged, and then rework #818 based on that, but we can make this work too... :-)

The thing is that @damianam is working on some changes on top of #818 (mostly w.r.t. the sanity checks), so he'll have to look into that on top of this PR now. Please ack @damianam?

W.r.t. `use_new_libnames`, I would take a different approach, rather than introduce a custom method. Let me get back to you with a suggestion there.

damianam:
@boegel yes, I was working on that and had it mostly done. But it doesn't matter (see below).

In my PR we used the MPICH easyblock for `psmpi`. The approach of this PR affects in these points:

* `psmpi` doesn't build `mpiexec` and similar, so the sanity check would fail.
* `psmpi` sets its own options for MPICH (its configure calls the MPICH configure), so the default MPICH options are not understood by the configure script of `psmpi`.

As it is now it won't work for us or any other `psmpi` user. We can, however, create a `psmpi`-specific easyblock. It might actually be the best solution. If we go down that way it would be helpful to separate the variable redefinition in a separate function, so we can override the `configure_step` without duplicating that piece of code. Thoughts (cc @ocaisa)?

boegel:
@besserox: please take a look at https://github.com/besserox/easybuild-easyblocks/pull/3 that cleans this up a bit

boegel:
(re)tested thoroughly with existing MVAPICH2 easyconfigs + with the MPICH/MVAPICH2 easyconfigs from https://github.com/hpcugent/easybuild-easyconfigs/pull/2589 (see test report at https://github.com/hpcugent/easybuild-easyconfigs/pull/2589#issuecomment-191173737)

So, going in!

Thanks a lot for your efforts on this @damianam and @besserox!

PR #848, opened 2016-03-02
H5py Cray with mpi support

gppezzi:
I couldn't figure out how to do this from the easyconfig file, so I created this easyblock.

### DIFF
--- a/easybuild/easyblocks/h/h5py.py
+++ b/easybuild/easyblocks/h/h5py.py
@@ -1,0 +1,15 @@
+from easybuild.easyblocks.generic.pythonpackage import PythonPackage
+from easybuild.tools.modules import get_software_root
+from easybuild.tools.run import run_cmd
+
+class h5py(PythonPackage):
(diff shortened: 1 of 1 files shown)


boegel:
@gppezzi:

```python
prebuildopts = "python setup.py configure --mpi"
```

like we do in the existing h5py 2.5.0 easyconfigs (e.g. https://github.com/hpcugent/easybuild-easyconfigs/blob/master/easybuild/easyconfigs/h/h5py/h5py-2.5.0-intel-2015b-Python-2.7.10-HDF5-1.8.15-patch1.eb)

Note also the extra `--hdf5` option in there...

nice job on totally skipping the sanity check btw ;)

gppezzi:
This was probably my first easyblock, i'm happy to delete it now that it was reviewed.

Anyway, I'm closing it since it is not needed anymore (fixed on https://github.com/hpcugent/easybuild-easyconfigs/pull/2613)

PR #849, opened 2016-03-02
Cray support for boost

gppezzi:
This easyblock is a tweaked version of boost.py for supporting Cray.

I don't know if it is preferable to include Cray support on the original one (but I don't want to break existing builds...)


ocaisa:
Wouldn't you be better off inheriting from the existing boost easyblock and only changing what you need to...or building the necessary logic into the existing boost easyblock?

gppezzi:
@ocaisa sure, but I'd like to hear what's the recommended way before tackling that...

pforai:
Just FIY: This block is based on my PR that I submitted some time ago and this has the problem that it expects aprun to be available on the machine where you build things.

boegel:
+1 for deriving from existing Boost easyblock, (way) too much copy-pasting now...

gppezzi:
@ocaisa  @boegel  @pforai  @lucamar I did some clean up, can you please review the new version?

easybuild/easyblocks/b/boostcray.py
```diff
@@ -0,0 +1,80 @@
...
+
+            self.toolchain.options['usempi'] = True
+            # configure the boost mpi module
+            # http://www.boost.org/doc/libs/1_47_0/doc/html/mpi/getting_started.html
+            # let Boost.Build know to look here for the config file
+            f = open('user-config.jam', 'a')
+            self.log.info("PRGENV_MODULE_NAME_SUFFIX is %s " % (self.toolchain.PRGENV_MODULE_NAME_SUFFIX))
+#	    self.log.info("TC_CONSTANT_CRAYPE is %s " % (self.toolchain.TC_CONSTANT_CRAYPE))
+            #if self.toolchain.TC_CONSTANT_CRAYPE in [toolchain.CRAYPE+'_GNU', toolchain.CRAYPE+'_Intel'] :
+            if self.toolchain.PRGENV_MODULE_NAME_SUFFIX in ['gnu', 'intel'] :
```
pforai: This would need to be a bit more specific and should also contain 'cce' and 'pgi' probably.

easybuild/easyblocks/b/boostcray.py
```diff
@@ -0,0 +1,80 @@
...
+                'using gcc ',
+		': %s' %(craygccversion),
+                ': CC ',
+                ': <compileflags>-I$(CRAY_MPICH2_DIR)/include ',
+                '  <linkflags>-L$(CRAY_MPICH2_DIR)/lib \ ',
+                '; ',
+                'using mpi ',
+                ': CC ',
+                ': <find-shared-library>mpich ',
+                ': aprun -n \ ',
```
pforai: We should warn the user that this might call aprun on the compilation machine

boegel:
Any chance to just enhance the existing `boost.py`?

Do we have a good way of checking whether we're running `eb` on a Cray system yet?

Or does `self.toolchain.name.startswith('Cray')` suffice (since it's really only if we're using the `Cray*` toolchains that we need to take special care; you can also build with `foss` on a Cray system but there's nothing really special about that imho).

gppezzi:
@boegel well I dont feel confident enough to modify the original easyblock without breaking other builds (meaning that I don't have much time for testing now, I'm just bulk pushing things that were waiting for the new cray toolchains).

But if that boils down to adding an extra test to check if TC is Cray*, then I think I can do it ;) I'll start with `self.toolchain.name.startswith('Cray')`

boegel:
@gppezzi we can help out with testing and making sure you're not breaking backwards compatibility;

It's good that you're pushing out the things you've worked on, and are willing to take the effort to contribute back, thanks for that!

easybuild/easyblocks/b/boostcray.py
gppezzi: @pforai  I've added cray to the list but since we check for a cray toolchain (and I'm not planning to fix boost on -cray or -intel right now), maybe I'll just drop this test for the moment?  

unless someone volunteers to work on the intel/cray boost support...

easybuild/easyblocks/b/boostcray.py
gppezzi: warn on screen or on log file?

gppezzi:
@boegel @pforai I've integrated the changes into boost.py, it seems to build, can you please review?

thx

easybuild/easyblocks/b/boost.py
```diff
@@ -121,9 +121,36 @@ def configure_step(self):
...
+                ': <find-shared-library>mpich ',
+                ': aprun -n \ ',
+                ';',
+                '',
+                ])
+                f.write(config)
+            else:
+                f.write("using mpi : %s ;" % os.getenv("MPICXX"))
 
+            f.close()
```
boegel: please replace this with:

```python
write_file('user-config.jam', txt, append=True)
```

and construct `txt` above

easybuild/easyblocks/b/boost.py
```diff
@@ -121,9 +121,36 @@ def configure_step(self):
             # http://www.boost.org/doc/libs/1_47_0/doc/html/mpi/getting_started.html
             # let Boost.Build know to look here for the config file
             f = open('user-config.jam', 'a')
-            f.write("using mpi : %s ;" % os.getenv("MPICXX"))
-            f.close()
+ 
+            # Check if using a Cray toolchain and configure MPI accordingly
+            if (self.toolchain.name.startswith('Cray') and 
+                self.toolchain.PRGENV_MODULE_NAME_SUFFIX in ['gnu', 'intel', 'cray']):
```
boegel: I think you need to split this up?

what if the toolchain name starts with `Cray`, but it's not a known PrgEnv suffix?

easybuild/easyblocks/b/boost.py
```diff
@@ -121,9 +121,36 @@ def configure_step(self):
...
+                self.log.info("PRGENV_MODULE_NAME_SUFFIX is %s " % (self.toolchain.PRGENV_MODULE_NAME_SUFFIX))
```
boegel: `log.debug`?

also, why is this relevant here? You can probably find this somewhere else in the log, and `PRGENV_MODULE_NAME_SUFFIX` isn't used at all below?

easybuild/easyblocks/b/boost.py
```diff
@@ -121,9 +121,36 @@ def configure_step(self):
...
+                craympichdir=os.getenv('CRAY_MPICH2_DIR')
+                craygccversion=os.getenv('GCC_VERSION')
```
boegel: What if we're not dealing with `PrgEnv-gnu`? Do we also want to grab the value of `$GCC_VERSION` then?

also: style remark: spaces around `=`

easybuild/easyblocks/b/boost.py
```diff
@@ -121,9 +121,36 @@ def configure_step(self):
...
+                f = open('user-config.jam','a')
+                config = '\n'.join([    
+                'import os ; ',
+                'local CRAY_MPICH2_DIR =  %s ;' %(craympichdir),
```
boegel: drop the `(`, `)`, spaces around `%`

easybuild/easyblocks/b/boost.py
```diff
@@ -121,9 +121,36 @@ def configure_step(self):
...
+ 
+            # Check if using a Cray toolchain and configure MPI accordingly
+            if (self.toolchain.name.startswith('Cray') and 
+                self.toolchain.PRGENV_MODULE_NAME_SUFFIX in ['gnu', 'intel', 'cray']):
+                self.log.info("PRGENV_MODULE_NAME_SUFFIX is %s " % (self.toolchain.PRGENV_MODULE_NAME_SUFFIX))
+                craympichdir=os.getenv('CRAY_MPICH2_DIR')
+                craygccversion=os.getenv('GCC_VERSION')
+                f = open('user-config.jam','a')
+                config = '\n'.join([    
+                'import os ; ',
```
boegel: why this `import`? `os` isn't used below?

easybuild/easyblocks/b/boost.py
```diff
@@ -121,9 +121,36 @@ def configure_step(self):
...
+                ': CC ',
+                ': <compileflags>-I$(CRAY_MPICH2_DIR)/include ',
+                '  <linkflags>-L$(CRAY_MPICH2_DIR)/lib \ ',
+                '; ',
+                'using mpi ',
+                ': CC ',
+                ': <find-shared-library>mpich ',
+                ': aprun -n \ ',
+                ';',
+                '',
```
boegel: each of the lines above should be indented with 4 spaces

easybuild/easyblocks/b/boost.py
gppezzi: This should only work for gnu, so I'd like to bail out for intel and cray. But how do I bail out ?

easybuild/easyblocks/b/boost.py
gppezzi: it was for debugging when I first fixed that part, not needed anymore.

easybuild/easyblocks/b/boost.py
gppezzi: If you're not with gnu you should update the easyblock :) 

I fixed that so that it is clear that only gnu is supported for now

easybuild/easyblocks/b/boost.py
gppezzi: @pforai  any idea? I've no idea (since I simply stole petar's PR :). But i'm removing it for now and re-testing.

gppezzi:
@boegel fixed most of remarks, I'll check tomorrow how to correctly bail out in case of unsupported PrgEnv

gppezzi:
@boegel @pforai @lucamar 

I've updated the easyblock to avoid the hardcoded reference to aprun/srun, now the mpi launcher can be specified through an extra_var. This argument is optional and it is only used for the regression tests (which we are currently not doing).

@pforai The argument `<find-shared-library>mpich` is actually needed for building the mpi enabled libraries, so apparently we need to leave it.

Build reports will be provided [here](https://github.com/hpcugent/easybuild-easyconfigs/pull/2615)

lucamar:
@gppezzi Thanks, it looks perfect

easybuild/easyblocks/b/boost.py
```diff
@@ -120,10 +125,35 @@ def configure_step(self):
...
+                        ': %s' % self.cfg['mpi_launcher'],
+                        ';',
+                        '',
+                    ])
+                else: 
+                    raise EasyBuildError("Bailing out: PrgEnv-intel and PrgEnv-cray not supported.")
+            else:
+                txt = "using mpi : %s ;" % os.getenv("MPICXX")
+
+        write_file('user-config.jam', txt, append=True)
```
boegel: @gppezzi indentation seems off here? should be intended one level deeper, so it's in the `if self.cfg['boost_mpi']:`, otherwise `txt` is undefined

boegel:
Tested with https://github.com/hpcugent/easybuild-easyconfigs/pull/2615 and existing Boost easyconfigs, good to go, thanks @gppezzi & @pforai!

gppezzi:
Thanks everyone for the feedback and help!

PR #852, opened 2016-03-02
new psmpi easyblock, and environment variable setup in mpich.py flush…

damianam:
New psmpi easyblock, and environment variable setup in mpich.py flushed out to its own method

### DIFF
--- a/easybuild/easyblocks/m/mpich.py
+++ b/easybuild/easyblocks/m/mpich.py
@@ -58,7 +58,13 @@ class EB_MPICH(ConfigureMake):
         })
         return extra_vars
 
-    def setup_env_vars(self):
+    # MPICH configure script complains when F90 or F90FLAGS are set,
+    # they should be replaced with FC/FCFLAGS instead.
+    # Additionally, there are a set of variables (FCFLAGS among them) that should not be set at configure time,
+    # or they will leak in the mpix wrappers.
+    # Specific variables to be included in the wrapper exists, but they changed between MPICH 3.1.4 and MPICH 3.2
+    # and in a typical scenario we probably don't want them.
+    def correct_mpich_build_env(self):
         """
         Method to correctly set the environment for MPICH and derivatives
         """
@@ -116,13 +122,7 @@ class EB_MPICH(ConfigureMake):
 
         self.cfg.update('configopts', ' '.join(add_configopts))
         
-        # MPICH configure script complains when F90 or F90FLAGS are set,
-        # they should be replaced with FC/FCFLAGS instead.
-        # Additionally, there are a set of variables (FCFLAGS among them) that should not be set at configure time,
-        # or they will leak in the mpix wrappers.
-        # Specific variables to be included in the wrapper exists, but they changed between MPICH 3.1.4 and MPICH 3.2
-        # and in a typical scenario we probably don't want them.
-        self.setup_env_vars()
+        self.correct_mpich_build_env()
 
         super(EB_MPICH, self).configure_step()
 
--- a/easybuild/easyblocks/m/mpich.py
+++ b/easybuild/easyblocks/m/mpich.py
@@ -85,7 +85,7 @@ class EB_MPICH(ConfigureMake):
                                          envvar, new_envvar, new_envvar, envvar, envvar_val)
         env.unset_env_vars(vars_to_unset)
 
-    def add_default_options(self):
+    def add_mpich_configopts(self):
         """
         Method to add common configure options for MPICH-based MPI libraries
         """
@@ -127,7 +127,7 @@ class EB_MPICH(ConfigureMake):
             super(EB_MPICH, self).make_dir(self.installdir, True, dontcreateinstalldir=True)
         
         if set_default_options:
-            self.add_default_options()
+            self.add_mpich_configopts()
 
         self.correct_mpich_build_env()
 
@@ -155,12 +155,9 @@ class EB_MPICH(ConfigureMake):
         else:
             libnames = ['fmpich', 'mpichcxx', 'mpichf90', 'mpich', 'mpl', 'opa']
 
-        launchers = ['mpiexec', 'mpiexec.hydra', 'mpirun']
-        compilers = ['mpicc', 'mpicxx', 'mpif77', 'mpif90']
-        if check_launchers is True:
-            binaries = launchers + compilers
-        else:
-            binaries = compilers
+        binaries = ['mpicc', 'mpicxx', 'mpif77', 'mpif90']
+        if check_launchers:
+            binaries = binaries.extend(['mpiexec', 'mpiexec.hydra', 'mpirun'])
 
         bins = [os.path.join('bin', x) for x in binaries]
         headers = [os.path.join('include', x) for x in ['mpi.h', 'mpicxx.h', 'mpif.h']]
--- a/easybuild/easyblocks/m/mpich.py
+++ b/easybuild/easyblocks/m/mpich.py
@@ -58,10 +58,28 @@ class EB_MPICH(ConfigureMake):
         })
         return extra_vars
 
-    # There is a number of configuration options that are typically needed that are not present
-    # here. The reason is that this easyblock is intended to be used as a parent for other
-    # easyblocks like MVAPICH2 and PSMPI. Not all of them support the same options, so they
-    # are not included here.
+    def setup_env_vars(self):
+        """
+        Method to correctly set the environment for MPICH and derivatives
+        """
+        env_vars = ['CFLAGS', 'CPPFLAGS', 'CXXFLAGS', 'FCFLAGS', 'FFLAGS', 'LDFLAGS', 'LIBS']
+        vars_to_unset = ['F90', 'F90FLAGS']
+        for envvar in env_vars:
+            envvar_val = os.getenv(envvar)
+            if envvar_val:
+                new_envvar = 'MPICHLIB_%s' % envvar
+                new_envvar_val = os.getenv(new_envvar)
+                vars_to_unset.append(envvar)
+                if envvar_val == new_envvar_val:
+                    self.log.debug("$%s == $%s, just defined $%s as empty", envvar, new_envvar, envvar)
+                elif new_envvar_val is None:
+                    env.setvar(new_envvar, envvar_val)
+                else:
+                    raise EasyBuildError("Both $%s and $%s set, can I overwrite $%s with $%s (%s) ?",
+                                         envvar, new_envvar, new_envvar, envvar, envvar_val)
+        env.unset_env_vars(vars_to_unset)
+
+
     def configure_step(self):
         """
         Custom configuration procedure for MPICH
@@ -97,29 +115,14 @@ class EB_MPICH(ConfigureMake):
         add_configopts.extend(['--enable-f77', '--enable-fc', '--enable-cxx'])
 
         self.cfg.update('configopts', ' '.join(add_configopts))
-
+        
         # MPICH configure script complains when F90 or F90FLAGS are set,
         # they should be replaced with FC/FCFLAGS instead.
         # Additionally, there are a set of variables (FCFLAGS among them) that should not be set at configure time,
         # or they will leak in the mpix wrappers.
         # Specific variables to be included in the wrapper exists, but they changed between MPICH 3.1.4 and MPICH 3.2
         # and in a typical scenario we probably don't want them.
-        env_vars = ['CFLAGS', 'CPPFLAGS', 'CXXFLAGS', 'FCFLAGS', 'FFLAGS', 'LDFLAGS', 'LIBS']
-        vars_to_unset = ['F90', 'F90FLAGS']
-        for envvar in env_vars:
-            envvar_val = os.getenv(envvar)
-            if envvar_val:
-                new_envvar = 'MPICHLIB_%s' % envvar
-                new_envvar_val = os.getenv(new_envvar)
-                vars_to_unset.append(envvar)
-                if envvar_val == new_envvar_val:
-                    self.log.debug("$%s == $%s, just defined $%s as empty", envvar, new_envvar, envvar)
-                elif new_envvar_val is None:
-                    env.setvar(new_envvar, envvar_val)
-                else:
-                    raise EasyBuildError("Both $%s and $%s set, can I overwrite $%s with $%s (%s) ?",
-                                         envvar, new_envvar, new_envvar, envvar, envvar_val)
-        env.unset_env_vars(vars_to_unset)
+        self.setup_env_vars()
 
         super(EB_MPICH, self).configure_step()
 
--- a/easybuild/easyblocks/m/mpich.py
+++ b/easybuild/easyblocks/m/mpich.py
@@ -128,7 +128,7 @@ class EB_MPICH(ConfigureMake):
 
     # make and make install are default
 
-    def sanity_check_step(self, custom_paths=None, use_new_libnames=None):
+    def sanity_check_step(self, custom_paths=None, use_new_libnames=None, check_launchers=True):
         """
         Custom sanity check for MPICH
         """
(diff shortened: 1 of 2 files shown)


damianam:
@boegel , as promised. New easyconfigs PR coming

boegel:
Jenkins: ok to test

easybuild/easyblocks/p/psmpi.py
```diff
@@ -0,0 +1,96 @@
...
+
+        # MPICH configure script complains when F90 or F90FLAGS are set,
+        # they should be replaced with FC/FCFLAGS instead.
+        # Additionally, there are a set of variables (FCFLAGS among them) that should not be set at configure time,
+        # or they will leak in the mpix wrappers.
+        # Specific variables to be included in the wrapper exists, but they changed between MPICH 3.1.4 and MPICH 3.2
+        # and in a typical scenario we probably don't want them.
+        self.setup_env_vars()
+
+        # Bypass the configure_step of EB_MPICH
```
boegel: mention why? (configure options defined there do not apply to psmpi?)

easybuild/easyblocks/m/mpich.py
```diff
@@ -97,29 +115,14 @@ def configure_step(self):
...
 
         self.cfg.update('configopts', ' '.join(add_configopts))
-
+        
         # MPICH configure script complains when F90 or F90FLAGS are set,
         # they should be replaced with FC/FCFLAGS instead.
         # Additionally, there are a set of variables (FCFLAGS among them) that should not be set at configure time,
         # or they will leak in the mpix wrappers.
         # Specific variables to be included in the wrapper exists, but they changed between MPICH 3.1.4 and MPICH 3.2
         # and in a typical scenario we probably don't want them.
```
boegel: move this comment into the method?

easybuild/easyblocks/m/mpich.py
```diff
@@ -58,10 +58,28 @@ def extra_options(extra_vars=None):
         })
         return extra_vars
 
-    # There is a number of configuration options that are typically needed that are not present
-    # here. The reason is that this easyblock is intended to be used as a parent for other
-    # easyblocks like MVAPICH2 and PSMPI. Not all of them support the same options, so they
-    # are not included here.
+    def setup_env_vars(self):
```
boegel: rename to `correct_mpich_build_env` (makes it clear this is specific to MPICH)

easybuild/easyblocks/p/psmpi.py
```diff
@@ -0,0 +1,96 @@
...
+        if not self.cfg['keeppreviousinstall']:
+            self.log.info("Making sure any old installation is removed before we start the build...")
+            super(EB_psmpi, self).make_dir(self.installdir, True, dontcreateinstalldir=True)
+
+        # MPICH configure script complains when F90 or F90FLAGS are set,
+        # they should be replaced with FC/FCFLAGS instead.
+        # Additionally, there are a set of variables (FCFLAGS among them) that should not be set at configure time,
+        # or they will leak in the mpix wrappers.
+        # Specific variables to be included in the wrapper exists, but they changed between MPICH 3.1.4 and MPICH 3.2
+        # and in a typical scenario we probably don't want them.
```
boegel: don't repeat the comment here if it's moved into the method

easybuild/easyblocks/p/psmpi.py
```diff
@@ -0,0 +1,96 @@
...
+            # cfr. http://git.mpich.org/mpich.git/blob_plain/v3.1.1:/CHANGES
+            # MPICH changed its library names sinceversion 3.1.1
+            use_new_libnames = LooseVersion(self.version) >= LooseVersion('5.1.0-1')
+
+        # Starting MPICH 3.1.1, libraries have been renamed
+        # cf http://git.mpich.org/mpich.git/blob_plain/v3.1.1:/CHANGES
+        if use_new_libnames:
+            libnames = ['mpi', 'mpicxx', 'mpifort']
+        else:
+            libnames = ['fmpich', 'mpichcxx', 'mpichf90', 'mpich', 'mpl', 'opa']
```
boegel: how close is this to the MPICH sanity check?

looks like a lot of copy-pasting to me, the difference between psmpi and MPICH doesn't really stand out (and it should, imho)

easybuild/easyblocks/p/psmpi.py
```diff
@@ -0,0 +1,96 @@
...
+
+        binaries = ['mpicc', 'mpicxx', 'mpif77', 'mpif90']
+        bins = [os.path.join('bin', x) for x in binaries]
+        headers = [os.path.join('include', x) for x in ['mpi.h', 'mpicxx.h', 'mpif.h']]
+        libs = [os.path.join('lib', 'lib%s.%s' % (l, e)) for l in libnames for e in ['a', shlib_ext]]
+
+        custom_paths.setdefault('dirs', []).extend(['bin', 'include', 'lib'])
+        custom_paths.setdefault('files', []).extend(bins + headers + libs)
+
+        # Bypass the sanity_check_step of EB_MPICH
```
boegel: mention *why* you're bypassing

but, I think we should allow tweaking the MPICH sanity check rather than reinventing it for 80-90%...

easybuild/easyblocks/p/psmpi.py
boegel: how close is this to the MPICH sanity check?

looks like a lot of copy-pasting to me, the difference between psmpi and MPICH doesn't really stand out (and it should, imho)

easybuild/easyblocks/p/psmpi.py
boegel: mention *why* you're bypassing

but, I think we should allow tweaking the MPICH sanity check rather than reinventing it for 80-90%...

easybuild/easyblocks/p/psmpi.py
```diff
@@ -0,0 +1,96 @@
...
+        # MPICH configure script complains when F90 or F90FLAGS are set,
+        # they should be replaced with FC/FCFLAGS instead.
+        # Additionally, there are a set of variables (FCFLAGS among them) that should not be set at configure time,
+        # or they will leak in the mpix wrappers.
+        # Specific variables to be included in the wrapper exists, but they changed between MPICH 3.1.4 and MPICH 3.2
+        # and in a typical scenario we probably don't want them.
+        self.setup_env_vars()
+
+        # Bypass the configure_step of EB_MPICH
+        super(EB_MPICH, self).configure_step()
```
boegel: another option is to move the MPICH-specific configure options block into a method, and then redefining that method here as:

```python
def define_mpich_config_opts(self):
    """Skip updating configure options for MPICH derivates; doesn't apply to psmpi."""
    pass
```

and then you can leave the `configure_step` of MPICH untouched; this avoids copy-pasting the `keeppreviousinstall` but, and the need to skip calling out to the `configure_step` of your parent like you're doing here

I think this may be cleaner...

easybuild/easyblocks/p/psmpi.py
damianam: I agree, but it was my understanding from yesterday's chat that you preferred this way. I'll change it now.

boegel: I guess I was expecting it to be more different, this is too similar and too much copy-pasting imho...

damianam:
@boegel all the suggested changes have been done. I haven't retested though.

easybuild/easyblocks/p/psmpi.py
```diff
@@ -0,0 +1,53 @@
...
+    """
+    Support for building the ParaStationMPI library.
+    - some compiler dependent configure options
+    """
+
+    def add_default_options(self):
+        """
+        Skip common options found in other MPICH-based runtimes
+        """
+        pass
```
damianam: I feel like it would be more elegant having an optional parameter in the `configure_step` in `mpich.py`. Something like:

`def configure_step(self, set_default_options=True):`

And then here, in `psmpi.py` something like:

```
def configure_step(self):
    [...] 
    super(EB_psmpi, self).configure_step(set_default_options=False)
```

@boegel would you agree?

easybuild/easyblocks/p/psmpi.py
```diff
@@ -0,0 +1,53 @@
...
+
+    # configure, make and make install are default
+
+    def sanity_check_step(self, custom_paths=None, use_new_libnames=None):
```
boegel: you can drop `use_new_libnames` here, since we don't have an easyblock that derives from `EB_psmpi` (and we'll probably won't have one soon)

easybuild/easyblocks/p/psmpi.py
```diff
@@ -0,0 +1,53 @@
...
+        """
+        Disable the checking of the launchers for ParaStationMPI
+        """
+        super(EB_psmpi, self).sanity_check_step(custom_paths=custom_paths,use_new_libnames=use_new_libnames,
```
boegel: isn't there a psmpi version yet where the new lib names are used?

easybuild/easyblocks/p/psmpi.py
boegel: same for `custom_paths` btw, just drop it here (and below)

easybuild/easyblocks/p/psmpi.py
boegel: I actually like it better like this, since it's very explicit.

Maybe replace the `pass` with `self.log.info("Not using MPICH-specific configure options to configure psmpi")

easybuild/easyblocks/m/mpich.py
```diff
@@ -145,7 +153,13 @@ def sanity_check_step(self, custom_paths=None, use_new_libnames=None):
         else:
             libnames = ['fmpich', 'mpichcxx', 'mpichf90', 'mpich', 'mpl', 'opa']
 
-        binaries = ['mpicc', 'mpicxx', 'mpiexec', 'mpiexec.hydra', 'mpif77', 'mpif90', 'mpirun']
+        launchers = ['mpiexec', 'mpiexec.hydra', 'mpirun']
+        compilers = ['mpicc', 'mpicxx', 'mpif77', 'mpif90']
+        if check_launchers is True:
+            binaries = launchers + compilers
```
boegel: I would do this:

```python
binaries = ['mpicc', 'mpicxx', 'mpif77', 'mpif90']
if check_launchers:
    binaries.extend(['mpiexec', 'mpiexec.hydra', 'mpirun'])
```

easybuild/easyblocks/m/mpich.py
```diff
@@ -145,7 +153,13 @@ def sanity_check_step(self, custom_paths=None, use_new_libnames=None):
         else:
             libnames = ['fmpich', 'mpichcxx', 'mpichf90', 'mpich', 'mpl', 'opa']
 
-        binaries = ['mpicc', 'mpicxx', 'mpiexec', 'mpiexec.hydra', 'mpif77', 'mpif90', 'mpirun']
+        launchers = ['mpiexec', 'mpiexec.hydra', 'mpirun']
+        compilers = ['mpicc', 'mpicxx', 'mpif77', 'mpif90']
+        if check_launchers is True:
```
boegel: just `if check_launchers` is sufficient ;)

easybuild/easyblocks/m/mpich.py
```diff
@@ -58,23 +58,37 @@ def extra_options(extra_vars=None):
...
+                    raise EasyBuildError("Both $%s and $%s set, can I overwrite $%s with $%s (%s) ?",
+                                         envvar, new_envvar, new_envvar, envvar, envvar_val)
+        env.unset_env_vars(vars_to_unset)
 
-        # things might go wrong if a previous install dir is present, so let's get rid of it
-        if not self.cfg['keeppreviousinstall']:
-            self.log.info("Making sure any old installation is removed before we start the build...")
-            super(EB_MPICH, self).make_dir(self.installdir, True, dontcreateinstalldir=True)
-
+    def add_default_options(self):
```
boegel: maybe rename to `add_mpich_configopts`?

it's a bit unclear now what this does based on the method name

easybuild/easyblocks/p/psmpi.py
```diff
@@ -0,0 +1,53 @@
...
+@author: Damian Alvarez (Forschungszentrum Juelich)
+"""
+
+from easybuild.easyblocks.mpich import EB_MPICH
+
+
+class EB_psmpi(EB_MPICH):
+    """
+    Support for building the ParaStationMPI library.
+    - some compiler dependent configure options
```
boegel: which ones?

easybuild/easyblocks/p/psmpi.py
damianam: I just don't like the idea of doing and undoing (defining a method in `mpich.py` and overriding it to make it do nothing).

With an extra optional parameter in the `configure_step` method we avoid that, and it is also pretty explicit (you see it in the signature of the method).

easybuild/easyblocks/p/psmpi.py
```diff
@@ -0,0 +1,110 @@
...
+        super(EB_psmpi, self).configure_step(set_default_options=False)
+
+    # make and make install are default
+
+    def sanity_check_step(self):
+        """
+        Disable the checking of the launchers for ParaStationMPI
+        """
+        # cfr. http://git.mpich.org/mpich.git/blob_plain/v3.1.1:/CHANGES
+        # MPICH changed its library names sinceversion 3.1.1
```
boegel: fix comment accordingly, any reference for psmpi?

easybuild/easyblocks/p/psmpi.py
```diff
@@ -0,0 +1,110 @@
...
+        use_new_libnames = LooseVersion(self.version) >= LooseVersion('5.1.0-1')
+
+        super(EB_psmpi, self).sanity_check_step(use_new_libnames=use_new_libnames,
+                check_launchers=False)
```
boegel: this fits on a single line now

easybuild/easyblocks/m/mpich.py
```diff
@@ -97,35 +111,31 @@ def configure_step(self):
...
+        * add common configure options for MPICH-based MPI libraries
+        * unset environment variables that leak into mpi* wrappers, and define $MPICHLIB_* equivalents instead
+        """
+
+        # things might go wrong if a previous install dir is present, so let's get rid of it
+        if not self.cfg['keeppreviousinstall']:
+            self.log.info("Making sure any old installation is removed before we start the build...")
+            super(EB_MPICH, self).make_dir(self.installdir, True, dontcreateinstalldir=True)
+        
+        if set_default_options:
```
boegel: rename to `add_mpich_configopts` (to make it match the function)?

boegel:
@damianam: almost there... :) :+1:

Can you look into a test report in the `psmpi` easyconfigs PR on top of this (after looking into the remarks)?

boegel:
@damianam: this looks good to go now, but we need a successful test report in https://github.com/hpcugent/easybuild-easyconfigs/pull/2619 before we can merge this...

boegel:
tested through https://github.com/hpcugent/easybuild-easyconfigs/pull/2619, looking good, so going in, thanks @damianam!

rjeschmi:
This is broken.

binaries = binaries.extend(['mpiexec', 'mpiexec.hydra', 'mpirun'])

sets binaries to None

easybuild/easyblocks/m/mpich.py
```diff
@@ -145,7 +155,10 @@ def sanity_check_step(self, custom_paths=None, use_new_libnames=None):
         else:
             libnames = ['fmpich', 'mpichcxx', 'mpichf90', 'mpich', 'mpl', 'opa']
 
-        binaries = ['mpicc', 'mpicxx', 'mpiexec', 'mpiexec.hydra', 'mpif77', 'mpif90', 'mpirun']
+        binaries = ['mpicc', 'mpicxx', 'mpif77', 'mpif90']
+        if check_launchers:
+            binaries = binaries.extend(['mpiexec', 'mpiexec.hydra', 'mpirun'])
```
boegel: oops, this is wrong indeed, thanks @rjeschmi...

boegel:
@rjeschmi urgh, this wasn't retested with any MPICH easyconfigs, so it slipped through.. :(

fixed in #868, thanks for reporting it!

Issue #157, opened 2013-03-27
trip over unknown configure options


boegel:
We need to make sure we trip when unknown configure options are being used.

The autoconf `configure` script will only issue a warning, but will happily continue, e.g. when using `--enable-mpi-threads` with OpenMPI v1.6.4:

```
configure: WARNING: unrecognized options: --enable-mpi-threads
```

This particular one was only figured out when OpenMPI was being used to build and test larger programs depending on its functionality, i.e. CP2K, so way too late in the build cycle.

Flamefire:
This came up again. There already is `parse_cmd_output`parse_cmd_output in `run.py` which can be used for that. However it only warns (by default) and is sometimes not fully useful.

Some use cases:
1. Fail on `WARNING: unrecognized options:`
2. Warn on "error" occurence
3. Maybe fail on "failed" (e.g. for Python there are test failures due to the sitecustomize.py used), so sort out "expected failures"

So my (backward compatible) suggestion would be to accept a list of regexes with an optional mode which can be "ignore", "warn" or "fail". Every line of output is matched against the regexes in-order and the first matching one wins and the corresponding action is executed. This list should be a property of the EB so descendants can change, overwrite or insert entries.

Example:

`regexes = ["WARNING: unrecognized options", ("test_site failed", IGNORE), ("checking.*failed", WARN), "failed"]`

boegel:
Error reporting is something I would like to use making improvements on, so I'm very happy to see contributors working on this.

I feeel your proposal makes a lot of sense @Flamefire, and could help a great deal is catching errors that otherwise go ignored.

Of course we'll need to be careful with suddenly treated the use of unrecognized `configure` options as failures, since that means breaking installations which have actually worked fine (although perhaps not entirely as was originally intended). For that particular one a very verbose warning would make more sense I think...

Perhaps the new method could also try and do a good job of trying to report the (first) error message that made a build fail. For example, the first `error: ...` compiler error + N lines above (for context), or `Error 1` + a couple preceding lines (but ignoring `Error 2` & such which typically pop up with parallel `make` runs).

Flamefire:
> Of course we'll need to be careful with suddenly treated the use of unrecognized configure options as failures, since that means breaking installations which have actually worked fine (although perhaps not entirely as was originally intended). For that particular one a very verbose warning would make more sense I think...

[...]

PR #853, opened 2016-03-03
Remove deprecated configure options from MVAPICH2 easyblock

besserox:
MVAPICH2 easyblock uses some deprecated configure options:
- `--enable-mpe` disappeared starting version >= 1.9
- `--enable-hwloc` disappeared starting version >= 2.0

This PR removes these deprecated options from the easyblock. 
PR https://github.com/hpcugent/easybuild-easyconfigs/pull/2626 for easyconfigs set these options directly in the easyconfig files when there are used.

PR https://github.com/hpcugent/easybuild-easyconfigs/pull/2626 have to be merged first.

### DIFF
--- a/easybuild/easyblocks/m/mvapich2.py
+++ b/easybuild/easyblocks/m/mvapich2.py
@@ -38,6 +38,7 @@ from distutils.version import LooseVersion
 
 from easybuild.easyblocks.mpich import EB_MPICH
 from easybuild.framework.easyconfig import CUSTOM
+from easybuild.tools.build_log import EasyBuildError
 
 
 class EB_MVAPICH2(EB_MPICH):
@@ -76,25 +77,36 @@ class EB_MVAPICH2(EB_MPICH):
             # However, the 'withmpe' option should be maintained for backward compatibility purpose
             if LooseVersion(self.version) < LooseVersion('1.9'):
                 add_configopts.append('--enable-mpe')
+
             else:
                 raise EasyBuildError("MPI Parallel Environment (MPE) is not available anymore starting MVAPICH2 1.9")
+
         if self.cfg['withlimic2']:
             add_configopts.append('--enable-limic2')
+
         if self.cfg['withchkpt']:
             add_configopts.extend(['--enable-checkpointing', '--with-hydra-ckpointlib=blcr'])
-        if self.cfg['withhwloc']:
-            # --with-hwloc/--without-hwloc option is not available anymore MVAPICH2 >= 2.0.
-            # Starting this version, HWLOC is apparently distributed with MVAPICH2 and always compiled with MVAPICH2,
-            # and it cannot be disabled.
-            # The 'withhwloc' option should be maintained for backward compatibility purpose.
-            # EasyBuild and MVAPCH2 will just silently ignore this option if it is used.
-            add_configopts.append('--with-hwloc')
+
+        # --with-hwloc/--without-hwloc option is not available anymore MVAPICH2 >= 2.0. Starting this version,
+        # HWLOC is apparently distributed with MVAPICH2 and always compiled with MVAPICH2, and it cannot be disabled.
+        # This check happens only if 'withhwloc = False' is explicitly specified in an easyconfig with MPIVACH2 >= 2.0
+        if not self.cfg['withhwloc'] and LooseVersion(self.version) >= LooseVersion('2.0'):
+            raise EasyBuildError("'withhwloc = False' is not supported anymore in easyconfigs for MVAPICH2 >= 2.0 because it cannot be disabled anymore")
+
+        if LooseVersion(self.version) < LooseVersion('2.0'):
+            if self.cfg['withhwloc']:
+                add_configopts.append('--with-hwloc')
+
+            else:
+                add_configopts.append('--without-hwloc')
 
         # pass BLCR paths if specified
         if self.cfg['blcr_path']:
             add_configopts.append('--with-blcr=%s' % self.cfg['blcr_path'])
+
         if self.cfg['blcr_inc_path']:
             add_configopts.append('--with-blcr-include=%s' % self.cfg['blcr_inc_path'])
+
         if self.cfg['blcr_lib_path']:
             add_configopts.append('--with-blcr-libpath=%s' % self.cfg['blcr_lib_path'])
 
--- a/easybuild/easyblocks/m/mvapich2.py
+++ b/easybuild/easyblocks/m/mvapich2.py
@@ -51,6 +51,7 @@ class EB_MVAPICH2(EB_MPICH):
         """Define custom easyconfig parameters specific to MVAPICH2."""
         extra_vars = {
             'withchkpt': [False, "Enable checkpointing support (required BLCR)", CUSTOM],
+            'withmpe': [False, "Build MPE routines", CUSTOM],
             'withlimic2': [False, "Enable LiMIC2 support for intra-node communication", CUSTOM],
             'rdma_type': ["gen2", "Specify the RDMA type (gen2/udapl)", CUSTOM],
             'blcr_path': [None, "Path to BLCR package", CUSTOM],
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/m/mvapich2.py
```diff
@@ -51,8 +51,6 @@ def extra_options():
         """Define custom easyconfig parameters specific to MVAPICH2."""
         extra_vars = {
             'withchkpt': [False, "Enable checkpointing support (required BLCR)", CUSTOM],
-            'withmpe': [False, "Build MPE routines", CUSTOM],
-            'withhwloc': [False, "Enable support for using hwloc support for process binding", CUSTOM],
```
boegel: these shouldn't be removed

although we can update the easyconfigs we ship ourselves accordingly, we should forget that there are other easyconfigs out there we do not control...

easybuild/easyblocks/m/mvapich2.py
```diff
@@ -69,14 +67,10 @@ def configure_step(self):
         add_configopts.append('--with-rdma=%s' % self.cfg['rdma_type'])
 
         # enable specific support options (if desired)
-        if self.cfg['withmpe']:
-            add_configopts.append('--enable-mpe')
```
boegel: how about issuing a warning that this is deprecated?

do we know when this is going to be removed?

easybuild/easyblocks/m/mvapich2.py
```diff
@@ -69,14 +67,10 @@ def configure_step(self):
...
         if self.cfg['withlimic2']:
             add_configopts.append('--enable-limic2')
         if self.cfg['withchkpt']:
             add_configopts.extend(['--enable-checkpointing', '--with-hydra-ckpointlib=blcr'])
-        if self.cfg['withhwloc']:
-            add_configopts.append('--with-hwloc')
```
boegel: warning, error if still used in version where it's no longer available?

boegel:
Jenkins: ok to test

easybuild/easyblocks/m/mvapich2.py
besserox: @boegel Yes, you're right.

What is the way to give a warning or an error in EasyBuild?

besserox:
Just some notes, for later:

- About MPE
  - `--enable-mpe` disappeared starting MVAPICH2 version >= 1.9
  - It is actually an option coming from MPICH, not MVAPICH2. 


- About HWLOC
  - `--enable-hwloc` disappeared from configure starting MVAPICH2 version >= 2.0
  - HWLOC is apparently distributed with MVAPICH2 and always compiled
  - `--enable-hwloc`/`--disable-hwloc` still appears in the user guide (but I guess they forgot to update this)

easybuild/easyblocks/m/mvapich2.py
boegel: you can use `self.log.warning` for warnings (but they'll only end up in the log file)

to raise an error, use `raise EasyBuildError("<msg goes here>")`

besserox:
@boegel This PR is not doing much now except documenting a bit more the issues.
- It raises an error if 'withmpe' is used for MVAPICH2 1.9 or above
- If 'withhwloc' is false or not specifed (false is the default value), MVAPICH2 2.0 or above will still use HWLOC anyway because it not optional anymore.

Let me know if you have a better behavior to propose.

easybuild/easyblocks/m/mvapich2.py
```diff
@@ -70,12 +70,24 @@ def configure_step(self):
...
             add_configopts.append('--enable-limic2')
         if self.cfg['withchkpt']:
             add_configopts.extend(['--enable-checkpointing', '--with-hydra-ckpointlib=blcr'])
         if self.cfg['withhwloc']:
+            # --with-hwloc/--without-hwloc option is not available anymore MVAPICH2 >= 2.0.
+            # Starting this version, HWLOC is apparently distributed with MVAPICH2 and always compiled with MVAPICH2,
+            # and it cannot be disabled.
+            # The 'withhwloc' option should be maintained for backward compatibility purpose.
+            # EasyBuild and MVAPCH2 will just silently ignore this option if it is used.
             add_configopts.append('--with-hwloc')
```
boegel: is this actually still a known configure option in MVAPICH >= 2.0?

if not, we should not use it anymore... 

cfr. https://github.com/hpcugent/easybuild-easyblocks/issues/157

easybuild/easyblocks/m/mvapich2.py
```diff
@@ -70,12 +70,24 @@ def configure_step(self):
 
         # enable specific support options (if desired)
         if self.cfg['withmpe']:
-            add_configopts.append('--enable-mpe')
+            # --enable-mpe is coming from MPICH.
```
boegel: I find this confusing, since the MPICH easyblock is unaware of `--enable-mpe`...

maybe say this is a configure option for all MPICH-based MPI libraries?

boegel:
@besserox: the logic makes sense to me, but we could do a little bit better (cfr. remarks)

easybuild/easyblocks/m/mvapich2.py
besserox: Two things:
- `--with-hwloc` is not recognized by the configure (but it is still in the documentation). I can clarify this with the MVAPICH2 team.
- The `configure` from MPICH/MVAPCH2 is not checking for the unrecognized options. Because they use a top-level configure which is calling many other configure scripts with all the same options (some options being used by some configure scripts, and not by other ones). So #157 will not apply here.

easybuild/easyblocks/m/mvapich2.py
besserox: @boegel  Yes, it makes sense. I can do so.

However, it also means that we have to do the check for the version in the 2 easyblocks (mpich and mvapich2), like we did for the `use_new_libnames` (cf PR https://github.com/hpcugent/easybuild-easyblocks/pull/844). This brings more complexity. Do we want to go for this?

One other solution to unify all this would be to add a extra option `mpich_version` in the MPICH easyblock. Then, every derived easyconfig would have to declare the MPICH version it's based on. For example, in a MVAPICH2 easyconfig, we would have:
```
name = 'MVAPICH2'
version = '2.0.1'
mpich_version = '3.1'

...
```
Then the version check for the MPICH features can be done using `mpich_version` instead of `version`. This way the derived easyblocks do not need to know anything. What do you think?

besserox: @boegel Any comment on my idea above to have an `mpich_version = 'X.X'` in the easyconfig file?

easybuild/easyblocks/m/mvapich2.py
boegel: OK, but does it make any sense to include it, since it has no effect?

So, we should just skip it and issue a warning for MVAPICH >= 2.0?

easybuild/easyblocks/m/mvapich2.py
boegel: I'm not a big fan of including two versions that need to be kept in sync in all the MVAPICH2 easyconfigs...

My remark was actually purely on the comment itself, not on the version check for `--enable-mpe` below... That's good as is imho.

As it is right now, the MPICH easyblock doesn't support a way to include `--enable-mpe`, so right now, it's a non-issue?

besserox:
@boegel This should be better now.

One change is that the default value for `withhwloc` is now True. But this does not change the behavior of old easyconfig files because:
    - previously, if `withhwloc` was False, we did not set explicitly `--without-hwloc`
    - and if `--without-hwloc` was not set, the actual default behavior of the configure was to enable it

So now I'm able to find out detect if the easyconfig set `withhwloc` to False for MVAPCH2 >= 2.0 and raise an error.

boegel:
@besserox: this needs https://github.com/besserox/easybuild-easyblocks/pull/6 to fix a merge conflict (because #854 was merged)

boegel:
retested with all existing MVAPICH2 easyconfigs through https://github.com/hpcugent/easybuild-easyconfigs/pull/2707, good to go, thanks @besserox!

PR #854, opened 2016-03-03
Use correct configure option for checkpoint/restart in MVAPICH2 easyblock

besserox:
MVAPICH2 easyblock uses the wrong configure options to enable the checkpoint/restart support. It uses the options from MPICH instead of the one from MVAPICH2. As a result, I'm pretty sure that the checkpoint/restart is not working properly with this build.

The issue with this change is that the result of the build will be different. However, this affect only one easyconfig file `MVAPICH2-1.7-GCC-4.6.3-hwloc-chkpt.eb`. 

For reference:
- Checkpoint/Restart support in MVAPICH2:
http://mvapich.cse.ohio-state.edu/static/media/mvapich/mvapich2-2.2b-userguide.html#x1-120004.4
- Checkpoint/Restart support in MPICH:
https://wiki.mpich.org/mpich/index.php/Checkpointing

### DIFF
--- a/easybuild/easyblocks/m/mvapich2.py
+++ b/easybuild/easyblocks/m/mvapich2.py
@@ -96,8 +96,12 @@ class EB_MVAPICH2(EB_MPICH):
         """
         Custom sanity check for MVAPICH2
         """
+        mv2_bins = ['bin/mpiexec.mpirun_rsh']
+        if self.cfg['withchkpt']:
+            mv2_bins += ['bin/mv2_checkpoint']
+
         custom_paths = {
-            'files': ['bin/mpiexec.mpirun_rsh'],
+            'files': mv2_bins,
         }
 
         # cfr. http://git.mpich.org/mpich.git/blob_plain/v3.1.1:/CHANGES
--- a/easybuild/easyblocks/m/mvapich2.py
+++ b/easybuild/easyblocks/m/mvapich2.py
@@ -74,7 +74,7 @@ class EB_MVAPICH2(EB_MPICH):
         if self.cfg['withlimic2']:
             add_configopts.append('--enable-limic2')
         if self.cfg['withchkpt']:
-            add_configopts.extend(['--enable-checkpointing', '--with-hydra-ckpointlib=blcr'])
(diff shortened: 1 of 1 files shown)


boegel:
Jenkins: ok to test

besserox:
@boegel This PR is independent of my other MVAPICH2-related PRs. 

Any thought about it?

boegel:
@besserox is there any way to check whether the checkpointing support is working or not?

Don't worry about this changing the build, this is a bug fix, so we're changing the build for good reasons.

We don't need to maintain backward compatibility with doing it wrong in the past... ;)

besserox:
@boegel As far I know there is nothing to test the checkpoint support. Probably the best thing would be to run a MPI application, checkpoint it, and restart it. But that's a bit of a heavy process for EasyBuild. That's more something for my MPI stack testing tool (https://github.com/besserox/ATIS/tree/master/MPI), but it's a dead project for now.

boegel:
@besserox Maybe with `ldd libmpich.so`, which may show that `libcr.so` is linked?

In any case: your fix is triggering something alright, the build of the one MVAPICH easyconfig that is supposed to enable checkpointing is now failing with:

```
cr.c:29:24: fatal error: attr/xattr.h: No such file or directory
compilation terminated.
make[8]: *** [cr.o] Error 1
```

Does it work on your end? Any idea what's going on?

besserox:
@boegel  

- About the test, I remember that MVAPICH2 installs an additional executable `mv2_checkpoint` when checkpoint/restart is enabled. I have added this check.

  We can also check if `libmpich.so` and/so `mpirun_rsh` are linked to `libcr.so` as you suggested. 
  What would be the proper way to do an `ldd mpirun_rsh | grep libcr.so` in Python?

- About the compilation error you reported. I have updated the `osdependencies` in the easyconfig in PR https://github.com/hpcugent/easybuild-easyconfigs/pull/2699

boegel:
@besserox just checking for `mv2_checkpoint` is good enough for me

Tested with https://github.com/hpcugent/easybuild-easyconfigs/pull/2699, good to go, thanks!

PR #857, opened 2016-03-04
Totalview and find_flexlm_license

damianam:
This PR updates the totalview easyblock to use the new `find_flexlm_license` function in the framework

### DIFF
--- a/easybuild/easyblocks/t/totalview.py
+++ b/easybuild/easyblocks/t/totalview.py
@@ -33,7 +33,7 @@ class EB_TotalView(EasyBlock):
         """Initialisation of custom class variables for Totalview"""
         super(EB_TotalView, self).__init__(*args, **kwargs)
         
-        self.license_var = 'UNKNOWN'
+        self.license_file = 'UNKNOWN'
         self.license_env_var = 'UNKNOWN'
 
     def configure_step(self):
@@ -52,7 +52,6 @@ class EB_TotalView(EasyBlock):
                 self.log.info("Using Totalview license specifications from %s: %s", self.license_env_var, lic_specs)
 
             self.license_file = os.pathsep.join(lic_specs)
-            self.cfg['license_file'] = self.license_file
             env.setvar(self.license_env_var, self.license_file)
 
         else:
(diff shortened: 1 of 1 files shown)


boegel:
Jenkins: ok to test

easybuild/easyblocks/t/totalview.py
```diff
@@ -28,13 +32,32 @@ class EB_TotalView(EasyBlock):
     def __init__(self, *args, **kwargs):
         """Initialisation of custom class variables for Totalview"""
         super(EB_TotalView, self).__init__(*args, **kwargs)
-        if not self.cfg['license_file']:
-            self.cfg['license_file'] = 'UNKNOWN'
+        
+        self.license_var = 'UNKNOWN'
```
boegel: `self.license_file`?

easybuild/easyblocks/t/totalview.py
```diff
@@ -28,13 +32,32 @@ class EB_TotalView(EasyBlock):
...
+
+        if lic_specs:
+            if self.license_env_var is None:
+                self.log.info("Using Totalview license specifications from 'license_file': %s", lic_specs)
+                self.license_env_var = default_lic_env_var
+            else:
+                self.log.info("Using Totalview license specifications from %s: %s", self.license_env_var, lic_specs)
+
+            self.license_file = os.pathsep.join(lic_specs)
+            self.cfg['license_file'] = self.license_file
```
boegel: there's no point in updating this, since it's not used later (only `self.license_file` is)

besides, it's better not to fiddle with the parsed easyconfig

easybuild/easyblocks/t/totalview.py
damianam: Yep.

easybuild/easyblocks/t/totalview.py
damianam: That is a good point. The same applies to the PGI and IntelBase easyblocks.

boegel:
lgtm

boegel:
tested with existing TotalView  easyconfigs, works like a charm, so going in, thanks @damianam

PR #858, opened 2016-03-05
ensure list of Python lib dirs always has a 'lib/...' entry

boegel:
I ran into this when bootstrapping EasyBuild on a SLES11 system.

This change is important to make sure that `$PYTHONPATH` is being defined correctly later on.


boegel:
@wpoely86: please review?

wpoely86:
:+1:

boegel:

PR #859, opened 2016-03-07
Easyblock for mutil + MakeCp enhancement.

wpoely86:
For the `MakeCp` block you can now specify:
```python
files_to_copy = [(['oldname1','oldname2'], ['newname1','newname2'], 'target-dir')]
```

@boegel please review

### DIFF
--- a/easybuild/easyblocks/generic/makecp.py
+++ b/easybuild/easyblocks/generic/makecp.py
@@ -95,7 +95,7 @@ class MakeCp(ConfigureMake):
                 if not os.path.exists(target):
                     os.makedirs(target)
 
-                for files_spec in files_specs:
+                for idx, files_spec in enumerate(files_specs):
                     # first look for files in start dir
                     filepaths = glob.glob(os.path.join(self.cfg['start_dir'], files_spec))
                     tup = (files_spec, self.cfg['start_dir'], filepaths)
@@ -112,7 +112,11 @@ class MakeCp(ConfigureMake):
                     if not filepaths:
                         raise EasyBuildError("No files matching '%s' found anywhere.", files_spec)
 
-                    for idx, filepath in enumerate(filepaths):
+                    if new_names and len(filepaths) != 1:
+                        raise EasyBuildError("When a list with new names has been specified, the original file spec can \
+                                             only match a single file yet it gives: %s", filepaths)
+
+                    for filepath in filepaths:
                         # copy individual file
                         if os.path.isfile(filepath):
                             if new_names:
--- a/easybuild/easyblocks/generic/makecp.py
+++ b/easybuild/easyblocks/generic/makecp.py
@@ -74,17 +74,9 @@ class MakeCp(ConfigureMake):
                     # ([src1, src2], targetdir)
                     if len(fil) == 2 and isinstance(fil[0], list) and isinstance(fil[1], basestring):
                         files_specs = fil[0]
-                        new_names = None
                         target = os.path.join(self.installdir, fil[1])
-                    # ([src1, src2], [src1_new, src2_new], targetdir)
-                    elif len(fil) == 3 and isinstance(fil[0], list) and isinstance(fil[0], list) and \
-                            len(fil[0]) == len(fil[1]) and isinstance(fil[2], basestring):
-                        files_specs = fil[0]
-                        new_names = fil[1]
-                        target = os.path.join(self.installdir, fil[2])
                     else:
-                        raise EasyBuildError("Only tuples of format '([<source files>], <target dir>)' or \
-                                              '([<source files>], [<new name files>], <target dir>)' are supported.")
+                        raise EasyBuildError("Only tuples of format '([<source files>], <target dir>)' supported.")
                 # 'src_file' or 'src_dir'
                 elif isinstance(fil, basestring):
                     files_specs = [fil]
@@ -95,7 +87,16 @@ class MakeCp(ConfigureMake):
                 if not os.path.exists(target):
                     os.makedirs(target)
 
-                for idx, files_spec in enumerate(files_specs):
+                for orig_files_spec in files_specs:
+                    if isinstance(orig_files_spec, tuple):
+                        files_spec = orig_files_spec[0]
+                        dest = orig_files_spec[1]
+                    else:
+                        files_spec = orig_files_spec
+                        dest = None
+
+                    self.log.debug("Found: %s to %s", files_spec, dest)
+
                     # first look for files in start dir
                     filepaths = glob.glob(os.path.join(self.cfg['start_dir'], files_spec))
                     tup = (files_spec, self.cfg['start_dir'], filepaths)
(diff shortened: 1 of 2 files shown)


easybuild/easyblocks/generic/makecp.py
```diff
@@ -74,9 +74,17 @@ def install_step(self):
                     # ([src1, src2], targetdir)
                     if len(fil) == 2 and isinstance(fil[0], list) and isinstance(fil[1], basestring):
                         files_specs = fil[0]
+                        new_names = None
                         target = os.path.join(self.installdir, fil[1])
+                    # ([src1, src2], [src1_new, src2_new], targetdir)
+                    elif len(fil) == 3 and isinstance(fil[0], list) and isinstance(fil[0], list) and \
```
boegel: you're checking `fil[0]` twice

also, both `list` and `tuple` are valid as types?

boegel:
@wpoely86: wouldn't supporting this make more sense?

```python
files_to_copy = [([('oldname1', 'newname1'), ('oldname2', 'newname2'), 'name3'], 'target-dir')]
```

easybuild/easyblocks/m/mutil.py
```diff
@@ -0,0 +1,63 @@
...
+"""
+import re
+import os
+
+from easybuild.easyblocks.generic.makecp import MakeCp
+from easybuild.tools.build_log import EasyBuildError
+
+
+class EB_mutil(MakeCp):
+    """Class to build mutil"""
```
boegel: `s/Class/Easyblock/g`, `s/build/build and install/g`

easybuild/easyblocks/m/mutil.py
```diff
@@ -0,0 +1,63 @@
...
+
+    def __init__(self, *args, **kwargs):
+        """Easyblock constructor, set correct options"""
+        super(EB_mutil, self).__init__(*args, **kwargs)
+
+        self.cfg['with_configure'] = True
```
boegel: include a log message for this?

easybuild/easyblocks/m/mutil.py
```diff
@@ -0,0 +1,63 @@
...
+        if not os.path.isfile(patch_path):
+            raise EasyBuildError("Could not find the patch for coreutils: %s", coreutils_patch)
+
+        # add the extract patch to the list of patches
+        newpatch = {
+            'name': coreutils_patch,
+            'path': patch_path,
+            'checksum': None,
+        }
+        self.patches.insert(0, newpatch)
```
boegel: why inject as first patch?

if you have a good reason, include a comment

easybuild/easyblocks/m/mutil.py
```diff
@@ -0,0 +1,63 @@
...
+    def extract_step(self):
+        """Set the start dir correctly and add patch from source."""
+        super(EB_mutil, self).extract_step()
+
+        # 1.822.3 -> 8.22
+        coreutils_version = re.sub(r"^\d+.(\d+)(\d\d).\d+", r"\1.\2", self.version)
+        coreutils_patch = "coreutils-%s.patch" % coreutils_version
+        patch_path = os.path.join(self.builddir, "%s-%s" % (self.name, self.version), "patch", coreutils_patch)
+        if not os.path.isfile(patch_path):
+            raise EasyBuildError("Could not find the patch for coreutils: %s", coreutils_patch)
```
boegel: mention `patch_path` too (just in case we construct it wrongly at some point)

easybuild/easyblocks/m/mutil.py
wpoely86: without the patch, it's just coreutils. This patch 'creates' the actual mutil so any other patch will be after it.

easybuild/easyblocks/m/mutil.py
```diff
@@ -0,0 +1,63 @@
...
+
+    def extract_step(self):
+        """Set the start dir correctly and add patch from source."""
+        super(EB_mutil, self).extract_step()
+
+        # 1.822.3 -> 8.22
+        coreutils_version = re.sub(r"^\d+.(\d+)(\d\d).\d+", r"\1.\2", self.version)
+        coreutils_patch = "coreutils-%s.patch" % coreutils_version
+        patch_path = os.path.join(self.builddir, "%s-%s" % (self.name, self.version), "patch", coreutils_patch)
+        if not os.path.isfile(patch_path):
```
boegel: include a log message if it's found:

```python
if os.path.isfile(patch_path):
    self.log.info("coreutils patch found at %s", patch_path)
else:
    raise ...
```

easybuild/easyblocks/m/mutil.py
wpoely86: if you want but it's pretty obvious if you run it? It's not like an optional thing.

easybuild/easyblocks/m/mutil.py
```diff
@@ -0,0 +1,63 @@
...
+            raise EasyBuildError("Could not find the patch for coreutils: %s", coreutils_patch)
+
+        # add the extract patch to the list of patches
+        newpatch = {
+            'name': coreutils_patch,
+            'path': patch_path,
+            'checksum': None,
```
boegel: hmm, you're defining it as if this was added through `fetch_patches`, so if we change something there, this will need to be updated accordingly

should we refactor `fetch_patches` in framework so you can just call out to `self.fetch_patch` here, rather than 'hardcoding' the definition of an 'internal' datastructure?

easybuild/easyblocks/m/mutil.py
boegel: shouldn't we just run `apply_patch` in `configure_step` instead then, rather than trying to convince the framework to apply the patch during `patch_step`?

that would be cleaner imho (see also other remark w.r.t. defining internal data structure)

easybuild/easyblocks/m/mutil.py
boegel: it's just that this is kind of an indirect way of doing things...

how about this instead?

```python
def configure_step(self):
    ConfigureMake.configure_step(self)
```

or, just allowing to hard specify that configure should be run in `MakeCp`:

```python
def configure_step(self):
    super(EB_mutil, self).configure_step(with_configure=True)
```

wpoely86: and that's not indirect? 

We have an option for it, why not use it? It's not restricted to easyconfigs.

boegel: well, the thing I don't like is that you're modifying the parsed easyconfig, we should probably move away from doing that...

wpoely86: whyso? I mean, we have this option and the are going to totally ignore it?

boegel: no, this option is there to be set from easyconfig files...

wpoely86: yes, and you want to ignore it completely. For all I known some neat extra features will be added in makecp and I want them. This option is the correct choice. We put by default to `True` instead of `False` in the generic block.

boegel: No, you're *not* just changing the default, you're overriding the value set in the easyconfig file, totally ignoring what may have been set in the easyconfig file...

Changing the default would be done via `extra_options`...

wpoely86: I'm not overriding something in the easyconfig, it's done in the `__init__`.

wpoely86:
@boegel ready

boegel:
@wpoely86: the new `files_to_copy` format is still a bit weird since the target path is split in two (filename and target dir), but I can't think of an easy/better option, so fine by me

Let's see this in action with your mutil easyconfig PR...

boegel:
https://github.com/hpcugent/easybuild-easyconfigs/pull/2201 is good to go, so this can be merged too

Thanks @wpoely86!

PR #860, opened 2016-03-08
New EasyBlock for SuperLU

besserox:


boegel:
Jenkins: ok to test

easybuild/easyblocks/s/superlu.py
```diff
@@ -0,0 +1,133 @@
...
+
+
+    def configure_step(self):
+        """
+        Set the CMake options for SuperLU
+        """
+        self.cfg['separate_build_dir'] = True
+
+        if self.cfg['build_shared_libs']:
+            self.cfg['configopts'] += "-DBUILD_SHARED_LIBS=ON "
```
boegel: when updating `self.cfg`, use `self.cfg.update`:

```
self.cfg.update('configopts', '-DBUILD_SHARED_LIBS=ON')
```

note that you don't need to worry about including a trailing space, `update` will do that for you

easybuild/easyblocks/s/superlu.py
```diff
@@ -0,0 +1,133 @@
...
+    def extra_options():
+        """
+        Define custom easyconfig parameters for SuperLU.
+        """
+        extra_vars = {
+            'build_shared_libs': [False, "Build shared library (instead of static library)", CUSTOM],
+        }
+        return CMakeMake.extra_options(extra_vars)
+
+
```
boegel: PEP8 styleguide prescribes single empty line between class methods (but two blank lines between 'top-level' function/class definitions)

easybuild/easyblocks/s/superlu.py
```diff
@@ -0,0 +1,133 @@
...
+        # Make sure to add -fPIC
+        self.cfg['configopts'] += "-DCMAKE_POSITION_INDEPENDENT_CODE=ON "
+
+        # Make sure not to build the slow BLAS library included in the package
+        self.cfg['configopts'] += "-Denable_blaslib=OFF "
+
+        # Set the BLAS library to use
+        # For this, use the BLA_VENDOR option from the FindBLAS module of CMake
+        # cf https://cmake.org/cmake/help/latest/module/FindBLAS.html
+        if get_software_root('IMKL'):
```
boegel: use `imkl` (the official software name), although it matters little in practice

you're also missing a version check here, since the value you set is specific to imkl 10.x?

easybuild/easyblocks/s/superlu.py
```diff
@@ -0,0 +1,133 @@
...
+            self.cfg['configopts'] += '-DBLA_VENDOR="ACML" '
+        elif get_software_root('ATLAS'):
+            self.cfg['configopts'] += '-DBLA_VENDOR="ATLAS" '
+        elif get_software_root('OpenBLAS'):
+            # Unfortunately, OpenBLAS is not recognized by FindBLAS from CMake,
+            # we have to specify the OpenBLAS library manually
+            self.cfg['configopts'] += '-DBLAS_LIBRARIES="${EBROOTOPENBLAS}/lib/libopenblas.a;-pthread" '
+        else:
+            # Fallback on Generic BLAS otherwise
+            self.cfg['configopts'] += '-DBLA_VENDOR="Generic" '
```
boegel: or raise an error? we should try and make very sure we don't build SuperLU unoptimized?

easybuild/easyblocks/s/superlu.py
```diff
@@ -0,0 +1,133 @@
...
+
+        super(EB_SuperLU, self).configure_step()
+
+
+    def test_step(self):
+        """
+        Run the testsuite of SuperLU
+        """
+        self.cfg['configopts'] = "test"
```
boegel: should you be using `configopts` here?

boegel: maybe you want to define `self.cfg['runtest'] = 'test'`, to trigger `make test`?

easybuild/easyblocks/s/superlu.py
```diff
@@ -0,0 +1,133 @@
...
+
+        if self.cfg['build_shared_libs']:
+            lib_ext = get_shared_lib_ext()
+        else:
+            lib_ext = "a"
+        expected_libpath = os.path.join(self.installdir, "lib", "libsuperlu.%s" % lib_ext)
+        actual_libpath   = os.path.join(self.installdir, "lib", "libsuperlu_%s.%s" % (self.cfg['version'],lib_ext) )
+
+        if not os.path.exists(expected_libpath):
+            os.symlink(actual_libpath, expected_libpath)
```
boegel: wrap this in a `try`:

```python
try:
    ....
except OSError as err:
    raise EasyBuildError("... %s ...", err)

File: easybuild/easyblocks/s/superlu.py [hunk 7]
```diff
@@ -0,0 +1,133 @@
...
+        """
+        self.cfg['separate_build_dir'] = True
+
+        if self.cfg['build_shared_libs']:
+            self.cfg['configopts'] += "-DBUILD_SHARED_LIBS=ON "
+        else:
+            self.cfg['configopts'] += "-DBUILD_SHARED_LIBS=OFF "
+
+        # Make sure to add -fPIC
+        self.cfg['configopts'] += "-DCMAKE_POSITION_INDEPENDENT_CODE=ON "
```
boegel: this should depend on `self.toolchains.options['pic']`

easybuild/easyblocks/s/superlu.py
fgeorgatos: no reason for error: that would break the functionality of --try-toolchain;
is there are a cause to not have that working on best-effort basis?!

fgeorgatos: perhaps a warning is missing there..

easybuild/easyblocks/s/superlu.py
```diff
@@ -0,0 +1,133 @@
+##
+# Copyright 2016 University of Luxembourg
```
fgeorgatos: AFAIK, this should either be dual-licensed MIT/GPL or Kenneth will ask you for extra work :)
(dual licensing is cheap and is based on precedent though)

easybuild/easyblocks/s/superlu.py
boegel: it depends on what 'generic' means here...

if this is the reference netlib BLAS, it's a big no-no to just fall back to that and have a 'working' SuperLU as result, since you'll actually be getting a functional yet crippled (performance-wise) install

easybuild/easyblocks/s/superlu.py
boegel: https://github.com/hpcugent/easybuild-easyblocks/blob/master/CONTRIBUTING.md

besserox:
@boegel @fgeorgatos Thanks for reviewing. Some comments below:

- About the version check for Intel MKL. The possible values of `BLA_VENDOR` for Intel are: 
  - `Intel10_32` -> Intel mkl v10 32 bit
  - `Intel10_64lp` -> Intel mkl v10 64 bit,lp thread model, lp64 model
  - `Intel10_64lp_seq` -> Intel mkl v10 64 bit,sequential code, lp64 model
  - `Intel` -> older versions of mkl 32 and 64 bit

  I have added a check and I pick either `Intel10_64lp` or `Intel`

- About the BLAS fallback:
  - `Generic` is looking for a library named `blas`. It picks `/usr/lib/libblas.so` on our cluster.
  - SuperLU also provide a BLAS implementation *not necessarily fast* (as they say) that we could fallback to if we set `enable_blaslib=ON`
  - We can also set `BLA_VENDOR="All"`. Then FindBLAS will try all and use the first one it finds. On our cluster this one find ATLAS, ie `/usr/lib/libf77blas.so.3gf;/usr/lib/libatlas.so.3gf` even the Intel toolchain is loaded. **I changed the fallback to this one**. 
  - Or do you prefer to fail?

  BTW, you can look at `<CMake install>/share/cmake-*/Modules/FindBLAS.cmake` to see how it searches for the BLAS libraries.

- About PIC, what is the default value? Do we need to add toolchainopts = {'pic': True} in the easyconfig?

- About the licence, no problem. I have added "Ghent University" in the copyright line. Is that fine like this, or do you want *Contributor Agreement* line too?

easybuild/easyblocks/s/superlu.py
```diff
@@ -0,0 +1,141 @@
...
+        elif get_software_root('ATLAS'):
+            self.cfg.update('configopts', '-DBLA_VENDOR="ATLAS"')
+        elif get_software_root('OpenBLAS'):
+            # Unfortunately, OpenBLAS is not recognized by FindBLAS from CMake,
+            # we have to specify the OpenBLAS library manually
+            self.cfg.update('configopts', '-DBLAS_LIBRARIES="${EBROOTOPENBLAS}/lib/libopenblas.a;-pthread"')
+        else:
+            # Fallback: try everything, pick the first one
+            # CMake should fail if none is found
+            self.cfg.update('configopts', '-DBLA_VENDOR="All"')
```
boegel: I prefer failing here.

The EB policy is to use the BLAS available in the toolchain. If it's not available, we shouldn't let SuperLU pick another random BLAS library it may find...

easybuild/easyblocks/s/superlu.py
```diff
@@ -0,0 +1,141 @@
...
+
+        # Make sure not to build the slow BLAS library included in the package
+        self.cfg.update('configopts', '-Denable_blaslib=OFF')
+
+        # Set the BLAS library to use
+        # For this, use the BLA_VENDOR option from the FindBLAS module of CMake
+        # cf https://cmake.org/cmake/help/latest/module/FindBLAS.html
+        if get_software_root('imkl'):
+            imkl_version = get_software_version('imkl')
+            if LooseVersion(imkl_version) >= LooseVersion('10'):
```
boegel: is `Intel10_64lp` also the right setting for imkl 11.x?

easybuild/easyblocks/s/superlu.py
```diff
@@ -0,0 +1,141 @@
...
+                self.cfg.update('configopts', '-DBLA_VENDOR="Intel10_64lp"')
```
boegel: please include a comment that explains what this relates to exactly (IMKL 10.x (or newer?), 64-bit, LP64 model, thread model)
and what the known alternatives are (`Intel10_32` and `Intel10_64lp_seq`)

boegel:
@besserox: `pic` is not enabled by default, so if you need it, you should enable it in the easyconfig; 

license looks fine now, but the important thing is that you're agreeing with the contributor agreement (if not, you should close this PR)

besserox:
@boegel Some more comments:

- So now, it fails if none of *imlk*, *ACML*, *ATLAS* or *OpenBLAS*. Is there any other to be supported?

  We could also have an easyblock option like `force_bla_vendor` so this can be overridden in the easyconfig. What do you think? 

  Is there a better way to find which BLAS is used in the toolchain? For example something like `toolchain.options['blas_vendor']`

- About Intel MKL version
  - `Intel10_32` (*Intel mkl v10 32 bit*) -> Does EasyBuild support 32 bit build? How to check?
  - `Intel10_64lp` (*Intel mkl v10 64 bit,lp thread model, lp64 model*) -> The one we use for IMKL >= 10
  - `Intel10_64lp_seq` (*Intel mkl v10 64 bit,sequential code, lp64 model*) -> NO, SuperLU requires multithread
  - `Intel` (*older versions of mkl 32 and 64 bit*) -> The one we use for IMKL < 10

  I assume/hope that `Intel10_64lp` also work for Intel MKL 11.x. SuperLU is compiling fine and all tests run fine. For the `FindBLAS.cmake` script, it looks like this version difference is only use to choose with library name to look for. 

- About `pic`, I have added this to the easyconfig files in PR https://github.com/hpcugent/easybuild-easyconfigs/pull/2665

- About the license, I agree with the contributor agreement

boegel:
@besserox:

* you've covered the main BLAS libraries with those, I'd say that's OK for now (imkl and OpenBLAS are by far the most prevalent);
* for now, we don't have something better than `get_software_root` to check for the BLAS library being used; I guess it would be useful to have the equivalent to `self.toolchain.mpi_family` which is already there... We do have `self.toolchain.definition()`, which returns a dictionary with toolchain components. If there's a BLAS in the toolchain, you could use `self.toolchain.definition()['BLAS']`, which would provide a list of names for the BLAS library (should be only one entry for BLAS)
So, I guess you can consider using something like:

```python
toolchain_blas = self.toolchain.definition().get('BLAS', ['NO_BLAS_IN_TOOLCHAIN'])[0]
if toolchain_blas == 'OpenBLAS':
    ...
elif ...
```
*

boegel:
@besserox: we do support installing the 32-bit version of imkl, but it's not easy to detect that currently... With the exception of an ancient toolchain, all imkl's we use in toolchains are 64-bit.

besserox:
@boegel So I try what you suggested, it seems to work well. So the behavior slightly changed now:

- If the toolchain has a BLAS library, it forces to uses it. If for some reason it could not find it, the configure/build should fail.

- If the toolchain has no BLAS library, it will try it best its best and look for one installed in the system. This allows to get it to work with GCC toolchain for example.

easybuild/easyblocks/s/superlu.py
```diff
@@ -0,0 +1,148 @@
...
+        else:
+            self.cfg.update('configopts', '-DCMAKE_POSITION_INDEPENDENT_CODE=OFF')
+
+        # Make sure not to build the slow BLAS library included in the package
+        self.cfg.update('configopts', '-Denable_blaslib=OFF')
+
+        # Set the BLAS library to use
+        # For this, use the BLA_VENDOR option from the FindBLAS module of CMake
+        # Check for all possible values at https://cmake.org/cmake/help/latest/module/FindBLAS.html
+        toolchain_blas = self.toolchain.definition().get('BLAS', ['NO_BLAS_IN_TOOLCHAIN'])[0]
```
boegel: my mistake, let's not use a random string like `'NO_BLAS_IN_TOOLCHAIN'`, use `None` instead (also below)

easybuild/easyblocks/s/superlu.py
```diff
@@ -0,0 +1,148 @@
...
+            self.cfg.update('configopts', '-DBLA_VENDOR="ACML"')
+        elif toolchain_blas == 'ATLAS':
+            self.cfg.update('configopts', '-DBLA_VENDOR="ATLAS"')
+        elif toolchain_blas == 'OpenBLAS':
+            # Unfortunately, OpenBLAS is not recognized by FindBLAS from CMake,
+            # we have to specify the OpenBLAS library manually
+            self.cfg.update('configopts', '-DBLAS_LIBRARIES="${EBROOTOPENBLAS}/lib/libopenblas.a;-pthread"')
+        elif toolchain_blas == 'NO_BLAS_IN_TOOLCHAIN':
+            # This toolchain has no BLAS library, let CMake find one
+            self.cfg.update('configopts', '-DBLA_VENDOR="All"')
```
boegel: I'm not in favour of this, it goes against the spirit of EB to try and be independent of what the OS provides (and what could disappear beyond our control at any given time).

I would drop this block, and just let is pass through to error out below.

easybuild/easyblocks/s/superlu.py
```diff
@@ -0,0 +1,148 @@
...
+                # 'Intel' -> For older versions of mkl 32 and 64 bit
+                self.cfg.update('configopts', '-DBLA_VENDOR="Intel"')
+        elif toolchain_blas == 'ACML':
+            self.cfg.update('configopts', '-DBLA_VENDOR="ACML"')
+        elif toolchain_blas == 'ATLAS':
+            self.cfg.update('configopts', '-DBLA_VENDOR="ATLAS"')
+        elif toolchain_blas == 'OpenBLAS':
+            # Unfortunately, OpenBLAS is not recognized by FindBLAS from CMake,
+            # we have to specify the OpenBLAS library manually
+            self.cfg.update('configopts', '-DBLAS_LIBRARIES="${EBROOTOPENBLAS}/lib/libopenblas.a;-pthread"')
```
boegel: use `get_software_libdir('OpenBLAS')` instead of `${EBROOTOPENBLAS}/lib`? That's clearer (full path specified), and it deals with the possibility that the library is actually in `lib64` (for whatever reason).

easybuild/easyblocks/s/superlu.py
```diff
@@ -0,0 +1,148 @@
...
+
+@author: Xavier Besseron (University of Luxembourg)
+"""
+
+import os
+from distutils.version import LooseVersion
+
+import easybuild.tools.environment as env
+from easybuild.easyblocks.generic.cmakemake import CMakeMake
+from easybuild.easyblocks.generic.configuremake import ConfigureMake
```
boegel: not used?

easybuild/easyblocks/s/superlu.py
```diff
@@ -0,0 +1,148 @@
...
+"""
+EasyBuild support for building and installing the SuperLU library, implemented as an easyblock
+
+@author: Xavier Besseron (University of Luxembourg)
+"""
+
+import os
+from distutils.version import LooseVersion
+
+import easybuild.tools.environment as env
```
boegel: not used?

easybuild/easyblocks/s/superlu.py
```diff
@@ -0,0 +1,148 @@
...
+from easybuild.easyblocks.generic.cmakemake import CMakeMake
+from easybuild.easyblocks.generic.configuremake import ConfigureMake
+from easybuild.framework.easyconfig import CUSTOM
+from easybuild.tools.build_log import EasyBuildError
+from easybuild.tools.systemtools import get_shared_lib_ext
+from easybuild.tools.modules import get_software_root,get_software_version
```
boegel: style nitpicking: space after `,`

easybuild/easyblocks/s/superlu.py
```diff
@@ -0,0 +1,148 @@
...
+        if self.cfg['build_shared_libs']:
+            self.cfg.update('configopts', '-DBUILD_SHARED_LIBS=ON')
+        else:
+            self.cfg.update('configopts', '-DBUILD_SHARED_LIBS=OFF')
+
+        # Add -fPIC flag if necessary
+        if self.toolchain.options['pic']:
+            self.cfg.update('configopts', '-DCMAKE_POSITION_INDEPENDENT_CODE=ON')
+        else:
+            self.cfg.update('configopts', '-DCMAKE_POSITION_INDEPENDENT_CODE=OFF')
```
boegel: you can make this shorter/cleaner:

```python
pic = ('OFF', 'ON')[self.toolchain.options['pic']]
self.cfg.update('configopts', '-DCMAKE_POSITION_INDEPENDENT_CODE=%s' % pic)
```

The trick here is using a boolean as an index (False: 0, True: 1).

easybuild/easyblocks/s/superlu.py
```diff
@@ -0,0 +1,148 @@
...
+                # It should work for Intel MKL 10 and above, as long as the library names stay the same
+                # SuperLU requires thread, 'Intel10_64lp_seq' will not work!
+                self.cfg.update('configopts', '-DBLA_VENDOR="Intel10_64lp"')
+            else:
+                # 'Intel' -> For older versions of mkl 32 and 64 bit
+                self.cfg.update('configopts', '-DBLA_VENDOR="Intel"')
+        elif toolchain_blas == 'ACML':
+            self.cfg.update('configopts', '-DBLA_VENDOR="ACML"')
+        elif toolchain_blas == 'ATLAS':
+            self.cfg.update('configopts', '-DBLA_VENDOR="ATLAS"')
```
boegel: you can collapse this to:

```python
 elif toolchain_blas in ['ACML', 'ATLAS']:
    self.cfg.update('configopts', '-DBLA_VENDOR="%s"' % toolchain_blas)
```

I also prefer using an empty line above the `elif`/`else`, it helps with readability.

easybuild/easyblocks/s/superlu.py
```diff
@@ -0,0 +1,148 @@
...
+            # This BLAS library is not supported yet
+            raise EasyBuildError("BLAS library '%s' is not supported yet" % toolchain_blas, err)
+
+        super(EB_SuperLU, self).configure_step()
+
+    def test_step(self):
+        """
+        Run the testsuite of SuperLU
+        """
+        self.cfg['runtest'] = "test"
```
boegel: this hard forces always running the tests, maybe you only want to set this if `runtest` was not specified in the easyconfig file (so you can disable running the tests from there):

```python
if self.cfg['runtest'] is None:
     self.cfg['runtest'] = 'test'
```

easybuild/easyblocks/s/superlu.py
```diff
@@ -0,0 +1,148 @@
...
+        Custom install procedure for SuperLU
+        """
+        super(EB_SuperLU, self).install_step()
+
+        if self.cfg['build_shared_libs']:
+            lib_ext = get_shared_lib_ext()
+        else:
+            lib_ext = "a"
+        expected_libpath = os.path.join(self.installdir, "lib", "libsuperlu.%s" % lib_ext)
+        actual_libpath   = os.path.join(self.installdir, "lib", "libsuperlu_%s.%s" % (self.cfg['version'],lib_ext) )
```
boegel: space after `,`, no need for space before last `)`

easybuild/easyblocks/s/superlu.py
```diff
@@ -0,0 +1,148 @@
...
+
+        if not os.path.exists(expected_libpath):
+            try:
+                os.symlink(actual_libpath, expected_libpath)
+            except OSError as err:
+                raise EasyBuildError("Failed to create symlink '%s' -> '%s" % (expected_libpath,actual_libpath), err)
```
boegel: space after `,`, and you actually want this (not the extra `%s`):

```python
raise EasyBuildError("Failed to create symlink '%s' -> '%s: %s", expected_libpath, actual_libpath, err)
```

easybuild/easyblocks/s/superlu.py
```diff
@@ -0,0 +1,148 @@
...
+
+    def sanity_check_step(self):
+        """
+        Check for main library files for SuperLU
+        """
+        if self.cfg['build_shared_libs']:
+            lib_ext = get_shared_lib_ext()
+        else:
+            lib_ext = "a"
```
boegel: copy-pasted, so let's do this higher up and define `self.lib_ext`?

easybuild/easyblocks/s/superlu.py
besserox: @boegel  BTW, the usage of `get_software_libdir()` break the `--extended-dry-run` functionality. In the sense that, because of this, it reports 
```
!!!
!!! WARNING: ignoring error 'NoneType' object has no attribute 'startswith'
!!!
```
instead of the actual configure command line.

Any way to fix this? (not here, but in a general way)

besserox:
@boegel I fixed your remarks

boegel:
@besserox informed me that this is tied to #863 (SuiteSparse easyblock update), which needs a bit of attention, so I'm retargeting this until after the EB v2.7.0 release

boegel:
@besserox the window for EasyBuild v2.8.0 is closing this week, any updates on this?

besserox:
@boegel I have not re-tested, but it should be ready to go

boegel:
looks great, testing the easyconfigs in https://github.com/hpcugent/easybuild-easyconfigs/pull/2665...

boegel:
Successful test reports in https://github.com/hpcugent/easybuild-easyconfigs/pull/2665, so good to go.

Thanks @besserox!

PR #861, opened 2016-03-09
hunt for usable 'python' command in PythonPackage easyblock when system Python is used

boegel:
This is required to fix the problem with bootstrapping EasyBuild in an environment where Python 3.x is the default `python`, and where `python2` is also available.

### DIFF
--- a/easybuild/easyblocks/generic/pythonpackage.py
+++ b/easybuild/easyblocks/generic/pythonpackage.py
@@ -254,14 +254,18 @@ class PythonPackage(ExtensionEasyBlock):
     def prepare_python(self):
         """Python-specific preperations."""
         # pick 'python' command to use
+        python = None
+
         python_root = get_software_root('Python')
         # keep in mind that Python may be listed as an allowed system dependency,
         # so just checking Python root is not sufficient
-        bin_python = os.path.join(python_root, 'bin', 'python')
-        if python_root and os.path.exists(bin_python) and os.path.samefile(which('python'), bin_python):
-            # if Python is listed as a (build) dependency, use 'python' command provided that way
-            python = os.path.join(python_root, 'bin', 'python')
-        else:
+        if python_root:
+            bin_python = os.path.join(python_root, 'bin', 'python')
+            if os.path.exists(bin_python) and os.path.samefile(which('python'), bin_python):
+                # if Python is listed as a (build) dependency, use 'python' command provided that way
+                python = os.path.join(python_root, 'bin', 'python')
+
+        if python is None:
             # if using system Python, go hunting for a 'python' command that satisfies the requirements
             python = pick_python_cmd(req_maj_ver=self.cfg['req_py_majver'], req_min_ver=self.cfg['req_py_minver'])
 
--- a/easybuild/easyblocks/generic/pythonpackage.py
+++ b/easybuild/easyblocks/generic/pythonpackage.py
@@ -128,7 +128,7 @@ def pick_python_cmd(req_maj_ver=None, req_min_ver=None):
                 res = python_cmd
             else:
                 res = which(python_cmd)
-            log.debug("Absolute path to retained Python command: %s", python_cmd)
+            log.debug("Absolute path to retained Python command: %s", res)
             break
         else:
             log.debug("Python command '%s' does not satisfy version requirements (maj: %s, min: %s), moving on",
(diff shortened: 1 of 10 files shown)


boegel:
@wpoely86: please review

easybuild/easyblocks/generic/pythonpackage.py
```diff
@@ -44,29 +45,95 @@
...
+        if req_maj_ver is not None:
+            # (strict) check for major version
+            out, _ = run_cmd("%s -c 'import sys; print sys.version_info[0]'" % python_cmd, simple=False)
+            if out.strip() != str(req_maj_ver):
+                return False
+
+            if req_min_ver is not None:
+                # check minor version
+                pycode = 'import sys; print "%s.%s" % sys.version_info[:2]'
+                out, _ = run_cmd("%s -c '%s'" % (python_cmd, pycode), simple=False)
```
wpoely86: maybe do 1 python call and use zero for `req_min_ver` if it's not specified?

easybuild/easyblocks/generic/pythonpackage.py
```diff
@@ -169,6 +240,26 @@ def set_pylibdirs(self):
...
+        if python_root and os.path.samefile(which('python'), os.path.join(python_root, 'bin', 'python')):
+            # if Python is listed as a (build) dependency, use 'python' command provided that way
+            python = 'python'
+        else:
+            # if using system Python, go hunting for a 'python' command that satisfies the requirements
+            python = pick_python_cmd(req_maj_ver=self.cfg['req_py_majver'], req_min_ver=self.cfg['req_py_minver'])
+
+        if python:
+            self.python_cmd = python
+            self.log.info("Python command being used: %s", self.python_cmd)
```
wpoely86: show full path to python?

easybuild/easyblocks/generic/pythonpackage.py
boegel: ok, makes sense, will do

wpoely86:
:+1:

easybuild/easyblocks/generic/pythonpackage.py
```diff
@@ -345,8 +466,10 @@ def sanity_check_step(self, *args, **kwargs):
         """
         Custom sanity check for Python packages
         """
-        if not 'exts_filter' in kwargs:
-            kwargs.update({'exts_filter': EXTS_FILTER_PYTHON_PACKAGES})
+        if 'exts_filter' not in kwargs:
+            orig_exts_filter = EXTS_FILTER_PYTHON_PACKAGES
+            exts_filter = (orig_exts_filter[0].replace('python', self.python_cmd), orig_exts_filter[1])
```
wpoely86: why is this needed?

boegel: for extensions being used with the system Python, we need to make sure we use the right Python command when doing `import` tests (which may be `python2` rather than `python`)

this is important for installing the GC3Pie bundle, for example (https://github.com/hpcugent/easybuild-easyconfigs/blob/master/easybuild/easyconfigs/g/GC3Pie/GC3Pie-2.4.2.eb)

boegel:
This is good to go, I've thoroughly tested this by rebuilding the following:

* EasyBuild bootstrap (e.g. on Arch Linux where `python` is Python 3.x)
* GC3Pie easyconfigs (on SL6, CentOS7 and Arch Linux)
* PyQt-4.11.4-intel-2015a-Python-2.7.9.eb
* SIP-4.16.8-intel-2015a-Python-2.7.9.eb
* numpy-1.10.4-intel-2016a-Python-2.7.11.eb
* scipy-0.17.0-intel-2016a-Python-2.7.11.eb
* vsc-base-1.7.3.eb
* vsc-mympirun-3.4.2.eb
* vsc-mympirun-scoop-3.4.1-intel-2015b-Python-2.7.10.eb
* vsc-processcontrol-1.0.eb
* vsc-processcontrol-1.0-vsc-base-2.1.2.eb
* libxml2-2.9.3-intel-2016a-Python-2.7.11.eb
* netcdf4-python-1.1.8-intel-2015b-Python-2.7.10.eb
* VSC-tools-0.1.2-ictce-5.3.0-Python-2.7.3.eb

boegel:

PR #862, opened 2016-03-09
SuiteSparse version 4.5+ makes shared objects and headers in lib/ and include/

hajgato:


boegel:
@hajgato shall we include the `MY_METIS_LIB` stuff from https://github.com/hpcugent/easybuild-easyconfigs/pull/2669/ too while we're at it?

PR #863, opened 2016-03-09
Update SuiteSparse easyblock for version >= 4.5

besserox:
The lastest version of SuiteSparse is build shared libray. 

This PR
- Set the `LD_LIBRARY_PATH` to the new directory where the libraries are located
- Add these new libraries to the sanity check


boegel:
@besserox how does this relate to #862?

cc @hajgato

boegel:
Jenkins: ok to test

hajgato:
@besserox @boegel This seems basically better than mine (which does not work at all), but you have to make sure that `MY_METIS_LIB` environmental variable is pointing to `libmetis.so` during the configure, build, and install steps; otherwise the bunded one will be built and used. (from `SuiteSparse` version 4.5.0)

hajgato:
plus add the locaton of `include` directory to `CPATH`

boegel:
@besserox Are you up for merging what @hajgato was working on in #862 in here?

besserox:
@boegel Yes, I can do this tomorrow

besserox:
@hajgato This appears to work for me. Do you want review and test?

hajgato:
@besserox For me it is fine, if it works. (The bundled `METIS` does not build.)

besserox:
@hajgato Yes, apparently the bundled METIS does not build:
- the `metis-5.1.0/build` sub-directory is empty
- there is no `libmetis.so` in `lib` (as there was before)

besserox:
@boegel Can we make sure a file does not exist during the sanity check?

Here we would like to make sure that there is no `libmetis.so` in the `lib` directory.

boegel:
@besserox not in a generic way, but you can just check that yourself when overriding `sanity_check_step`?

```python
def sanity_check_step(self):
    """Custom sanity check for SuiteSparse."""
    if os.path.exists(os.path.join(self.installdir, 'lib', 'libmetis.so')):
        raise EasyBuildError("<some meaningful error message>")

    ...
```

easybuild/easyblocks/s/suitesparse.py
```diff
@@ -162,13 +172,21 @@ def install_step(self):
...
+        else:
+            ld_library_path = ['UMFPACK/Lib', 'AMD/Lib']
+            include_dirs = [self.config_name]
+
         guesses = super(EB_SuiteSparse, self).make_module_req_guess()
         guesses.update({
-            'CPATH': [self.config_name],
-            'LD_LIBRARY_PATH': ['UMFPACK/Lib', 'AMD/Lib'],
+            'CPATH': include_dirs,
+            'LD_LIBRARY_PATH': ld_library_path,
```
boegel: @besserox also update `'LIBRARY_PATH'`? cfr. #885

boegel:
@besserox the window for EasyBuild v2.8.0 is closing this week, any updates on this?

easybuild/easyblocks/s/suitesparse.py
boegel: @besserox please also consider the change request in #882 when tackling the remark above

besserox:
@boegel I have included your last comments

easybuild/easyblocks/s/suitesparse.py
```diff
@@ -162,30 +172,48 @@ def install_step(self):
...
-            'CPATH': [self.config_name],
-            'LD_LIBRARY_PATH': ['UMFPACK/Lib', 'AMD/Lib'],
-        })
+
+        # Previous versions of SuiteSparse used specific directories for includes and libraries
+        if LooseVersion(self.version) < LooseVersion('4.5'):
+            include_dirs = [self.config_name]
+            ld_library_path = ['AMD/lib', 'BTF/lib', 'CAMD/lib', 'CCOLAMD/lib', 'CHOLAMD/lib', 'CHOLMOD/lib',
+                               'COLAMD/lib/', 'CSparse/lib', 'CXSparse/lib', 'KLU/lib', 'LDL/lib', 'RBio/lib',
+                               'UMFPACK/lib', 'SuiteSparse_config']
```
boegel: @besserox use `self.config_name` rather than hardcoding `SuiteSparse_config`?

easybuild/easyblocks/s/suitesparse.py
```diff
@@ -70,24 +72,32 @@ def configure_step(self):
...
         metis = get_software_root('METIS')
         parmetis = get_software_root('ParMETIS')
         if parmetis:
             metis_path = parmetis
-            metis_libs = ' '.join([
-                os.path.join(parmetis, 'lib', 'libparmetis.a'),
-                os.path.join(parmetis, 'lib', 'metis.a'),
-            ])
+            metis_include = os.path.join( parmetis, 'include')
+            metis_libs = os.path.join(parmetis, get_software_libdir('ParMETIS'), 'libmetis.a')
```
boegel: @besserox sure about this bit? no more `libparmetis.a`?

easybuild/easyblocks/s/suitesparse.py
```diff
@@ -162,30 +172,48 @@ def install_step(self):
...
+          raise EasyBuildError("SuiteSparse has compiled its own Metis. This will conflict with the Metis build."
+                               " The SuiteSparse EasyBlock need to be updated!")
+
         if LooseVersion(self.version) < LooseVersion('4.0'):
             csparse_dir = 'CSparse3'
         else:
             csparse_dir = 'CSparse'
 
+        libnames = ["AMD", "BTF", "CAMD", "CCOLAMD", "CHOLMOD", "COLAMD", "CXSparse", "KLU", "LDL", "RBio", "SPQR", "UMFPACK"]
+        libs = [os.path.join(x, 'lib', 'lib%s.a' % x.lower()) for x in libnames] + [os.path.join(csparse_dir, 'lib', 'libcsparse.a')]
```
boegel: lines too long...

boegel:
@besserox fixes for style remarks in https://github.com/besserox/easybuild-easyblocks/pull/7

One open question on `libparmetis.a` though...

easybuild/easyblocks/s/suitesparse.py
besserox: I believe ParMETIS builds libmetis.a and libparmetis.a

boegel: Yes, but don't you want to instruct `SuiteSparse` to link to `libparmetis.a`? (the link to `libmetis.a` is then also required to resolve all symbols, I think)

besserox: I don't think `SuiteSparse` knows anything about `ParMETIS`. It cannot link to it. It only depends on `METIS`. And depending on `ParMETIS` is an alternative way to provide `METIS`.

A look at the README.txt of `SuiteSparse` only mentions `METIS`, never `ParMETIS`.

boegel: OK, I'll see if anything pops up when retesting the existing SuiteSparse easyconfigs then...

besserox:
@boegel Updated to set BLAS and LAPACK properly on the build command line (cf https://github.com/hpcugent/easybuild-easyconfigs/pull/2766)

I found another bug at the same time.

easybuild/easyblocks/s/suitesparse.py
```diff
@@ -122,7 +124,7 @@ def configure_step(self):
         if cfgvars:
             try:
                 f = open(fp, "a")
-                f.write("# lines below added automatically by EasyBuild")
+                f.write("# lines below added automatically by EasyBuild\n")
                 for (var, val) in cfgvars.items():
                     f.write("%s = %s\n" % (var, val))
                 f.close()
```
boegel: this whole block could be cleaned up quite a bit with:

```python
cfgtxt = '\n'.join(["%s = %s" % (var, val) for (var, val) in cfgvars.items()])
write_file(fp, cfgtxt, append=True)
```

You can also move it out of the `try`/`except`, since `write_file` already takes care catching `IOError`

easybuild/easyblocks/s/suitesparse.py
```diff
@@ -122,7 +124,7 @@ def configure_step(self):
         if cfgvars:
             try:
                 f = open(fp, "a")
-                f.write("# lines below added automatically by EasyBuild")
+                f.write("# lines below added automatically by EasyBuild\n")
```
boegel: oh crap, this is pretty bad :)

one of the entries in `cfgvars` was being ignored, and to top it off, it's a random one (ordering of `cfgvars.items()` list is random)

besserox: Yes. I saw it when I removed the LAPACK and BLAS variables, and thanks to the sanity check on libmetis.so not being built.

boegel:
retested latest version of this both with all existing SuiteSparse easyconfigs and https://github.com/hpcugent/easybuild-easyconfigs/pull/3050, works like a charm

going in, thanks a lot for your efforts @besserox!

PR #864, opened 2016-03-10
Fixed the checking of rpmrebuild

damianam:
https://github.com/hpcugent/easybuild-framework/issues/1587#issuecomment-194744938

### DIFF
--- a/easybuild/easyblocks/generic/rpm.py
+++ b/easybuild/easyblocks/generic/rpm.py
@@ -47,7 +47,6 @@ from easybuild.framework.easyconfig import CUSTOM
 from easybuild.tools.build_log import EasyBuildError
 from easybuild.tools.filetools import which
 from easybuild.tools.run import run_cmd
-from easybuild.tools.systemtools import check_os_dependency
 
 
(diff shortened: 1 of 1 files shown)


boegel:
Jenkins: ok to test

easybuild/easyblocks/generic/rpm.py
```diff
@@ -45,6 +45,7 @@
 from easybuild.easyblocks.generic.binary import Binary
 from easybuild.framework.easyconfig import CUSTOM
 from easybuild.tools.build_log import EasyBuildError
+from easybuild.tools.filetools import which
 from easybuild.tools.run import run_cmd
 from easybuild.tools.systemtools import check_os_dependency
```
boegel: drop this import?

damianam: It is needed here in line 125. But I think that one can be also replaced by `which`. Would you agree?

boegel: Yes, I see no reason not to replace that too.

boegel:
lgtm (and retested), going in, thanks @damianam!

PR #866, opened 2016-03-11
Add PGI support to imkl easyblock

ocaisa:
Depends on ~~hpcugent/easybuild-framework#1342~~

### DIFF
--- a/easybuild/easyblocks/i/imkl.py
+++ b/easybuild/easyblocks/i/imkl.py
@@ -347,7 +347,7 @@ class EB_imkl(IntelBase):
                 if get_software_root('GCC'):
                     compsuff = '_gnu'
                 elif get_software_root('PGI'):
-                    compopt = 'compiler=pgi'
+                    compopt = '_pgi'
                 else:
                     raise EasyBuildError("Not using Intel/GCC/PGI, don't know compiler suffix for FFTW libraries.")
 
--- a/easybuild/easyblocks/i/imkl.py
+++ b/easybuild/easyblocks/i/imkl.py
@@ -300,6 +300,12 @@ class EB_imkl(IntelBase):
                     except OSError, err:
                         raise EasyBuildError("Can't change to interface %s directory %s: %s", lib, intdir, err)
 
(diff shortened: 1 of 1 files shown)


ocaisa:
Not so trivial, no PGI support for `fftw2x_cdft`. I can fix it with a simple patch on their makefile or we could leave it out?

boegel:
@ocaisa The toolchain support for the FFTW wrappers in Intel MKL (`intelfftw.py`) will check whether the FFTW wrapper are there, but it seems like it only checks the `fftw3` stuff...

ocaisa:
@boegel Success! This works now.

easybuild/easyblocks/i/imkl.py
```diff
@@ -298,6 +301,14 @@ def post_install_step(self):
                     except OSError, err:
                         raise EasyBuildError("Can't change to interface %s directory %s: %s", lib, intdir, err)
 
+                    if lib in cdftlibs and get_software_root('PGI') and lib not in patched:
+                        # Have to patch the makefile for this library
+                        run_cmd("sed -i s/gnu/pgi/g makefile", log_all=True, simple=True)
+                        run_cmd("sed -i s/gcc/pgcc/g makefile", log_all=True, simple=True)
+                        run_cmd("sed -i s/-Wall//g makefile", log_all=True, simple=True)
+                        run_cmd("sed -i s/-std=c99/-c99/g makefile", log_all=True, simple=True)
```
boegel: these should be done with `apply_regex_substitutions`?

@ocaisa: I can try and tackle that and send you a PR, or wrap your existing PR...

ocaisa: regex not my strong point, happy to take a PR

boegel:
tested with imkl easyconfig using `pompi` toolchain from https://github.com/hpcugent/easybuild-easyconfigs/pull/2900 (on top of https://github.com/hpcugent/easybuild-framework/pull/1664 and https://github.com/hpcugent/easybuild-framework/pull/1724), and with several exsting imkl easyconfigs, works great

So, good to go, thanks for figuring this out @ocaisa!

PR #868, opened 2016-03-11
fix extending list of binaries in MPICH easyblock

boegel:
fix for bug introduced in #852 

reported by @rjeschmi


boegel:
I guess @damianam gets to keep claiming he doesn't know Python that well ;)

damianam:
I guess I do :-P. But this has more to do with not paying attention than knowing Python. Either way, my fault.

PR #870, opened 2016-03-15
change suffix of original file to .easybuild when using fileinput in impi easyblock

ocaisa:
The .eb extension is a bad choice because it actually means something


ocaisa:
This messes up scripts I have that automatically pull easyconfigs from existing installations

boegel:
@ocaisa looks ok to include, but your script shouldn't be picked up `.eb` files from random locations?

ocaisa:
I never said it was my script...

boegel:
going in, thanks @ocaisa

PR #872, opened 2016-03-16
fix test_step in UFC easyblock

boegel:
the UFC easyblock has been broken for a while (since #698 got merged), but this was not considered a big issue since UFC is deprecated (cfr. https://bitbucket.org/fenics-project/ufc-deprecated)

however, since the fix is pretty trivial...

### DIFF
--- a/easybuild/easyblocks/u/ufc.py
+++ b/easybuild/easyblocks/u/ufc.py
@@ -79,6 +79,12 @@ class EB_UFC(CMakePythonPackage):
 
         super(EB_UFC, self).configure_step()
 
+    def test_step(self):
+        """No test suite available for UFC."""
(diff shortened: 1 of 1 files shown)


ehiggs:
The comment says it wants a string value. Should this return `''`?

boegel:
No, one of the superclasses excepts a string value for the `runtest` easyconfig parameter (`self.cfg['runtest']`) in it's `test_step`; we now bypass the parent's `test_step` entirely (which makes sense, since UFC doesn't provide a test suite)

ehiggs:
Then it seems good.

boegel:
This fixes the installing UFC *and* FFC on top, so well worth merging.

boegel:
Thanks for the review @ehiggs!

PR #876, opened 2016-03-26
don't hardcode Python version in test_make_module_pythonpackage

boegel:


boegel:
Jenkins: test this please

PR #878, opened 2016-03-29
EasyBlock for Amber14

thiell:
Modified from existing Amber12 EB
Originally installed with EasyBuild 2.3

### DIFF
--- a/easybuild/easyblocks/a/amber.py
+++ b/easybuild/easyblocks/a/amber.py
@@ -1,0 +1,223 @@
+##
+# Copyright 2009-2015 Ghent University
+# Copyright 2015-2016 Stanford University
+#
+# This file is part of EasyBuild,
+# originally created by the HPC team of Ghent University (http://ugent.be/hpc/en),
+# with support of Ghent University (http://ugent.be/hpc),
+# the Flemish Supercomputer Centre (VSC) (https://vscentrum.be/nl/en),
+# the Hercules foundation (http://www.herculesstichting.be/in_English)
+# and the Department of Economy, Science and Innovation (EWI) (http://www.ewi-vlaanderen.be/en).
+#
+# http://github.com/hpcugent/easybuild
+#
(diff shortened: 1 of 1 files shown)


boegel:
@thiell is this based on or related to #491?

boegel:
Jenkins: ok to test

thiell:
@boegel: Yes indeed, it's based on Amber12 EB found in #491. Maybe not the latest version though. It helped me for installing Amber 14 with EasyBuild. Even if not perfect yet, this may be useful to others and we wanted to share this.

boegel:
@thiell I'll pick up on this, since we just got an install request for Amber+AmberTools too...

I'll take it where you left off, check if there's anything missing from #491, and get it merge-ready.

Thanks a lot for sharing your work!

cc @valtandor

easybuild/easyblocks/a/amber.py
```diff
@@ -0,0 +1,223 @@
...
+        return ConfigureMake.extra_options(extra_vars)
+
+    def __init__(self, *args, **kwargs):
+        super(EB_Amber, self).__init__(*args, **kwargs)
+        self.already_extracted = False
+        self.build_in_installdir = True
+
+    def extract_step(self):
+        """Only extract from the tarball if this has not already been done."""
+        if not self.already_extracted:
```
boegel: @thiell, @valtandor: any idea why this is here at all?

thiell: @boegel you mean the already_extracted? I'm not sure at all sorry :) that doesn't too good indeed.

boegel: Yes, I was referring to the `already_extracted` conditional... I guess it's a leftover of earlier attempts by @valtandor.

I'm cleaning up this easyblock a bit, so I'll chuck this and see what happens.

boegel:
merged via #958, thanks @thiell!

PR #880, opened 2016-03-29
EasyBlock for OpenMM 6.3

thiell:
Originally installed with EasyBuild 2.3

### DIFF
--- a/easybuild/easyblocks/o/openmm.py
+++ b/easybuild/easyblocks/o/openmm.py
@@ -1,0 +1,132 @@
+##
+# Copyright 2009-2015 Ghent University
+# Copyright 2015-2016 Stanford University
+#
+# This file is part of EasyBuild,
+# originally created by the HPC team of Ghent University (http://ugent.be/hpc/en),
+# with support of Ghent University (http://ugent.be/hpc),
+# the Flemish Supercomputer Centre (VSC) (https://vscentrum.be/nl/en),
+# the Hercules foundation (http://www.herculesstichting.be/in_English)
+# and the Department of Economy, Science and Innovation (EWI) (http://www.ewi-vlaanderen.be/en).
+#
+# http://github.com/hpcugent/easybuild
+#
+# EasyBuild is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation v2.
+#
+# EasyBuild is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with EasyBuild.  If not, see <http://www.gnu.org/licenses/>.
+##
+"""
+EasyBuild support for building and installing OpenMM, implemented as an easyblock
+
+@author: Stephane Thiell (Stanford University)
+"""
+import os
+from distutils.version import LooseVersion
+
+import easybuild.tools.environment as env
+import easybuild.tools.toolchain as toolchain
+from easybuild.easyblocks.generic.cmakemake import CMakeMake
+from easybuild.framework.easyconfig import CUSTOM, MANDATORY
+from easybuild.tools.run import run_cmd
+from easybuild.tools.modules import get_software_root, get_software_version
+from easybuild.tools.build_log import EasyBuildError
+
+
+class EB_OpenMM(CMakeMake):
+    """Support for building/installing OpenMM."""
+
+    def __init__(self, *args, **kwargs):
+        """Initialisation of custom class variables for OpenMM."""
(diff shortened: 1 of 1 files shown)


boegel:
Jenkins: ok to test

easybuild/easyblocks/o/openmm.py
```diff
@@ -0,0 +1,132 @@
...
+        """Custom configuration procedure for OpenMM."""
+
+        if LooseVersion(self.version) < LooseVersion('6.3'):
+            self.log.warn("OpenMM easyblocks not tested for older version of OpenMM")
+
+        # build a release build
+        self.cfg.update('configopts', "-DCMAKE_BUILD_TYPE=Release")
+
+        #self.cfg.update('configopts', "-DOPENMM_BUILD_SHARED_LIB=OFF")
+        #self.cfg.update('configopts', "-DOPENMM_BUILD_STATIC_LIB=ON")
```
boegel: @thiell Is there missing logic to control this?

easybuild/easyblocks/o/openmm.py
```diff
@@ -0,0 +1,132 @@
...
+
+        super(EB_OpenMM, self).sanity_check_step(custom_paths=custom_paths)
+
+    def make_module_extra(self):
+        """Custom extra module file entries for OpenMM."""
+
+        txt = super(EB_OpenMM, self).make_module_extra()
+
+        txt += self.module_generator.set_environment("OPENMM_CUDA_COMPILER", "nvcc")
+        txt += self.module_generator.load_module("CUDA/%s" % get_software_version("CUDA"))
```
boegel: this is not needed, since the easyblock is enforcing that CUDA is listed as a dependency?

easybuild/easyblocks/o/openmm.py
```diff
@@ -0,0 +1,132 @@
...
+        cmd = "cd python; python setup.py build && python setup.py install %s" % args
+        run_cmd(cmd, log_all=True, simple=True, log_ok=True)
+
+    def sanity_check_step(self):
+        """Custom sanity check for OpenMM."""
+
+        custom_paths = {
+                        'files': ['lib/libOpenMM.so', 'lib/plugins/libOpenMMCUDA.so'],
+                        'dirs': ['lib', 'lib/plugins', 'include/openmm'],
+                       }
```
boegel: please fix indentation

easybuild/easyblocks/o/openmm.py
```diff
@@ -0,0 +1,132 @@
...
+        else:
+            raise EasyBuildError("CUDA not found")
+
+        doxygen = get_software_root('Doxygen')
+        if not doxygen:
+            raise EasyBuildError("Doxygen not found")
+
+        doxygen = get_software_root('SWIG')
+        if not doxygen:
+            raise EasyBuildError("SWIG not found")
```
boegel: this can be simplified a bit by using a for loop over the 3 required dependencies

easybuild/easyblocks/o/openmm.py
```diff
@@ -0,0 +1,132 @@
...
+
+
+class EB_OpenMM(CMakeMake):
+    """Support for building/installing OpenMM."""
+
+    def __init__(self, *args, **kwargs):
+        """Initialisation of custom class variables for OpenMM."""
+        super(EB_OpenMM, self).__init__(*args, **kwargs)
+
+        self.example = None
```
boegel: not needed at all, so the whole `__init__` block can be trashed

easybuild/easyblocks/o/openmm.py
```diff
@@ -0,0 +1,132 @@
...
+
+@author: Stephane Thiell (Stanford University)
+"""
+import os
+from distutils.version import LooseVersion
+
+import easybuild.tools.environment as env
+import easybuild.tools.toolchain as toolchain
+from easybuild.easyblocks.generic.cmakemake import CMakeMake
+from easybuild.framework.easyconfig import CUSTOM, MANDATORY
```
boegel: please remove unused imports

boegel:
@thiell: I played around with this a bit, but it doesn't seem to work for me... The configure step is failing with:

```
CMake Error: The following variables are used in this project, but they are set to NOTFOUND.                                                                                                                                                                               Please set them or make sure they are set and tested correctly in the CMake files:                                                                                                                                                                                         CUDA_CUDA_LIBRARY (ADVANCED)                                                                                                                                                                                                                                                   linked by target "OpenMMCUDA" in directory /tmp/vsc40023/easybuild_build/OpenMM/6.3/foss-2016a/openmm-6.3/platforms/cuda/sharedTarget  
```

Also, in your OpenMM easyconfig in https://github.com/hpcugent/easybuild-easyconfigs/issues/2798, `CUDA` is not listed as a dependency, while the easyblock enforces this?
It seems like that could be related to the problem I'm running into.

W.r.t. the style remarks & suggestions: I tackled those myself, see https://github.com/stanford-rc/easybuild-easyblocks/pull/1. If you merge than in, this PR will be updated.

Issue #882, opened 2016-04-08
SuiteSparse additional module paths


stachon:
To compile FreeFem++ with SuiteSparse, I needed to add additional module paths. Not sure if this should go into easyblock or easyconfig, this is what I have in easyconfig as a workaround for now:

```python
modextrapaths = {
  'LD_LIBRARY_PATH': ['SuiteSparse_config'],
  'CPATH': ['UMFPACK/include', 'AMD/include', 'SuiteSparse_config'],
  'LIBRARY_PATH': ['AMD/lib', 'BTF/lib', 'CAMD/lib', 'CCOLAMD/lib', 'CHOLAMD/lib', 'CHOLMOD/lib', 'COLAMD/lib/', 'CSparse/lib', 'CXSparse/lib', 'KLU/lib', 'LDL/lib', 'RBio/lib', 'UMFPACK/lib', 'SuiteSparse_config']
}
```

boegel:
@stachon since there is a custom easyblock specific to SuiteSparse, it should be fixed there, in the `make_module_req_guess` method that's already implemented there

Are you up for tackling that?

ocaisa:
For me the easyblock seems to be broken, it's not adding `lib` to `LD_LIBRARY_PATH`, only to `LIBRARY_PATH`, I think you are suffering from the same problem.

ocaisa:
#885 solved configuration problems I had with Octave, I think the suitesparse easyconfig may need a bot of an overhaul though

ocaisa:
@stachon Can you try the easyblock from #885 and see if it solves your problem, I think it will (unless you want a static build)

boegel:
The intention is now to fix this via #863

boegel:
now fixed in #863, should get merged soon

PR #885, opened 2016-04-11
CPATH and LD_LIBRARY_PATH missing defaults for SuiteSparse

ocaisa:
Default values were not included in `CPATH` and `LD_LIBRARY_PATH`


easybuild/easyblocks/s/suitesparse.py
```diff
@@ -167,8 +167,8 @@ def make_module_req_guess(self):
         """
         guesses = super(EB_SuiteSparse, self).make_module_req_guess()
         guesses.update({
-            'CPATH': [self.config_name],
-            'LD_LIBRARY_PATH': ['UMFPACK/Lib', 'AMD/Lib'],
+            'CPATH': ['include', self.config_name],
+            'LD_LIBRARY_PATH': ['lib', 'UMFPACK/Lib', 'AMD/Lib'],
```
boegel: how about changing it to:

```python
guesses['CPATH'].append(self.config_name)

extra_libs = ['AMD/Lib', 'UMFPACK/Lib']
guesses['LD_LIBRARY_PATH'].extend(extra_libs)
guesses['LIBRARY_PATH'].extend(extra_libs)
```

extra bonus points :cookie: if you also tackle the issue reported in #882 while you're at it :-)

ocaisa:
I think #882 might be directly resolved with this change, the problem before (at least for Octave) was that libraries were not being found because the `lib` directory was not in the LD path, if you looked at the config.log you could see it was simply failing to resolve libraries. Adding the additional paths was only a patch for it to resolve libraries since the static libs are still in those paths (I wonder if the static libs they should be moved to `lib` by the easyblock since that is where the so libs are).

boegel:
@ocaisa: see also #863 (mostly unrelated, but it would be nice to tackle the updates all in one go, to avoid having to deal with conflicts)

or, I should get #863 merged first, since the changes in here are fairly trivial

cc @besserox

besserox:
For me #863 is ready to be merged. A minor improvement could be an additional sanity check (cf my last comment in #863) but it can be a separate PR later.

Unfortunately, I won't have any time to look at or test anything in the next 2 weeks.

ocaisa:
I think #863 already fixed this issue...and does a much more comprehensive job, I'm closing.

PR #886, opened 2016-04-13
PGI: move localrc adjustments to post_install_step()

bartoldeman:
Together with self.load_dependency_modules() this makes really
sure the GCC version in localrc is correct.

The problem was that when the toolchain GNU/4.9.3-2.25 was used
the GNU module was loaded during installation but the GCCcore
dependency is not for the *GCC-4.9.3-2.25* easyconfigs.

### DIFF
--- a/easybuild/easyblocks/p/pgi.py
+++ b/easybuild/easyblocks/p/pgi.py
@@ -112,6 +112,13 @@ class EB_PGI(EasyBlock):
         cmd = "%s ./install" % ' '.join(['%s=%s' % x for x in sorted(pgi_env_vars.items())])
         run_cmd(cmd, log_all=True, simple=True)
 
+    def post_install_step(self):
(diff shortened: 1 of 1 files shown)


boegel:
@bartoldeman: it looks like something else is going on here... during `post_install_step`, dependencies should be loaded just fine

Which easyconfig were you using when you ran into this problem?

Did it include this:

```python
toolchain = {'name': 'dummy', 'version': 'dummy'}
```

rather than this?

```python
```

Due to a weird historical fluke, using `dummy` as a version for the `dummy` toolchain results in *not* loading the dependencies at build time...

boegel:
Jenkins: ok to test

bartoldeman:
You are spot on about the dummy. For some reason I had version as dummy in my test easyconfig but not in the submitted one, I had no idea it would cause this behaviour. I will close this PR and submit a new one for the libnuma issue.

PR #887, opened 2016-04-14
easyblock for Tau

boegel:
easyblock to simplify the complex easyconfig in https://github.com/hpcugent/easybuild-easyconfigs/pull/509, and deal with querying the toolchain

may need a bit of freshening up before it's good to go

cc @geimer, @berndmohr

### DIFF
--- a/easybuild/easyblocks/t/tau.py
+++ b/easybuild/easyblocks/t/tau.py
@@ -156,13 +156,9 @@ class EB_TAU(ConfigureMake):
         comp_fam = self.toolchain.comp_family()
         if comp_fam in known_compilers:
             self.cc, self.cxx, self.fortran = known_compilers[comp_fam]
-        else:
-            raise EasyBuildError("Compiler family not supported yet: %s", comp_fam)
 
         # determine values for MPI flags
         self.mpi_inc_dir, self.mpi_lib_dir = os.getenv('MPI_INC_DIR'), os.getenv('MPI_LIB_DIR')
-        if self.mpi_inc_dir is None or self.mpi_lib_dir is None:
-            raise EasyBuildError("Failed to determine MPI include/library paths, no MPI available in toolchain?")
 
         # determine value for optional packages option template
         self.opt_pkgs_opts = ''
@@ -209,6 +205,11 @@ class EB_TAU(ConfigureMake):
 
     def configure_step(self):
         """Custom configuration procedure for TAU: template configuration options before using them."""
+        if self.cc is None or self.cxx is None or self.fortran is None:
+            raise EasyBuildError("Compiler family not supported yet: %s", self.toolchain.comp_family())
+
+        if self.mpi_inc_dir is None or self.mpi_lib_dir is None:
+            raise EasyBuildError("Failed to determine MPI include/library paths, no MPI available in toolchain?")
 
         # inform which backend/variant is being handled
         backend = (['tau'] + self.cfg['extra_backends'])[self.variant_index // 3]
--- a/easybuild/easyblocks/t/tau.py
+++ b/easybuild/easyblocks/t/tau.py
@@ -176,7 +176,7 @@ class EB_TAU(ConfigureMake):
 
                     variant_label = 'tau'
                     # For non-GCC builds, the compiler name is encoded in the variant
-                    if self.cxx != 'g++':
+                    if self.cxx and self.cxx != 'g++':
                         variant_label += '-' + self.cxx
                     if get_software_root('PAPI'):
                         variant_label += '-papi'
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/t/tau.py
```diff
@@ -0,0 +1,262 @@
...
+        # number of iterations: # backends + 1 (for basic)
+        iter_cnt = len(self.cfg['extra_backends']) + 1
+
+        self.cfg['configopts'] = [mpi_tmpl, openmp_tmpl, hybrid_tmpl] * iter_cnt
+
+        # custom prefix option for configure command
+        self.cfg['prefix_opt'] = '-prefix='
+
+        # installation command is 'make install clean'
+        self.cfg['installopts'] = 'clean'
```
wpoely86: wtf?

easybuild/easyblocks/t/tau.py
```diff
@@ -0,0 +1,262 @@
...
+
+        # compose templates
+        mpi_tmpl = comp_opts + mpi_opts + opt_pkgs_opts + backend_opt
+        openmp_tmpl = comp_opts + openmp_opts + opt_pkgs_opts + backend_opt
+        hybrid_tmpl = comp_opts + openmp_opts + mpi_opts + opt_pkgs_opts + backend_opt
+
+        # number of iterations: # backends + 1 (for basic)
+        iter_cnt = len(self.cfg['extra_backends']) + 1
+
+        self.cfg['configopts'] = [mpi_tmpl, openmp_tmpl, hybrid_tmpl] * iter_cnt
```
wpoely86: maybe print (to debug) all of this?

easybuild/easyblocks/t/tau.py
```diff
@@ -0,0 +1,262 @@
...
+        backend = (['tau'] + self.cfg['extra_backends'])[self.variant_index // 3]
+        variant = ['mpi', 'openmp', 'hybrid'][self.variant_index % 3]
+        print_msg("starting with %s backend (%s variant)" % (backend, variant), log=self.log, silent=self.silent)
+
+    def prepare_step(self):
+        """Custom prepare step for Tau: check required dependencies and collect information on them."""
+        super(EB_TAU, self).prepare_step()
+
+        # install prefixes for selected backends
+        for dep in ['Scalasca', 'Score-P', 'Vampirtrace']:
```
wpoely86: put this together with the `KNOWN_BACKENDS` (as dict)

easybuild/easyblocks/t/tau.py
```diff
@@ -0,0 +1,262 @@
...
+            root = get_software_root(dep)
+            backend_name = dep.lower().replace('-', '')
```
wpoely86: or with this

easybuild/easyblocks/t/tau.py
boegel: :)

easybuild/easyblocks/t/tau.py
```diff
@@ -0,0 +1,262 @@
...
+            if root:
+                self.opt_pkgs_opts += ' -%s=%s' % (opt, root)
+
+        # determine list of labels, based on selected (extra) backends, variants and optional packages
+        self.variant_labels = []
+        backend_labels = ['', '-epilog-scalasca-trace', '-scorep', '-vampirtrace-trace']
+        for backend, backend_label in zip(['tau'] + KNOWN_BACKENDS, backend_labels):
+            #if backend != 'tau' and backend not in self.cfg['extra_backends']:
+            if backend == 'tau' or backend in self.cfg['extra_backends']:
+
```
wpoely86: less whitespace in this part?

geimer:
@boegel:  I noticed that you removed the `self.cfg['keeppreviousinstall'] = True` from the prepare step.  Was this intentional, i.e., is there another way to achieve the same behavior?  I'm asking because the PDT easyblock uses this technique as well...

boegel:
@geimer I redefined `make_installdir` as `pass` instead, which is cleaner imho

easybuild/easyblocks/t/tau.py
```diff
@@ -0,0 +1,278 @@
...
+        avail_makefiles = ['Makefile.' + l for l in self.variant_labels]
+        if self.cfg['tau_makefile'] not in avail_makefiles:
+            raise EasyBuildError("Specified tau_makefile %s will not be available (only: %s)",
+                                 self.cfg['tau_makefile'], avail_makefiles)
+
+        # create install directory and make sure it does not get cleaned up again in the install step;
+        # the first configure iteration already puts things in place in the install directory,
+        # so that shouldn't get cleaned up afterwards...
+        self.log.info("Creating install dir %s before starting configure-build-install iterations", self.installdir)
+        mkdir(self.installdir, parents=True)
```
boegel: TODO: remove install dir if it's already there (e.g. because of forced reinstallation)

boegel:
tested with https://github.com/hpcugent/easybuild-easyconfigs/pull/509, good to go now (finally...)

Thanks @geimer and @berndmohr!

PR #888, opened 2016-04-14
PGI: delete libnuma symbolic links in installation directory

bartoldeman:
If the PGI installer does NOT find an OS libnuma, makelocalrc creates
symbolic links to libpgnuma.so.
Since we use the EasyBuild libnuma, delete those symbolic links.

### DIFF
--- a/easybuild/easyblocks/p/pgi.py
+++ b/easybuild/easyblocks/p/pgi.py
@@ -121,6 +121,13 @@ class EB_PGI(EasyBlock):
 
         cmd = "%s -x %s -g77 /" % (filename, install_abs_subdir)
         run_cmd(cmd, log_all=True, simple=True)
+        
+        # if an OS libnuma is NOT found, makelocalrc creates symbolic links to libpgnuma.so
+        # since we use the EB libnuma, delete those symbolic links
+        for filename in ["libnuma.so", "libnuma.so.1"]:
+            path = os.path.join(install_abs_subdir, "lib", filename)
+            if os.path.islink(path):
+                os.remove(path)
 
     def sanity_check_step(self):
         """Custom sanity check for PGI"""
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/p/pgi.py
```diff
@@ -121,6 +121,13 @@ def install_step(self):
 
         cmd = "%s -x %s -g77 /" % (filename, install_abs_subdir)
         run_cmd(cmd, log_all=True, simple=True)
+        
+        # if an OS libnuma is NOT found, makelocalrc creates symbolic links to libpgnuma.so
+        # since we use the EB libnuma, delete those symbolic links
+        for filename in ["libnuma.so", "libnuma.so.1"]:
+            path = os.path.join(install_abs_subdir, "lib", filename)
+            if os.path.islink(path):
+                os.remove(path)
```
boegel: @bartoldeman: you should only do this if `numactl` is actually included as a dependency?

use `if get_software_root('numactl')` to check

boegel:
Jenkin
Jenkins: ok to test

easybuild/easyblocks/p/pgi.py
```diff
@@ -121,6 +121,14 @@ def install_step(self):
 
         cmd = "%s -x %s -g77 /" % (filename, install_abs_subdir)
         run_cmd(cmd, log_all=True, simple=True)
+        
+        # If an OS libnuma is NOT found, makelocalrc creates symbolic links to libpgnuma.so
+        # If we use the EB libnuma, delete those symbolic links to ensure they are not used
+        if get_software_root("numactl"):
```
boegel: you'll need to import this function too, i.e. add this on top (in alphabetical order compared to the ones already there):

```python
```

bartoldeman: Thanks and sorry for the mishap! Done, now tested correctly.

boegel:
Tested with https://github.com/hpcugent/easybuild-easyconfigs/pull/2879, looks good, go going in, thanks @bartoldeman!

PR #889, opened 2016-04-18
Score-P easyblock update (REVIEW)

geimer:
- [x] Fix bug passing `--with-libbfd` path
- [x] Add `libunwind` suppport
- [x] Add `SIONlib` support
- [x] Add IBM XL compiler support
- [x] Add PGI compiler support (depends on ~~https://github.com/hpcugent/easybuild-framework/pull/1342~~)
- [x] Add MVAPICH2 support
- [x] Add Cray support

### DIFF
--- a/easybuild/easyblocks/s/score_p.py
+++ b/easybuild/easyblocks/s/score_p.py
@@ -63,6 +63,7 @@ class EB_Score_minus_P(ConfigureMake):
                 toolchain.GCC: 'gcc',
                 toolchain.IBMCOMP: 'ibm',
                 toolchain.INTELCOMP: 'intel',
+                toolchain.PGI: 'pgi',
             }
             comp_fam = self.toolchain.comp_family()
             if comp_fam in comp_opts:
--- a/easybuild/easyblocks/s/score_p.py
+++ b/easybuild/easyblocks/s/score_p.py
@@ -1,5 +1,5 @@
 ##
-# Copyright 2013 Ghent University
+# Copyright 2013-2016 Ghent University
 #
 # This file is part of EasyBuild,
 # originally created by the HPC team of Ghent University (http://ugent.be/hpc/en),
@@ -73,7 +73,7 @@ class EB_Score_minus_P(ConfigureMake):
 
         # auto-detection for dependencies mostly works fine, but hard specify paths anyway to have full control
         deps = {
-            'binutils': ['--with-libbfd=%%s/%s' % get_software_libdir('binutils', fs=['libbfd.a'])],
+            'binutils': ['--with-libbfd=%s'],
             'Cube': ['--with-cube=%s/bin'],
             'CUDA': ['--with-libcudart=%s'],
             'OTF2': ['--with-otf2=%s/bin'],
--- a/easybuild/easyblocks/s/score_p.py
+++ b/easybuild/easyblocks/s/score_p.py
@@ -61,9 +61,14 @@ class EB_Score_minus_P(ConfigureMake):
         mpi_opts = {
             toolchain.INTELMPI: 'intel2',  # intel: Intel MPI v1.x (ancient); intelpoe: IBM POE MPI for Intel platforms
             toolchain.OPENMPI: 'openmpi',
-            toolchain.MPICH: 'mpich3',  # In EB terms, MPICH means MPICH 3.x; MPICH 1.x is ancient and unsupported
+            toolchain.MPICH: 'mpich3',     # In EB terms, MPICH means MPICH 3.x; MPICH 1.x is ancient and unsupported
             toolchain.MPICH2: 'mpich2',
+            toolchain.MVAPICH2: 'mpich3',  # Though MVAPICH2 is based on MPICH 3.x only since v1.9b, this setting only
+                                           # affects options passed to the MPI (Fortran) compiler wrappers.  And
+                                           # MPICH 3.x compatible options were already supported in MVAPICH2 1.7.
         }
+        # With minimal toolchains, packages using this easyblock may be built with a non-MPI toolchain (e.g., OTF2).
+        # In this case, skip passing the '--with-mpi' option.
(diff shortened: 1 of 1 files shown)


boegel:
Jenkins: ok to test

boegel:
@geimer Please ping me when you consider this done (or if you want me to take a close look & provide feedback).

boegel:
@geimer: I'm slowly closing the window for EasyBuild v2.8.0, PRs that don't make it in by the end of this week will have to wait until v2.9.0 (end of June'16); what's the status of this?

I've already tagged it for v2.9.0 since it's still marked as WIP, but I can reconsider that if you want.

geimer:
@boegel: Now that ~~https://github.com/hpcugent/easybuild-framework/pull/1342~~ has been merged, I could add the PGI compiler suite support.  Cray should also be OK (worked for me), but I'm still waiting for feedback from @jgphpc.  Anyway, I think it's ready for review now.

easybuild/easyblocks/s/score_p.py
```diff
@@ -38,49 +39,83 @@
...
+                    self.cfg.update('configopts', "--with-mpi=%s" % mpi_opts[mpi_fam])
+                else:
+                    raise EasyBuildError("MPI family %s not supported yet (only: %s)",
+                                         mpi_fam, ', '.join(mpi_opts.keys()))
 
-        # auto-detection for dependencies mostly works fine, but hard specify paths anyway to have full control
+        # Auto-detection for dependencies mostly works fine, but hard specify paths anyway to have full control
         deps = {
-            'binutils': ['--with-libbfd=%%s/%s' % get_software_libdir('binutils', fs=['libbfd.a'])],
+            'binutils': ['--with-libbfd=%s'],
```
boegel: @geimer: the only change that raises a question for me is this one...

There must have been a good reason we're using `get_software_libdir` here?

geimer: @boegel: There was no good reason; it simply was a bug that went by unnoticed for quite some time.  I've checked all Score-P releases back to v1.1 (Oct 2012) and the `configure` option always worked like this:
```
  --with-libbfd=(yes|no|<Path to libbfd installation>)
                          If you want to build with libbfd support but do not
                          have a libbfd in a standard location, you need to
                          explicitly specify the directory where it is
                          installed. On non-cross-compile systems we search
                          the system include and lib paths per default [yes];
                          on cross-compile systems, however, you have to
                          specify a path [no]. --with-libbfd is a shorthand
                          for --with-libbfd-include=<Path/include> and
                          --with-libbfd-lib=<Path/lib>. If these shorthand
                          assumptions are not correct, you can use the
                          explicit include and lib options directly.
  --with-libbfd-include=<Path to libbfd headers>
  --with-libbfd-lib=<Path to libbfd libraries>
```
That is, some absolute path of libbfd.a was passed in, the configure check tried to use it as a prefix for 'include' and 'lib', this obviously didn't help in making the compile/link check pass, and the compiler adapter fell back to using `nm` instead of libbfd. `configure` prints a big fat warning that the provided path does not exist, but it is so well hidden in the EB log file that basically nobody finds it...

boegel: @geimer: That sounds very reasonable, but... We're missing something.

On top of current `develop`, all builds using this `Score_minus_P` easyblock work fine; that includes: `Cube` (v4.x), `LWM2`, `OTF2`, `Scalasca` and `Score-P` itself.

On top of this updated easyblock however, two of the `Score-P` builds fail (`Score-P-1.2.1-gompi-1.4.12-no-OFED.eb` and `Score-P-1.2.3-goolf-1.5.14.eb`), while another one works fine `Score-P-1.4-foss-2015a.eb`... The other stuff using this easyblock works fine (`Cube` (v4.x), `LWM2`, `OTF2`, `Scalasca`).

So, it seems like we've introduced some kind of regression for Score-P 1.2.x?

Here's the relevant error message (reformatted for readability):

```
 == 2016-05-13 18:58:24,165 build_log.py:152 ERROR cmd " ./configure --prefix=...  --with-libbfd=$EBROOTBINUTILS " exited with exitcode 1 and output: 
...
checking bfd.h usability... yes                                                                                                                                                                                                                                           
checking bfd.h presence... yes                                                                                                                                                                                                                                            
checking for bfd.h... yes                                                                                                                                                                                                                                                 
: error: cannot find libbfd.a, libbfd.so, or libbfd.dylib                                                                                                                                                                                                                 
configure: error: ./configure failed for build-mpi  
```

Note that without this patch, it's not the absolute path of `libbfd.a` being passed, but the path to the directory in which `libbfd.a` resides, i.e. `--with-libbfd=$EBROOTBINUTILS/lib`.

This apparently works with Score-P 1.2.x and 1.4 (despite it being wrong according to the help info for `--with-libbfd` vs `--with-libbfd-lib`.

boegel: @geimer The original PR for the Score-P easyblock (https://github.com/hpcugent/easybuild-easyblocks/pull/304) mentions the need for specifying the `lib` subdir for `--with-libbfd` (and `--with-papi-lib`) rather than specifying the root, so it seems like we did have a good reason for this...

Could this have been a bug in the handling of `--with-libbfd` option in Score 1.2.x, and this behaviour being retained as a fallback in later versions for backward compatibility?

Maybe @berndmohr remembers why we needed to specify the path to the `/lib` subdirectory of the `binutils` install directory rather than the root install prefix as suggested with the help output for `--with-libbfd`?

My suggestion would be to put a version check in place for versions older than Score-P 1.4, and stick to using the `/lib` approach for those, to avoid breaking stuff?

boegel: @geimer: this workaround does the job for me: https://github.com/geimer/easybuild-easyblocks/pull/4

boegel:
tested last update, all good to go now, thanks @geimer!

PR #890, opened 2016-04-20
add easyblock for libQGLViewer (REVIEW)

jrbosch:
add easyblock for libQGLViewer

### DIFF
--- a/easybuild/easyblocks/l/libqglviewer.py
+++ b/easybuild/easyblocks/l/libqglviewer.py
@@ -30,7 +30,7 @@ EasyBuild support for building and installing libQGLViewer, implemented as an ea
 
 from easybuild.easyblocks.generic.configuremake import ConfigureMake
 from easybuild.tools.run import run_cmd
-
+from easybuild.tools.systemtools import get_shared_lib_ext
 
 class EB_libQGLViewer(ConfigureMake):
     """Support for building/installing libQGLViewer."""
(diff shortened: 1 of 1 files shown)


jrbosch:
@boegel , I have not been able to test this, because Qt version I include as dependency is 4.8.7 and it does not include GL/glu.h is necessary for libQGLViewer, see https://forum.qt.io/topic/13851/include-gl-glu-h-was-removed-from-qtopengl-qgl-h-while-upgrading-from-qt-4-7-3-to-4-8 I do not know if should I use a Qt oldest version (<= 4.7.3 that include GL/glu.h) in the libQGLViewer easyconfig as dependency or if there is another way to solve that problem?

boegel:
@jrbosch: see whether including libGLU 9.0.0 (which depends on Mesa) as a dependency to libQGLViewer fixes your problem

easybuild/easyblocks/l/libqglviewer.py
```diff
@@ -0,0 +1,100 @@
...
+            # setting am_cv_prog_tar_ustar avoids that configure tries to figure out (in this case qmake)
+            # which command should be used for tarring/untarring
+            # am__tar and am__untar should be set to something decent (tar should work)
+            tar_vars = {
+                'am__tar': 'tar chf - "$$tardir"',
+                'am__untar': 'tar xf -',
+                'am_cv_prog_tar_ustar': 'easybuild_avoid_ustar_testing'
+            }
+            for (key, val) in tar_vars.items():
+                self.cfg.update('preconfigopts', "%s='%s'" % (key, val))
```
boegel: you can simplify this significantly by dropping the support for `configure_cmd_prefix` and `tar_config_opts`, which is important for the generic `ConfigureMake` easyblock, but not here

so, something like this is sufficient:

```python
def configure_step(self):
    """Custom configuration procedure for libQGLViewer: ./qmake --prefix=/install/path ..."""
    cmd = "%(preconfigopts)s ./qmake PREFIX=%(installdir)s %(configopts)s" % {
    }
```

easybuild/easyblocks/l/libqglviewer.py
```diff
@@ -0,0 +1,100 @@
...
+            'cmd_prefix': cmd_prefix,
+            'installdir': self.installdir,
+            'configopts': self.cfg['configopts'],
+        }
+
+        (out, _) = run_cmd(cmd, log_all=True, simple=False)
+
+        return out
+
+    def build_step(self, verbose=False, path=None):
```
boegel: if you haven't made any modifications compared to the ConfigureMake implementation (I don't think you have), just drop this whole block, it will be inherited from `ConfigureMake` (just like you're doing already for `install_step` btw...)

easybuild/easyblocks/l/libqglviewer.py
```diff
@@ -0,0 +1,100 @@
...
+        (out, _) = run_cmd(cmd, path=path, log_all=True, simple=False, log_output=verbose)
+
+        return out
+
+    def sanity_check_step(self):
+        """Custom sanity check for libQGLViewer."""
+
+        custom_paths = {
+                        'files': ['lib/libQGLViewer.prl', 'lib/libQGLViewer.so', 'lib/libQGLViewer.so.2', 'lib/libQGLViewer.so.2.5', 'lib/libQGLViewer.so.2.5.1'],
+                        'dirs': ['include/QGLViewer'],
```
boegel: please fix indentation (use 4 spaces per indent level), and be careful with checking for `.so.2` and `.so.2.5`, these are typically very specific to the version being installed; since the `.so` is a symlink to these, it's sufficient to just check the `.so`

also, bonus point for taking both `lib` and `lib64` into account (on some operating systems, libraries get installed to `lib64`, and to not hardcode `.so` but use `get_shared_lib_ext`, as such:

```python
custom_paths = {
    'files': [('lib/libQGLViewer.prl', 'lib64/libQGLViewer.prl'),
              ('lib/libQGLViewer.%s' % shlib_ext, 'lib64/libQGLViewer.%s' % shlib_ext)],
    'dirs': ['include/QGLViewer'],
}
```

note that you'll need an extra import statement on top for `get_shared_lib_ext` (keep them alphabetically sorted):

```python
```

boegel:
Jenkins: ok to test

jrbosch:
@boegel if it is not necessary to include files with the specific version in sanity_check_step method doing it this new way please let me know again and I will remove it.

easybuild/easyblocks/l/libqglviewer.py
jrbosch: @boegel yes, I overworte the method to do some tests to make but I forgot to remove it before creating the PR.

easybuild/easyblocks/l/libqglviewer.py
```diff
@@ -0,0 +1,66 @@
...
+        ver_min = self.version.split('.')[1]
+        ver_maj_min = ".".join([ver_maj, ver_min])
+        shlib_ext = get_shared_lib_ext()
+        
+        custom_paths = {            
+            'files': [('lib/libQGLViewer.prl', 'lib64/libQGLViewer.prl'), 
+		      ('lib/libQGLViewer.%s' % shlib_ext, 'lib64/libQGLViewer.%s' % shlib_ext),
+		      ('lib/libQGLViewer.%s.%s' % (shlib_ext, ver_maj), 'lib64/libQGLViewer.%s.%s' % (shlib_ext, ver_maj)),
+		      ('lib/libQGLViewer.%s.%s' % (shlib_ext, ver_maj_min), 'lib64/libQGLViewer.%s.%s' % (shlib_ext, ver_maj_min)), 
+		      ('lib/libQGLViewer.%s.%s' % (shlib_ext, self.version), 'lib64/libQGLViewer.%s.%s' % (shlib_ext, self.version))],
```
boegel: @jrbosch I think this is overkill, I would just check the `.so` itself...

easybuild/easyblocks/l/libqglviewer.py
```diff
@@ -0,0 +1,66 @@
...
+    def sanity_check_step(self):
+        """Custom sanity check for libQGLViewer."""
+        ver_maj = self.version.split('.')[0]
+        ver_min = self.version.split('.')[1]
+        ver_maj_min = ".".join([ver_maj, ver_min])
+        shlib_ext = get_shared_lib_ext()
+        
+        custom_paths = {            
+            'files': [('lib/libQGLViewer.prl', 'lib64/libQGLViewer.prl'), 
+		      ('lib/libQGLViewer.%s' % shlib_ext, 'lib64/libQGLViewer.%s' % shlib_ext),
```
boegel: to nitpick: one more space indentation (i.e. line up the `(` vertically)

easybuild/easyblocks/l/libqglviewer.py
```diff
@@ -0,0 +1,66 @@
...
+            'installdir': self.installdir,
+            'configopts': self.cfg['configopts'],
+        }
+        run_cmd(cmd, log_all=True, simple=True)
+
+    def sanity_check_step(self):
+        """Custom sanity check for libQGLViewer."""
+        ver_maj = self.version.split('.')[0]
+        ver_min = self.version.split('.')[1]
+        ver_maj_min = ".".join([ver_maj, ver_min])
```
boegel: if you drop the more specific checks below, please drop these 3 lines too

easybuild/easyblocks/l/libqglviewer.py
```diff
@@ -0,0 +1,66 @@
...
+from easybuild.easyblocks.generic.configuremake import ConfigureMake
+from easybuild.tools.run import run_cmd
+from easybuild.tools.systemtools import get_shared_lib_ext
+
+class EB_libQGLViewer(ConfigureMake):
+    """Support for building/installing libQGLViewer."""
+
+    def __init__(self, *args, **kwargs):
+        """Initialisation of custom class variables for libQGLViewer."""
+        super(EB_libQGLViewer, self).__init__(*args, **kwargs)
```
boegel: @jrbosch this is not needed at all, you only need to define `__init__` if you're doing something non-standard, but here you're just calling out to the parent constructor...

boegel:
@jrbosch this looks great now! Thanks!

boegel:
tested with https://github.com/hpcugent/easybuild-easyconfigs/pull/2923, good to go in, thanks @jrbosch!

PR #869, opened 2016-03-12
add sanity check in easybuild/__init__.py w.r.t. current working dir

boegel:
companion PR for https://github.com/hpcugent/easybuild-framework/pull/1667

PR #891, opened 2016-04-20
remove check whether 'easybuild' is being imported from dir that contains easybuild/__init__.py

boegel:
This reverts #869, which was needed because of https://github.com/hpcugent/easybuild-framework/pull/1593, but is no longer relevant after the use of fixup_namespace_packages proved not be required (cfr. https://github.com/hpcugent/easybuild-framework/pull/1680).


boegel:
good to go, see also https://github.com/hpcugent/easybuild-framework/pull/1730

PR #894, opened 2016-04-21
PDT easyblock fixes

geimer:
- Actually pass compiler flag to configure
- Allow building with dummy toolchain

### DIFF
--- a/easybuild/easyblocks/p/pdt.py
+++ b/easybuild/easyblocks/p/pdt.py
@@ -53,14 +53,16 @@ class EB_PDT(ConfigureMake):
 
         # determine values for compiler flags to use
         known_compilers = {
+            toolchain.DUMMY: '-GNU',  # Assume that dummy toolchain uses a system-provided GCC
             toolchain.GCC: '-GNU',
             toolchain.INTELCOMP: '-icpc',
         }
         comp_fam = self.toolchain.comp_family()
         if comp_fam in known_compilers:
-            self.compiler_opt = known_compilers[comp_fam]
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/p/pdt.py
```diff
@@ -53,14 +53,16 @@ def configure_step(self):
...
             toolchain.GCC: '-GNU',
             toolchain.INTELCOMP: '-icpc',
         }
         comp_fam = self.toolchain.comp_family()
         if comp_fam in known_compilers:
-            self.compiler_opt = known_compilers[comp_fam]
+            compiler_opt = known_compilers[comp_fam]
         else:
             raise EasyBuildError("Compiler family not supported yet: %s" % comp_fam)
+        self.cfg.update('configopts', compiler_opt)
```
boegel: so, `compiler_opt` was not used at all before...?

boegel:
Jenkins: ok to test

easybuild/easyblocks/p/pdt.py
geimer: Apparently, no.  So, one could argue that this flag doesn't matter as the easyblock has worked fine until now, but I can't say for sure which consequences omitting this flag really has. I guess that `configure` always picked up GCC, regardless of the actual toolchain...

boegel: Yeah, I guess so... Nice catch!

boegel:
Tested with existing PDT easyconfigs and the ones from #2928, works great, so going in.

Thanks @geimer!

PR #895, opened 2016-04-22
add test configuration for Travis

boegel:
depends on ~~https://github.com/hpcugent/easybuild-framework/pull/1733~~

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ addons:
 install:
     # install easybuild-framework (and dependencies)
     # use framework 'develop' branch, except when testing 'master'
-    if [ "x$TRAVIS_BRANCH" = 'xmaster' ]; then FRAMEWORK_BRANCH=master; else FRAMEWORK_BRANCH=develop; fi
+    - if [ "x$TRAVIS_BRANCH" = 'xmaster' ]; then FRAMEWORK_BRANCH=master; else FRAMEWORK_BRANCH=develop; fi
     - easy_install https://github.com/hpcugent/easybuild-framework/archive/${FRAMEWORK_BRANCH}.tar.gz
     # install environment modules tool using 'install_eb_dep.sh' script provided by easybuild-framework
     - source $(which install_eb_dep.sh) modules-3.2.10 $HOME && source $MOD_INIT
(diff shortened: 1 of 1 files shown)


boegel:
@wpoely86 please review?

.travis.yml
```diff
@@ -0,0 +1,17 @@
...
+  apt:
+    packages:
+      - tcl8.5
+install:
+    # install easybuild-framework (and dependencies)
+    # use framework 'develop' branch, except when testing 'master'
+    - if [ "x$TRAVIS_BRANCH" = 'xmaster' ]; then FRAMEWORK_BRANCH=master; else FRAMEWORK_BRANCH=develop; fi
+    - easy_install https://github.com/hpcugent/easybuild-framework/archive/${FRAMEWORK_BRANCH}.tar.gz
+    # install environment modules tool using 'install_eb_dep.sh' script provided by easybuild-framework
+    - source $(which install_eb_dep.sh) modules-3.2.10 $HOME && source $MOD_INIT
```
wpoely86: only test Tmod?

.travis.yml
```diff
@@ -0,0 +1,17 @@
...
+  - 2.7
+addons:
+  apt:
+    packages:
+      - tcl8.5
+install:
+    # install easybuild-framework (and dependencies)
+    # use framework 'develop' branch, except when testing 'master'
+    - if [ "x$TRAVIS_BRANCH" = 'xmaster' ]; then FRAMEWORK_BRANCH=master; else FRAMEWORK_BRANCH=develop; fi
+    - easy_install https://github.com/hpcugent/easybuild-framework/archive/${FRAMEWORK_BRANCH}.tar.gz
```
wpoely86: does it somewhere clearly show the commit of framework that is being used?

.travis.yml
boegel: For the easyblock unit tests, it doesn't matter at all, no module commands are run.

I'm even wondering now whether this is needed at all...
Maybe it needs to be there to make the framework happy.

.travis.yml
boegel: Hmm, good point... I'll check on that.

boegel: OK, seems like it doesn't. I'll figure out the `git` (or `curl`) command we can use to determine the commit ID that corresponds to `develop` at that time.

boegel:
@wpoely86 good to go, I think?

boegel:

PR #897, opened 2016-04-23
fix bugs w.r.t. --module-only, also test with Lmod as modules tool on Travis

boegel:
@wpoely86's remark on this in #895 was justified, the modules tool *is* indeed used during the easyblock tests too...

Even better, this is exposing a bug, when using Lmod + Lua syntax for modules (probably just a missing `.lua` extension):

```
======================================================================
FAIL: Test for using --module-only with easyblock /home/travis/build/boegel/easybuild-easyblocks/easybuild/easyblocks/generic/craytoolchain.py
----------------------------------------------------------------------
  File "<string>", line 1, in innertest
  File "/home/travis/build/boegel/easybuild-easyblocks/test/easyblocks/module.py", line 212, in template_module_only_test
    self.assertTrue(os.path.exists(modfile), "Module file %s was generated" % modfile)
AssertionError: Module file /tmp/eb-VbRtsI/eb-oyTAci/modules/all/foo/1.3.2 was generated
----------------------------------------------------------------------
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,17 @@
 language: python
-python:
-  - 2.6
-  - 2.7
+python: 2.6
+env:
+  matrix:
+    - ENV_MOD_VERSION=3.2.10
+    - LMOD_VERSION=5.6.3 EASYBUILD_MODULES_TOOL=Lmod
+    - LMOD_VERSION=6.3.1 EASYBUILD_MODULES_TOOL=Lmod EASYBUILD_MODULE_SYNTAX=Lua
+matrix:
+  # mark build as finished as soon as job has failed
+  fast_finish: true
+  include:
+    # also test default configuration with Python 2.7
+    - python: 2.7
+      env: ENV_MOD_VERSION=3.2.10
 addons:
   apt:
     packages:
@@ -13,7 +23,9 @@ install:
     - curl -si https://api.github.com/repos/hpcugent/easybuild-framework/git/refs/heads/${BRANCH} | grep sha
     - easy_install https://github.com/hpcugent/easybuild-framework/archive/${BRANCH}.tar.gz
     # install environment modules tool using 'install_eb_dep.sh' script provided by easybuild-framework
-    - source $(which install_eb_dep.sh) modules-3.2.10 $HOME && source $MOD_INIT
+    - if [ ! -z $ENV_MOD_VERSION ]; then source $(which install_eb_dep.sh) modules-${ENV_MOD_VERSION} $HOME; fi
+    - if [ ! -z $LMOD_VERSION ]; then source $(which install_eb_dep.sh) lua-5.1.4.8 $HOME; fi
+    - if [ ! -z $LMOD_VERSION ]; then source $(which install_eb_dep.sh) Lmod-${LMOD_VERSION} $HOME; fi
 script:
     - export PYTHONPATH=$TRAVIS_BUILD_DIR
     - python -O -m test.easyblocks.suite
--- a/easybuild/easyblocks/p/psi.py
+++ b/easybuild/easyblocks/p/psi.py
@@ -184,9 +184,11 @@ class EB_PSI(CMakeMake):
     def make_module_extra(self):
         """Custom variables for PSI module."""
         txt = super(EB_PSI, self).make_module_extra()
-        psi4datadir = glob.glob(os.path.join(self.installdir, 'share', 'psi*'))
-        if len(psi4datadir) != 1:
-            raise EasyBuildError("Could not determine the PSI4 data dir, there are multiple possibilities: ",
-                                 psi4datadir)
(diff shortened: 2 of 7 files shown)


test/easyblocks/module.py
```diff
@@ -249,8 +249,6 @@ def suite():
     # add dummy PrgEnv-gnu/1.2.3 module, required for testing CrayToolchain easyblock
     write_file(os.path.join(TMPDIR, 'modules', 'all', 'PrgEnv-gnu', '1.2.3'), "#%Module")
 
-    easyblocks = [e for e in easyblocks if 'cray' in e]
```
boegel: oh boy, this should've never slipped in...

.travis.yml
```diff
@@ -13,7 +23,9 @@ install:
     - curl -si https://api.github.com/repos/hpcugent/easybuild-framework/git/refs/heads/${BRANCH} | grep sha
     - easy_install https://github.com/hpcugent/easybuild-framework/archive/${BRANCH}.tar.gz
     # install environment modules tool using 'install_eb_dep.sh' script provided by easybuild-framework
-    - source $(which install_eb_dep.sh) modules-3.2.10 $HOME && source $MOD_INIT
+    - if [ ! -z $ENV_MOD_VERSION ]; then source $(which install_eb_dep.sh) modules-${ENV_MOD_VERSION} $HOME; fi
+    - if [ ! -z $LMOD_VERSION ]; then source $(which install_eb_dep.sh) lua-5.1.4.8 $HOME; fi
```
wpoely86: why the `which`?

boegel: because of the `source`, which needs to location to the script?

.travis.yml
```diff
@@ -13,7 +23,9 @@ install:
     - curl -si https://api.github.com/repos/hpcugent/easybuild-framework/git/refs/heads/${BRANCH} | grep sha
```
wpoely86: can we have an easy click-able link here to the commit in the framework?

boegel: why in particular do you want that?

wpoely86: to be able to quickly view what was used?

boegel: so, you want the log to include something like:

```
https://github.com/hpcugent/easybuild-framework/tree/d9016c2e22a6b0c6eb3763251e8790fb4bdea85e
```

which is basically

```
SHA=`curl -si https://api.github.com/repos/hpcugent/easybuild-framework/git/refs/heads/${BRANCH} | grep sha`
echo "Using easybuild-framework @ https://github.com/hpcugent/easybuild-framework/tree/$SHA"
```

@wpoely86 please open a separate PR for that (also to easyconfigs repo)

boegel: done in #900

.travis.yml
```diff
@@ -1,7 +1,17 @@
...
-  - 2.7
+python: 2.6
+env:
+  matrix:
+    - ENV_MOD_VERSION=3.2.10
+    - LMOD_VERSION=5.6.3 EASYBUILD_MODULES_TOOL=Lmod
+    - LMOD_VERSION=6.3.1 EASYBUILD_MODULES_TOOL=Lmod EASYBUILD_MODULE_SYNTAX=Lua
+matrix:
+  # mark build as finished as soon as job has failed
+  fast_finish: true
```
wpoely86: why? The tests take 60s or so?

easybuild/easyblocks/p/psi.py
```diff
@@ -184,9 +184,11 @@ def sanity_check_step(self):
     def make_module_extra(self):
         """Custom variables for PSI module."""
         txt = super(EB_PSI, self).make_module_extra()
-        psi4datadir = glob.glob(os.path.join(self.installdir, 'share', 'psi*'))
-        if len(psi4datadir) != 1:
-            raise EasyBuildError("Could not determine the PSI4 data dir, there are multiple possibilities: ",
-                                 psi4datadir)
-        txt += self.module_generator.set_environment('PSI4DATADIR', psi4datadir[0])
+        share_dir = os.path.join(self.installdir, 'share')
```
wpoely86: why split this up?

wpoely86:
:+1:

.travis.yml
boegel: sure, but maybe they won't start all at the same time, depending on the resources that are available

quick feedback in case of failing tests is useful, and it doesn't hurt

easybuild/easyblocks/p/psi.py
boegel: the value is used in two places...

boegel:

easybuild/easyblocks/p/psi.py
wpoely86: that's not the question, does the glob not work when the `share` directory doesn't exist?

boegel: the `glob` doesn't work if the directory is not there, e.g. with `--module-only --force` (which gets tested)

PR #898, opened 2016-04-25
Add support for OpenMP in Clang

wpoely86:


wpoely86:
@boegel your way of showing the commit id fails:
```
The command "curl -si https://api.github.com/repos/hpcugent/easybuild-framework/git/refs/heads/${BRANCH} | grep sha" failed and exited with 1 during .
```
Which is logical because the branch only exists in my repo

boegel:
@wpoely86 ah, yes, ok, it works fine when the PR is tested (cfr. https://travis-ci.org/hpcugent/easybuild-easyblocks/jobs/125520148), but not when your branch is tested...

Not only is it your repo, it's also the wrong repo for the branch (framework vs easyblocks).

boegel:
failed test for @wpoely86's branch is at https://travis-ci.org/wpoely86/easybuild-easyblocks/jobs/125519851

@wpoely86: this looks like a fluke to me, since it was doing exactly what is intended, i.e. grabbing the SHA for the `develop` branch of `hpcugent:easybuild-framework` ?

easybuild/easyblocks/c/clang.py
```diff
@@ -150,6 +152,12 @@ def extract_step(self):
 
         src_dirs[clang_src_dir] = os.path.join(self.llvm_src_dir, 'tools', 'clang')
 
+        if LooseVersion(self.version) >= LooseVersion('3.8'):
+            openmp_src_dirs = glob.glob('openmp-*')
+            if len(openmp_src_dirs) != 1:
+                raise EasyBuildError("Failed to find exactly one openmp source directory: %s", openmp_src_dirs)
+            src_dirs[openmp_src_dirs[0]] = os.path.join(self.llvm_src_dir, 'projects', 'openmp')
```
boegel: these four lines are becoming a common pattern, so flesh this out into a function that just takes the glob pattern (`openmp-*`), and returns the dir name (or raises an error)?

wpoely86: it's not really common enough for that.

wpoely86:
Could be, I've seen Travis go weird several times.

easybuild/easyblocks/c/clang.py
boegel: I see 4 occurrences. How is that not enough?

It's not about reducing LoC, I don't really care about that, it's about not copy-pasting code.

easybuild/easyblocks/c/clang.py
```diff
@@ -126,29 +128,28 @@ def extract_step(self):
...
-            raise EasyBuildError("Failed to find exactly one clang source directory: %s", clang_src_dirs)
-        clang_src_dir = clang_src_dirs[0]
+        find_source_dir(['clang-*', 'cfe-*'], os.path.join(self.llvm_src_dir, 'tools', 'clang'))
 
-        src_dirs[clang_src_dir] = os.path.join(self.llvm_src_dir, 'tools', 'clang')
+        if LooseVersion(self.version) >= LooseVersion('3.8'):
+            find_source_dir('openmp-*', os.path.join(self.llvm_src_dir, 'projects', 'openmp'))
 
         for tmp in self.src:
             for (dir, new_path) in src_dirs.items():
```
boegel: @wpoely86: do you mind renaming `tmp` to `src` and `dir` to `dirname` while you're at it?

boegel:
lgtm

boegel:
tested with existing Clang easyconfigs, and the one from https://github.com/hpcugent/easybuild-easyconfigs/pull/2949, good to go

PR #901, opened 2016-04-26
fix extension filter for Ruby, drop deprecated 'testrb' from sanity check

boegel:


boegel:
@rjeschmi please review?

rjeschmi:
looks fine. Was there any problem with the previous extensions check (ie did it miss somethign?)

boegel:
@rjeschmi apparently the argument is a regex, I ran into false positives with short names like `gh` (see also https://github.com/hpcugent/easybuild-easyconfigs/pull/2953)

boegel:
Tested with https://github.com/hpcugent/easybuild-easyconfigs/pull/2953, good to go.

Thanks for the review @rjeschmi!

rjeschmi:
Good catch.

PR #902, opened 2016-04-26
ENH: Support for WRF >= 3.7

andreas-h:
WRF changed its `configure` script in version 3.7.  This PR checks for WRF version and handles both cases (version <= 3.6 and version >= 3.7)

### DIFF
--- a/easybuild/easyblocks/w/wrf.py
+++ b/easybuild/easyblocks/w/wrf.py
@@ -118,12 +118,16 @@ class EB_WRF(EasyBlock):
         build_type_option = None
         self.comp_fam = self.toolchain.comp_family()
         if self.comp_fam == toolchain.INTELCOMP:  #@UndefinedVariable
-            build_type_option = ("INTEL\ \(ifort\/icc\)" if LooseVersion(self.version) >= LooseVersion('3.7')
-                                 else "Linux x86_64 i486 i586 i686, ifort compiler with icc")
+            if LooseVersion(self.version) >= LooseVersion('3.7'):
+                build_type_option = "INTEL\ \(ifort\/icc\)"
+            else:
+                build_type_option = "Linux x86_64 i486 i586 i686, ifort compiler with icc"
 
         elif self.comp_fam == toolchain.GCC:  #@UndefinedVariable
-            build_type_option = ("GNU\ \(gfortran\/gcc\)" if LooseVersion(self.version) >= LooseVersion('3.7')
-                                 else "x86_64 Linux, gfortran compiler with gcc")
+            if LooseVersion(self.version) >= LooseVersion('3.7'):
+                build_type_option = "GNU\ \(gfortran\/gcc\)"
+            else:
+                build_type_option = "x86_64 Linux, gfortran compiler with gcc"
 
         else:
             raise EasyBuildError("Don't know how to figure out build type to select.")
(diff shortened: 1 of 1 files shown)


boegel:
Jenkins: ok to test

easybuild/easyblocks/w/wrf.py
```diff
@@ -116,10 +118,12 @@ def configure_step(self):
         build_type_option = None
         self.comp_fam = self.toolchain.comp_family()
         if self.comp_fam == toolchain.INTELCOMP:  #@UndefinedVariable
-            build_type_option = "Linux x86_64 i486 i586 i686, ifort compiler with icc"
+            build_type_option = ("INTEL\ \(ifort\/icc\)" if LooseVersion(self.version) >= LooseVersion('3.7')
+                                 else "Linux x86_64 i486 i586 i686, ifort compiler with icc")
```
boegel: @andreas-h I would reformat this to this, for readability:

```python
if LooseVersion(self.version) >= LooseVersion('3.7'):
    build_type_option = "INTEL\ \(ifort\/icc\)"
else:
    build_type_option = "Linux x86_64 i486 i586 i686, ifort compiler with icc"
```

same below

easybuild/easyblocks/w/wrf.py
```diff
@@ -133,7 +137,10 @@ def configure_step(self):
             raise EasyBuildError("Unknown build type: '%s'. Supported build types: %s", bt, known_build_types)
 
         # fetch option number based on build type option and selected build type
-        build_type_question = "\s*(?P<nr>[0-9]+).\s*%s\s*\(%s\)" % (build_type_option, bt)
+        if LooseVersion(self.version) >= LooseVersion('3.7'):
+            build_type_question = "\s*(?P<nr>[0-9]+)\.\ \(%s\).*%s" % (bt, build_type_option)
```
boegel: would it make sense to include an example as a comment above? same below

boegel:
@andreas-h this looks very good, thanks a lot for looking into this!

Do you have an updated easyconfig file as well, that can be used as a testcase for this change?

andreas-h:
@boegel I have an easyconfig file which I suppose is correct, but it currently does not build successfully on my system.  But I guess this isn't easybuild related, as I also have trouble building directly (i.e., without EasyBuild).

I uploaded the easyconfig file I'm using here: https://gist.github.com/andreas-h/de2af52d1348a00b308997717986655d but not all dependencies are in the easybuild-easyconfigs repo yet, I guess.

I uploaded the build output showing my problems here: https://gist.github.com/andreas-h/c9ec2df5dea7a0ac18bed2e3643ad2e7 but I guess that's is off-topic for this issue ;)

boegel:
@andreas-h that log doesn't correspond to the `foss` easyconfig you showed?

The problem in the log is that `gcc` is being combined with these `$CFLAGS`, which are for the Intel compilers:

```
== 2016-04-26 17:42:10,258 environment.py:97 INFO Environment variable CFLAGS set to -O2 -xHost -ftz -fp-speculation=safe -fp-model source
```

boegel:
@andreas-h ah, sorry, you have both easyconfigs in your gist... are you hitting similar problems when using `foss`?

boegel:
@andreas-h one extra change is required for recent versions of WRF, cfr. https://github.com/andreas-h/easybuild-easyblocks/pull/1

boegel:
@andreas-h I opened a PR with an easyconfig for WRF 3.8, see https://github.com/hpcugent/easybuild-easyconfigs/issues/2963

Works for me using your updated easyblock, thanks!

boegel:
tested with both existing WRF easyconfigs, and the one in https://github.com/hpcugent/easybuild-easyconfigs/pull/2963, so good to go

Thanks @andreas-h!

PR #884, opened 2016-04-09
make PythonPackage easyblock compatible with --module-only

boegel:
When using `--module-only`, the `configure_step` is skipped, resulting in `self.python_cmd` not being defined.

This leads to an error like:

```
== temporary log file in case of crash /tmp/eb-D0fij6/easybuild-zoU8DE.log
== processing EasyBuild easyconfig /vscmnt/gent_vulpix/_/user/data/gent/gvo000/gvo00002/vsc40023/wxPython-3.0.2.0-intel-2016-Python-2.7.11.eb
== building and installing wxPython/3.0.2.0-intel-2016a-Python-2.7.11...
== fetching files [skipped]
== creating build dir, resetting environment...
== unpacking [skipped]
== patching [skipped]
== preparing...
== configuring [skipped]
== building [skipped]
== testing [skipped]
== installing [skipped]
== taking care of extensions [skipped]
== postprocessing [skipped]
== sanity checking...
ERROR: Traceback (most recent call last):
  File "/user/scratchdelcatty/gent/gvo000/gvo00002/vsc40023/easybuild_easy_installed/lib/python2.6/site-packages/easybuild_framework-2.8.0dev0-py2.6.egg/easybuild/main.py", line 113
, in build_and_install_software                                                                                                                                                      
    (ec_res['success'], app_log, err) = build_and_install_one(ec, init_env)
  File "/user/scratch/gent/vsc400/vsc40023/easybuild_easy_installed/lib/python2.6/site-packages/easybuild_framework-2.8.0dev0-py2.6.egg/easybuild/framework/easyblock.py", line 2292,
 in build_and_install_one                                                                                                                                                            
    result = app.run_all_steps(run_test_cases=run_test_cases)
  File "/user/scratch/gent/vsc400/vsc40023/easybuild_easy_installed/lib/python2.6/site-packages/easybuild_framework-2.8.0dev0-py2.6.egg/easybuild/framework/easyblock.py", line 2208,
 in run_all_steps                                                                                                                                                                    
    self.run_step(step_name, step_methods)
  File "/user/scratch/gent/vsc400/vsc40023/easybuild_easy_installed/lib/python2.6/site-packages/easybuild_framework-2.8.0dev0-py2.6.egg/easybuild/framework/easyblock.py", line 2087,
 in run_step                                                                                                                                                                         
    step_method(self)()
  File "/tmp/eb-D0fij6/included-easyblocks/easybuild/easyblocks/wxpython.py", line 64, in sanity_check_step
[...]

PR #906, opened 2016-05-01
bug fix: use ConfigureMake's test_step in libxml2 easyblock to run 'make check'

boegel:
without this, installing libxml2 with Python support fails with:

```
...
== testing...
== FAILED: Installation ended unsuccessfully (build directory: /tmp/vsc40023/easybuild_build/libxml2/2.9.3/intel-2016a-Python-2.7.11): build failed (first 300 chars): cmd "make True
" exited with exitcode 2 and output:
make: *** No rule to make target `True'.  Stop.
```


boegel:
This bug was introduced in #884, and occurred because `runtest` is `True` for `PythonPackage`, but `ConfigureMake.test_step` doesn't deal with that well...

PR #907, opened 2016-05-04
add easyblock for binutils (REVIEW)

boegel:
see also discussion in https://github.com/hpcugent/easybuild-easyconfigs/pull/2524

cc @wpoely86, @ocaisa

### DIFF
--- a/easybuild/easyblocks/b/binutils.py
+++ b/easybuild/easyblocks/b/binutils.py
@@ -28,9 +28,12 @@ EasyBuild support for building and installing binutils, implemented as an easybl
 @author: Kenneth Hoste (HPC-UGent)
 """
 import os
+import re
 
 from easybuild.easyblocks.generic.configuremake import ConfigureMake
+from easybuild.tools.build_log import EasyBuildError
 from easybuild.tools.modules import get_software_libdir, get_software_root
+from easybuild.tools.run import run_cmd
 from easybuild.tools.systemtools import get_shared_lib_ext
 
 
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/b/binutils.py
```diff
@@ -0,0 +1,94 @@
...
+from easybuild.tools.systemtools import get_shared_lib_ext
+
+
+class EB_binutils(ConfigureMake):
+    """Support for building/installing binutils."""
+
+    def configure_step(self):
+        """Custom configuration procedure for binutils: statically link to zlib, configure options."""
+
+        libdirs = ['/lib', '/lib64', '/usr/lib', '/usr/lib64', '/usr/lib/x86_64-linux-gnu/']
```
wpoely86: check if `/lib` and `/usr/lib` are the same and only add one if needed (same for `lib64`)

wpoely86:
:+1:

rjeschmi:
It looks fine.

probably a good use of RPATH somewhere in here.

The sanity check is probably ok for now, but it would be better to have the sanity check I need for RPATH so we can check in a more robust way.

this is all non-blocking though, just ideas for the future.

boegel:
should be good to go now, see test reports in https://github.com/hpcugent/easybuild-easyconfigs/pull/3006#issuecomment-217641650 (existing binutils easyconfigs) and https://github.com/hpcugent/easybuild-easyconfigs/pull/2524#issuecomment-217652335 (binutils 2.26)

wpoely86:
:+1:

boegel:
Thanks for the review @wpoely86 and @rjeschmi!

PR #911, opened 2016-05-09
CMake: do release build by default

wpoely86:
@boegel please review

### DIFF
--- a/easybuild/easyblocks/generic/cmakemake.py
+++ b/easybuild/easyblocks/generic/cmakemake.py
@@ -60,7 +60,7 @@ class CMakeMake(ConfigureMake):
     def __init__(self, *args, **kwargs):
         """Initialize generic CMake easyblock."""
 
-        super(EB_Clang, self).__init__(*args, **kwargs)
+        super(CMakeMake, self).__init__(*args, **kwargs)
 
         if self.cfg['buildtypes'] not in CMAKE_BUILD_TARGETS:
             raise EasyBuildError("The specified build type for CMake is not known. Accepted values: " \
--- a/easybuild/easyblocks/generic/cmakemake.py
+++ b/easybuild/easyblocks/generic/cmakemake.py
@@ -97,7 +97,7 @@ class CMakeMake(ConfigureMake):
         options = ['-DCMAKE_INSTALL_PREFIX=%s' % self.installdir]
 
         if '-DCMAKE_BUILD_TYPE' not in self.cfg['configopts']:
-            options.append("-DCMAKE_BUILD_TYPE=%s" % self.cfg['buildtypes'])
+            options.append("-DCMAKE_BUILD_TYPE=%s" % self.cfg['buildtype'])
 
         env_to_options = {
             'CC': 'CMAKE_C_COMPILER',
--- a/easybuild/easyblocks/generic/cmakemake.py
+++ b/easybuild/easyblocks/generic/cmakemake.py
@@ -62,7 +62,7 @@ class CMakeMake(ConfigureMake):
 
         super(CMakeMake, self).__init__(*args, **kwargs)
 
-        if self.cfg['buildtypes'] not in CMAKE_BUILD_TARGETS:
+        if self.cfg['buildtype'] not in CMAKE_BUILD_TARGETS:
             raise EasyBuildError("The specified build type for CMake is not known. Accepted values: " \
                                  + ', '.join(CMAKE_BUILD_TARGETS))
 
--- a/easybuild/easyblocks/generic/cmakemake.py
+++ b/easybuild/easyblocks/generic/cmakemake.py
@@ -64,8 +64,8 @@ class CMakeMake(ConfigureMake):
         super(CMakeMake, self).__init__(*args, **kwargs)
 
         if self.cfg['buildtype'] not in CMAKE_BUILD_TYPES:
-            raise EasyBuildError("The specified build type for CMake is not known. Accepted values: " \
-                                 + ', '.join(CMAKE_BUILD_TYPES))
+            raise EasyBuildError("The specified build type for CMake is not known. Accepted values: %s",
+                                 ', '.join(CMAKE_BUILD_TYPES))
 
     def configure_step(self, srcdir=None, builddir=None):
         """Configure build using cmake"""
--- a/easybuild/easyblocks/generic/cmakemake.py
+++ b/easybuild/easyblocks/generic/cmakemake.py
@@ -40,7 +40,8 @@ from easybuild.tools.build_log import EasyBuildError
 from easybuild.tools.environment import setvar
 from easybuild.tools.run import run_cmd
 
-CMAKE_BUILD_TARGETS = ['Debug', 'Release', 'RelWithDebInfo', 'MinSizeRel']
+# Reference: https://cmake.org/cmake/help/v3.0/variable/CMAKE_BUILD_TYPE.html
+CMAKE_BUILD_TYPES = ['Debug', 'Release', 'RelWithDebInfo', 'MinSizeRel']
 
 class CMakeMake(ConfigureMake):
     """Support for configuring build with CMake instead of traditional configure script"""
@@ -52,7 +53,7 @@ class CMakeMake(ConfigureMake):
         extra_vars.update({
             'srcdir': [None, "Source directory location to provide to cmake command", CUSTOM],
             'separate_build_dir': [False, "Perform build in a separate directory", CUSTOM],
-            'buildtype': ['Release', "Build type for CMake. Accepted values: " + ', '.join(CMAKE_BUILD_TARGETS),
+            'buildtype': ['Release', "Build type for CMake. Accepted values: " + ', '.join(CMAKE_BUILD_TYPES),
                           CUSTOM],
         })
         return extra_vars
@@ -62,9 +63,9 @@ class CMakeMake(ConfigureMake):
 
         super(CMakeMake, self).__init__(*args, **kwargs)
 
-        if self.cfg['buildtype'] not in CMAKE_BUILD_TARGETS:
+        if self.cfg['buildtype'] not in CMAKE_BUILD_TYPES:
             raise EasyBuildError("The specified build type for CMake is not known. Accepted values: " \
-                                 + ', '.join(CMAKE_BUILD_TARGETS))
+                                 + ', '.join(CMAKE_BUILD_TYPES))
 
     def configure_step(self, srcdir=None, builddir=None):
         """Configure build using cmake"""
--- a/easybuild/easyblocks/generic/cmakemake.py
+++ b/easybuild/easyblocks/generic/cmakemake.py
@@ -41,7 +41,7 @@ from easybuild.tools.environment import setvar
 from easybuild.tools.run import run_cmd
 
 # Reference: https://cmake.org/cmake/help/v3.0/variable/CMAKE_BUILD_TYPE.html
-CMAKE_BUILD_TYPES = ['Debug', 'Release', 'RelWithDebInfo', 'MinSizeRel']
+CMAKE_BUILD_TYPES = ['Release', 'Debug', 'RelWithDebInfo', 'MinSizeRel']
 
 class CMakeMake(ConfigureMake):
     """Support for configuring build with CMake instead of traditional configure script"""
@@ -53,7 +53,7 @@ class CMakeMake(ConfigureMake):
         extra_vars.update({
             'srcdir': [None, "Source directory location to provide to cmake command", CUSTOM],
             'separate_build_dir': [False, "Perform build in a separate directory", CUSTOM],
-            'buildtype': ['Release', "Build type for CMake. Accepted values: " + ', '.join(CMAKE_BUILD_TYPES),
+            'buildtype': [CMAKE_BUILD_TYPES[0], "Build type for CMake. Accepted values: " + ', '.join(CMAKE_BUILD_TYPES),
                           CUSTOM],
         })
         return extra_vars
--- a/easybuild/easyblocks/generic/cmakemake.py
+++ b/easybuild/easyblocks/generic/cmakemake.py
@@ -51,6 +51,7 @@ class CMakeMake(ConfigureMake):
         extra_vars.update({
             'srcdir': [None, "Source directory location to provide to cmake command", CUSTOM],
             'separate_build_dir': [False, "Perform build in a separate directory", CUSTOM],
+            'releasebuild': [True, "Build with a Release version", CUSTOM],
         })
         return extra_vars
 
@@ -83,6 +84,10 @@ class CMakeMake(ConfigureMake):
                 srcdir = default_srcdir
 
         options = ['-DCMAKE_INSTALL_PREFIX=%s' % self.installdir]
+
+        if self.cfg.get('releasebuild', True):
+            options.append("-DCMAKE_BUILD_TYPE=Release")
+
         env_to_options = {
             'CC': 'CMAKE_C_COMPILER',
             'CFLAGS': 'CMAKE_C_FLAGS',
--- a/easybuild/easyblocks/generic/cmakemake.py
+++ b/easybuild/easyblocks/generic/cmakemake.py
@@ -40,6 +40,7 @@ from easybuild.tools.build_log import EasyBuildError
 from easybuild.tools.environment import setvar
 from easybuild.tools.run import run_cmd
 
+CMAKE_BUILD_TARGETS = ['', 'Debug', 'Release', 'RelWithDebInfo', 'MinSizeRel']
 
 class CMakeMake(ConfigureMake):
     """Support for configuring build with CMake instead of traditional configure script"""
@@ -51,9 +52,19 @@ class CMakeMake(ConfigureMake):
         extra_vars.update({
             'srcdir': [None, "Source directory location to provide to cmake command", CUSTOM],
             'separate_build_dir': [False, "Perform build in a separate directory", CUSTOM],
-            'releasebuild': [True, "Build with a Release version", CUSTOM],
+            'buildtype': ['Release', "Build type for CMake. Accepted values: " + ', '.join(CMAKE_BUILD_TARGETS),
+                          CUSTOM],
         })
         return extra_vars
+
+    def __init__(self, *args, **kwargs):
+        """Initialize generic CMake easyblock."""
+
+        super(EB_Clang, self).__init__(*args, **kwargs)
+
+        if self.cfg['buildtypes'] not in CMAKE_BUILD_TARGETS:
+            raise EasyBuildError("The specified build type for CMake is not known. Accepted values: " \
+                                 + ', '.join(CMAKE_BUILD_TARGETS))
 
     def configure_step(self, srcdir=None, builddir=None):
         """Configure build using cmake"""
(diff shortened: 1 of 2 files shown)


boegel:
@wpoely86 hmm, do you think this is backwards compatible? Is it possible that some applications are unaware of this configure option, and that it'll break builds?

easybuild/easyblocks/generic/cmakemake.py
```diff
@@ -83,6 +84,10 @@ def configure_step(self, srcdir=None, builddir=None):
                 srcdir = default_srcdir
 
         options = ['-DCMAKE_INSTALL_PREFIX=%s' % self.installdir]
+
+        if self.cfg.get('releasebuild', True):
+            options.append("-DCMAKE_BUILD_TYPE=Release")
```
boegel: what if `-DCMAKE_BUILD_TYPE` is already specified?

wpoely86:
It's part of CMake for as long as I know of. It's not application specific.

easybuild/easyblocks/generic/cmakemake.py
wpoely86: The last one will win (which means the one specified by someone else)

boegel: wouldn't it be nicer to check whether `-DCMAKE_BUILD_TYPE` is already specified via `self.cfg['configopts']`?

```python
if self.cfg.get('releasebuild', True) and '-DDCMAKE_BUILD_TYPE' not in self.cfg['configopts']:
    options.append("-DCMAKE_BUILD_TYPE=Release")
```

in that case, you can even drop the `releasebuild` parameter?

also, it would be nicer to support `build_type = 'Release'` instead anyway?

easybuild/easyblocks/generic/cmakemake.py
```diff
@@ -51,9 +52,20 @@ def extra_options(extra_vars=None):
...
+                          CUSTOM],
         })
         return extra_vars
 
+    def __init__(self, *args, **kwargs):
+        """Initialize generic CMake easyblock."""
+
+        super(CMakeMake, self).__init__(*args, **kwargs)
+
+        if self.cfg['buildtype'] not in CMAKE_BUILD_TARGETS:
```
boegel: use `self.get('buildtype', None)`

wpoely86:
@boegel ready

boegel:
@wpoely86 strictly speaking, we should retest *all* easyconfigs/easyblocks that use `CMakeMake`... I suspect that list will be big (see also https://github.com/hpcugent/easybuild-easyblocks/pull/616)

easybuild/easyblocks/generic/cmakemake.py
```diff
@@ -40,6 +40,7 @@
 from easybuild.tools.environment import setvar
 from easybuild.tools.run import run_cmd
 
+CMAKE_BUILD_TARGETS = ['Debug', 'Release', 'RelWithDebInfo', 'MinSizeRel']
```
boegel: add a docs URL for this?

easybuild/easyblocks/generic/cmakemake.py
```diff
@@ -51,9 +52,20 @@ def extra_options(extra_vars=None):
         extra_vars.update({
             'srcdir': [None, "Source directory location to provide to cmake command", CUSTOM],
             'separate_build_dir': [False, "Perform build in a separate directory", CUSTOM],
+            'buildtype': ['Release', "Build type for CMake. Accepted values: " + ', '.join(CMAKE_BUILD_TARGETS),
```
boegel: use a `CMAKE_BUILD_TYPE_RELEASE` constant, and rename to `build_type`?

wpoely86: why?

easybuild/easyblocks/generic/cmakemake.py
```diff
@@ -51,9 +52,20 @@ def extra_options(extra_vars=None):
...
         return extra_vars
 
+    def __init__(self, *args, **kwargs):
+        """Initialize generic CMake easyblock."""
+
+        super(CMakeMake, self).__init__(*args, **kwargs)
+
+        if self.cfg['buildtype'] not in CMAKE_BUILD_TARGETS:
+            raise EasyBuildError("The specified build type for CMake is not known. Accepted values: " \
+                                 + ', '.join(CMAKE_BUILD_TARGETS))
```
boegel: please use this to avoid the `\`:

```python
raise EasyBuildError("The specified build type for CMake is not known. Accepted values: %s",
                     ', '.join(CMAKE_BUILD_TARGETS))
```

easybuild/easyblocks/generic/cmakemake.py
```diff
@@ -83,6 +95,10 @@ def configure_step(self, srcdir=None, builddir=None):
                 srcdir = default_srcdir
 
         options = ['-DCMAKE_INSTALL_PREFIX=%s' % self.installdir]
+
+        if '-DCMAKE_BUILD_TYPE' not in self.cfg['configopts']:
+            options.append("-DCMAKE_BUILD_TYPE=%s" % self.cfg['buildtype'])
```
boegel: add some logging in both cases?

easybuild/easyblocks/generic/cmakemake.py
boegel: magic string, you now have it in two places...

wpoely86: which string?

easybuild/easyblocks/generic/cmakemake.py
wpoely86: you already get the full command run with all options

boegel: sure, but it may not be clear where the `-DCMAKE_BUILD_TYPE` suddenly comes from, or why `build_type` is not being picked up...

easybuild/easyblocks/generic/cmakemake.py
boegel: `'Release'`

boegel: a nice screwup may be to only change it to `RELEASE` in one place

easybuild/easyblocks/generic/cmakemake.py
wpoely86: that goes for all the things that happen here...

wpoely86: which is why we print the full command run

boegel: no, because the other configure options don't have a matching easyconfig parameter that can be set?

the problem is that you can specify the value for `-DCMAKE_BUILD_TYPE` in two ways now, it should be clear from the logs which one wins

wpoely86: we don't do it for any other option (like seperate build dir).

It will be clear which one wins, the command run is shown?

boegel: people may get confused where it comes from, and why the `build_type` they specify is totally ignored...

separate build dir or not is less of an issue, since it doesn't affect the end result (it'll either work, or it won't)

imagine an easyblock deriving from `CMakeMake` updating `configopts` to include `-DCMAKE_BUILD_TYPE`, for example; clear log messages will help in figuring that out

wpoely86: The better solution is to just add it the configopts and not bother with checking it? Having multiple ways is just a source of problems here with zero benefit.

boegel: we should honor backward compatibility, and not blindly add configure options that (may) affect the output of a build, and especially not generate confusing configure commands where two different values are specified for the same option, making it unclear which wins (last wins is a decent assumption, but it's not always that way, so people may wonder, etc.)

easybuild/easyblocks/generic/cmakemake.py
```diff
@@ -51,9 +53,20 @@ def extra_options(extra_vars=None):
         extra_vars.update({
             'srcdir': [None, "Source directory location to provide to cmake command", CUSTOM],
             'separate_build_dir': [False, "Perform build in a separate directory", CUSTOM],
+            'buildtype': [CMAKE_BUILD_TYPES[0], "Build type for CMake. Accepted values: " + ', '.join(CMAKE_BUILD_TYPES),
```
boegel: I don't like this approach, we might miss the impact of reordering the `CMAKE_BUILD_TYPES` alphabetically

I prefer using `CMAKE_BUILD_TYPE_RELEASE`...

easybuild/easyblocks/generic/cmakemake.py
```diff
@@ -51,9 +53,20 @@ def extra_options(extra_vars=None):
...
+                          CUSTOM],
         })
         return extra_vars
 
+    def __init__(self, *args, **kwargs):
+        """Initialize generic CMake easyblock."""
+
+        super(CMakeMake, self).__init__(*args, **kwargs)
+
```
boegel: needs more whitespace? ;)

besserox:
Just one global comment: if you don't specify `CMAKE_BUILD_TYPE` in CMake it will not add any optimization/debug flag on the compiler command line, no `-g`, no `-Ox`, no `-DNDEBUG`. So I think setting "CMAKE_BUILD_TYPE=Release" is a good choice for EasyBuild.

Another point: is the `-march` option still defined for CMake EasyBlock? ie by setting environment variable `CFLAGS`, `CXXFLAGS`, etc.?

easybuild/easyblocks/generic/cmakemake.py
```diff
@@ -51,9 +53,20 @@ def extra_options(extra_vars=None):
...
+        if self.cfg['buildtype'] not in CMAKE_BUILD_TYPES:
+            raise EasyBuildError("The specified build type for CMake is not known. Accepted values: %s",
+                                 ', '.join(CMAKE_BUILD_TYPES))
+
```
besserox: Possibly, an application can define a custom `CMAKE_BUILD_TYPES`. This line will make it impossible to use.

For example, in my case I define a new build type called `Optimized` that corresponds to Release mode + `-march`.

boegel:
@besserox according to @fgeorgatos, you're a CMake expert...

Your thoughts on this, i.e. always specifying `-DCMAKE_BUILD_TYPE=Release` unless `-DCMAKE_BUILD_TYPE` is already specified in `configopts`?

besserox:
Yes, I think you should always specify `-DCMAKE_BUILD_TYPE=Release` unless `-DCMAKE_BUILD_TYPE` is already specified. As I commented before, if nothing specified, there won't be any optimization flag, and that's not what we want with EasyBuild.

Some other comments:

- Ideally, I would forbid `-DCMAKE_BUILD_TYPE` in the `configopts` and recommend instead to set `self.cfg['buildtype'] = "Whatever"`. Of course, as you said, for backward compatibility, you cannot enforce this. But maybe EasyBuild could give some warnings about this?

- Application can possibly bypass the `CMAKE_BUILD_TYPE` of CMake and add additional flags like `-O2` on the compiler command line. This would be a poor design (but I won't be surprise to discover it somewhere). I would suggest to allow the value `None` for the `buildtype` option, and in such case, no `-DCMAKE_BUILD_TYPE` should be generated. This would allow some flexibility.

- Application can also create custom build modes. In such case it should be possible to use it for the `buildtype` option. This is even documented https://cmake.org/Wiki/CMake_FAQ#How_can_I_extend_the_build_modes_with_a_custom_made_one_.3F

boegel:
@besserox EasyBuild takes control over `$CFLAGS` etc. and passes them down to CMake via `-DCMAKE_C_CFLAGS="$CFLAGS"`, etc.

So whether or not the build type is set makes little difference imho, but we should do it for good practice.

I'm just a little bit hesitant to push this in so close to a new release, I expect some builds to be affected (in the negative sense), for whatever reason...

easybuild/easyblocks/generic/cmakemake.py
boegel: OK, then this check doesn't make sense, indeed, we should allow the build type to be specified freely, and simply use `Release` as default.

besserox:
Yes, I understand your problem.

If you want to let EasyBuild manage the flags, then you should probably force `CMAKE_BUILD_TYPE=""`.

FYI, this are the flags added by CMake (in my case, this might depend on CMake version, compiler, etc.)
- Debug -> `-g`
- MinSizeRel -> `-Os -DNDEBUG`
- Release -> `-O3 -DNDEBUG`
- RelWithDebInfo -> `-O2 -g -DNDEBUG`

besserox:
If you specify `-DCMAKE_C_FLAGS` and `-DCMAKE_BUILD_TYPE`, the flags from both will appear on the compiler command line.

boegel:
@besserox https://cmake.org/cmake/help/v3.5/variable/CMAKE_LANG_FLAGS.html?highlight=flags

```
CMAKE_<LANG>_FLAGS¶
Flags for all build types.

<LANG> flags used regardless of the value of CMAKE_BUILD_TYPE.
```

So, hard specified flags will win, regardless of the specified build type. But maybe specifying the build type may have other side effects?

So, it still makes sense to me to use `Release` as default?

boegel:
From https://cmake.org/cmake/help/v3.5/variable/CMAKE_BUILD_TYPE.html#variable:CMAKE_BUILD_TYPE:

```
For example, in a build tree configured to build type Debug, CMake will see to having CMAKE_C_FLAGS_DEBUG settings get added to the CMAKE_C_FLAGS settings.
```

So, it makes sense to still properly define it, there may be critical flags that get appended via `CMAKE_C_FLAGS_RELEASE`?

boegel:
There are 289 easyconfigs that use CMake in current `develop`, that's doable for testing overnight on top of this...

besserox:
For Release mode `-O3 -DNDEBUG` flags are added (cf my comment above)

What happens if you have conflicting flags? ie `-O2` and `-O3`

boegel:
@besserox Last flag wins, so changing the default to `Release` (or anything, frankly), would have the unwanted side-effect of using `-O3` for everything. Eek.

@wpoely86 So, use `None` as default then?

besserox:
Yes, I think you should go for `None` as default

boegel:
works for me, since that retains backwards compatibility of not fiddling with `-DCMAKE_BUILD_TYPE` by default...

boegel:
but then, the value of this PR is rather low too, since you could just add `-DCMAKE_BUILD_TYPE` to configopts instead?

thoughts @wpoely86?

besserox:
@boegel Have you considered adding `-DNDEBUG` in the default flags in EasyBuild?
According to `assert` manpage, it disables assertions. It might also be used by developers to disable other debug code.

boegel:
@besserox that probably makes sense... @wpoely86?

wpoely86:
I don't know, for Clang for example, we have an option for that. We can create a 'Release' build of clang but still with assertions enabled.

Flamefire:
Current solution LGTM (except the erroneous bamtools inclusion). It is minimal and does "The Right Thing" by default

boegel:
> Current solution LGTM (except the erroneous bamtools inclusion). It is minimal and does "The Right Thing" by default

@Flamefire It does make sense, but the question remains whether a change like this is going to have fallout. Perhaps there are creative ways where using `-DCMAKE_BUILD_TYPE=Release` would make the installation fail (for example because the developers figured it would be a good idea to come up with their own list of known build types)...

Flamefire:
The default build type (empty string) is a valid one but means only a baseline. All optimizations are usually not turned on, so the compiler will be running without any `-O` options. This is probably OK because EB defaults CFLAGS to include `-O2`. However CMake also adds e.g. `-DNDEBUG` to release builds to disable assertions, and probably there are also other optimizations. As EB is for HPC having not-fully-optimized libs installed can cost real money (in compute time). Of course impact depends on the actual code, but still...

While I understand your concern I don't think it will be a problem in practice. Most (if not all) CMake projects are built on CI with `-DCMAKE_BUILD_TYPE=Release` and/or `-DCMAKE_BUILD_TYPE=Debug`. I'm not sure of no build type is actually tested. Some CMake projects have code to change the empty build type to Release to mitigate this issue. Coming up with their own build types is of course possible, but usually they append to the common ones (e.g. for a profiling or sanitized build) And for the very unlikely case that there is no Release type you got the  `buildtype` option to change it per EC.

So bottom line: IMO this should have been merged already ;-)

if @wpoely86 is no longer active, you could cherry-pick the changes and make the final adjustments or I could open a PR and mention @wpoely86 in the commit message

akesandgren:
"Most (if not all) CMake projects are built on CI with -DCMAKE_BUILD_TYPE=Release and/or -DCMAKE_BUILD_TYPE=Debug"

This is, sadly, not true. Way to many projects using CMake doesn't have a clue about CMAKE_BUILD_TYPE, nor about proper optimization flags.

And there is one thing that always trumps performance... correctness of results.
And WAY to many codes are poorly written and breaks when optimization is turned up.

I.e., using Release by default is likely to break a lot of codes that we have in the tree.
So before we do that I think we should setup proper tests suites, with actual runs of the codes with real user inputs, to compare.
This is sadly very time consuming to setup...

Just to make things clear. I'm all for performance, as long as the results are correct. I've just seen too much code that breaks if you just look at it...

Flamefire:
> This is, sadly, not true. Way to many projects using CMake doesn't have a clue about CMAKE_BUILD_TYPE, nor about proper optimization flags.

I guess a check is due then. An easy way I'd imagine: Could EB be used to download and extract all sources and break further processing just before calling `cmake ...`? This will put all sources into a subfolder of a /tmp folder and one can grep for CMAKE_BUILD_TYPE and CMAKE_CONFIGURATION_TYPES (or what the name was). All projects using that are probably ok. Same for projects using `CMAKE_<CXX|C>_FLAGS_RELEASE`

At the least: Could we add this option with empty default? Then ECs/EBs can add it where they know

> And WAY to many codes are poorly written and breaks when optimization is turned up.

Can you say which and how/why?

Counter-example: Caffee is very much aware of it and expects it to be either Release or Debug.

akesandgren:
My pet project for broken code was (until I got enough patches accepted upstream) VASP. Now it's just half bad. Not using CMake though, but that wasn't my main point on that part.

I don't have any specifics at the moment, been too busy installing software on user requests and making EB better. But you can probably take any bioinformatics code as an example. Or Amber...
or QuantumESPRESSO, esp the code they have to produce the vdW_kernel table which breaks horribly when going to -O2 if i remember correctly.

And yes, there are lots of really good codes out there, like GROMACS, and for instance the blas/lapack/scalapack code. (And yes I've found more than one bug in them over the years causing them to break when increasing optimization)

And when it comes to correctness, I, personally, would not trust even code that by default uses Release, to deliver correct results...

Anyway, this is a discussion that could go on for ages, there is always new code coming out, both broken by design and good examnples.

Flamefire:
> But you can probably take any bioinformatics code as an example. Or Amber...
or QuantumESPRESSO, esp the code they have to produce the vdW_kernel table which breaks horribly when going to -O2 if i remember correctly.

Well, EB adds `-O2` by default. So if there are broken packages included, they'll be broken already.

For current usage in EB:
- There are 212 software packages using `CMakeMake`
- 36 of those use `-DCMAKE_BUILD_TYPE`
- from a quick scan about half of them are "good CMake" and (probably) support build types (just judging names I know)
- no (to me) known bad examples, like the mentioned Amber, QuantumESpresso

IMO Release should be the default just as `CFLAGS=-O2` is. There is always the option to opt-out. If one can get EB to download and extract the relevant packages a grep can help filtering out "good" candidates. Depending on how many are left those can be checked or have (probably automatically) the opt-out added.

branfosj:
#1929 (CMake release build by default) and #1656 (Bamtools), so closing.

Issue #795, opened 2016-01-04
Problem with logic for use of SCOTCH_PTHREAD


ocaisa:
Currenty in [this line](https://github.com/hpcugent/easybuild-easyblocks/blob/develop/easybuild/easyblocks/s/scotch.py#L111) we switch off the use of `SCOTCH_PTHREAD` for some MPI implementattions. ParaStation MPI has both threadsafe and non-threadsafe versions so we need the ability to turn this feature on or off from the easyconfig file too.

boegel:
up for tackling this yourself @ocaisa? or @damianam?

ocaisa:
We've done a modified EasyBlock that accepts `threadedmpi = True` in the easyconfig. Will contribute back when we get a chance to take a breath

boegel:
@ocaisa great!

fgeorgatos:
@ocaisa, @boegel: thanks, this ringed a bell! #166 could have been helped by it.
When I have a crumb of time, I'd give it another spin just out of curiosity.

boegel:
@ocaisa breathing yet? ;)

boegel:
follow-up in PR #914 that tackles this, thanks @ocaisa

PR #914, opened 2016-05-10
Make threading an explicit option rather than relying on the MPI impl…

ocaisa:
…ementation (ParaStaion MPI can support both possibilities)

### DIFF
--- a/easybuild/easyblocks/s/scotch.py
+++ b/easybuild/easyblocks/s/scotch.py
@@ -40,6 +40,7 @@ from distutils.version import LooseVersion
 
 import easybuild.tools.toolchain as toolchain
 from easybuild.framework.easyblock import EasyBlock
+from easybuild.framework.easyconfig import CUSTOM
 from easybuild.tools.build_log import EasyBuildError
 from easybuild.tools.filetools import copytree
 from easybuild.tools.run import run_cmd
@@ -47,6 +48,14 @@ from easybuild.tools.run import run_cmd
 
 class EB_SCOTCH(EasyBlock):
     """Support for building/installing SCOTCH."""
+
+    @staticmethod
+    def extra_options(extra_vars=None):
+        """Define custom easyconfig parameters specific to Scotch."""
+        extra_vars = {
+            'threadedmpi': [False, "Use threaded MPI calls.", CUSTOM],
+        }
+        return EasyBlock.extra_options(extra_vars)
 
     def configure_step(self):
         """Configure SCOTCH build: locate the template makefile, copy it to a general Makefile.inc and patch it."""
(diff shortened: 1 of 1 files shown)


ocaisa:
Resolves #795

easybuild/easyblocks/s/scotch.py
```diff
@@ -108,7 +117,7 @@ def build_step(self):
         if self.toolchain.options['i8']:
             cflags += " -DINTSIZE64"
 
-        if not self.toolchain.mpi_family() in [toolchain.INTELMPI, toolchain.QLOGICMPI]:  #@UndefinedVariable
+        if self.cfg['threadedmpi']:
```
boegel: @ocaisa this implies that for builds on top of OpenMPI (for example), `threadedmpi = True` must be now specified to obtain the same installation, so this should be something like:

```python
if self.cfg['threadedmpi'] or if self.toolchain.mpi_family() not in [toolchain.INTELMPI, toolchain.QLOGICMPI]
```

to ensure backward compatibility?

ocaisa: Ok, but I'm pretty sure this wasn't working for people, to my eyes it's mostly there for OpenFOAM and threaded scotch does not work with OpenFOAM.
I'll set the default to true for OpenMPI but personally I think this is probably not doing what people actually expect.

ocaisa: How do I check a variable has been set in the easyconfig? I need to know if they have explicitly set the value to false in the easyconfig so I can ensure I only change the value to true when it's not defined there.

boegel: @ocaisa you could use `None` as default value, and assume people will set it to `False` in the easyconfig?

and then check `self.cfg['threadedmpi'] is None`

there's no easy way to distinguish between a default value and the same value but specified by the easyconfig

ocaisa:
@boegel Tested with ParaStation MPI and Intel MPI, backwards compatibility is now there and the new logic works

easybuild/easyblocks/s/scotch.py
```diff
@@ -108,9 +117,12 @@ def build_step(self):
         if self.toolchain.options['i8']:
             cflags += " -DINTSIZE64"
 
-        if not self.toolchain.mpi_family() in [toolchain.INTELMPI, toolchain.QLOGICMPI]:  #@UndefinedVariable
+        if self.cfg['threadedmpi']: 
             cflags += " -DSCOTCH_PTHREAD"
-
+        # TODO For backwards compatability of v2.8.0 the following is necessary but could be removed on a major version upgrade
+        if self.cfg['threadedmpi'] is None and not self.toolchain.mpi_family() in [toolchain.INTELMPI, toolchain.QLOGICMPI]:
```
boegel: please change to `not in`:

```python
and self.toolchain.mpi_family() not in [toolchain.INTELMPI, toolchain.QLOGICMPI]:
```

boegel:
going in, thanks @ocaisa!

PR #916, opened 2016-05-12
also patch out -Werror when building imkl FFTW interfaces with PGI

boegel:
Ran into this with imkl 11.3.3 (https://github.com/hpcugent/easybuild-easyconfigs/pull/3044).


ocaisa:
lgtm

boegel:

Issue #918, opened 2016-05-12
Investigate creation of a `siterc` file for PGI to support `$LIBRARY_PATH`


geimer:
Quoting from https://github.com/hpcugent/easybuild/issues/217:

Apparently `$LIBRARY_PATH` is ignored by the PGI compilers. However, the GCCcore stuff is correctly picked up via the `localrc` file generated by the `makelocalrc` script run during installation. To make `$LIBRARY_PATH` work, PGI proposes a workaround on their web pages, see https://www.pgroup.com/support/link.htm#lib_path_ldflags.

Note that there is a typo in the code snippet. The second line should read `$LIBRARY_PATH` from the environment, not `$LD_LIBRARY_PATH`.

It should also be straightforward to add `/usr/lib/x86_64-linux-gnu` to `$LIBRARY_PATH` in the module file directly from the PGI easyblock when on a Debian/Ubuntu system, to work around the issue linked above.

boegel:
@damianam, @bartoldeman thoughts on this?

PGI not picking up `$LIBRARY_PATH` is going to bite us in the ass sooner or later, we're bound to rely on it in several places since `$LIBRARY_PATH` is defined by the modules generated by EB...

damianam:
That's a PITA. I see 2 approaches:

* We patch the installation as suggested in Marku's link. The change is permanent for everyone.
* We redefine `LIBRARY_PATHS` to `LDLIBARGS` in EB (I am not sure where though).

The first approach changes the default behaviour, but will make things easier for new PGI users. The second approach respect the default behaviour, but it will be more work to implement and will left new users exposed to this annoyance. I'd suggest to go for the first one.

boegel:
Since installing a custom `siterc` file is an official recommendation by PGI, I'd definitely go with that...

damianam:
Seems more like a workaround than a recommendation, but yes, I agree, it looks like the best option.

bartoldeman:
Agreed, siterc is the way to go

boegel:
fixed in #919, please review?

boegel:
let's follow up in #919, there's no use in tracking this in two issues

PR #919, opened 2016-05-13
update PGI easyblock to install siterc file so PGI picks up $LIBRARY_PATH + enhance sanity check paths for compiler commands

boegel:
fix for #918

cc @geimer, @damianam, @bartoldeman: please review?

### DIFF
--- a/easybuild/easyblocks/p/pgi.py
+++ b/easybuild/easyblocks/p/pgi.py
@@ -57,7 +57,11 @@ variable library_path is
 default($if($LIBRARY_PATH,-L$replace($LIBRARY_PATH,":", -L)));
 
 # add the -L arguments to the link line
-append LDLIBARGS=$library_path; 
+append LDLIBARGS=$library_path;
+
+# also include the location where libm & co live on Debian-based systems
+# cfr. https://github.com/hpcugent/easybuild-easyblocks/pull/919
+append LDLIBARGS=-L/usr/lib/x86_64-linux-gnu;
 """
 
 
--- a/easybuild/easyblocks/p/pgi.py
+++ b/easybuild/easyblocks/p/pgi.py
@@ -41,9 +41,25 @@ import easybuild.tools.environment as env
 from easybuild.framework.easyblock import EasyBlock
 from easybuild.framework.easyconfig import CUSTOM
 from easybuild.tools.build_log import EasyBuildError
-from easybuild.tools.filetools import find_flexlm_license
+from easybuild.tools.filetools import find_flexlm_license, write_file
 from easybuild.tools.run import run_cmd
 from easybuild.tools.modules import get_software_root
+
+
+# contents for siterc file to make PGI pick up $LIBRARY_PATH
+# cfr. https://www.pgroup.com/support/link.htm#lib_path_ldflags
+SITERC_LIBRARY_PATH = """
+# get the value of the environment variable LIBRARY_PATH
+variable LIBRARY_PATH is environment(LIBRARY_PATH);
+
+# split this value at colons, separate by -L, prepend 1st one by -L
+variable library_path is
+default($if($LIBRARY_PATH,-L$replace($LIBRARY_PATH,":", -L)));
+
+# add the -L arguments to the link line
+append LDLIBARGS=$library_path; 
+"""
+
 
 class EB_PGI(EasyBlock):
     """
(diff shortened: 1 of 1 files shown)


damianam:
It looks good to me but I think we should check if `siterc` is really honoured (just in case the documentation is not accurate, like with `PGI_JAVA`). We don't have siterc in our system, but we have a bunch of other rc files. I wonder if `siterc` got deprecated in favour of `localrc` or something like that.

boegel:
@damianam works as advertised:

```
$ file $EBROOTPGI/linux86-64/16.3/bin/siterc
/user/scratch/gent/vsc400/vsc40023/easybuild_REGTEST/SL6/sandybridge/software/PGI/16.3-GCC-4.9.3-2.25/linux86-64/16.3/bin/siterc

$ pgc++ -v hello.cpp 2>&1 | tr ' ' '\n' | grep '\-L'
-L/user/scratch/gent/vsc400/vsc40023/easybuild_REGTEST/SL6/sandybridge/software/PGI/16.3-GCC-4.9.3-2.25/linux86-64/16.3/lib
-L/user/scratch/gent/vsc400/vsc40023/easybuild_REGTEST/SL6/sandybridge/software/numactl/2.0.11-GCCcore-4.9.3/lib
-L/user/scratch/gent/vsc400/vsc40023/easybuild_REGTEST/SL6/sandybridge/software/binutils/2.25-GCCcore-4.9.3/lib
-L/apps/gent/SL6/sandybridge/software/GCCcore/4.9.3/lib64
-L/apps/gent/SL6/sandybridge/software/GCCcore/4.9.3/lib
-L/user/scratchdelcatty/gent/gvo000/gvo00002/vsc40023/easybuild_REGTEST/SL6/sandybridge/software/PGI/16.3-GCC-4.9.3-2.25/linux86-64/16.3/lib
-L/usr/lib64
-L/vscmnt/gent_vulpix_apps/_apps_gent/SL6/sandybridge/software/GCCcore/4.9.3/bin/../lib/gcc/x86_64-unknown-linux-gnu/4.9.3
-L/vscmnt/gent_vulpix_apps/_apps_gent/SL6/sandybridge/software/GCCcore/4.9.3/bin/../lib/gcc/x86_64-unknown-linux-gnu/4.9.3/../../../../lib64
```

compare this without having the `siterc`:

```
$ ls $EBROOTPGI/linux86-64/16.3/bin/siterc
ls: cannot access /user/scratch/gent/vsc400/vsc40023/easybuild_REGTEST/CO7/haswell-ib/software/PGI/16.3-GCC-4.9.3-2.25/linux86-64/16.3/bin/siterc: No such file or directory

$ pgc++ -v hello.cpp 2>&1 | tr ' ' '\n' | grep '\-L'
-L/user/scratchdelcatty/gent/gvo000/gvo00002/vsc40023/easybuild_REGTEST/CO7/haswell-ib/software/PGI/16.3-GCC-4.9.3-2.25/linux86-64/16.3/lib
-L/usr/lib64
-L/user/scratchdelcatty/gent/gvo000/gvo00002/vsc40023/easybuild_REGTEST/CO7/haswell-ib/software/GCCcore/4.9.3/bin/../lib/gcc/x86_64-unknown-linux-gnu/4.9.3
-L/user/scratchdelcatty/gent/gvo000/gvo00002/vsc40023/easybuild_REGTEST/CO7/haswell-ib/software/GCCcore/4.9.3/bin/../lib/gcc/x86_64-unknown-linux-gnu/4.9.3/../../../../lib64
```

rjeschmi:
This is probably a good idea.

I think a lot of software that uses pgi works around it.

geimer:
@boegel: lgtm, but...

To make it work on my Debian system, I need to add `/usr/lib/x86_64-linux-gnu` to `$LIBRARY_PATH`.  However, I can't simply add a `modextrapaths` line to the easyconfig file for this, as `modextrapaths` doesn't allow absolute paths... Is there any workaround or do we have to handle this in the easyblock as well?

boegel:
@geimer ah right, I forgot about that part...

There's some discussion on making `modextrapaths` accept absolute paths too, but I still think it would be better to handle this for PGI specifically in the `siterc` file being added by the easyblock.

Can you verify whether adding this in the `siterc` as well fixes the problem?

```
# also include the location where libm & co live on Debian-based systems
append LDLIBARGS=/usr/lib/x86_64-linux-gnu;
```

Do we need to worry about `lib64` too here?

boegel:
@geimer actually, I think it should be this (i.e. include the `-L`)

```
# also include the location where libm & co live on Debian-based systems
append LDLIBARGS=-L/usr/lib/x86_64-linux-gnu
```

geimer:
@boegel: Yep, noticed that already ;-)  With the `-L` it seems to work well.

There is no `/usr/lib64` on a 64-bit Debian multiarch system.  64-bit libraries are in `/usr/lib/x86_64-linux-gnu`, 32-bit libs in `/usr/lib/i386-linux-gnu`.  However, I don't think we have to care much about the 32-bit stuff as all available easyconfigs only install a 64-bit compiler which cannot produce 32-bit code:
```
$ pgcc -m32 -o hello hello.c 
pgcc-Fatal--m32 is not supported in this installation
```

boegel:
@geimer: OK, added, thanks!

Should be good to go now?

geimer:
@boegel: Reinstalled PGI with this PR applied and it works for me, so in that sense it's good to go.  But do you want to include this directory unconditionally?  Should be OK, but maybe someone with a non-Debian system should verify that it doesn't break anything?

boegel:
@geimer: I checked that with `pgcc -v hello.c`; works fine, path is not mentioned in that output, so seems like PGI checks for existence itself.

Going in, thanks for the feedback @damianam, @rjeschmi and @geimer!

PR #923, opened 2016-05-17
add easyblock for siesta (WIP)

migueldiascosta:
corresponding easyconfig in hpcugent/easybuild-easyconfigs#2965

### DIFF
--- a/easybuild/easyblocks/s/siesta.py
+++ b/easybuild/easyblocks/s/siesta.py
@@ -30,6 +30,7 @@ EasyBuild support for Siesta, implemented as an easyblock
 
 import os
 import easybuild.tools.toolchain as toolchain
+from distutils.version import LooseVersion
 from easybuild.framework.easyconfig import CUSTOM
 from easybuild.easyblocks.generic.configuremake import ConfigureMake
 from easybuild.easyblocks.generic.makecp import MakeCp
@@ -37,6 +38,13 @@ from easybuild.easyblocks.generic.makecp import MakeCp
 
 class EB_Siesta(MakeCp):
     """Support for building and installing Siesta."""
+
+    def __init__(self, *args, **kwargs):
+        super(EB_Siesta, self).__init__(*args, **kwargs)
+        if LooseVersion(self.version) < LooseVersion('4.0'):
+            self.vibra_filename = 'vibrator'
+        else:
+            self.vibra_filename = 'vibra'
 
     @staticmethod
     def extra_options(extra_vars=None):
@@ -106,7 +114,7 @@ class EB_Siesta(MakeCp):
             utils.extend(os.path.join('SiestaSubroutine/FmixMD/Src', b) for b in ['driver', 'para', 'simple'])
             utils.append(os.path.join('TBTrans', 'tbtrans'))
             utils.extend(os.path.join('VCA', b) for b in ['fractional', 'mixps'])
-            utils.extend(os.path.join('Vibra/Src', b) for b in ['fcbuild', 'vibrator'])
+            utils.extend(os.path.join('Vibra/Src', b) for b in ['fcbuild', self.vibra_filename])
             utils.extend(os.path.join('WFS', b) for b in ['info_wfsx', 'readwf', 'readwfx', 'wfs2wfsx', 'wfsnc2wfsx',
                                                           'wfsx2wfs'])
             utils.append(os.path.join('pdosxml', 'pdosxml'))
@@ -134,8 +142,8 @@ class EB_Siesta(MakeCp):
                          'grid2cube', 'grid2val', 'hs2hsx', 'hsx2hs', 'info_wfsx', 'ioncat', 'ionplot.sh', 'lwf2cdf',
                          'macroave', 'md2axsf', 'mixps', 'mprop', 'new.gnubands', 'orbmol_proj', 'para', 'pdosxml',
                          'pi3', 'plstm', 'readwf', 'readwfx', 'rho2xsf', 'simple', 'simplex', 'surf.py', 'swarm',
-                         'tbtrans', 'test-xml', 'vib2xsf', 'vibrator', 'wfs2wfsx', 'wfsnc2wfsx', 'wfsx2wfs', 'xml2psf',
-                         'xv2xsf'])
+                         'tbtrans', 'test-xml', 'vib2xsf', self.vibra_filename, 'wfs2wfsx', 'wfsnc2wfsx', 'wfsx2wfs',
+                         'xml2psf', 'xv2xsf'])
 
         custom_paths = {
             'files': bins,
--- a/easybuild/easyblocks/s/siesta.py
+++ b/easybuild/easyblocks/s/siesta.py
@@ -29,10 +29,13 @@ EasyBuild support for Siesta, implemented as an easyblock
 """
 
 import os
+import shutil
 import easybuild.tools.toolchain as toolchain
+from easybuild.tools.run import run_cmd
 from distutils.version import LooseVersion
 from easybuild.framework.easyconfig import CUSTOM
 from easybuild.easyblocks.generic.makecp import MakeCp
+from easybuild.tools.filetools import apply_regex_substitutions
 
 
 class EB_Siesta(MakeCp):
@@ -71,19 +74,37 @@ class EB_Siesta(MakeCp):
         cfg_cmd += '--with-blacs="' + os.environ['LIBBLACS'] + '" '
         cfg_cmd += '--with-scalapack="' + os.environ['LIBSCALAPACK'] + '"'
 
-        # make sure packaged lapack is not on generated arch.make
-        sed_cmd = "sed -i 's/COMP_LIBS=dc_lapack.a/COMP_LIBS=/' arch.make"
+        try:
+            os.chdir('Obj')
+        except OSError, err:
+            raise EasyBuildError("Failed to move into build dir: %s", err)
 
-        self.cfg.update('prebuildopts', 'cd Obj && ' + cfg_cmd + ' && ' + sed_cmd + ' && ')
+        run_cmd(cfg_cmd, log_all=True, simple=True, log_output=True)
+
+        # make sure packaged lapack is not on generated arch.make
+        apply_regex_substitutions('arch.make', [('dc_lapack.a', ''), ('libsiestaLAPACK.a', '')])
 
         if self.cfg['with_transiesta']:
-            self.cfg.update('buildopts', ' && cd .. && mkdir Obj2 && cd Obj2 && ')
-            self.cfg.update('buildopts', cfg_cmd + ' && ' + sed_cmd + ' && make transiesta ')
+            try:
+                shutil.copytree('../Obj', '../Obj2')
+            except OSError, err:
+                raise EasyBuildError("Failed to copy build dir: %s", err)
+
+        run_cmd("make", log_all=True, simple=True, log_output=True)
+
+        if self.cfg['with_transiesta']:
+            try:
+                os.chdir('../Obj2')
+            except OSError, err:
+                raise EasyBuildError("Failed to move to transiesta build dir: %s", err)
+            run_cmd("make transiesta", log_all=True, simple=True, log_output=True)
 
         if self.cfg['with_utils']:
-            self.cfg.update('buildopts', ' && cd ../Util && sh ./build_all.sh')
-
-        super(EB_Siesta, self).build_step()
+            try:
+                os.chdir('../Util')
+            except OSError, err:
+                raise EasyBuildError("Failed to move to Util dir: %s", err)
+            run_cmd("./build_all.sh", log_all=True, simple=True, log_output=True)
 
     def install_step(self):
         """Custom install procedure for Siesta."""
--- a/easybuild/easyblocks/s/siesta.py
+++ b/easybuild/easyblocks/s/siesta.py
@@ -71,11 +71,14 @@ class EB_Siesta(MakeCp):
         cfg_cmd += '--with-blacs="' + os.environ['LIBBLACS'] + '" '
         cfg_cmd += '--with-scalapack="' + os.environ['LIBSCALAPACK'] + '"'
 
-        self.cfg.update('prebuildopts', 'cd Obj && ' + cfg_cmd + ' && ')
+        # make sure packaged lapack is not on generated arch.make
+        sed_cmd = "sed -i 's/COMP_LIBS=dc_lapack.a/COMP_LIBS=/' arch.make"
+
+        self.cfg.update('prebuildopts', 'cd Obj && ' + cfg_cmd + ' && ' + sed_cmd + ' && ')
 
         if self.cfg['with_transiesta']:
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/s/siesta.py
```diff
@@ -0,0 +1,162 @@
...
+        """
+        extra = {
+            'files_to_copy': [[], "List of files or dirs to copy", CUSTOM],
+            'with_transiesta': [True, "Build transiesta", CUSTOM],
+            'with_utils': [True, "Build all utils", CUSTOM],
+        }
+        if extra_vars is None:
+            extra_vars = {}
+        extra.update(extra_vars)
+        return ConfigureMake.extra_options(extra_vars=extra)
```
boegel: since you're deriving directly from `MakeCp`, you should use `MakeCp.extra_options`?

easybuild/easyblocks/s/siesta.py
```diff
@@ -0,0 +1,162 @@
...
+
+    def __init__(self, *args, **kwargs):
```
boegel: override `self.build_step` and `self.install_step` instead to update `self.cfg['buildopts']` and `self.cfg['installopts']`?

also, use `self.cfg.update`:

```python
def build_step(self):
    """Custom build procedure for Siesta."""
    self.cfg.update('prebuildopts', "cd Obj && " + cfg_cmd + " && ")
    ...
    super(MakeCp, self).build_step()

File: easybuild/easyblocks/s/siesta.py [hunk 3]
```diff
@@ -0,0 +1,162 @@
...
+        super(EB_Siesta, self).__init__(*args, **kwargs)
+
+        cfg_cmd = '../Src/obj_setup.sh && ../Src/configure'
+
+        if self.toolchain.options.get('usempi', None):
+            cfg_cmd += ' --enable-mpi '
+
+        build_vars = "COMP_LIBS='' "
+        build_vars += 'BLAS_LIBS="$LIBBLAS" LAPACK_LIBS="$LIBLAPACK" '
```
boegel: grab the actual value for `$LIBBLAS` & co rather than using `$LIBBLAS` using `os.environ['LIBBLAS']`, etc.; that'll make the logs clearer

easybuild/easyblocks/s/siesta.py
```diff
@@ -0,0 +1,162 @@
...
+    def sanity_check_step(self):
+        """Custom sanity check for Siesta."""
+
+        bins = ['bin/siesta']
+
+        if self.cfg['with_transiesta']:
+            bins.extend(['bin/transiesta'])
+
+        if self.cfg['with_utils']:
+            bins.extend(['bin/' + util for util in 'eigfat2plot', 'new.gnubands', 'gnubands', 'ccViz', 'mprop',
```
boegel: please use `os.path.join`: `os.path.join('bin', util)

easybuild/easyblocks/s/siesta.py
```diff
@@ -0,0 +1,162 @@
...
+                         'ioncat', 'grid2cdf', 'cdf2xsf', 'cdf2grid', 'grid2val', 'grid2cube', 'hsx2hs', 'hs2hsx',
+                         'get_chem_labels', 'pi3', 'macroave', 'lwf2cdf', 'swarm', 'simplex', 'orbmol_proj', 'plstm',
+                         'simple', 'driver', 'para', 'tbtrans', 'mixps', 'fractional', 'fcbuild', 'vibrator', 'readwf',
+                         'readwfx', 'info_wfsx', 'wfs2wfsx', 'wfsx2wfs', 'wfsnc2wfsx', 'pdosxml', 'xml2psf', 'test-xml'
+                         ])
+
+        custom_paths = {
+            'files': bins,
+            'dirs': []
+            }
```
boegel: dedent `}` to same column as `c` in `custom_paths`

easybuild/easyblocks/s/siesta.py
```diff
@@ -0,0 +1,162 @@
...
+            bins.extend(['Obj2/transiesta'])
+
+        if self.cfg['with_utils']:
+            self.cfg['buildopts'] += ' && cd ../Util && sh ./build_all.sh'
+
+            bins.extend(['Util/Bands/eigfat2plot', 'Util/Bands/new.gnubands', 'Util/Bands/gnubands'])
+
+            bins.extend(['Util/CMLComp/ccViz'])
+
+            bins.extend(['Util/COOP/mprop', 'Util/COOP/dm_creator', 'Util/COOP/fat'])
```
boegel: I feel this can be made shorter quite a bit... There's no need for the intermittent empty lines if you structure things a bit more imho.

Try to avoid copy-pasting the same dir names (but don't take that to the extreme, using `Examples` twice for `Util/Denchar` is OK imho), and use `os.path.join`.

Also, sort binary names alphabetically (makes it easier to spot missing things or problems later).

```python
bins.extend(os.path.join('Bands', b) for b in ['eigfat2plot', 'gnubands', 'new.gnubands'])
bins.append(os.path.join('CMLComp', 'ccViz'))
bins.extend(os.path.join('COOP', b) for b in ['dm_creator', 'fat', 'mprop'])
...
self.cfg['files_to_copy'] = [([os.path.join('Utils', b) for b in bins], 'bin')]
```

easybuild/easyblocks/s/siesta.py
```diff
@@ -0,0 +1,162 @@
...
+            bins.extend(['bin/transiesta'])
+
+        if self.cfg['with_utils']:
+            bins.extend(['bin/' + util for util in 'eigfat2plot', 'new.gnubands', 'gnubands', 'ccViz', 'mprop',
+                         'dm_creator', 'fat', 'eig2bxsf', 'xv2xsf', 'md2axsf', 'rho2xsf', 'vib2xsf', 'fmpdos',
+                         '2dplot.py', 'surf.py', 'denchar', 'dm2cdf', 'cdf2dm', 'Eig2DOS', 'ionplot.sh', 'gen-basis',
+                         'ioncat', 'grid2cdf', 'cdf2xsf', 'cdf2grid', 'grid2val', 'grid2cube', 'hsx2hs', 'hs2hsx',
+                         'get_chem_labels', 'pi3', 'macroave', 'lwf2cdf', 'swarm', 'simplex', 'orbmol_proj', 'plstm',
+                         'simple', 'driver', 'para', 'tbtrans', 'mixps', 'fractional', 'fcbuild', 'vibrator', 'readwf',
+                         'readwfx', 'info_wfsx', 'wfs2wfsx', 'wfsx2wfs', 'wfsnc2wfsx', 'pdosxml', 'xml2psf', 'test-xml'
```
boegel: this list is quite exhaustive, what are the chances of binaries being added or disappearing in future versions?

easybuild/easyblocks/s/siesta.py
migueldiascosta: well, I meant to override, not extend, the `MakeCp.extra_options` (mainly because in `MakeCp`, `files_to_copy` is mandatory in the sense that it must be present in the easyconfig, and here I'm only using it in the easyblock)...

easybuild/easyblocks/s/siesta.py
migueldiascosta: High, I suppose...

Just saw that they are changing their license and development platform (http://departments.icmab.es/leem/siesta/News/siesta-GPL.txt), but the build system is the same.

Looked at siesta-4.0-b2, at least gnubands is not longer built, only new.gnubands, so at least than can be dropped, will check more differences.

We could of course make all of this simpler and more generic if we build in install dir and use `ConfigureMake` instead of `MakeCp`, keeping all the utils in their original place, which I guess is what the developers intend...

easybuild/easyblocks/s/siesta.py
boegel: @ migueldiascosta  You can do this by passing the `extra` dict you have here to `MakeCp.extra_options`, no need to go through `ConfigureMake`.

You can also simplify things a bit, I don't think letting the `extra_options` of `EB_Siesta` accept `extra_vars` is very useful, since this is a software-specific easyblock. So, I'd go with:

```python
extra = {
    'files_to_copy': [[], "List of files or dirs to copy", CUSTOM],
    'with_transiesta': [True, "Build transiesta", CUSTOM],
    'with_utils': [True, "Build all utils", CUSTOM],
}
return MakeCp.extra_options(extra_vars=extra)
```

easybuild/easyblocks/s/siesta.py
boegel: I would either only check a handful of binaries here, or go with your other suggestions of building-in-installdir.

The latter has the disadvantage that the installation directory will not be 'clean' (source & object files will be around there too). But that's not a terrible problem imho.

easybuild/easyblocks/s/siesta.py
```diff
@@ -0,0 +1,113 @@
...
+
+    def __init__(self, *args, **kwargs):
+        """Initialisation of custom class variables for Siesta."""
+        super(EB_Siesta, self).__init__(*args, **kwargs)
+        self.build_in_installdir = True
+
+    def extract_step(self):
+        """Extract sources."""
+        # strip off 'siesta-<version>' part to avoid having everything in a subdirectory
+        self.cfg['unpack_options'] = "--strip-components=1"
```
boegel: @migueldiascosta please use `self.cfg.update`, in order not to overwrite any potential additional unpack options that are specified in the easyconfig file:

```python
self.cfg.update('unpack_options', "--strip-components=1")
```

easybuild/easyblocks/s/siesta.py
```diff
@@ -0,0 +1,113 @@
...
+        if self.toolchain.options.get('usempi', None):
+            cfg_cmd += ' --enable-mpi '
+
+        cfg_cmd += '--with-blas="' + os.environ['LIBBLAS'] + '" '
+        cfg_cmd += '--with-lapack="' + os.environ['LIBLAPACK'] + '" '
+        cfg_cmd += '--with-blacs="' + os.environ['LIBBLACS'] + '" '
+        cfg_cmd += '--with-scalapack="' + os.environ['LIBSCALAPACK'] + '"'
+
+        # make sure packaged lapack is not on generated arch.make
+        sed_cmd = "sed -i 's/COMP_LIBS=dc_lapack.a/COMP_LIBS=/' arch.make"
```
boegel: @migueldiascosta not a big fan of this, since we have a dedicated function for this kind of runtime patching, see http://easybuild.readthedocs.io/en/latest/Extended_dry_run.html#extended-dry-run-guidelines-easyblocks-framework-functions-runtime-patching

easybuild/easyblocks/s/siesta.py
```diff
@@ -0,0 +1,113 @@
...
+        self.cfg.update('prebuildopts', 'cd Obj && ' + cfg_cmd + ' && ' + sed_cmd + ' && ')
+
+        if self.cfg['with_transiesta']:
+            self.cfg.update('buildopts', ' && cd .. && mkdir Obj2 && cd Obj2 && ')
+            self.cfg.update('buildopts', cfg_cmd + ' && ' + sed_cmd + ' && make transiesta ')
+
+        if self.cfg['with_utils']:
+            self.cfg.update('buildopts', ' && cd ../Util && sh ./build_all.sh')
+
+        super(EB_Siesta, self).build_step()
```
boegel: @migueldiascosta rather than relying on the `build_step` of the parent, it would be better if you actually execute each of the command separately yourself via `run_cmd`, and take care of the changing into a different directory via Python (`os.chdir` wrapper in `try`/`except OSError`)

This way, it will be *a lot* easier to debug in case something goes wrong, since it will be clear which command is failing...

Let me know what you think about this.

easybuild/easyblocks/s/siesta.py
```diff
@@ -0,0 +1,113 @@
...
+
+    def install_step(self):
+        """Custom install procedure for Siesta."""
+
+        bins = ['Obj/siesta']
+
+        if self.cfg['with_transiesta']:
+            bins.extend(['Obj2/transiesta'])
+
+        self.cfg['files_to_copy'] = [(bins, 'bin')]
```
boegel: @migueldiascosta isn't there anything additional to copy in case of `with_utils`?

boegel:
@migueldiascosta the current easyblock works (cfr. test reports in https://github.com/hpcugent/easybuild-easyconfigs/pull/2965), but I feel it could be improved a bit (to make our future self happy ;-))

Let me know what you think about the remarks, and whether there's something I can do to help.

easybuild/easyblocks/s/siesta.py
migueldiascosta: @boegel will do, good to know about that function

migueldiascosta: @boegel more generally, the configure script in siesta is apparently unmaintained (https://bugs.launchpad.net/siesta/+bug/1627936). They recommend adapting one of the example arch.make files instead. One could copy and patch one of those, but in that case one might as well create one from scratch, it would be more robust - what do you think?

easybuild/easyblocks/s/siesta.py
migueldiascosta: @boegel there are a lot of things, we ended up setting `build_in_installdir` and leaving them in their own folders (https://github.com/hpcugent/easybuild-easyblocks/pull/923#r64148570, d94ebc8418001f2554a5925f8853c2aaddcf4892)

easybuild/easyblocks/s/siesta.py
migueldiascosta: will do

easybuild/easyblocks/s/siesta.py
boegel: Ah, OK, but then there's no point at all in calling out to the `install_step` of the parent, is there? You should then just *skip* the install step?

migueldiascosta: well, this is telling it to copy `siesta` and `transiesta` to the `bin` folder (so that those two are in the path). It's the parent's (`MakeCp`) `install_step` that actually copies `files_to_copy`, one can do it here but it seems redundant, no?

easybuild/easyblocks/s/siesta.py
migueldiascosta: @boegel attached is the one generated by configure with the current easyblock and easyconfig, can be simplified

[arch.make.txt](https://github.com/hpcugent/easybuild-easyblocks/files/509952/arch.make.txt)

boegel:
@migueldiascosta is this still WIP?

migueldiascosta:
@boegel in the latest release (4.1-b2), upstream removed the configure script (which was not working correctly) and now ships two template arch.make files (one for gfortran and one for intel) which need to be customized

I have a local version of the easyblock and easyconfig for this version but since this is still a beta release, thought I'd wait for the next stable release

boegel:
@migueldiascosta OK, let's revisit this once there's a stable release

boegel:
@migueldiascosta This can be closed in favour of #1105?

migueldiascosta:
naturally :)

Issue #917, opened 2016-05-12
impi MIC_LD_LIBRARY_PATH


stachon:
I think impi should set MIC_LD_LIBRARY_PATH (in the same way as imkl already does) to mic/lib/

boegel:
@stachon makes sense... up for looking into a PR for that?

stachon:
ok I will prepare a PR

boegel:
@stachon fixed in #925, so can be closed?

stachon:
yes, thanks!

PR #925, opened 2016-05-17
#917 add MIC_LD_LIBRARY_PATH to impi

stachon:
The directory should be valid for all versions I was able to check :
4.1.1.036, 5.0.3.048, 5.1.1.109, 5.1.2.150, and 2017 beta #917


boegel:
lgtm, thanks @stachon!

PR #928, opened 2016-05-18
Add the ability to specify a custom parallel launcher

valtandor:
This addition to the EasyBlock allows the user to specify, in the easyconfig, a launching string if a different MPI launcher is used (for example, `srun`).

### DIFF
--- a/easybuild/easyblocks/m/molpro.py
+++ b/easybuild/easyblocks/m/molpro.py
@@ -220,7 +220,7 @@ class EB_Molpro(ConfigureMake, Binary):
                 }
                 run_cmd_qa(cmd, qa=qa, std_qa=stdqa, log_all=True, simple=True)
 
-            molpro_path = os.path.join(self.install_dir, 'bin', 'molpro')
+            molpro_path = os.path.join(self.installdir, 'bin', 'molpro')
         else:
             if os.path.isfile(self.license_token):
                 run_cmd("make tuning")
--- a/easybuild/easyblocks/m/molpro.py
+++ b/easybuild/easyblocks/m/molpro.py
@@ -53,6 +53,7 @@ class EB_Molpro(ConfigureMake, Binary):
         extra_vars = ConfigureMake.extra_options(extra_vars)
         extra_vars.update({
             'precompiled_binaries': [False, "Are we installing precompiled binaries?", CUSTOM],
+            'parallel_launcher': [None, "Custom parallel launcher string", CUSTOM],
         })
         return EasyBlock.extra_options(extra_vars)
 
@@ -218,6 +219,8 @@ class EB_Molpro(ConfigureMake, Binary):
                     r"directory .* does not exist, try to create [Y]/n\n": '',
                 }
                 run_cmd_qa(cmd, qa=qa, std_qa=stdqa, log_all=True, simple=True)
+
+            molpro_path = os.path.join(self.install_dir, 'bin', 'molpro')
         else:
             if os.path.isfile(self.license_token):
                 run_cmd("make tuning")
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/m/molpro.py
```diff
@@ -218,6 +225,8 @@ def install_step(self):
                     r"directory .* does not exist, try to create [Y]/n\n": '',
                 }
                 run_cmd_qa(cmd, qa=qa, std_qa=stdqa, log_all=True, simple=True)
+
+            molpro_path = os.path.join(self.installdir, 'bin', 'molpro')
```
boegel: path is the same in both `if` and `else` bodies, so hoist it above the `if`?

easybuild/easyblocks/m/molpro.py
```diff
@@ -227,7 +236,14 @@ def install_step(self):
...
             molpro_path = os.path.join(self.full_prefix, 'bin', 'molpro')
-            apply_regex_substitutions(molpro_path, [(r"^(LAUNCHER\s*=\s*).*$", r"\1 %s" % self.orig_launcher)])
+
+        # Make sure that the launcher text we replace (where this is done) is not
+        # the instance of setting LAUNCHER that contains the eval.
+        launchertext = r"^(LAUNCHER\s*=\s*)((?!\"`eval echo \$LAUNCHER` \$MOLPRO_OPTIONS\").)*$"
+        if self.cfg['parallel_launcher'] is not None:
+            apply_regex_substitutions(molpro_path, [(launchertext, r'\1"%s"' % self.cfg['parallel_launcher'])])
+        elif self.orig_launcher is not None:
+            apply_regex_substitutions(molpro_path, [(launchertext, r"\1%s" % self.orig_launcher)])
```
boegel: you dropped a space before the `%s`, not sure how space-sensitive things are, but better leave it in place? same above

boegel:
@valtandor some minor remarks, please ping me when they're tackled since Travis doesn't notify of updates

valtandor:
@boegel I will do, but it will need to wait for about two or three weeks as I'm on leave as of today with very limited access to my usual machines.

easybuild/easyblocks/m/molpro.py
valtandor: In fact, in one case `self.installdir` was used, and in the other, `self.full_prefix`, so not the same. However, I have set a variable `destroot` and used that instead, so that only one call to `os.path.join` is needed, and can be done outside the if statement.

valtandor:
@boegel: This should have the required changes implemented.

easybuild/easyblocks/m/molpro.py
```diff
@@ -218,16 +227,24 @@ def install_step(self):
...
+
+        molpro_exe = os.path.join(destroot, 'bin', 'molpro')
+
+        # Make sure that the launcher text we replace (where this is done) is not
+        # the instance of setting LAUNCHER that contains the eval.
+        launchertext = r"^(LAUNCHER\s*=\s*)((?!\"`eval echo \$LAUNCHER` \$MOLPRO_OPTIONS\").)*$"
+        if self.cfg['parallel_launcher'] is not None:
+            apply_regex_substitutions(molpro_exe, [(launchertext, r'\1 "%s"' % self.cfg['parallel_launcher'])])
+        elif self.orig_launcher is not None:
+            apply_regex_substitutions(molpro_exe, [(launchertext, r"\1 %s" % self.orig_launcher)])
```
boegel: @valtandor while testing this, I noticed that the syntax here is not correct, i.e.:

```
$ grep '^LAUNCHER' $EASYBUILD_PREFIX/software/Molpro/*/*/bin/molpro
LAUNCHER= /path/to/software/impi/5.0.2.044-iccifort-2015.1.133-GCC-4.9.2/bin64/mpiexec -machinefile %h -n %n %x
```

the space after the `=`, that's just not right in a bash script...

This issue was already there in the easyblock however, you didn't introduce it. But we should fix this while we're at it.

This additional change fixes that problem, please review/merge: https://github.com/valtandor/easybuild-easyblocks/pull/4

boegel: @valtandor ping?

PR #940, opened 2016-05-30
added support for java6

pescobar:
I know this is not ready to merge. @boegel please review? :)

This is an easyconfig to test
https://gist.github.com/10a77ea9b266f3020b40bbfb8c582f51

To download java jdk6: 
`wget --no-check-certificate --no-cookies --header "Cookie: oraclelicense=accept-securebackup-cookie" http://download.oracle.com/otn-pub/java/jdk/6u24-b07/jdk-6u24-linux-x64.bin`

### DIFF
--- a/easybuild/easyblocks/j/java.py
+++ b/easybuild/easyblocks/j/java.py
@@ -27,13 +27,32 @@ EasyBlock for installing Java, implemented as an easyblock
 
 @author: Jens Timmerman (Ghent University)
 """
+from distutils.version import LooseVersion
 from easybuild.easyblocks.generic.packedbinary import PackedBinary
+from easybuild.tools.run import run_cmd
 
 
 class EB_Java(PackedBinary):
+#class PackedBinary(Binary, EasyBlock):
     """Support for installing Java as a packed binary file (.tar.gz)
     Use the PackedBinary easyblock and set some extra paths.
     """
+    
+    def extract_step(self):
+        """Unpack the source"""
+        #print LooseVersion(self.version)
+        #print self.src
+        print self.builddir
+        if LooseVersion(self.version) < LooseVersion('1.7'):
+            # uncompress java6 to build dir
+            cmd = 'cd %s; yes | bash %s' % (self.builddir, self.src[0]['path'])
+            #print cmd
+            run_cmd(cmd, log_all=True, simple=True)
+            cmd2 = 'cp -r %s/* %s' % (self.builddir, self.installdir)
+            run_cmd(cmd2, log_all=True, simple=True)
+        else:
+            PackedBinary.extract_step(self)
+
     def make_module_extra(self):
         """
         Set JAVA_HOME to install dir


easybuild/easyblocks/j/java.py
```diff
@@ -27,13 +27,32 @@
...
     """
+    
+    def extract_step(self):
+        """Unpack the source"""
+        #print LooseVersion(self.version)
+        #print self.src
+        print self.builddir
+        if LooseVersion(self.version) < LooseVersion('1.7'):
+            # uncompress java6 to build dir
+            cmd = 'cd %s; yes | bash %s' % (self.builddir, self.src[0]['path'])
```
boegel: seems like you should use `run_cmd_qa` instead? what type of question is being asked?

easybuild/easyblocks/j/java.py
```diff
@@ -27,13 +27,32 @@
...
+            #print cmd
+            run_cmd(cmd, log_all=True, simple=True)
+            cmd2 = 'cp -r %s/* %s' % (self.builddir, self.installdir)
```
boegel: this should be done in `install_step`?

boegel: and using `shutil.copytree` rather than `cp`/`run_cmd`

easybuild/easyblocks/j/java.py
```diff
@@ -27,13 +27,32 @@
 
 @author: Jens Timmerman (Ghent University)
 """
+from distutils.version import LooseVersion
 from easybuild.easyblocks.generic.packedbinary import PackedBinary
+from easybuild.tools.run import run_cmd
 
 
 class EB_Java(PackedBinary):
+#class PackedBinary(Binary, EasyBlock):
```
boegel: drop this?

easybuild/easyblocks/j/java.py
pescobar: this is what you get...you just need to press enter that's why I did the `yes` trick...seems simpler to me but I will implement it as you prefer...

```
Java(TM) SE Development Kit 6 successfully installed.

Product Registration is FREE and includes many benefits:
* Notification of new versions, patches, and updates
* Special offers on Oracle products, services and training
* Access to early releases and documentation

Product and system data will be collected. If your configuration
supports a browser, the JDK Product Registration form will
be presented. If you do not register, none of this information
will be saved. You may also register your JDK later by
opening the register.html file (located in the JDK installation
directory) in a browser.

For more information on what data Registration collects and 
how it is managed and used, see:
http://java.sun.com/javase/registration/JDKRegistrationPrivacy.html

Press Enter to continue.....

 
Done.
```

boegel: @pescobar alternatively, use the `inp` argument to `run_cmd` to specify input to `stdin`; you probably need to pass a space (rather than an empty string):

```python
run_cmd(cmd, log_all=True, simple=True, inp=' ')
```

pescobar:
I tested with a java6 easyconfig and also with a java7 easyconfig.

seem to be ready for merging

easybuild/easyblocks/j/java.py
```diff
@@ -27,13 +27,42 @@
...
 
 class EB_Java(PackedBinary):
     """Support for installing Java as a packed binary file (.tar.gz)
     Use the PackedBinary easyblock and set some extra paths.
     """
+    
+    def extract_step(self):
+        """Unpack the source"""
+        if LooseVersion(self.version) < LooseVersion('1.7'):
+            cmd = 'cd %s; bash %s' % (self.builddir, self.src[0]['path'])
```
boegel: @pescobar please do the `cd` in Python, and there's no need for `bash`, `run_cmd` already runs commands in a subshell

```python
try:
    os.chdir(self.builddir)
except OSError, err:
    raise EasyBuildError("Failed to move to build dir: %s", err)

run_cmd(self.src[0]['path'], log_all=True, simple=True, inp='')
```

easybuild/easyblocks/j/java.py
```diff
@@ -27,13 +27,42 @@
...
+        if LooseVersion(self.version) < LooseVersion('1.7'):
+            src = self.builddir + '/' + 'jdk%s' % self.version
+            dst = self.installdir
+            for item in os.listdir(src):
+                s = os.path.join(src, item)
+                d = os.path.join(dst, item)
+                if os.path.isdir(s):
+                    shutil.copytree(s, d, symlinks=True, ignore=None)
+                else:
+                    shutil.copy2(s, d)
```
boegel: @pescobar you can simplify this by removing the existing (empty) installdir and just use `copytree`:

```python
if LooseVersion(self.version) < LooseVersion('1.7'):
    try:
        os.rmdir(self.installdir)
        shutil.copytree(os.path.join(self.builddir, 'jdk%s' % self.version), self.installdir)
    except OSError, err:
        raise EasyBuildError("Failed to install by copying: %s", err)
```

easybuild/easyblocks/j/java.py
```diff
@@ -27,13 +27,45 @@
...
     Use the PackedBinary easyblock and set some extra paths.
     """
+    
+    def extract_step(self):
+        """Unpack the source"""
+        if LooseVersion(self.version) < LooseVersion('1.7'):
+            try: 
+                os.chmod(self.src[0]['path'], 0755)
+            except OSError, err:
+                raise EasyBuildError("Failed adding execution permission to java installer: %s", err)
```
boegel: @pescobar please use `adjust_permissions` (from `filetools`):

```python
adjust_permissions(self.src[0]['path'], stat.S_IXUSR, add=True)
```

Also, this will change permissions of the actual source file itself, should we worry about that (and copy it to build dir before changing permissions & running it)?

boegel:
@pescobar looks good, but I'd like to see a PR for an easyconfig for Java 1.6.x so we have a test case for this?

pescobar:
@boegel here you have https://github.com/hpcugent/easybuild-easyconfigs/pull/3577

boegel:
tested with https://github.com/hpcugent/easybuild-easyconfigs/pull/3577, looks good, going in, thanks @pescobar!

PR #941, opened 2016-05-31
also install scripts with MRtrix 0.3.14

boegel:


easybuild/easyblocks/m/mrtrix.py
```diff
@@ -79,11 +79,22 @@ def install_step(self):
...
         elif LooseVersion(self.version) >= LooseVersion('0.3.14'):
             release_dir = os.path.join(self.builddir, 'release')
+            scripts_dir = os.path.join(self.builddir, 'scripts')
             try:
                 os.rmdir(self.installdir)
                 shutil.copytree(release_dir, self.installdir)
+                shutil.copytree(scripts_dir, os.path.join(self.installdir, 'scripts'))
             except OSError as err:
-                raise EasyBuildError("Failed to copy %s to %s: %s", release_dir, self.installdir, err)
+                raise EasyBuildError("Failed to copy %s/%s to %s: %s", release_dir, scripts_dir, self.installdir, err)
```
wpoely86: `%s/%s` can be very confusing.

boegel: yeah, good point, I'll change it to ` & `

boegel:

PR #942, opened 2016-05-31
libxsmm support for CP2K

wpoely86:
@hajgato 

@boegel please review

### DIFF
--- a/easybuild/easyblocks/c/cp2k.py
+++ b/easybuild/easyblocks/c/cp2k.py
@@ -76,6 +76,7 @@ class EB_CP2K(EasyBlock):
         self.debug = ''
         self.fpic = ''
 
+        # used for both libsmm and libxsmm
         self.libsmm = ''
         self.modincpath = ''
         self.openmp = ''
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/c/cp2k.py
```diff
@@ -94,7 +95,7 @@ def extra_options():
             'extradflags': ['', "Extra DFLAGS to be added", CUSTOM],
             'ignore_regtest_fails': [False, ("Ignore failures in regression test "
                                              "(should be used with care)"), CUSTOM],
-            'maxtasks': [3, ("Maximum number of CP2K instances run at "
+            'maxtasks': [4, ("Maximum number of CP2K instances run at "
```
boegel: @wpoely86 why bump this, any specific reason?

boegel:
Looks fine.

@wpoely86: points to easyconfigs that need this?

easybuild/easyblocks/c/cp2k.py
wpoely86: this is for the hybrid build. It must be a multiple of two: each process gets two threads.

boegel: OK, fine then

hajgato:
@boegel ^

wpoely86:
Also needed for hpcugent/easybuild-easyconfigs#2889

boegel:
Thanks for the pointers... I'd like to see a successful test report for either of those (#3099 first?) before we merge this in.

boegel:
tested with https://github.com/hpcugent/easybuild-easyconfigs/pull/3099, good to go, thanks @hajgato and @wpoely86!

PR #944, opened 2016-05-31
enhance Qt easyblock to support Qt3

boegel:


easybuild/easyblocks/q/qt.py
```diff
@@ -62,7 +63,13 @@ def configure_step(self):
...
             platform = 'linux-g++-64'
         elif comp_fam in [toolchain.INTELCOMP]:  #@UndefinedVariable
-            platform = 'linux-icc-64'
+            if LooseVersion(self.version) >= LooseVersion('4'):
+                platform = 'linux-icc-64'
+            else:
+                platform = 'linux-icc'
+                # fix -fPIC flag (-KPIC is not correct for recent Intel compilers)
+                qmake_conf = os.path.join('mkspecs', platform, 'qmake.conf')
+                apply_regex_substitutions(qmake_conf, [('-KPIC', '-fPIC')])
```
wpoely86: does this have a clean error is the file does not exist?

boegel: @wpoely86 

```
>>> from easybuild.tools.filetools import apply_regex_substitutions
>>> apply_regex_substitutions('/tmp/sadfghmythergsdfdadgh', [])
  File "easybuild/tools/filetools.py", line 747, in apply_regex_substitutions
    for line in fileinput.input(path, inplace=1, backup='.orig.eb'):
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/fileinput.py", line 254, in next
    line = self.readline()
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/fileinput.py", line 325, in readline
    os.rename(self._filename, self._backupfilename)
OSError: [Errno 2] No such file or directory
```

If you want to change that, it should be tackled in `apply_regex_substitutions` itself imho, not in this easyblock.

wpoely86: true, open a PR for it :grin: 
The error should tell which file it's not finding.

boegel: @wpoely86 https://github.com/hpcugent/easybuild-framework/pull/1788

boegel:

PR #945, opened 2016-06-01
Gromacs dp

paderijk:
Hi,

This is my pull request related to the support for gromacs with double precision as discussed on the mailing list in the `easyblock` of gromacs.

In the `easyconfig` of gromacs to the following line needs to be added:

```
configopts='-DGMX_DOUBLE=on'
```

With this modification the suffix `_d` will be added, since that will be added when compiled with double precision.

Regards,

Pieter

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -35,20 +35,12 @@ Related Python packages:
 
 *Build status overview:*
 
-* **master** branch *(Python 2.4, Python 2.6, Python 2.7)*
+* **master** branch:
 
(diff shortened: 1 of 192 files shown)


paderijk:
Looks like more files got added to the merge, it's only the gromacs easyblock. Sorry for the confusion.. it's my first pull request.

boegel:
@paderijk You should target the `develop` branch when opening pull requests, that probably explains why the diff is bigger than you expected.

PR #946, opened 2016-06-01
Added support for the double-precision for gromacs

paderijk:
Hi,

This is my pull request related to the support for gromacs with double precision as discussed on the mailing list in the `easyblock` of gromacs.

In the `easyconfig` of gromacs to the following line needs to be added:

```
configopts='-DGMX_DOUBLE=on'
```

With this modification the suffix `_d` will be added, since that will be added when compiled with double precision.

Regards,

Pieter

### DIFF
--- a/easybuild/easyblocks/g/gromacs.py
+++ b/easybuild/easyblocks/g/gromacs.py
@@ -182,6 +182,11 @@ class EB_GROMACS(CMakeMake):
         if self.toolchain.options.get('usempi', None):
             suff = '_mpi'
 
+        # Add the _d suffix to the suffix, in case of the double precission
+        if '-DGMX_DOUBLE=on' in self.cfg['configopts']:
+            suff = suff+'_d'
+
+
         # in GROMACS v5.1, only 'gmx' binary is there
         # (only) in GROMACS v5.0, other binaries are symlinks to 'gmx'
         binaries = []
--- a/easybuild/easyblocks/g/gromacs.py
+++ b/easybuild/easyblocks/g/gromacs.py
@@ -184,7 +184,9 @@ class EB_GROMACS(CMakeMake):
 
         # Add the _d suffix to the suffix, in case of the double precission
         if '-DGMX_DOUBLE=on' in self.cfg['configopts']:
-            suff = suff+'_d'
+            suff = suff + '_d'
+        elif '-DGMX_DOUBLE=ON' in self.cfg['configopts']:
+            suff = suff + '_d'
 
 
         # in GROMACS v5.1, only 'gmx' binary is there
--- a/easybuild/easyblocks/g/gromacs.py
+++ b/easybuild/easyblocks/g/gromacs.py
@@ -188,7 +188,6 @@ class EB_GROMACS(CMakeMake):
         elif '-DGMX_DOUBLE=ON' in self.cfg['configopts']:
             suff = suff + '_d'
 
-
         # in GROMACS v5.1, only 'gmx' binary is there
         # (only) in GROMACS v5.0, other binaries are symlinks to 'gmx'
         binaries = []


boegel:
@paderijk Thanks for taking the effort, this is indeed a useful extension to the GROMACS easyblock.

One question though: do you think the hard check for `-DGMX_DOUBLE=on` in `configotps` is flexible enough? Or, in other words, are there other ways of setting `GMX_DOUBLE` such that builds with double precision is enabled (e.g. `-DGMX_DOUBLE=yes`)?

If so, adding support for a new custom easyconfig parameter `double_precision`, and using `double_precision = True` in the easyconfig file, leading to adding `-DGMX_DOUBLE=on` in `configopts` (in the easyblock) seems like a better approach?
I would rather go that way than try and check whether `GMX_DOUBLE` is enabled in a variety of ways.

Of course, the drawback of this would be that when people add `-DGMX_DOUBLE=on` to `configopts` in the easyconfig file with enabling `double_precision = True`, the sanity check is still going to fail...

So, maybe this is good enough, and will lead to the least surprises...Thoughts?

easybuild/easyblocks/g/gromacs.py
```diff
@@ -182,6 +182,11 @@ def sanity_check_step(self):
         if self.toolchain.options.get('usempi', None):
             suff = '_mpi'
 
+        # Add the _d suffix to the suffix, in case of the double precission
+        if '-DGMX_DOUBLE=on' in self.cfg['configopts']:
+            suff = suff+'_d'
```
boegel: @paderijk just to be really picky w.r.t. style: please add spaces around `+` for readability, and avoid duplicate empty lines below (i.e. drop on empty line)

If you just make a new commit with those changes and push it to the same branch `gromacs_dp`, this pull request will be updated (i.e. no need to open a new PR).

paderijk:
@boegel updated the code as you requested, agree on readability. 

Added the check so it also recognize the `-DGMX_DOUBLE=ON` as officially mentioned in the [gromacs documentation](http://www.gromacs.org/Documentation/Installation_Instructions/Cmake#double_precision_build).

Related to your initial question, I've no thoughts so far on what the best approach might be.

easybuild/easyblocks/g/gromacs.py
```diff
@@ -182,6 +182,12 @@ def sanity_check_step(self):
         if self.toolchain.options.get('usempi', None):
             suff = '_mpi'
 
+        # Add the _d suffix to the suffix, in case of the double precission
+        if '-DGMX_DOUBLE=on' in self.cfg['configopts']:
+            suff = suff + '_d'
+        elif '-DGMX_DOUBLE=ON' in self.cfg['configopts']:
+            suff = suff + '_d'
```
boegel: @paderijk this is exactly what I was 'afraid' of, lot's of copy-pasting and checking for something that's actually very trivial, there are probably other ways of setting `GMX_DOUBLE` such that double precision is enabled...

From the CMake documentation: "True if the constant is 1, ON, YES, TRUE, Y, or a non-zero number".

@besserox since this is CMake basically, thoughts?

paderijk: @boegel in that case I think it might be worth to incorporate it in the easyconfig file with the parameter `double_precision`.

besserox: The CMake-recommend way is normally not to change the library name, and instead build and install in different directories.

But apparently, GROMACS people decided to do so, cf file `cmake/gmxManageSuffixes.cmake` in the source tarball. 

Looking more in this file, we can apparently redefine this. I would say that adding `-DGMX_DEFAULT_SUFFIX=OFF -DGMX_LIBS_SUFFIX="" -DGMX_BINARY_SUFFIX=""` should disable this customized prefixes (not tested).

The problem with this is that scripts/build systems/users/... might expect these prefixes.

besserox: I realized that I did not answer the question :-)

One other option could be to check directly the CMake cache (after the configure step):
```
cmake -LA . | grep 'GMX_DOUBLE:BOOL=ON'
```
or (maybe less portable)
```
grep 'GMX_DOUBLE:BOOL=ON' CMakeCache.txt
```
I believe it should always be `ON` or `OFF` in there.

boegel: @besserox I think this is a good option...

@paderijk up for giving this a shot? I would run `cmake -LA .`, and then use Python regular expressions to check for `GMX_DOUBLE:BOOL=ON` in the output rather than using `grep`...

Quick & dirty try (untested):

```python
out, _ = run_cmd("cmake -LA .", simple=False)
if re.search("GMX_DOUBLE:BOOL=ON", out):
    suff = suff + '_d'
```

besserox: Sorry for the wrong hint. This is not working. The CMake cache still has the same value has on the command line.

The other approach that I just tested needs an addition CMake script. For example `cmake_test_variable.cmake`:
```
if(${VARNAME})
  message("${VARNAME} is ON")
else()
  message("${VARNAME} is OFF")
endif()
```
that you have to execute like this:
```
cmake -DVARNAME="GMX_DOUBLE" -P cmake_test_variable.cmake .
```

besserox: Still looking for a better solution, what about:
```
cmake -LA . | egrep '^GMX_LIBS_SUFFIX'
cmake -LA . | egrep '^GMX_BINARY_SUFFIX'
```

It should return the suffixes that have be decided by GROMACS configuration.
(I have not tested that though)

boegel: Hmm, no, all I'm getting is this:

```
$ cmake -LA .  | grep SUF
GMX_DEFAULT_SUFFIX:BOOL=ON
$
```

boegel: I do see this though... Does this mean we need to parse `CMakeCache.txt`? :(

```
$ grep SUFF CMakeCache.txt 
GMX_DEFAULT_SUFFIX:BOOL=ON
GMX_BINARY_SUFFIX:INTERNAL=
//Previous value of GMX_DEFAULT_SUFFIX for change tracking
GMX_DEFAULT_SUFFIX_PREVIOUS_VALUE:INTERNAL=ON
GMX_LIBS_SUFFIX:INTERNAL=
```

boegel: I tackled this in #960 by adding a custom easyconfig parameter `double_precision`, and using a regex to check whether `GMX_DOUBLE` is set, cfr. 812c304

boegel:
merged via #960, thanks @paderijk!

PR #950, opened 2016-06-03
Easyblocks for Anaconda and Conda Env

jerowe:
Here are two easyblocks, referenced here: https://github.com/hpcugent/easybuild-framework/issues/1791

The first, anaconda.py, installs the python distro anaconda as a module.

The second, condacreate.py, uses the conda create utility to create a conda environment, and optionally installs requirements into that environment. http://conda.pydata.org/docs/using/envs.html

**Acknowledgements** This work is being sponsored by NYU Abu Dhabi Center for Genomics and Systems Biology


boegel:
@jerowe do you have example easyconfigs for these too? if so, please open a PR for those too (and refer to this PR as a requirement in the PR description)

jerowe:
@boegel, the PR for the configs is here. [https://github.com/hpcugent/easybuild-easyconfigs/pull/3173](PR 3173)

boegel:
@cfenoy Would you have some time to take a look at this?

jerowe:
Configs are now here:

https://github.com/hpcugent/easybuild-easyconfigs/pull/3337

Dropbox ruined my morning.

jerowe:
Hi there.

I've been on vacation and dropped the ball on this. I'm going to go over the comments. I know I needed to add more files for tests, so I will get on that.

In regards to changing 'EB_anaconda' to just 'anaconda', when I try that it doesn't find the module. I would imagine that is because I'm running it in dev mode with 

```
eb --include-easyblocks=/path/to/easybuild/easyblocks/a/anaconda.py anaconda-2.4.0.0.eb
```

?

boegel:
@jerowe why would you want to change the class name to just `anaconda`? If you want EasyBuild to automagically pick up the easyblock for installing `anaconda`, you need to stick to the class naming scheme which involves prefixing the class name with `EB_`...

jerowe:
I may have misread that.  In any case I didn't change it.  ;)

jerowe:
I think it was that my configs should have 

```
easyblock = 'anaconda'
``

Instead of 

```
easyblock = 'EB_anaconda'
```

But when I try this I get an error saying the easyblock is not found.

**Comment by boegel** [comment]
@jerowe easyblocks that do not start with `EB_` are only looked for in the `generic` subnamespace, so `anaconda.py` should be located in `easybuild/easyblocks/generic/anaconda.py`, or be included with `--include-easyblocks /path/to/generic/anaconda.py`

However, it's better to rely on the automagic derivation of which easyblock should be used based on the software name in this case imho (`name = 'anaconda'` => `EB_anaconda`)

**Comment by jerowe** [comment]
Thanks for the clarification. I think I get it now!

**Comment by jerowe** [comment]
I had one other question I forgot to ask.

In each of these modules, Anaconda, CondaCreate, and CondaEnv, I have a few functions that are the same, such as post_install_cmd. Where would be a good place to put these? Since they are all specific to the anaconda module should they just go there? I don't like having duplicate functionality hanging out in there, but I wasn't sure where would make the most sense to put it.

**Comment by boegel** [comment]
@jerowe I think it makes sense to add them in the Anaconda easyblock, and then either import those functions in the other easyblocks, or use inheritance to derive `CondaCreate` from `Anaconda` and override the methods that are different.

Let me know if this makes sense.

**Comment by jerowe** [comment]
@boegel, that's what I was thinking. Thanks for the input. ;-) I will implement them and push the changes.

Do you think these will be in the 2.9.0 release? I see them listed in 2.x.

**Comment by phineas42** [comment]
@jerowe Can you briefly describe how a user would use software built using CondaEnv? When I tried it, the build worked marvelously, and I could load with Lmod, but there was no supplemental environment listed when I ran "conda info --envs" and the custom modules were not found when running import within Python.

For what it's worth, I modified your example anaconda easyconfig to match my group's existing style, which basically tacks on the Python short version to the easyconfig name:
```
# Anaconda Easyconfig
easyblock = 'EB_anaconda'
name = "Anaconda2"
anaconda_version = 2
version = "4.1.1"
sources = ["%s-%s-%s.sh" % (name,version,variant)]
checksums = [ 'f7bb3c0ccf23c9789bb895335aa68bf3' ]
```

```
# CondaEnv easyconfig
easyblock = 'EB_CondaEnv'
name = "condaenv-test"
builddependencies = [('Anaconda2', '4.1.1')]
sources=['environment.yml']
```

```
# environment.yml
name: test
channels:
  - ncar
  - khallock
dependencies:
  - basemap=1.0.7    #-c anaconda
  - pynio=1.4.3      #-c ncar
  - pyngl=dev_r16541 #-c khallock
```

...and if it looks like my local changes broke something, feel free to say so and I won't bug you any more.  :sweat_smile:

**Comment by jerowe** [comment]
@phineas, this has to do with the design choices made by conda env.

If you create conda envs through the usual mechanism you run a 

```
conda env create #with an environment.yml file in the cwd
conda env create --file environment-1.0.yml
```
These environments are added to $ANACONDA_ROOT/envs/

We are creating the conda envs using the --prefix opt, and conda does not 'know' about them until you activate them.

You can get around this a bit by adding your self.install_dir minus the version number to the ~/.condarc envs_dirs options.

So if I create an condaenv easyblock module test/1.0

```
conda config --add envs_dirs $EASYBUILDS_ROOT/software/test/
```

But then you're env names are the version names, which is not terribly pretty or informative. ;-) Conda env does not have a mechanism to figure out version numbers (yet!).

Another way to think of it is:

```
conda env create
```
Conda manages your software stack

```
conda env create -p
```
YOU manage your software stack. I would think you would be a sane person and manage it with environment modules, but maybe you decide to go crazy with something else.

To make a long story short just load the module as you would normally. `module load condaenv-test/1.0' Its like running a source activate. All that source activate does is prepend your path and set the CONDA_ENV and CONDA_DEFAULT_ENV to the root dir of that env.

jerowe:
@phineas42, glad you're finding the modules useful!  ;-)

I just wanted to point out that using anaconda_version in the easyconfig was something I was playing with initially, but got rid of. It shouldn't be doing anything.

jerowe:
@boegel  - I moved the duplicate functions on over to the anaconda module, reran tests, and all is well now. ;-)

boegel:
@jerowe OK, I'll try and find time to give this a detailed look soon, but I don't think I'll be able to do that before the EasyBuild v2.9.0 release (there are a couple of things I'd like to take care of in the framework part first, and I was hoping to release v2.9.0 early Sept).

jerowe:
There no rush.  I just would like to give my boss a timeline.

Thanks for all your help!

phineas42:
@jerowe Thanks for the explanation! With that knowledge I was able to discern that the cause of my issue...

> custom modules were not found when running import within Python.

...was due to my attempting "import pynio" instead of "import Nio".

I like your approach of prepending the path and setting CONDA_ENV and CONDA_DEFAULT_ENV.

jerowe:
Hi @boegel , how is it going?

Any word on which release I can aim for?

boegel:
@jerowe I think we should be able to get this in for EasyBuild v3.0.0 (planned for mid Nov).

I just need to find time to test this.

jerowe:
@boegel, that sounds great! Thanks! We've been using Easybuild in production since August, and have been really pleased with it so far. ;-)

jerowe:
@boegel , I don't want to be pushy, but is there anything I can do to move this along? We have been using this method in production for our modules since August, and its been marvelous. There is internal documentation on the process, but I've been holding off on public documentation until these blocks are a part of easybuild core. 

If you go [here](https://github.com/nyuad-cgsb/nyuad-hpc-module-configs/tree/master/_easybuild) you will see that we have a build process that goes from conda env config -> easyblock config. As you can see, easybuild has become a huge part of our infrastructure. ;-)

For my part I will take a closer look at the changes you made to the anaconda block, and see if I can figure out some ways to clean up the CondaCreate and CondaEnv blocks.

boegel:
Don't worry about coming over as pushy, I appreciate that people bump PRs that are open for a while and are (still) very valuable to get merged.
I've been wanting to get back to it, but haven't found timefor it yet. I really want to get this in for the next EasyBuild release. I don't think it needsa lot of work, but there's just too much copy-pasting going on thatisgoing to come back to haunt us later.
Let me know if you've had any luck on reworking the Conda* easyblocks. I may be able to find some time myself this weekend...

boegel:
@jerowe I gave this another look, there's quite a bit of opportunity for further cleanup imho.

See https://github.com/jerowe/easybuild-easyblocks/pull/2/files as another step in the right (according to me) direction. Let me know what you think.

I also think the `(EB_)CondaCreate` and `EB_CondaEnv` easyblocks should be merged into a single generic easyblock named `Conda` instead, with a switch to pick between `conda create` + `conda install` (i.e. what the `CondaCreate` easyblock now implements), and `conda env create`.

Maybe it's worthwhile to set up a short conf call to discuss this further, e.g. sometime next week?

jerowe:
@boegel , thanks for taking the time to look things over! I'm really not a python developer, so this is very helpful for me.

I could do a conference call. I live in the UAE. Gulf Time Zone. UTC+04:00. Or just google 'time in dubai' ;-) What time zone are you in? So far any day but Thursday is good for me.

boegel:
@jerowe I sent you a mail to figure out a good time for a conf call; I may spend some time this weekend on merging together `CondaCreate` and `CondaEnv` into a single generic `Conda` easyblock, I think that makes most sense.

boegel:
Tested thoroughly with https://github.com/hpcugent/easybuild-easyconfigs/pull/3337, good to go now.

Thanks a lot of all your work on this @jerowe, and for your patience!

PR #952, opened 2016-06-08
fix make_installdir in Inspector & VTune easyblocks

boegel:
Fix for problem reported by @jhein32: making `make_installdir` a no-op has the unwanted side-effect of not cleaning up the installation directory  when doing a forced reinstallation, which leads to:

```
 == 2016-06-08 11:40:31,549 run.py:99 DEBUG run_cmd: running cmd ./install.sh  -s /local/easybuild/build/VTune/2016_update3/dummy-dummy/vtune_amplifier_xe_2016_update3/silent.cfg (in /local/easybuild/build
/VTune/2016_update3/dummy-dummy/vtune_amplifier_xe_2016_update3)
== 2016-06-08 11:40:31,549 run.py:107 DEBUG run_cmd: Command output will be logged to /tmp/eb-c_vKEE/easybuild-run_cmd-N5ooC7.log
== 2016-06-08 11:40:42,289 run.py:392 DEBUG cmd "./install.sh  -s /local/easybuild/build/VTune/2016_update3/dummy-dummy/vtune_amplifier_xe_2016_update3/silent.cfg" exited with exitcode 2 and output:
WARNING: Destination directory already exists. 
Installing to this directory could overwrite existing files and folders within
this directory.
--------------------------------------------------------------------------------
```

This change makes sure that the install directory is not created, but it does get cleaned up when it needs to be, e.g. during a forced re-installation.

### DIFF
--- a/easybuild/easyblocks/i/inspector.py
+++ b/easybuild/easyblocks/i/inspector.py
@@ -49,7 +49,7 @@ class EB_Inspector(IntelBase):
 
     def make_installdir(self):
         """Do not create installation directory, install script handles that already."""
-        pass
+        super(EB_Inspector, self).make_installdir(dontcreate=True)
 
     def install_step(self):
         """
--- a/easybuild/easyblocks/v/vtune.py
+++ b/easybuild/easyblocks/v/vtune.py
@@ -49,7 +49,7 @@ class EB_VTune(IntelBase):
 
     def make_installdir(self):
         """Do not create installation directory, install script handles that already."""
-        pass
+        super(EB_VTune, self).make_installdir(dontcreate=True)
 
     def install_step(self):
         """


boegel:
@jhein32: can you please verify this fix? Should be trivial enough to apply this patch manually, or you can download the updated easyblock (from https://raw.githubusercontent.com/boegel/easybuild-easyblocks/202c69b3704605ac2c6cbe277e79b407bea2236d/easybuild/easyblocks/v/vtune.py) and use `--include-easyblocks`

jhein32:
Hi Kenneth,
I manually downloaded and it appears to work.  I only got it to pick up the new file by specifying it in absolute path (didn't try relative path).  From the documentation (e.g. eb --help) I assumed the --include-easyblocks asks for the directly where the vtune.py lives.
Best wishes
J.

boegel:
@jhein32 no, you need to list the full paths to the easyblocks, but it does support wildcards, see also http://easybuild.readthedocs.io/en/latest/Including_additional_Python_modules.html#include-easyblocks

Thanks for verifying the fix, going in!

PR #954, opened 2016-06-10
Make QE easyblock aware of multithreaded FFT

ocaisa:
No need for `hybrid`, just get the trigger from the `toolchainopts` (which is needed anyway to get the threaded libs for all the math libraries)

edit: depends on https://github.com/hpcugent/easybuild-framework/pull/1802

### DIFF
--- a/easybuild/easyblocks/q/quantumespresso.py
+++ b/easybuild/easyblocks/q/quantumespresso.py
@@ -127,7 +127,7 @@ class EB_QuantumESPRESSO(ConfigureMake):
         repls.append(('DFLAGS', ' '.join(dflags), False))
 
         # complete C/Fortran compiler and LD flags
-        if self.toolchain.options.get('openmp', True) or self.cfg['hybrid']:
+        if self.toolchain.options.get('openmp', False) or self.cfg['hybrid']:
             repls.append(('LDFLAGS', self.toolchain.get_flag('openmp'), True))
             repls.append(('(?:C|F90|F)FLAGS', self.toolchain.get_flag('openmp'), True))
 
--- a/easybuild/easyblocks/q/quantumespresso.py
+++ b/easybuild/easyblocks/q/quantumespresso.py
@@ -102,7 +102,7 @@ class EB_QuantumESPRESSO(ConfigureMake):
         dflags.append(comp_fam_dflags[self.toolchain.comp_family()])
 
         if self.toolchain.options.get('openmp', False):
-            libffi = os.getenv('LIBFFT_MT')
+            libfft = os.getenv('LIBFFT_MT')
         else:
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/q/quantumespresso.py
```diff
@@ -49,7 +49,6 @@ class EB_QuantumESPRESSO(ConfigureMake):
     def extra_options():
         """Custom easyconfig parameters for Quantum ESPRESSO."""
         extra_vars = {
-            'hybrid': [False, "Enable hybrid build (with OpenMP)", CUSTOM],
```
boegel: I think we need to retain this (for now), otherwise people who use `hybrid = True` now without having `openmp` enabled in `toolchainopts` will run into a different build?

ocaisa: You're right...but `hybrid = True` is broken for FFTW, it's a bit of a fluke that it works with MKL. Anyway the actual [performance is worse than with pure MPI](http://www.prace-ri.eu/IMG/pdf/enabling_ffte_library_and_fftw3_threading_in_quantum_espresso.pdf)

ocaisa: Or is the case that people would want to enable hybrid without the toolchain option?

boegel: well, they do now, and it works (at least for Intel MKL)

ocaisa:
@boegel I've left the backwards compatability there but also now the `openmp` flag will trigger the same things. Works with `foss`, will double-check against `intel` and then provide easyconfigs for QE 5.4.0

easybuild/easyblocks/q/quantumespresso.py
```diff
@@ -101,7 +101,10 @@ def configure_step(self):
         }
         dflags.append(comp_fam_dflags[self.toolchain.comp_family()])
 
-        libfft = os.getenv('LIBFFT')
+        if self.toolchain.options.get('openmp', False):
```
boegel: @ocaisa missing `or self.cfg['hybrid']`?

boegel: hmm, or maybe not, w.r.t. backwards compatibility...

ocaisa: Yeah, that's the problem, if you enable openmp in toolchainopts and hybrid using the old easyblock it wouldn't work anyway (hence your previous build error) but many people probably did not set the toolchainopt

boegel:
tested with https://github.com/hpcugent/easybuild-easyconfigs/pull/3425 and existing easyconfigs for QuantumESPRESSO, good to go, thanks @ocaisa!

PR #955, opened 2016-06-11
easyblock for Extrae

boegel:
wrapper PR for #892 by @geimer with some minor style cleanups on top


boegel:
tested with https://github.com/hpcugent/easybuild-easyconfigs/pull/507, good to go as reviewed version of #892, thanks @geimer!

PR #893, opened 2016-04-21
Initial easyblock for Paraver (WIP)

geimer:
To go in conjunction with https://github.com/hpcugent/easybuild-easyconfigs/pull/508

@boegel: Status after out meeting in Gent; may need some cleanup and further testing

PR #956, opened 2016-06-12
add easyblock for Paraver

boegel:
wrapper PR for #893 by @geimer, adding some style fixes on top


boegel:
tested with https://github.com/hpcugent/easybuild-easyconfigs/pull/508, good to go

Thanks @geimer!

PR #957, opened 2016-06-13
also install vsc-install in EasyBuildMeta easyblock, if tarball is provided

boegel:
This is required to support offline installation of EasyBuild, cfr. https://lists.ugent.be/wws/arc/easybuild/2016-06/msg00087.html

The change being made is backward-compatible, in the sense that the `vsc-install` source tarball does not *need* to be there.


boegel:
@wpoely86 please review?

wpoely86:
:+1:

boegel:

PR #958, opened 2016-06-14
easyblock for Amber (cleaned up)

boegel:
This is a cleaned up version of the easyblock provided by @thiell in #878, which itself is based on the easyblock provided by @valtandor in #491.

### DIFF
--- a/easybuild/easyblocks/a/amber.py
+++ b/easybuild/easyblocks/a/amber.py
@@ -1,0 +1,223 @@
+##
+# Copyright 2009-2015 Ghent University
+# Copyright 2015-2016 Stanford University
+#
+# This file is part of EasyBuild,
+# originally created by the HPC team of Ghent University (http://ugent.be/hpc/en),
+# with support of Ghent University (http://ugent.be/hpc),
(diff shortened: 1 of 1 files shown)


boegel:
Tested with https://github.com/hpcugent/easybuild-easyconfigs/pull/3200, `$TMPDIR` issue (discussed at https://github.com/hpcugent/easybuild-easyblocks/pull/953#issuecomment-225613062) will be handled via framework (see https://github.com/hpcugent/easybuild-framework/pull/1808).

So, good to go, thanks for your efforts on this @valtandor and @thiell!

avapirev:
Hi, are there any plans for an easyblock to be pushed in at any time which supports the Amber Tools 16 install and which is compatible with newer versions of NETcdf and HDF5? In particular, the current easyblock declares a flag "--with-netcdf-fort" which is not relevant for Amber 16 in some cases.

valtandor:
Hi @avapirev, I don't really have access to Amber 16 for building any more, so it wouldn't be by me, but I'm sure you're welcome to put together a PR for an updated EasyBlock.

boegel:
@avapirev PR for easyconfigs for Amber16 is at https://github.com/hpcugent/easybuild-easyconfigs/pull/3688

PR #951, opened 2016-06-07
Gromacs cray GPU + dynamic libs

gppezzi:

PR #960, opened 2016-06-17
update GROMACS easyblock for Cray, double precision & old GROMACS versions

boegel:
This is a combination of #569 (by @valtandor), #946 (@paderijk) and ~~#951~~ (@gppezzi), with some ~~sprinkles~~cleanup/fixes on top...

WIP because:

* [x] fix sanity check, now suffixes get included twice sometimes resulting in `libgromacs_mpi_mpi.a`
* [x] need to find better way of enabling double precision & using correct suffix in sanity check
  * cfr. discussion in #946
* [x] double-check what the intention was of the 'mdrun-only' configure options that @valtandor added in #569 ~~(see `FIXME`s)~~
   * cfr. https://github.com/hpcugent/easybuild-easyblocks/pull/569/files#r68004579
* [x] verify that none of the existing easyconfigs are broken (only tested a couple so far)
* [x] test with:
  * [x] https://github.com/hpcugent/easybuild-easyconfigs/pull/1446
  * [x] https://github.com/hpcugent/easybuild-easyconfigs/pull/2348
  * [x] https://github.com/hpcugent/easybuild-easyconfigs/pull/3051
  * [x] https://github.com/hpcugent/easybuild-easyconfigs/pull/3138
  * [x] https://github.com/hpcugent/easybuild-easyconfigs/pull/3125
  * [x] ~~https://github.com/hpcugent/easybuild-easyconfigs/pull/3195~~
  * [x] https://github.com/hpcugent/easybuild-easyconfigs/pull/3249

### DIFF
--- a/easybuild/easyblocks/g/gromacs.py
+++ b/easybuild/easyblocks/g/gromacs.py
@@ -156,10 +156,17 @@ class EB_GROMACS(CMakeMake):
                 for libname in ['BLAS', 'LAPACK']:
                     libdir = os.getenv('%s_LIB_DIR' % libname)
                     if self.toolchain.toolchain_family() == toolchain.CRAYPE:
-                        self.cfg.update('configopts', '-DGMX_%s_USER="%s/libsci_gnu_mpi_mp.a"' % (libname, libdir))
+                        libsci_mpi_mp_lib = glob.glob(os.path.join(libdir, 'libsci_*_mpi_mp.a'))
+                        if libsci_mpi_mp_lib:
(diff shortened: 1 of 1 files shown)


gppezzi:
I've noticed that the install part is also compiling code on our builds.

If it makes sense, we could add -j also to the install step. What do you think?

gppezzi:
Tested for https://github.com/hpcugent/easybuild-easyconfigs/pull/3195

Build report is [here](https://gist.github.com/gppezzi/9245089a2c865ed4263e68842a85d671)

boegel:
retested with all existing GROMACS easyconfigs + with the ones from the various GROMACS PRs, so good to go!

PR #961, opened 2016-06-20
fixes for Geant4 easyblock

boegel:
fix for https://github.com/hpcugent/easybuild-easyconfigs/issues/3213 plus making sure that environment variables that are defined don't include `None` in their value


boegel:
tested with all existing Geant4 easyconfigs, problems reported in https://github.com/hpcugent/easybuild-easyconfigs/issues/3213 are confirmed to be fixed, so going in

Thanks for reporting this @lekshmideepu!

PR #964, opened 2016-06-30
correct full path to ALADIN config file & patch it to use right Fortran compiler flags

boegel:


easybuild/easyblocks/a/aladin.py
```diff
@@ -144,16 +145,16 @@ def configure_step(self):
 
         # generate gmkpack configuration file
         self.conf_file = 'ALADIN_%s' % self.version
-        self.conf_filepath = os.path.join(self.builddir, 'arch', '%s.x' % self.conf_file)
+        self.conf_filepath = os.path.join(self.builddir, 'gmkpack_support', 'arch', '%s.x' % self.conf_file)
```
wpoely86: no version if's?

easybuild/easyblocks/a/aladin.py
```diff
@@ -144,16 +145,16 @@ def configure_step(self):
...
 
         try:
             if os.path.exists(self.conf_filepath):
                 os.remove(self.conf_filepath)
                 self.log.info("Removed existing gmpack config file %s" % self.conf_filepath)
 
-            archdir = os.path.join(self.builddir, 'arch')
+            archdir = os.path.dirname(self.conf_file)
```
wpoely86: `self.conf_file = 'ALADIN_%s' % self.version` ? It's not a path?

easybuild/easyblocks/a/aladin.py
boegel: no, this was wrong all along...

easybuild/easyblocks/a/aladin.py
boegel: yeah, should be `self.conf_filepath`, fixing that

boegel:
tested with https://github.com/hpcugent/easybuild-easyconfigs/pull/3292, works like a charm, going in

PR #965, opened 2016-07-04
PSI: prepare for PSI4 1.0

wpoely86:
- Check for CMake based on file presence instead of version
- The binaries is always called psi4

cc @tovrstra

### DIFF
--- a/easybuild/easyblocks/p/psi.py
+++ b/easybuild/easyblocks/p/psi.py
@@ -98,7 +98,7 @@ class EB_PSI(CMakeMake):
             raise EasyBuildError("Boost module not loaded.")
 
         # pre 4.0b5, they were using autotools, on newer it's CMake
-        if not os.path.exists(os.path.join(self.builddir, self.psi_srcdir, "CMakeLists.txt")):
+        if LooseVersion(self.version) <= LooseVersion("4.0b5") and self.cfg.name == "PSI":
             self.log.info("Using configure based build")
             env.setvar('PYTHON', os.path.join(pythonroot, 'bin', 'python'))
             env.setvar('USE_SYSTEM_BOOST', 'TRUE')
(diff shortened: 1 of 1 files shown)


tovrstra:
I have no idea how easyblocks work exactly but as far as I can tell this looks reasonable.

easybuild/easyblocks/p/psi.py
```diff
@@ -136,8 +137,9 @@ def configure_step(self):
 
             ConfigureMake.configure_step(self, cmd_prefix=self.cfg['start_dir'])
         else:
-            self.cfg['configopts'] += "-DPYTHON_INTERPRETER=%s " % os.path.join(pythonroot, 'bin', 'python')
-            self.cfg['configopts'] += "-DCMAKE_BUILD_TYPE=Release "
+            self.log.info("Using CMake based build")
+            self.cfg['configopts'] += " -DPYTHON_INTERPRETER=%s " % os.path.join(pythonroot, 'bin', 'python')
+            self.cfg['configopts'] += "-DCMAKE_BUILD_TYPE=RelWithDebInfo "
```
boegel: please use `self.cfg.update('configopts', "...")`

boegel: also: is just changing the build type OK? should we make that a separate parameter?

wpoely86: oeps, this shouldn't be added it

easybuild/easyblocks/p/psi.py
```diff
@@ -98,7 +98,8 @@ def configure_step(self):
             raise EasyBuildError("Boost module not loaded.")
 
         # pre 4.0b5, they were using autotools, on newer it's CMake
-        if LooseVersion(self.version) <= LooseVersion("4.0b5"):
+        if not os.path.exists(os.path.join(self.builddir, self.psi_srcdir, "CMakeLists.txt")):
+            self.log.info("Using configure based build")
```
boegel: This looks OK, but it causes problems with existing PSI easyconfigs, in case they already include `CMakeLists.txt`, but were still being installed with `./configure`, and thus don't include `CMake` as a (build) dependency.

So, this introduces a regression of sorts...

I think you need to keep the version check in place, but make an exception for installations where `PSI4` is the software name?

boegel:
tested with all existing PSI easyconfigs, good to go, thanks @wpoely86!

PR #966, opened 2016-07-06
ensure correct compiler command/flags are used for SAMtools

boegel:
fix for https://github.com/hpcugent/easybuild-easyconfigs/issues/1959

cc @backelj

### DIFF
--- a/easybuild/easyblocks/s/samtools.py
+++ b/easybuild/easyblocks/s/samtools.py
@@ -71,7 +71,7 @@ class EB_SAMtools(ConfigureMake):
 
     def configure_step(self):
         """Ensure correct compiler command & flags are used via arguments to 'make' build command"""
-        for var in ['CC', 'CFLAGS']:
+        for var in ['CC', 'CXX', 'CFLAGS', 'CXXFLAGS']:
             if var in os.environ:
                 self.cfg.update('buildopts', '%s="%s"' % (var, os.getenv(var)))
 
--- a/easybuild/easyblocks/s/samtools.py
+++ b/easybuild/easyblocks/s/samtools.py
@@ -70,10 +70,10 @@ class EB_SAMtools(ConfigureMake):
         self.lib_files = ["libbam.a"]
 
     def configure_step(self):
-        """
-        No configure
-        """
-        pass
+        """Ensure correct compiler command & flags are used via arguments to 'make' build command"""
+        for var in ['CC', 'CFLAGS']:
+            if var in os.environ:
+                self.cfg.update('buildopts', '%s="%s"' % (var, os.getenv(var)))
 
     def install_step(self):
         """


boegel:
@wpoely86 please review

tested with all 29 existing SAMtools easyconfigs, works like a charm

fgeorgatos:
entertaining how long this escaped under the radar;

do we have any means to verify when icc compiler was really picked up? is there any kind of stamp?

easybuild/easyblocks/s/samtools.py
```diff
@@ -70,10 +70,10 @@ def __init__(self, *args, **kwargs):
...
 
     def configure_step(self):
-        """
-        No configure
-        """
-        pass
+        """Ensure correct compiler command & flags are used via arguments to 'make' build command"""
+        for var in ['CC', 'CFLAGS']:
+            if var in os.environ:
+                self.cfg.update('buildopts', '%s="%s"' % (var, os.getenv(var)))
```
fgeorgatos: lgtm, btw

easybuild/easyblocks/s/samtools.py
```diff
@@ -70,10 +70,10 @@ def __init__(self, *args, **kwargs):
         self.lib_files = ["libbam.a"]
 
     def configure_step(self):
-        """
-        No configure
-        """
-        pass
+        """Ensure correct compiler command & flags are used via arguments to 'make' build command"""
+        for var in ['CC', 'CFLAGS']:
```
wpoely86: does it hurt to do it also for CXX?

boegel: it doesn't hurt, but doesn't matter, BAMtools is C-only

wpoely86:
@wpoely86 feels a warm glow of happiness

boegel:
Going in, thanks for the review @fgeorgatos an @wpoely86!

@fgeorgatos I think @pforai knows of a way to check with which compiler stuff was built, so we could look into a generic detection mechanism in framework to catch occurrences like this...

backelj:
I submitted the original ticket in Sep 10, 2015 already (https://github.com/hpcugent/easybuild-easyconfigs/issues/1959)…

I don’t recall whether we’ve discussed this somewhere, but I do recall thinking “how many other easyconfigs will also have this issue”?

Anyway, thanks for the fix, Kenneth…

— Franky

pforai:
I've played around parsing ELF structures where compilers insert identification strings(usually .comment) and the basic approach was to generate a dummy ELF with the toolchain that is selected and check the ELF for whatever string was inserted by the compiler, then go through all of the generated binaries and do string comparison there. 

The problem is that some compilers insert very precise and you can easily do comparisons while others are just inserting strings like "Intel Corporation" by default but compilation settings can be changed to get better visibility (-sox for ICC etc.).

GCC for example inserts additional ELF sections regarding ABI compatibility etc. 

I guess we can make this work as part of sanity checking.

wpoely86:
If we're going to do compiler wrappers for rpath, we might as well use them for this too?

boegel:
@wpoely86 where do compilers wrappers come into play here?

PR #976, opened 2016-08-11
Boost root env variable

jhein32:
have boost modules set up a BOOST_ROOT environment variable for use by e.g. cmake

### DIFF
--- a/easybuild/easyblocks/b/boost.py
+++ b/easybuild/easyblocks/b/boost.py
@@ -215,3 +215,10 @@ class EB_Boost(EasyBlock):
             custom_paths["files"].append('lib/libboost_python.%s' % shlib_ext)
 
         super(EB_Boost, self).sanity_check_step(custom_paths=custom_paths)
+
+    def make_module_extra(self):
+        """Set up a BOOST_ROOT environment variable to e.g. ease Boost handling by cmake"""
+        txt = super(EB_Boost, self).make_module-extra()
+        txt = self.module_generator.set_environment('BOOST_ROOT', self.installdir)
+        return txt
+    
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/b/boost.py
```diff
@@ -215,3 +215,10 @@ def sanity_check_step(self):
             custom_paths["files"].append('lib/libboost_python.%s' % shlib_ext)
 
         super(EB_Boost, self).sanity_check_step(custom_paths=custom_paths)
+
+    def make_module_extra(self):
+        """Set up a BOOST_ROOT environment variable to e.g. ease Boost handling by cmake"""
+        txt = super(EB_Boost, self).make_module-extra()
```
boegel: @jhein32 you have a syntax error here, should be `make_module_extra()`, not `make_module-extra()`

when you make changes to an existing easyblock, please make sure you test it too, by (re)building a `Boost` module in this case

you can make sure that your modified easyblock is being used via `--include-easyblocks` (see http://easybuild.readthedocs.io/en/latest/Including_additional_Python_modules.html#include-easyblocks)

easybuild/easyblocks/b/boost.py
```diff
@@ -215,3 +215,10 @@ def sanity_check_step(self):
...
+        txt = self.module_generator.set_environment('BOOST_ROOT', self.installdir)
```
boegel: @jhein32 you want `txt +=` here, to avoid overwriting the existing contents of `txt

jhein32:
Noticed the travis failures.  Edited the file via website.

jhein32:
@boegel I was fighting with the git-hub desktop and it seems to have done a merge of master into a develop (not sure my private or the hpcugent).  Could you review that this has no site-effect?

boegel:
@jhein32 you added an empty merge commit indeed, but don't worry about that too much, we're not purists there and it has no negative side-effects

PR looks OK now, I'll give it a test on my end but I don't expect problems, should be good to go

jhein32:
Ran a manual check, rebuilding one of our boost modules.  The module file contains the BOOST_ROOT variable and the value appears correct :)

boegel:
lgtm, thanks @jhein32!

PR #979, opened 2016-08-19
enhance ConfigureMake generic easyblock to allow using it to install extensions

boegel:


boegel:
@wpoely86, @hajgato: please review

@ocaisa: this is a small enhancement to the `ConfigureMake` easyblock in order to support installing a bundle of X11 packages... (PR with easyconfig coming up)

hajgato:
@boegel Its seems fine for me. I do not see what would be the consequences for other easyblocks, like in the case of `PERL` module. Is this the only one that affected?

wpoely86:
:+1:

boegel:
@hajgato the `PerlModule` generic easyblock should be the only one affected, yes; the change is needed because by making `ConfigureMake` derive from `ExtensionEasyblock`, we block multiple inheritance there...

I've been thinking some more about this, and I've been wondering whether this is the right approach. The extensions mechanism is really intended for stuff like Pyhton packages, Perl modules, R libraries, etc., i.e. language-specific packages that require a particular software package as a dependency.

By enhancing `ConfigureMake` like we do here, we've somewhat abusing this mechanism for other purposes.

I think maybe we should be enhancing the `Bundle` easyblock to support specifying a list of `components` instead, where you can indicate using which easyblock each of these components should be installed? That's a proper and more generic approach, but I'm not sure how easy/hard it would be to implement that... Thoughts?

hajgato:
@boegel I do think that `Bunde` should be enhanced as you said. The final goal should be the ability to install more than one package to a common "prefix". (But I do see soo deep into framework/blocks)

boegel:
superseded by #980

PR #981, opened 2016-08-26
add 'include/libxml2' to $CPATH in libxml2 easyblock

boegel:


boegel:
tested via https://github.com/hpcugent/easybuild-easyconfigs/pull/3479, going in

PR #983, opened 2016-08-31
Added cppcheck easyblock

damianam:
This easyblock enables the installation of cppcheck. The easyconfigs will be PR'ed when this gets accepted, I guess.

### DIFF
--- a/easybuild/easyblocks/c/cppcheck.py
+++ b/easybuild/easyblocks/c/cppcheck.py
@@ -58,7 +58,7 @@ class EB_cppcheck(ConfigureMake):
         Run qmake on the GUI, if necessary
         """
         if self.cfg['build_gui']:
-            cmd = 'qmake QMAKE_CXX="$CXX" QMAKE_LINK=$CXX'
+            cmd = 'qmake QMAKE_CXX="$CXX" QMAKE_LINK="$CXX"'
 
             if self.cfg['have_rules']:
                 cmd = ' '.join([cmd, 'HAVE_RULES=yes'])
--- a/easybuild/easyblocks/c/cppcheck.py
+++ b/easybuild/easyblocks/c/cppcheck.py
@@ -124,8 +124,16 @@ class EB_cppcheck(ConfigureMake):
         """
         Custom sanity check for cppcheck.
         """
-        custom_paths = {
-            'files': ['bin/cppcheck', 'bin/cppcheck-gui' ],
-            'dirs': ['.', 'bin/cfg']
-        }
+
+        if self.cfg['build_gui']:
+            custom_paths = {
+                'files': ['bin/cppcheck', 'bin/cppcheck-gui' ],
+                'dirs': ['.', 'bin/cfg']
+            }
+        else:
+            custom_paths = {
+                'files': ['bin/cppcheck' ],
+                'dirs': ['.']
+            }
+
         super(EB_cppcheck, self).sanity_check_step(custom_paths=custom_paths)
--- a/easybuild/easyblocks/c/cppcheck.py
+++ b/easybuild/easyblocks/c/cppcheck.py
@@ -59,9 +59,9 @@ class EB_cppcheck(ConfigureMake):
         """
         if self.cfg['build_gui']:
             if self.cfg['have_rules']:
-                cmd = 'qmake HAVE_RULES=yes'
+                cmd = 'qmake QMAKE_CXX="$CXX" QMAKE_LINK=$CXX HAVE_RULES=yes'
             else:
-                cmd = 'qmake'
+                cmd = 'qmake QMAKE_CXX="$CXX" QMAKE_LINK=$CXX'
             
             gui_dir = os.path.join(self.cfg['start_dir'], 'gui')
 
--- a/easybuild/easyblocks/c/cppcheck.py
+++ b/easybuild/easyblocks/c/cppcheck.py
@@ -58,10 +58,10 @@ class EB_cppcheck(ConfigureMake):
         Run qmake on the GUI, if necessary
         """
         if self.cfg['build_gui']:
+            cmd = 'qmake QMAKE_CXX="$CXX" QMAKE_LINK=$CXX'
+
             if self.cfg['have_rules']:
-                cmd = 'qmake QMAKE_CXX="$CXX" QMAKE_LINK=$CXX HAVE_RULES=yes'
-            else:
-                cmd = 'qmake QMAKE_CXX="$CXX" QMAKE_LINK=$CXX'
+                cmd = ' '.join([cmd, 'HAVE_RULES=yes'])
             
             gui_dir = os.path.join(self.cfg['start_dir'], 'gui')
 
--- a/easybuild/easyblocks/c/cppcheck.py
+++ b/easybuild/easyblocks/c/cppcheck.py
@@ -63,8 +63,9 @@ class EB_cppcheck(ConfigureMake):
             else:
                 cmd = 'qmake'
             
+            gui_dir = os.path.join(self.cfg['start_dir'], 'gui')
+
             try:
-                gui_dir = os.path.join(self.cfg['start_dir'], 'gui')
                 os.chdir(gui_dir)
                 run_cmd(cmd, log_all=True, simple=True)
                 os.chdir(self.cfg['start_dir'])
@@ -73,7 +74,6 @@ class EB_cppcheck(ConfigureMake):
 
         else:
             self.log.debug("Configuration of the GUI skipped")
-            pass
 
     def build_step(self, verbose=False):
         """
@@ -99,13 +99,11 @@ class EB_cppcheck(ConfigureMake):
        
         # The GUI doesn't have make install, so we copy it manually
         if self.cfg['build_gui']:
+            filepath = os.path.join(self.cfg['start_dir'], 'gui/', 'cppcheck-gui')
+
             try:
                 # make sure we're (still) in the start dir
                 os.chdir(self.cfg['start_dir'])
-
-                file_to_copy = 'cppcheck-gui'
-
-                filepath = os.path.join(self.cfg['start_dir'], 'gui/', file_to_copy)
 
                 # Perform the copy
                 target_dest = os.path.join(self.installdir, 'bin')
@@ -113,9 +111,9 @@ class EB_cppcheck(ConfigureMake):
                     self.log.debug("Copying file %s to %s" % (filepath, target_dest))
                     shutil.copy2(filepath, target_dest)
                 else:
-                    raise EasyBuildError("Can't copy non-existing path %s to %s", filepath, target_dest)
+                    raise EasyBuildError("Can't copy non-existing file %s to %s", filepath, target_dest)
                 
-                # Create cfg symlink. It shouldn't be necessary, but cppcheck-gui complaints otherwise
+                # Create cfg symlink. It shouldn't be necessary, but cppcheck-gui complains otherwise
                 # unless called with --data-dir
                 src = os.path.join(self.installdir, 'cfg')
                 dst = os.path.join(self.installdir, 'bin/cfg')
@@ -123,15 +121,15 @@ class EB_cppcheck(ConfigureMake):
                 os.symlink(src, dst)
 
             except OSError as err:
-                raise EasyBuildError("Copying %s to installation dir failed: %s", file_to_copy, err)
+                raise EasyBuildError("Copying %s to installation dir failed: %s", filepath, err)
 
     def sanity_check_step(self):
         """
         Custom sanity check for cppcheck.
         """
         custom_paths = {
-            'files': ['bin/cppcheck' ],
-            'dirs': ['.']
+            'files': ['bin/cppcheck'],
+            'dirs': []
         }
 
         if self.cfg['build_gui']:
--- a/easybuild/easyblocks/c/cppcheck.py
+++ b/easybuild/easyblocks/c/cppcheck.py
@@ -105,15 +105,15 @@ class EB_cppcheck(ConfigureMake):
 
                 file_to_copy = 'cppcheck-gui'
 
-                filepath = os.path.join(self.cfg['start_dir'], '/gui/', file_to_copy)
+                filepath = os.path.join(self.cfg['start_dir'], 'gui/', file_to_copy)
 
                 # Perform the copy
+                target_dest = os.path.join(self.installdir, 'bin')
                 if os.path.isfile(filepath):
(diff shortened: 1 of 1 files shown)


boegel:
@damianam Please open the easyconfig PR already, and refer to this PR in the description, so we have something to test this with.

easybuild/easyblocks/c/cppcheck.py
```diff
@@ -0,0 +1,139 @@
...
+    """
+    
+    @staticmethod
+    def extra_options(extra_vars=None):
+        """
+        Define if we are using rules or not, and if we are building the GUI
+        """
+        extra = {
+            'have_rules': [False, "Use rules", BUILD],
+            'build_gui': [False, "Build GUI", BUILD],
```
boegel: usually we use `CUSTOM` rather than `BUILD` here

easybuild/easyblocks/c/cppcheck.py
```diff
@@ -0,0 +1,139 @@
...
+        }
+        if extra_vars is None:
+            extra_vars = {}
+        extra.update(extra_vars)
+        return ConfigureMake.extra_options(extra_vars=extra)
+
+
```
boegel: one empty line between methods is enough

easybuild/easyblocks/c/cppcheck.py
```diff
@@ -0,0 +1,139 @@
...
+        Run qmake on the GUI, if necessary
+        """
+        if self.cfg['build_gui']:
+            os.chdir(self.cfg['start_dir']+'/gui')
+            if self.cfg['have_rules']:
+                cmd = 'qmake HAVE_RULES=yes'
+            else:
+                cmd = 'qmake'
+            run_cmd(cmd, log_all=True, simple=True)
+            os.chdir(self.cfg['start_dir'])
```
boegel: please wrap the `os.chdir`s (+ `run_cmd` in between) into a `try-except OSError` block, and determine `cmd` before the `try`

easybuild/easyblocks/c/cppcheck.py
```diff
@@ -0,0 +1,139 @@
...
+        else:
+            pass
```
boegel: at least do a `log.debug` here stating that building of the GUI is being skipped

easybuild/easyblocks/c/cppcheck.py
```diff
@@ -0,0 +1,139 @@
...
+        extra.update(extra_vars)
+        return ConfigureMake.extra_options(extra_vars=extra)
+
+
+    def configure_step(self):
+        """
+        Run qmake on the GUI, if necessary
+        """
+        if self.cfg['build_gui']:
+            os.chdir(self.cfg['start_dir']+'/gui')
```
boegel: use `os.path.join(self.cfg['start_dir'], 'gui')` please

easybuild/easyblocks/c/cppcheck.py
```diff
@@ -0,0 +1,139 @@
...
+        """
+        self.cfg.update('buildopts', 'CFGDIR=%(installdir)s/cfg')
+        
+        if self.cfg['have_rules']:
+            self.cfg.update('buildopts', 'HAVE_RULES=yes')
+
+        super(EB_cppcheck, self).build_step(verbose=verbose)
+        
+        if self.cfg['build_gui']:
+            super(EB_cppcheck, self).build_step(verbose=verbose,path='gui')
```
boegel: style nitpicking: space after `,`

easybuild/easyblocks/c/cppcheck.py
```diff
@@ -0,0 +1,139 @@
...
+        """
+        Run qmake on the GUI, if necessary
+        """
+        if self.cfg['build_gui']:
+            os.chdir(self.cfg['start_dir']+'/gui')
+            if self.cfg['have_rules']:
+                cmd = 'qmake HAVE_RULES=yes'
+            else:
+                cmd = 'qmake'
+            run_cmd(cmd, log_all=True, simple=True)
```
boegel: @damianam just a single `configure` run is enough? the `build_step` seems to suggest you may need to run `configure` twice?

easybuild/easyblocks/c/cppcheck.py
```diff
@@ -0,0 +1,139 @@
...
+                
+                # Create cfg symlink. It shouldn't be necessary, but cppcheck-gui complaints otherwise
+                # unless called with --data-dir
+                src = os.path.join(self.installdir, 'cfg')
+                dst = os.path.join(self.installdir, 'bin/cfg')
+                self.log.debug("Creating symlink %s to %s" % (src, dst))
+                os.symlink(src, dst)
+
+            except OSError, err:
+                raise EasyBuildError("Copying %s to installation dir failed: %s", fil, err)
```
boegel: please use `OSError as err`, and `fil` is not defined?

easybuild/easyblocks/c/cppcheck.py
```diff
@@ -0,0 +1,139 @@
...
+        # The GUI doesn't have make install, so we copy it manually
+        if self.cfg['build_gui']:
+            try:
+                # make sure we're (still) in the start dir
+                os.chdir(self.cfg['start_dir'])
+
+                file_to_copy = 'cppcheck-gui'
+                self.log.debug("Copying cppcheck-gui")
+
+                filepath = self.cfg['start_dir'] + '/gui/' + file_to_copy
```
boegel: please use `os.path.join(self.cfg['start_dir'], 'gui', 'cpp-check-gui')` (there's no need for `file_to_copy` since it's used only once?)

easybuild/easyblocks/c/cppcheck.py
```diff
@@ -0,0 +1,139 @@
...
+        super(EB_cppcheck, self).install_step()
+       
+        # The GUI doesn't have make install, so we copy it manually
+        if self.cfg['build_gui']:
+            try:
+                # make sure we're (still) in the start dir
+                os.chdir(self.cfg['start_dir'])
+
+                file_to_copy = 'cppcheck-gui'
+                self.log.debug("Copying cppcheck-gui")
```
boegel: the `log.debug` below is more informative, just drop this one?

easybuild/easyblocks/c/cppcheck.py
```diff
@@ -0,0 +1,139 @@
...
+        if self.cfg['build_gui']:
+            custom_paths = {
+                'files': ['bin/cppcheck', 'bin/cppcheck-gui' ],
+                'dirs': ['.', 'bin/cfg']
+            }
+        else:
+            custom_paths = {
+                'files': ['bin/cppcheck' ],
+                'dirs': ['.']
+            }
```
boegel: @damianam I would make it more explicit what is different in case of `build_gui` (and don't use `.` as an entry):

```python
custom_paths = {
    'files': ['bin/cppcheck'],
    'dirs': [],
}
if self.cfg['build_gui']:
    custom_paths['files'].append('bin/cppcheck-gui')
    custom_paths['dirs'].append('bin/cfg')
```

easybuild/easyblocks/c/cppcheck.py
```diff
@@ -0,0 +1,139 @@
...
+            extra_vars = {}
+        extra.update(extra_vars)
+        return ConfigureMake.extra_options(extra_vars=extra)
+
+
+    def configure_step(self):
+        """
+        Run qmake on the GUI, if necessary
+        """
+        if self.cfg['build_gui']:
```
boegel: @damianam should we make sure that `Qt` is listed as a dep in this case (to avoid picking up `qmake` from the OS?)

damianam:
The PR with the easyconfig, do you have any preference regarding toolchain? We have it in GCCcore, but maybe you want it somewhere else.

boegel:
@damianam up to you, it can be tweaked to a different toolchain as needed

easybuild/easyblocks/c/cppcheck.py
damianam: The CLI interface doesn't have `configure`, that's why.

easybuild/easyblocks/c/cppcheck.py
damianam: Mmmh, I am not sure with what you mean by `OSError as err`

easybuild/easyblocks/c/cppcheck.py
damianam: That's exactly what I wanted to do yesterday, but my python-fu still is too weak, and it was already late. I figured out that you would review it throughly ;-).

I saw the `.` in `bowtie2.py`, I didn't notice it wasn't really used anywhere else. It looked kind of useless to me.

easybuild/easyblocks/c/cppcheck.py
damianam: From my point of view no. What if somebody wants to use the `Qt` from the OS for whatever reason?

easybuild/easyblocks/c/cppcheck.py
boegel: replace `except OSError, err:` with `except OSError as err:` (which is compatible with both Py3 and Py2)

easybuild/easyblocks/c/cppcheck.py
boegel: hmm, ok, fine then

damianam:
Easyconfig PR: https://github.com/hpcugent/easybuild-easyconfigs/pull/3508

easybuild/easyblocks/c/cppcheck.py
```diff
@@ -0,0 +1,141 @@
...
+            if self.cfg['have_rules']:
+                cmd = 'qmake HAVE_RULES=yes'
+            else:
+                cmd = 'qmake'
+            
+            try:
+                gui_dir = os.path.join(self.cfg['start_dir'], 'gui')
```
boegel: move this line outside of the `try` to, so it's definitely defined in the `except` scope

easybuild/easyblocks/c/cppcheck.py
```diff
@@ -0,0 +1,141 @@
...
+                os.chdir(gui_dir)
+                run_cmd(cmd, log_all=True, simple=True)
+                os.chdir(self.cfg['start_dir'])
+            except OSError as err:
+                raise EasyBuildError("Moving to %s and configure the GUI build failed: %s", gui_dir, err)
+
+        else:
+            self.log.debug("Configuration of the GUI skipped")
+            pass
```
boegel: drop this, no need for a `pass` is you have other statements already ther

easybuild/easyblocks/c/cppcheck.py
```diff
@@ -0,0 +1,141 @@
...
+            except OSError as err:
+                raise EasyBuildError("Copying %s to installation dir failed: %s", file_to_copy, err)
+
+    def sanity_check_step(self):
+        """
+        Custom sanity check for cppcheck.
+        """
+        custom_paths = {
+            'files': ['bin/cppcheck' ],
+            'dirs': ['.']
```
boegel: please drop the `'.'`, the current directory will always be there and non-empty, so this is useless to check

easybuild/easyblocks/c/cppcheck.py
```diff
@@ -0,0 +1,141 @@
...
+
+            except OSError as err:
+                raise EasyBuildError("Copying %s to installation dir failed: %s", file_to_copy, err)
+
+    def sanity_check_step(self):
+        """
+        Custom sanity check for cppcheck.
+        """
+        custom_paths = {
+            'files': ['bin/cppcheck' ],
```
boegel: style nitpicking: drop the space before `]` ;)

easybuild/easyblocks/c/cppcheck.py
```diff
@@ -0,0 +1,141 @@
...
+                
+                # Create cfg symlink. It shouldn't be necessary, but cppcheck-gui complaints otherwise
+                # unless called with --data-dir
+                src = os.path.join(self.installdir, 'cfg')
+                dst = os.path.join(self.installdir, 'bin/cfg')
+                self.log.debug("Creating symlink %s to %s" % (src, dst))
+                os.symlink(src, dst)
+
+            except OSError as err:
+                raise EasyBuildError("Copying %s to installation dir failed: %s", file_to_copy, err)
```
boegel: drop `file_to_copy`, use `filepath` here instead (and define `filepath` outside of the `try`

easybuild/easyblocks/c/cppcheck.py
```diff
@@ -0,0 +1,141 @@
...
+
+                filepath = os.path.join(self.cfg['start_dir'], 'gui/', file_to_copy)
+
+                # Perform the copy
+                target_dest = os.path.join(self.installdir, 'bin')
+                if os.path.isfile(filepath):
+                    self.log.debug("Copying file %s to %s" % (filepath, target_dest))
+                    shutil.copy2(filepath, target_dest)
+                else:
+                    raise EasyBuildError("Can't copy non-existing path %s to %s", filepath, target_dest)
```
boegel: `s/path/file/g` (what if cppcheck-gui` is a directory :))

easybuild/easyblocks/c/cppcheck.py
```diff
@@ -0,0 +1,141 @@
...
+                
+                # Create cfg symlink. It shouldn't be necessary, but cppcheck-gui complaints otherwise
```
boegel: `s/complaints/complains/`

easybuild/easyblocks/c/cppcheck.py
```diff
@@ -0,0 +1,139 @@
...
+            extra_vars = {}
+        extra.update(extra_vars)
+        return ConfigureMake.extra_options(extra_vars=extra)
+
+    def configure_step(self):
+        """
+        Run qmake on the GUI, if necessary
+        """
+        if self.cfg['build_gui']:
+            cmd = 'qmake QMAKE_CXX="$CXX" QMAKE_LINK=$CXX'
```
boegel: @damianam better wrap the 2nd `$CXX` in quotes too, in case its value contains spaces?

easybuild/easyblocks/c/cppcheck.py
```diff
@@ -0,0 +1,139 @@
...
+
+            if self.cfg['have_rules']:
+                cmd = ' '.join([cmd, 'HAVE_RULES=yes'])
```
boegel: The below is good enough here, but I won't nitpick over that ;)

```python
cmd += ' HAVE_RULES=yes'
```

boegel:
tested with https://github.com/hpcugent/easybuild-easyconfigs/pull/3508, good to go, thanks @damianam!

PR #984, opened 2016-09-01
enhance & clean up libxml2 easyblock

boegel:
with this applied, we can drop the `easyblock = 'ConfigureMake'` from libxml2 easyconfigs that do not include Python bindings, and simplifis them quite a bit

### DIFF
--- a/easybuild/easyblocks/l/libxml2.py
+++ b/easybuild/easyblocks/l/libxml2.py
@@ -80,10 +80,8 @@ class EB_libxml2(ConfigureMake, PythonPackage):
             if not get_software_root('Python'):
                 raise EasyBuildError("Python module not loaded")
 
-            # we will do the python bindings ourselves so force them off
-            # this is required to ensure that the libxml2 Python bindings are installed in the right location,
-            # i.e. in the libxml2 install prefix rather than in the Python install prefix
-            self.cfg.update('configopts', '--without-python')
+            libxml2_pylibdir = os.path.join(self.installdir, self.pylibdir)
+            self.cfg.update('configopts', "--with-python --with-python-install-dir=%s" % libxml2_pylibdir)
         else:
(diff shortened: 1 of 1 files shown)


boegel:
@wpoely86 please review?

easybuild/easyblocks/l/libxml2.py
```diff
@@ -50,72 +52,73 @@ def extra_options(extra_vars=None):
...
+
+        if self.toolchain.options.get('pic', False):
+            self.cfg.update('configopts', '--with-pic')
+
+        zlib = get_software_root('zlib')
+        if zlib:
+            self.cfg.update('configopts', '--with-zlib=%s' % zlib)
+
+        if self.with_python_bindings:
+            if not get_software_root('Python'):
```
wpoely86: why do this twice?

easybuild/easyblocks/l/libxml2.py
```diff
@@ -50,72 +52,73 @@ def extra_options(extra_vars=None):
...
+                raise EasyBuildError("Python module not loaded")
+
+            libxml2_pylibdir = os.path.join(self.installdir, self.pylibdir)
+            self.cfg.update('configopts', "--with-python --with-python-install-dir=%s" % libxml2_pylibdir)
```
wpoely86: pass the python binary to `---with-python`

easybuild/easyblocks/l/libxml2.py
boegel: good point, thanks

boegel:
@wpoely86 please rereview?

wpoely86:
:+1:

boegel:
tested with all existing easyconfigs via https://github.com/hpcugent/easybuild-easyconfigs/pull/3509, good to go, thanks for the review @wpoely86!

PR #986, opened 2016-09-05
Small sanity check change to openfoam block

wpoely86:
In openfoam 4.0, the `vorticity` binary is superseded by postproces. I've replaced it by `wdot` which seems to be present in all releases so far.

### DIFF
--- a/easybuild/easyblocks/o/openfoam.py
+++ b/easybuild/easyblocks/o/openfoam.py
@@ -87,7 +87,8 @@ class EB_OpenFOAM(EasyBlock):
             try:
                 contents_installdir = os.listdir(self.installdir)
                 # it's one directory but has a wrong name
-                if len(contents_installdir) == 1 and os.path.isdir(os.path.join(self.installdir, contents_installdir[0])):
+                if len(contents_installdir) == 1 and \
+                   os.path.isdir(os.path.join(self.installdir, contents_installdir[0])):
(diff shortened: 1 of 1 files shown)


wpoely86:
@hajgato FYI

easybuild/easyblocks/o/openfoam.py
```diff
@@ -87,7 +87,8 @@ def extract_step(self):
             try:
                 contents_installdir = os.listdir(self.installdir)
                 # it's one directory but has a wrong name
-                if len(contents_installdir) == 1 and os.path.isdir(os.path.join(self.installdir, contents_installdir[0])):
+                if len(contents_installdir) == 1 and \
+                   os.path.isdir(os.path.join(self.installdir, contents_installdir[0])):
                     source = os.path.join(self.installdir, contents_installdir[0])
```
boegel: @wpoely86 better move `source` outside of the `if`, I don't like this use of `\`

wpoely86: if that rocks your boat but there's nothing wrong with backslashes

boegel: we rarely use them elsewhere in the EB codebase, just a matter of consistency

boegel:
looks ok & tested with a couple of existing OpenFOAM easyconfigs, so good to go, thanks @wpoely86!

PR #987, opened 2016-09-06
Updated neuron easyblock to support the lack of hoc_ed in 7.4, and to…

damianam:
… allow the bypassing of the tests. The reason to allow the bypassing is that `mpi_cmd_for` is broken and makes it unnecessarily difficult to perform tests with anything else than `mpirun`. In our setup, we use `srun` for all the toolchains, so `mpi_cmd_for` doesn't work.


easybuild/easyblocks/n/neuron.py
```diff
@@ -57,6 +59,7 @@ def extra_options():
 
         extra_vars = {
             'paranrn': [True, "Enable support for distributed simulations.", CUSTOM],
+            'enable_tests': [True, "Enable extensive testing in the sanity check step.", CUSTOM],
```
boegel: @damianam why not use the existing `runtest` easyconfig parameter?

damianam: @boegel for 2 reasons: 1) I forgot about it; 2) This is normally used to control the `test_step` method. A good followup question is: why are these tests run in the sanity check instead of in the test step? Maybe we should move them out?

boegel: @damianam we could move this to the `test_step` method, but then we need to do some extra effort to make sure that the environment is such that the NEURON that was built can be used; in the `sanity_check_step` we simply rely on (a temporary version of) the generated module to handle that for us

So, the test being done in the sanity check step somewhat validates the installation + the module that will be generated...

boegel:
@damianam this needs a sync with `develop` because of #993

damianam:
@boegel done

damianam:
@boegel, what about this for 3.1.0?

boegel:
Looks good, tested with https://github.com/hpcugent/easybuild-easyconfigs/pull/4080, good to go.

Thanks @damianam!

boegel:
Oops, didn't mean to close this... :)

PR #993, opened 2016-09-17
make easyblocks that run MPI tests aware of 'mpi_tests' build option

boegel:
depends on ~~https://github.com/hpcugent/easybuild-framework/pull/1920~~


boegel:

Testing this thoroughly revealed a missing `import` in the WRF easyblock, but other than that, works fine, so going in.

Issue #776, opened 2015-12-14
using --module-only for icc results in $UNKNOWN being defined to 'UNKNOWN'


boegel:
```
-bash-4.1$ module show icc/2015.3.187-GNU-4.9.3-2.25 
-------------------------------------------------------------------
/user/scratch/gent/vsc400/vsc40023/easybuild_REGTEST/SL6/sandybridge/modules/all/icc/2015.3.187-GNU-4.9.3-2.25:

module-whatis	 C and C++ compiler from Intel - Homepage: http://software.intel.com/en-us/intel-compilers/ 
conflict	 icc 
module		 load GNU/4.9.3-2.25 
prepend-path	 CPATH /user/scratch/gent/vsc400/vsc40023/easybuild_REGTEST/SL6/sandybridge/software/icc/2015.3.187-GNU-4.9.3-2.25/ipp/include 
prepend-path	 CPATH /user/scratch/gent/vsc400/vsc40023/easybuild_REGTEST/SL6/sandybridge/software/icc/2015.3.187-GNU-4.9.3-2.25/mkl/include 
prepend-path	 CPATH /user/scratch/gent/vsc400/vsc40023/easybuild_REGTEST/SL6/sandybridge/software/icc/2015.3.187-GNU-4.9.3-2.25/tbb/include 
prepend-path	 IPPROOT /user/scratch/gent/vsc400/vsc40023/easybuild_REGTEST/SL6/sandybridge/software/icc/2015.3.187-GNU-4.9.3-2.25/ipp 
prepend-path	 LD_LIBRARY_PATH /user/scratch/gent/vsc400/vsc40023/easybuild_REGTEST/SL6/sandybridge/software/icc/2015.3.187-GNU-4.9.3-2.25/lib 
prepend-path	 LD_LIBRARY_PATH /user/scratch/gent/vsc400/vsc40023/easybuild_REGTEST/SL6/sandybridge/software/icc/2015.3.187-GNU-4.9.3-2.25/ipp/lib/intel64 
prepend-path	 LD_LIBRARY_PATH /user/scratch/gent/vsc400/vsc40023/easybuild_REGTEST/SL6/sandybridge/software/icc/2015.3.187-GNU-4.9.3-2.25/mkl/lib/intel64 
[...]

PR #994, opened 2016-09-18
Move setting up intel license to prepare_step (for --modules-only)

wpoely86:
If you run with --modules-only, the prepare_step gets run so we need to
set up the license here. Else creating the module will fail with
incorrect or no license at all.

### DIFF
--- a/easybuild/easyblocks/generic/intelbase.py
+++ b/easybuild/easyblocks/generic/intelbase.py
@@ -201,12 +201,9 @@ class IntelBase(EasyBlock):
         except OSError, err:
             raise EasyBuildError("Failed to symlink %s to %s: %s", self.home_subdir_local, self.home_subdir, err)
 
-    def configure_step(self):
-        """Configure: handle license file and clean home dir."""
-
-        # prepare (local) 'intel' home subdir
-        self.setup_local_home_subdir()
-        self.clean_home_subdir()
+    def prepare_step(self):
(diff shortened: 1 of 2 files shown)


wpoely86:
It's a bit ironic that the `--module-only` tests fail while I try to fix them :laughing:

boegel:
@wpoely86 can you clarify whether this is this a fix for #776?

wpoely86:
@boegel yes, that's the idea. But actually, it should be picked up from the existing license...

boegel:
@wpoely86 only if `$INTEL_LICENSE_FILE` is defined before the module is loaded?

wpoely86:
well yes, if it's set in the tcl module, it should be the same value in the lua module (and vice versa)

boegel:
@wpoely86 the tests fail for a good reason, there's no valid license file available

This is fixed with https://github.com/wpoely86/easybuild-easyblocks/pull/27

I disagree that we should pick up the `$INTEL_LICENSE_FILE` definition from an existing module, since the whole reason for using `--module-only` may be to *correct* the value of `$INTEL_LICENSE_FILE` that is currently there...

We can't make any assumptions there imho, it's up to the person using `--module-only` on the Intel tools to make sure the license file being picked up is correct, EasyBuild has no way of knowing what the correct setting is here.

boegel:
tested and approved, indeed fixes #776 , thanks @wpoely86!

PR #996, opened 2016-09-20
Copy CP2K provided basis set directory to installdir, and use it

hajgato:


easybuild/easyblocks/c/cp2k.py
```diff
@@ -781,6 +782,19 @@ def install_step(self):
...
+        targetdir = os.path.join(self.installdir, 'data')
+        if os.path.exists(targetdir):
+            self.log.info("Won't copy data dir. Destination directory %s already exists" % targetdir)
+        elif not os.path.exists(datadir):
+            self.log.info("Won't copy data dir. Source directory %s does not exist" % datadir)
+        else:
+            try:
+                shutil.copytree(datadir, targetdir)
+            except:
+                raise EasyBuildError("Copying tests from %s to %s failed", srctests, targetdir)
```
boegel: @hajgato positive logic is easier to follow, `srctests` doesn't exist, please use `except OSError` and include actual error in error message, so:

```python
if os.path.exists(targetdir):
    self.log.info("Won't copy data dir. Destination directory %s already exists", targetdir)
elif os.path.exists(datadir):
    try:
        shutil.copytree(datadir, targetdir)
    except OSError as err:
        raise EasyBuildError("Copying tests from %s to %s failed: %s", datadir, targetdir, err)
else:
    self.log.info("Won't copy data dir. Source directory %s does not exist", datadir)
```

easybuild/easyblocks/c/cp2k.py
```diff
@@ -816,3 +830,12 @@ def sanity_check_step(self):
...
 
         super(EB_CP2K, self).sanity_check_step(custom_paths=custom_paths)
+
+    def make_module_extra(self):
+        """Set up a CP2K_DATA_DIR environment variable to find CP2K provided basis sets"""
+
+        txt = super(EB_CP2K, self).make_module_extra()
+        if os.path.exists(os.path.join(self.installdir, 'data')):
+            txt += self.module_generator.set_environment('CP2K_DATA_DIR',
+                                                          os.path.join(self.installdir, 'data'))
```
boegel: @hajgato if you need a value multiple times, it's cleaner to just construct it once:

```python
data_dir = os.path.join(self.installdir, 'data')
if os.path.exists(data_dir):
    txt += self.module_generator.set_environment('CP2K_DATA_DIR', data_dir)
````

File: easybuild/easyblocks/c/cp2k.py (hunk 1 above)
Comment (hajgato) [inline easybuild/easyblocks/c/cp2k.py]: done

File: easybuild/easyblocks/c/cp2k.py (hunk 2 above)
Comment (hajgato) [inline easybuild/easyblocks/c/cp2k.py]: done

File: easybuild/easyblocks/c/cp2k.py [hunk 3]
```diff
@@ -781,6 +782,19 @@ def install_step(self):
...
+        # copy data dir
+        datadir = os.path.join(self.cfg['start_dir'], 'data')
+        targetdir = os.path.join(self.installdir, 'data')
+        if os.path.exists(targetdir):
+            self.log.info("Won't copy data dir. Destination directory %s already exists" % targetdir)
+        elif os.path.exists(datadir):
+            try:
+                shutil.copytree(datadir, targetdir)
+            except:
+                raise EasyBuildError("Copying data dir from %s to %s failed", srctests, targetdir)
```
hajgato: @boegel It comes from copy paste ;)
Anyway, I corrected it

boegel:
merged via #1084, thanks @hajgato!

PR #1002, opened 2016-09-28
Enable GCCcore as toolchain for clang, add logging, and fail if no proper prefix is found

damianam:
This PR implements the following changes:

-Enables the use of `GCCcore` as base GCC for clang
-Adds logging to show which path was used as `GCC_INSTALL_PREFIX`
-If no correct path is found, it fails instead of keep going with `-DGCC_INSTALL_PREFIX='None'`

### DIFF
--- a/easybuild/easyblocks/c/clang.py
+++ b/easybuild/easyblocks/c/clang.py
@@ -198,7 +198,19 @@ class EB_Clang(CMakeMake):
 
         # GCC and Clang are installed in different prefixes and Clang will not
         # find the GCC installation on its own.
-        self.cfg['configopts'] += "-DGCC_INSTALL_PREFIX='%s' " % get_software_root('GCC')
+        # First try with GCC
+        gcc_prefix = get_software_root('GCC')
+
+        # If that doesn't work, try with GCCcore
+        if gcc_prefix == None:
+            gcc_prefix = get_software_root('GCCcore')
+        
+        # If that doesn't work either, print error and exit
+        if gcc_prefix == None:
+            raise EasyBuildError("Can't find GCC prefix")
+
+        self.cfg['configopts'] += "-DGCC_INSTALL_PREFIX='%s' " % gcc_prefix
+        self.log.debug("Using %s as GCC_INSTALL_PREFIX" % gcc_prefix)
 
         self.cfg['configopts'] += "-DCMAKE_BUILD_TYPE=Release "
         if self.cfg['assertions']:
--- a/easybuild/easyblocks/c/clang.py
+++ b/easybuild/easyblocks/c/clang.py
@@ -207,7 +207,7 @@ class EB_Clang(CMakeMake):
             gcc_prefix = get_software_root('GCC')
         
         # If that doesn't work either, print error and exit
-        if gcc_prefix == None:
+        if gcc_prefix is None:
             raise EasyBuildError("Can't find GCC or GCCcore to use")
 
         self.cfg.update('configopts', "-DGCC_INSTALL_PREFIX='%s' " % gcc_prefix)
--- a/easybuild/easyblocks/c/clang.py
+++ b/easybuild/easyblocks/c/clang.py
@@ -198,16 +198,17 @@ class EB_Clang(CMakeMake):
 
         # GCC and Clang are installed in different prefixes and Clang will not
         # find the GCC installation on its own.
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/c/clang.py
```diff
@@ -198,7 +198,19 @@ def configure_step(self):
...
+        # First try with GCC
+        gcc_prefix = get_software_root('GCC')
+
+        # If that doesn't work, try with GCCcore
+        if gcc_prefix == None:
+            gcc_prefix = get_software_root('GCCcore')
+        
+        # If that doesn't work either, print error and exit
+        if gcc_prefix == None:
+            raise EasyBuildError("Can't find GCC prefix")
```
wpoely86: make this more explicit: `Cannot find GCC or GCCcore to use`

damianam: Done

easybuild/easyblocks/c/clang.py
```diff
@@ -198,7 +198,20 @@ def configure_step(self):
 
         # GCC and Clang are installed in different prefixes and Clang will not
         # find the GCC installation on its own.
-        self.cfg['configopts'] += "-DGCC_INSTALL_PREFIX='%s' " % get_software_root('GCC')
+        # First try with GCCcore, as GCC built on top of GCCcore is just a wrapper for GCCcore and binutils,
+        # instead of a full-fledge compiler
+        gcc_prefix = get_software_root('GCCcore')
+
+        # If that doesn't work, try with GCC
+        if gcc_prefix == None:
```
boegel: @damianam please use `if gcc_prefix is None` (style nitpicking); same below

easybuild/easyblocks/c/clang.py
```diff
@@ -198,7 +198,20 @@ def configure_step(self):
...
+            gcc_prefix = get_software_root('GCC')
+        
+        # If that doesn't work either, print error and exit
+        if gcc_prefix == None:
+            raise EasyBuildError("Can't find GCC or GCCcore to use")
+
+        self.cfg['configopts'] += "-DGCC_INSTALL_PREFIX='%s' " % gcc_prefix
```
boegel: @damianam please update to use `self.cfg.update`:

```python
self.cfg.update('configopts', "-DGCC_INSTALL_PREFIX='%s'" % gcc_prefix)
```

easybuild/easyblocks/c/clang.py
```diff
@@ -198,7 +198,20 @@ def configure_step(self):
...
+        self.log.debug("Using %s as GCC_INSTALL_PREFIX" % gcc_prefix)
```
boegel: @damianam more style nitpicking: don't use `%` in `self.log.debug`, use `,`:

```python
self.log.debug("Using %s as GCC_INSTALL_PREFIX",  gcc_prefix)
```

(this results in only doing the actual templating when `self.log.debug` is called)

damianam: Done, like the other comments

boegel:
This looks good, it's good to go in, but I wonder what made you make this change...

The `GCC` bundle of `GCCcore` and `binutils` usually includes `altroot = 'GCCcore'` (cfr. https://github.com/hpcugent/easybuild-easyconfigs/blob/master/easybuild/easyconfigs/g/GCC/GCC-6.2.0-2.27.eb#L21), which tells EasyBuild to define `$EBROOTGCC` to the same value as `$EBROOTGCCCORE`, so in that case this doesn't make any difference at all?

damianam:
You are right. It just seemed a bit more robust to do this double check, but with the current setup it is pointless.

boegel:
@damianam so, what shall we do? Merge it, or leave it as is? Do you think this could be worthwhile in the end?

damianam:
I think that leaving it as it is won't cause any harm and have some potential (but remote) benefit. But if you want to keep things cleaner and more focused on current issues, then we can clean that bit and remove the `GCC` check. It is more of a policy than a technical question IMO.

I would check for `GCCcore` instead of `GCC` though, as the original easyblock don't allow `clang` to be installed there.

boegel:
@damianam ok, your last point actually explains why you've made this change: so you can install `Clang` on top of `GCCcore` instead of on top of the `GCC` bundle, I missed that part...

So, going in!

PR #1003, opened 2016-09-30
fix ATLAS easyblock for non-x86 systems

boegel:
patch from https://github.com/openhpc/ohpc/pull/312 by @ericvh


boegel:
@ericvh: if you need any help with EasyBuild on `aarch64`, let us know

I know that @geimer has been playing around with it as well; there are some upcoming minor tweaks to the EasyBuild framework that will improve EasyBuild support for `aarch64` as well...

boegel:

PR #1004, opened 2016-09-30
Added jhbuild easyblock

damianam:
This PR adds a jhbuild easyblock. Useful for installing a variety of packages (X.Org, GTK, GIMP, etc). A test PR in https://github.com/hpcugent/easybuild-easyconfigs will follow (next week). This is a first step to clean https://github.com/hpcugent/easybuild-easyconfigs/pull/3463

Test easyconfig PR: https://github.com/hpcugent/easybuild-easyconfigs/pull/3641

### DIFF
--- a/easybuild/easyblocks/j/jhbuild.py
+++ b/easybuild/easyblocks/j/jhbuild.py
@@ -1,0 +1,128 @@
+##
+# Copyright 2016 Forschungszentrum Juelich GmbH
+#
+# This file is part of EasyBuild,
+# originally created by the HPC team of Ghent University (http://ugent.be/hpc/en),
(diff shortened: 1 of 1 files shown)


damianam:
@boegel I guess this one won't make it to 3.1.0, right?

damianam:
@boegel, what should we do with this? It is not that useful for us anymore since we use your `X11`, but throwing it away is kind of a waste.

boegel:
@damianam Well, if it's not going to be used we should close the PR?

If we ever need to get back to it, we can revisit this PR...

damianam:
Sounds reasonable. I am closing it.

PR #1005, opened 2016-10-05
Add parallel build to Boost

dolfim:


akesandgren:
Nice. Made it ALOT faster :-)

boegel:
tested with all existing easyconfigs for `Boost`, works like a charm for all, so going in

Thanks a lot @dolfim!

boegel:
@dolfim just something for future contributions: I had to change the target branch to `develop` (instead of `master`), since we never merge PRs directly into `master` (other than when we release a new version of EasyBuild)

PR #1006, opened 2016-10-06
Fix g2clib easyblock to allow to install 1.6.0 or higher

damianam:
Self descriptive title


boegel:
lgtm, tested, going in, thanks @damianam!

PR #1008, opened 2016-10-07
add support to Scons generic easyblock to provide argument to specify installation prefix

boegel:
Apparently `PREFIX=` is not always the correct way to specify the installation prefix, sometimes it's `prefix=` (or maybe something else)...


boegel:

PR #1010, opened 2016-10-11
Modified scipy easyblock to allow installing the latest version

damianam:
Otherwise the sanity check fails


boegel:
merged via #1037, thanks @damianam!

PR #1013, opened 2016-10-13
Fixing usempi and with_mpi usage to allow for a serial build of Amber 16

edmondac:
The previous logic mixed up the existence of mpiroot and the toolchain usempi option. Now both must be specified for with_mpi to be True. And with_mpi is used for all checks.


boegel:
looks perfect, verified via for backward compatibility via `eb Amber-14-intel-2016a-AmberTools-15-patchlevel-13-13.eb -x`

Thanks @edmondac!

PR #1014, opened 2016-10-14
leverage toolchain.linalg in ScaLAPACK easyblock, use copy_file

boegel:
required to get rid of implementation that was deprecated for EasyBuild v3.0

this should not affect existing easyconfigs, it's just implementation details really...

### DIFF
--- a/easybuild/easyblocks/s/scalapack.py
+++ b/easybuild/easyblocks/s/scalapack.py
@@ -39,6 +39,7 @@ from distutils.version import LooseVersion
 import easybuild.tools.toolchain as toolchain
 from easybuild.easyblocks.blacs import det_interface  #@UnresolvedImport
 from easybuild.easyblocks.generic.configuremake import ConfigureMake
+from easybuild.toolchains.linalg.acml import Acml
 from easybuild.toolchains.linalg.atlas import Atlas
 from easybuild.toolchains.linalg.gotoblas import GotoBLAS
 from easybuild.toolchains.linalg.lapack import Lapack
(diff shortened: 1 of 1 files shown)


akesandgren:
Actually MKL should probably be in the dep_grp for scalapack too
There are situations where one have to use the netlib scalapack and mkl together....
there be bugs in mkl scalapack in some cases (at least there used to be)

boegel:
@akesandgren I suggest we add that in later as needed? I have yet to see a ScaLAPACK easyconfig that sit on top of `imkl`, let alone a toolchain that leverages it...

boegel:
@wpoely86 please review?

easybuild/easyblocks/s/scalapack.py
```diff
@@ -56,33 +59,28 @@ def configure_step(self):
...
+        else:
+            copy_file(src, dest)
 
         self.loosever = LooseVersion(self.version)
 
-        # make sure required dependencies are available
-        deps = [("LAPACK", "ACML", "OpenBLAS")]
-        self.log.deprecated("EB_ScaLAPACK.configure_step uses hardcoded list of LAPACK libs", '3.0')
+        # make sure required dependencies are available for BLAS/LAPACK
+        dep_grps = [('ACML', 'LAPACK', 'OpenBLAS')]
```
wpoely86: didn't you add constants for these?

boegel: these are the software names, but I guess I can use `Acml.BLAS_MODULE_NAME[0]` & co here, like I do in #1016?

boegel:
@wpoely86: please rereview?

boegel:
tested on all existing ScaLAPACK easyconfigs, good to go, thanks for the review @wpoely86!

PR #1016, opened 2016-10-14
remove deprecated 'get_blas_lib' function from LAPACK easyblock + use copy_file

boegel:
see also https://github.com/hpcugent/easybuild/pull/260

edit: requires #1014, where dependency on `get_blas_lib` in ScaLAPACK easyblock is removed


boegel:
@wpoely86 please review

boegel:
Tested with various existing LAPACK easyconfigs, good to go (after fixing a stupid bug that @wpoely86 didn't notice ;)). Thanks for the review!

PR #1019, opened 2016-10-20
Allow netCDF-C++4 to be used with ESMF

damianam:


boegel:
lgtm, testing checks out, thanks @damianam!

PR #1024, opened 2016-10-27
update CBLAS easyblock to build with foss toolchain

boegel:


boegel:

PR #1025, opened 2016-11-02
Make the allinea easyblock search for the templates in the easyconfig paths

akesandgren:
Also add support for specifying a system.config to install.

This makes it possible to keep the templates in the sites easyconfig repos and also handles installing a system.config file if requested.


easybuild/easyblocks/a/allinea.py
```diff
@@ -81,11 +82,34 @@ def install_step(self):
...
+        # copy system.config if requested
+	sysconf_path = os.path.join(self.installdir, 'system.config')
+	if self.cfg['sysconfig'] is not None:
+	    path = self.obtain_file(self.cfg['sysconfig'], extension=False)
+	    if path:
+		self.log.debug('system.config file %s found' % path)
+	    else:
+		raise EasyBuildError('No system.config file named %s found', sysconfig)
+            try:
+                shutil.copy(path, sysconf_path)
```
boegel: @akesandgren please use `copy_file` (from `filetools` module) instead

easybuild/easyblocks/a/allinea.py
```diff
@@ -81,11 +82,34 @@ def install_step(self):
...
+            except OSError, err:
+                raise EasyBuildError("Failed to copy %s to %s: %s", path, sysconf_path, err)
+	    try:
+		os.chmod(sysconf_path, 0444)
```
boegel: @akesandgren please use `adjust_permissions` function for this?

easybuild/easyblocks/a/allinea.py
```diff
@@ -81,11 +82,34 @@ def install_step(self):
...
         for templ in self.cfg['templates']:
+	    path = self.obtain_file(templ, extension='qtf')
+	    if path:
+		self.log.debug('Template file %s found' % path)
+	    else:
+		raise EasyBuildError('No template file named %s found', templ)
+
             try:
-                shutil.copy2(templ, templ_path)
+                shutil.copy(path, templ_path)
```
boegel: @akesandgren why this change?

akesandgren: If you mean the change from copy2 to copy, the installed file should have the same permissions as the rest of the installation, not whatever the template file had before the copy.

akesandgren:
Anything else?

easybuild/easyblocks/a/allinea.py
boegel: ok, thanks for the clarification

easybuild/easyblocks/a/allinea.py
```diff
@@ -81,11 +84,30 @@ def install_step(self):
...
             except OSError, err:
                 raise EasyBuildError("Failed to copy template %s to %s: %s", templ, templ_path, err)
 
+        # copy system.config if requested
+	sysconf_path = os.path.join(self.installdir, 'system.config')
```
boegel: 🚨 tabs detected! 🚨

fixed in https://github.com/akesandgren/easybuild-easyblocks/pull/2

boegel:
tested with existing Allinea easyconfigs, thanks @akesandgren!

PR #1026, opened 2016-11-08
only RPATH in /usr/lib* directories when building bintuils with dummy toolchain

boegel:
fix for issue raised in https://github.com/hpcugent/easybuild-easyconfigs/issues/3409 by @akesandgren 

I seem to be also being bitten by this while testing out the RPATH support being implemented in https://github.com/hpcugent/easybuild-framework/pull/1942

cc @wpoely86


boegel:
Haven't seen any issues pop up with this in place, so going in, thanks for the review @wpoely86!

PR #1027, opened 2016-11-08
Make hdf5 easyblock handle filter-deps correctly.

akesandgren:
If zlib or szip is filtered out then just activate them and let
configure find them from the OS install.

### DIFF
--- a/easybuild/easyblocks/h/hdf5.py
+++ b/easybuild/easyblocks/h/hdf5.py
@@ -39,6 +39,7 @@ from easybuild.easyblocks.generic.configuremake import ConfigureMake
 from easybuild.tools.build_log import EasyBuildError
 from easybuild.tools.modules import get_software_root
 from easybuild.tools.systemtools import get_shared_lib_ext
+from easybuild.tools.config import build_option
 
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/h/hdf5.py
```diff
@@ -51,12 +52,16 @@ def configure_step(self):
...
         for (dep, opt) in deps:
             root = get_software_root(dep)
             if root:
                 self.cfg.update('configopts', '%s=%s' % (opt, root))
             else:
-                raise EasyBuildError("Dependency module %s not loaded.", dep)
+                if dep in filter_deps:
+                    self.cfg.update('configopts', '%s' % opt)
+                else:
+                    raise EasyBuildError("Dependency module %s not loaded.", dep)
```
akesandgren: Yeah, that was the other option. I just wasn't sure if whoever wrote that easyblock had any real reason for doing it that way.

boegel: I guess it's like this since this was written long before `--filter-deps`...

Let's go with the non-error option instead then?

boegel:
Thanks @akesandgren!

PR #1028, opened 2016-11-09
update Gurobi easyblock to use copy_file

boegel:
@bedroge please review (since you contributed this easyblock originally)

In terms of functionality, nothing has changed.


boegel:
Thanks for the review @bedroge!

Tested with https://github.com/hpcugent/easybuild-easyconfigs/pull/3769 and the existing Gurobi easyconfig, so good to go.

Issue #969, opened 2016-07-08
GCC: Libiberty 


stachon:
Hello,
can we add Libiberty to GCC build? (adding --enable-install-libiberty to configure) ? In versions 4.8.3 and 4.7.4 it is present, it seems that it is no longer installed by default since GCC 4.8.x and binutils-2.24. Or perhaps this should go to binutils? (it seems that either GCC, binutils and GDB can provide it).

[1] https://gcc.gnu.org/onlinedocs/libiberty/Using.html#Using

boegel:
@stachon I see no reason not to do it...

We have an easyblock for both `binutils` and `GCC`, and since we're using our own `binutils` in recent toolchains, that seems a good place to add it?

@wpoely86 thoughts?

wpoely86:
Fine by me but wasn't there a reason we don't build it? 

But doesn't hurt to try it again.

ocaisa:
See https://github.com/hpcugent/easybuild-easyconfigs/pull/1885#issuecomment-133078819 and the few comments after for why we didn't include libiberty in the past

boegel:
enabling the installation of `libiberty` (by default) is done in #1030

boegel:
It seems like this was our reason for not enabling the installation of `libiberty` by default (error I got when building `binutils` 2.25 with `GCC` 4.9.3 that has `libiberty` included:

```
/gpfs/scratch/projects/project_gpilot/vsc40023/eb_rpath/software/binutils/2.25/bin/ld.gold: error: /gpfs/scratch/projects/project_gpilot/vsc40023/eb_rpath/software/GCCcore/4.9.3/lib64/libiberty.a(cplus-dem.o): requires unsupported dynamic reloc 11; recompile with -fPIC
```

boegel:
#1030 (which will be included for EasyBuild v3.0) now enables the installation if `libiberty` by default through the `binutils` easyblock, so I'm considering this fixed

Thanks for reporting this @stachon!

Issue #921, opened 2016-05-16
GCC: /lib64 and /usr/lib64 comes before $root/lib when linking


sebth:
When GCC searches for libraries during linking, system `/lib64` and `/usr/lib64` comes before `/lib` of a loaded module. This can cause GCC to link programs against system libraries instead of the libraries of a loaded module.

This is the output of `gcc -print-search-dirs`, which shows the incorrect search order, when GCC-4.8.4 has been installed and loaded (with EasyBuild):

```
install: /home/sebth/eb/software/GCC/4.8.4/lib/gcc/x86_64-unknown-linux-gnu/4.8.4/
programs: =/home/sebth/eb/software/GCC/4.8.4/libexec/gcc/x86_64-unknown-linux-gnu/4.8.4/:/home/sebth/eb/software/GCC/4.8.4/libexec/gcc/x86_64-unknown-linux-gnu/4.8.4/:/home/sebth/eb/software/GCC/4.8.4/libexec/gcc/x86_64-unknown-linux-gnu/:/home/sebth/eb/software/GCC/4.8.4/lib/gcc/x86_64-unknown-linux-gnu/4.8.4/:/home/sebth/eb/software/GCC/4.8.4/lib/gcc/x86_64-unknown-linux-gnu/:/home/sebth/eb/software/GCC/4.8.4/lib/gcc/x86_64-unknown-linux-gnu/4.8.4/../../../../x86_64-unknown-linux-gnu/bin/x86_64-unknown-linux-gnu/4.8.4/:/home/sebth/eb/software/GCC/4.8.4/lib/gcc/x86_64-unknown-linux-gnu/4.8.4/../../../../x86_64-unknown-linux-gnu/bin/
libraries: =
/home/sebth/eb/software/GCC/4.8.4/lib64/x86_64-unknown-linux-gnu/4.8.4/:
/home/sebth/eb/software/GCC/4.8.4/lib64/../lib64/:
/home/sebth/eb/software/GCC/4.8.4/lib/x86_64-unknown-linux-gnu/4.8.4/:
/home/sebth/eb/software/GCC/4.8.4/lib/../lib64/:
/home/sebth/eb/software/GCC/4.8.4/lib/gcc/x86_64-unknown-linux-gnu/4.8.4/:
[...]

PR #1030, opened 2016-11-11
give /lib preference over /lib64 & co, enable installation of libliberty by default

boegel:
See suggestion in #921 by @sebth on giving preference to `/lib` subdirectories, and #969 by @stachon suggesting to enable the installation of `libiberty`.

I'm going to test https://github.com/hpcugent/easybuild-easyconfigs/pull/3776, https://github.com/hpcugent/easybuild-easyconfigs/pull/3777 and https://github.com/hpcugent/easybuild-easyconfigs/pull/3778 on top of this PR.

### DIFF
--- a/easybuild/easyblocks/b/binutils.py
+++ b/easybuild/easyblocks/b/binutils.py
@@ -161,7 +161,10 @@ class EB_binutils(ConfigureMake):
         }
 
         if self.cfg['install_libiberty']:
-            custom_paths['files'].extend([os.path.join('lib', 'libiberty.a'), os.path.join('include', 'libiberty.h')])
+            custom_paths['files'].extend([
+                (os.path.join('lib', 'libiberty.a'), os.path.join('lib64', 'libiberty.a')),
+                os.path.join('include', 'libiberty.h'),
+            ])
 
         # if zlib is listed as a dependency, it should get linked in statically
         if get_software_root('zlib'):
--- a/easybuild/easyblocks/b/binutils.py
+++ b/easybuild/easyblocks/b/binutils.py
@@ -155,6 +155,9 @@ class EB_binutils(ConfigureMake):
             'dirs': [],
         }
 
+        if self.cfg['install_libiberty']:
+            custom_paths['files'].extend([os.path.join('lib', 'libiberty.a'), os.path.join('include', 'libiberty.h')])
+
         # if zlib is listed as a dependency, it should get linked in statically
         if get_software_root('zlib'):
             for binary in binaries:
--- a/easybuild/easyblocks/b/binutils.py
+++ b/easybuild/easyblocks/b/binutils.py
@@ -32,7 +32,6 @@ import os
 import re
 from distutils.version import LooseVersion
 
-import easybuild.tools.environment as env
 from easybuild.easyblocks.generic.configuremake import ConfigureMake
 from easybuild.framework.easyconfig import CUSTOM
 from easybuild.tools.build_log import EasyBuildError
@@ -113,11 +112,11 @@ class EB_binutils(ConfigureMake):
         if self.cfg['install_libiberty']:
             cflags = os.getenv('CFLAGS')
             if cflags:
-                env.setvar('CFLAGS', "%s -fPIC" % cflags)
+                self.cfg.update('buildopts', 'CFLAGS="$CFLAGS -fPIC"')
             else:
                 # if $CFLAGS is not defined, make sure we retain "-g -O2",
                 # since not specifying any optimization level implies -O0...
-                env.setvar('CFLAGS', "-g -O2 -fPIC")
+                self.cfg.update('buildopts', 'CFLAGS="-g -O2 -fPIC"')
 
     def install_step(self):
         """Install using 'make install', also install libiberty if desired."""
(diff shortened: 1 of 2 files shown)


boegel:
two updates:

* we can't enable `libiberty` by default for now, since it's not built with `-fPIC` which causes problems when building e.g. `binutils` with that GCC (@stachon: are you up for figuring this out?)

* I added a custom easyconfig parameter `prefer_lib_subdir` that can be used to disable the patching of `MULTILIB_OSDIRNAMES`

* patching of `MULTILIB_OSDIRNAMES` is disabled when the `multilib` easyconfig parameter is enabled, since it causes problems (it breaks building the GCC easyconfigs that have `CLooG` support enabled, cfr. https://gist.github.com/boegel/7a828f3370e8ebccee56e7e2f20b1e27)

ocaisa:
It is possible to enable the libiberty default in binutils instead (that's what we do)

boegel:
@ocaisa do you think it makes sense to always enable `libiberty` through the `binutils` easyblock (by default, we can provide an easyconfig parameter to disable it)?

ocaisa:
I guess its a question of being consistent, if our defaults consistently only provide it in binutils then we should be safe...we could even bomb out of GCC if someone sets the configopt?

boegel:
@ocaisa is `binutils` then by default built with `-fPIC`? Do you only pass `--enable-install-libiberty` when building `binutils`, or do you do more than that?

Also, *why* did you enable that exactly? What's the use case for `libiberty`?

boegel:
@wpoely86, @geimer: thoughts on this? This seems to be working great, I'm very much keen on merging this in, unless there are any objections.

geimer:
@boegel I was having some trouble with `--enable-install-libiberty` recently, though it may have been due to not setting up `-fPIC` correctly (especially in the `dummy` build which is used whenever you bild something for `GCCcore` and where you have to enable `-fPIC` manually).  I have to try again, but won't have time for it until after SC...

ocaisa:
Lgtm

ocaisa:
...at least the libiberty part ;)

ocaisa:
Missing a sanity check for libiberty in the binutils easyblock

boegel:

Tested this thoroughly using a large bunch of GCC rebuilds, and by rebuilding all existing `binutils` easyconfigs.

Going in!

easybuild/easyblocks/b/binutils.py
```diff
@@ -99,6 +109,34 @@ def configure_step(self):
...
+                copy_file(os.path.join(self.cfg['start_dir'], 'include', 'libiberty.h'),
+                          os.path.join(self.installdir, 'include', 'libiberty.h'))
+
+            if not glob.glob(os.path.join(self.installdir, 'lib*', 'libiberty.a')):
+                copy_file(os.path.join(self.cfg['start_dir'], 'libiberty', 'libiberty.a'),
+                          os.path.join(self.installdir, 'lib', 'libiberty.a'))
+
+            if not os.path.exists(os.path.join(self.installdir, 'info', 'libiberty.texi')):
+                copy_file(os.path.join(self.cfg['start_dir'], 'libiberty', 'libiberty.texi'),
+                          os.path.join(self.installdir, 'info', 'libiberty.texi'))
```
geimer: Although this has been merged already: does it really make sense to install the `.texi` file?  I thought that this is the input file for `texinfo` to generate the actual `.info` file to be installed...

boegel: @ocaisa was doing that too, but I don't mind cutting that out. Honestly, I didn't spend much time questioning it's usefulness. ;)

ocaisa:
Its a long time since I first did this, not sure why I included that, I may have just searched for everything libiberty related...or I may have gone away and checked what was being installed with old binutils builds

easybuild/easyblocks/g/gcc.py
```diff
@@ -287,6 +294,16 @@ def configure_step(self):
             self.log.info("Performing regular GCC build...")
             configopts += " --prefix=%(p)s --with-local-prefix=%(p)s" % {'p': self.installdir}
 
+        # prioritize lib over lib{64,32,x32} for all architectures by overriding default MULTILIB_OSDIRNAMES config
+        # only do this when multilib is not enabled
+        if self.cfg['prefer_lib_subdir'] and not self.cfg['multilib']:
+            cfgfile = 'gcc/config/i386/t-linux64'
+            multilib_osdirnames = "MULTILIB_OSDIRNAMES = m64=../lib:../lib64 m32=../lib:../lib32 mx32=../lib:../libx32"
+            self.log.info("Patching MULTILIB_OSDIRNAMES in %s with '%s'", cfgfile, multilib_osdirnames)
+            write_file(cfgfile, multilib_osdirnames, append=True)
```
boegel: @sebth it looks like this is causing problems with older GCC versions, i.e. GCC 4.7.2 and older (except for GCC 4.6.4)... GCC 4.7.3 and newer are OK.

This is the error:
```
/usr/lib/crti.o: could not read symbols: File in wrong format
collect2: ld returned 1 exit status
```

with:

```
$ file /usr/lib/crti.o
/usr/lib/crti.o: ELF 32-bit LSB relocatable, Intel 80386, version 1 (SYSV), not stripped
```

I'll add a `prefer_lib_subdir = False` in those easyconfigs, but do you happen to know what the problem is?

PR #1031, opened 2016-11-11
avoid CMake fiddling with the RPATHs injected by EasyBuild via --rpath

boegel:
Passing `-DCMAKE_SKIP_RPATH=ON` when using `--rpath` is required to avoid that CMake strips *all* RPATHs from the built binaries/libraries before actually installing them (which it does by default, urgh).

Required to get METIS to build in conjunction with `--rpath` (cfr. https://github.com/hpcugent/easybuild-framework/pull/1942).

### DIFF
--- a/easybuild/easyblocks/m/metis.py
+++ b/easybuild/easyblocks/m/metis.py
@@ -37,7 +37,8 @@ from distutils.version import LooseVersion
 
 from easybuild.easyblocks.generic.configuremake import ConfigureMake
 from easybuild.tools.build_log import EasyBuildError
-from easybuild.tools.filetools import mkdir
+from easybuild.tools.config import build_option
+from easybuild.tools.filetools import apply_regex_substitutions, mkdir
 from easybuild.tools.run import run_cmd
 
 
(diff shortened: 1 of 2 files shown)


boegel:
@wpoely86 please review?

easybuild/easyblocks/m/metis.py
```diff
@@ -47,14 +48,17 @@ class EB_METIS(ConfigureMake):
...
         self.lib_exts = []
 
     def configure_step(self, *args, **kwargs):
         """Configure build using 'make config' (only for recent versions (>= v5))."""
 
         if LooseVersion(self.version) >= LooseVersion("5"):
 
+            if build_option('rpath'):
+                # patch Makefile to tell CMake not to wipe the RPATHs we inject...
+                apply_regex_substitutions('Makefile', [(r'^(CONFIG_FLAGS\s*=\s*)', r'\1-DCMAKE_SKIP_RPATH=ON ')])
```
wpoely86: no space between `\1` and `-D...`?

boegel: Should be there already, but doesn't hurt to make sure, you're right, I'll add it.

boegel:

PR #1038, opened 2016-11-15
also build Boost MPI library in parallel

boegel:
wrapper PR for #1011 by @dolfim, fixes typo in adding parallel option to `bjam` command


boegel:
tested on a bunch of existing Boost easyconfigs (old & recent), looking good, thanks @dolfim!

PR #1041, opened 2016-11-16
Handle gipaw correctly

### LINKED ISSUE #1039: Offline QuantumEspresso installation not possible
It turns out that QuantumEspresso has a (new? I'm using 6.0) feature to download the necessary sources for the extensions. Since this is not handled by EB, on an offline node you end up with the following error:
```
*** Unable to download qe-gipaw. Test whether curl or wget is installed and working,
*** if you have direct access to internet. If not, copy into archive/ the file
*** located here http://files.qe-forge.org/index.php?file=qe-gipaw-6.0.tar.gz
```
If your node is online you won't notice this

ocaisa:
Resolves #1039 

The unpacked gipaw directory begins with `qe-gipaw`, take this into account (and do it for any similar cases too).


boegel:
lgtm, tested with all existing QE easyconfigs, thanks @ocaisa!

PR #1020, opened 2016-10-21
Add support for ELPA, CP2K v4.1 and fix warning with intel compilers

damianam:

PR #1043, opened 2016-11-16
Add psmp build support for foss

ocaisa:
Add proper support for a psmp build with foss, this requires that openmp is set as a toolchain option otherwise you will get a link error for fftw.

Built on top of #1020


boegel:
merged via #1084, thanks @ocaisa!

PR #1048, opened 2016-11-23
update SAMtools easyblock for recent versions

boegel:
use './configure --prefix' and 'make install' for SAMtools v1.3 and newer, but also still copy include/lib and missing files in bin/


boegel:
Thanks for the review @rjeschmi.

I retested this updated SAMtools easyblock using all existing SAMtools easyconfigs + https://github.com/hpcugent/easybuild-easyconfigs/pull/3855, all good.

PR #1052, opened 2016-11-29
add support to Toolchain generic easyblock to include statements in generated module file that set up build environment (WIP)

boegel:
(WIP because it doesn't fully work yet, this is trickier than I expected it to be...)


pescobar:
+1

PR #1055, opened 2016-12-01
Add ambertools, derived from amber

verdurin:
Simple conversion of the existing `amber` for use with `ambertools`, which doesn't require a license.

In testing I found that the updating functionality was semi-broken in that it truncated the `patchlevels` value to a single digit, with the result that when set to 20, the current patch level, it instead only updated to patch 2. As a result, I've applied a workaround to use the original value from the easyconfig.

### DIFF
--- a/easybuild/easyblocks/a/ambertools.py
+++ b/easybuild/easyblocks/a/ambertools.py
@@ -1,0 +1,224 @@
+##
+# Copyright 2009-2016 Ghent University
+# Copyright 2015-2016 Stanford University
+#
+# This file is part of EasyBuild,
+# originally created by the HPC team of Ghent University (http://ugent.be/hpc/en),
+# with support of Ghent University (http://ugent.be/hpc),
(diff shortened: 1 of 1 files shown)


boegel:
Hmm, with this approach there's quite a bit of copy-pasting between the easyblocks for `Amber` and `AmberTools`, is there not @verdurin?

Should we look into a more integrated solution, where both easyblocks leverage common functionality (which is hosted in either `amber*.py` Python module)?

verdurin:
@boegel that would certainly be much more elegant, and preferred.

With a parameter to distinguish between the two, the need for the duplication would be obviated.

boegel:
@verdurin My PR (https://github.com/akesandgren/easybuild-easyblocks/pull/7) to @akesandgren branch for https://github.com/easybuilders/easybuild-easyblocks/pull/1305 make `EB_Amber` compatible with `AmberTools`, so this can be closed probably?

Easyconfig PR for `AmberTools` is coming soon

verdurin:
Okay, closing. Thanks @boegel

PR #1056, opened 2016-12-01
skip RPATH sanity check for binary installations

pescobar:
I have tested this patch to install Java and it seems to work. It solves this issue: https://github.com/hpcugent/easybuild-framework/issues/2039

### DIFF
--- a/easybuild/easyblocks/a/ansys.py
+++ b/easybuild/easyblocks/a/ansys.py
@@ -97,3 +97,8 @@ class EB_ANSYS(EasyBlock):
            'dirs': [os.path.join(self.ansysver, x) for x in ["ansys", "aisol", "CFD-Post","CFX"]]
         }
         super(EB_ANSYS, self).sanity_check_step(custom_paths=custom_paths)
+    
+    def sanity_check_rpath(self):
+        """Skip the rpath sanity check, this is binary software"""
+        self.log.info("RPATH sanity check is skipped when using %s easyblock.",
+                      self.__class__.__name__)
--- a/easybuild/easyblocks/a/ansys.py
+++ b/easybuild/easyblocks/a/ansys.py
@@ -31,26 +31,18 @@ EasyBuild support for installing ANSYS, implemented as an easyblock
 import os
 import stat
 
-from easybuild.framework.easyblock import EasyBlock
+from easybuild.easyblocks.generic.packedbinary import PackedBinary
 from easybuild.tools.run import run_cmd
 from easybuild.tools.filetools import adjust_permissions
 
 
-class EB_ANSYS(EasyBlock):
+class EB_ANSYS(PackedBinary):
     """Support for installing ANSYS."""
 
     def __init__(self, *args, **kwargs):
         """Initialize ANSYS-specific variables."""
         super(EB_ANSYS, self).__init__(*args, **kwargs)
         self.ansysver = "v%s" % ''.join(self.version.split('.')[0:2])
-
-    def configure_step(self):
-        """No configuration for ANSYS."""
-        pass
-
-    def build_step(self):
-        """No building for ANSYS."""
-        pass
 
     def install_step(self):
         """Custom install procedure for ANSYS."""
@@ -97,8 +89,3 @@ class EB_ANSYS(EasyBlock):
            'dirs': [os.path.join(self.ansysver, x) for x in ["ansys", "aisol", "CFD-Post","CFX"]]
         }
         super(EB_ANSYS, self).sanity_check_step(custom_paths=custom_paths)
-    
-    def sanity_check_rpath(self):
-        """Skip the rpath sanity check, this is binary software"""
-        self.log.info("RPATH sanity check is skipped when using %s easyblock.",
-                      self.__class__.__name__)
--- a/easybuild/easyblocks/generic/binary.py
+++ b/easybuild/easyblocks/generic/binary.py
@@ -127,7 +127,7 @@ class Binary(EasyBlock):
     def sanity_check_rpath(self):
         """Skip the rpath sanity check, this is binary software"""
         self.log.info("RPATH sanity check is skipped when using %s easyblock (derived from Binary)",
-              self.__class__.__name__)
+                      self.__class__.__name__)
 
     def make_module_extra(self):
         """Add the install directory to the PATH."""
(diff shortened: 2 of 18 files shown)


easybuild/easyblocks/generic/binary.py
```diff
@@ -124,6 +124,10 @@ def post_install_step(self):
 
         super(Binary, self).post_install_step()
 
+    def sanity_check_rpath(self):
+        """Skip the rpath sanity check, this is binary software"""
+        pass
```
boegel: @pescobar I would emit a clear log message rather than just skipping this method:

```python
self.log.info("Skipping RPATH sanity check for binary installations")
```

boegel:
@pescobar are you up for extending this to other easyblocks as well, e.g. definitely `IntelBase`, `PackedBinary` & `Tarball`, and some other software-specific easyblocks that are basically binary installations (ANSYS, MATLAB, FLUENT, PGI, Maple, MCR, ...)?

You can use http://easybuild.readthedocs.io/en/latest/version-specific/easyblocks.html for inspiration.

Of course, you don't need to touch easyblocks that derive from e.g. `Binary` & other easyblocks that have this included.

cc @EricDeveaud

pescobar:
@boegel I have applied your suggestion to print a message to log

I will look at `IntelBase`, `PackedBinary` and `Tarball` easyblocks

pescobar:
@boegel trying to do this more generic I tested this:

```
    def sanity_check_rpath(self):
        """Skip the rpath sanity check, this is binary software"""
        self.log.info("Skipping RPATH sanity check for %s installations" % self.__class__.__name__
```

And then I get this in the logs:

```== 2016-12-01 11:12:06,841 binary.py:130 INFO Skipping RPATH sanity check for EB_Java installations```

do you like this approach or do you suggest a different one?

**Comment by boegel** [comment]
@pescobar makes sense!

**Comment by boegel** [comment]
@pescobar just use:

```python
self.log.info("Skipping RPATH sanity check for %s installations", self.__class__.__name__)
```

(`, `, not `%`)

That way the string will only be templated if it's really printed.

pescobar:
@boegel thanks for the tip, I will always be a python noob ;)

what I don't like of this first approach I tried is that you get `EB_Java ` in the log and then to figure out where is the code which actually skips the sanity check you have to go to the `java` block to see that it extends `packedbinary` which then extends `binary` and then you find the code. 

Do you think it's fine like that? or is there anything more useful which could be printed in the log?

boegel:
@pescobar well, you could hardcode `Binary` (and change it in all easyblock you do the skip), or you could travel up the object inheritance tree and try and figure out where it really comes from...

I'm not sure how to do the latter, but it's certainly possible. Maybe see http://stackoverflow.com/questions/2654113/python-how-to-get-the-callers-method-name-in-the-called-method ?

pescobar:
@boegel I think I will hardcode the name. There are not so many exceptions I think. See this comment: http://stackoverflow.com/questions/2654113/python-how-to-get-the-callers-method-name-in-the-called-method#comment45178767_2654130

pescobar:
@boegel I have applied the change. If you agree I will apply the same for `IntelBase` and `Tarball` easyblocks

I think it's not needed for `PackedBinary` as it extends `Binary`

easybuild/easyblocks/generic/binary.py
```diff
@@ -124,6 +124,10 @@ def post_install_step(self):
 
         super(Binary, self).post_install_step()
 
+    def sanity_check_rpath(self):
+        """Skip the rpath sanity check, this is binary software"""
+        self.log.info("Skipping RPATH sanity check for %s installations. This method is inherited from the Binary easyblock", self.__class__.__name__)
```
boegel: @pescobar line is too long now (should be 120 chars or less)...

How about:

```python
self.log.info("RPATH sanity check is skipped when using %s easyblock (derived from Binary)",
              self.__class__.__name__)
```

pescobar:
@boegel thanks for the input. fix applied as you suggested

boegel:
@pescobar lgtm

pescobar:
here you have the PR for the other blocks:

https://github.com/hpcugent/easybuild-easyblocks/pull/1057
https://github.com/hpcugent/easybuild-easyblocks/pull/1058

boegel:
@pescobar no need to open separate PRs, just update those easyblocks via this PR...

The change is pretty trivial (easy to review), and won't affect existing installations at all (other than fixing the bug under `--rpath`).

pescobar:
@boegel I have closed all the previous PR and added all the commits in this one

Let me know if you have any remark or if there is any other block where the rpath sanity check should be skipped

easybuild/easyblocks/a/ansys.py
```diff
@@ -97,3 +97,8 @@ def sanity_check_step(self):
            'dirs': [os.path.join(self.ansysver, x) for x in ["ansys", "aisol", "CFD-Post","CFX"]]
         }
         super(EB_ANSYS, self).sanity_check_step(custom_paths=custom_paths)
+    
+    def sanity_check_rpath(self):
+        """Skip the rpath sanity check, this is binary software"""
+        self.log.info("RPATH sanity check is skipped when using %s easyblock.",
+                      self.__class__.__name__)
```
boegel: @pescobar short enough for one single line

boegel:
@pescobar more easyblocks that need this: `EB_ant`, `EB_BiSearch`, `EB_Chimera`, `EB_Eigen`, `EB_FDTD_underscore_Solutions`, `EB_IronPython`, `EB_picard`, `EB_Samcef`, `EB_TotalView`

For several of these, the right approach is probably to let these derive from `Binary` (or `Tarball`) rather than from `EasyBlock`, although that needs to be checked carefully...

pescobar:
I just checked and all those derive from `EasyBlock`

boegel:
@pescobar please review/merge https://github.com/pescobar/easybuild-easyblocks/pull/6

I'll make sure these changes don't break anything by retesting all easyconfigs affected by these changes before I merge this PR.
Luckily, that'll be rather fast in this case since it's all binary installations... :)

easybuild/easyblocks/m/mcr.py
```diff
@@ -38,15 +38,14 @@
 import shutil
 import stat
 
-from distutils.version import LooseVersion
```
boegel: this should be removed... testing always pays off ;)

boegel:
Extensively tested these changes, all good, thanks for diving in @pescobar!

PR #1060, opened 2016-12-01
skip rpath sanity check for matlab

pescobar:


boegel:
@pescobar please include this in #1056, no need for separate PRs

PR #1062, opened 2016-12-01
Boost

fizwit:
bjam has been superseded by b2 since Boost 1.47.  This patch adds the option of using b2 for building Boost. This patch changes the default behavior of using MPIC wrapper from the environment for MPI. b2 is much better about finding MPI libraries than bjam.  MPI should be defined with EB toolchain and not from the OS.
"b2 = True" must be specified in easyconfig to enable this new feature. Otherwise, the default bjam will be used. This will ensure that older of versions of Boost will build the same as they always have.
I have tested this with Boost-1.62 on Ubuntu 14.04 and Ubuntu 16.04 with foss-2016b (OPENMPI)
This should be reviewed by someone who is more knowledgeable about Boost than I am.

### DIFF
--- a/easybuild/easyblocks/b/boost.py
+++ b/easybuild/easyblocks/b/boost.py
@@ -157,16 +157,8 @@ class EB_Boost(EasyBlock):
                 else: 
                     raise EasyBuildError("Bailing out: only PrgEnv-gnu supported for now")
             elif self.cfg['b2']:
-                # use MPI from EB toolchain
-                mpilibs = [os.getenv('EBROOTMPIC'),os.getenv("EBROOTOPENMPI"),os.getenv("EBROOTGOMPI")]
-                if len(mpilibs):
-                    for mpi in mpilibs:
-                        if mpi is not None:
-                            self.log.info('found: MPI libraries from environement: %s'% mpi)
-                            txt += 'using mpi : %s ;\n' % mpi
-                            break
-                else:
-                    raise EasyBuildError("Failed to find MPI in toolchain")
+                # b2 will find MPI libraries; Be sure that your EB toolchain contains MPI support
+                txt = 'using mpi ;\n'
             else:
                 txt = "using mpi : %s ;\n" % os.getenv("MPICXX")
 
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/b/boost.py
```diff
@@ -99,13 +100,18 @@ def configure_step(self):
...
-        try:
-            self.objdir = os.path.join(self.builddir, 'obj')
-            os.mkdir(self.objdir)
-            self.log.debug("Succesfully created directory %s" % self.objdir)
-        except OSError, err:
-            raise EasyBuildError("Failed to create directory %s: %s", self.objdir, err)
+        if self.cfg['b2']:
+            #  Boost >1.47 b2 creates "stage" directory as build target; No need to create build directory
+            self.objdir = os.path.join(self.builddir, 'stage')
+            pass
```
wpoely86: you can drop the `pass` here?

easybuild/easyblocks/b/boost.py
```diff
@@ -189,21 +209,38 @@ def build_step(self):
...
-                    shutil.copy2(src, dst)
-        except OSError, err:
-            raise EasyBuildError("Copying %s to installation dir %s failed: %s", self.objdir, self.installdir, err)
+    def install_step(self):
+        """Install Boost. b2 install installs to prefix. Else by copying file to install dir."""
+
+        if self.cfg['b2']:
+            cmd = './b2 --user-config=user-config.jam --prefix=%s install' % self.installdir
+            self.log.info('Install with b2; cmd=%s' % cmd )
+            (out, ec) = run_cmd(cmd, log_ok=False, log_all=True )
```
wpoely86: run this file through `pep8` for style please 😉

easybuild/easyblocks/b/boost.py
```diff
@@ -150,12 +156,26 @@ def configure_step(self):
...
+    def build_boost_b2(self):
+        """as of Boost 1.47.0, the Boost build tool has changed from bjam t0o b2
+           All cconfiguration should be done in <configure> step. ( bootstrap and user-config.jam)
+        """
+
+        cmd = "./b2 --user-config=user-config.jam"
+        self.log.info('Building Boost with b2; cmd=%s' % cmd )
+        (out, ec) = run_cmd(cmd, log_all=True, log_ok=False, simple=False )
+        if ec == 1: 
+            raise EasyBuildError("Boost built to completion, but not libraries were built. Check your Boost configuration")
```
wpoely86: `not all libraries`?

wpoely86:
Looking good @fizwit 
We will need to run some more test.

boegel:
follow-up in #1070

PR #1064, opened 2016-12-07
Pass CXXFLAGS and LDFLAGS to Boost Bjam

dolfim:
This makes the easyblock more consistent with the others and specially allows to change the C++ std just using the toolchainopt.

### DIFF
--- a/easybuild/easyblocks/b/boost.py
+++ b/easybuild/easyblocks/b/boost.py
@@ -159,9 +159,14 @@ class EB_Boost(EasyBlock):
     def build_step(self):
         """Build Boost with bjam tool."""
 
-        cxxflags = os.environ['CXXFLAGS']
-        ldflags = os.environ['LDFLAGS']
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/b/boost.py
```diff
@@ -158,7 +159,9 @@ def configure_step(self):
     def build_step(self):
         """Build Boost with bjam tool."""
 
-        bjamoptions = " --prefix=%s" % self.objdir
+        cxxflags = os.environ['CXXFLAGS']
+        ldflags = os.environ['LDFLAGS']
```
boegel: @dolfim looks good, but it's safer to use `os.getenv` or `os.environ.get`, so you can specify a fallback value in case `$CXXFLAGS` or `$LDFLAGS` are not defined?

Or, alternatively, these should only be passed down *if* they are defined (which is usually the case in EasyBuild, but not always, e.g. when installing with `dummy` toolchain).

dolfim:
@boegel Just updated as requested.

boegel:
lgtm, tested with a bunch of existing Boost easyconfigs, thanks @dolfim!

PR #1069, opened 2016-12-09
update sanity check for Mono to support recent versions

boegel:
cc @backelj


boegel:
tested with https://github.com/hpcugent/easybuild-easyconfigs/pull/3871 & existing `Mono` easyconfigs

boegel:

PR #1073, opened 2016-12-19
Make the Rosetta EasyBlock work in --module-only mode

tobbez:
Previously the sanity check would fail since None was inserted into the
paths of the files to be checked (rather than the actual compiler name).


boegel:
lgtm, thanks for looking into this @tobbez!

PR #1074, opened 2016-12-19
fix eigen sanity check for latest 3.3.1

pescobar:


boegel:
This makes the sanity check slightly less strict, but that's OK imho, there's little that could go wrong anyway since Eigen is only a set of header files.

Thanks @pescobar!

PR #974, opened 2016-07-27
ENH: add support for WPS3.8

andreas-h:
- fix broken download link (l. 69)
- replace geodata download with larger file which also contains GMTED2010 topography which is needed by WPS3.8 (l. 70) see http://www2.mmm.ucar.edu/wrf/users/download/get_sources_wps_geog.html
- support `ungrib` testing for WPS3.8, which doesn't contain `Vtable.ARW` any more. However, when using `Vtable.ARW.UPP`, the test in WPS3.8 succeeds.

PR #1078, opened 2017-01-04
ENH: add support for WPS 3.8

andreas-h:
after I messed up the history of #974, here's a new attempt at adding WPS3.8 support.


andreas-h:
sorry for the noise, picked wrong branch ...

PR #1079, opened 2017-01-04
ENH: add support for WPS 3.8

andreas-h:
after messing up the history of #974, here's my new attempt at adding support for WPS3.8

### DIFF
--- a/easybuild/easyblocks/w/wps.py
+++ b/easybuild/easyblocks/w/wps.py
@@ -34,7 +34,6 @@ EasyBuild support for building and installing WPS, implemented as an easyblock
 import fileinput
 import os
 import re
-import shutil
 import sys
 import tempfile
 from distutils.version import LooseVersion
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/w/wps.py
```diff
@@ -63,17 +62,19 @@ def __init__(self, *args, **kwargs):
         self.comp_fam = None
         self.wrfdir = None
         self.compile_script = None
+        testdata_urls = ["http://www2.mmm.ucar.edu/wrf/src/data/avn_data.tar.gz"]
+        if LooseVersion(self.version) < LooseVersion('3.8'):
+            testdata_urls += ["http://www2.mmm.ucar.edu/wrf/src/wps_files/geog.tar.gz"]  # 697MB download, 16GB unpacked!
+        else:
+            testdata_urls += ["http://www2.mmm.ucar.edu/wrf/src/wps_files/geog_complete.tar.bz2"]  # 2.3GB download!
+        self.cfg['testdata'] = testdata_urls
```
boegel: @andreas-h we should anticipate that people can potentially specify URLs in their easyconfig too, so we should check whether `self.cfg['testdata'` is stil `None` before we decide to override it...

easybuild/easyblocks/w/wps.py
```diff
@@ -63,17 +62,19 @@ def __init__(self, *args, **kwargs):
         self.comp_fam = None
         self.wrfdir = None
         self.compile_script = None
+        testdata_urls = ["http://www2.mmm.ucar.edu/wrf/src/data/avn_data.tar.gz"]
+        if LooseVersion(self.version) < LooseVersion('3.8'):
+            testdata_urls += ["http://www2.mmm.ucar.edu/wrf/src/wps_files/geog.tar.gz"]  # 697MB download, 16GB unpacked!
```
boegel: @andreas-h please use `test_data_urls.append("http://...")` (also below)

boegel:
Changes look good, tested with https://github.com/hpcugent/easybuild-easyconfigs/pull/3970, so good to go.

Thanks @andreas-h!

PR #1080, opened 2017-01-04
Cleanup build before proceeding with full Boost

dolfim:
While porting the easyconfig to Boost 1.63 I got failures like:
```
error: Name clash for '...'
error: 
error: Tried to build the target twice, with property sets having 
error: these incompabile properties:
error: ...
```


boegel:
I ran into this exact same problem with Boost 1.62.0, see https://github.com/hpcugent/easybuild-easyconfigs/pull/3650, thanks for figuring this out @dolfim!

boegel:
Tested extensively with older Boost easyconfigs, works like a charm, thanks @dolfim!

dolfim:
Thanks you for the quick review!

PR #1082, opened 2017-01-05
adding in conda fix

jerowe:
Use rmtree2 to remove the installdir prior to conda create command.

### DIFF
--- a/easybuild/easyblocks/generic/conda.py
+++ b/easybuild/easyblocks/generic/conda.py
@@ -35,6 +35,7 @@ from easybuild.easyblocks.generic.binary import Binary
 from easybuild.framework.easyconfig import CUSTOM
 import easybuild.tools.environment as env
 from easybuild.tools.run import run_cmd
+from easybuild.tools.filetools import rmtree2
 
 
 class Conda(Binary):
@@ -70,6 +71,9 @@ class Conda(Binary):
         cmd = "conda config --add create_default_packages setuptools"
         run_cmd(cmd, log_all=True, simple=True)
 
+        # conda requires that the installdir is empty
+        rmtree2(self.installdir)
+
         if self.cfg['environment_file'] or self.cfg['remote_environment']:
 
             if self.cfg['environment_file']:
@@ -79,7 +83,6 @@ class Conda(Binary):
 
             self.set_conda_env()
 
-            # use --force to ignore existing installation directory
             cmd = "%s conda env create --force %s -p %s" % (self.cfg['preinstallopts'], env_spec, self.installdir)
             run_cmd(cmd, log_all=True, simple=True)
 
--- a/easybuild/easyblocks/generic/conda.py
+++ b/easybuild/easyblocks/generic/conda.py
@@ -83,11 +83,11 @@ class Conda(Binary):
 
             self.set_conda_env()
 
-            cmd = "%s conda env create --force %s -p %s" % (self.cfg['preinstallopts'], env_spec, self.installdir)
+            cmd = "%s conda env create %s -p %s" % (self.cfg['preinstallopts'], env_spec, self.installdir)
             run_cmd(cmd, log_all=True, simple=True)
 
         else:
-            cmd = "%s conda create --force -y -p %s" % (self.cfg['preinstallopts'], self.installdir)
+            cmd = "%s conda create -y -p %s" % (self.cfg['preinstallopts'], self.installdir)
             run_cmd(cmd, log_all=True, simple=True)
 
             if self.cfg['requirements']:


easybuild/easyblocks/generic/conda.py
```diff
@@ -79,7 +83,6 @@ def install_step(self):
 
             self.set_conda_env()
 
-            # use --force to ignore existing installation directory
             cmd = "%s conda env create --force %s -p %s" % (self.cfg['preinstallopts'], env_spec, self.installdir)
```
boegel: @jerowe so, what *does* `--force` do then? Should we remove that too?

boegel:
@jerowe please open PRs to `develop` rather than `master` (I changed this myself for this PR)

easybuild/easyblocks/generic/conda.py
jerowe: Its usage is rather mysterious to me. I think it means that it will create the env even if one or more of the install packages fails.

jerowe: From the conda docs
```
    -f, --force
              Force install (even when  package  already  installed),  implies
              --no-deps.
```

jerowe: Maybe we should have a separate option for conda options?

```
conda_options=['--force', '--copy']
```

There are a lot of options in conda.

FYI, I think I initially thought that --force would make conda behave nicely with an existing directory, but that does not see to be the case.

jerowe:
The issue was added to the 3.1 milestone. When you think that will be out?

boegel:
3.1.0 is planed for end of this month, but we have another intermediate release planned (3.0.3) hopefully by the end of next week, I can try and include it there.

I'll try and see if I can reproduce the issue with EasyBuild 3.0.2, since I'm really puzzled as to why we missed this...

jerowe:
Alright. Just wondering.

boegel:
@jerowe I just tried to install the two easyconfigs included in EasyBuild 3.0.2 that rely on the `conda` easyblock, and neither of them seems to suffer from this issue...

Installing `perl-app-cpanminus-1.7039.eb` still works fine.

Installing `gencore_variant_detection-1.0.eb` fails, but for a different reason:

```
== 2017-01-12 20:28:42,452 run.py:433 DEBUG cmd " conda env create --force nyuad-cgsb/gencore_variant_detection_1.0 -p /user/home/gent/vsc400/vsc40023/eb_phanpyscratch/SL6/sandybridge/software/gencore_variant_detection/1.0" exited with exitcode 1 and output:
Fetching package metadata .............
Solving package specifications: .
Using Anaconda Cloud api site https://api.anaconda.org


NoPackagesFoundError: Dependency missing in current linux-64 channels: 
  - circos 0.69.2* -> perl-readonly -> perl 5.22.0*
```

(full log @ https://gist.github.com/boegel/c397a5642fef6bd2316a5641602fc6ad)

With which easyconfig are you hitting the problem that made you look into this change?
From what I can tell, it should not be needed, but it may depend on the `conda` version you are using?

We have this in the `Anaconda{2,3}` easyconfig to update the `conda` included there, are you missing that?

```python
# a newer version of conda is required to run 'conda env create -p'
prep_env = "PATH=%(installdir)s/bin:$PATH "
postinstallcmds = [prep_env + "conda install -f -p %(installdir)s -c conda conda=4.2.12 ruamel_yaml=0.11.14"]
```

jerowe:
@boegel , this has to do with changes in the bioconda channel, actually. They recently joined forces with the conda-forge channel. I need to add that to all my envs, and it should be added to the perl-app-cpanminus eb in the channels list. 

I just fixed the Let me work up that PR for the perl-app-cpanminus config.

jerowe:
Alright, let me start again. I just got up and read your message, and I should have read it more carefully.

So you aren't encountering the problem? It may be the conda version. I had to roll back to conda=4.2.9, because multiuser support gets broken in >4.2.9. I'm hoping it gets fixed soon and I can move on with my life, but I don't know when that will be.

If that's the case it might be a good idea to delete the directories to support earlier versions of conda.

In any case I should do some testing.

jerowe:
I don't have the problem I described earlier now either, using conda version 4.2.12. Let me try it again with conda=4.2.9. Maybe that is where my troubles are.

boegel:
@jerowe Have you figured out what the problem was here?

In any case, this change doesn't seem to be required, so lets close this PR?

jerowe:
Sorry,  I meant to close this. I think it was something particular to my
set up,  and I have not been able to reproduce it.

boegel:
OK, closing this then, thanks for the feedback @jerowe!

PR #1083, opened 2017-01-06
add custom easyblock for FFTW (REVIEW)

boegel:
This easyblock takes care of building FFTW with use of AVX* instructions enabled when suitable, i.e. when the host on which it is being built supports these instructions.

(note: requires https://github.com/hpcugent/easybuild-framework/pull/2074)

Using this easyblock for building FFTW results in the same multi-cycle build as is currently done in the FFTW easyconfigs (by specifying `configopts` as a list):

```
$ eb FFTW.eb -x | grep configure_step -A 1
[configure_step method]
  running command " ./configure --prefix=/Users/kehoste/.local/easybuild/software/FFTW/3.3.5-gompi-2017a --enable-single --enable-mpi --with-pic --enable-openmp --enable-threads --enable-avx --enable-avx2 --enable-sse2"
--
[configure_step method]
  running command " ./configure --prefix=/Users/kehoste/.local/easybuild/software/FFTW/3.3.5-gompi-2017a --enable-mpi --with-pic --enable-openmp --enable-threads --enable-avx --enable-avx2 --enable-sse2"
--
[configure_step method]
  running command " ./configure --prefix=/Users/kehoste/.local/easybuild/software/FFTW/3.3.5-gompi-2017a --enable-long-double --enable-mpi --with-pic --enable-openmp --enable-threads"
--
[configure_step method]
  running command " ./configure --prefix=/Users/kehoste/.local/easybuild/software/FFTW/3.3.5-gompi-2017a --enable-quad-precision --with-pic --enable-openmp --enable-threads"
```

The use of advanced SIMD instructions (SSE2, AVX*), building of the libraries for different precisions and for MPI/OpenMP/threads can be controlled via custom easyconfig parameters supported by this easyblock, see below.

```
$ eb -e EB_FFTW -a

...

EASYBLOCK-SPECIFIC
------------------
...
use_avx*                 Configure with --enable-avx (if None, auto-detect support for AVX) [default: None]
use_avx2*                Configure with --enable-avx2 (if None, auto-detect support for AVX2) [default: None]
use_avx512*              Configure with --enable-avx512 (if None, auto-detect support for AVX512) [default: None]
use_sse2*                Configure with --enable-sse2 (if None, auto-detect support for SSE2) [default: None]
with_double_prec*        Enable building of double precision library [default: True]
with_long_double_prec*   Enable building of long-double precision library [default: True]
with_mpi*                Enable building of FFTW MPI library [default: True]
with_openmp*             Enable building of FFTW OpenMP library [default: True]
with_quad_prec*          Enable building of quad precision library [default: True]
with_single_prec*        Enable building of single precision library [default: True]
with_threads*            Enable building of FFTW threads library [default: True]

...
```

Example easyconfig to use along with this easyblock (`eb FFTW.eb --include-easyblocks fftw.py`):

```python
name = 'FFTW'
version = '3.3.5'

homepage = 'http://www.fftw.org'
description = """FFTW is a C subroutine library for computing the discrete Fourier transform (DFT)
 in one or more dimensions, of arbitrary input size, and of both real and complex data."""

toolchain = {'name': 'gompi', 'version': '2016b'}
toolchainopts = {'pic': True}

sources = [SOURCELOWER_TAR_GZ]
source_urls = [homepage]

moduleclass = 'numlib'
```

TODO:

* [x] take into account `--optarch` (in particular `--optarch=GENERIC`)
* [x] extend to also auto-detecting for ARM (`neon`, `asimd`) and POWER (`altivec`) (thoughts on this @geimer, @JackPerdue?)

cc: @akesandgren, @jhein32, @wpoely86, @rjeschmi, @verdurin

### DIFF
--- a/easybuild/easyblocks/f/fftw.py
+++ b/easybuild/easyblocks/f/fftw.py
@@ -34,11 +34,11 @@ from easybuild.tools.systemtools import X86_64, get_cpu_features
 from easybuild.tools.toolchain.compiler import OPTARCH_GENERIC
 
 
-# SSE2, AVX* (x86_64 only)
-FFTW_CPU_FEATURE_FLAGS_SINGLE_DOUBLE = ['avx', 'avx2', 'avx512', 'sse2']
+# AVX*, FMA, SSE2 (x86_64 only)
+FFTW_CPU_FEATURE_FLAGS_SINGLE_DOUBLE = ['avx', 'avx2', 'avx512', 'fma', 'sse2']
 # Altivec (POWER), SSE (x86), NEON (ARM), FMA (x86_64)
 # asimd is CPU feature for extended NEON on AARCH64
-FFTW_CPU_FEATURE_FLAGS = FFTW_CPU_FEATURE_FLAGS_SINGLE_DOUBLE + ['altivec', 'asimd', 'fma', 'neon', 'sse']
+FFTW_CPU_FEATURE_FLAGS = FFTW_CPU_FEATURE_FLAGS_SINGLE_DOUBLE + ['altivec', 'asimd', 'neon', 'sse']
 FFTW_PRECISION_FLAGS = ['single', 'double', 'long-double', 'quad-precision']
 
 
@@ -129,7 +129,10 @@ class EB_FFTW(ConfigureMake):
                 if prec in ['single', 'double']:
                     for flag in FFTW_CPU_FEATURE_FLAGS_SINGLE_DOUBLE:
                         if getattr(self, flag):
-                            prec_configopts.append('--enable-%s' % flag)
+                            if flag == 'fma':
+                                prec_configopts.append('--enable-avx-128-fma')
+                            else:
+                                prec_configopts.append('--enable-%s' % flag)
 
                 # Altivec (POWER) and SSE only for single precision
                 for flag in ['altivec', 'sse']:
@@ -139,10 +142,6 @@ class EB_FFTW(ConfigureMake):
                 # NEON (ARM) only for single precision and double precision (on AARCH64)
                 if (prec == 'single' and self.neon) or (prec == 'double' and self.asimd):
                     prec_configopts.append('--enable-neon')
-
-                # FMA support (only on recent x86_64)
-                if self.fma:
-                    prec_configopts.append('--enable-avx-128-fma')
 
                 # append additional configure options (may be empty string, but that's OK)
                 self.cfg.update('configopts', [' '.join(prec_configopts) + common_config_opts])
--- a/easybuild/easyblocks/f/fftw.py
+++ b/easybuild/easyblocks/f/fftw.py
@@ -1,0 +1,175 @@
+##
+# Copyright 2009-2017 Ghent University
+#
+# This file is part of EasyBuild,
+# originally created by the HPC team of Ghent University (http://ugent.be/hpc/en),
+# with support of Ghent University (http://ugent.be/hpc),
+# the Flemish Supercomputer Centre (VSC) (https://www.vscentrum.be),
+# Flemish Research Foundation (FWO) (http://www.fwo.be/en)
+# and the Department of Economy, Science and Innovation (EWI) (http://www.ewi-vlaanderen.be/en).
+#
+# http://github.com/hpcugent/easybuild
+#
+# EasyBuild is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation v2.
+#
+# EasyBuild is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with EasyBuild.  If not, see <http://www.gnu.org/licenses/>.
+##
+"""
+EasyBuild support for building and installing FFTW, implemented as an easyblock
+"""
+from vsc.utils.missing import nub
+
+from easybuild.easyblocks.generic.configuremake import ConfigureMake
+from easybuild.framework.easyconfig import CUSTOM
+from easybuild.tools.systemtools import X86_64, get_cpu_features
+
+
+FFTW_CPU_FEATURE_FLAGS = ['avx', 'avx2', 'avx512', 'sse2']
+FFTW_PRECISION_FLAGS = ['single', 'double', 'long-double', 'quad-precision']
+
+
+class EB_FFTW(ConfigureMake):
+    """Support for building/installing FFTW."""
+
+    @staticmethod
+    def _prec_param(prec):
+        """Determine parameter name for specified precision"""
+        return 'with_%s_prec' % prec.replace('-', '_').replace('_precision', '')
+
+    @staticmethod
+    def extra_options():
+        """Custom easyconfig parameters for FFTW."""
+        extra_vars = {
+            'with_mpi': [True, "Enable building of FFTW MPI library", CUSTOM],
+            'with_openmp': [True, "Enable building of FFTW OpenMP library", CUSTOM],
+            'with_threads': [True, "Enable building of FFTW threads library", CUSTOM],
+        }
+
+        for flag in FFTW_CPU_FEATURE_FLAGS:
+            help_msg = "Configure with --enable-%s (if None, auto-detect support for %s)" % (flag, flag.upper())
+            extra_vars['use_%s' % flag] = [None, help_msg, CUSTOM]
+
+        for prec in FFTW_PRECISION_FLAGS:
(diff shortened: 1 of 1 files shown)


JackPerdue:
Cool... can you add an option for altivec on the Power7/X?

There is an example here:

http://www.siliconslick.com/easybuild/easyconfigs/curie/curie/FFTW/FFTW-3.3.4-gompi-2016b.eb

jack (who has AVX, AVX2 and Altivec to support at the moment)

boegel:
@JackPerdue done!

We should also do this for AARCH64, but I'm tempted to postpone that to a future PR... Thoughts @geimer?

boegel:
Last commit (cd2781d) also takes into account `--optarch=GENERIC`, resulting in the `--enable-avx*` configure options not being used when it is used, see below.

Thoughts on this @pescobar?

```
$ eb FFTW.eb -x | grep configure_step -A 1
[configure_step method]
  running command " ./configure --prefix=/Users/kehoste/.local/easybuild/software/FFTW/3.3.5-gompi-2017a --enable-single --enable-mpi --with-pic --enable-openmp --enable-threads --enable-avx --enable-avx2 --enable-sse2"
--
[configure_step method]
  running command " ./configure --prefix=/Users/kehoste/.local/easybuild/software/FFTW/3.3.5-gompi-2017a --enable-mpi --with-pic --enable-openmp --enable-threads --enable-avx --enable-avx2 --enable-sse2"
--
[configure_step method]
  running command " ./configure --prefix=/Users/kehoste/.local/easybuild/software/FFTW/3.3.5-gompi-2017a --enable-long-double --enable-mpi --with-pic --enable-openmp --enable-threads"
--
[configure_step method]
  running command " ./configure --prefix=/Users/kehoste/.local/easybuild/software/FFTW/3.3.5-gompi-2017a --enable-quad-precision --with-pic --enable-openmp --enable-threads"

$ eb FFTW.eb -x --optarch=GENERIC | grep configure_step -A 1
[configure_step method]
  running command " ./configure --prefix=/Users/kehoste/.local/easybuild/software/FFTW/3.3.5-gompi-2017a --enable-single --enable-mpi --with-pic --enable-openmp --enable-threads"
--
[configure_step method]
  running command " ./configure --prefix=/Users/kehoste/.local/easybuild/software/FFTW/3.3.5-gompi-2017a --enable-mpi --with-pic --enable-openmp --enable-threads"
--
[configure_step method]
  running command " ./configure --prefix=/Users/kehoste/.local/easybuild/software/FFTW/3.3.5-gompi-2017a --enable-long-double --enable-mpi --with-pic --enable-openmp --enable-threads"
--
[configure_step method]
  running command " ./configure --prefix=/Users/kehoste/.local/easybuild/software/FFTW/3.3.5-gompi-2017a --enable-quad-precision --with-pic --enable-openmp --enable-threads"
```

geimer:
@boegel I believe enabling NEON on AArch64 is now fairly trivial. The question is more whether the resulting library works.  I'm up for giving this new easyblock a try on my ARM box, though I have no clue how to test FFTW...

boegel:
@geimer it seems like FFTW has a test suite included, you should be able to run it by including `runtest = 'check'` in your FFTW easyconfig file...

Are you up for looking into updating this easyblock to also cover AARCH64, or shall I?

I think it's just checking for the `neon` (only for single precision unless it's AARCH64 then also for double precision) and `asimd` (single/double only)?

geimer:
@boegel `neon` should only show up on AArch32.  As far as I know, the AArch64 architecture always includes a double-precision-capable NEON unit which is flagged as `asimd` (Advanced SIMD). So, yes, two additional `if`'s should suffice.

easybuild/easyblocks/f/fftw.py
```diff
@@ -0,0 +1,197 @@
...
+                        if getattr(self, flag):
+                            prec_configopts.append('--enable-%s' % flag)
+
+                # Altivec (POWER) and SSE only for single precision
+                for flag in ['altivec', 'sse']:
+                    if prec == 'single' and getattr(self, flag):
+                        prec_configopts.append('--enable-%s' % flag)
+
+                # NEON (ARM) only for single precision and double precision (on AARCH64)
+                if (prec == 'single' and self.neon) or (prec == 'double' and self.asimd):
```
geimer: I'm not sure this `if` condition is correct. I believe one can pass `--enable-neon` for both single- and double-precision if `asimd` is flagged.

boegel: Doesn't asimd imply neon?

I can update this to use `prec == 'single' and (self.asimd or self.neon)`, but it wouldn't make a difference in practice?

geimer: The thing is: the `neon` flag is only available on AArch32 if NEON support is available. And it is single-precision only. On AArch64, there is no `neon` flag, but only the `asimd` flag indicating single- and double-precision NEON support. So yes, I believe it would make a difference.

boegel: OK, thanks for the clarification, fixed in c7899c2

geimer:
@JackPerdue You may actually also be interested in VSX support, which seems to be available in POWER7 and above. However, as far as I can see there is no flag in `/proc/cpuinfo` that indicates availability. But maybe there is some other pseudo-file (e.g., under `/sys`)?

JackPerdue:
processor       : 0
cpu             : POWER7 (architected), altivec supported
clock           : 4228.000000MHz
revision        : 2.1 (pvr 004a 0201)

Given: https://en.wikipedia.org/wiki/IBM_POWER_microprocessors  I'd say grepping /proc/cpuinfo for POWER7 would do.   

Thanks for the suggestion.  I'll have to look at that here at some point (we are about to reprovision our Power7 cluster)

easybuild/easyblocks/f/fftw.py
```diff
@@ -0,0 +1,196 @@
...
+# EasyBuild is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with EasyBuild.  If not, see <http://www.gnu.org/licenses/>.
+##
+"""
+EasyBuild support for building and installing FFTW, implemented as an easyblock
```
wpoely86: `:author`?

boegel:
@JackPerdue VSC support detection implemented in https://github.com/hpcugent/easybuild-framework/pull/2078, I'll update this easyblock accordingly once that is merged

boegel:
My tests with this easyblock on latest `develop` of EasyBuild framework (on Intel Haswell, Sandy Bridge and AMD Opteron Magny Cours) all works out fine: FFTW built with this easyblock and running the provided test suite via `make check` works as expected.

easybuild/easyblocks/f/fftw.py
```diff
@@ -0,0 +1,198 @@
...
+
+        # on macOS, AVX is indicated with 'avx1.0' rather than 'avx'
+        if 'avx1.0' in cpu_features:
+            cpu_features.append('avx')
+
+        self.log.info("List of CPU features: %s", cpu_features)
+
+        # auto-detect CPU features that can be used and are not enabled/disabled explicitly,
+        # but only if --optarch=GENERIC is not being used
+        if self.cfg['auto_detect_cpu_features'] and build_option('optarch') != OPTARCH_GENERIC:
```
geimer: All x86_64 CPUs support at least SSE2, so this could be enabled even with `OPTARCH_GENERIC`.

easybuild/easyblocks/f/fftw.py
boegel: fixed, thanks

easybuild/easyblocks/f/fftw.py
boegel: Hmm, ok, that complicates things a little bit, but I'll look into it.

boegel: implemented in cb14353, please rereview?

boegel:
Thanks for the review/feedback @wpoely86, @geimer and @JackPerdue!

easybuild/easyblocks/f/fftw.py
```diff
@@ -0,0 +1,198 @@
...
+                if prec in ['single', 'double']:
+                    for flag in FFTW_CPU_FEATURE_FLAGS_SINGLE_DOUBLE:
+                        if getattr(self, flag):
+                            if flag == 'fma':
+                                prec_configopts.append('--enable-avx-128-fma')
+                            else:
+                                prec_configopts.append('--enable-%s' % flag)
+
+                # Altivec (POWER) and SSE only for single precision
+                for flag in ['altivec', 'sse']:
```
ocaisa: `altivec` does not work for GCC 5 for single precision with Power, see https://github.com/FFTW/fftw3/issues/59 (I also tested GCC6 and the tests fail there as well)

Issue #975, opened 2016-07-29
cp2k easyblock ignores --optarch=GENERIC


simo-tuomisto:
In cp2k.py:438:
optarch = ''
if self.toolchain.options['optarch']:
   optarch = '-xHOST'

and cp2k.py:484:
optarch = ''
if self.toolchain.options['optarch']:
   optarch = '-march=native'

Is this behaviour intended?

boegel:
@simo-tuomisto I just noticed this was left unanswered...

This is indeed a bug in the CP2K easyblock that we should fix.

The `-xHOST` and `-march=native` should indeed not be hardcoded in the CP2K easyblock, they should come from the toolchain mechanism. In that case, CP2K would also be built as intended under `--optarch=GENERIC` or `--optarch=<custom flags>`.

boegel:
@damianam while you're at it... ;)

boegel:
Or maybe I'll tackle this myself in the CP2K easyblock update at https://github.com/hpcugent/easybuild-easyblocks/pull/1084, and ask @damianam to review it, since he's familiar both with the CP2K easyblock and the `optarch` use in there.

boegel:
fixedin #1084

PR #1084, opened 2017-01-10
update CP2K easyblock: copy data dir, support version 4.1, support ELPA, fix psmp build with foss toolchain

boegel:
combo of #996 by @hajgato, #1020 by @damianam and #1043 by @ocaisa


easybuild/easyblocks/c/cp2k.py
```diff
@@ -207,6 +209,14 @@ def configure_step(self):
             options['LIBS'] += ' -lplumed'
             options['DFLAGS'] += ' -D__PLUMED2'
 
+        # ELPA
+        if self.cfg["elpa"]:
```
boegel: @damianam I know you modelled this after the `PLUMED` support, but I'm now unsure why we actually introduced the `plumed` configure parameter at all (@wpoely86, any ideas?)...

Instead, why not go with this approach:

```python
elpa = get_software_root('ELPA')
if elpa:
    options['LIBS'] += ' -lelpa'
    options['DFLAGS'] += ' -D__ELPA3'
    options['FCFLAGSOPT'] += ' -I%s ' % os.path.join(elpa,'include/elpa-%s/modules' % get_software_version('ELPA'))

boegel: changed in 00e4403

damianam:
@boegel IMO the `if self.cfg["plumed"]:` approach is a bit more flexible. If plumed gets pulled in by some dependency, but you don't want to enable the plumed support on CP2K for whatever reason (a bug with it for instance), you can't with your approach. That'd be a corner case though. At the end is a design decision:

-Smart easyblock that automatically detects and adds stuff
vs
-Smart user that manually chooses how to build CP2K

boegel:
@damianam OK, what I want to avoid is that you must both specify `PLUMED` as a dependency *and* set `plumed = True` in the easyconfig file, so how about this...

Use `None` as a default value for the `plumed` parameter, implying "no decision is made", and then enable PLUMED support if PLUMED is included as a dependency. If `plumed` is set to `False`, don't enable PLUMED support. If `plumed` is set to `True`, enforce that PLUMED support can be enabled, i.e. that PLUMED is included as a dependency.

That way, just specifying PLUMED as a dependency is sufficient, and you can disable PLUMED support if you want to.

In the case of CP2K, it's unlikely that PLUMED would be an indirect dependency I think, but I guess you never know going forward.

If that's OK, shall I implement the same mechanism for ELPA?

ocaisa:
and libxsmm

boegel:
One thing that bothers me is that this approach we're taking here is different from what we usually do... Why are we adding a custom parameter for some of the CP2K dependencies, but we're not doing that elsewhere?

Isn't it quite unlikely anyway that `libxsmm`, `PLUMED` and `ELPA` are included as indirect dependencies? If so, why bother?

wpoely86:
I agree it's very unlikely that PLUMED or libxsmm will be loaded as indirect deps when building CP2K...

ocaisa:
I think I probably agree, if you enough to know CP2K's dependencies are optional then you should be able to figure out that you need to exclude the dep in the easyconfig. 

One thing I like about a unified yeb format is that you could actually comment (in the easyconfig where someone would see it) why certain dependencies are there and in what scenarios they are a good/bad idea...at the moment something like that would be way too tedious to do consistently.

damianam:
@boegel your proposal seems very reasonable. I also agree that the discussed issue is unlikely to happen, but I see no strong reason to don't be proactive and avoid a potential problem, unlikely as it might be.

boegel:
@damianam I'll make the proposed change for PLUMED, since we have that parameter in place already and we should keep it to avoid breaking existing easyconfigs

I won't add it for ELPA and libxsmm though, we can do that when the need arises, i.e. if either of them is an indirect dep and we need to disable ELPA/libxsmm support for some reason.

boegel:
@damianam updated, please rerereview?

damianam:
I am not fully convinced about treating plumed differently than other dependencies/options, but I understand your reasons for it.

boegel:
@damianam I'm not sure why the `plumed` parameter was ever included, I think it was a mistake, but I want to avoid breaking backward compatibility, hence I prefer not removing it.

Since I don't see the value for now, I'd rather not extend this approach to ELPA & co though, since it's double bookkeeping, and too easy to by accident build CP2K without ELPA support even though you intended to build it with it...

boegel:
@damianam please rereview (especially fb7ee49), this now also fixes #975

damianam:
@boegel lgtm

akesandgren:
The CP2K block should actually build all versions of the binary at one go. There is no reason not to.
I.e., sopt, ssmp, popt, and psmp, instead of having separate easyconfigs for it.

akesandgren:
There is also a bug in the if ifortver >= logic.
If ifortver >= 2012 it should't do anything special (probably). Currently it ends up in the
elif ifortver >= LooseVersion("11"): section

akesandgren:
And when intel toolchain is active one should add -D__MKL to the defines when FFTW3 module is missing.
I.e., in the if not get_software_root('FFTW') section
And __FFTMKL is not used from cp2k ver 3 and probably onwards (haven't downloaded ver 4 and 5 yet)

akesandgren:
Support for "D__PW_CUDA CUDA FFT and associated gather/scatter on the GPU" should probably be added too.

boegel:
@akesandgren I was hoping to get this merged ASAP, it's pulling stuff together from 3 existing PRs + some extra changes on top.

Are any of the things you mentioned a blocker in your opinion?

If not, I suggest we follow up in another PR...

akesandgren:
The  ifortver >= logic. (easy to solve) is probably the most important at the moment.
Noe of them are blockers.

boegel:
@akesandgren The `ifortver` logic looks OK to me. 

According to https://software.intel.com/en-us/articles/build-cp2k-using-intel-fortran-compiler-professional-edition, you should compile `qs_vxc_atom` with `-O1`, which is basically what is being done there.
This suggestion also seems to apply to more recent `ifort` versions.

If you disagree, let's follow-up in another issue or PR.

W.r.t. your suggestions for enhancements, see #1086

I'm going to go ahead and merge this, thanks for all the feedback everyone!

Issue #1085, opened 2017-01-16
OpenSSL 1.1.0 sanity check issue


vanzod:
The sanity check For OpenSSL 1.1.0 fails since the library files names are hard coded into the easyblock and with the newer version number they do not match anymore.
I would suggest moving the path to the easyconfig files. Thoughts?

boegel:
@vanzod we usually deal with this sort of thing by including a version check in the easyblock, where we use the new library names for newer versions, and stick to the old names for old versions

Doing this in the easyblock is a do-once-and-forget thing, while moving the `sanity_check_paths` in the easyconfigs means we're duplicating that info all over the place (which is already a problem for other stuff though).

Anything involving logic/version checks should ideally be done in the easyblock.

boegel:
follow-up in #1087

PR #1087, opened 2017-01-17
Added sanity check support for OpenSSL 1.1

vanzod:
This fix addresses the issue in #1085 when installing OpenSSL ver. 1.1 due to change in library naming.

### DIFF
--- a/easybuild/easyblocks/o/openssl.py
+++ b/easybuild/easyblocks/o/openssl.py
@@ -63,20 +63,18 @@ class EB_OpenSSL(ConfigureMake):
         if libdir is None:
             raise EasyBuildError("Failed to determine library directory.")
 
+        custom_paths = {
+            'files': [os.path.join(libdir, x) for x in ['libcrypto.a', 'libcrypto.so', 'libssl.a', 'libssl.so']] +
+                     ['bin/openssl'],
+            'dirs': [],
+        }
+
         if LooseVersion(self.version) < LooseVersion("1.1"):
-            custom_paths = {
-                'files': [os.path.join(libdir, x) for x in ['libcrypto.a', 'libcrypto.so', 'libcrypto.so.1.0.0',
-                                                            'libssl.a', 'libssl.so', 'libssl.so.1.0.0']] +
-                         ['bin/openssl'],
-                'dirs': [os.path.join(libdir, 'engines')],
-            }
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/o/openssl.py
```diff
@@ -57,15 +59,24 @@ def sanity_check_step(self):
...
+                                                            'libssl.a', 'libssl.so', 'libssl.so.1.0.0']] +
+                         ['bin/openssl'],
+                'dirs': [os.path.join(libdir, 'engines')],
+            }
+        else:
+            custom_paths = {
+                'files': [os.path.join(libdir, x) for x in ['libcrypto.a', 'libcrypto.so', 'libcrypto.so.1.1',
+                                                            'libssl.a', 'libssl.so', 'libssl.so.1.1']] +
+                         ['bin/openssl'],
+                'dirs': [os.path.join(libdir, 'engines-1.1')],
```
boegel: @vanzod I think it makes sense to avoid duplication, by fleshing out the common stuff, something like:

```python
custom_paths = {
    'files': [os.path.join(libdir, x) for x in ['libcrypto.a', 'libcrypto.so', 'libssl.a', 'libssl.so']] +
             ['bin/openssl'],
    'dirs': [],
}

if LooseVersion(self.version) < LooseVersion("1.1"):
    custom_paths['files'].extend([os.path.join(libdir, 'libcrypto.so.1.0.0'), os.path.join(libdir, 'libssl.so.1.0.0')])
    custom_paths['dirs'].append(os.path.join(libdir, 'engines'))
else:
    custom_paths['files'].extend([os.path.join(libdir, 'libcrypto.so.1.1'), os.path.join(libdir, 'libssl.so.1.1')])
    custom_paths['dirs'].append(os.path.join(libdir, 'engines-1.1'))
```

vanzod: Agree

boegel:
looks good now, do you have an OpenSSL 1.1.x easyconfig too @vanzod?

We prefer having an actual test case included in EB too for updates like this...

boegel:
checked this with a bunch of existing OpenSSL 1.0.0* and 1.0.1* easyconfigs, still works, so I'll go ahead and merge this since it doesn't introduce any regressions afaict

@vanzod please open a PR with an easyconfig for OpenSSL 1.1.* so we have a test case for this

PR #1088, opened 2017-01-20
Add ncurses tinfo static library to link path

vanzod:
If statically linking against the ncurses library built with separate terminfo library (i.e. configured with `--with-termlib`), Python building step fails to find required symbols unless the terminfo library is added to the linking paths.

### DIFF
--- a/easybuild/easyblocks/p/python.py
+++ b/easybuild/easyblocks/p/python.py
@@ -98,7 +98,13 @@ class EB_Python(ConfigureMake):
                 ncurses_libdir = get_software_libdir('ncurses')
                 readline_static_lib = os.path.join(libreadline, readline_libdir, 'libreadline.a')
                 ncurses_static_lib = os.path.join(ncurses, ncurses_libdir, 'libncurses.a')
-                readline = "readline readline.c %s %s" % (readline_static_lib, ncurses_static_lib)
+                tinfo_static_lib = os.path.join(ncurses, ncurses_libdir, 'libtinfo.a')
+
+                if os.path.isfile(tinfo_static_lib):
+                    readline = "readline readline.c %s %s %s" % (readline_static_lib, ncurses_static_lib, tinfo_static_lib)
+                else:
+                    readline = "readline readline.c %s %s" % (readline_static_lib, ncurses_static_lib)
+
                 for line in fileinput.input(modules_setup_dist, inplace='1', backup='.readline'):
                     line = re.sub(r"^#readline readline.c.*", readline, line)
                     sys.stdout.write(line)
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/p/python.py
```diff
@@ -98,7 +98,13 @@ def configure_step(self):
                 ncurses_libdir = get_software_libdir('ncurses')
                 readline_static_lib = os.path.join(libreadline, readline_libdir, 'libreadline.a')
                 ncurses_static_lib = os.path.join(ncurses, ncurses_libdir, 'libncurses.a')
-                readline = "readline readline.c %s %s" % (readline_static_lib, ncurses_static_lib)
+                tinfo_static_lib = os.path.join(ncurses, ncurses_libdir, 'libtinfo.a')
+
+                if os.path.isfile(tinfo_static_lib):
+                    readline = "readline readline.c %s %s %s" % (readline_static_lib, ncurses_static_lib, tinfo_static_lib)
+                else:
+                    readline = "readline readline.c %s %s" % (readline_static_lib, ncurses_static_lib)
```
boegel: @vanzod I'd prefer avoiding to copy-paste as much as possible, how about this:

```python
ncurses_static_libs = os.path.join(ncurses, ncurses_libdir, 'libncurses.a')
tinfo_static_lib = os.path.join(ncurses, ncurses_libdir, 'libtinfo.a')
# if libtinfo.a exists, we also need to link it in to resolve all ncurses symbols
if os.path.exists(tinfo_static_lib):
    ncurses_static_libs += ' ' + tinfo_static_lib
readline = "readline readline.c %s %s" % (readline_static_lib, ncurses_static_libs)
```

vanzod:
@boegel Do not merge this PR yet since there are still issues with the terminfo library.

vanzod:
Closing as issue has been solved. See https://github.com/hpcugent/easybuild-easyconfigs/issues/4049

boegel:
@pforai Just mentioned to me that he was hitting the problem being fixed here after enabling the use of ` --add-dummy-to-minimal-toolchains` with `--minimal-toolchains`.

Due to this, `ncurses-6.0.eb` is being picked up, where `libtinfo.a` is split off, as opposed to in the other `ncurses` easyconfigs.

So maybe this change should be applied after all?

@vanzod Did you run into other problems with the Python you obtained with this patch applied? Which other issues with the terminfo library made you put this on hold?

vanzod:
No, this fix does not solve the problem either since I still got an error because the `libtinfo.a` is not position independent despite `-fPIC`. See my Jan 23 comment in https://github.com/easybuilders/easybuild-easyconfigs/issues/4049.

ocaisa:
Hopefully this is solved with https://github.com/easybuilders/easybuild-easyconfigs/pull/5067

PR #1089, opened 2017-01-23
take 'unpack_sources' also into account for extensions installed with PythonPackage

boegel:
cfr. https://github.com/hpcugent/easybuild-easyconfigs/pull/4057/files#r97373212


boegel:
This change also requires that the `PythonPackage`-specific `unpack_sources` easyconfig parameter is actually passed down to extensions in a `Bundle`...

Cfr. https://github.com/hpcugent/easybuild-easyconfigs/pull/4057#issuecomment-276328396

boegel:
superseded by #1166

PR #1092, opened 2017-01-31
(correctly) define $ROSETTA3_DB in Rosetta easyblock

boegel:
`$ROSETTA_DB` was being incorrectly defined in the Rosetta easyconfigs (cfr. https://github.com/hpcugent/easybuild-easyconfigs/pull/4083)

### DIFF
--- a/easybuild/easyblocks/r/rosetta.py
+++ b/easybuild/easyblocks/r/rosetta.py
@@ -71,7 +71,7 @@ class EB_Rosetta(EasyBlock):
             self.srcdir = os.path.join(prefix, 'rosetta_source')
             if not os.path.exists(self.srcdir):
                 self.srcdir = os.path.join(prefix, 'main', 'source')
-            if not os.path.exists(self.srcdir): 
+            if not os.path.exists(self.srcdir):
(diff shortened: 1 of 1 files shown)


damianam:
@boegel your changes lgtm, even though I think the title of the PR is misleading (you are setting `ROSETTA3_DB` not `ROSETTA_DB`). But I have 2 comments regarding old pieces of code that you might want to fix. Since I can't add line comments on lines that haven't been changed on the PR, I'll write them here:

* Line 70: Strictly speaking, that can be wrong. `len(subdirs)` could be `0`, and the error would be misleading. I doubt you'll ever get to that though.

* Method `detect_cxx`: Shouldn't you be checking for `CXX_SEQ` or `CXX`, rather than `CC_SEQ` or `CC`? If for whatever reason that is correct, I'd suggest to rename the method to `detect_cc`, or at least add a comment to clarify it.

boegel:
@damianam thanks for the thorough review! Remarks fixed, please rereview?

damianam:
@boegel lgtm

boegel:
tested with https://github.com/hpcugent/easybuild-easyconfigs/pull/4083 and https://github.com/hpcugent/easybuild-easyconfigs/pull/4085, good to go, thanks @damianam!

Issue #1090, opened 2017-01-24
EasyBuild fails to build on SLES11 SP3 LTS, due to changes in python's setuptools


dominikborkowski:
Sometime after November 2016, our SLES11 SP3 LTS systems no longer could upgrade or install any EasyBuild past version 2.5.0 (or 2.6.0). In fact, prior to that date, we were able to upgrade to 2.9.0 and then to 3.0.0 just fine. In January trying to repeat the process with 3.0.2, or forcing to reinstall any versions past 2.5.0 failed.

Further investigation showed that the systems in question were also unable to bootstrap clean EasyBuild: there were no errors, even in debug mode. Stage 1 was prepared in /tmp, but it was never copied to the prefix dir.

There have been no changes on the systems, so signs were pointing to something outside. Closer investigation indicated that setuptools may be the culprit. SLES 11 SP3 provides the following:

```
~> rpm -qa --last python python-setuptools
python-setuptools-0.6c11-6.1                  Tue 01 Mar 2016 12:56:13 PM EST
python-2.6.9-0.33.1                           Mon 30 Mar 2015 10:42:33 AM EDT
```

Seems that 0.6c11 was not meeting the evaluation criteria for being greater than 0.6, so new version was always downloaded for each iteration of EasyBuild. For example:

```
> grep EASY-INSTALL-ENTRY-SCRIPT EasyBuild/*/bin/easy_install
EasyBuild/2.2.0/bin/easy_install:# EASY-INSTALL-ENTRY-SCRIPT: 'setuptools==0.6c8','console_scripts','easy_install'
EasyBuild/2.3.0/bin/easy_install:# EASY-INSTALL-ENTRY-SCRIPT: 'distribute==0.6.34','console_scripts','easy_install'
EasyBuild/2.5.0/bin/easy_install:# EASY-INSTALL-ENTRY-SCRIPT: 'setuptools==0.6c8','console_scripts','easy_install'
EasyBuild/2.6.0/bin/easy_install:# EASY-INSTALL-ENTRY-SCRIPT: 'setuptools==0.6c8','console_scripts','easy_install'
EasyBuild/2.7.0/bin/easy_install:# EASY-INSTALL-ENTRY-SCRIPT: 'setuptools==20.6.7','console_scripts','easy_install'
EasyBuild/2.8.0/bin/easy_install:# EASY-INSTALL-ENTRY-SCRIPT: 'setuptools==21.2.1','console_scripts','easy_install'
[...]

PR #1093, opened 2017-02-01
also consider setuptools in EasyBuildMeta easyblock

boegel:
also consider `setuptools` in case it is listed as one of the sources in the `EasyBuild` easyconfig file

This is required to work around the problem reported in #1090

### DIFF
--- a/easybuild/easyblocks/e/easybuildmeta.py
+++ b/easybuild/easyblocks/e/easybuildmeta.py
@@ -61,6 +61,8 @@ class EB_EasyBuildMeta(PythonPackage):
             # when installing the easybuild-* packages, the vsc-base version in easy-install.pth may be 'bumped'
             # if a newer vsc-base version is found somewhere (e.g. provided by the OS)
             self.easybuild_pkgs.extend(['vsc-base', 'vsc-install'])
+            # consider setuptools first, in case it is listed as a sources
+            self.easybuild_pkgs.insert(0, 'setuptools')
 
     def check_readiness_step(self):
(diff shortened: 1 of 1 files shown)


damianam:
@boegel I find it confusing that you add packages in `__init__`, even if they aren't included as sources, and then, if they are not in the sources, just ignore them. Wouldn't it be more clear (and cleaner) if `setuptools` (and possibly `vsc-install` and `vsc-base`, but I don't fully understand why you need them twice) is added to `easybuild_pkgs` if, and just if, it is listed in `self.src`?

boegel:
@damianam I guess we could do that too, yeah. Do you want me to make that change now though? I'd like to avoid spending too much time on this now... Making changes to the `EasyBuildMeta` easyblock is a bit tricky, since it also affects the bootstrap script, etc.

W.r.t. `vsc-base` and `vsc-install`: this is only because those may be picked up from the OS, if a more recent `vsc-base` or `vsc-install` is available there. That mostly a concern for HPC-UGent/VSC systems, other systems most likely don't have these installed system-wide.

damianam:
@boegel: "jain". Doing it now or later I think it doesn't matter. But I think it should be done. To fix https://github.com/hpcugent/easybuild-easyblocks/issues/1090 and don't delay `3.1.0`, I guess you can keep it as it is now, after all it doesn't affect the functionality, but open an issue with target `3.1.1` or `3.2.0` to don't forget about it later.

boegel:
@damianam issue opened at https://github.com/hpcugent/easybuild-easyblocks/issues/1095, thanks for the review!

PR #1096, opened 2017-02-02
changed the sanity check for 2016b since the directory structure has …

mboisson:
MCR R2016b fails to build only because the sanity checks are no longer valid. I added a condition to have the right sanity checks for this version specifically.

### DIFF
--- a/easybuild/easyblocks/m/mcr.py
+++ b/easybuild/easyblocks/m/mcr.py
@@ -119,16 +119,17 @@ class EB_MCR(EasyBlock):
 
     def sanity_check_step(self):
         """Custom sanity check for MCR."""
-        if self.version == "2016b":
-            custom_paths = {
-                'files': [],
-                'dirs': [os.path.join(self.subdir, x, 'glnxa64') for x in ['bin', 'cefclient/sys/os']],
-            }
+        custom_paths = {
+            'files': [],
+            'dirs': [os.path.join(self.subdir, 'bin', 'glnxa64')],
+        }
+        if LooseVersion(self.version) >= LooseVersion('2016b'):
+            custom_paths.append(os.path.join(self.subdir, 'cefclient', 'sys', 'os', 'glnxa64'))
         else:
-            custom_paths = {
-                'files': [],
-                'dirs': [os.path.join(self.subdir, x, 'glnxa64') for x in ['runtime', 'bin', 'sys/os']],
-            }
+            custom_paths.extend([
+                os.path.join(self.subdir, 'runtime', 'glnxa64'),
+                os.path.join(self.subdir, 'sys', 'os', 'glnxa64'),
+            ])
         super(EB_MCR, self).sanity_check_step(custom_paths=custom_paths)
 
     def make_module_extra(self):
--- a/easybuild/easyblocks/m/mcr.py
+++ b/easybuild/easyblocks/m/mcr.py
@@ -71,6 +71,7 @@ class EB_MCR(EasyBlock):
         if LooseVersion(self.version) < LooseVersion('R2015a'):
             shutil.copyfile(os.path.join(self.cfg['start_dir'], 'installer_input.txt'), configfile)
             config = read_file(configfile)
+            # compile regex first since re.sub doesn't accept re.M flag for multiline regex in Python 2.6
             regdest = re.compile(r"^# destinationFolder=.*", re.M)
             regagree = re.compile(r"^# agreeToLicense=.*", re.M)
             regmode = re.compile(r"^# mode=.*", re.M)
--- a/easybuild/easyblocks/m/mcr.py
+++ b/easybuild/easyblocks/m/mcr.py
@@ -119,10 +119,16 @@ class EB_MCR(EasyBlock):
 
     def sanity_check_step(self):
         """Custom sanity check for MCR."""
-        custom_paths = {
-            'files': [],
-            'dirs': [os.path.join(self.subdir, x, 'glnxa64') for x in ['runtime', 'bin', 'sys/os']],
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/m/mcr.py
```diff
@@ -119,10 +119,16 @@ def install_step(self):
...
+        if self.version == "2016b":
+            custom_paths = {
+                'files': [],
+                'dirs': [os.path.join(self.subdir, x, 'glnxa64') for x in ['bin', 'cefclient/sys/os']],
+            }
+        else:
+            custom_paths = {
+                'files': [],
+                'dirs': [os.path.join(self.subdir, x, 'glnxa64') for x in ['runtime', 'bin', 'sys/os']],
+            }
```
boegel: @mboisson it's cleaner to only include the version-specific stuff in the if/else bodies:

```python
custom_paths = {
    'files': [],
    'dirs': [os.path.join(self.subdir, 'bin', 'glnxa64')],
}
if LooseVersion(self.version) >= LooseVersion('2016b'):
    custom_paths.append(os.path.join(self.subdir, 'cefclient', 'sys', 'os', 'glnxa64'))
else:
    custom_paths.extend([
        os.path.join(self.subdir, 'runtime, 'glnxa64'),
        os.path.join(self.subdir, 'sys', 'os', 'glnxa64'),
    ])
```

easybuild/easyblocks/m/mcr.py
```diff
@@ -119,10 +119,16 @@ def install_step(self):
 
     def sanity_check_step(self):
         """Custom sanity check for MCR."""
-        custom_paths = {
-            'files': [],
-            'dirs': [os.path.join(self.subdir, x, 'glnxa64') for x in ['runtime', 'bin', 'sys/os']],
-        }
+        if self.version == "2016b":
```
boegel: better make this future proof, with:

```python
if LooseVersion(self.version) >= LooseVersion('2016b'):
```

boegel:
@mboisson please change the target branch to `develop` (you can do this for open PRs now via the 'Edit' button)

mboisson:
I think I updated the pull request correctly (although I'm not sure how it happened) ? The new commit shows in the history in any case.

boegel:
@mboisson looks good, thanks; you did the right thing, i.e. add a new commit to the branch you used for the PR and then push it to GitHub, which updates the PR

Can you also open a PR for an MCR easyconfig that requires this change?

mboisson:
Done. I included a couple more MCR easyconfig files as well as one Java, since MCR depends on Java and we installed an up-to-date version of Java in our software stack.

bartoldeman:
This change is buggy. Instead of ```custom_paths.append``` it should use ```custom_paths['dirs'].append```.
Maxime, can you push a change? I am changing it for ComputeCanada now.

boegel:
Urgh, that's my bad, serves me right for making suggestions without actually testing it myself... :)

Thanks for the notification @bartoldeman, I knew I did the right thing by not merging this into EB v3.1.0 last-minute ;)

bartoldeman:
There is another issue with the LooseVersion check. It should check for R2015a etc. instead of 2015a. I am debugging that now, as the code for < 2015a seems not to work.

mboisson:
Done, I think.

mboisson:
On a side note, I really should have made this PR from a custom branch....

bartoldeman:
I fixed the LooseVersion and regexp stuff in the top 2 commits here:
https://github.com/ComputeCanada/easybuild-easyblocks/commits/computecanada-master

mboisson:
I added Bart's changes to this pull request.

easybuild/easyblocks/m/mcr.py
```diff
@@ -68,12 +68,16 @@ def configure_step(self):
...
-            config = re.sub(r"^# destinationFolder=.*", "destinationFolder=%s" % self.installdir, config, re.M)
-            config = re.sub(r"^# agreeToLicense=.*", "agreeToLicense=Yes", config, re.M)
-            config = re.sub(r"^# mode=.*", "mode=silent", config, re.M)
+            regdest = re.compile(r"^# destinationFolder=.*", re.M)
+            regagree = re.compile(r"^# agreeToLicense=.*", re.M)
+            regmode = re.compile(r"^# mode=.*", re.M)
+
+            config = regdest.sub("destinationFolder=%s" % self.installdir, config)
+            config = regagree.sub("agreeToLicense=Yes", config)
+            config = regmode.sub("mode=silent", config)
```
boegel: @bartoldeman why was this changed? compiling a regex first before using it only makes sense if you reuse the compiled regex's multiple times, which is not the case here?

bartoldeman: because of re.M (=re.MULTILINE)
In python2.7 you can do  ```re.sub(regexp,subst,string,flags=re.M)``` but this is impossible in python2.6. The 4th argument of ```re.sub``` is a count so for that re.M is bogus.

boegel: Hah, OK, I wasn't aware of that...

We should include a comment to clarify that, can you add something like this @mboisson?

```python
# compile regex first since re.sub doesn't accept re.M flag for multiline regex in Python 2.6
```

boegel:
I tested this updated MCR easyblock with all existing MCR easyconfigs, and the ones in https://github.com/hpcugent/easybuild-easyconfigs/pull/4098, so good to go when the clarifying comment is included.

mboisson:
Done

boegel:
Good to go, thanks @mboisson and @bartoldeman!

PR #1102, opened 2017-02-09
disable_quad_on_power

d-rdgz:
disables attempts to build libquadmath
versions of FFTW on Power


easybuild/easyblocks/f/fftw_no_quad_on_power.patch
```diff
@@ -0,0 +1,36 @@
+--- easybuild/easyblocks/f/fftw.py.orig	2017-02-08 12:53:50.207423000 -0600
```
boegel: @d-rdgz please remove this file?

```
git rm easybuild/easyblocks/f/fftw_no_quad_on_power.patch
git commit -m "remove patch file for FFTW easyblock"
git push ...
```

easybuild/easyblocks/f/fftw.py
```diff
@@ -41,7 +41,12 @@
 # Altivec (POWER), SSE (x86), NEON (ARM), FMA (x86_64)
 # asimd is CPU feature for extended NEON on AARCH64
 FFTW_CPU_FEATURE_FLAGS = FFTW_CPU_FEATURE_FLAGS_SINGLE_DOUBLE + ['altivec', 'asimd', 'neon', 'sse']
-FFTW_PRECISION_FLAGS = ['single', 'double', 'long-double', 'quad-precision']
+
+cpu_arch = get_cpu_architecture()
```
boegel: @d-rdgz I'm not sure it is a good idea to call `get_cpu_architecture()` when this easyblock is imported...

Also, it seems like overkill, just doing the `continue` on `quad` + `POWER` that you do before is sufficient?

boegel: Ah, no, that's in the sanity check...

I think it's better to leave all 4 precision 'types' to be listed in `FFTW_PRECISION_FLAGS`, and to hard set the value of `self.cfg['with_quad_prec']` to `False` on POWER, with a decent log message to go along with it (if it was set to `True` at that time)

easybuild/easyblocks/f/fftw.py
```diff
@@ -179,7 +184,10 @@ def sanity_check_step(self):
         }
 
         extra_files = []
+        cpu_arch = get_cpu_architecture()
         for (prec, letter) in [('double', ''), ('long_double', 'l'), ('quad', 'q'), ('single', 'f')]:
+            if prec == 'quad' and cpu_arch == POWER:
+                continue
```
boegel: this checking should be done earlier, and `with_quad_prec` should be hard set to `False` instead, i.e. in `run_all_steps` before the `for prec in FFTW_PRECISION_FLAGS:` loop...

geimer:
With GCC 6, one *may* be able to get quad-precision to work using the `-mfloat128` command-line switch (untested), see https://gcc.gnu.org/onlinedocs/gcc-6.3.0/gcc/RS_002f6000-and-PowerPC-Options.html#RS_002f6000-and-PowerPC-Options. And while at it, ARM doesn't support quad-precision either.

JackPerdue:
Ah, man, Kenneth, c'mon.... you know my Python skills are marginally better than my git skills (and big thanks from me to Dylan for getting us here with the git bits).

On a related note (since somebody needs to rewrite this anyway),  I've been struggling with 3.1.0 and the new EB_FFTW easyblock when it comes to quad while using newer Intel compilers.   Looking over all the FFTW .eb's before EB3.1.0, I noticed that there was a trend recently to exclude the quad build for Intel compilers, but include it for the GCC compilers.

I can telll you, the current (3.1.0) logic is not happy building quad with Intel compilers

I ended up using a 3.0.9-style easyconfig (using ConfigureMake, where I still have some control).

So I think there is a bit more work here.   And don't ask Dylan..... :)   He only does the git stuff at this point (he just started at the help desk).

geimer:
@JackPerdue You still have full control using the various `with_foo_prec` custom easyconfig parameters. And if I see it right, this is already used in Intel easyconfigs for FFTW to disable quad-precision (e.g., https://github.com/hpcugent/easybuild-easyconfigs/blob/master/easybuild/easyconfigs/f/FFTW/FFTW-3.3.5-intel-2016b.eb).

Also, see https://github.com/d-rdgz/easybuild-easyblocks/pull/1 for an (untested) attempt to address the remarks.

JackPerdue:
Ah... OK... thanks for the clarification... I guess I missed that bit in the latest FFTW intel easyconfig or I didn't find it applicable (in fact I found it confusing since the comment says that it only works with GCC  _GREATER_ than 4.6 and foss-2016b is built with 5.4.0... seems like it used to work but now it doesn't so I wonder if that should be _LESS_ than)..  

I guess I can use the same trick on Power (which is what Kenneth was saying but I didn't quite grasp at the time).

boegel:
@d-rdgz I would like to include these changes in EasyBuild v3.1.1 which I hope to release by the end of the week.

Can you take a look at the changes @geimer applied in https://github.com/d-rdgz/easybuild-easyblocks/pull/1?

boegel:
ping @JackPerdue & @d-rdgz?

If OK by you, we can go forward with this with a wrapper PR using @geimer's forked branch, i.e. including the additional changes in https://github.com/d-rdgz/easybuild-easyblocks/pull/1...

boegel:
@JackPerdue Well, sure, you can disable building with quad precision yourself via `with_quad_prec = False`, but detecting POWER and automagically disabling quad precision is better if you ask me...

@geimer Can you open a new PR to the `develop` branch of the central repo for your `dylan_branch2` branch to get a good view on things? Merging that PR will also merge this one, since the commits included here will be included in your PR too.

boegel:
merged via #1138, thanks for your efforts @d-rdgz!

PR #1104, opened 2017-02-13
Update nwchem easyblock to handle different versions of OpenMPI.

akesandgren:
Also fix missing libs in linking when building with OpenIB as armci
network.

And turn on USE_NOIO by default.

### DIFF
--- a/easybuild/easyblocks/n/nwchem.py
+++ b/easybuild/easyblocks/n/nwchem.py
@@ -178,18 +178,27 @@ class EB_NWChem(ConfigureMake):
         if LooseVersion(self.version) >= LooseVersion("6.5"):
             env.setvar('EACCSD','y') # enable EOM electron-attachemnt coupled cluster capability
             env.setvar('IPCCSD','y') # enable EOM ionization-potential coupled cluster capability
+            env.setvar('USE_NOIO', 'TRUE') # avoid doing I/O for the ddscf, mp2 and ccsd modules
 
         for var in ['USE_MPI', 'USE_MPIF', 'USE_MPIF4']:
             env.setvar(var, 'y')
         for var in ['CC', 'CXX', 'F90']:
             env.setvar('MPI_%s' % var, os.getenv('MPI%s' % var))
+        #if LooseVersion(self.version) <= LooseVersion("6.5"):
         env.setvar('MPI_LOC', os.path.dirname(os.getenv('MPI_INC_DIR')))
         env.setvar('MPI_LIB', os.getenv('MPI_LIB_DIR'))
         env.setvar('MPI_INCLUDE', os.getenv('MPI_INC_DIR'))
         libmpi = None
         mpi_family = self.toolchain.mpi_family()
         if mpi_family in toolchain.OPENMPI:
-            libmpi = "-lmpi_f90 -lmpi_f77 -lmpi -ldl -Wl,--export-dynamic -lnsl -lutil"
+            ompi_ver = get_software_version('OpenMPI')
+            if LooseVersion(ompi_ver) < LooseVersion("1.10"):
+                if LooseVersion(ompi_ver) < LooseVersion("1.8"):
+                    libmpi = "-lmpi_f90 -lmpi_f77 -lmpi -ldl -Wl,--export-dynamic -lnsl -lutil"
+                else:
+                    libmpi = "-lmpi_usempi -lmpi_mpifh -lmpi"
+            else:
+                libmpi = "-lmpi_usempif08 -lmpi_usempi_ignore_tkr -lmpi_mpifh -lmpi"
         elif mpi_family in [toolchain.INTELMPI]:
             if self.cfg['armci_network'] in ["MPI-MT"]:
                 libmpi = "-lmpigf -lmpigi -lmpi_ilp64 -lmpi_mt"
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/n/nwchem.py
```diff
@@ -178,18 +178,27 @@ def configure_step(self):
         if LooseVersion(self.version) >= LooseVersion("6.5"):
             env.setvar('EACCSD','y') # enable EOM electron-attachemnt coupled cluster capability
             env.setvar('IPCCSD','y') # enable EOM ionization-potential coupled cluster capability
+            env.setvar('USE_NOIO', 'TRUE') # avoid doing I/O for the ddscf, mp2 and ccsd modules
 
         for var in ['USE_MPI', 'USE_MPIF', 'USE_MPIF4']:
             env.setvar(var, 'y')
         for var in ['CC', 'CXX', 'F90']:
             env.setvar('MPI_%s' % var, os.getenv('MPI%s' % var))
+        #if LooseVersion(self.version) <= LooseVersion("6.5"):
```
boegel: please remove?

easybuild/easyblocks/n/nwchem.py
```diff
@@ -199,6 +208,9 @@ def configure_step(self):
             libmpi = "-lmpichf90 -lmpich -lopa -lmpl -lrt -lpthread"
         else:
             raise EasyBuildError("Don't know how to set LIBMPI for %s", mpi_family)
+        if self.cfg['armci_network'] in ["OPENIB"]:
+            libmpi += " -libumad -libverbs -lpthread"
+        #if LooseVersion(self.version) <= LooseVersion("6.5"):
```
boegel: please remove?

easybuild/easyblocks/n/nwchem.py
```diff
@@ -178,18 +178,27 @@ def configure_step(self):
...
         if mpi_family in toolchain.OPENMPI:
-            libmpi = "-lmpi_f90 -lmpi_f77 -lmpi -ldl -Wl,--export-dynamic -lnsl -lutil"
+            ompi_ver = get_software_version('OpenMPI')
+            if LooseVersion(ompi_ver) < LooseVersion("1.10"):
+                if LooseVersion(ompi_ver) < LooseVersion("1.8"):
+                    libmpi = "-lmpi_f90 -lmpi_f77 -lmpi -ldl -Wl,--export-dynamic -lnsl -lutil"
+                else:
+                    libmpi = "-lmpi_usempi -lmpi_mpifh -lmpi"
+            else:
+                libmpi = "-lmpi_usempif08 -lmpi_usempi_ignore_tkr -lmpi_mpifh -lmpi"
```
boegel: oh boy, what a mess :)

do you have additional easyconfigs to test these changes @akesandgren?

akesandgren: Yes i do have an easyconfig that need that fix, i.e. using a iomkl/2017.01 which uses OpenMPI/2.0.1

It dies horribly for us when using intel/2017.x so openmpi was the way to go.

easybuild/easyblocks/n/nwchem.py
akesandgren: That #if Looseversion (both of them) should probably be activated but i haven't had time to verify that it works if i do...

easybuild/easyblocks/n/nwchem.py
boegel: can you open a pull request for that?

easybuild/easyblocks/n/nwchem.py
boegel: so, `$LIBMPI` doesn't need to be set anymore for NWChem 6.6 and newer?

akesandgren: Thats what their docco says, unfortunately that probably destroys the ARMCI fix, so i haven't taken the time to test that.

easybuild/easyblocks/n/nwchem.py
akesandgren: Will do...

akesandgren: I accidentally tried to build my new nwchem 6.6 using iomkl/2017a (openmpi 2.0.2) with the old nwchem.py and it subsequently failed. Using this new nwchem.py worked though, so it got "properly" tested ...

easybuild/easyblocks/n/nwchem.py
boegel: Do you have a reference to the documentation?

boegel: confirmed, no need to set these environment variables anymore with NWChem 6.6, see https://github.com/akesandgren/easybuild-easyblocks/pull/3

boegel:
lgtm, I'll retest with existing NWChem easyconfigs and https://github.com/hpcugent/easybuild-easyconfigs/issues/4231, but should be good to go, thanks @akesandgren!

easybuild/easyblocks/n/nwchem.py
```diff
@@ -173,33 +173,54 @@ def configure_step(self):
...
-            if self.cfg['armci_network'] in ["MPI-MT"]:
-                libmpi = "-lmpigf -lmpigi -lmpi_ilp64 -lmpi_mt"
+
+        libmpi = ""
+
+        # for NWChem 6.6 and newer, $LIBMPI & co should no longer be
+        # set, the correct values are determined by the NWChem build
+        # procedure automatically, see
+        # http://www.nwchem-sw.org/index.php/Compiling_NWChem#MPI_variables
+        if LooseVersion(self.version) <= LooseVersion("6.5"):
```
boegel: @akesandgren We'll need to make this "< 6.6", since otherwise the build of NWChem version 6.5.revision26243 is broken:

```
>>> from distutils.version import LooseVersion
>>> LooseVersion('6.5.revision26243') <= LooseVersion('6.5')
False
>>> LooseVersion('6.5.revision26243') < LooseVersion('6.6')
True
```

akesandgren:
test changed to < 6.6

boegel:
verified with existing NWChem easyconfigs and https://github.com/hpcugent/easybuild-easyconfigs/pull/4231, good to go

Thanks @akesandgren!

PR #1105, opened 2017-02-13
Add an easyblock for Siesta.

akesandgren:
This is an easyblock for Siesta 4.0 and later


akesandgren:
@boegel If you could get this one into 3.1.1 i'd appreciate it.

boegel:
@akesandgren since I intended to release EB v3.1.1 last Friday already, this won't make it in anymore, I'm sorry.

Someone should review & test this; do you have any easyconfig files for this too?

Moreover, as you point out, this and #923 should be merged somehow... Maybe @migueldiascosta is up for taking a look at your pull request, and seeing whether both approaches align?

migueldiascosta:
@boegel @akesandgren the problem with this (and my #923) is that it will only work with 4.0 versions - 4.1-b1 broke the configure script (although it's trivial to fix it in the easyblock) and 4.1-b2 removed it completely - from upstream, "We have discontinued support for the configure script due to missing updates and dependency checks. Hence, later revisions have removed configure" (to be fair, they have always advocated tweaking the arch.make file instead of configure).

From the manual of 4.1-b2, "You are strongly encouraged to look at Obj/DOCUMENTED-TEMPLATE.make for information about the fine points of the arch.make file. There are two sample make files for compilation of SIESTA with gfortran and ifort named Obj/gfortran.make and Obj/intel.make, respectively. Please use those as guidelines for creating the final arch.make."

For version 4.1-b2, I used [this](https://gist.github.com/migueldiascosta/78f6fbf6123d2de399d60884158732bf). As mentioned in #923, I had decided to wait for a stable version before changing that PR, but it hasn't been released yet.

So, I guess the first question is, do we want to support the versions where a configure script still exists and works? Since Siesta hasn't been supported so far, we could simply start with 4.1-b2...

The second question is, regardless of the first, what's the best way of supporting the new versions without the configure script? I don't really like all those sed commands in my easyblock...

akesandgren:
Ohh..... that's not something i have looked at. Since we have Siesta 4.0 installled there is no reason for us to rush this. I won't have time to look at next ver of siesta for quite a while. (Still lots of sw to install and new HW to experiment with). So from my point of view, waiting for an official 4.1 is early enough before starting to look at how to build it.

boegel:
@migueldiascosta Is the `arch.make` procedure also supported for versions earlier than 4.1? If so, why not use that approach for all versions?

We have a dedicated function to do "runtime patching", so you don't have to use `sed`, i.e. `apply_regex_substitutions`, see for example https://github.com/hpcugent/easybuild-easyblocks/blob/master/easybuild/easyblocks/b/binutils.py#L88

migueldiascosta:
@boegel we would still need to patch different example 'arch.make' files depending on the version (hopefully the new ones won't change much...), but yes, we could

regarding `apply_regex_substitutions`, of course, there had to be a better way (and now I think you had already mentioned this to me and I forgot)

migueldiascosta:
@boegel in the older versions `configure` needs to be run anyway, even if `arch.make` is then adjusted/replaced, so might as well keep it

@akesandgren something like this should make your easyblock also work for the newer versions without `configure`

``` python

start_dir = self.cfg['start_dir']
obj_dir = os.path.join(start_dir, 'Obj')
arch_make = os.path.join(obj_dir, 'arch.make')

netcdff_loc = get_software_root('NetCDF-Fortran')

regex_subs = [
    ('dc_lapack.a', ''),
    ('libsiestaBLAS.a', ''),
    ('libsiestaLAPACK.a', ''),
]

if LooseVersion(self.version) < LooseVersion("4.1-b2"):

    # MPI?
        self.cfg.update('configopts', '--enable-mpi')

    # BLAS and LAPACK
    self.cfg.update('configopts', '--with-blas="$LIBBLAS"')
    self.cfg.update('configopts', '--with-lapack="$LIBLAPACK"')

    # ScaLAPACK (and BLACS)
    self.cfg.update('configopts', '--with-scalapack="$LIBSCALAPACK"')
    self.cfg.update('configopts', '--with-blacs="$LIBSCALAPACK"')

    # NetCDF-Fortran
    if netcdff_loc:
        self.cfg.update('configopts', '--with-netcdf=-lnetcdff')

    super(EB_Siesta, self).configure_step(cmd_prefix='../Src/')

    apply_regex_substitutions(arch_make, regex_subs)

else:  # there's no configure on newer versions

    if self.toolchain.comp_family() in [toolchain.INTELCOMP]:
        copy_file(os.path.join(obj_dir, 'intel.make'), arch_make)
    else:
        copy_file(os.path.join(obj_dir, 'gfortran.make'), arch_make)

        regex_subs.extend([
            (r"^(CC\s*=\s*).*$", r"\1%s" % os.getenv('MPICC')),
            (r"^(FC\s*=\s*).*$", r"\1%s" % os.getenv('MPIF90')),
            (r"^(FPPFLAGS\s*=.*)$", r"\1 -DMPI"),
            (r"^(FPPFLAGS\s*=.*)$", r"\1\nMPI_INTERFACE = libmpi_f90.a\nMPI_INCLUDE = ."),
        ])
        complibs = os.environ['LIBSCALAPACK']
    else:
        complibs = os.environ['LIBLAPACK']

    regex_subs.extend([
        (r"^(LIBS\s*=\s).*$", r"\1 %s" % complibs),
    ])

    if netcdff_loc:
        regex_subs.extend([
            (r"^(LIBS\s*=.*)$", r"\1 -lnetcdff"),
            (r"^(FPPFLAGS\s*=.*)$", r"\1 -DCDF"),
        ])

    apply_regex_substitutions(arch_make, regex_subs)
```

akesandgren:
I'll come up with a new easyblock for this that covers all the bases. Might take a day or so due to other stuff...

akesandgren:
@migueldiascosta How does this look?

easybuild/easyblocks/s/siesta.py
```diff
@@ -0,0 +1,308 @@
...
+
+        if self.cfg['with_utils']:
+            bins.extend(['bin/denchar'])
+
+        custom_paths = {
+            'files': bins,
+            'dirs': [],
+        }
+
+        super(EB_Siesta, self).sanity_check_step(custom_paths=custom_paths)
```
migueldiascosta: yesterday I had only tested with an `intel` toolchain, today checked with `foss`

[this](https://gist.github.com/migueldiascosta/0fbe52345abe7d59eb62a0cbab909264#file-siesta-py-patch) is what I changed to get all versions to work on both toolchains

akesandgren:
This last set of changes fixes everything @migueldiascosta has mentioned (except the Src/fdict/Makefile since that isn't needed as far as i can see).

It builds 4.0-b1/b2, 4.0, 4.1-b1/b2 with foss/2017a and intel/2017a.
Does parallel and openmp builds for 4.1 versions, and a bunch och other minor fixes.

It can probably be made a bit cleaner, (esp if one only cares about full release versions) but i think it is good to go.

I'll update the corresponding easyconfig file for Siesta too.

migueldiascosta:
@akesandgren you're right, everything's working now. Thanks!

akesandgren:
@boegel Care to merge this?

easybuild/easyblocks/s/siesta.py
```diff
@@ -0,0 +1,342 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with EasyBuild.  If not, see <http://www.gnu.org/licenses/>.
+##
+"""
+EasyBuild support for building and installing Siesta, implemented as an easyblock
+
+@author: Ake Sandgren (Umea University)
```
boegel: missing @migueldiascosta?

easybuild/easyblocks/s/siesta.py
```diff
@@ -0,0 +1,342 @@
+##
+# Copyright 2009-2016 Ghent University
```
boegel: 2017

easybuild/easyblocks/s/siesta.py
```diff
@@ -0,0 +1,342 @@
...
+# You should have received a copy of the GNU General Public License
+# along with EasyBuild.  If not, see <http://www.gnu.org/licenses/>.
+##
+"""
+EasyBuild support for building and installing Siesta, implemented as an easyblock
+
+@author: Ake Sandgren (Umea University)
+"""
+import os
+import re
```
boegel: not used

easybuild/easyblocks/s/siesta.py
```diff
@@ -0,0 +1,342 @@
...
+                    (r"^(LIBS\s*=.*)$", r"\1 $(NETCDF_LIBS)"),
+                    (r"^(FPPFLAGS\s*=.*)$", r"\1 -DCDF"),
+                ])
+
+        apply_regex_substitutions(arch_make, regex_subs)
+
+        run_cmd('make %s' % par, log_all=True, simple=True, log_output=True)
+
+        # Put binary in temporary install dir
+        shutil.copy(os.path.join(self.cfg['start_dir'], 'Obj', 'siesta'),
```
boegel: please use `copy_file` from `filetools`

easybuild/easyblocks/s/siesta.py
```diff
@@ -0,0 +1,342 @@
...
+            if LooseVersion(self.version) >= LooseVersion("4.1"):
+                expected_utils.extend([
+                    'Bands/gnubands',
+                    'Grimme/fdf2grimme',
+                    'SpPivot/pvtsp',
+                    'TS/ts2ts/ts2ts', 'TS/tshs2tshs/tshs2tshs', 'TS/TBtrans/tbtrans',
+                ])
+
+            for f in expected_utils:
+                shutil.copy(os.path.join(self.cfg['start_dir'], 'Util', f), bindir)
```
boegel: use `copy_files` from `filetools`?

easybuild/easyblocks/s/siesta.py
```diff
@@ -0,0 +1,342 @@
...
+            # Build transiesta
+            try:
+                os.chdir(obj_dir)
+            except OSError, err:
+                raise EasyBuildError("Failed to change back to Obj dir: %s", err)
+
+            run_cmd('make clean', log_all=True, simple=True, log_output=True)
+            run_cmd('make %s transiesta' % par, log_all=True, simple=True, log_output=True)
+
+            shutil.copy(os.path.join(self.cfg['start_dir'], 'Obj', 'transiesta'),
```
boegel: use `copy_dir`?

easybuild/easyblocks/s/siesta.py
```diff
@@ -0,0 +1,342 @@
...
+        """No build step for Siesta."""
+        pass
+
+    def install_step(self):
+        """Custom install procedure for Siesta."""
+
+        try:
+            # binary
+            bindir = os.path.join(self.installdir, 'bin')
+            shutil.copytree(os.path.join(self.cfg['start_dir'], 'bin'), bindir)
```
boegel: `copy_dir`?

easybuild/easyblocks/s/siesta.py
```diff
@@ -0,0 +1,342 @@
...
+                        bindir)
+
+    def build_step(self):
+        """No build step for Siesta."""
+        pass
+
+    def install_step(self):
+        """Custom install procedure for Siesta."""
+
+        try:
```
boegel: no need for the `try` anymore if you use `copy_*` functions, they do proper error reporting already

easybuild/easyblocks/s/siesta.py
```diff
@@ -0,0 +1,342 @@
...
+# along with EasyBuild.  If not, see <http://www.gnu.org/licenses/>.
+##
+"""
+EasyBuild support for building and installing Siesta, implemented as an easyblock
+
+@author: Ake Sandgren (Umea University)
+"""
+import os
+import re
+import shutil
```
boegel: not used if you switch to `copy_*` functions (see below)

easybuild/easyblocks/s/siesta.py
```diff
@@ -0,0 +1,342 @@
...
+import stat
+import tempfile
```
boegel: not used

easybuild/easyblocks/s/siesta.py
```diff
@@ -0,0 +1,342 @@
...
+
+import easybuild.tools.config as config
```
boegel: not used?

easybuild/easyblocks/s/siesta.py
```diff
@@ -0,0 +1,342 @@
...
+import easybuild.tools.environment as env
```
boegel: not used?

easybuild/easyblocks/s/siesta.py
```diff
@@ -0,0 +1,342 @@
...
+
+class EB_Siesta(ConfigureMake):
+    """
+    Support for building/installing Siesta.
+    - avoid parallel build, doesn't work
+    """
+
+    def __init__(self, *args, **kwargs):
+        """Initialisation of custom class variables for Siesta."""
+        super(EB_Siesta, self).__init__(*args, **kwargs)
```
boegel: this can be dropped, you're not customising anything

easybuild/easyblocks/s/siesta.py
```diff
@@ -0,0 +1,342 @@
...
+from easybuild.tools.build_log import EasyBuildError
+from easybuild.tools.filetools import adjust_permissions, apply_regex_substitutions, copy_file, mkdir, write_file
+from easybuild.tools.modules import get_software_libdir, get_software_root, get_software_version
+from easybuild.tools.run import run_cmd
+
+
+class EB_Siesta(ConfigureMake):
+    """
+    Support for building/installing Siesta.
+    - avoid parallel build, doesn't work
```
boegel: drop this, since for recent versions it seems to be supported?

easybuild/easyblocks/s/siesta.py
```diff
@@ -0,0 +1,342 @@
...
+import tempfile
+
+import easybuild.tools.config as config
+import easybuild.tools.environment as env
+import easybuild.tools.toolchain as toolchain
+from distutils.version import LooseVersion
+from easybuild.easyblocks.generic.configuremake import ConfigureMake
+from easybuild.framework.easyconfig import CUSTOM
+from easybuild.tools.build_log import EasyBuildError
+from easybuild.tools.filetools import adjust_permissions, apply_regex_substitutions, copy_file, mkdir, write_file
```
boegel: `write_file` is not used?

easybuild/easyblocks/s/siesta.py
```diff
@@ -0,0 +1,342 @@
...
+from easybuild.tools.modules import get_software_libdir, get_software_root, get_software_version
```
boegel: only `get_software_root` is actually used

easybuild/easyblocks/s/siesta.py
```diff
@@ -0,0 +1,342 @@
...
+        netcdff_loc = get_software_root('NetCDF-Fortran')
+
+        par = ''
+        if LooseVersion(self.version) >= LooseVersion("4.1"):
+            par = '-j %s' % self.cfg['parallel']
+
+        # enable OpenMP support if desired
+        openmp = self.toolchain.options.get('openmp', None)
+        if openmp:
+            scalapack = os.environ['LIBSCALAPACK_MT']
```
boegel: you can shorten this quite a bit:

```python
env_var_suff = ''
if openmp:
    env_var_suff = '_MT'

scalapack = os.environ['LIBSCALAPACK' + env_var_suff]
```

easybuild/easyblocks/s/siesta.py
```diff
@@ -0,0 +1,342 @@
...
+            (r"^(LDFLAGS\s*=).*$", r"\1 %s %s" % (os.environ['FCFLAGS'], os.environ['LDFLAGS'])),
+        ]
+
+	if netcdff_loc:
+	    # Needed for gfortran at least
+	    regex_subs.extend([
+		(r"^(ARFLAGS_EXTRA\s*=.*)$", r"\1\nNETCDF_INCFLAGS = -I%s/include" % netcdff_loc),
+	    ])
+
+        if fftw is not None:
```
boegel: just `if fftw:`?

easybuild/easyblocks/s/siesta.py
```diff
@@ -0,0 +1,342 @@
...
+            regex_subs.extend([
+                (r'(FPPFLAGS\s*=.*)$', r'\1\nFFTW_INCFLAGS = -I%s\nFFTW_LIBS = -L%s %s' % (os.environ['FFTW_INC_DIR'], os.environ['FFTW_LIB_DIR'], fftw)),
```
boegel: line too long

```python
fftw_inc, fftw_lib = os.environ['FFTW_INC_DIR'], os.environ['FFTW_LIB_DIR']
(..., ... % (fftw_inc, fftw_lib, fftw))
```

easybuild/easyblocks/s/siesta.py
```diff
@@ -0,0 +1,342 @@
...
+            ])
+
+        # Make a temp installdir during the build of the various parts
+        try:
+            mkdir(bindir)
```
boegel: no need for `try` with `mkdir`, it already does proper error reporting

easybuild/easyblocks/s/siesta.py
```diff
@@ -0,0 +1,342 @@
...
+        except OSError, err:
+            raise EasyBuildError("Failed to create temp installdir %s: %s", bindir, err)
+
+        # change to actual build dir
+        try:
+            os.chdir(obj_dir)
```
boegel: use `change_dir` from `filetools`, no need for `try`

easybuild/easyblocks/s/siesta.py
```diff
@@ -0,0 +1,308 @@
...
+            # ScaLAPACK (and BLACS)
+            self.cfg.update('configopts', '--with-scalapack="%s"' % scalapack)
+            self.cfg.update('configopts', '--with-blacs="$%s"' % blacs)
+
+            # NetCDF-Fortran
+            if netcdff_loc:
+                self.cfg.update('configopts', '--with-netcdf=-lnetcdff')
+
+            super(EB_Siesta, self).configure_step(cmd_prefix='../Src/')
+
```
boegel: that's weird...

maybe include a comment on that above the lines below to make that clear?

easybuild/easyblocks/s/siesta.py
```diff
@@ -0,0 +1,342 @@
...
+                regex_subs_Makefile = [
+                    (r'CFLAGS\)-c', r'CFLAGS) -c'),
+                ]
+                apply_regex_substitutions('Makefile', regex_subs_Makefile)
+
+        else: # there's no configure on newer versions
+
+            if self.toolchain.comp_family() in [toolchain.INTELCOMP]:
+                copy_file(os.path.join(obj_dir, 'intel.make'), arch_make)
+            else:
```
boegel: check for `GCC` with `elif`, raise error on `else` (unknown compiler used)?

easybuild/easyblocks/s/siesta.py
```diff
@@ -0,0 +1,342 @@
...
+        run_cmd('make %s' % par, log_all=True, simple=True, log_output=True)
+
+        # Put binary in temporary install dir
+        shutil.copy(os.path.join(self.cfg['start_dir'], 'Obj', 'siesta'),
+                    bindir)
+
+        if self.cfg['with_utils']:
+            # Make the utils
+            try:
+                os.chdir(os.path.join(start_dir, 'Util'))
```
boegel: use `change_dir`, no need for `try`

easybuild/easyblocks/s/siesta.py
```diff
@@ -0,0 +1,342 @@
...
+
+            if LooseVersion(self.version) >= LooseVersion("4.1"):
+                regex_subs_TS = [
+                    (r"^default:.*$", r""),
+                    (r"^EXE\s*=.*$", r""),
+                    (r"^(include\s*..ARCH_MAKE.*)$", r"EXE=tshs2tshs\ndefault: $(EXE)\n\1"),
+                    (r"^(INCFLAGS.*)$", r"\1 -I%s" % obj_dir),
+                ]
+
+                apply_regex_substitutions(os.path.join(self.cfg['start_dir'], 'Util', 'TS', 'tshs2tshs', 'Makefile'), regex_subs_TS)
```
boegel: line too long

```python
makefile = os.path.join(self.cfg['start_dir'], 'Util', 'TS', 'tshs2tshs', 'Makefile')
apply_regex_substitutions(makefile, regex_subs_TS)
```

easybuild/easyblocks/s/siesta.py
```diff
@@ -0,0 +1,342 @@
...
+
+            # SUFFIX rules in wrong place
+            regex_subs_suffix = [
+                (r'^(\.SUFFIXES:.*)$', r''),
+                (r'^(include\s*\$\(ARCH_MAKE\).*)$', r'\1\n.SUFFIXES:\n.SUFFIXES: .c .f .F .o .a .f90 .F90'),
+            ]
+            apply_regex_substitutions(os.path.join(self.cfg['start_dir'], 'Util', 'Sockets', 'Makefile'), regex_subs_suffix)
+            apply_regex_substitutions(os.path.join(self.cfg['start_dir'], 'Util', 'SiestaSubroutine', 'SimpleTest', 'Src', 'Makefile'), regex_subs_suffix)
```
boegel: line too long...

easybuild/easyblocks/s/siesta.py
```diff
@@ -0,0 +1,342 @@
...
+
+            regex_subs_UtilLDFLAGS = [
+                (r'(\$\(FC\)\s*-o\s)', r'$(FC) %s %s -o ' % (os.environ['FCFLAGS'], os.environ['LDFLAGS'])),
+            ]
+            apply_regex_substitutions(os.path.join(self.cfg['start_dir'], 'Util', 'Optimizer', 'Makefile'), regex_subs_UtilLDFLAGS)
+            apply_regex_substitutions(os.path.join(self.cfg['start_dir'], 'Util', 'JobList', 'Src', 'Makefile'), regex_subs_UtilLDFLAGS)
```
boegel: lines too long...

easybuild/easyblocks/s/siesta.py
```diff
@@ -0,0 +1,342 @@
...
+
+            run_cmd('./build_all.sh', log_all=True, simple=True, log_output=True)
+
+            # Now move all the built utils to the temp installdir
+            expected_utils = [
```
boegel: any risk of this list changing over times?

can we come up with a smarter way to copy all utils?

easybuild/easyblocks/s/siesta.py
```diff
@@ -0,0 +1,342 @@
...
+                    'TS/ts2ts/ts2ts', 'TS/tshs2tshs/tshs2tshs', 'TS/TBtrans/tbtrans',
+                ])
+
+            for f in expected_utils:
+                shutil.copy(os.path.join(self.cfg['start_dir'], 'Util', f), bindir)
+
+        if self.cfg['with_transiesta']:
+            # Build transiesta
+            try:
+                os.chdir(obj_dir)
```
boegel: use `change_dir`

easybuild/easyblocks/s/siesta.py
akesandgren: @migueldiascosta Text to use for you will be?

easybuild/easyblocks/s/siesta.py
akesandgren: @boegel Is there a corresponding copytree?

akesandgren: apparently there is :-)

easybuild/easyblocks/s/siesta.py
migueldiascosta: not really necessary but if you want to, "Miguel Dias Costa (National University of Singapore)"

easybuild/easyblocks/s/siesta.py
akesandgren: What is weird?

Most correct way of having error on unsupported compiler?

easybuild/easyblocks/s/siesta.py
akesandgren: It WILL change over time, but figuring out what is actually built needs to be done by hand i think

easybuild/easyblocks/s/siesta.py
akesandgren: copy_dir?? not found in filetools, copytree exists though

easybuild/easyblocks/s/siesta.py
migueldiascosta: more lazier than smarter, I've simply used `buildininstalldir=True` instead, also because there are useful scripts and examples in the Utils folders

cleaner would be to copy the whole tree but without source and object files, and create links to the binaries in the bin folder

but none of this solves the need for a list of expected utils to sanity check - the `build_all.sh` iterates over all subdirs that have a makefile, we could get the default targets from each of them (?)

easybuild/easyblocks/s/siesta.py
boegel:
```python
if ...:  # GCC
    ...
elif ...:  # Intel
   ...
else:
  raise EasyBuildError("Don't know how to handle compiler family %s", comp_fam)

File: easybuild/easyblocks/s/siesta.py (hunk 23 above)
Comment (boegel) [inline easybuild/easyblocks/s/siesta.py]: @akesandgren having to patch the `$(CFLAGS)-c` to include the missing space is weird

File: easybuild/easyblocks/s/siesta.py (hunk 7 above)
Comment (boegel) [inline easybuild/easyblocks/s/siesta.py]: only available in `develop` for now, cfr. https://github.com/hpcugent/easybuild-framework/blob/develop/easybuild/tools/filetools.py#L1480

will be included in EasyBuild v3.2.0, just like this easyblock hopefully...

File: easybuild/easyblocks/s/siesta.py (hunk 29 above)
Comment (boegel) [inline easybuild/easyblocks/s/siesta.py]: well, an explicit sanity check doesn't hurt, it'll just be a PITA to maintain

but we can adjust accordingly later if it really becomes a problem

File: easybuild/easyblocks/s/siesta.py [hunk 31]
```diff
@@ -0,0 +1,320 @@
...
+    def build_step(self):
+        """No build step for Siesta."""
+        pass
+
+    def install_step(self):
+        """Custom install procedure for Siesta."""
+
+        # binary
+        bindir = os.path.join(self.installdir, 'bin')
+        copytree(os.path.join(self.cfg['start_dir'], 'bin'), bindir)
```
boegel: @akesandgren we deprecated `copytree` in `develop` for EasyBuild v3.2.0, in favour of `copy_dir`

easybuild/easyblocks/s/siesta.py
```diff
@@ -0,0 +1,320 @@
...
+            ('libsiestaLAPACK.a', ''),
+            # Needed here to allow 4.1-b1 to be built with openmp
+            (r"^(LDFLAGS\s*=).*$", r"\1 %s %s" % (os.environ['FCFLAGS'], os.environ['LDFLAGS'])),
+        ]
+
+        if netcdff_loc:
+            # Needed for gfortran at least
+            regex_subs.extend([
+                (r"^(ARFLAGS_EXTRA\s*=.*)$", r"\1\nNETCDF_INCFLAGS = -I%s/include" % netcdff_loc),
+            ])
```
boegel: just use `append` to add a single item

```python
regex_subs.append((r"^(ARFLAGS_EXTRA\s*=.*)$", r"\1\nNETCDF_INCFLAGS = -I%s/include" % netcdff_loc))

File: easybuild/easyblocks/s/siesta.py [hunk 33]
```diff
@@ -0,0 +1,320 @@
...
+
+        if fftw:
+            fftw_inc, fftw_lib = os.environ['FFTW_INC_DIR'], os.environ['FFTW_LIB_DIR']
+            regex_subs.extend([
```
boegel: same here, just use .append?

easybuild/easyblocks/s/siesta.py
boegel: @akesandgren line still too long?

easybuild/easyblocks/s/siesta.py
boegel: ah, wait, this isn't the sanity check...

well, it'll fail anyway when we try to copy files that aren't there

easybuild/easyblocks/s/siesta.py
```diff
@@ -0,0 +1,320 @@
...
+
+            if LooseVersion(self.version) >= LooseVersion("4.1"):
+                expected_utils.extend([
+                    'Bands/gnubands',
+                    'Grimme/fdf2grimme',
+                    'SpPivot/pvtsp',
+                    'TS/ts2ts/ts2ts', 'TS/tshs2tshs/tshs2tshs', 'TS/TBtrans/tbtrans',
+                ])
+
+            for f in expected_utils:
```
boegel: @akesandgren don't use single-letter variable names outside of list comprehensions please, `for util in expected_utils:`

akesandgren:
Hmm, time to start finishing this ...

What did you think was weird above? Comment Apr 6th "maybe include a comment on that above the lines below to make that clear"

boegel:
@akesandgren having to patch `$(CFLAGS)-c`, see https://github.com/hpcugent/easybuild-easyblocks/pull/1105/files#r108053518

migueldiascosta:
@boegel, @akesandgren regarding the `$(CFLAGS)-c` patch, I guess they didn't notice because their `CFLAGS`, from their `configure` or the template `arch.make`, must have been empty, but they did fix it as soon as I mentioned it to them: https://bugs.launchpad.net/siesta/+bug/1582978

easybuild/easyblocks/s/siesta.py
```diff
@@ -0,0 +1,318 @@
...
+            # Needed here to allow 4.1-b1 to be built with openmp
+            (r"^(LDFLAGS\s*=).*$", r"\1 %s %s" % (os.environ['FCFLAGS'], os.environ['LDFLAGS'])),
+        ]
+
+        if netcdff_loc:
+            # Needed for gfortran at least
+            regex_subs.append((r"^(ARFLAGS_EXTRA\s*=.*)$", r"\1\nNETCDF_INCFLAGS = -I%s/include" % netcdff_loc))
+
+        if fftw:
+            fftw_inc, fftw_lib = os.environ['FFTW_INC_DIR'], os.environ['FFTW_LIB_DIR']
```
boegel: You should be using `FFT_INC_DIR` and `FFT_LIB_DIR` here, which are also defined if `imkl` is being used

This was fixed in https://github.com/akesandgren/easybuild-easyblocks/pull/4, together with a bunch of minor style fixes.

boegel:
Thanks for all the hard work on this @akesandgren and @migueldiascosta!

Tested with https://github.com/hpcugent/easybuild-easyconfigs/pull/2965, looking good, so going in!

If you have any other Siesta easyconfigs (different versions/toolchains), please open a PR for them.

PR #1106, opened 2017-02-13
[WIP] Systemmpi

ocaisa:
Allow the use of a (recognised) MPI implementation from the system as part of a toolchain.

Current easyblock works for OpenMPI but requires some additional checks to ensure a consistent toolchain stack.

### DIFF
--- a/easybuild/easyblocks/generic/systemcompiler.py
+++ b/easybuild/easyblocks/generic/systemcompiler.py
@@ -69,6 +69,8 @@ class SystemCompiler(Bundle):
 
         # Determine compiler path (real path, with resolved symlinks)
         compiler_name = self.cfg['name'].lower()
+        if compiler_name == 'gcccore':
+            compiler_name = 'gcc'
         path_to_compiler = which(compiler_name)
         if path_to_compiler:
             path_to_compiler = os.path.realpath(path_to_compiler)
--- a/easybuild/easyblocks/generic/systemmpi.py
+++ b/easybuild/easyblocks/generic/systemmpi.py
@@ -1,0 +1,160 @@
+##
+# Copyright 2015-2016 Ghent University
+#
+# This file is part of EasyBuild,
+# originally created by the HPC team of Ghent University (http://ugent.be/hpc/en),
+# with support of Ghent University (http://ugent.be/hpc),
+# the Flemish Supercomputer Centre (VSC) (https://www.vscentrum.be),
+# Flemish Research Foundation (FWO) (http://www.fwo.be/en)
+# and the Department of Economy, Science and Innovation (EWI) (http://www.ewi-vlaanderen.be/en).
+#
+# http://github.com/hpcugent/easybuild
+#
+# EasyBuild is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation v2.
+#
+# EasyBuild is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with EasyBuild.  If not, see <http://www.gnu.org/licenses/>.
+##
+"""
+EasyBuild support for using (already installed/existing) system MPI instead of a full install via EasyBuild.
+
+@author Alan O'Cais (Juelich Supercomputing Centre)
+"""
+import os
+import re
+
+from easybuild.easyblocks.generic.bundle import Bundle
+from easybuild.tools.filetools import read_file, which
+from easybuild.tools.run import run_cmd
+from easybuild.framework.easyconfig.easyconfig import ActiveMNS
+from easybuild.tools.build_log import EasyBuildError
+
+
+class SystemMPI(Bundle):
+    """
+    Support for generating a module file for the system compiler with specified name.
+
+    The compiler is expected to be available in $PATH, required libraries are assumed to be readily available.
+
+    Specifying 'system' as a version leads to using the derived compiler version in the generated module;
+    if an actual version is specified, it is checked against the derived version of the system compiler that was found.
+    """
+
+    def extract_ompi_version(self, txt):
+        """Extract MPI version from provided string."""
+        # look for 3-4 digit version number, surrounded by spaces
+        # examples:
+        # gcc (GCC) 4.4.7 20120313 (Red Hat 4.4.7-11)
+        # Intel(R) C Intel(R) 64 Compiler XE for applications running on Intel(R) 64, Version 15.0.1.133 Build 20141023
+        version_regex = re.compile(r'\s(v[0-9]+(?:\.[0-9]+){1,3})\s', re.M)
+        res = version_regex.search(txt)
+        if res:
+            self.mpi_version = res.group(1)[1:]
+            self.log.debug("Extracted MPI version '%s' from: %s", self.mpi_version, txt)
+        else:
+            raise EasyBuildError("Failed to extract OpenMPI version using regex pattern '%s' from: %s",
+                                 version_regex.pattern, txt)
+
+    def __init__(self, *args, **kwargs):
+        """Extra initialization: determine system compiler version and prefix."""
+        super(SystemMPI, self).__init__(*args, **kwargs)
+
+        # Determine MPI path (real path, with resolved symlinks)
+        mpi_name = self.cfg['name'].lower()
+        mpi_c_compiler = 'mpicc'
+        path_to_mpi_c_compiler = which(mpi_c_compiler)
+        if path_to_mpi_c_compiler:
+            path_to_mpi_c_compiler = os.path.realpath(path_to_mpi_c_compiler)
+            self.log.info("Found path to MPI implementation '%s' mpicc compiler (with symlinks resolved): %s",
+                          mpi_name, path_to_mpi_c_compiler)
+        else:
+            raise EasyBuildError("%s not found in $PATH", mpi_c_compiler)
+
+        # Determine compiler version and installation prefix
+        if mpi_name == 'openmpi':
+            out, _ = run_cmd("ompi_info --version", simple=False)
+            self.extract_ompi_version(out)
+
+            # extract the installation prefix
+            self.mpi_prefix, _ = run_cmd("ompi_info --path prefix|awk '{print $2}'", simple=False)
+            # drop the carriage return
+            self.mpi_prefix = self.mpi_prefix[:-1]
+            # verify that toolchain compiler and compiler in MPI wrappers match
+            #TODO
+
+            # extract any OpenMPI environment variables in the current environment and ensure they are added to the
+            # final module
+            #TODO
+
+        else:
+            raise EasyBuildError("Unknown system MPI implementation %s", mpi_name)
+
+        self.log.debug("Derived version/install prefix for system MPI %s: %s, %s",
+                       mpi_name, self.mpi_version, self.mpi_prefix)
+
+        # If EasyConfig specified "real" version (not 'system' which means 'derive automatically'), check it
+        if self.cfg['version'] == 'system':
+            self.log.info("Found specified version '%s', going with derived compiler version '%s'",
+                          self.cfg['version'], self.mpi_version)
+        elif self.cfg['version'] != self.mpi_version:
+            raise EasyBuildError("Specified version (%s) does not match version reported by MPI (%s)" %
+                                 (self.cfg['version'], self.mpi_version))
+
+        # fix installdir and module names (may differ because of changes to version)
+        mns = ActiveMNS()
+        self.cfg.full_mod_name = mns.det_full_module_name(self.cfg)
+        self.cfg.short_mod_name = mns.det_short_module_name(self.cfg)
+        self.cfg.mod_subdir = mns.det_module_subdir(self.cfg)
+
+        # keep track of original values, for restoring later
+        self.orig_version = self.cfg['version']
+        self.orig_installdir = self.installdir
+
+    def make_installdir(self, dontcreate=None):
+        """Custom implementation of make installdir: do nothing, do not touch system compiler directories and files."""
+        pass
+
+    def make_module_req_guess(self):
+        """
+        A dictionary of possible directories to look for.  Return empty dict for a system MPI.
+        """
+        return {}
+
+    def make_module_step(self, fake=False):
+        """
+        Custom module step for SystemMPI: make 'EBROOT' and 'EBVERSION' reflect actual system MPI version
+        and install path.
+        """
+        # For module file generation: temporarily set version and installdir to system compiler values
+        self.cfg['version'] = self.mpi_version
+        self.installdir = self.mpi_prefix
+
+        # Generate module
+        res = super(SystemMPI, self).make_module_step(fake=fake)
+
+        # Reset version and installdir to EasyBuild values
+        self.installdir = self.orig_installdir
+        self.cfg['version'] = self.orig_version
+        return res
+
+    def make_module_extend_modpath(self):
+        """
+        Custom prepend-path statements for extending $MODULEPATH: use version specified in easyconfig file (e.g.,
+        "system") rather than the actual version (e.g., "4.8.2").
+        """
+        # temporarily set switch back to version specified in easyconfig file (e.g., "system")
+        self.cfg['version'] = self.orig_version
+
+        # Retrieve module path extensions
+        res = super(SystemMPI, self).make_module_extend_modpath()
+
+        # Reset to actual compiler version (e.g., "4.8.2")
+        self.cfg['version'] = self.mpi_version
+        return res


easybuild/easyblocks/generic/systemmpi.py
```diff
@@ -0,0 +1,160 @@
...
+        else:
+            raise EasyBuildError("%s not found in $PATH", mpi_c_compiler)
+
+        # Determine compiler version and installation prefix
+        if mpi_name == 'openmpi':
+            out, _ = run_cmd("ompi_info --version", simple=False)
+            self.extract_ompi_version(out)
+
+            # extract the installation prefix
+            self.mpi_prefix, _ = run_cmd("ompi_info --path prefix|awk '{print $2}'", simple=False)
```
ocaisa: Very happy if someone would suggest some regex that gives me the same end result (@geimer)

ocaisa:
The failing tests are because there is no `OpenMPI` installed on Travis (and I don't know if that is an option)

easybuild/easyblocks/generic/systemmpi.py
geimer: @ocaisa You mean something like `res = re.search('^\s+Open MPI: (.*)$', output_of_ompi_info)` and then `res.group(1)`?

ocaisa: Thanks @geimer, that gives me enough to help me figure out how to extract the rest of the info I'll need too

ocaisa:
@boegel Ready for review

easybuild/easyblocks/generic/systemmpi.py
```diff
@@ -0,0 +1,195 @@
...
+    Specifying 'system' as a version leads to using the derived mpi version in the generated module;
+    if an actual version is specified, it is checked against the derived version of the system mpi that was found.
+    """
+
+    def extract_ompi_setting(self, search, txt):
+        """Extract MPI version from provided string."""
+        # look for 3-4 digit version number, surrounded by spaces
+        # examples:
+        # gcc (GCC) 4.4.7 20120313 (Red Hat 4.4.7-11)
+        # Intel(R) C Intel(R) 64 Compiler XE for applications running on Intel(R) 64, Version 15.0.1.133 Build 20141023
```
ocaisa: Update this comment!

easybuild/easyblocks/generic/systemmpi.py
```diff
@@ -0,0 +1,192 @@
...
+        if self.mpi_c_compiler != c_compiler_name and self.compiler_version != compiler_version:
+            raise EasyBuildError("C compiler for toolchain (%s/%s) and underneath MPI (%s/%s) do not match!",
+                                 c_compiler_name, compiler_version, self.mpi_c_compiler, self.compiler_version)
+
+        # For module file generation: temporarily set version and installdir to system compiler values
+        self.cfg['version'] = self.mpi_version
+        self.installdir = self.mpi_prefix
+
+        # Generate module
+        res = super(SystemMPI, self).make_module_step(fake=fake)
```
ocaisa: There's a bug here but I cannot find it. In the final module it is setting
```
setenv("EBVERSIONOPENMPI","5.4.0")
```
where `5.4.0` is the version of the compiler underneath MPI. I've checked the value of `self.mpi_version` so I don't understand why this value is winning...is it somehow connected to `altversion`?

ocaisa:
@boegel Remaining error is simply that there is no `mpicc` on travis...how do I fudge this? Can `OpenMPI` be installed there?

besserox:
I saw this PR so I'm jumping in to share the solution I've made some time ago to address the same issue.

Basically, instead of creating a dedicated easyblock, I just create an 'empty' easyconfig file named  OpenMPI/system, ie it creates a module named `OpenMPI` with version `system`.

The advantage is that it is automatically recognized as an OpenMPI toolchain by EasyBuild, so there's no need for any kind of magic to properly set **MPICC**, **MPICXX** internally. EasyBuild will automatically set them to the right 'OpenMPI' values.

Same thing for software that needs to know the type of the toolchain in the easyblock (eg OpenFOAM). It works transparently.

See my 'implementation' in those files:
- **GCC/system**: https://github.com/besserox/easybuild-easyconfigs/blob/openfoam-system/easybuild/easyconfigs/o/OpenMPI/OpenMPI-system.eb
- **OpenMPI/system**: https://github.com/besserox/easybuild-easyconfigs/blob/openfoam-system/easybuild/easyconfigs/o/OpenMPI/OpenMPI-system.eb
- **gompi/system**: https://github.com/besserox/easybuild-easyconfigs/blob/openfoam-system/easybuild/easyconfigs/g/gompi/gompi-system.eb
- **OpenFOAM/system**: https://github.com/besserox/easybuild-easyconfigs/blob/openfoam-system/easybuild/easyconfigs/o/OpenFOAM/OpenFOAM-2.0.1-gompi-system.eb
- and all the other OpenFOAM dependencies

This allowed me to use EasyBuild to compile OpenFOAM on my Ubuntu workstation using the system compilers.


I believe it will work smoothly with the other compilers/MPIs.

ocaisa:
@besserox What you have should be fine for a subset of what I'm interested in. What I'm concerned about is more the case where there are various compilers and MPI implementations on the system and I want to chose a particular one to use as a toolchain. To do this I need to know exactly where it is installed and if there are any environment variables set that affect it's behaviour (which is frequently true for a tuned MPI instance).
My goal was to pull in the MPI form the system once to incorporate them from the system but then to forget about the modules available on the system (this is necessary for me because BSC uses a tweaked `module` tool that has module files that Lmod or any other common module tool can understand). It also locks the versions used by EB so that we don't get unnoticed compiler or MPI updates from the system.

easybuild/easyblocks/generic/systemcompiler.py
```diff
@@ -38,6 +38,29 @@
...
+        out, _ = run_cmd("%s -V" % compiler_name, simple=False)
+    else:
+        raise EasyBuildError("Unknown compiler %s" % compiler_name)
+
+    version_regex = re.compile(r'\s([0-9]+(?:\.[0-9]+){1,3})\s', re.M)
+    res = version_regex.search(out)
+    if res:
+        compiler_version = res.group(1)
+        debug_msg = "Extracted compiler version '%s' for %s from: %s", compiler_version, compiler_name, out
+        return compiler_version, debug_msg
```
boegel: don't pass down the debug log message, just log here; also, we try to avoid 'inline' `return` statements

```python
_log = fancylogger.getLogger('easyblocks.generic.systemcompiler')
...
def extract_compiler_version(compiler_name):

    ...
    if res:
        compiler_version = res.group(1)
        _log.debug("Extracted compiler version '%s' for %s from: %s", compiler_version, compiler_name, out)
    else:
        raise EasyBuildError("...")

    return compiler_version
```

easybuild/easyblocks/generic/systemmpi.py
```diff
@@ -0,0 +1,192 @@
...
+        self.cfg.full_mod_name = mns.det_full_module_name(self.cfg)
+        self.cfg.short_mod_name = mns.det_short_module_name(self.cfg)
+        self.cfg.mod_subdir = mns.det_module_subdir(self.cfg)
+
+        # keep track of original values, for restoring later
+        self.orig_version = self.cfg['version']
+        self.orig_installdir = self.installdir
+
+    def make_installdir(self, dontcreate=None):
+        """Custom implementation of make installdir: do nothing, do not touch system compiler directories and files."""
```
boegel: `s/compiler/MPI/g`

easybuild/easyblocks/generic/systemmpi.py
```diff
@@ -0,0 +1,192 @@
...
+        pass
+
+    def make_module_req_guess(self):
+        """
+        A dictionary of possible directories to look for.  Return empty dict for a system compiler.
```
boegel: `s/compiler/MPI/g`

easybuild/easyblocks/generic/systemmpi.py
```diff
@@ -0,0 +1,192 @@
...
+
+        # Retrieve module path extensions
+        res = super(SystemMPI, self).make_module_extend_modpath()
+
+        # Reset to actual MPI version (e.g., "4.8.2")
+        self.cfg['version'] = self.mpi_version
+        return res
+
+    def make_module_extra(self):
+        """Add all the build environment variables."""
```
boegel: fix docstring (nothing is being built?)

easybuild/easyblocks/generic/systemmpi.py
```diff
@@ -0,0 +1,192 @@
...
+
+    def extract_ompi_setting(self, search, txt):
+        """Extract a particular OpenMPI setting from provided string."""
+
+        version_regex = re.compile(r'^\s+%s: (.*)$' % search, re.M)
+        res = version_regex.search(txt)
+        if res:
+            setting = res.group(1)
+            self.log.debug("Extracted OpenMPI setting %s: '%s' from search text", search, setting)
+            return setting
```
boegel: avoid inline `return` please

easybuild/easyblocks/generic/systemmpi.py
```diff
@@ -0,0 +1,192 @@
+##
+# Copyright 2015-2016 Ghent University
```
boegel: `2015-2017`

easybuild/easyblocks/generic/systemmpi.py
```diff
@@ -0,0 +1,192 @@
...
+
+@author Alan O'Cais (Juelich Supercomputing Centre)
+"""
+import os
+import re
+
+from easybuild.easyblocks.generic.bundle import Bundle
+from easybuild.easyblocks.generic.systemcompiler import extract_compiler_version
+from easybuild.tools.modules import get_software_version
+from easybuild.tools.filetools import read_file, which
```
boegel: `read_file` is not used

easybuild/easyblocks/generic/systemmpi.py
```diff
@@ -0,0 +1,192 @@
...
+    The mpi compiler is expected to be available in $PATH, required libraries are assumed to be readily available.
+
+    Specifying 'system' as a version leads to using the derived mpi version in the generated module;
+    if an actual version is specified, it is checked against the derived version of the system mpi that was found.
+    """
+
+    def extract_ompi_setting(self, search, txt):
+        """Extract a particular OpenMPI setting from provided string."""
+
+        version_regex = re.compile(r'^\s+%s: (.*)$' % search, re.M)
```
boegel: rename `search` to `pattern`?

easybuild/easyblocks/generic/systemmpi.py
```diff
@@ -0,0 +1,192 @@
...
+            return setting
+        else:
+            raise EasyBuildError("Failed to extract OpenMPI setting '%s' using regex pattern '%s' from: %s",
+                                 search, version_regex.pattern, txt)
+
+    def __init__(self, *args, **kwargs):
+        """Extra initialization: determine system MPI version, prefix and any associated envvars."""
+        super(SystemMPI, self).__init__(*args, **kwargs)
+
+        # Determine MPI path (real path, with resolved symlinks) to ensure it exists
```
boegel: comment is in wrong place?

easybuild/easyblocks/generic/systemmpi.py
```diff
@@ -0,0 +1,192 @@
...
+        mpi_name = self.cfg['name'].lower()
+
+        # Check we have MPI wrappers
+        if mpi_name != 'impi':
```
boegel: use positive logic:

```python
if mpi_name == 'impi':
    mpi_c_wrapper = 'mpiicc'
else:
    mpi_c_wrapper = 'mpicc'

File: easybuild/easyblocks/generic/systemmpi.py [hunk 14]
```diff
@@ -0,0 +1,192 @@
...
+            output_of_ompi_info, _ = run_cmd("ompi_info", simple=False)
+            # Extract the version of OpenMPI
+            self.mpi_version = self.extract_ompi_setting("Open MPI", output_of_ompi_info)
+
+            # Extract the installation prefix
+            self.mpi_prefix = self.extract_ompi_setting("Prefix", output_of_ompi_info)
+
+            # Extract any OpenMPI environment variables in the current environment and ensure they are added to the
+            # final module
+            raw_env = os.environ
```
boegel: no need to introduce `raw_env`, just use `os.environ` on line below?

easybuild/easyblocks/generic/systemmpi.py
```diff
@@ -0,0 +1,192 @@
...
+        self.c_compiler_version, debug_msg = extract_compiler_version(self.mpi_c_compiler)
+        self.log.debug(debug_msg)
+        self.log.debug("Derived compiler/version for C compiler underneath system MPI %s: %s, %s",
+                       mpi_name, self.mpi_c_compiler, self.c_compiler_version)
+
+        # If EasyConfig specified "real" version (not 'system' which means 'derive automatically'), check it
+        if self.cfg['version'] == 'system':
+            self.log.info("Found specified version '%s', going with derived MPI version '%s'",
+                          self.cfg['version'], self.mpi_version)
+        elif self.cfg['version'] != self.mpi_version:
```
boegel: use positive logic, use `log.info` to inform that specified version matches found version

easybuild/easyblocks/generic/systemmpi.py
```diff
@@ -0,0 +1,192 @@
...
+    def make_module_step(self, fake=False):
+        """
+        Custom module step for SystemMPI: make 'EBROOT' and 'EBVERSION' reflect actual system MPI version
+        and install path.
+        """
+        # First let's verify that the toolchain and the compilers under MPI match
+        c_compiler_name = self.toolchain.COMPILER_CC
+        compiler_version = get_software_version(self.toolchain.COMPILER_MODULE_NAME[0])
+
+        if self.mpi_c_compiler != c_compiler_name and self.c_compiler_version != compiler_version:
```
boegel: `s/and/or/g`?

boegel:
@ocaisa To ensure that OpenMPI is available in Travis, see https://github.com/hpcugent/hanythingondemand/blob/master/.travis.yml#L3

easybuild/easyblocks/generic/systemmpi.py
```diff
@@ -0,0 +1,231 @@
...
+
+            # Extract the C compiler used underneath OpenMPI, check for the definition of OMPI_MPICC
+            self.mpi_c_compiler = self.extract_ompi_setting("C compiler", output_of_ompi_info)
+
+        elif mpi_name == 'impi':
+            # Extract the version of IntelMPI
+            # The prefix in the the mpiicc script can be used to extract the explicit version
+            contents_of_mpiicc, _ = read_file(path_to_mpi_c_wrapper)
+            prefix_regex = re.compile(r'(?<=compilers_and_libraries_)(.*)(?=/linux/mpi)', re.M)
+            self.mpi_version = prefix_regex.search(contents_of_mpiicc)
```
boegel: double-check to make sure the version was indeed found (i.e. not `None`)

easybuild/easyblocks/generic/systemmpi.py
```diff
@@ -0,0 +1,231 @@
...
+
+            # Extract the installation prefix
+            # If I_MPI_ROOT is defined, let's use that
+            raw_env = os.environ
+            if raw_env['I_MPI_ROOT']:
+                self.mpi_prefix = raw_env['I_MPI_ROOT']
```
boegel: no need to introduce `raw_env`, and you can do this in one go:

```python
self.mpi_prefix = os.getenv('I_MPI_ROOT', os.path.dirname(os.path.dirname(mpi_c_wrapper)))
```

In any case: also check that the directory you come up with actually exists?

easybuild/easyblocks/generic/systemmpi.py
```diff
@@ -0,0 +1,231 @@
...
+
+            # Extract any IntelMPI environment variables in the current environment and ensure they are added to the
+            # final module
+            raw_env = os.environ
+            self.mpi_envvars = dict((key, value) for key, value in raw_env.iteritems() if key.startswith("I_MPI_"))
+            self.mpi_envvars += dict((key, value) for key, value in raw_env.iteritems() if key.startswith("MPICH_"))
+            self.mpi_envvars += dict((key, value) for key, value in raw_env.iteritems()
+                                     if key.startswith("MPI") and key.endswith("PROFILE"))
+
+            # Extract the C compiler used underneath OpenMPI
```
boegel: not OpenMPI here, but Intel MPI

easybuild/easyblocks/generic/systemmpi.py
```diff
@@ -0,0 +1,231 @@
...
+            compile_info, _ = run_cmd("%s -compile-info", simple=False)
+            self.mpi_c_compiler = compile_info.group(0)
```
boegel: `compile_info` is just a string, so no `.group(0)`?

ocaisa: Sorry, still working on this, bit of a pain to test

besserox:
@ocaisa That makes sense

easybuild/easyblocks/generic/systemmpi.py
ocaisa: I've implemented that further down for the generic case

easybuild/easyblocks/generic/systemmpi.py
ocaisa: It is now :)

easybuild/easyblocks/generic/systemmpi.py
```diff
@@ -0,0 +1,249 @@
...
+                                 pattern, version_regex.pattern, txt)
+
+        return setting
+
+    def __init__(self, *args, **kwargs):
+        """Extra initialization: determine system MPI version, prefix and any associated envvars."""
+        super(SystemMPI, self).__init__(*args, **kwargs)
+
+        # If we are testing on Travis let's cheat (since it has been set up to have OpenMPI installed)
+        if self.cfg['name'] == 'foo':
```
boegel: hmm, no :)

we should find another way for this...

easybuild/easyblocks/generic/systemmpi.py
```diff
@@ -0,0 +1,255 @@
...
+
+        return return_dict
+
+    def make_module_step(self, fake=False):
+        """
+        Custom module step for SystemMPI: make 'EBROOT' and 'EBVERSION' reflect actual system MPI version
+        and install path.
+        """
+        # First let's verify that the toolchain and the compilers under MPI match
+        if self.travis:
```
boegel: Are we reimplementing https://github.com/auchenberg/volkswagen ?

This kind of code should never make it into easyblocks; instead, we should modify the tests to test specific easyblocks under controlled circumstances. See how we handle `systemcompiler.py` in `test/easyblocks/init_easyblocks.py`

ocaisa: Ok, ok

ocaisa:
This is almost done but systemcompiler needs some improvement for intel compilers that are not by default in the path

ocaisa:
Actually I think both blocks should be inheriting from configuremake

boegel:
@ocaisa Why would you derive from `ConfigureMake`? That doesn't make sense to me...

ocaisa:
I don't understand enough about the inheritance of easyblocks, which class makes the default choices for path extensions in the module file?

ocaisa:
Ok made the necessary changes to system compiler so that it also sets the paths. @geimer may want to take a look. I've added it as an option to the easyconfigs so it is backwards compatible.
Need to do something similar for systemmpi now.

ocaisa:
@boegel Should be ready for a re-review now, tested with OpenMPI, testing with impi now

easybuild/easyblocks/generic/systemcompiler.py
```diff
@@ -27,18 +27,48 @@
...
+def extract_compiler_version(compiler_name):
+    """Extract compiler version for provided compiler_name."""
+    # look for 3-4 digit version number, surrounded by spaces
+    # examples:
+    # gcc (GCC) 4.4.7 20120313 (Red Hat 4.4.7-11)
+    # Intel(R) C Intel(R) 64 Compiler XE for applications running on Intel(R) 64, Version 15.0.1.133 Build 20141023
+    if compiler_name == 'gcc':
+        out, _ = run_cmd("gcc --version", simple=False)
+    elif compiler_name in ['icc', 'ifort']:
+        out, _ = run_cmd("%s -V" % compiler_name, simple=False)
```
ocaisa: @geimer The version extracted here doesn't actually match the version used by EasyBuild. I could just extract that from the product path (for <2016 one search string, for later stuff another)

ocaisa:
Ok, I'm done, correct versions for Intel and also picks up the licence (if it can)

easybuild/easyblocks/generic/systemmpi.py
```diff
@@ -0,0 +1,317 @@
...
+            guesses = {}
+        return guesses
+
+    def make_module_step(self, fake=False):
+        """
+        Custom module step for SystemMPI: make 'EBROOT' and 'EBVERSION' reflect actual system MPI version
+        and install path.
+        """
+        # First let's verify that the toolchain and the compilers under MPI match
+        c_compiler_name = self.toolchain.COMPILER_CC
```
boegel: move this into the `else`

easybuild/easyblocks/generic/systemmpi.py
```diff
@@ -0,0 +1,317 @@
...
+        if c_compiler_name == 'DUMMYCC':
```
boegel: change to

```python
if self.toolchain.name == DUMMY_TOOLCHAIN_NAME:
```

(and add the missing import for `DUMMY_TOOLCHAIN_NAME`)

easybuild/easyblocks/generic/systemmpi.py
```diff
@@ -0,0 +1,317 @@
...
+            c_compiler_version = self.c_compiler_version
+            self.log.info("Found dummy toolchain so assuming GCC as compiler underneath MPI and faking the version")
+        else:
+            c_compiler_version = get_software_version(self.toolchain.COMPILER_MODULE_NAME[0])
+
+        if self.mpi_c_compiler != c_compiler_name or self.c_compiler_version != c_compiler_version:
+            raise EasyBuildError("C compiler for toolchain (%s/%s) and underneath MPI (%s/%s) do not match!",
+                                 c_compiler_name, c_compiler_version, self.mpi_c_compiler, self.c_compiler_version)
+
+        # For module file generation: temporarily set version and installdir to system compiler values
```
boegel: `s/compiler/MPI/g`

easybuild/easyblocks/generic/systemmpi.py
```diff
@@ -0,0 +1,317 @@
...
+        Custom prepend-path statements for extending $MODULEPATH: use version specified in easyconfig file (e.g.,
+        "system") rather than the actual version (e.g., "4.8.2").
+        """
+        # temporarily set switch back to version specified in easyconfig file (e.g., "system")
+        self.cfg['version'] = self.orig_version
+
+        # Retrieve module path extensions
+        res = super(SystemMPI, self).make_module_extend_modpath()
+
+        # Reset to actual MPI version (e.g., "4.8.2")
```
boegel: use `2.0.2` since this refers to Open MPI version for example?

easybuild/easyblocks/generic/systemmpi.py
```diff
@@ -0,0 +1,317 @@
...
+
+        # Reset version and installdir to EasyBuild values
+        self.installdir = self.orig_installdir
+        self.cfg['version'] = self.orig_version
+        return res
+
+    def make_module_extend_modpath(self):
+        """
+        Custom prepend-path statements for extending $MODULEPATH: use version specified in easyconfig file (e.g.,
+        "system") rather than the actual version (e.g., "4.8.2").
```
boegel: `s/4.8.2/2.0.2/g`

easybuild/easyblocks/generic/systemmpi.py
```diff
@@ -0,0 +1,317 @@
...
+        self.log.debug("Cleaning up after testing loading of module")
+        self.clean_up_fake_module(fake_mod_data)
+
+    def cleanup_step(self):
+        """Do nothing."""
+        pass
+
+    def permissions_step(self):
+        """Do nothing."""
+        pass
```
boegel: no need to override these last three if you first derive from `Bundle`?

easybuild/easyblocks/generic/systemcompiler.py
```diff
@@ -102,13 +135,24 @@ def __init__(self, *args, **kwargs):
...
         else:
             raise EasyBuildError("Unknown system compiler %s" % self.cfg['name'])
 
+        if not os.path.exists(self.compiler_prefix):
+            raise EasyBuildError("Path derived for system compiler (%s) does not exist: %s!",
+                                 compiler_name, self.compiler_prefix)
         self.log.debug("Derived version/install prefix for system compiler %s: %s, %s",
                        compiler_name, self.compiler_version, self.compiler_prefix)
 
+        if self.compiler_prefix in ['/usr']:
```
ocaisa: I thought people might want to extend this list (but I couldn't think of another obvious example)

boegel: well, we can revert it to `in [...]` if that actually happens?

easybuild/easyblocks/generic/systemcompiler.py
```diff
@@ -168,3 +255,28 @@ def make_module_extend_modpath(self):
         # Reset to actual compiler version (e.g., "4.8.2")
         self.cfg['version'] = self.compiler_version
         return res
+
+    def make_module_extra(self):
```
boegel: `make_module_extra` takes optional arguments (see https://github.com/hpcugent/easybuild-framework/blob/master/easybuild/framework/easyblock.py#L982), so you need to make sure you pass those down:

```python
def make_module_extra(self, *args, **kwargs):
    ...
        extras = EB_GCC.make_module_extra(self, *args, **kwargs)
    ...
        extras = super(SystemCompiler, self).make_module_extra(*args, **kwargs)

File: easybuild/easyblocks/generic/systemcompiler.py [hunk 32]
```diff
@@ -168,3 +249,35 @@ def make_module_extend_modpath(self):
...
+        self.log.debug("Cleaning up after testing loading of module")
+        self.clean_up_fake_module(fake_mod_data)
+
+    def cleanup_step(self):
+        """Do nothing."""
+        pass
+
+    def permissions_step(self):
+        """Do nothing."""
+        pass
```
ocaisa: These are required because they are not inherited from bundle (which doesn't define them)

ocaisa:
@boegel Should be good to go now, tested for local OpenMPI and remote Intel

I also updated the easyconfigs in https://github.com/hpcugent/easybuild-easyconfigs/issues/4136

ocaisa:
@boegel When is this going to get some love? I need to update it so I can play with Power8, I have to be dynamic with the XL compilers and Spectrum MPI since the available versions are still beta (and I have no sources). I'd prefer to do this in a new PR if I could.
Also there is no easyblock or existing easyconfigs for XL compilers, so I am just going to assume they are being delivered as RPMs.

easybuild/easyblocks/generic/systemmpi.py
```diff
@@ -0,0 +1,316 @@
...
+
+    def __init__(self, *args, **kwargs):
+        """Extra initialization: determine system MPI version, prefix and any associated envvars."""
+        super(SystemMPI, self).__init__(*args, **kwargs)
+
+        mpi_name = self.cfg['name'].lower()
+
+        # Determine MPI wrapper path (real path, with resolved symlinks) to ensure it exists
+        if mpi_name == 'impi':
+            mpi_c_wrapper = 'mpiicc'
```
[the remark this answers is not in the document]
ocaisa: Oh, that is a bit of a tough one because the line in that script I'm using to figure out the version doesn't exist in the impi `mpicc`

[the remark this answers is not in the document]
ocaisa: That's weird, it does exist for `mpif90` and `mpigcc`, I guess I can use `mpigcc`...but isn't that making another assumption that that exists?

In general, are you saying that if you didn't have intel compilers available that the `mpiicc` script would not exist? As it stands I only use it to extract the version info for `impi` so I am indifferent to which command I use...I just want to be certain the script exists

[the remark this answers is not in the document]
ocaisa: I guess I can it make it barf if `mpiicc` or `mpigcc` do not exist.

ocaisa:
Error from tests is
```
EasyBuildError: 'Failed to obtain class for SystemCompiler easyblock (not available?): Error when calling the metaclass bases\n    Cannot create a consistent method resolution\norder (MRO) for bases EasyBlock, EB_icc, EB_ifort'
```
Is that a recent addition? Any hints on a fix @boegel

easybuild/easyblocks/generic/systemmpi.py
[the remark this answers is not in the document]
boegel: If you're only using it to determine the version, it's fine.

Maybe trying to use `mpiicc` will just fail if the Intel compilers are not available, I'm not sure.

We can deal with that if the problem occurs though...

boegel:
@ocaisa For future reference: the `Cannot create a consistent method resolution order` problem occurs because you're deriving from both `EB_Icc` and `EB_ifort`, while `EB_ifort` itself also derives from `EB_icc`; you should just derive from `EB_ifort`, not both.

See also https://stackoverflow.com/questions/29214888/typeerror-cannot-create-a-consistent-method-resolution-order-mro

ocaisa:
Works for GCC/OpenMPI with easyconfigs from https://github.com/easybuilders/easybuild-easyconfigs/pull/4136

easybuild/easyblocks/generic/systemcompiler.py
```diff
@@ -168,3 +258,35 @@ def make_module_extend_modpath(self):
...
+            if self.cfg['name'] in ['GCC','GCCcore']:
+                extras = EB_GCC.make_module_extra(self, *args, **kwargs)
+            elif self.cfg['name'] in ['icc']:
+                extras = EB_icc.make_module_extra(self, *args, **kwargs)
+            elif self.cfg['name'] in ['ifort']:
+                extras = EB_ifort.make_module_extra(self, *args, **kwargs)
+            else:
+                raise EasyBuildError("I don't know how to generate extra module text for %s", self.cfg['name'])
+        else:
+            extras = super(SystemCompiler, self).make_module_extra(*args, **kwargs)
```
ocaisa: @boegel For a system `icc` this seems to kick an error:
```
File "/tmp/eb-JeUdnR/included-easyblocks/easybuild/easyblocks/generic/systemcompiler.py", line 290, in sanity_check_step
    fake_mod_data = self.load_fake_module(purge=True)
  File "/usr/local/software/jureca/Stages/2017a/software/EasyBuild/3.3.1/lib/python2.7/site-packages/easybuild_framework-3.3.1-py2.7.egg/easybuild/framework/easyblock.py", line 1251, in load_fake_module
    fake_mod_path = self.make_module_step(fake=True)
  File "/tmp/eb-JeUdnR/included-easyblocks/easybuild/easyblocks/generic/systemcompiler.py", line 240, in make_module_step
    res = super(SystemCompiler, self).make_module_step(fake=fake)
  File "/usr/local/software/jureca/Stages/2017a/software/EasyBuild/3.3.1/lib/python2.7/site-packages/easybuild_framework-3.3.1-py2.7.egg/easybuild/framework/easyblock.py", line 2195, in make_module_step
    txt += self.make_module_extra()
  File "/tmp/eb-JeUdnR/included-easyblocks/easybuild/easyblocks/generic/systemcompiler.py", line 274, in make_module_extra
    extras = super(SystemCompiler, self).make_module_extra(*args, **kwargs)
  File "/usr/local/software/jureca/Stages/2017a/software/EasyBuild/3.3.1/lib/python2.7/site-packages/easybuild_easyblocks-3.3.1-py2.7.egg/easybuild/easyblocks/generic/bundle.py", line 165, in make_module_extra
    return super(Bundle, self).make_module_extra(altroot=self.altroot, altversion=self.altversion)
TypeError: make_module_extra() got an unexpected keyword argument 'altroot'
```

ocaisa: This error only occurs when `generate_standalone_module = False`, otherwise it works

ocaisa: Should I be doing nothing if `generate_standalone_module = False` to avoid this problem?

boegel: You're not using the customised `bundle.py` easyblock from this PR which does `EasyBlock.make_module_extra(...)` rather than `super(Bundle, self).make_module_extra(...)`

boegel: I ran into this same problem when testing the `SystemMPI` easyblock, and it was fixed when also using the updated `bundle.py`.

ocaisa: Yes, confirmed, works with the `bundle.py` in the PR

ocaisa:
Looks good to go, tested with https://github.com/easybuilders/easybuild-easyconfigs/pull/4136 for GCC/OpenMPI and Intel compilers/MPI (2017), works as expected.

easybuild/easyblocks/generic/systemmpi.py
```diff
@@ -0,0 +1,325 @@
...
+            # For impi the version information is only found in *some* of the wrappers it ships, in particular it is
+            # not in mpicc
+            mpi_c_wrapper = 'mpiicc'
+            path_to_mpi_c_wrapper = which(mpi_c_wrapper)
+            if not path_to_mpi_c_wrapper:
+                mpi_c_wrapper = 'mpigcc'
+                path_to_mpi_c_wrapper = which(mpi_c_wrapper)
+                if not path_to_mpi_c_wrapper:
+                    raise EasyBuildError("Could not find suitable MPI wrapper to extract version for impi",
+                                         mpi_c_wrapper)
```
boegel: this extra argument shouldn't be here, see https://github.com/ocaisa/easybuild-easyblocks/pull/11

easybuild/easyblocks/generic/bundle.py
```diff
@@ -162,7 +162,7 @@ def install_step(self):
 
     def make_module_extra(self):
         """Set extra stuff in module file, e.g. $EBROOT*, $EBVERSION*, etc."""
-        return super(Bundle, self).make_module_extra(altroot=self.altroot, altversion=self.altversion)
+        return EasyBlock.make_module_extra(self, altroot=self.altroot, altversion=self.altversion)
```
boegel: there's a better way of fixing the issue you bumped into, see https://github.com/ocaisa/easybuild-easyblocks/pull/12

boegel:
Tested with https://github.com/easybuilders/easybuild-easyconfigs/pull/4136, good to go, thanks for your efforts and patience @ocaisa!

PR #1107, opened 2017-02-13
WIP (don't merge just yet) - GROMACS easyblock to install double precision alongside single precision

ostueker:
Gromacs builds with single and double precision can be installed
in the same directory, as dp builds get a `_d` suffix (this is
Gromacs standard).

The Easyblock has been changed to install double precision additionally
to the single precision build when option `double_precision` is
selected.

Therefore one Gromacs (with double_precision and MPI)  build will now
produce four binaries:

* gmx       (single precision; non-MPI)
* gmx_mpi   (single precision; MPI)
* gmx_d     (double precision; non-MPI)
* gmx_mpi_d (double precisionl MPI)

I'm not entirely happy with the solution I came up with because:

1. It would be nice to have the option to install (a) only single, (b) single and double or (c) only double precision. With the old easyblock it was only (a) and (c), now it's (a) and (b).
2. It would be nice to be able to have the option to inject `-DGMX_BUILD_MDRUN_ONLY=on` for the MPI builds. This will cause that instead of `gmx_mpi` and `gmx_mpi_d` only `mdrun_mpi` and `mdrun_mpi_d` are created as mdrun is the only part of Gromacs that uses MPI anyways.

I'm open for suggestions how to achieve those two goals.

### DIFF
--- a/easybuild/easyblocks/g/gromacs.py
+++ b/easybuild/easyblocks/g/gromacs.py
@@ -56,7 +56,7 @@ class EB_GROMACS(CMakeMake):
     @staticmethod
     def extra_options():
         extra_vars = {
-            'double_precision': [False, "Build with double precision enabled (-DGMX_DOUBLE=ON)", CUSTOM],
+            'double_precision': [False, "Additional build(s) with double precision enabled (-DGMX_DOUBLE=ON)", CUSTOM],
             'mpisuffix': ['_mpi', "Suffix to append to MPI-enabled executables (only for GROMACS < 4.6)", CUSTOM],
             'mpiexec': ['mpirun', "MPI executable to use when running tests", CUSTOM],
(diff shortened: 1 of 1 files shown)


mboisson:
For goal 1), why don't you make a different option which would be : 
precision = "single", "double", "both" ?

akesandgren:
I think we can do this a bit more streamlined and with less copying of code.
Something that's been in my TODO for a while.

ostueker:
Hi,
It's not longer relevant.
After finally grasping how `configopts` works, I changed our easyconfigs to use:

```
configopts = [
    " -DGMX_DOUBLE=OFF" ,
    " -DGMX_DOUBLE=ON" ,
]
```
instead of `precision = "single"` or something else.

PR #1108, opened 2017-02-14
Bundle: Allow per-component source_urls with templating

geimer:
Including a package with template-dependent `source_urls` in a bundle doesn't work.  If specified on the bundle level, the templating resolves to the bundle's values, which is in most cases plain wrong.  This PR allows to specify per-component `source_urls` with the templates resolved to the component's values.


boegel:
@geimer Looks good, thanks! Do you have an example use case for this?

boegel:
Makes perfect sense, I see no reason to hold this back. Thanks for the fix @geimer!

PR #1109, opened 2017-02-15
Qt: Use '-prefix <path>' rather than '--prefix=<path>' for configure

geimer:
Passing `--prefix=<path>` to `configure` leads to a hard error with Qt 5.8.0.  After checking the `configure --help` output of Qt 3.1.0, Qt 4.8.7, and Qt 5.8.0, it seems that `-prefix <path>` is the correct syntax (which was probably silently converted by versions prior to 5.8.0).


boegel:
lgtm, I'll get this tested with the existing Qt(5) easyconfigs, but I don't expect any problems (also works with Qt 3.3.8)

boegel:
testing didn't bring forward any problems, so going in, thanks @geimer!

PR #800, opened 2016-01-07
update sanity check in PETSc/SLEPc easyblocks for v3.6.x

boegel:
`conf` directory is moved in PETSc/SLEPc 3.6, cfr. http://www.mcs.anl.gov/petsc/documentation/changes/36.html and http://slepc.upv.es/download/changes.htm

PR #1111, opened 2017-02-21
Fixed indentation problem in petsc.py

mboisson:
Fixed an indentation problem which caused sanity checks to fail with petsc.py when shared libraries are enabled.


boegel:
Urgh, nice catch @mboisson!

This was introduced in #800, so my bad (and also @wpoely86 for missing it during review ;-)).

PR #1112, opened 2017-02-22
libVTmc.so needs to be reachable for itac.

akesandgren:
Itac needs to add slib to the LD_LIBRARY_PATH so mpirun -check_mpi can
find it.


boegel:
lgtm, tested with existing `itac` easyconfigs, thanks @akesandgren!

PR #1113, opened 2017-02-22
consider both lib and lib64 in CGAL sanity check

boegel:
for some reason, CGAL libraries now get installed into `lib64` rather than `lib` (seeing this with CGAL 4.9 on CentOS 7.3)...


boegel:

PR #1114, opened 2017-02-23
Allowing "download install" for petsc

mboisson:
When compiling petsc with "--download-<package>" configure options, these are installed in the installation directory in the configure step. 

The installation directory should therefore be created (and purged if there was an existing one) _BEFORE_ the configure step. It should also be kept after the configure step (rather than cleaned up before the install step).

This patch enables this when "downloadinstall" is set in the easy config.

### DIFF
--- a/easybuild/easyblocks/p/petsc.py
+++ b/easybuild/easyblocks/p/petsc.py
@@ -62,8 +62,8 @@ class EB_PETSc(ConfigureMake):
             'papi_lib': ['/usr/lib64/libpapi.so', "Path for PAPI library", CUSTOM],
             'runtest': ['test', "Make target to test build", BUILD],
             'download_packages_static': [None, "Lists the packages that should be downloaded and installed without shared objects", CUSTOM],
-            'download_packages_shared': [None, "Lists the packages that should be downloaded and installed with shared objects", CUSTOM]
-            'download_packages': [None, "Lists the packages that should be downloaded and installed without specifying static or shared (used the default for the package)", CUSTOM]
+            'download_packages_shared': [None, "Lists the packages that should be downloaded and installed with shared objects", CUSTOM],
+            'download_packages': [None, "Lists the packages that should be downloaded and installed without specifying static or shared (used the default for the package)", CUSTOM],
         }
         return ConfigureMake.extra_options(extra_vars)
 
--- a/easybuild/easyblocks/p/petsc.py
+++ b/easybuild/easyblocks/p/petsc.py
@@ -61,9 +61,9 @@ class EB_PETSc(ConfigureMake):
             'papi_inc': ['/usr/include', "Path for PAPI include files", CUSTOM],
             'papi_lib': ['/usr/lib64/libpapi.so', "Path for PAPI library", CUSTOM],
             'runtest': ['test', "Make target to test build", BUILD],
-            'download_packages_static': [None, "Lists the packages that should be downloaded and installed without shared objects", CUSTOM],
-            'download_packages_shared': [None, "Lists the packages that should be downloaded and installed with shared objects", CUSTOM],
-            'download_packages': [None, "Lists the packages that should be downloaded and installed without specifying static or shared (used the default for the package)", CUSTOM],
+            'download_deps_static': [[], "Dependencies that should be downloaded and installed static", CUSTOM],
+            'download_deps_shared': [[], "Dependencies that should be downloaded and installed shared", CUSTOM],
+            'download_deps': [[], "Dependencies that should be downloaded and installed", CUSTOM]
         }
         return ConfigureMake.extra_options(extra_vars)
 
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/p/petsc.py
```diff
@@ -79,6 +80,11 @@ def configure_step(self):
         Configure procedure is much more concise for older versions (< v3).
         """
         if LooseVersion(self.version) >= LooseVersion("3"):
+            # make the install dir first if we are doing a download install, then keep it for the rest of the way
+            if self.cfg["downloadinstall"]:
+                self.log.info("Creating the installation directory before the configure step because we are doing a download install.")
```
boegel: style nitpicking: this line is too long, should be <= 120 characters...

boegel:
@mboisson maybe the name of the new easyconfig parameter should be more directly related to what is actually happening, e.g. `create_installdir_before_configure`

Either that, or enabling `downloadinstall` should result in automatically adding all the `--download-*` configuration options (not sure if that's feasible or desirable)

mboisson:
Well, it also prevents the install dir from being removed after the configure step, so I'm not sure the new name is better. 

Adding all the "--download-*" configuration option is not possible. There are ~300 of those, most of them not frequently used by end users.

boegel:
@mboisson the fact that the installdir isn't being removed is implied, without that there's little point in creating it early...

I just find it confusing that the `downloadinstall` has nothing to do with downloading stuff directly...

mboisson:
Would it be less confusing if there was a second parameter called
download_packages
in which one would list the packages they want downloaded and then petsc.py would add them to the configure options ?

boegel:
@mboisson that makes sense, and then you don't need to have `downloadinstall` anymore, since listing stuff in `download_packages` would imply that the install dir needs to be created early

mboisson:
Allright. I'll modify the PR for that.

mboisson:
I ended up adding 3 options
download_packages   => --download-X=1
download_packages_shared  => --download-X=1 --download-X-shared=1
download_packages_static   => --download-X=1 --download-X-shared=0

easybuild/easyblocks/p/petsc.py
mboisson: The new comment line is also shorter.

easybuild/easyblocks/p/petsc.py
```diff
@@ -79,6 +82,22 @@ def configure_step(self):
...
+            # make the install dir first if we are doing a download install, then keep it for the rest of the way
+            if self.cfg["download_packages"] or self.cfg["download_packages_static"] or self.cfg["download_packages_shared"]:
+                self.log.info("Creating the installation directory before the configure.")
+                self.make_installdir()
+                self.cfg["keeppreviousinstall"] = True
+                if self.cfg["download_packages"]:
+                    for p in self.cfg["download_packages"]:
+                        self.cfg.update('configopts', '--with-download-%s=1' % p)
+                if self.cfg["download_packages_static"]:
+                    for p in self.cfg["download_packages_static"]:
```
boegel: @mboisson maybe make the default value an empty list `[]`, which avoids the extra `if` check?

easybuild/easyblocks/p/petsc.py
```diff
@@ -79,6 +82,22 @@ def configure_step(self):
...
+                        self.cfg.update('configopts', '--with-download-%s=1' % p)
+                        self.cfg.update('configopts', '--with-download-%s-shared=0' % p)
+                if self.cfg["download_packages_shared"]:
+                    for p in self.cfg["download_packages_shared"]:
```
boegel: no single-letter variables outside of list comprehensions please

easybuild/easyblocks/p/petsc.py
```diff
@@ -79,6 +82,22 @@ def configure_step(self):
...
+                        self.cfg.update('configopts', '--with-download-%s=1' % p)
```
boegel: quite a bit of duplicate code here, so how about this (I also renamed to `download_deps*`, which is shorter and avoids confusion with 'packages' like extra Python packages/extensions, etc...)

```python
for dep in self.cfg['download_deps'] + self.cfg['download_deps_shared'] + self.cfg['download_deps_static']:
    self.cfg.update('configopts', '--with-download-%s=1' % dep)
for dep in self.cfg['download_packages_shared']:
    self.cfg.update('configopts', '--with-download-%s-shared=1' % dep)
for dep in self.cfg['download_packages_static']:
    self.cfg.update('configopts', '--with-download-%s-shared=0' % dep)
```

easybuild/easyblocks/p/petsc.py
```diff
@@ -61,6 +61,9 @@ def extra_options():
             'papi_inc': ['/usr/include', "Path for PAPI include files", CUSTOM],
             'papi_lib': ['/usr/lib64/libpapi.so', "Path for PAPI library", CUSTOM],
             'runtest': ['test', "Make target to test build", BUILD],
+            'download_packages_static': [None, "Lists the packages that should be downloaded and installed without shared objects", CUSTOM],
+            'download_packages_shared': [None, "Lists the packages that should be downloaded and installed with shared objects", CUSTOM],
+            'download_packages': [None, "Lists the packages that should be downloaded and installed without specifying static or shared (used the default for the package)", CUSTOM],
```
boegel: keep the max. line length of 120 characters in mind please

mboisson: Shortened the lines

easybuild/easyblocks/p/petsc.py
mboisson: done

easybuild/easyblocks/p/petsc.py
mboisson: changed it for "dep" (and download_deps_* for the lists)

easybuild/easyblocks/p/petsc.py
mboisson: Done. I also added a "set" directive to make sure that if dependencies are listed multiple times, they are not included in the configure option multiple times.

easybuild/easyblocks/p/petsc.py
```diff
@@ -79,6 +82,20 @@ def configure_step(self):
...
         """
         if LooseVersion(self.version) >= LooseVersion("3"):
+            # make the install dir first if we are doing a download install, then keep it for the rest of the way
+            if self.cfg["download_deps"] or self.cfg["download_deps_static"] or self.cfg["download_deps_shared"]:
+                self.log.info("Creating the installation directory before the configure.")
+                self.make_installdir()
+                self.cfg["keeppreviousinstall"] = True
+                deps = set( self.cfg["download_deps"]
+                            + self.cfg["download_deps_static"]
+                            + self.cfg["download_deps_shared"] )
```
boegel: the wrapping makes this a bit messy, and you need to full list of deps in two places, so how about:

```python
        if LooseVersion(self.version) >= LooseVersion("3"):
            deps = self.cfg["download_deps"] + self.cfg["download_deps_static"] + self.cfg["download_deps_shared"]
            if deps:
                # make the install dir first if we are downloading dependencies, then keep it for the rest of the way
                ...
                for dep in set(deps):
                    ...
```

mboisson: done

mboisson:
I believe all the changes requested have been done.

boegel:
lgtm, doesn't affect existing PETSc easyconfigs, and already tested in the wild (cfr. https://lists.ugent.be/wws/arc/easybuild/2017-03/msg00061.html)

so, going in, thanks @mboisson!

PR #1115, opened 2017-02-23
Clang: Auto-detect default build_targets

geimer:
Uses the CPU architecture of the build host to determine the default build targets.

@boegel, @wpoely86: What do you think?  Does this make sense now that EasyBuild also supports ARM and Power?

### DIFF
--- a/easybuild/easyblocks/c/clang.py
+++ b/easybuild/easyblocks/c/clang.py
@@ -57,8 +57,8 @@ CLANG_TARGETS = ["all", "AArch64", "ARM", "CppBackend", "Hexagon", "Mips",
                  "MBlaze", "MSP430", "NVPTX", "PowerPC", "R600", "Sparc",
                  "SystemZ", "X86", "XCore"]
 
-# Mapping of EasyBuild CPU architecture names to list of LLVM target names
-TARGETS_MAP = {
+# Mapping of EasyBuild CPU architecture names to list of default LLVM target names
+DEFAULT_TARGETS_MAP = {
     AARCH32: ['ARM'],
     AARCH64: ['AArch64'],
     POWER: ['PowerPC'],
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/c/clang.py
```diff
@@ -84,16 +94,26 @@ def __init__(self, *args, **kwargs):
         self.llvm_obj_dir_stage3 = None
         self.make_parallel_opts = ""
 
-        unknown_targets = [target for target in self.cfg['build_targets'] if target not in CLANG_TARGETS]
+        build_targets = self.cfg['build_targets']
```
wpoely86: why the extra variable here? You only use it once in the line below?

wpoely86:
Seems like a good idea to me.

easybuild/easyblocks/c/clang.py
geimer: It's used three more times below.  But meanwhile you probably noticed ;)

wpoely86: yes but why not use the original? It's confusing to rename it?

geimer: It hasn't been renamed; `unknown_targets` is still there (see last line of the inserted block).  It just looks strange in the diff. `build_targets` is a newly introduced variable that is used later on.

easybuild/easyblocks/c/clang.py
```diff
@@ -84,16 +94,26 @@ def __init__(self, *args, **kwargs):
...
+        if build_targets is None:
+            arch = get_cpu_architecture()
+            default_target = TARGETS_MAP.get(arch, None)
+            if default_target:
+                self.cfg['build_targets'] = build_targets = default_target
```
boegel: This should be using `[default_target]` rather than `default_target`, since a list is expected later to define `DLLVM_TARGETS_TO_BUILD`:

```python
';'..join(self.cfg['build_targets'])
```

geimer: @boegel Hmm... `TARGETS_MAP.get()` already returns a list, no?

geimer:
@boegel I renamed `default_target` to `default_targets` and `TARGETS_MAP` to `DEFAULT_TARGETS_MAP` this may make things clearer.

boegel:
@geimer Sorry for the lack of follow-up here @geimer.

I finally managed to get back to this, tested it with all existing `Clang` easyconfigs + https://github.com/easybuilders/easybuild-easyconfigs/pull/5532.

Looks good, so going in, thanks!

Issue #1116, opened 2017-02-23
MKL now free


ocaisa:
https://software.intel.com/en-us/articles/free-mkl

Since the MKL is now free, it looks like the `silent.cfg` no longer requires an `activation_type`. An attempted build of last weeks release yields:
```
== 2017-02-23 11:11:23,156 build_log.py:147 ERROR EasyBuild crashed with an error (at ?:124 in __init__): build failed (first 300 chars): cmd "./install.sh  -s /gpfsdata/aocais/ECAM/build/imkl/2017.2.174/iimpi-system-GCC-system-2.27/l_mkl_2017.2.174/silent.cfg" exited with exitcode 2 and output:
Name "ACTIVATION_TYPE" is not valid
```

ocaisa:
For me the following silent.cfg seemed to work for an offline install:
```
SIGNING_ENABLED=yes
PSET_INSTALL_DIR=/gpfsdata/aocais/ECAM/software/imkl/2017.2.174-iimpi-system-GCC-system-2.27
ACCEPT_EULA=accept
INSTALL_MODE=NONRPM
CONTINUE_WITH_OPTIONAL_ERROR=yes
COMPONENTS=ALL
```

ocaisa:
Although it still tried to do `curl POST` at the end:
```
curl -X POST --max-time 5 --retry 3 -s -d @/gpfslocal/pub/tmp/install.3a3HQj/stat_data.json -o /gpfslocal/pub/tmp/install.3a3HQj/transfer.log -k https://ircws.intel.com/wsregcentertools/wsRegCenterISIP.asmx/RecordISIPData?deviceId=00000000-0000-0000-0000-000000000001
```

boegel:
A similar issue occurs when trying to install Intel Python, see https://github.com/hpcugent/easybuild-easyconfigs/issues/3922 

I've been working on an update of the `IntelBase` easyblock to make it more permissive w.r.t. providing a license file, PR coming soon...

ocaisa:
How about a new extravar for intelbase:
```
licensed = False
```
which could just clean out anything with activation_ in silent.cfg

boegel:
That's exactly what I was working on, but I got stuck a bit too deep down the rabbit hole with other stuff to get Intel Python to work...

ocaisa:
Ok, I will give it a shot with this version of MKL

DanielTosello:
[...]

PR #1117, opened 2017-02-24
add support for installing Intel tools that do not require license at installation time, incl. impi and imkl 2017.2.174

ocaisa:
Just add a new variable `licensed` to the intelbase to switch off licence related stuff

### DIFF
--- a/easybuild/easyblocks/generic/intelbase.py
+++ b/easybuild/easyblocks/generic/intelbase.py
@@ -112,6 +112,7 @@ class IntelBase(EasyBlock):
         extra_vars = EasyBlock.extra_options(extra_vars)
         extra_vars.update({
             'license_activation': [ACTIVATION_LIC_SERVER, "License activation type", CUSTOM],
+            'licensed': [True, "Set to False if Intel software does not require a runtime licence", CUSTOM],
             # 'usetmppath':
             # workaround for older SL5 version (5.5 and earlier)
             # used to be True, but False since SL5.6/SL6
@@ -206,32 +207,33 @@ class IntelBase(EasyBlock):
         """Custom prepare step for IntelBase. Set up the license"""
         super(IntelBase, self).prepare_step()
 
-        default_lic_env_var = 'INTEL_LICENSE_FILE'
-        lic_specs, self.license_env_var = find_flexlm_license(custom_env_vars=[default_lic_env_var],
-                                                              lic_specs=[self.cfg['license_file']])
+        if self.cfg['licensed']:
+            default_lic_env_var = 'INTEL_LICENSE_FILE'
+            lic_specs, self.license_env_var = find_flexlm_license(custom_env_vars=[default_lic_env_var],
+                                                                  lic_specs=[self.cfg['license_file']])
 
-        if lic_specs:
-            if self.license_env_var is None:
-                self.log.info("Using Intel license specifications from 'license_file': %s", lic_specs)
-                self.license_env_var = default_lic_env_var
+            if lic_specs:
+                if self.license_env_var is None:
+                    self.log.info("Using Intel license specifications from 'license_file': %s", lic_specs)
+                    self.license_env_var = default_lic_env_var
+                else:
+                    self.log.info("Using Intel license specifications from $%s: %s", self.license_env_var, lic_specs)
+
+                self.license_file = os.pathsep.join(lic_specs)
+                env.setvar(self.license_env_var, self.license_file)
+
+                # if we have multiple retained lic specs, specify to 'use a license which exists on the system'
+                if len(lic_specs) > 1:
+                    self.log.debug("More than one license specs found, using '%s' license activation instead of '%s'",
+                                   ACTIVATION_EXIST_LIC, self.cfg['license_activation'])
+                    self.cfg['license_activation'] = ACTIVATION_EXIST_LIC
+
+                    # $INTEL_LICENSE_FILE should always be set during installation with existing license
+                    env.setvar(default_lic_env_var, self.license_file)
             else:
-                self.log.info("Using Intel license specifications from $%s: %s", self.license_env_var, lic_specs)
-
-            self.license_file = os.pathsep.join(lic_specs)
-            env.setvar(self.license_env_var, self.license_file)
-
-            # if we have multiple retained lic specs, specify to 'use a license which exists on the system'
-            if len(lic_specs) > 1:
-                self.log.debug("More than one license specs found, using '%s' license activation instead of '%s'",
-                               ACTIVATION_EXIST_LIC, self.cfg['license_activation'])
-                self.cfg['license_activation'] = ACTIVATION_EXIST_LIC
-
-                # $INTEL_LICENSE_FILE should always be set during installation with existing license
-                env.setvar(default_lic_env_var, self.license_file)
-        else:
-            msg = "No viable license specifications found; "
-            msg += "specify 'license_file', or define $INTEL_LICENSE_FILE or $LM_LICENSE_FILE"
-            raise EasyBuildError(msg)
+                msg = "No viable license specifications found; "
+                msg += "specify 'license_file', or define $INTEL_LICENSE_FILE or $LM_LICENSE_FILE"
+                raise EasyBuildError(msg)
 
     def configure_step(self):
         """Configure: handle license file and clean home dir."""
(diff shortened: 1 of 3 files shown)


ocaisa:
Hopes to solve #1116

easybuild/easyblocks/generic/intelbase.py
```diff
@@ -260,31 +262,38 @@ def install_step(self, silent_cfg_names_map=None, silent_cfg_extras=None):
...
+                lic_file_entry,
+                ""  # Add a newline at the end, so we can easily append if needed
+            ]) % {
+                'activation_name': silent_cfg_names_map.get('activation_name', ACTIVATION_NAME),
+                'license_file_name': silent_cfg_names_map.get('license_file_name', LICENSE_FILE_NAME),
+                'activation': self.cfg['license_activation'],
+                'license_file': self.license_file,
+            }
+        else:
+            silent = '\n'
```
wpoely86: print a debug message that you are not going to use any license?

easybuild/easyblocks/generic/intelbase.py
```diff
@@ -112,6 +112,7 @@ def extra_options(extra_vars=None):
         extra_vars = EasyBlock.extra_options(extra_vars)
         extra_vars.update({
             'license_activation': [ACTIVATION_LIC_SERVER, "License activation type", CUSTOM],
+            'licensed': [True, "Set to False if Intel software does not require a runtime licence", CUSTOM],
```
ocaisa: Is this name potentially misleading, of course it is still licenced but doesn't require a runtime licence, how about `runtime_license`?

ocaisa:
This successfully builds the latest MKL for me

easybuild/easyblocks/generic/intelbase.py
ocaisa: Should I just be leveraging the existing `license_activation` variable, repurposing it to also handle a False value.

ocaisa: Thoughts @boegel

boegel: I had a WIP branch where I named this `requires_license`, but I guess `requires_runtime_license` would be more accurate...

easybuild/easyblocks/generic/intelbase.py
```diff
@@ -206,32 +207,33 @@ def prepare_step(self):
         """Custom prepare step for IntelBase. Set up the license"""
         super(IntelBase, self).prepare_step()
 
-        default_lic_env_var = 'INTEL_LICENSE_FILE'
-        lic_specs, self.license_env_var = find_flexlm_license(custom_env_vars=[default_lic_env_var],
-                                                              lic_specs=[self.cfg['license_file']])
+        if self.cfg['licensed']:
```
boegel: ideally, we would have an equivalent named argument to `prepare_step`, which we can leverage from easyblocks deriving from `IntelBase`:

something like this (logic may be a bit off, since the point of this is to avoid having to use `requires_runtime_license = False` in all future `imkl` easyconfigs):

```python
def prepare_step(self, requires_runtime_license=True):
    ...
    if self.cfg['requires_runtime_license'] or requires_runtime_license:
        ...
```

easybuild/easyblocks/generic/intelbase.py
```diff
@@ -260,31 +262,39 @@ def install_step(self, silent_cfg_names_map=None, silent_cfg_extras=None):
...
-        if self.cfg['license_activation'] in lic_file_server_activations:
-            lic_file_entry = "%(license_file_name)s=%(license_file)s"
-        elif not self.cfg['license_activation'] in other_activations:
-            raise EasyBuildError("Unknown type of activation specified: %s (known :%s)",
-                                 self.cfg['license_activation'], ACTIVATION_TYPES)
-
-        silent = '\n'.join([
-            "%(activation_name)s=%(activation)s",
-            lic_file_entry,
+        if self.cfg['licensed']:
```
boegel: maybe it's better to use `if self.runtime_license:` here, which is set to something not `None` in `prepare_step` in case there is a runtime license?

easybuild/easyblocks/generic/intelbase.py
```diff
@@ -372,8 +382,9 @@ def make_module_extra(self):
         """Custom variable definitions in module file."""
         txt = super(IntelBase, self).make_module_extra()
 
-        txt += self.module_generator.prepend_paths(self.license_env_var, [self.license_file],
-                                                   allow_abs=True, expand_relpaths=False)
+        if self.cfg['licensed']:
```
boegel: `if self.cfg['license_file']:` ?

ocaisa:
Ok, that still seems to be working for me, just did a rebuild

easybuild/easyblocks/generic/intelbase.py
```diff
@@ -203,11 +204,17 @@ def setup_local_home_subdir(self):
...
+    def prepare_step(self, requires_runtime_license=True):
         """Custom prepare step for IntelBase. Set up the license"""
         super(IntelBase, self).prepare_step()
 
-        if self.cfg['licensed']:
+        # Decide if we need a license or not
+        if not self.cfg['requires_runtime_license'] or not requires_runtime_license:
+            self.requires_runtime_license = False
+        else:
+            self.requires_runtime_license = True
```
boegel: how about

```python
self.requires_runtime_license = elf.cfg['requires_runtime_license'] or requires_runtime_license

Comment (ocaisa) [inline easybuild/easyblocks/generic/intelbase.py]: Isn't this always `True` regardless of the easyconfig setting since the default value of `requires_runtime_license` is `True`

Comment (ocaisa) [inline easybuild/easyblocks/generic/intelbase.py]: Rather than `or`, shouldn't it be `and`

Comment (boegel) [inline easybuild/easyblocks/generic/intelbase.py]: :+1: on `and`

**Comment by ocaisa** [comment]
Still works on latest MKL, I think this is good to go now

File: easybuild/easyblocks/generic/intelbase.py [hunk 7]
```diff
@@ -202,36 +204,40 @@ def setup_local_home_subdir(self):
...
-    def prepare_step(self):
+    def prepare_step(self, requires_runtime_license=True):
         """Custom prepare step for IntelBase. Set up the license"""
         super(IntelBase, self).prepare_step()
 
-        default_lic_env_var = 'INTEL_LICENSE_FILE'
-        lic_specs, self.license_env_var = find_flexlm_license(custom_env_vars=[default_lic_env_var],
-                                                              lic_specs=[self.cfg['license_file']])
+        # Decide if we need a license or not (default is True because of defaults of individual Booleans)
+        self.requires_runtime_license = self.cfg['requires_runtime_license'] and requires_runtime_license
```
boegel: please initialise `self.requires_runtime_license` in the constructor `__init__`, probably to `True`; this is import for `--module-only` for example

easybuild/easyblocks/generic/intelbase.py
```diff
@@ -372,8 +389,9 @@ def make_module_extra(self):
         """Custom variable definitions in module file."""
         txt = super(IntelBase, self).make_module_extra()
 
-        txt += self.module_generator.prepend_paths(self.license_env_var, [self.license_file],
-                                                   allow_abs=True, expand_relpaths=False)
+        if self.requires_runtime_license:
```
boegel: we need to think about what this does under `--module-only`, although it may be OK since `prepare_step` isn't skipped then?

easybuild/easyblocks/generic/intelbase.py
ocaisa: Done

easybuild/easyblocks/generic/intelbase.py
ocaisa: Yes, should be fine, if it worked before it should still work now (otherwise you would be seeing `UNKNOWN` being defined in modules by the existing code)

easybuild/easyblocks/generic/intelbase.py
```diff
@@ -260,31 +269,42 @@ def install_step(self, silent_cfg_names_map=None, silent_cfg_extras=None):
...
+                'license_file': self.license_file,
+            }
+        else:
+            self.log.debug(
+                "Ignoring license checks: self.requires_runtime_license=%s self.cfg['requires_runtime_license']=%s"
+                % (self.requires_runtime_license, self.cfg['requires_runtime_license'])
+            )
+            silent = '\n'
+
+        silent = silent.join([
```
boegel: this doesn't look right, this should be:

```
silent += '\n'.join([
```

Otherwise you'll be repeating the text that specifies the license several times.

See:

```python
>>> 'x'.join(['a', 'b', 'c'])
'axbxc'
```

easybuild/easyblocks/generic/intelbase.py
```diff
@@ -260,31 +269,42 @@ def install_step(self, silent_cfg_names_map=None, silent_cfg_extras=None):
...
+                'license_file_name': silent_cfg_names_map.get('license_file_name', LICENSE_FILE_NAME),
+                'activation': self.cfg['license_activation'],
+                'license_file': self.license_file,
+            }
+        else:
+            self.log.debug(
+                "Ignoring license checks: self.requires_runtime_license=%s self.cfg['requires_runtime_license']=%s"
+                % (self.requires_runtime_license, self.cfg['requires_runtime_license'])
+            )
+            silent = '\n'
```
boegel: this should be 

```python
silent = ''
```

easybuild/easyblocks/generic/intelbase.py
```diff
@@ -260,31 +269,42 @@ def install_step(self, silent_cfg_names_map=None, silent_cfg_extras=None):
...
+                'activation_name': silent_cfg_names_map.get('activation_name', ACTIVATION_NAME),
+                'license_file_name': silent_cfg_names_map.get('license_file_name', LICENSE_FILE_NAME),
+                'activation': self.cfg['license_activation'],
+                'license_file': self.license_file,
+            }
+        else:
+            self.log.debug(
+                "Ignoring license checks: self.requires_runtime_license=%s self.cfg['requires_runtime_license']=%s"
+                % (self.requires_runtime_license, self.cfg['requires_runtime_license'])
+            )
```
boegel: this looks overly verbose, I'd go with

```python
self.log.debug("No license required, so not including license specifications in silent.cfg")
```

easybuild/easyblocks/generic/intelbase.py
boegel: ok

easybuild/easyblocks/generic/intelbase.py
```diff
@@ -112,6 +115,8 @@ def extra_options(extra_vars=None):
         extra_vars = EasyBlock.extra_options(extra_vars)
         extra_vars.update({
             'license_activation': [ACTIVATION_LIC_SERVER, "License activation type", CUSTOM],
+            'requires_runtime_license': [True, "Set to False if Intel software does not require a runtime licence",
```
boegel: maybe use `"Boolean indicates whether or not a runtime license is required"` ?

boegel:
@ocaisa shall we also include an update to the `imkl` easyblock, where we call `prepare_step` with `requires_runtime_license=False` for recent `imkl` versions, to resolve #1116?

ocaisa:
Gonna need some guidance for imkl, how do I say
```
>= 2016.2
```

ocaisa:
Just tested again, works for me

easybuild/easyblocks/i/impi.py
```diff
@@ -58,6 +58,12 @@ def extra_options():
         }
         return IntelBase.extra_options(extra_vars)
 
+    def prepare_step(self):
+        if LooseVersion(self.version) >= LooseVersion('2017.2.174'):
+            super(EB_impi, self).prepare_step(requires_runtime_license=False)
```
boegel: wait... do you have a reference for this?

I know you can get the Intel MPI runtime for free, but also the full thing that allows you to build/link with/to it?

@DanielTosello?

ocaisa: https://software.intel.com/en-us/intel-mpi-library

ocaisa: There is says free download at least, same as MKL, this link is maybe a bit more explicit:
https://registrationcenter.intel.com/en/forms/?productid=2558&licensetype=2

ocaisa: For MKL, impi, TBB, DAAL, IPP your licence is now only getting you premium support

ocaisa: Eek, abort, abort! That is a named user licence, not valid for HPC:

> iii. NAMED-USER LICENSE: If You obtained the Materials under a named-user license, You may
allow only one (1) individual to install and use the Materials on no more than three (3)
computers provided that same individual is using the Materials only on one (1) computer at a
time. If You obtained a time-limited named-user license, the term of Your license and your
ability to use the Materials is limited to the time period of the obtained license, which is specified
on Intel’s download website, specified in the applicable documentation or controlled by the
license key for the Materials.

ocaisa: I think we can leave the option to switch off checking the licence there for people who respect this license agreement but they should need to explicitly add this to their easyconfig (and since we know the restrictions, we should perhaps warn them if they use this option).

boegel: Yes, we can provide the option, but it shouldn't be done automagically...

boegel: Doesn't matter, download of `impi` download for v2017.2.174 is identical, and doesn't allow you to provide a license file anymore either, so this change is required

ocaisa: Ok, so SHA256 sums on the 'licenced' downloads and the 'free' downloads are the same. Therefore it is no longer possible to provide your licence to these software packages and this PR can go in. People need to be careful about what they are agreeing to when they download the sources as the installation packages will no longer check your licencing during the install...so it is a case of *Freeloader beware!*

boegel:
verified with https://github.com/hpcugent/easybuild-easyconfigs/pull/4248, good to go, thanks @ocaisa!

PR #1118, opened 2017-02-28
tkinter easyblock - WIP

verdurin:
This is a prerequisite for `Python-2.7.13` and the separate `Tkinter` easyconfig.

I was unsure how much of the existing easyblock to remove, given that we're building the whole thing and deleting most of it.

Very open to cleaner ways of doing this.

### DIFF
--- a/easybuild/easyblocks/t/tkinterpython.py
+++ b/easybuild/easyblocks/t/tkinterpython.py
@@ -1,0 +1,187 @@
+##
+# Copyright 2009-2016 Ghent University
+#
+# This file is part of EasyBuild,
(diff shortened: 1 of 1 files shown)


boegel:
@verdurin Can you also open PRs for the Python without `Tkinter` + the `Tkinter` easyconfig?

I indeed think that this easyblock can be cleaned up quite a bit. Also, it should be named `EB_Tkinter` (there is no other `Tkinter` afaict)?

verdurin:
New PR 4225 for the easyconfigs depending on this.

boegel:
@wpoely86 do you have time to give this a thorough review or help @verdurin with cleaning this up?

PR #1119, opened 2017-02-28
New qt.py to address prefix opt change in 5.8

lucamar:
I propose to insert the following version check in the configure step of the easyblock qt.py:
`LooseVersion(self.version) >= LooseVersion('5.8')`
It is necessary since version 5.8.0 has changed the prefix option to `-prefix installdir` (it used to be `--prefix=installdir`).
I propose to let the user define `prefix_opt` as well, to comply with the help screen given by the command `eb -a -e EB_Qt` (as in ConfigureMake):
`prefix_opt* Prefix command line option for configure script ('--prefix=' if None) [default: None]`

### DIFF
--- a/easybuild/easyblocks/q/qt.py
+++ b/easybuild/easyblocks/q/qt.py
@@ -26,6 +26,7 @@
 EasyBuild support for building and installing Qt, implemented as an easyblock
 
 @author: Kenneth Hoste (Ghent University)
+@author: Luca Marsella (CSCS)
 """
 import os
 from distutils.version import LooseVersion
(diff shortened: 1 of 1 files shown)


boegel:
@lucamar This problem is already fixed in `develop`, see https://github.com/hpcugent/easybuild-easyblocks/pull/1109

lucamar:
Thanks @boegel, I had overlooked it, sorry for that. Do you think it would still make sense to insert the definition of `prefix_opt` in qt.py, as one can read in the easyblock specific options?

boegel:
@lucamar Yeah, I understand that the Qt easyblock not listening to `prefix_opt` may be confusing... 

But since the option is no longer needed, maybe it's best to simply remove `prefix_opt` as a known easyconfig parameter for `Qt`?

lucamar:
@boegel I agree with you, therefore I'm closing this pull request.

PR #1120, opened 2017-02-28
remove prefix_opt as custom easyconfig paramter for Qt since -prefix is hardcoded

boegel:
See remark by @lucamar in #1119

cc @geimer

without this patch:

```
$ eb -e EB_Qt -a | grep 'EASYBLOCK-SPECIFIC' -A 5
EASYBLOCK-SPECIFIC
------------------
configure_cmd_prefix*   Prefix to be glued before ./configure [default: ""]
platform*               Target platform to build for (e.g. linux-g++-64, linux-icc-64) [default: None]
prefix_opt*             Prefix command line option for configure script ('--prefix=' if None) [default: None]
tar_config_opts*        Override tar settings as determined by configure. [default: False]
```

with this patch:

```
$ eb -e EB_Qt -a | grep 'EASYBLOCK-SPECIFIC' -A 4
EASYBLOCK-SPECIFIC
------------------
configure_cmd_prefix*   Prefix to be glued before ./configure [default: ""]
platform*               Target platform to build for (e.g. linux-g++-64, linux-icc-64) [default: None]
tar_config_opts*        Override tar settings as determined by configure. [default: False]
```

### DIFF
--- a/easybuild/easyblocks/q/qt.py
+++ b/easybuild/easyblocks/q/qt.py
@@ -48,7 +48,12 @@ class EB_Qt(ConfigureMake):
         extra_vars = {
              'platform': [None, "Target platform to build for (e.g. linux-g++-64, linux-icc-64)", CUSTOM],
         }
-        return ConfigureMake.extra_options(extra_vars)
+        extra_vars = ConfigureMake.extra_options(extra_vars)
+
+        # allowing to specify prefix_opt doesn't make sense for Qt, since -prefix is hardcoded in configure_step
+        del extra_vars['prefix_opt']
+
+        return extra_vars
 
     def configure_step(self):
         """Configure Qt using interactive `configure` script."""


boegel:
@wpoely86: please review?

geimer:
@boegel The change itself makes sense to me.  But I'll leave it up to @wpoely86 to review the implementation.

easybuild/easyblocks/q/qt.py
```diff
@@ -48,7 +48,12 @@ def extra_options():
         extra_vars = {
              'platform': [None, "Target platform to build for (e.g. linux-g++-64, linux-icc-64)", CUSTOM],
         }
-        return ConfigureMake.extra_options(extra_vars)
+        extra_vars = ConfigureMake.extra_options(extra_vars)
+
+        # allowing to specify prefix_opt doesn't make sense for Qt, since -prefix is hardcoded in configure_step
+        del extra_vars['prefix_opt']
```
wpoely86: maybe at least a debug message that a value disappears?

boegel: I'm not sure that's worth the trouble here, since you won't see the `prefix_opt` pop up anywhere (e.g. the output of `eb -a -e EB_Qt`, etc.).

Also, we'd need to create a separate logger for it, since this is a static method (hence no `self.log` available here...). I would rather not to do...

wpoely86: fine by me then.

boegel:
Thanks for the review @geimer and @wpoely86!

PR #1121, opened 2017-03-03
enhance GROMACS easyblock to build with PLUMED support

akesandgren:
This enables a simplistic approach to patch Gromacs with PLUMED if PLUMED is a dependency.

### DIFF
--- a/easybuild/easyblocks/g/gromacs.py
+++ b/easybuild/easyblocks/g/gromacs.py
@@ -45,7 +45,7 @@ from easybuild.easyblocks.generic.cmakemake import CMakeMake
 from easybuild.framework.easyconfig import CUSTOM
 from easybuild.tools.build_log import EasyBuildError
 from easybuild.tools.filetools import download_file, extract_file, which
-from easybuild.tools.modules import get_software_libdir, get_software_root
+from easybuild.tools.modules import get_software_libdir, get_software_root, get_software_version
 from easybuild.tools.run import run_cmd
 from easybuild.tools.systemtools import get_platform_name , get_shared_lib_ext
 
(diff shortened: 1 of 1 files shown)


akesandgren:
An easyconfig that uses this is pending, waiting for a merge of PLUMED 2.3.0 with foss/2017a (easyconfig PR   4251)

easybuild/easyblocks/g/gromacs.py
```diff
@@ -70,6 +70,26 @@ def __init__(self, *args, **kwargs):
         self.lib_subdir = ''
         self.pre_env = ''
 
+        # check whether PLUMED is listed as a dependency
+        self.with_plumed = 'PLUMED' in [dep['name'] for dep in self.cfg['dependencies']]
```
boegel: why not just check this in `prepare_step`, since `self.with_plumed` isn't used anywhere else?

also, we usually check this via `get_software_root('PLUMED')` (even though that's not perfect, see https://github.com/hpcugent/easybuild-framework/issues/1269)

easybuild/easyblocks/g/gromacs.py
```diff
@@ -70,6 +70,26 @@ def __init__(self, *args, **kwargs):
...
+        super(EB_GROMACS, self).prepare_step()
+
+        if self.with_plumed:
+            # Need to check if plumed has an engine for this version
+            engine = 'gromacs-%s' % self.version
+
+            # Should use shared, static, or runtime patch depending on 
+            # setting of self.toolchain.options.get('dynamic')
+            # and adapt cmake flags accordingly as per instructions
+            # from plumed patch
```
boegel: reference for those instructions?

both this comment and the one above are TODO's, so maybe mark this PR as `(WIP)`?

easybuild/easyblocks/g/gromacs.py
akesandgren: I mimiced the code from dl_poly_classic.py which was the only one i could find that used plumed. But I can move it to the prepare step.

easybuild/easyblocks/g/gromacs.py
akesandgren: Mark it how? The reference is "plumed patch -h" :-)

boegel: mark it `(WIP)` in the PR title?

easybuild/easyblocks/g/gromacs.py
boegel: Well, we can certainly do better than `get_software_root`, yes, but it's not implemented in framework yet...

easybuild/easyblocks/g/gromacs.py
akesandgren: Like that?

boegel: yep

boegel:
@akesandgren Do you have an example GROMACS easyconfig that uses PLUMED that can be used as a testcase for these changes?

akesandgren:
easyconfig PR #4561 Gromacs 2016.1 with plumed

boegel:
Tested with https://github.com/hpcugent/easybuild-easyconfigs/pull/4561 and existing GROMACS easyconfigs, good to go.

Thanks @akesandgren!

PR #1127, opened 2017-03-09
add easyblock for QScintilla

boegel:
for now only works if QScintilla is built on top of Qt v4.x, can be enhanced later to also work with Qt5

### DIFF
--- a/easybuild/easyblocks/q/qscintilla.py
+++ b/easybuild/easyblocks/q/qscintilla.py
@@ -1,0 +1,144 @@
+##
+# Copyright 2009-2017 Ghent University
+#
+# This file is part of EasyBuild,
+# originally created by the HPC team of Ghent University (http://ugent.be/hpc/en),
+# with support of Ghent University (http://ugent.be/hpc),
+# the Flemish Supercomputer Centre (VSC) (https://www.vscentrum.be),
+# Flemish Research Foundation (FWO) (http://www.fwo.be/en)
+# and the Department of Economy, Science and Innovation (EWI) (http://www.ewi-vlaanderen.be/en).
+#
+# http://github.com/hpcugent/easybuild
+#
+# EasyBuild is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation v2.
+#
+# EasyBuild is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with EasyBuild.  If not, see <http://www.gnu.org/licenses/>.
+##
+"""
+EasyBuild support for building and installing QScintilla, implemented as an easyblock
+
+author: Kenneth Hoste (HPC-UGent)
+"""
+import os
+from distutils.version import LooseVersion
+
+from easybuild.easyblocks.generic.configuremake import ConfigureMake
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/q/qscintilla.py
```diff
@@ -0,0 +1,144 @@
...
+    def configure_step(self):
+        """Custom configuration procedure for QScintilla."""
+
+        srcdir = os.path.join(self.cfg['start_dir'], 'Qt4Qt5')
+        try:
+            os.chdir(srcdir)
+        except OSError as err:
+            raise EasyBuildError("Failed to change to %s: %s", srcdir, err)
+
+        regex_subs = [
```
wpoely86: add a comment why we do this?

easybuild/easyblocks/q/qscintilla.py
```diff
@@ -0,0 +1,144 @@
...
+
+    def install_step(self):
+        """Custom install procedure for QScintilla."""
+        
+        super(EB_QScintilla, self).install_step()
+
+        # also install Python bindings if Python is included as a dependency
+        python = get_software_root('Python')
+        if python:
+
```
wpoely86: less white space

easybuild/easyblocks/q/qscintilla.py
```diff
@@ -0,0 +1,144 @@
...
+            pydir = os.path.join(self.cfg['start_dir'], 'Python')
+            try:
+                os.chdir(pydir)
+            except OSError as err:
+                raise EasyBuildError("Failed to change to %s: %s", pydir, err)
```
wpoely86: shouldn't we wrap this into something nicer for dry run?

easybuild/easyblocks/q/qscintilla.py
```diff
@@ -0,0 +1,144 @@
...
+                raise EasyBuildError("Failed to symlink PyQt Python bindings in %s: %s", target_dir, err)
+
+            # also requires empty __init__.py file to ensure Python modules can be imported from this location
+            write_file(os.path.join(target_dir, '__init__.py'), '')
+
+    def sanity_check_step(self):
+        """Custom sanity check for QScintilla."""
+
+        if LooseVersion(self.version) >= LooseVersion('2.10'):
+            qsci_lib = 'libqscintilla2_qt4.so'
```
wpoely86: hardcoded `.so`

easybuild/easyblocks/q/qscintilla.py
boegel: ok, done

easybuild/easyblocks/q/qscintilla.py
boegel: we should have a `change_dir` function in framework, it's much overdue, but I won't block this PR over it (since I want to include this in the next EasyBuild release which is likely going to be v3.1.2, and `change_dir` wouldn't be supported until v3.2.0)

easybuild/easyblocks/q/qscintilla.py
boegel: ok, thx

easybuild/easyblocks/q/qscintilla.py
boegel: sure

easybuild/easyblocks/q/qscintilla.py
boegel: `change_dir` support implemented in https://github.com/hpcugent/easybuild-framework/pull/2155

I'll issue a large PR to introduce the use of `change_dir` in all easyblocks when working towards EB v3.2.0...

boegel:
Tested with https://github.com/hpcugent/easybuild-easyconfigs/pull/4306, good to go

PR #1129, opened 2017-03-10
[intelbase, pgi, totalview] Allow list of license files/servers via 'license_file' easyconfig parameter

geimer:
So far, passing the license via the easyconfig parameter `license_file` only worked with a single file/server.  To set multiple files/servers, one had to define the corresponding environment variable before invoking `eb`.

**Edit:** Now requires https://github.com/hpcugent/easybuild-framework/pull/2157

### DIFF
--- a/easybuild/easyblocks/generic/intelbase.py
+++ b/easybuild/easyblocks/generic/intelbase.py
@@ -44,6 +44,7 @@ from distutils.version import LooseVersion
 import easybuild.tools.environment as env
 from easybuild.framework.easyblock import EasyBlock
 from easybuild.framework.easyconfig import CUSTOM
+from easybuild.framework.easyconfig.types import ensure_iterable_license_specs
 from easybuild.tools.build_log import EasyBuildError
 from easybuild.tools.filetools import find_flexlm_license, read_file
 from easybuild.tools.run import run_cmd
@@ -215,17 +216,8 @@ class IntelBase(EasyBlock):
         self.requires_runtime_license = self.cfg['requires_runtime_license'] and requires_runtime_license
 
         if self.requires_runtime_license:
-            license_file = self.cfg['license_file']
-            if (license_file is None) or isinstance(license_file, basestring):
-                license_specs = [license_file]
-            elif isinstance(license_file, (list, tuple)) and all(isinstance(x, basestring) for x in license_file):
-                license_specs = license_file
-            else:
-                msg = "Unsupported type %s for easyconfig parameter 'license_file'! " % type(license_file)
-                msg += "Can either be None, a string, or a tuple/list of strings."
-                raise EasyBuildError(msg)
-
             default_lic_env_var = 'INTEL_LICENSE_FILE'
+            license_specs = ensure_iterable_license_specs(self.cfg['license_file'])
             lic_specs, self.license_env_var = find_flexlm_license(custom_env_vars=[default_lic_env_var],
                                                                   lic_specs=license_specs)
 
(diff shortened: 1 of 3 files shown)


geimer:
@boegel Please review?

easybuild/easyblocks/generic/intelbase.py
```diff
@@ -215,9 +215,19 @@ def prepare_step(self, requires_runtime_license=True):
         self.requires_runtime_license = self.cfg['requires_runtime_license'] and requires_runtime_license
 
         if self.requires_runtime_license:
+            license_file = self.cfg['license_file']
+            if isinstance(license_file, basestring):
+                license_specs = [license_file]
+            elif isinstance(license_file, list):
```
boegel: maybe also allow `tuple`?

```python
elif isinstance(license_file, (list, tuple)):`
```

easybuild/easyblocks/generic/intelbase.py
```diff
@@ -215,9 +215,19 @@ def prepare_step(self, requires_runtime_license=True):
...
+                license_specs = license_file
+            else:
+                msg = "Unsupported type for easyconfig parameter 'license_file'! "
```
boegel: mention which type you did get, using `type(license_file)`

easybuild/easyblocks/generic/intelbase.py
```diff
@@ -215,9 +215,19 @@ def prepare_step(self, requires_runtime_license=True):
...
+                msg += "Can either be a string or a list of strings."
```
boegel: you're not checking whether you're getting a list of strings, just whether you're getting a list...

If you want to, you could use something like:

```python
if not all(isinstance(x, basestring) for x in license_file):
```

boegel:
@geimer You broke the tests, but I'm not entirely sure why... If you include the type you are actually getting in the error message, you should be able to tell.

Maybe it's related to the default value for `license_file`, which may be `None`?

Also, I don't the copy-pasting very much, maybe we should provide a utility function for this in framework instead?
I know I did suggest to just handle this in the easyblocks, but I overlooked the copy-pasting aspect...

geimer:
@boegel Yeah, I forgot to handle `None`.  Fix is already there, but not pushed yet.

Regarding the "convert stuff to list" thing: Where to put it and how to name it? `convert_string_list_of_strings_or_tuple_of_strings_to_list_of_strings` seems excessively verbose :wink:

boegel:
@geimer How about `ensure_iterable_value`? You could make it return a value so you can wrap a single string into a single-element list?

```python
license_specs = ensure_iterable_value(self.cfg['license_file'])
```

easybuild/easyblocks/generic/intelbase.py
boegel: W.r.t. `tuple`, better check whether `find_flexlm_license` would handle that correctly (I think it would, but I'm not sure)

geimer:
@boegel Tuple works as expected.

W.r.t. `ensure_iterable_value`: A list of dicts is also iterable.  Thus, it feels like the error handling would still have to be done in the individual easyblocks (with code replication) with such a generic name. Maybe `ensure_iterable_type_of_strings`? Though this wouldn't cover a `None` element properly.

geimer:
`ensure_iterable_license_specs`?  Would be very specific, but one could keep the precise error message.  Put it in `framework/tools/utilities.py`?

boegel:
`ensure_iterable_license_specs` makes sense, maybe `easybuild.framework.easyconfig.types` makes sense as location?

boegel:
lgtm

boegel:
@geimer Ignore the latter, it's obviously because https://github.com/hpcugent/easybuild-framework/issues/2157 isn't merged yet.

I'm testing a script that reports back on failed Travis test runs back in the PR (and it seems to be working, \o/).

boegel:
Good to go now, thanks @geimer!

PR #1132, opened 2017-03-13
add support for EASYBUILD_ROBOT_PATH

foxik0070:
@boegel Added support for keeping EASYBUILD_ROBOT_PATH setting.

### DIFF
--- a/easybuild/easyblocks/e/easybuildmeta.py
+++ b/easybuild/easyblocks/e/easybuildmeta.py
@@ -223,3 +223,13 @@ class EB_EasyBuildMeta(PythonPackage):
             self.log.debug("Restored copy of original environment")
 
         return modpath
+
+    def make_module_extra(self):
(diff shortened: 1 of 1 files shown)


boegel:
@foxik0070 Hmm, it looks like a more generic approach that allows you to either pick up the whole configuration or selected parts of it, would be more useful? This seems a bit too specific to me?

Can you clarify the context for this, how are you installing EasyBuild (using `eb`, using the bootstrap script, etc.)?

foxik0070:
@boegel I installed EasyBuild using eb. We would like ours setting robot-path to propagate our users.

boegel:
@foxik0070 I think we should try to make this more generic, so either any `$EASYBUILD_*` environment variables are picked up, or so you can specify which ones should be picked up somehow?

foxik0070:
@boegel Modified. Tested, Works...

PR #1133, opened 2017-03-13
modified trinity sanity check for versions 2.3 and above

mboisson:
In versions 2.3 and 2.4, the name of the folder has changed. Need to adjust the sanity check to take it into consideration.


boegel:
(close/open to wake up Travis...)

boegel:
lgtm, thanks @mboisson!

PR #1134, opened 2017-03-14
Make unpacking Python extensions optional

wpoely86:

PR #1135, opened 2017-03-14
Added block for MXNet

wpoely86:
@hajgato this works
~~It needs #1134~~

To Do:
- [x] Found out why the python sanity check fails. Most likely `PYTHONPATH` is missing.

### DIFF
--- a/easybuild/easyblocks/m/mxnet.py
+++ b/easybuild/easyblocks/m/mxnet.py
@@ -43,7 +43,7 @@ from easybuild.tools.run import run_cmd
 from easybuild.tools.systemtools import get_shared_lib_ext
 
 # the namespace file for the R extension
-r_namespace = """# Export all names
+R_NAMESPACE = """# Export all names
 exportPattern(".")
 
 # Import all packages listed as Imports or Depends
@@ -94,19 +94,16 @@ class EB_MXNet(MakeCp):
         else:
             raise EasyBuildError("Failed to find/isolate MXNet source directory: %s", mxnet_dirs)
 
-        try:
-            for srcdir in [d for d in os.listdir(self.builddir) if not d.startswith('mxnet-')]:
-                submodule, _, _ = srcdir.rpartition('-')
-                newdir = os.path.join(self.mxnet_src_dir, submodule)
-                olddir = os.path.join(self.builddir, srcdir)
-                # first remove empty existing directory
-                rmtree2(newdir)
-                try:
-                    shutil.move(olddir, newdir)
-                except IOError, err:
-                    raise EasyBuildError("Failed to move %s to %s: %s", olddir, newdir, err)
-        except (IOError, OSError) as err:
-            raise EasyBuildError("Failed to move the different source directories to the main directory: %s", err)
+        for srcdir in [d for d in os.listdir(self.builddir) if not d.startswith('mxnet-')]:
+            submodule, _, _ = srcdir.rpartition('-')
+            newdir = os.path.join(self.mxnet_src_dir, submodule)
+            olddir = os.path.join(self.builddir, srcdir)
+            # first remove empty existing directory
+            rmtree2(newdir)
+            try:
+                shutil.move(olddir, newdir)
+            except IOError, err:
+                raise EasyBuildError("Failed to move %s to %s: %s", olddir, newdir, err)
 
         # the nnvm submodules has dmlc-core as a submodule too. Let's put a symlink in place.
         newdir = os.path.join(self.mxnet_src_dir, "nnvm", "dmlc-core")
@@ -169,7 +166,7 @@ class EB_MXNet(MakeCp):
         symlink(os.path.join(self.installdir, "include"), os.path.join("inst", "include"))
 
         # MXNet doesn't provide a list of its R dependencies by default
-        write_file("NAMESPACE", r_namespace)
+        write_file("NAMESPACE", R_NAMESPACE)
         change_dir(self.mxnet_src_dir)
         self.r_ext.prerun()
         # MXNet is just weird. To install the R extension, we have to:
@@ -198,10 +195,10 @@ class EB_MXNet(MakeCp):
             raise EasyBuildError("Loading fake module failed: %s", err)
 
         if not self.py_ext.sanity_check_step():
-            raise EasyBuildError("The sanity check of the Python bindings failed")
+            raise EasyBuildError("The sanity check for the Python bindings failed")
 
         if not self.r_ext.sanity_check_step():
-            raise EasyBuildError("The sanity check of the R bindings failed")
+            raise EasyBuildError("The sanity check for the R bindings failed")
 
         self.clean_up_fake_module(fake_mod_data)
 
--- a/easybuild/easyblocks/m/mxnet.py
+++ b/easybuild/easyblocks/m/mxnet.py
@@ -117,8 +117,8 @@ class EB_MXNet(MakeCp):
 
     def prepare_step(self):
         """Prepare for building and installing MXNet."""
-        self.py_ext.prepare_step()
         super(EB_MXNet, self).prepare_step()
+        self.py_ext.prepare_step()
 
     def configure_step(self):
         """Patch 'config.mk' file to use EB stuff"""
--- a/easybuild/easyblocks/m/mxnet.py
+++ b/easybuild/easyblocks/m/mxnet.py
@@ -37,8 +37,7 @@ from easybuild.easyblocks.generic.pythonpackage import PythonPackage
 from easybuild.easyblocks.generic.rpackage import RPackage
 from easybuild.framework.easyconfig import CUSTOM
 from easybuild.tools.build_log import EasyBuildError
-from easybuild.tools.filetools import apply_regex_substitutions, copy_file, change_dir, mkdir
-from easybuild.tools.filetools import rmtree2, symlink, write_file
+from easybuild.tools.filetools import change_dir, mkdir, rmtree2, symlink, write_file
 from easybuild.tools.modules import get_software_root, get_software_version
 from easybuild.tools.run import run_cmd
 from easybuild.tools.systemtools import get_shared_lib_ext
@@ -118,26 +117,20 @@ class EB_MXNet(MakeCp):
     def prepare_step(self):
         """Prepare for building and installing MXNet."""
         super(EB_MXNet, self).prepare_step()
-        self.py_ext.prepare_step()
+        self.py_ext.prepare_python()
 
     def configure_step(self):
         """Patch 'config.mk' file to use EB stuff"""
-        copy_file('make/config.mk', '.')
-
-        regex_subs = [
-            (r"export CC = gcc", r"# \g<0>"),
-            (r"export CXX = g\+\+", r"# \g<0>"),
-            (r"(?P<var>ADD_CFLAGS\s*=)\s*$", r"\g<var> %s" % os.environ['CFLAGS']),
-            (r"(?P<var>ADD_LDFLAGS\s*=)\s*$", r"\g<var> %s" % os.environ['LDFLAGS']),
-        ]
+        for (var, env_var) in [('CC', 'CC'), ('CXX', 'CXX'), ('ADD_CFLAGS', 'CFLAGS'), ('ADD_LDFLAGS', 'LDFLAGS')]:
+            self.cfg.update('buildopts', '%s="%s"' % (var, os.getenv(env_var)))
 
         toolchain_blas = self.toolchain.definition().get('BLAS', None)[0]
         if toolchain_blas == 'imkl':
             blas = "mkl"
             imkl_version = get_software_version('imkl')
             if LooseVersion(imkl_version) >= LooseVersion('17'):
-                regex_subs.append(("USE_MKL2017 = 0", "USE_MKL2017 = 1"))
-            regex_subs.append((r"(?P<var>MKLML_ROOT=).*$", r"# \g<var>%s" % os.environ["MKLROOT"]))
+                self.cfg.update('buildopts', 'USE_MKL2017=1')
+            self.cfg.update('buildopts', 'MKLML_ROOT="%s"' % os.getenv("MKLROOT"))
         elif toolchain_blas in ['ACML', 'ATLAS']:
             blas = "atlas"
         elif toolchain_blas == 'OpenBLAS':
@@ -145,12 +138,10 @@ class EB_MXNet(MakeCp):
         elif toolchain_blas is None:
             raise EasyBuildError("No BLAS library found in the toolchain")
 
-        regex_subs.append((r'USE_BLAS =.*', 'USE_BLAS = %s' % blas))
+        self.cfg.update('buildopts', 'USE_BLAS="%s"' % blas)
 
         if get_software_root('NNPACK'):
-            regex_subs.append(("USE_NNPACK = 0", "USE_NNPACK = 1"))
-
-        apply_regex_substitutions('config.mk', regex_subs)
+            self.cfg.update('buildopts', 'USE_NNPACK=1')
 
         super(EB_MXNet, self).configure_step()
 
--- a/easybuild/easyblocks/m/mxnet.py
+++ b/easybuild/easyblocks/m/mxnet.py
@@ -43,6 +43,21 @@ from easybuild.tools.modules import get_software_root, get_software_version
 from easybuild.tools.run import run_cmd
 from easybuild.tools.systemtools import get_shared_lib_ext
 
+# the namespace file for the R extension
+r_namespace = """# Export all names
+exportPattern(".")
+
+# Import all packages listed as Imports or Depends
+import(
+methods,
+Rcpp,
+DiagrammeR,
+data.table,
+jsonlite,
+magrittr,
+stringr
+)
+"""
 
 class EB_MXNet(MakeCp):
     """Easyblock to build and install MXNet"""
@@ -80,8 +95,8 @@ class EB_MXNet(MakeCp):
         else:
             raise EasyBuildError("Failed to find/isolate MXNet source directory: %s", mxnet_dirs)
 
-        for srcdir in os.listdir(self.builddir):
-            if not srcdir.startswith('mxnet-'):
+        try:
+            for srcdir in [d for d in os.listdir(self.builddir) if not d.startswith('mxnet-')]:
                 submodule, _, _ = srcdir.rpartition('-')
                 newdir = os.path.join(self.mxnet_src_dir, submodule)
                 olddir = os.path.join(self.builddir, srcdir)
@@ -91,12 +106,19 @@ class EB_MXNet(MakeCp):
                     shutil.move(olddir, newdir)
                 except IOError, err:
                     raise EasyBuildError("Failed to move %s to %s: %s", olddir, newdir, err)
+        except (IOError, OSError) as err:
+            raise EasyBuildError("Failed to move the different source directories to the main directory: %s", err)
 
         # the nnvm submodules has dmlc-core as a submodule too. Let's put a symlink in place.
         newdir = os.path.join(self.mxnet_src_dir, "nnvm", "dmlc-core")
         olddir = os.path.join(self.mxnet_src_dir, "dmlc-core")
         rmtree2(newdir)
         symlink(olddir, newdir)
+
+    def prepare_step(self):
+        """Prepare for building and installing MXNet."""
+        self.py_ext.prepare_step()
+        super(EB_MXNet, self).prepare_step()
 
     def configure_step(self):
         """Patch 'config.mk' file to use EB stuff"""
@@ -156,21 +178,7 @@ class EB_MXNet(MakeCp):
         symlink(os.path.join(self.installdir, "include"), os.path.join("inst", "include"))
 
         # MXNet doesn't provide a list of its R dependencies by default
-        namespace = """# Export all names
-exportPattern(".")
-
-# Import all packages listed as Imports or Depends
-import(
-methods,
-Rcpp,
-DiagrammeR,
-data.table,
-jsonlite,
-magrittr,
-stringr
-)
-"""
-        write_file("NAMESPACE", namespace)
+        write_file("NAMESPACE", r_namespace)
         change_dir(self.mxnet_src_dir)
         self.r_ext.prerun()
         # MXNet is just weird. To install the R extension, we have to:
@@ -186,6 +194,12 @@ stringr
 
     def sanity_check_step(self):
         """Check for main library files for MXNet"""
+        custom_paths = {
+            'files': ['lib/libmxnet.a', 'lib/libmxnet.%s' % get_shared_lib_ext()],
+            'dirs': [],
+        }
+        super(EB_MXNet, self).sanity_check_step(custom_paths=custom_paths)
+
         # for the extension we are doing the loading of the fake module ourself
         try:
             fake_mod_data = self.load_fake_module()
@@ -200,17 +214,10 @@ stringr
 
         self.clean_up_fake_module(fake_mod_data)
 
-        custom_paths = {
-            'files': ["lib/libmxnet.%s" % ext for ext in ['a', get_shared_lib_ext()]],
-            'dirs': [],
-        }
-        super(EB_MXNet, self).sanity_check_step(custom_paths=custom_paths)
-
     def make_module_extra(self, *args, **kwargs):
         """Custom variables for MXNet module."""
         txt = super(EB_MXNet, self).make_module_extra(*args, **kwargs)
 
-        self.py_ext.set_pylibdirs()
         for path in self.py_ext.all_pylibdirs:
             fullpath = os.path.join(self.installdir, path)
             # only extend $PYTHONPATH with existing, non-empty directories
--- a/easybuild/easyblocks/m/mxnet.py
+++ b/easybuild/easyblocks/m/mxnet.py
@@ -173,17 +173,26 @@ class EB_MXNet(MakeCp):
 
     def sanity_check_step(self):
         """Check for main library files for MXNet"""
+        self.py_ext.sanity_check_step()
+        self.r_ext.sanity_check_step()
+
         custom_paths = {
             'files': ["lib/libmxnet.%s" % ext for ext in ['a', get_shared_lib_ext()]],
             'dirs': [],
         }
         super(EB_MXNet, self).sanity_check_step(custom_paths=custom_paths)
-        self.py_ext.sanity_check_step()
-        self.r_ext.sanity_check_step()
 
     def make_module_extra(self, *args, **kwargs):
         """Custom variables for MXNet module."""
         txt = super(EB_MXNet, self).make_module_extra(*args, **kwargs)
-        txt += self.py_ext.make_module_extra(*args, **kwargs)
-        txt += self.r_ext.make_module_extra(*args, **kwargs)
+
+        self.py_ext.set_pylibdirs()
+        for path in self.py_ext.all_pylibdirs:
+            fullpath = os.path.join(self.installdir, path)
+            # only extend $PYTHONPATH with existing, non-empty directories
+            if os.path.exists(fullpath) and os.listdir(fullpath):
+                txt += self.module_generator.prepend_paths('PYTHONPATH', path)
+
+        txt += self.module_generator.prepend_paths("R_LIBS", [''])  # prepend R_LIBS with install path
+
         return txt
--- a/easybuild/easyblocks/m/mxnet.py
+++ b/easybuild/easyblocks/m/mxnet.py
@@ -27,6 +27,7 @@ EasyBuild support for MXNet, implemented as an easyblock
 
 @author: Ward Poelmans (Free University of Brussels)
 """
+import glob
 import os
 import shutil
 from distutils.version import LooseVersion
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/m/mxnet.py
```diff
@@ -0,0 +1,191 @@
...
+        self.r_ext.postrun()
+
+    def sanity_check_step(self):
+        """Check for main library files for MXNet"""
+        custom_paths = {
+            'files': ["lib/libmxnet.%s" % ext for ext in ['a', get_shared_lib_ext()]],
+            'dirs': [],
+        }
+        super(EB_MXNet, self).sanity_check_step(custom_paths=custom_paths)
+        self.py_ext.sanity_check_step()
```
wpoely86: This one fails:
```
== 2017-03-14 10:54:04,601 run.py:137 DEBUG run_cmd: running cmd /apps/brussel/broadwell/software/Python/2.7.12-foss-2016b/bin/python -c "import mxnet" (in /svub4/wapoelma/.local/easybuild/software/MXNet/0.9.3-foss-2016b-Python-2.7.12)
== 2017-03-14 10:54:04,601 run.py:162 INFO running cmd: /apps/brussel/broadwell/software/Python/2.7.12-foss-2016b/bin/python -c "import mxnet" 
== 2017-03-14 10:54:04,691 run.py:444 DEBUG cmd "/apps/brussel/broadwell/software/Python/2.7.12-foss-2016b/bin/python -c "import mxnet"" exited with exitcode 1 and output:
  File "<string>", line 1, in <module>
ImportError: No module named mxnet

== 2017-03-14 10:54:04,692 extension.py:163 WARNING Extension: mxnet failed to install, cmd '/apps/brussel/broadwell/software/Python/2.7.12-foss-2016b/bin/python -c "import mxnet"' (stdin: None) output: Traceback (most recent call last):
  File "<string>", line 1, in <module>
ImportError: No module named mxnet

== 2017-03-14 10:54:04,692 extensioneasyblock.py:137 WARNING Sanity check for mxnet failed: mxnet failed to install, cmd '/apps/brussel/broadwell/software/Python/2.7.12-foss-2016b/bin/python -c "import mxnet"' (stdin: None) output: Traceback (most recent call last):
  File "<string>", line 1, in <module>
ImportError: No module named mxnet

== 2017-03-14 10:54:04,692 easyconfig.py:1208 WARNING Unable to resolve template value library(%(ext_name)s) with dict {'nameletterlower': 'm', 'pyshortver': '2.7', 'versionsuffix': '-Python-2.7.12', 'rshortver': '3.3', 'toolchain_name': 'foss', 'namelower': 'mxnet', 'version_minor': '9', 'name': 'MXNet', 'versionprefix': '', 'version_major': '0', 'nameletter': 'M', 'version_major_minor': '0.9', 'pyver': '2.7.12', 'version': '0.9.3', 'rver': '3.3.3', 'toolchain_version': '2016b'}
== 2017-03-14 10:54:04,693 extensioneasyblock.py:114 DEBUG starting sanity check for extension with filter ('R -q --no-save', 'library(%(ext_name)s)')
```
but it continues nevertheless. The extension works if you load the final module so I'm guessing the issue is that the `PYTHONPATH` is not set?

easybuild/easyblocks/m/mxnet.py
```diff
@@ -0,0 +1,191 @@
...
+        self.r_ext.sanity_check_step()
+
+    def make_module_extra(self, *args, **kwargs):
+        """Custom variables for MXNet module."""
+        txt = super(EB_MXNet, self).make_module_extra(*args, **kwargs)
+        txt += self.py_ext.make_module_extra(*args, **kwargs)
+        txt += self.r_ext.make_module_extra(*args, **kwargs)
+        return txt
```
wpoely86: this gives:
```tcl
prepend-path	CPATH		$root/include
prepend-path	LD_LIBRARY_PATH		$root/lib
prepend-path	LIBRARY_PATH		$root/lib
prepend-path	PATH		$root/bin
setenv	EBROOTMXNET		"$root"
setenv	EBVERSIONMXNET		"0.9.3"
setenv	EBDEVELMXNET		"$root/easybuild/MXNet-0.9.3-foss-2016b-Python-2.7.12-easybuild-devel"

setenv	EBROOTMXNET		"$root"
setenv	EBVERSIONMXNET		None
setenv	EBDEVELMXNET		"$root/easybuild/mxnet-None-foss-2016b-Python-2.7.12-easybuild-devel"

prepend-path	PYTHONPATH		$root/lib/python2.7/site-packages
setenv	EBROOTMXNET		"$root"
setenv	EBVERSIONMXNET		None
setenv	EBDEVELMXNET		"$root/easybuild/mxnet-None-foss-2016b-Python-2.7.12-easybuild-devel"

prepend-path	R_LIBS		$root
```

Any way to use the logic from the extensions easyblock to figure out the `PYTHONPATH` and `R_LIBS` without have duplicate entries?

easybuild/easyblocks/m/mxnet.py
boegel: @wpoely86 You need to check the return value of sanity_check_step for extensions, or maybe just call the ones for the extensions first, and then call to super

easybuild/easyblocks/m/mxnet.py
boegel: The way you're using it is (clearly) not how it's intended to be used, since `make_module_extra` also defines `$EBROOT*` and `$EBVERSION*`.

You should just extend `txt` with update statements for `$PYTHONPATH` and `$R_LIBS` yourself, we don't have a better way of dealing with that unless we do some refactoring (e.g. flesh out the bit that defines `$PYTHONPATH` into a separate method that we can call, etc.)

easybuild/easyblocks/m/mxnet.py
```diff
@@ -0,0 +1,189 @@
...
+        # Find the full path to the directory that was unpacked from mxnet-*.tar.gz.
+        for srcdir in os.listdir(self.builddir):
+            if srcdir.startswith("mxnet-"):
+                self.mxnet_src_dir = os.path.join(self.builddir, srcdir)
+                break
+
+        if self.mxnet_src_dir is None:
+            raise EasyBuildError("Could not find the MXNet source directory")
+
+        self.log.debug("MXNet dir is: %s", self.mxnet_src_dir)
```
boegel:
```python
mxnet_dirs = glob.glob(os.path.join(self.builddir, 'mxnet-*'))
if len(mxnet_dirs) == 1:
    self.mxnet_src_dir = mxnet_dirs[0]
    self.log.debug("MXNet dir is: %s", self.mxnet_src_dir)
else:
    raise EasyBuildError("Failed to find/isolate MXNet source directory: %s", mxnet_dirs)

File: easybuild/easyblocks/m/mxnet.py [hunk 4]
```diff
@@ -0,0 +1,189 @@
...
+
+        for srcdir in os.listdir(self.builddir):
+            if srcdir.startswith("mxnet-"):
+                continue
+            else:
```
boegel:
```python
if not srcdir.startswith('mxnet-'):
```

or

```python
if srcdir != os.path.basename(self.mxnet_src_dir):
```

easybuild/easyblocks/m/mxnet.py
```diff
@@ -0,0 +1,189 @@
...
+                submodule, _, _ = srcdir.rpartition('-')
+                try:
+                    newdir = os.path.join(self.mxnet_src_dir, submodule)
+                    olddir = os.path.join(self.builddir, srcdir)
```
boegel: move these out of the `try`

easybuild/easyblocks/m/mxnet.py
```diff
@@ -0,0 +1,189 @@
...
+                    # first remove empty existing directory
+                    os.rmdir(newdir)
```
boegel: use http://easybuild.readthedocs.io/en/latest/api/easybuild.tools.filetools.html#easybuild.tools.filetools.rmtree2 (outside of `try`)

easybuild/easyblocks/m/mxnet.py
```diff
@@ -0,0 +1,189 @@
...
+                    shutil.move(olddir, newdir)
+                except IOError, err:
+                    raise EasyBuildError("Failed to move %s to %s: %s", olddir, newdir, err)
+
+        # the nnvm submodules has dmlc-core as a submodule too. Let's up a symlink in place.
```
boegel: `s/up/put/g`?

easybuild/easyblocks/m/mxnet.py
```diff
@@ -0,0 +1,189 @@
...
+        try:
+            newdir = os.path.join(self.mxnet_src_dir, "nnvm", "dmlc-core")
+            olddir = os.path.join(self.mxnet_src_dir, "dmlc-core")
+            os.rmdir(newdir)
```
boegel: use `rmtree2`, then there's no need to the `try` at all anymore

easybuild/easyblocks/m/mxnet.py
```diff
@@ -0,0 +1,189 @@
...
+
+    def configure_step(self):
+        """Patch 'config.mk' file to use EB stuff"""
+        copy_file('make/config.mk', '.')
+
+        regex_subs = [
+            (r"export CC = gcc", r"# \g<0>"),
+            (r"export CXX = g\+\+", r"# \g<0>"),
+            (r"(?P<var>ADD_LDFLAGS\s*=)\s*$", r"\g<var> %s" % os.environ['EBVARCFLAGS']),
+            (r"(?P<var>ADD_CFLAGS\s*=)\s*$", r"\g<var> %s" % os.environ['EBVARLDFLAGS']),
```
boegel: you have these mixed up?

also, don't use `EBVAR`, cfr. https://github.com/hpcugent/easybuild-framework/issues/1219

easybuild/easyblocks/m/mxnet.py
```diff
@@ -0,0 +1,189 @@
...
+            imkl_version = get_software_version('imkl')
+            if LooseVersion(imkl_version) >= LooseVersion('17'):
+                regex_subs.append(("USE_MKL2017 = 0", "USE_MKL2017 = 1"))
+            regex_subs.append((r"(?P<var>MKLML_ROOT=).*$", r"# \g<var>%s" % os.environ["MKLROOT"]))
+        elif toolchain_blas in ['ACML', 'ATLAS']:
+            blas = "atlas"
+        elif toolchain_blas == 'OpenBLAS':
+            blas = "openblas"
+        elif toolchain_blas is None:
+            # This toolchain has no BLAS library
```
boegel: this comment is not really needed, situation is pretty clear?

easybuild/easyblocks/m/mxnet.py
```diff
@@ -0,0 +1,189 @@
...
+        if get_software_root('NNPACK'):
+            regex_subs.append(("USE_NNPACK = 0", "USE_NNPACK = 1"))
+
+        apply_regex_substitutions('config.mk', regex_subs)
+
+        super(EB_MXNet, self).configure_step()
+
+    def install_step(self):
+        """Specify list of files to copy"""
+        self.cfg['files_to_copy'] = ['bin', 'include', 'lib',
```
boegel: if `files_to_copy` would be specified in the easyconfig, you're hard overwriting it here?

easybuild/easyblocks/m/mxnet.py
```diff
@@ -0,0 +1,189 @@
...
+        os.chdir(self.py_ext.src)
+
+        self.py_ext.prerun()
+        self.py_ext.run(unpack_src=False)
+        self.py_ext.postrun()
+
+        # next up, the R bindings
+        self.r_ext.src = os.path.join(self.mxnet_src_dir, "R-package")
+        try:
+            os.chdir(self.r_ext.src)
```
boegel: `change_dir` => no need for `try`

easybuild/easyblocks/m/mxnet.py
```diff
@@ -0,0 +1,189 @@
...
+            mkdir("inst")
+            symlink(os.path.join(self.installdir, "lib"), "inst/libs")
```
boegel: be consistent about using `os.path.join`

easybuild/easyblocks/m/mxnet.py
```diff
@@ -0,0 +1,189 @@
...
+        namespace = ["import(Rcpp)", "import(methods)", ""]
+        write_file("NAMESPACE", "\n".join(namespace))
+        self.r_ext.prerun()
+        self.r_ext.run()
+        self.r_ext.postrun()
+
+    def sanity_check_step(self):
+        """Check for main library files for MXNet"""
+        custom_paths = {
+            'files': ["lib/libmxnet.%s" % ext for ext in ['a', get_shared_lib_ext()]],
```
boegel: also check `bin`, etc.?

easybuild/easyblocks/m/mxnet.py
wpoely86: Good point. And it still compiles 😆

easybuild/easyblocks/m/mxnet.py
wpoely86: yes. You're the one always telling me to put stuff in the block :wink:

wpoely86:
This one is ready @boegel 
It's not going to win any beauty contests but it works.

easybuild/easyblocks/m/mxnet.py
```diff
@@ -0,0 +1,222 @@
...
+
+        mxnet_dirs = glob.glob(os.path.join(self.builddir, 'mxnet-*'))
+        if len(mxnet_dirs) == 1:
+            self.mxnet_src_dir = mxnet_dirs[0]
+            self.log.debug("MXNet dir is: %s", self.mxnet_src_dir)
+        else:
+            raise EasyBuildError("Failed to find/isolate MXNet source directory: %s", mxnet_dirs)
+
+        for srcdir in os.listdir(self.builddir):
+            if not srcdir.startswith('mxnet-'):
```
boegel: collapse to:

```python
for srcdir in [d for d in os.listdir(self.builddir) if not d.startswith('mxnet-')]:
```

since you're using `os.listdir` here, better wrap the whole thing in `try`/`except (IOError, OSError) as err`?

easybuild/easyblocks/m/mxnet.py
```diff
@@ -0,0 +1,222 @@
...
+                submodule, _, _ = srcdir.rpartition('-')
+                newdir = os.path.join(self.mxnet_src_dir, submodule)
+                olddir = os.path.join(self.builddir, srcdir)
+                # first remove empty existing directory
+                rmtree2(newdir)
```
boegel: make sure it's empty?

easybuild/easyblocks/m/mxnet.py
```diff
@@ -0,0 +1,222 @@
...
+                try:
+                    shutil.move(olddir, newdir)
```
boegel: why not just symlink these too like you do for `nnvm`?

easybuild/easyblocks/m/mxnet.py
```diff
@@ -0,0 +1,222 @@
...
+    def configure_step(self):
+        """Patch 'config.mk' file to use EB stuff"""
+        copy_file('make/config.mk', '.')
+
+        regex_subs = [
+            (r"export CC = gcc", r"# \g<0>"),
+            (r"export CXX = g\+\+", r"# \g<0>"),
+            (r"(?P<var>ADD_CFLAGS\s*=)\s*$", r"\g<var> %s" % os.environ['CFLAGS']),
+            (r"(?P<var>ADD_LDFLAGS\s*=)\s*$", r"\g<var> %s" % os.environ['LDFLAGS']),
+        ]
```
boegel: use `buildopts` instead to override these, no need to do runtime patching?

```python
for (var, env_var) in [('CC', 'CC'), ('CXX', 'CXX'), ('ADD_CFLAGS', 'CFLAGS'), ('ADD_LDFLAGS', 'LDFLAGS')]:
  self.update.cfg('buildopts', '%s="%s"' % (var, os.getenv(env_var)))

File: easybuild/easyblocks/m/mxnet.py [hunk 19]
```diff
@@ -0,0 +1,222 @@
...
+
+        toolchain_blas = self.toolchain.definition().get('BLAS', None)[0]
+        if toolchain_blas == 'imkl':
+            blas = "mkl"
+            imkl_version = get_software_version('imkl')
+            if LooseVersion(imkl_version) >= LooseVersion('17'):
+                regex_subs.append(("USE_MKL2017 = 0", "USE_MKL2017 = 1"))
+            regex_subs.append((r"(?P<var>MKLML_ROOT=).*$", r"# \g<var>%s" % os.environ["MKLROOT"]))
```
boegel: same here, do this via `buildopts`?

easybuild/easyblocks/m/mxnet.py
```diff
@@ -0,0 +1,222 @@
...
+        elif toolchain_blas in ['ACML', 'ATLAS']:
+            blas = "atlas"
+        elif toolchain_blas == 'OpenBLAS':
+            blas = "openblas"
+        elif toolchain_blas is None:
+            raise EasyBuildError("No BLAS library found in the toolchain")
+
+        regex_subs.append((r'USE_BLAS =.*', 'USE_BLAS = %s' % blas))
```
boegel: also via `buildopts`?

easybuild/easyblocks/m/mxnet.py
```diff
@@ -0,0 +1,222 @@
...
+
+        if get_software_root('NNPACK'):
+            regex_subs.append(("USE_NNPACK = 0", "USE_NNPACK = 1"))
```
boegel: via `buildopts`?

easybuild/easyblocks/m/mxnet.py
boegel: at least issue a warning if `files_to_copy` was specified and you're hard overwriting it now?

easybuild/easyblocks/m/mxnet.py
```diff
@@ -0,0 +1,222 @@
...
+
+        # next up, the R bindings
+        self.r_ext.src = os.path.join(self.mxnet_src_dir, "R-package")
+        change_dir(self.r_ext.src)
+        mkdir("inst")
+        symlink(os.path.join(self.installdir, "lib"), os.path.join("inst", "libs"))
+        symlink(os.path.join(self.installdir, "include"), os.path.join("inst", "include"))
+
+        # MXNet doesn't provide a list of its R dependencies by default
+        namespace = """# Export all names
```
boegel: this should be a constant defined outside of the `class`?

easybuild/easyblocks/m/mxnet.py
```diff
@@ -0,0 +1,222 @@
...
+        # - Reinstal the exported version
+        self.r_ext.run()
+        run_cmd("R_LIBS=%s Rscript -e \"require(mxnet); mxnet:::mxnet.export(\\\"R-package\\\")\"" % self.installdir)
+        change_dir(self.r_ext.src)
+        self.r_ext.run()
+        self.r_ext.postrun()
+
+    def sanity_check_step(self):
+        """Check for main library files for MXNet"""
+        # for the extension we are doing the loading of the fake module ourself
```
boegel: also mention why you are sanity checking the extensions first?

easybuild/easyblocks/m/mxnet.py
```diff
@@ -0,0 +1,222 @@
...
+        if not self.py_ext.sanity_check_step():
+            raise EasyBuildError("The sanity check of the Python bindings failed")
+
+        if not self.r_ext.sanity_check_step():
+            raise EasyBuildError("The sanity check of the R bindings failed")
+
+        self.clean_up_fake_module(fake_mod_data)
+
+        custom_paths = {
+            'files': ["lib/libmxnet.%s" % ext for ext in ['a', get_shared_lib_ext()]],
```
boegel: I'd go with

```python
'files': ['lib/libmxnet.a', 'lib/libmxnet.%s' % get_shared_lib_ext()],
```

easybuild/easyblocks/m/mxnet.py
```diff
@@ -0,0 +1,222 @@
...
+            'dirs': [],
+        }
+        super(EB_MXNet, self).sanity_check_step(custom_paths=custom_paths)
+
+    def make_module_extra(self, *args, **kwargs):
+        """Custom variables for MXNet module."""
+        txt = super(EB_MXNet, self).make_module_extra(*args, **kwargs)
+
+        self.py_ext.set_pylibdirs()
```
boegel: hmm, do this right in `prepare_step`, since it's now being done twice now (when fake module is created, and again when real module is created)

easybuild/easyblocks/m/mxnet.py
```diff
@@ -0,0 +1,222 @@
...
+        for path in self.py_ext.all_pylibdirs:
```
boegel: `s/path/pylibdir/`

easybuild/easyblocks/m/mxnet.py
wpoely86: It's always empty? Else the submodules won't work.

easybuild/easyblocks/m/mxnet.py
wpoely86: Then you need to patch the build system to respect these. Patching their own config file is the easiest way.

And we need to patch it anyway for other stuff too (see below)

easybuild/easyblocks/m/mxnet.py
wpoely86: same answer as above 😉

easybuild/easyblocks/m/mxnet.py
wpoely86: same answer as above 😉

easybuild/easyblocks/m/mxnet.py
wpoely86: same answer as above 😉

easybuild/easyblocks/m/mxnet.py
wpoely86: That's no different then any other Python Extension?

easybuild/easyblocks/m/mxnet.py
wpoely86: I guess it's a matter of taste? I move because that it what it would look like if you did real git submodules?

easybuild/easyblocks/m/mxnet.py
wpoely86: We don't do that in any of the blocks that have a similar approach?

wpoely86:
@boegel done

wpoely86:
@boegel Can you have another look at this?

easybuild/easyblocks/m/mxnet.py
boegel: well, ok then

easybuild/easyblocks/m/mxnet.py
boegel: well, ok, but we should be using a `filetools` function here...

easybuild/easyblocks/m/mxnet.py
boegel: Howso? You're deriving from `MakeCp` and not customising the `build_step`, so this is standard `make`?

easybuild/easyblocks/m/mxnet.py
boegel: Which doesn't mean we shouldn't... ;)

easybuild/easyblocks/m/mxnet.py
```diff
@@ -0,0 +1,220 @@
...
+from easybuild.easyblocks.generic.rpackage import RPackage
+from easybuild.framework.easyconfig import CUSTOM
+from easybuild.tools.build_log import EasyBuildError
+from easybuild.tools.filetools import change_dir, mkdir, rmtree2, symlink, write_file
+from easybuild.tools.modules import get_software_root, get_software_version
+from easybuild.tools.run import run_cmd
+from easybuild.tools.systemtools import get_shared_lib_ext
+
+# the namespace file for the R extension
+r_namespace = """# Export all names
```
boegel: constants should be all caps, so `R_NAMESPACE` or `MXNET_R_NAMESPACE`

easybuild/easyblocks/m/mxnet.py
```diff
@@ -0,0 +1,220 @@
...
+exportPattern(".")
+
+# Import all packages listed as Imports or Depends
+import(
+methods,
```
boegel: I'd just make this a single line? There's no big advantage here in spreading this vertically?

easybuild/easyblocks/m/mxnet.py
wpoely86: yeah maybe.

I can extend if that makes you happy

easybuild/easyblocks/m/mxnet.py
wpoely86: There isn't one.

Lets not block this PR for that. If we add it we should use it in all blocks.

boegel: I agree to not blocking this PR for it, but then we'll never end up adding it. ;)

easybuild/easyblocks/m/mxnet.py
```diff
@@ -0,0 +1,220 @@
...
+                submodule, _, _ = srcdir.rpartition('-')
+                newdir = os.path.join(self.mxnet_src_dir, submodule)
+                olddir = os.path.join(self.builddir, srcdir)
+                # first remove empty existing directory
+                rmtree2(newdir)
+                try:
+                    shutil.move(olddir, newdir)
+                except IOError, err:
+                    raise EasyBuildError("Failed to move %s to %s: %s", olddir, newdir, err)
+        except (IOError, OSError) as err:
```
boegel: why the nested `try`?

easybuild/easyblocks/m/mxnet.py
wpoely86: It's readable?

easybuild/easyblocks/m/mxnet.py
boegel: I would, just to avoid surprises if people try to specify `files_to_copy` in the easyconfig (which would get bluntly ignored right now)

easybuild/easyblocks/m/mxnet.py
```diff
@@ -0,0 +1,220 @@
...
+        write_file("NAMESPACE", r_namespace)
+        change_dir(self.mxnet_src_dir)
+        self.r_ext.prerun()
+        # MXNet is just weird. To install the R extension, we have to:
+        # - First install the extension like it is
+        # - Let R export the extension again. By doing this, all the dependencies get
+        #   correctly filled and some mappings are done
+        # - Reinstal the exported version
+        self.r_ext.run()
+        run_cmd("R_LIBS=%s Rscript -e \"require(mxnet); mxnet:::mxnet.export(\\\"R-package\\\")\"" % self.installdir)
```
boegel: wrap the cmd in single quotes to avoid the escaping (a least partially)?

easybuild/easyblocks/m/mxnet.py
```diff
@@ -0,0 +1,220 @@
...
+        super(EB_MXNet, self).sanity_check_step(custom_paths=custom_paths)
+
+        # for the extension we are doing the loading of the fake module ourself
+        try:
+            fake_mod_data = self.load_fake_module()
+        except EasyBuildError, err:
+            raise EasyBuildError("Loading fake module failed: %s", err)
+
+        if not self.py_ext.sanity_check_step():
+            raise EasyBuildError("The sanity check of the Python bindings failed")
```
boegel: `s/of/for/`?

easybuild/easyblocks/m/mxnet.py
```diff
@@ -0,0 +1,220 @@
...
+
+    def make_module_extra(self, *args, **kwargs):
+        """Custom variables for MXNet module."""
+        txt = super(EB_MXNet, self).make_module_extra(*args, **kwargs)
+
+        for path in self.py_ext.all_pylibdirs:
+            fullpath = os.path.join(self.installdir, path)
+            # only extend $PYTHONPATH with existing, non-empty directories
+            if os.path.exists(fullpath) and os.listdir(fullpath):
+                txt += self.module_generator.prepend_paths('PYTHONPATH', path)
```
boegel: this is just copy-pasting from the `PythonPackage` easyblock, we should hoist this into a separate method we can just call out to with `txt += self.py_ext.prepend_pythonpath()` or something?

or can we just all `txt += self.py_ext.make_module_extra()`?

easybuild/easyblocks/m/mxnet.py
```diff
@@ -0,0 +1,220 @@
...
+
+        txt += self.module_generator.prepend_paths("R_LIBS", [''])  # prepend R_LIBS with install path
```
boegel: same here, call out to `self.r_ext.make_module_extra()` rather than copy-pasting?

easybuild/easyblocks/m/mxnet.py
boegel: well, pfft, who cares in this case...

Also, I actually find it less readable almost :P

easybuild/easyblocks/m/mxnet.py
wpoely86: tried that. It's copy&pasted for a reason.

easybuild/easyblocks/m/mxnet.py
wpoely86: If you really insist but it was a pain to get working and I'm sure this one works.

easybuild/easyblocks/m/mxnet.py
wpoely86: Indeed, who cares? Let's keep it like it is. Doesn't make any difference

boegel:
tested with https://github.com/hpcugent/easybuild-easyconfigs/pull/4765, good to go, thanks @wpoely86!

PR #1137, opened 2017-03-16
PGI: Rename member variable to avoid conflict with member of base class

geimer:
The PGI EasyBlock currently (unintentionally) overwrites the `install_subdir` member variable of the base class `EasyBlock`, which leads to an empty `linux86-64` subdirectory in the `modules` dir.  This PR renames the member to `pgi_install_subdir` which in the end fixes this issue.


boegel:
lgtm

boegel:
Tested with a couple of PGI easyconfigs, fix confirmed, no more empty `linux86-64` subdirectory in the modules install dir.

Thanks @geimer!

PR #1139, opened 2017-03-16
SAMtools: Modifications for version 1.4

klust:
SAMtools 1.4 no longer provides the header files errmod.h and kprobaln.h nor are they needed. So the easyblock should no longer try to find these files. Tested by recompiling a 1.3.1 version and a 1.4 version, the latter with an easyconfig provided in a pull request in that repository.


boegel:
lgtm, thanks @klust!

PR #1140, opened 2017-03-16
BUG: don't set $FPATH environment variable

### LINKED ISSUE #820: easyblocks shouldn't be defining $FPATH (imkl, acml, inspector, vtune)
`$FPATH` is not what we think it may be... It's _not_ the equivalent of `$CPATH` for Fortran.

And it screws up tab completion in zsh.

cc @rtmclay

andreas-h:
fixes #820

### DIFF
--- a/easybuild/easyblocks/a/acml.py
+++ b/easybuild/easyblocks/a/acml.py
@@ -98,9 +98,8 @@ class EB_ACML(EasyBlock):
         txt += self.module_generator.set_environment('ACML_BASEDIR', basepaths[0])
         txt += self.module_generator.set_environment('ACML_BASEDIR_MT', basepaths[1])
 
-        for key in ['CPATH', 'FPATH']:
-            for path in basepaths:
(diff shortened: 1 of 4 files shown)


boegel:
I did a quick check on why we actually have this, and it seems like older version of Intel MKL actually recommended to have it set:

```
$ grep FPATH software/imkl/10.3.6.233/mkl/bin/mklvars.sh
            if [ -n "$FPATH" ]; then OLD_FPATH=":${FPATH}"; fi
            export FPATH="${MKLROOT}/include${MKL_MOD}${OLD_FPATH}"
                echo FPATH=${FPATH}
```

That being said, as long as `$CPATH` is defined to the same location, it shouldn't matter, see also https://software.intel.com/pt-br/comment/1745015 .

The nickname of Intel's Steve Lionel is "Dr. Fortran", so I trust his judgement here. :)

boegel:
Going in, thanks @andreas-h!

PR #1142, opened 2017-03-19
FFTW: --enable-avx-128-fma depends on the fma4 CPU feature (AMD), not fma

bartoldeman:
So use use_fma4 instead of use_fma, but keep the use_fma flag for backwards
compatibility if use_fma4 is not set.

### DIFF
--- a/easybuild/easyblocks/f/fftw.py
+++ b/easybuild/easyblocks/f/fftw.py
@@ -36,8 +36,8 @@ from easybuild.tools.systemtools import AARCH32, AARCH64, POWER, X86_64, get_cpu
 from easybuild.tools.toolchain.compiler import OPTARCH_GENERIC
 
 
-# AVX*, FMA, SSE2 (x86_64 only)
-FFTW_CPU_FEATURE_FLAGS_SINGLE_DOUBLE = ['avx', 'avx2', 'avx512', 'fma', 'sse2', 'vsx']
+# AVX*, FMA4 (AMD Bulldozer+ only), SSE2 (x86_64 only)
+FFTW_CPU_FEATURE_FLAGS_SINGLE_DOUBLE = ['avx', 'avx2', 'avx512', 'fma4', 'sse2', 'vsx']
 # Altivec (POWER), SSE (x86), NEON (ARM), FMA (x86_64)
 # asimd is CPU feature for extended NEON on AARCH64
 FFTW_CPU_FEATURE_FLAGS = FFTW_CPU_FEATURE_FLAGS_SINGLE_DOUBLE + ['altivec', 'asimd', 'neon', 'sse']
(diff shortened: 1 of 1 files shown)


boegel:
Hmm, I guess this explains why we had to use `use_fma = False` when building FFTW with the `intel` toolchain, see for example https://github.com/hpcugent/easybuild-easyconfigs/pull/4334/files#r105414986.

I do wonder why it seems to work fine when FFTW is build with the `gompi` toolchain on Haswell though, where `fma` is a supported CPU feature...

Do you have a reference for this @bartoldeman that states that FMA4 is required?

How did you figure this out?

bartoldeman:
You need to keep `use_fma = False` (or use `use_fma4 = False`) in the easyconfigs for `intel` since Intel compilers do not understand FMA4 (ref:
https://github.com/FFTW/fftw3/blob/master/configure.ac
) otherwise compilation will fail on AMD Bulldozers and Piledrivers.

I found this and this:
https://github.com/FFTW/fftw3/blob/master/configure.ac
```
dnl 128-bit AVX is special. There is no reason to use it on Intel processors
dnl since SSE2 is just as fast. However, on AMD processors we can both use
dnl FMA4, and 128-bit SIMD is better than 256-bit since core pairs in a
dnl compute unit can execute two 128-bit instructions independently.
AC_ARG_ENABLE(avx-128-fma, [AC_HELP_STRING([--enable-avx-128-fma],[enable AVX128/FMA optimizations])], have_avx_128_fma=$enableval, have_avx_128_fma=no)
if test "$have_avx_128_fma" = "yes"; then
        AC_DEFINE(HAVE_AVX_128_FMA,1,[Define to enable 128-bit FMA AVX optimization])
        AVX_128_FMA_CFLAGS="${AVX_CFLAGS} -mfma4"
        AC_SUBST(AVX_128_FMA_CFLAGS)
fi
AM_CONDITIONAL(HAVE_AVX_128_FMA, test "$have_avx_128_fma" = "yes")
```
You can also check:
https://github.com/FFTW/fftw3/blob/master/simd-support/avx-128-fma.c
and
https://github.com/FFTW/fftw3/blob/master/simd-support/simd-avx-128-fma.h
```
#if !(defined(__AVX__) && defined(__FMA4__)) /* sanity check */
#error "compiling simd-avx-128-fma.h without -mavx or -mfma4"
#endif
```

Yup FMA3 (cpu features flag fma) is supported on Haswell, but that is not what `--enable-avx-128-fma` enables. Possibly `--enable-avx2` does that though.

bartoldeman:
I mean this ref about compiler support:
https://en.wikipedia.org/wiki/FMA_instruction_set#Compiler_and_assembler_support
this explains why GCC >= 4.5 has no trouble with `--enable-avx-128-fma`. It's just that if you enable that on Haswell you are essentially cross-compiling that code.

boegel:
@bartoldeman OK, thanks for clarifying.

I would handle the backward-compatibility w.r.t. `use_fma` a bit different, to make it very clear it should not be used anymore, see https://github.com/bartoldeman/easybuild-easyblocks/pull/3 .

The FFTW easyconfigs that use `use_fma` now should be updated to use `use_fma4` instead (and maybe the comment why it's hard disabled should be clarified too).

Also, the easyconfigs for FFTW 3.3.3 and 3.3.4 that use the `intel` or `ictce` toolchains should also hard disable `use_fma4`, no?

boegel:
@bartoldeman OK, thanks for clarifying.

I would handle the backward-compatibility w.r.t. `use_fma` a bit different, to make it very clear it should not be used anymore, see https://github.com/bartoldeman/easybuild-easyblocks/pull/3 .

The FFTW easyconfigs that use `use_fma` now should be updated to use `use_fma4` instead (and maybe the comment why it's hard disabled should be clarified too).

Also, the easyconfigs for FFTW 3.3.3 and 3.3.4 that use the `intel` or `ictce` toolchains should also hard disable `use_fma4`, no?

bartoldeman:
Thanks, yes that is better.
FFTW 3.3.[34] easyconfigs don't use this easyblock (using ConfigureMake) so are not affected.

boegel:
Tested with https://github.com/hpcugent/easybuild-easyconfigs/pull/4384, good to go, thanks @bartoldeman!

PR #1143, opened 2017-03-20
fix 'prefix' subdirectory for older versions of icc/ifort

boegel:
also add '`which <compilername>`' as sanity check command

fix for bug introduced in #1126, where the subdirectory where `icc` and `ifort` was no longer included in `$PATH` for versions where `bin/intel64` does not exist but `composer_xe_<version>/bin/intel64` does, which includes versions 2013_sp* and 2015.*...

the value for `prefix` was wrong all along, but this was never noticed since the non-prefixed subdirectories were also being included, including `bin`

serves me right for doing last-minute merges ;)

cc @bartoldeman

### DIFF
--- a/easybuild/easyblocks/i/icc.py
+++ b/easybuild/easyblocks/i/icc.py
@@ -121,7 +121,9 @@ class EB_icc(IntelBase):
             'dirs': [],
         }
 
-        super(EB_icc, self).sanity_check_step(custom_paths=custom_paths)
+        custom_commands = ["which icc"]
+
+        super(EB_icc, self).sanity_check_step(custom_paths=custom_paths, custom_commands=custom_commands)
 
     def make_module_req_guess(self):
         """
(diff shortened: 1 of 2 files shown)


bartoldeman:
Right, only the top-levels were included in my existing easyconfigs for modules for 2015.*...

I checked some of our folders and found that old Intel compilers used -'s instead of underscores though. Looks like it changed here:
```
2011-10-12.1.4/composerxe
2011-10-12.1.4/composer_xe_2011_sp1
2011-10-12.1.4/composer_xe_2011_sp1.10.319
2011-4-12.0.4/composerxe
2011-4-12.0.4/composerxe-2011
2011-4-12.0.4/composerxe-2011.4.191
```

bartoldeman:
so we need to double-check the icc-11* and icc-2011* easyconfigs...

boegel:
For `icc-11.*.eb` and `icc-2011.*.eb, there is no `composer*` subdirectory, but there is a `bin/intel64`, so that's fine.

bartoldeman:
Some more history searching:

* oldest commit 63edc6f uses `composer-xe` which was correct at the time (April 2012). So newer compilers only had `/bin`(this is what I have for our production icc/2015.3.187 since that module was installed Summer 2015).

* commit 15b48de5 (Sep 2015) changes that to from composer-xe to composer_xe_.
So with this change both shallow and deeper directories were included.

* commit  86980d0 (Dec 2015) changed from composer_xe_ to composer-xe- which is now corrected to the behaviour of Sep 2015 (except either shallow or deep is done but not both).

PR #1146, opened 2017-03-21
enhance Bowtie2 easyblock: set correct build options, copy more files, extend sanity check

boegel:
Triggered by addition of `RELEASE_FLAGS` in `buildopts` by @klust in https://github.com/hpcugent/easybuild-easyconfigs/pull/4372

This makes the Bowtie2 installations across toolchains consistent, and allows to clean up the Bowtie2 easyconfigs significantly (cfr. https://github.com/hpcugent/easybuild-easyconfigs/pull/4380)

### DIFF
--- a/easybuild/easyblocks/b/bowtie2.py
+++ b/easybuild/easyblocks/b/bowtie2.py
@@ -16,12 +16,13 @@ EasyBuild support for building and installing Bowtie2, implemented as an easyblo
 @author: Fotis Georgatos (Uni.Lu)
 @author: Kenneth Hoste (Ghent University)
 """
-
+from distutils.version import LooseVersion
 import os
 import shutil
 
 from easybuild.easyblocks.generic.configuremake import ConfigureMake
 from easybuild.tools.build_log import EasyBuildError
+from easybuild.tools.filetools import copy_file
 
 
 class EB_Bowtie2(ConfigureMake):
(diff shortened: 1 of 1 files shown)


boegel:
@wpoely86 please review?

easybuild/easyblocks/b/bowtie2.py
```diff
@@ -30,35 +31,60 @@ class EB_Bowtie2(ConfigureMake):
...
+
+        cc = os.getenv('CC')
+        if cc:
+            self.cfg.update('buildopts', 'CC="%s"' % cc)
+
+        cxx = os.getenv('CXX')
+        if cxx:
+            self.cfg.update('buildopts', 'CPP="%s" CXX="%s"' % (cxx, cxx))
+
+        cxxflags = os.getenv('CXXFLAGS')
```
wpoely86: no `CFLAGS`? Or `LDFLAGS`?

easybuild/easyblocks/b/bowtie2.py
```diff
@@ -30,35 +31,60 @@ class EB_Bowtie2(ConfigureMake):
...
-        except OSError, err:
-            raise EasyBuildError("Copying %s to installation dir %s failed: %s", srcfile, destdir, err)
+        for filename in self.bowtie2_files:
+            copy_file(os.path.join(self.cfg['start_dir'], filename), os.path.join(self.installdir, 'bin', filename))
+
+        for dirname in ['doc', 'example', 'scripts']:
+            srcdir = os.path.join(self.cfg['start_dir'], dirname)
+            targetdir = os.path.join(self.installdir, dirname)
+            try:
+                shutil.copytree(srcdir, targetdir)
```
wpoely86: we have a `copytree`?

easybuild/easyblocks/b/bowtie2.py
```diff
@@ -16,12 +16,13 @@
...
+from distutils.version import LooseVersion
 import os
 import shutil
 
 from easybuild.easyblocks.generic.configuremake import ConfigureMake
 from easybuild.tools.build_log import EasyBuildError
+from easybuild.tools.filetools import copy_file
 
 
 class EB_Bowtie2(ConfigureMake):
```
wpoely86: why not `MakeCp`?

easybuild/easyblocks/b/bowtie2.py
boegel: no, this is just about making sure the optimisation compiler flags specified by EasyBuild are used, rather than the hardcoded defaults

easybuild/easyblocks/b/bowtie2.py
boegel: because then I have to fiddle with the mandatory `files_to_copy` easyconfig parameter, not worth the trouble here imho

easybuild/easyblocks/b/bowtie2.py
boegel: yeah, but we should probably get rid of it, it was there to use it when we were still supporting Python 2.4

Instead, we should have a `copy_dir` function that wraps around `shutil.copytree` or something...

wpoely86: also fine by me. Just not directly shutil.copytree

easybuild/easyblocks/b/bowtie2.py
wpoely86: So you basically recreate the `MakeCp` block? It's definitely worth the trouble not recreating existing blocks.

boegel: `MakeCp` implements a generic procedure where files specified in `files_to_copy` are being copied

Here we implement a custom `install_step` for Bowtie2 that copies very specific files/directories, so whether we derive from `ConfigureMake` or `MakeCp` doesn't make a difference, except that if we derive from `MakeCp` we have to take extra action to avoid that `files_to_copy` *has* to be specified (while it doesn't make sense to specify it in this specific case).

wpoely86: We have a whole bunch of blocks that already do this (deriving from `MakeCp` and filling in the `files_to_copy`) so I don't see why this should be any different. Any future enhancements made to `MakeCp` will also need to be ported to this block. 

Now you have to take the action of letting the configure_step be `pass` while in the MakeCp case you need to specify that `files_to_copy` is not mandatory in the easyconfig. So I don't see much difference in that regard.

boegel: Fixed, please rereview?

boegel:
tested with https://github.com/hpcugent/easybuild-easyconfigs/pull/4380, good to go, thanks for the review @wpoely86!

PR #879, opened 2016-03-29
EasyBlock for VMD

thiell:
http://www.ks.uiuc.edu/Research/vmd/
Originally installed with EasyBuild 2.3

PR #1148, opened 2017-03-23
easyblock for VMD

boegel:
This is a cleaned up/enhanced version of the VMD easyblock contributed by @thiell in #879, works with VMD 1.9.3.

The CUDA/OptiX support added by @thiell should still work, but this has not been tested.

### DIFF
--- a/easybuild/easyblocks/v/vmd.py
+++ b/easybuild/easyblocks/v/vmd.py
@@ -1,0 +1,208 @@
+##
+# Copyright 2009-2015 Ghent University
+# Copyright 2015-2016 Stanford University
+#
+# This file is part of EasyBuild,
+# originally created by the HPC team of Ghent University (http://ugent.be/hpc/en),
+# with support of Ghent University (http://ugent.be/hpc),
+# the Flemish Supercomputer Centre (VSC) (https://vscentrum.be/nl/en),
+# the Hercules foundation (http://www.herculesstichting.be/in_English)
+# and the Department of Economy, Science and Innovation (EWI) (http://www.ewi-vlaanderen.be/en).
+#
+# http://github.com/hpcugent/easybuild
+#
+# EasyBuild is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation v2.
+#
+# EasyBuild is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/v/vmd.py
```diff
@@ -0,0 +1,159 @@
...
+        # LP64: build VMD as 64-bit binary
+        # IMD: enable support for Interactive Molecular Dynamics (e.g. to connect to NAMD for remote simulations)
+        # PTHREADS: enable support for POSIX threads
+        # COLVARS: enable support for collective variables (related to NAMD/LAMMPS)
+        # NOSILENT: verbose build command
+        self.cfg.update('configopts', "LINUXAMD64 LP64 IMD PTHREADS COLVARS NOSILENT")
+
+        # add additional configopts based on available dependencies
+        for key in deps:
+            if deps[key]:
```
wpoely86: why this `if`? Can it not be avoided?

easybuild/easyblocks/v/vmd.py
```diff
@@ -0,0 +1,159 @@
...
+            env.setvar('NUMPY_INCLUDE_DIR', out.strip())
+
+        # compiler commands
+        self.cfg.update('buildopts', 'CC="%s"' % os.getenv('CC'))
+        self.cfg.update('buildopts', 'CCPP="%s"' % os.getenv('CXX'))
+
+        # source tarballs contains a 'plugins' and 'vmd-<version>' directory
+        vmddir = os.path.join(self.cfg['start_dir'], '%s-%s' % (self.name.lower(), self.version))
+
+        # first, build plugins
```
wpoely86: you have to explain this one

easybuild/easyblocks/v/vmd.py
boegel: some dependencies are optional, so some of these may be `None`

easybuild/easyblocks/v/vmd.py
boegel: well, I can provide a reference to the VMD docs, but that's how it should be done, see http://www.ks.uiuc.edu/Research/vmd/doxygen/compiling.html or http://robinbetz.com/blog/2015/01/08/compiling-vmd-with-python-support/

boegel:
Tested with https://github.com/hpcugent/easybuild-easyconfigs/pull/4391, good to go in as revised version of #879. Thanks for the review @wpoely86!

PR #1150, opened 2017-03-23
allow libpng as OS dependency for WPS

mboisson:
Need to define libpnglib and libpnginc to an empty value first, because they are used later. If the libpng and zlib are not found as module, expect to have them find as an OS package.

### DIFF
--- a/easybuild/easyblocks/w/wps.py
+++ b/easybuild/easyblocks/w/wps.py
@@ -110,6 +110,8 @@ class EB_WPS(EasyBlock):
         # libpng dependency check
         libpng = get_software_root('libpng')
         zlib = get_software_root('zlib')
+        libpnglib = ""
+        libpnginc = ""
         if libpng:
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/w/wps.py
```diff
@@ -110,14 +110,14 @@ def configure_step(self):
         # libpng dependency check
         libpng = get_software_root('libpng')
         zlib = get_software_root('zlib')
+        libpnglib = ""
+        libpnginc = ""
```
boegel: @mboisson I'd throw this in the (original) `else` part, with a clear comment like

```
# define these as empty, assume that libpng will be available via OS (e.g. due to --filter-deps=libpng)
```

mboisson: done

boegel:
lgtm, thanks @mboisson!

PR #1155, opened 2017-03-29
Add block for building the OpenKIM library.

schiotz:
The Open Knowledgebase of Interatomic Models (openkim.org) is a library
of interatomic models that can be used by some simulation packages, including
LAMMPS and Asap.  An easybuild of the latter will be contributed shortly.

This easyblock compiles the library itself; then it downloads and compiles
all models currently available from the OpenKIM project.  Finally, it installs
everything and sets an environment variable pointing to the models.

### DIFF
--- a/easybuild/easyblocks/o/openkim.py
+++ b/easybuild/easyblocks/o/openkim.py
@@ -1,5 +1,5 @@
 ##
-# Copyright 2009-2016 Ghent University
+# Copyright 2009-2017 Ghent University
 #
 # This file is part of EasyBuild,
 # originally created by the HPC team of Ghent University (http://ugent.be/hpc/en),
@@ -34,14 +34,14 @@ from easybuild.tools.config import build_option
 import urllib2
 import json
 
-class OpenKIM(ConfigureMake):
+class EB_OpenKIM(ConfigureMake):
     def build_step(self, verbose=False, path=None):
         """
         Start the actual build
         """
         # Build the API.
         self.log.info("Building the OpenKIM API")
-        out1 = ConfigureMake.build_step(self, verbose=verbose, path=path)
+        out1 = super(EB_OpenKIM, self).build_step(verbose=verbose, path=path)
         
         # Download all models
         self.log.info("Downloading all OpenKIM models")
@@ -49,7 +49,7 @@ class OpenKIM(ConfigureMake):
 
         # Build the models
         self.log.info("Building the OpenKIM models")
-        out3 = ConfigureMake.build_step(self, verbose=verbose, path=path)
+        out3 = super(EB_OpenKIM, self).build_step(verbose=verbose, path=path)
         
         return out1+out2+out3
 
--- a/easybuild/easyblocks/o/openkim.py
+++ b/easybuild/easyblocks/o/openkim.py
@@ -1,0 +1,82 @@
+##
+# Copyright 2009-2016 Ghent University
+#
+# This file is part of EasyBuild,
+# originally created by the HPC team of Ghent University (http://ugent.be/hpc/en),
+# with support of Ghent University (http://ugent.be/hpc),
+# the Flemish Supercomputer Centre (VSC) (https://www.vscentrum.be),
+# Flemish Research Foundation (FWO) (http://www.fwo.be/en)
+# and the Department of Economy, Science and Innovation (EWI) (http://www.ewi-vlaanderen.be/en).
+#
+# http://github.com/hpcugent/easybuild
+#
+# EasyBuild is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation v2.
+#
+# EasyBuild is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with EasyBuild.  If not, see <http://www.gnu.org/licenses/>.
+##
+"""
+EasyBuild support for building and installing Python, implemented as an easyblock
+
+@author: Jakob Schiotz (Tech. Univ. Denmark)
+"""
+
+from easybuild.easyblocks.generic.configuremake import ConfigureMake
+from easybuild.tools.run import run_cmd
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/o/openkim.py
```diff
@@ -0,0 +1,82 @@
+##
+# Copyright 2009-2016 Ghent University
```
boegel: 2017 :)

easybuild/easyblocks/o/openkim.py
```diff
@@ -0,0 +1,82 @@
...
+import json
+
+class OpenKIM(ConfigureMake):
+    def build_step(self, verbose=False, path=None):
+        """
+        Start the actual build
+        """
+        # Build the API.
+        self.log.info("Building the OpenKIM API")
+        out1 = ConfigureMake.build_step(self, verbose=verbose, path=path)
```
boegel: should be `super(EB_OpenKIM, self).build_step(...)`

easybuild/easyblocks/o/openkim.py
```diff
@@ -0,0 +1,82 @@
...
+@author: Jakob Schiotz (Tech. Univ. Denmark)
+"""
+
+from easybuild.easyblocks.generic.configuremake import ConfigureMake
+from easybuild.tools.run import run_cmd
+from easybuild.tools.config import build_option
+import urllib2
+import json
+
+class OpenKIM(ConfigureMake):
```
boegel: should be `EB_OpenKIM`, otherwise EasyBuild will not pick up this easyblock automatically (see http://easybuild.readthedocs.io/en/latest/Implementing-easyblocks.html#class-name-for-software-specific-easyblocks)

easybuild/easyblocks/o/openkim.py
```diff
@@ -0,0 +1,82 @@
...
+        """
+        Start the actual build
+        """
+        # Build the API.
+        self.log.info("Building the OpenKIM API")
+        out1 = ConfigureMake.build_step(self, verbose=verbose, path=path)
+        
+        # Download all models
+        self.log.info("Downloading all OpenKIM models")
+        out2 = self.download_kim_models(verbose, path)
```
boegel: can we avoid downloading on the fly by just listing these in `sources`?

easybuild/easyblocks/o/openkim.py
```diff
@@ -0,0 +1,82 @@
...
+        # We have to do this one model at a time.
+        query = '&fields={"kimcode":1,"kim-api-version":1}&database=obj&query={"type":"mo"}'
+        url = 'https://query.openkim.org/api'
+        timeout = build_option('download_timeout')
+        if timeout is None:
+            # default to 10sec timeout if none was specified
+            # default system timeout may be infinite (?)
+            timeout = 10        
+        # We should maybe do some error handling here?
+        url_fd = urllib2.urlopen(url, data=query, timeout=timeout)
```
boegel: should use `download_file` from `easybuild.tools.filetools`?

but it's better to not download on the fly and provide these via `sources` instead?

easybuild/easyblocks/o/openkim.py
schiotz: I wish we could.  The OpenKIM project does not make download of all models available in a single tar file; instead you can request a list of model names from their database, and then get each model one at a time as a tar file.  Fortunately, their makefile handles this (including dependencies between "Models" and "Model Drivers").

Thank you for taking the time to review this.  I will fix the things in the other comments.

easybuild/easyblocks/o/openkim.py
schiotz: This is not actually downloading a file, but getting a list of model names from a database.

I could probably use download_file to download the list into a temporary file, and then read that file.  Would that be preferable?

schiotz:
@boegel I have followed your suggestions above, except for using download_file instead of urllib2.urlopen.  It is not a file download, but a database query.  

Nevertheless, it would probably make sense to benefit from the error handling etc in download_file, so I tried to use it to download to a temporary file "openkim_model_names.txt".  As download_file does not take a separate data argument for the database query, and as the query just looks like URL arguments, I just appended it to the URL itself.  

Unfortunately, that caused an error in one of the functions download_file calls (see below).  Either this means that download_file is not the right tool for this, or I do something stupid. :-)

```
== 2017-04-05 09:21:18,795 build_log.py:147 ERROR EasyBuild crashed with an error (at ?:124 in __init__): Unexpected error occurred when trying to download openkim_model_names.txt to .: unknown url type: openkim_model_names.txt (at easybuild/tools/filetools.py:390 in download_file)
== 2017-04-05 09:21:18,795 easyblock.py:2518 WARNING build failed (first 300 chars): Unexpected error occurred when trying to download openkim_model_names.txt to .: unknown url type: openkim_model_names.txt
== 2017-04-05 09:21:18,796 easyblock.py:276 INFO Closing log for application name OpenKIM version 1.7.3
```

easybuild/easyblocks/o/openkim.py
boegel: This kind of goes against the idea of ensuring reproducible installations...

If someone uses the same easyconfig later, they'll likely end up with a different installation.

Does it make sense to hardcode the list of models in the easyconfig file rather than doing this database query?

And if we do, will stuff be downloaded on the fly via the `make add-<model>` commands being run? Can we provide the sources rather than having it download on the fly?

schiotz: @boegel 
I have discussed this with the OpenKIM developers.  I think the solution is to list all modules in the easyconfig, and let it download all these tar files.  I am unsure of the correct way of doing this.  There will be two lists of .tgz files, they need to be unpacked into two different places in the source tree.  

What is the correct way to do this?  I could define two lists in the easyconfig, and redefine the method unpacking the sources in the easyblock to unpack everything the right place.  Would this be an acceptable solution?  There would literally be hundreds of source archived, is that a problem?

It would be nice if all models could be downloaded as one tgz file, but they cannot do that since the models have different licenses, and may not be distributed together.

schiotz:
@boegel 
I would like to withdraw this pull request.  As I worked on your suggestion to include all KIM model source files in the EasyConfig (for reproducibility), I gradually needed less and less code in EB_OpenKIM, until I realized that I could do everything in the configuration block.  I will update https://github.com/hpcugent/easybuild-easyconfigs/pull/4411 in a moment.

PR #1156, opened 2017-03-30
enhance Python sanity check to check for Tkinter support if Tk is included as a dependency

boegel:
sort of related to https://github.com/hpcugent/easybuild-easyconfigs/pull/4260, where we are no longer included `Tk` as a dependency to Python

I'm deliberately not enforcing the `Tkinter` is *not* there if `Tk` is not included as a dependency, since `Tk` may be picked up from the OS (which may be deliberate via `--filter-deps`).


boegel:
Tested with handful of existing Python easyconfigs + https://github.com/hpcugent/easybuild-easyconfigs/pull/4260, works fine as expected

boegel:

PR #1158, opened 2017-03-30
fix Tkinter checks for Python 3.x, change check to see whether Tkinter should be built

boegel:
follow up for #1156

cc @wpoely86

### DIFF
--- a/easybuild/easyblocks/p/python.py
+++ b/easybuild/easyblocks/p/python.py
@@ -152,12 +152,6 @@ class EB_Python(ConfigureMake):
         pyver = 'python' + '.'.join(self.version.split('.')[:2])
         shlib_ext = get_shared_lib_ext()
 
-        if '--with-tcltk' in self.cfg['configopts']:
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/p/python.py
```diff
@@ -156,6 +150,13 @@ def sanity_check_step(self):
         """Custom sanity check for Python."""
 
         pyver = 'python' + '.'.join(self.version.split('.')[:2])
+        shlib_ext = get_shared_lib_ext()
+
+        if '--with-tcltk' in self.cfg['configopts']:
```
wpoely86: why this change?

boegel: Well, it's a better way to check if Tkinter should be there or not; it gives the flexibility to add `--with-tcltk` to `configopts` in the easyconfig itself without having `Tk` as a dependency, e.g. to use the system Tk.

wpoely86: well, we don't usually count on any system dep

boegel:
Thanks for the review @wpoely86, verified that this fixes the sanity check for Python v3.x, so going in.

PR #1153, opened 2017-03-28
add easyblock for WRF-Fire

boegel:

PR #1159, opened 2017-03-31
also build WPS part of WRF-Fire

boegel:
follow-up for #1153


boegel:

PR #1160, opened 2017-03-31
Preinstallopts for binary and packed binary easyblocks

damianam:
These easyblocks were missing (and silently ignoring) preinstallopts. It might be necessary to delete stuff prior to copying to the installation directory. Of course it can be done too with `postinstallcmds`, but this option can affect the module generator for instance, if the library dir is deleted.

### DIFF
--- a/easybuild/easyblocks/generic/binary.py
+++ b/easybuild/easyblocks/generic/binary.py
@@ -98,6 +98,11 @@ class Binary(EasyBlock):
         """Copy all files in build directory to the install directory"""
         if self.cfg['install_cmd'] is None:
             try:
+                cmd = self.cfg['preinstallopts']
+                if cmd:
+                    self.log.info("Running preinstallopts of %s using command '%s'..." % (self.name, cmd))
+                    run_cmd(cmd, log_all=True, simple=True)
+                self.log.info("Copying %s into '%s'..." % (self.cfg['start_dir'], self.installdir))
                 # shutil.copytree doesn't allow the target directory to exist already
                 rmtree2(self.installdir)
                 shutil.copytree(self.cfg['start_dir'], self.installdir, symlinks=self.cfg['keepsymlinks'])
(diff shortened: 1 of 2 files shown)


easybuild/easyblocks/generic/binary.py
```diff
@@ -98,6 +98,11 @@ def install_step(self):
         """Copy all files in build directory to the install directory"""
         if self.cfg['install_cmd'] is None:
             try:
+                cmd = self.cfg['preinstallopts']
+                if cmd:
+                    self.log.info("Running preinstallopts of %s using command '%s'..." % (self.name, cmd))
+                    run_cmd(cmd, log_all=True, simple=True)
+                self.log.info("Copying %s into '%s'..." % (self.cfg['start_dir'], self.installdir))
```
boegel: hmm, this is not how `preinstallopts` is usually being used

not that this also mismatches the name, these are not 'options', but an actual command (granted, something like `foo &&` isn't an option either, `pre*opts` gets sort of abused easily for that)

damianam: Yes, I know, normally you have something like:

```
cmd = ' '.join([self.cfg['preinstallopts'], self.cfg['install_cmd'], self.cfg['installopts']])
self.log.info("Installing %s using command '%s'..." % (self.name, cmd))
```

Which 90% of the time gets "abused" and becomes `foo &&` as you said. But regardless of if this is abuse or not, it should be consistent. Otherwise you have `preinstallopts` silently ignored.

boegel: Well, I'm not sure if this is consistent... `preinstallopts` is for "options to the install command", which is not what you are doing here...

damianam: I'd say that "options to the install command" describes better `installopts` than `preinstallopts`

damianam:
@boegel ping?

easybuild/easyblocks/generic/binary.py
boegel: Well, `preinstallopts` is for options that go *before* the install command.

For example `preinstallopts = 'PREFIX=%(installdir)s'`.

boegel:
@damianam I think introducing a custom `preinstall_cmds` easyconfig parameter is more appropriate here?

It would also be nice to try and avoid code duplication, for example by defining a `run_preinstall_cmds` method in `Binary` that you can call out to in `install_step` in both `Binary` and `PackedBinary`?

boegel:
@damianam Thoughts on my last suggestion?

PR #1154, opened 2017-03-28
add easyblock for Doris

boegel:

PR #1161, opened 2017-04-05
also build SARtools and Envisat tools for Doris

boegel:
follow-up for #1154

### DIFF
--- a/easybuild/easyblocks/d/doris.py
+++ b/easybuild/easyblocks/d/doris.py
@@ -48,10 +48,10 @@ class EB_Doris(ConfigureMake):
         # create installation directory (and /bin subdirectory) early, make sure it doesn't get removed later
         self.make_installdir()
         mkdir(os.path.join(self.installdir, 'bin'))
-        self.cfg["keeppreviousinstall"] = True
+        self.cfg['keeppreviousinstall'] = True
 
         # configure/build/install should be done from 'src' subdirectory
-        change_dir('src')
+        change_dir(os.path.join(self.cfg['start_dir'], 'src'))
 
         qa = {
             "===> Press enter to continue.": '',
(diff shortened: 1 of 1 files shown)


boegel:
@wpoely86 please review?

easybuild/easyblocks/d/doris.py
```diff
@@ -73,17 +73,63 @@ def configure_step(self):
...
 
+        super(EB_Doris, self).build_step()
+
+        # build SARtools
+        change_dir(os.path.join(self.cfg['start_dir'], 'SARtools'))
+
+        self.cfg['buildopts'] = common_buildopts
+        self.cfg.update('buildopts', 'CC="%s"' % os.getenv('CXX'))
+        cflags = os.getenv('CFLAGS') + " -D_FILE_OFFSET_BITS=64 -D_LARGEFILE_SOURCE -D_LARGEFILE64_SOURCE"
+        self.cfg.update('buildopts', 'CFLAGS="%s"' % cflags)
```
wpoely86: `CXXFLAGS`?

easybuild/easyblocks/d/doris.py
```diff
@@ -73,17 +73,63 @@ def configure_step(self):
...
+        self.cfg.update('installopts', 'CFLAGS="%s"' % os.getenv('CFLAGS'))
+        super(EB_Doris, self).install_step()
+
     def sanity_check_step(self):
         """Custom sanity check for Doris."""
         custom_paths = {
-            'files': ['bin/cpx2ps', 'bin/doris', 'bin/plotcpm', 'bin/run'],
+            'files': [os.path.join('bin', x) for x in ['cpx2ps', 'doris', 'plotcpm', 'run']] +
+                     [os.path.join('bin', x) for x in ['bkconvert', 'cpxfiddle', 'flapjack', 'floatmult', 'wrap']] +
+                     [os.path.join('bin', x) for x in ['envisat_dump_header', 'envisat_dump_data']],
```
wpoely86: why not put them all in a list variable outside this dict?

easybuild/easyblocks/d/doris.py
boegel: well, I need to define `CFLAGS`, but I should use the `$CXXFLAGS` value as defined by EasyBuild here, yes

easybuild/easyblocks/d/doris.py
boegel: I grouped them by origin (Doris vs SARtools vs ENVISAT_TOOLS), I'll include a comment to make that clear

boegel:

PR #1162, opened 2017-04-06
Fix problems when usempi not defined in toolchain

ppomorsk:
Many common build configurations of NAMD do not use MPI.  This change makes namd.py work with toolchains without MPI.

### DIFF
--- a/easybuild/easyblocks/n/namd.py
+++ b/easybuild/easyblocks/n/namd.py
@@ -59,7 +59,7 @@ class EB_NAMD(MakeCp):
 
         # complete Charm ++ and NAMD architecture string with compiler family
         comp_fam = self.toolchain.comp_family()
-        if self.toolchain.options['usempi']:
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/n/namd.py
```diff
@@ -59,7 +59,7 @@ def configure_step(self):
 
         # complete Charm ++ and NAMD architecture string with compiler family
         comp_fam = self.toolchain.comp_family()
-        if self.toolchain.options['usempi']:
+        if 'usempi' in self.toolchain.options and self.toolchain.options['usempi']:
```
boegel: I'd rather go with something like:

```python
if self.toolchain.options.get('usempi', False):
```

which is already being used in other easyblocks

boegel:
@ppomorsk pull requests should be opened targeting `develop` rather than `master`, I've changed the target branch

boegel:
Going in, thanks @ppomorsk!

PR #1163, opened 2017-04-11
GROMACS: build for target architecture based on optarch

ostueker:
Gromac's CMake tries persistently to detect the CPU capabilities on the build node, which makes
problems if the target platform is older than that of the build node.

This patch therefore selects appopriate GMX_SIMD cmake flags based on flags present in the optarch
string (originating from `EASYBUILD_OPTARCH`).  If no known flags are found, it falls back to cmake's auto detection of capabilities on the build node.

### DIFF
--- a/easybuild/easyblocks/g/gromacs.py
+++ b/easybuild/easyblocks/g/gromacs.py
@@ -89,7 +89,9 @@ class EB_GROMACS(CMakeMake):
             comp_fam = self.toolchain.comp_family()
             optarch = optarch.get(comp_fam, '').upper()
 
-        if 'AVX2' in optarch and LooseVersion(self.version) >= LooseVersion('5.0'):
+        if 'AVX512' in optarch and LooseVersion(self.version) >= LooseVersion('2016'):
+            res = 'AVX_512'
+        elif 'AVX2' in optarch and LooseVersion(self.version) >= LooseVersion('5.0'):
             res = 'AVX2_256'
         elif 'AVX' in optarch:
             res = 'AVX_256'
--- a/easybuild/easyblocks/g/gromacs.py
+++ b/easybuild/easyblocks/g/gromacs.py
@@ -124,9 +124,11 @@ class EB_GROMACS(CMakeMake):
             # disable GUI tools
             self.cfg.update('configopts', "-DGMX_X11=OFF")
 
-            # convince to build for an older architecture then present on the build node by setting GMX_SIMD CMake flag.
-            if build_option('optarch', default=None) and self.toolchain.comp_family() in build_option('optarch'):
-                build_arch = build_option('optarch')[self.toolchain.comp_family()]
+            # convince to build for an older architecture than present on the build node by setting GMX_SIMD CMake flag.
+            optarch = build_option('optarch', default=None)
+            comp_fam = self.toolchain.comp_family()
+            if isinstance(optarch, dict) and comp_fam in build_option('optarch'):
+                build_arch = optarch[comp_fam]
                 gmx_simd = self.get_gromacs_arch(build_arch)
                 if gmx_simd:
                     if LooseVersion(self.version) < LooseVersion('5.0'):
@@ -378,11 +380,11 @@ class EB_GROMACS(CMakeMake):
         [2] http://www.gromacs.org/Documentation/Acceleration_and_parallelization
         """
         build_arch = build_arch.upper()
-        if build_arch.endswith(("AVX2")) and not LooseVersion(self.version) < LooseVersion('5.0'):
+        if 'AVX2' in build_arch and LooseVersion(self.version) >= LooseVersion('5.0'):
             return "AVX2_256"
-        elif build_arch.endswith(("AVX")):
+        elif 'AVX' in build_arch:
             return "AVX_256"
-        elif build_arch.endswith(("SSE3" ,"SSE2") or build_arch.contains("MARCH=NOCONA")):
+        elif 'SSE3' in build_arch  or  'SSE2' in build_arch  or  'MARCH=NOCONA' in build_arch:
             # Gromacs doesn't have any GMX_SIMD=SSE3 but only SSE2 and SSE4.1 [1].
             # According to [2] the performance difference between SSE2 and SSE4.1 is minor on x86
             # and SSE4.1 is not supported by AMD Magny-Cours[1].
--- a/easybuild/easyblocks/g/gromacs.py
+++ b/easybuild/easyblocks/g/gromacs.py
@@ -44,6 +44,7 @@ from easybuild.easyblocks.generic.configuremake import ConfigureMake
 from easybuild.easyblocks.generic.cmakemake import CMakeMake
 from easybuild.framework.easyconfig import CUSTOM
 from easybuild.tools.build_log import EasyBuildError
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/g/gromacs.py
```diff
@@ -123,6 +124,16 @@ def configure_step(self):
             # disable GUI tools
             self.cfg.update('configopts', "-DGMX_X11=OFF")
 
+            # convince to build for an older architecture then present on the build node by setting GMX_SIMD CMake flag.
```
boegel: `s/then/than`

easybuild/easyblocks/g/gromacs.py
```diff
@@ -123,6 +124,16 @@ def configure_step(self):
...
+            if build_option('optarch', default=None) and self.toolchain.comp_family() in build_option('optarch'):
```
boegel: You should also check here whether `optarch` is specified by compiler?

In theory, you could also have a hit here with `GCC` when `optarch` is just specified as a set of options (regardless of compiler), which would make this crash and burn below?

You could use something like:
```python
if isinstance(optarch, dict) and ...
```

easybuild/easyblocks/g/gromacs.py
```diff
@@ -123,6 +124,16 @@ def configure_step(self):
...
+                build_arch = build_option('optarch')[self.toolchain.comp_family()]
```
boegel: try to only grab these values once, introduce variable above:

```python
optarch = build_option('optarch', default=None)
```

easybuild/easyblocks/g/gromacs.py
```diff
@@ -357,3 +368,25 @@ def sanity_check_step(self):
...
+    def get_gromacs_arch(self, build_arch):
+        """Determine value of GMX_SIMD CMake flag based on optarch string.
+
+        Refs:
+        [0] http://manual.gromacs.org/documentation/2016.3/install-guide/index.html#typical-installation
+        [1] http://manual.gromacs.org/documentation/2016.3/install-guide/index.html#simd-support
+        [2] http://www.gromacs.org/Documentation/Acceleration_and_parallelization
+        """
+        build_arch = build_arch.upper()
+        if build_arch.endswith(("AVX2")) and not LooseVersion(self.version) < LooseVersion('5.0'):
```
boegel: Does it need to *end* with `avx2`, or is having `avx2` anywhere in the list of options sufficient? If the latter, use  something like

```python
if 'AVX2' in build_arch:
```

Also, simplify the version check using positive logic (easier to understand):

```python
LooseVersion(self.version) >= LooseVersion('5.0')
```

boegel:
@ostueker ping on the remarks?

ostueker:
@boegel: Sorry for the delay but I was very busy the past two months.

I think I have addressed all of your comments in commit c8a947e.

boegel:
@ostueker Sorry for not getting back to this earlier...

After taking a closer look at this, I reworked this a bit in https://github.com/ostueker/easybuild-easyblocks/pull/1, please review & merge to update this PR.

Mostly style fixes & moving things around, largest change is taking into account that `optarch` maybe also be a string value (for example when using `eb --optarch=mavx2 ...`

akesandgren:
This should probably only be done if explicitly requested in the easyconfig by a specific option.
The reason is that this is a bad thing to do for normal builds.

Basically, always build directly on the target arch!

And the targeted gromacs simd options is not complete.

mboisson:
@akesandgren, that is exactly what this PR does. It only gives GROMACS a specific architecture if the easybuild optarch is specified. Otherwise it keeps doing autodetection.

boegel:
I was wondering about this too, but figured that deciding whether or not to enable auto-detection of architecture based on the `optarch` configuration setting makes sense, since that's exactly intended to avoid building for the host system architecture (which is the default in EasyBuild), see http://easybuild.readthedocs.io/en/latest/Controlling_compiler_optimization_flags.html.

Maybe we should disable auto-detection when the `optarch` toolchain option is set to `False` though (just like we don't use `-xHost` or `-march=native` then)...
Not sure what the alternative is then though, can we tell GROMACS *not* to target a specific architecture but to build from some common denominator?

akesandgren:
Isn't optarch by default true? Which would make that code kick in, or am i missing something?

boegel:
@akesandgren The `optarch` toolchain option is `True` by default, yes.

What is being checked here is the value for `--optarch` or `$EASYBUILD_OPTARCH`, i.e. the value for the `optarch` EasyBuild configuration option, which can be something like `mavx2` or `Intel:xCORE-AVX2;GCC:mavx2`...

akesandgren:
Ah, slightly confusing....

ostueker:
@boegel No worries, I have been crazy busy all year as well.

> After taking a closer look at this, I reworked this a bit in ostueker#1, please review & merge to update this PR.

Looks good to me, however [Gromacs 2016](http://manual.gromacs.org/documentation/2016/install-guide/index.html#simd-support) also introduced a setting `AVX_512` for Intel Haswell and newer.  
I hope to find time early in the new year to add that as well. If till PR hasn't been merged by then I'll add it, otherwise I'll just file a new one.

boegel:
@akesandgren We usually only have one `return` per function, at the very end. Not jumping out in the middle of a function makes it easier to parse.

There's a tiny performance loss, but it won't be measurable, we'll just be checking an empty `optarch` value and nothing will happen.

akesandgren:
Ok, wasn't 100% sure on how the rest of the code acts on optarch being Null.
It wasn't for performance reason :-)

boegel:
@akesandgren This line makes sure that `optarch` is either a proper value or an empty string if it's not defined:

```python
optarch = build_option('optarch', default='')
```

easybuild/easyblocks/g/gromacs.py
```diff
@@ -70,6 +72,42 @@ def __init__(self, *args, **kwargs):
...
+
+        Refs:
+        [0] http://manual.gromacs.org/documentation/2016.3/install-guide/index.html#typical-installation
+        [1] http://manual.gromacs.org/documentation/2016.3/install-guide/index.html#simd-support
+        [2] http://www.gromacs.org/Documentation/Acceleration_and_parallelization
+        """
+        # default: fall back on autodetection
+        res = None
+
+        optarch = build_option('optarch', default='')
```
boegel: @ostueker This isn't correct, see https://github.com/ostueker/easybuild-easyblocks/pull/2

easybuild/easyblocks/g/gromacs.py
```diff
@@ -70,6 +72,42 @@ def __init__(self, *args, **kwargs):
...
+        # take into account that optarch value is a dictionary if it is specified by compiler family
+        if isinstance(optarch, dict):
+            comp_fam = self.toolchain.comp_family()
+            optarch = optarch.get(comp_fam, '').upper()
```
boegel: `.upper()` is only done in case `optarch` value is a `dict`, it should always be done

also fixed in https://github.com/ostueker/easybuild-easyblocks/pull/2

boegel:
Looks good to go now after another quick test, thanks for your efforts on this @ostueker, and @akesandgren & @mboisson for the feedback!

Issue #1164, opened 2017-04-13
The generic/cmakemake.py use of CMAKE_INCLUDE_PATH and CMAKE_LIBRARY_PATH doesn't work


akesandgren:
There is a problems with the current use of CMAKE_INCLUDE_PATH/CMAKE_LIBRARY_PATH.

It is setting CMAKE_INCLUDE_PATH and CMAKE_LIBRARY_PATH based only on the toolchains opinion of LDFLAGS/CPPFLAGS.
I think it would make much more sense to set it to what CPATH and LD_LIBRARY_PATH is instead.

After all, the modules that are (build)dependencies for a piece of code should be picked up by cmake when doing find_path/find_library.

wpoely86:
I can be wrong but aren't both build on the same info? LDFLAGS and CFLAGS use CPATH and LD_LIBRARY_PATH?

akesandgren:
No, self.toolchain.get_variable("CPPFLAGS", list) only contains the paths from the toolchain in use, not any CPATHs added by extra modules, like hwloc which is part of the module chain for foss, but not the actual toolchain.

The problem showed up when building gromacs using foss and actually looking at the build log. It was failing to find the hwloc that is part of the foss/2017a module chain.

boegel:
Let's follow up on this in #1165

PR #1165, opened 2017-04-13
Use CPATH/LD_LIBRARY_PATH for Cmakes find_path/find_library functions.

akesandgren:
Using only self.toolchain.get_variable("CPPFLAGS", list) when setting
CMAKE_INCLUDE_PATH results in cmake incapable of finding things coming
from (build)dependencies modules.

One example is GROMACS which uses find_path/library to find hwloc and
either fails completely or finds any hwloc installed from the OS. When
using foss it should pick up the path to hwloc from the corresponding
module.

See issue #1164

### DIFF
--- a/easybuild/easyblocks/generic/cmakemake.py
+++ b/easybuild/easyblocks/generic/cmakemake.py
@@ -62,8 +62,12 @@ class CMakeMake(ConfigureMake):
             self.log.nosupport("CMakeMake.configure_step: named argument 'builddir' (should be 'srcdir')", "2.0")
 
         # Set the search paths for CMake
-        include_paths = os.pathsep.join(self.toolchain.get_variable("CPPFLAGS", list))
-        library_paths = os.pathsep.join(self.toolchain.get_variable("LDFLAGS", list))
+        tc_ipaths = self.toolchain.get_variable("CPPFLAGS", list)
+        tc_lpaths = self.toolchain.get_variable("LDFLAGS", list)
+        cpaths = os.getenv('CPATH', '').split(os.pathsep)
+        lpaths = os.getenv('LD_LIBRARY_PATH', '').split(os.pathsep)
+        include_paths = os.pathsep.join(tc_ipaths + cpaths)
+        library_paths = os.pathsep.join(tc_lpaths + lpaths)
         setvar("CMAKE_INCLUDE_PATH", include_paths)
         setvar("CMAKE_LIBRARY_PATH", library_paths)
 
--- a/easybuild/easyblocks/generic/cmakemake.py
+++ b/easybuild/easyblocks/generic/cmakemake.py
@@ -40,6 +40,7 @@ from easybuild.tools.build_log import EasyBuildError
 from easybuild.tools.config import build_option
 from easybuild.tools.environment import setvar
 from easybuild.tools.run import run_cmd
+from vsc.utils.missing import nub
 
 
 class CMakeMake(ConfigureMake):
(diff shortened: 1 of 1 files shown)


boegel:
lgtm

I'll try and test this thoroughly to see if this raises any problems, but it's indeed the right thing to do.

I hope to include this in the upcoming EasyBuild v3.2.0...

boegel:
Testing showed that this change somehow breaks the installation of `ANTs-2.1.0rc3-goolf-1.5.14.eb` (and possibly others).

I'm not quite sure how though, since the output of the `cmake` command in the configure step looks identical.

The build fails with errors like this (with EasyBuild configured to use `/prefix`):

```
cp: cannot create regular file ‘/prefix/software/ANTs/2.1.0rc3-goolf-1.5.14/ITKv4/Modules/Remote/MGHIO/.git/objects/pack/pack-f3f9b3791c291e460f99b75a7eb3f4f4e1e02508.idx’: Permission denied
cp: cannot create regular file ‘/prefix/software/ANTs/2.1.0rc3-goolf-1.5.14/ITKv4/Modules/Remote/MGHIO/.git/objects/pack/pack-f3f9b3791c291e460f99b75a7eb3f4f4e1e02508.pack’: Permission denied
cp: cannot create regular file ‘/prefix/software/ANTs/2.1.0rc3-goolf-1.5.14/ITKv4/.git/objects/pack/pack-4364526d8e63450d626d092bd29932e7a2f390af.pack’: Permission denied
cp: cannot create regular file ‘/prefix/software/ANTs/2.1.0rc3-goolf-1.5.14/ITKv4/.git/objects/pack/pack-4364526d8e63450d626d092bd29932e7a2f390af.idx’: Permission denied
```

This problem does not occur without this patch to the `CMakeMake` easyblock...

akesandgren:
Why is something running around copying stuff beneath .git??

And that change can't cause permission changes in itself...

wpoely86:
@boegel please upload the cmake logs.

akesandgren:
Is this still a problem?

akesandgren:
ANTs should be scrapped.
The latest version is 2.1.0 which still doesn't understand GCC>4

akesandgren:
@boegel How about this one? is the ant stuff still giving you problems?

easybuild/easyblocks/generic/cmakemake.py
```diff
@@ -62,8 +62,12 @@ def configure_step(self, srcdir=None, builddir=None):
...
 
         # Set the search paths for CMake
-        include_paths = os.pathsep.join(self.toolchain.get_variable("CPPFLAGS", list))
-        library_paths = os.pathsep.join(self.toolchain.get_variable("LDFLAGS", list))
+        tc_ipaths = self.toolchain.get_variable("CPPFLAGS", list)
+        tc_lpaths = self.toolchain.get_variable("LDFLAGS", list)
+        cpaths = os.getenv('CPATH', '').split(os.pathsep)
+        lpaths = os.getenv('LD_LIBRARY_PATH', '').split(os.pathsep)
+        include_paths = os.pathsep.join(tc_ipaths + cpaths)
+        library_paths = os.pathsep.join(tc_lpaths + lpaths)
```
boegel: Should we avoid duplicates here? You can use `nub` (which you can import from `vsc.utils.missing`):

```python
include_paths = os.pathsep.join(nub(tc_ipaths + cpaths))
library_paths = os.pathsep.join(nub(tc_lpaths + lpaths))
```

akesandgren:
Any more comments on this?
It is still needed to correctly build GROMACS with CUDA.

Without it it fails to find NVML at least.

akesandgren:
@migueldiascosta Adding you too since you seem to be very active today :-)

migueldiascosta:
@akesandgren no good deed goes unpunished? :)

this looks sensible to me, will test it on some `CMakeMake` easyconfigs

akesandgren:
As you might guess I'd like to get this into 3.6.2 too.

We've been using it for quite some time now without and problems.

boegel:
Sorry for letting this sit for so long @akesandgren...

I think this is good to go, it shouldn't cause any problems, and if it does, it likely was a problem already that we just weren't aware of yet.

We'll see if anything serious pops up during the upcoming regression test for EasyBuild v3.6.2...

PR #1166, opened 2017-04-20
add support to install Python extensions without unpacking

boegel:
Thanks to https://github.com/hpcugent/easybuild-framework/pull/2194, `unpack_sources` can now also be specified for an individual extension, which is not taken into account currently. This patch fixes that.

### DIFF
--- a/easybuild/easyblocks/generic/pythonpackage.py
+++ b/easybuild/easyblocks/generic/pythonpackage.py
@@ -327,8 +327,12 @@ class PythonPackage(ExtensionEasyBlock):
             # specify current directory
             loc = '.'
         else:
-            # specify path to 1st source file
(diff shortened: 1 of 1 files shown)


boegel:
@wpoely86 please review?

boegel:

fizwit:
Can this feature be used to install Python Wheel modules within the exts_list when building Python? Can you give an example?

boegel:
@fizwit Yes it can, see https://github.com/hpcugent/easybuild-easyconfigs/blob/master/easybuild/easyconfigs/i/IPython/IPython-5.3.0-intel-2017a-Python-2.7.13.eb#L112. You need EasyBuild v3.2.0 or more recent for this.

PR #1167, opened 2017-04-26
Fixed installation directory for ITAC, so VT_ROOT gets correctly defined in itacvars.sh

damianam:
@boegel could you check if the versions compared in the if are correct?

### DIFF
--- a/easybuild/easyblocks/i/itac.py
+++ b/easybuild/easyblocks/i/itac.py
@@ -39,6 +39,7 @@ from distutils.version import LooseVersion
 from easybuild.easyblocks.generic.intelbase import IntelBase
 from easybuild.framework.easyconfig import CUSTOM
 from easybuild.tools.build_log import EasyBuildError
+from easybuild.tools.config import install_path
 from easybuild.tools.run import run_cmd
 
 class EB_itac(IntelBase):
@@ -62,11 +63,18 @@ class EB_itac(IntelBase):
         """
 
         if LooseVersion(self.version) >= LooseVersion('8.1'):
-            super(EB_itac, self).install_step(silent_cfg_names_map=None)
+            silent_cfg_names_map = {}
+            if LooseVersion(self.version) >= LooseVersion('9.0'):
+                # To generate the correct itacvars.sh (called by traceanalyzer) we need to change the default install dir
+                silent_cfg_names_map = {
+                    'install_dir': install_path(),
+                }
+
+            super(EB_itac, self).install_step(silent_cfg_names_map=silent_cfg_names_map)
 
             # itac v9.0.1 installer create itac/<version> subdir, so stuff needs to be moved afterwards
-            if LooseVersion(self.version) >= LooseVersion('9.0'):
-                super(EB_itac, self).move_after_install()
+            #if LooseVersion(self.version) >= LooseVersion('9.0'):
+            #    super(EB_itac, self).move_after_install()
         else:
             silent = \
 """


easybuild/easyblocks/i/itac.py
```diff
@@ -62,11 +63,18 @@ def install_step(self):
...
+                    'install_dir': install_path(),
+                }
+
+            super(EB_itac, self).install_step(silent_cfg_names_map=silent_cfg_names_map)
 
             # itac v9.0.1 installer create itac/<version> subdir, so stuff needs to be moved afterwards
-            if LooseVersion(self.version) >= LooseVersion('9.0'):
-                super(EB_itac, self).move_after_install()
+            #if LooseVersion(self.version) >= LooseVersion('9.0'):
+            #    super(EB_itac, self).move_after_install()
```
boegel: these lines should be removed instead, together with the comment above?

easybuild/easyblocks/i/itac.py
```diff
@@ -62,11 +63,18 @@ def install_step(self):
         """
 
         if LooseVersion(self.version) >= LooseVersion('8.1'):
-            super(EB_itac, self).install_step(silent_cfg_names_map=None)
+            silent_cfg_names_map = {}
+            if LooseVersion(self.version) >= LooseVersion('9.0'):
+                # To generate the correct itacvars.sh (called by traceanalyzer) we need to change the default install dir
+                silent_cfg_names_map = {
+                    'install_dir': install_path(),
```
boegel: I'm not sure that this is the right solution here...

The installation directory is by default named like the module is, so this is OK when your modules are named `<name>/<version>`.

If they're not, because you care installing with a different module naming scheme, `itac` will not get installed where it's supposed to.

With HMNS for example, you'd get `itac/<version>` directly in your installation prefix, while it should be in `Core/itac/<version>`...

So, the proper fix is to actually leave the `move_after_install` in place, and patch the `VT_ROOT=` line in `itacvars.sh`?

boegel:
@damianam version-check looks OK, the `VT_ROOT` definition is wrong for itac v9 and newer:

```
$ find software/itac -name itacvars.sh | xargs grep '^VT_ROOT=' | grep intel64
software/itac/8.1.4.045/intel64/bin/itacvars.sh:VT_ROOT="/foo/software/itac/8.1.4.045"
software/itac/2017.1.024/intel64/bin/itacvars.sh:VT_ROOT="/foo/software/itac/2017.1.024/itac/2017.1.024"
software/itac/9.0.3.051/intel64/bin/itacvars.sh:VT_ROOT="/foo/software/itac/9.0.3.051/itac/9.0.3.051"
```

easybuild/easyblocks/i/itac.py
damianam: HMNS with `EASYBUILD_FIXED_INSTALLDIR_NAMING_SCHEME` would still work, but I get your point. Patching `itacvars.sh` might be safer, but I have to admit that I dislike to do something (install in `installdir/itac/version`) to undo it later (`move_after_install` and patch `itacvars.sh`). Doesn't look right.

boegel: Well, we can also simply not do `move_after_install`, and accept the silly double subdirectory .

I agree it isn't right, but it's not EB's fault imho, it's the `itac` installer.

damianam: > I agree it isn't right, but it's not EB's fault imho, it's the itac installer.
 
Agreed

boegel: So, shall we revert to simply not doing `move_after_install`?

damianam: Sounds reasonable.

boegel:
@damianam Is this solving all the problems reported in #1231?

boegel:
@damianam Ping?

damianam:
@boegel I kind of lost track of what was the issue here, but I proceeded as suggested in our discussion.

boegel:
@damianam In its current state, this can not be merged, since now the sanity check is broken for `itac-2017.1.024.eb` for example...

Maybe a better approach is to do the move but patch `itacvars.sh` so the right path is used?

akesandgren:
Why care about itacvars.(c)sh in the first place?
Do chmod 0 on them and use the settings from the module file instead, that one seem to be correct?
@damianam @boegel  ??

damianam:
@akesandgren good question. This PR has been long forgotten, but I remember that the reason to fix it was because someone was sourcing `itacvars.sh` for some reason. I don't see the point of it though..... But if we do `chmod 0` on these scripts, it makes sense to do the same for all the other intel packages, for consistency. Which at the end might create other issues (like not being able to verify that the module is doing the correct thing, in comparison with the script). I guess it makes sense to just ignore/close this PR and https://github.com/easybuilders/easybuild-easyblocks/issues/1231 and simply tell users to rely on the module?

akesandgren:
Yeah, educating users is one way.
We used to do chmod 0 on these in our old pre-EB setup.

PR #1168, opened 2017-05-03
enhance TBB easyblock to also support building open source versions

pescobar:
this still requires some work before it can be merged.


boegel:
checked with https://github.com/hpcugent/easybuild-easyconfigs/pull/4538 and https://github.com/hpcugent/easybuild-easyconfigs/pull/4541 + existing `tbb` eaysconfigs, works like a charm

PR #1173, opened 2017-05-08
Add openmp support to namd easyblock.

akesandgren:
Also add TCL support.

### DIFF
--- a/easybuild/easyblocks/n/namd.py
+++ b/easybuild/easyblocks/n/namd.py
@@ -24,7 +24,7 @@ from easybuild.easyblocks.generic.makecp import MakeCp
 from easybuild.framework.easyconfig import CUSTOM, MANDATORY
 from easybuild.tools.build_log import EasyBuildError
 from easybuild.tools.config import build_option
-from easybuild.tools.filetools import extract_file
+from easybuild.tools.filetools import apply_regex_substitutions, change_dir, extract_file
 from easybuild.tools.modules import get_software_root, get_software_version
 from easybuild.tools.run import run_cmd
 
@@ -54,6 +54,17 @@ class EB_NAMD(MakeCp):
         super(EB_NAMD, self).__init__(*args, **kwargs)
         self.namd_arch = None
 
+    def extract_step(self):
+        """Custom extract step for NAMD, we need to extract charm++ so we can patch it."""
+        super(EB_NAMD, self).extract_step()
+
+        change_dir(self.src[0]['finalpath'])
+        self.charm_tarballs = glob.glob('charm-*.tar')
+        if len(self.charm_tarballs) != 1:
+            raise EasyBuildError("Expected to find exactly one tarball for Charm++, found: %s", self.charm_tarballs)
+
+        extract_file(self.charm_tarballs[0], os.getcwd())
+
     def configure_step(self):
         """Custom configure step for NAMD, we build charm++ first (if required)."""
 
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/n/namd.py
```diff
@@ -74,33 +85,42 @@ def configure_step(self):
...
-        charm_subdir = '.'.join(os.path.basename(charm_tarballs[0]).split('.')[:-1])
+        cmd = "./build charm++ %s %s --with-numa -j%s %s -DMPICH_IGNORE_CXX_SEEK" % tup
+        charm_subdir = '.'.join(os.path.basename(self.charm_tarballs[0]).split('.')[:-1])
         self.log.debug("Building Charm++ using cmd '%s' in '%s'" % (cmd, charm_subdir))
         run_cmd(cmd, path=charm_subdir)
 
         # compiler (options)
         self.cfg.update('namd_cfg_opts', '--cc "%s" --cc-opts "%s"' % (os.environ['CC'], os.environ['CFLAGS']))
-        self.cfg.update('namd_cfg_opts', '--cxx "%s" --cxx-opts "%s"' % (os.environ['CXX'], os.environ['CXXFLAGS']))
+        self.cfg.update('namd_cfg_opts', '--cxx "%s --std=c++11" --cxx-opts "%s"' % (os.environ['CXX'], os.environ['CXXFLAGS']))
```
boegel: @akesandgren should `--std=c++11` be used regardless of NAMD version?

easybuild/easyblocks/n/namd.py
```diff
@@ -74,33 +85,42 @@ def configure_step(self):
...
-        charm_tarballs = glob.glob('charm-*.tar')
-        if len(charm_tarballs) != 1:
-            raise EasyBuildError("Expected to find exactly one tarball for Charm++, found: %s", charm_tarballs)
-
-        extract_file(charm_tarballs[0], os.getcwd())
 
         tup = (self.cfg['charm_arch'], self.cfg['charm_opts'], self.cfg['parallel'], os.environ['CXXFLAGS'])
-        cmd = "./build charm++ %s %s -j%s %s -DMPICH_IGNORE_CXX_SEEK" % tup
-        charm_subdir = '.'.join(os.path.basename(charm_tarballs[0]).split('.')[:-1])
+        cmd = "./build charm++ %s %s --with-numa -j%s %s -DMPICH_IGNORE_CXX_SEEK" % tup
```
boegel: `--with-numa` also works for old versions?

easybuild/easyblocks/n/namd.py
```diff
@@ -112,7 +132,7 @@ def configure_step(self):
                 self.cfg.update('namd_cfg_opts', "--with-fftw")
             self.cfg.update('namd_cfg_opts', "--fftw-prefix %s" % fftw)
 
-        namd_charm_arch = "--charm-arch %s" % '-'.join(self.cfg['charm_arch'].strip().split(' '))
+        namd_charm_arch = "--charm-arch %s" % '-'.join(self.cfg['charm_arch'].strip().split())
```
boegel: won't this break existing easyconfigs that use spaces in `charm_arch`?

easybuild/easyblocks/n/namd.py
```diff
@@ -131,8 +151,12 @@ def test_step(self):
             namdcmd = os.path.join(self.cfg['start_dir'], self.namd_arch, 'namd%s' % self.version.split('.')[0])
             if self.cfg['charm_arch'].startswith('mpi'):
                 namdcmd = self.toolchain.mpi_cmd_for(namdcmd, 2)
-            cmd = "%(namd)s %(testdir)s" % {
+            ppn = ''
+            if self.toolchain.options.get('openmp', False):
+                ppn = '+ppn 2'
```
boegel: Why 2, I'd expect `1` here (or nothing at all, as it was before)?

easybuild/easyblocks/n/namd.py
akesandgren: Well, didn't really check namd 2.9 or 2.10b1 (which are the only non 2.12 versions currently listed in EB) I'd vote for deprecating them anyway.

easybuild/easyblocks/n/namd.py
akesandgren: Yes numa works for 2.9 at least

easybuild/easyblocks/n/namd.py
akesandgren: There is no space in --charm-arch's argument. it's always a "a-b-c-d" string

easybuild/easyblocks/n/namd.py
akesandgren: This is the number of threads to use, and since we're testing openmp, in this case, it's better to turn it on and set to 2 than let it fly off on it's own and try to use #cores as threads.

easybuild/easyblocks/n/namd.py
boegel: but `charm_arch` may contain a space (see https://github.com/hpcugent/easybuild-easyconfigs/search?utf8=%E2%9C%93&q=charm_arch&type=), you're changing the `split`

easybuild/easyblocks/n/namd.py
boegel: Maybe they should be archived, yes, but only with EasyBuild 4.0.

Until then, we should maintain backward compatibility.

easybuild/easyblocks/n/namd.py
boegel: OK, sorry, this is OK. You're still splitting on spaces in `charm_arch` here, but also handling consecutive spaces correctly.

easybuild/easyblocks/n/namd.py
boegel: tested, using --std=c++11 breaks older NAMD easyconfig, so let's inject this conditionally

done in https://github.com/akesandgren/easybuild-easyblocks/pull/5

boegel:
Tested with existing NAMD easyconfigs + those in https://github.com/hpcugent/easybuild-easyconfigs/pull/4516 and https://github.com/hpcugent/easybuild-easyconfigs/pull/3199, good to go.

Thanks @akesandgren!

PR #1169, opened 2017-05-03
update NCL easyblock to support building in parallel

boegel:

PR #1179, opened 2017-05-15
revert building in parallel for NCL, enhance sanity check

boegel:
building in parallel was enabled in #1169, but it turns out this break the installation in subtle ways, i.e. fonts are no longer being found when generating plots with NCL (using `ng4ex gsun01n -clean`) because the installation is incomplete (`fontcaps` failed to build)

cc @akesandgren


akesandgren:
As a revert this looks good.

boegel:
Thanks for the review @akesandgren!

PR #1180, opened 2017-05-15
enable building of shared FFTW libraries

boegel:
fix for https://github.com/hpcugent/easybuild-easyconfigs/issues/4466

I need to double-check whether this causes problems with any of the existing FFTW easyconfigs...


boegel:
works like a charm with all existing FFTW modules

@wpoely86 please review?

akesandgren:
lrgtm

boegel:
Thanks for the review @akesandgren!

PR #1184, opened 2017-05-21
Tkinter easyblock

wpoely86:
Follow up from #1118


easybuild/easyblocks/t/tkinter.py
```diff
@@ -0,0 +1,120 @@
...
+##
+"""
+EasyBuild support for building and installing Tkinter. This is the Python core
+module to use Tcl/Tk.
+
+@author: Stijn De Weirdt (Ghent University)
+@author: Dries Verdegem (Ghent University)
+@author: Kenneth Hoste (Ghent University)
+@author: Pieter De Baets (Ghent University)
+@author: Jens Timmerman (Ghent University)
```
boegel: drop these above, only you and Adam are actual authors of this?

easybuild/easyblocks/t/tkinter.py
```diff
@@ -0,0 +1,120 @@
...
+        tkinter_so = os.path.join(self.installdir, 'lib', pyver, '_tkinter*.' + shlib_ext)
+        tkinter_so_hits = glob.glob(tkinter_so)
+        if len(tkinter_so_hits) == 1:
+            self.log.info("Found exactly one _tkinter*.so: %s", tkinter_so_hits[0])
+        else:
+            raise EasyBuildError("Expected to find exactly one _tkinter*.so: %s", tkinter_so_hits)
+
+        custom_paths = {
+            'files': [],
+            'dirs': ['lib']
```
boegel: Make this more specific, just include `'files': [tkinter_so]`?

Should we also check whether nothing else is in the install dir?

easybuild/easyblocks/t/tkinter.py
```diff
@@ -0,0 +1,120 @@
...
+        }
+
+        super(EB_Python, self).sanity_check_step(custom_commands=custom_commands, custom_paths=custom_paths)
+
+    def make_module_extra(self):
+        """Set PYTHONPATH"""
+        txt = super(EB_Tkinter, self).make_module_extra()
+        pyver = '.'.join(self.version.split('.')[:2])
+        txt += self.module_generator.prepend_paths('PYTHONPATH', os.path.join("lib", "python%s" % pyver))
```
boegel: use `self.all_pylibdirs`?

easybuild/easyblocks/t/tkinter.py
```diff
@@ -0,0 +1,120 @@
...
+            tkinter = 'tkinter'
+        else:
+            tkinter = 'Tkinter'
+        custom_commands = ["python -c 'import %s'" % tkinter]
+
+        pyver = 'python' + '.'.join(self.version.split('.')[:2])
+        shlib_ext = get_shared_lib_ext()
+
+        # check whether _tkinter*.so is found, exact filename doesn't matter
+        tkinter_so = os.path.join(self.installdir, 'lib', pyver, '_tkinter*.' + shlib_ext)
```
boegel: use `all_pylibdirs` rather than hardcoding `lib`?

easybuild/easyblocks/t/tkinter.py
```diff
@@ -0,0 +1,120 @@
...
+    def install_step(self):
+        """Install python but only keep the bits we need"""
+        super(EB_Tkinter, self).install_step()
+
+        tmpname = "eb-tmp"
+        tmpdir = os.path.join(self.installdir, tmpname)
+        mkdir(tmpdir)
+
+        pyver = '.'.join(self.version.split('.')[:2])
+        libdir = os.path.join(self.installdir, "lib", "python%s" % pyver)
```
boegel: same here, use `all_pylibdirs`?

easybuild/easyblocks/t/tkinter.py
```diff
@@ -0,0 +1,120 @@
...
+            raise EasyBuildError("Tk is mandatory to build Tkinter")
+
+        super(EB_Tkinter, self).configure_step()
+
+    def install_step(self):
+        """Install python but only keep the bits we need"""
+        super(EB_Tkinter, self).install_step()
+
+        tmpname = "eb-tmp"
+        tmpdir = os.path.join(self.installdir, tmpname)
```
boegel: I'd make this a subdir of `self.builddir`, which is cleaned up automagically?

easybuild/easyblocks/t/tkinter.py
```diff
@@ -0,0 +1,120 @@
...
+        mkdir(tmpdir)
+
+        pyver = '.'.join(self.version.split('.')[:2])
+        libdir = os.path.join(self.installdir, "lib", "python%s" % pyver)
+        tkdirs = ["lib-tk", "lib-dynload/_tkinter.so"]
+
+        copy([os.path.join(libdir, x) for x in tkdirs], tmpdir)
+
+        delete_dirs = [x for x in os.listdir(self.installdir) if not x == tmpname]
```
boegel: `x != tmpname`

boegel: and if you use `self.builddir` in `tmpdir`, you don't even need to do this, you can just wipe `self.installdir` entirely?

easybuild/easyblocks/t/tkinter.py
```diff
@@ -0,0 +1,120 @@
...
+
+        for deldir in delete_dirs:
+            rmtree2(os.path.join(self.installdir, deldir))
+
+        mkdir(libdir, parents=True)
+        shutil.move(os.path.join(tmpdir, tkdirs[0]), libdir)
+        shutil.move(os.path.join(tmpdir, os.path.basename(tkdirs[1])), libdir)
+
+        rmtree2(tmpdir)
```
boegel: no need for this if `tmpdir` uses `self.builddir`

easybuild/easyblocks/t/tkinter.py
boegel: also, use `tempfile.mkdtemp`, you can tell it in which location to create a directory, and specify a prefix for the name (`eb-tmp-` for example)

easybuild/easyblocks/t/tkinter.py
```diff
@@ -0,0 +1,120 @@
...
+        """Install python but only keep the bits we need"""
+        super(EB_Tkinter, self).install_step()
+
+        tmpname = "eb-tmp"
+        tmpdir = os.path.join(self.installdir, tmpname)
+        mkdir(tmpdir)
+
+        pyver = '.'.join(self.version.split('.')[:2])
+        libdir = os.path.join(self.installdir, "lib", "python%s" % pyver)
+        tkdirs = ["lib-tk", "lib-dynload/_tkinter.so"]
```
boegel: not all of these are dirs, but they are all files (somewhat) ;)

easybuild/easyblocks/t/tkinter.py
```diff
@@ -0,0 +1,120 @@
...
+
+
+class EB_Tkinter(EB_Python):
+    """Support for building/installing the Python Tkinter module
+    based on the normal Python module. We build a normal python
+    but only install the Tkinter bits.
+    """
+
+    def configure_step(self):
+        """Check for Tk"""
```
boegel: `Check for Tk before configuring`

easybuild/easyblocks/t/tkinter.py
wpoely86: that possible on a different filesystem. The reason why I do it like this is to have a real 'move' on the same fs.

boegel: Is that really going to matter for the amount of files involved?

Are you worried about speed, or something else?

wpoely86: Not sure, some people do weird things out there. But I don't feel strongly about this. I can change if that makes you happy.

boegel: Well, it's a lot cleaner to use the build dir for this, since it's cleaned up automagically...

easybuild/easyblocks/t/tkinter.py
wpoely86: Well, we delete everything in there and copy only the tk parts back?

`tkinter_so` is a glob.

easybuild/easyblocks/t/tkinter.py
```diff
@@ -0,0 +1,100 @@
...
+@author: Adam Huffman (The Francis Crick Institute)
+@author: Ward Poelmans (Free University of Brussels)
+"""
+import os
+import shutil
+import tempfile
+from distutils.version import LooseVersion
+
+from easybuild.easyblocks.generic.pythonpackage import det_pylibdir
+from easybuild.easyblocks.p.python import EB_Python
```
boegel: don't use the `.p.`, import directly from `easybuild.easyblocks.python`

easybuild/easyblocks/t/tkinter.py
```diff
@@ -0,0 +1,100 @@
...
+        pylibdir = os.path.join(self.installdir, os.path.dirname(det_pylibdir()))
+        tkparts = ["lib-tk", "lib-dynload/_tkinter.so"]
+
+        copy([os.path.join(pylibdir, x) for x in tkparts], tmpdir)
+
+        rmtree2(self.installdir)
+
+        mkdir(pylibdir, parents=True)
+        shutil.move(os.path.join(tmpdir, tkparts[0]), pylibdir)
+        shutil.move(os.path.join(tmpdir, os.path.basename(tkparts[1])), pylibdir)
```
boegel: wrap this in `try`/`except OSError as err:`?

easybuild/easyblocks/t/tkinter.py
```diff
@@ -0,0 +1,100 @@
...
+        custom_commands = ["python -c 'import %s'" % tkinter]
+
+        shlib_ext = get_shared_lib_ext()
+        pylibdir = os.path.dirname(det_pylibdir())
+
+        custom_paths = {
+            'files': ['%s/_tkinter.%s' % (pylibdir, shlib_ext)],
+            'dirs': ['lib']
+        }
+
```
boegel: drop this empty line?

easybuild/easyblocks/t/tkinter.py
```diff
@@ -0,0 +1,102 @@
...
+        copy([os.path.join(pylibdir, x) for x in tkparts], tmpdir)
+
+        rmtree2(self.installdir)
+
+        mkdir(pylibdir, parents=True)
+        try:
+            shutil.move(os.path.join(tmpdir, tkparts[0]), pylibdir)
+            shutil.move(os.path.join(tmpdir, os.path.basename(tkparts[1])), pylibdir)
+        except:
```
boegel: include the exception in the error message, so it's clear what exactly went wrong...

```python
except OSError as err:
    raise EasyBuildError("Failed to move Tkinter back to the install directory: %s", err)
```

boegel:
lgtm, and tested with https://github.com/hpcugent/easybuild-easyconfigs/pull/4620, thanks @wpoely86 and @verdurin for levelling the ground for this!

PR #1185, opened 2017-05-22
stop using deprecated 'copytree' function from easybuild.tools.filetools

boegel:
`copytree` function was deprecated in https://github.com/hpcugent/easybuild-framework/pull/2177

`copy_dir` is not a drop-in replacement yet for (`shutil.`)`copytree`, at least not without https://github.com/hpcugent/easybuild-framework/pull/2230

Once https://github.com/hpcugent/easybuild-framework/pull/2230 is merged, I plan to get rid of all `shutil.copytree` uses in the existing easyblocks.

### DIFF
--- a/easybuild/easyblocks/d/dl_poly_classic.py
+++ b/easybuild/easyblocks/d/dl_poly_classic.py
@@ -33,7 +33,7 @@ import os
 import shutil
 
 from easybuild.tools.build_log import EasyBuildError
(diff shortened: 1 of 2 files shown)


boegel:
I'll merge this in to stop using deprecated behaviour, hoping to find time to clean up the use of `shutil.copytree` and replace it by `copy_dir` in all easyblocks...

PR #1186, opened 2017-05-24
individually apply regex substitutions that add lines

migueldiascosta:
some of these (the ones with repeated matching patterns) were not being applied because of the way apply_regex_substitutions works

### DIFF
--- a/easybuild/easyblocks/s/siesta.py
+++ b/easybuild/easyblocks/s/siesta.py
@@ -86,6 +86,7 @@ class EB_Siesta(ConfigureMake):
         else:
             fftw = None
 
+        regex_newlines = []
         regex_subs = [
             ('dc_lapack.a', ''),
             (r'^NETCDF_INTERFACE\s*=.*$', ''),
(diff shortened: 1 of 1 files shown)


migueldiascosta:
@akesandgren please review? MPI is not getting enabled in `4.1-b2` because of this....

akesandgren:
Doh, didn't notice..

poatch looks good to me, but i haven't tested it.

boegel:
@migueldiascosta Can we enhance the sanity check somehow to catch that MPI wasn't built? Checking for a particular file, or maybe running some command?

migueldiascosta:
@boegel hm... `echo "SystemName test" | mpirun -np 2 siesta 2>/dev/null | grep PARALLEL` should return `PARALLEL version`...

akesandgren:
A ldd check of siesta should reveal a mpi library and avoid using mpirun.

ldd %(instalpath)s/bin/siesta | grep libmpi

should be non-null

migueldiascosta:
@akesandgren that was actually not enough to detect, because it was linked with mpi but -DMPI was not defined

akesandgren:
Ohhh, evil :-)

migueldiascosta:
yep :)

boegel:
@migueldiascosta So, let's include that as a sanity check command then?

```python
custom_commands = ["echo 'SystemName test' | mpirun -np 2 siesta 2>/dev/null | grep PARALLEL"]

super(EB_Siesta, self).sanity_check_step(custom_paths=custom_paths, custom_commands=custom_commands)
```

If `grep` doesn't find a matching line, it will exit with non-zero, and the sanity check will fail.

I tested this on top of your PR, works like a charm both to detect the broken installation and verify the fixed one, see https://github.com/migueldiascosta/easybuild-easyblocks/pull/1

boegel:
Thanks for the fix @migueldiascosta!

PR #1187, opened 2017-05-24
Ferret 7.1 update

hajgato:


hajgato:
This `EasyBlock` is replaced by #1349

PR #1189, opened 2017-05-29
Make samtools able to handle version 0.1.17

hajgato:


boegel:
lgtm, thanks @hajgato!

PR #1190, opened 2017-06-02
Support for Simpack-2017

onkelmonokel:


easybuild/easyblocks/s/simpack.py
```diff
@@ -0,0 +1,72 @@
...
+from easybuild.tools.run import run_cmd
+
+
+class EB_Simpack(PackedBinary):
+    """Support for installing Simpack."""
+
+    def __init__(self, *args, **kwargs):
+        super(EB_Simpack, self).__init__(*args, **kwargs)
+	self.license = self.cfg['license_server'] + ':' + self.cfg['license_server_port']
+	self.defaultconf = os.path.join(self.installdir, 'defaults/settings/default.ini')
```
boegel: @onkelmonokel Please avoid using tab characters for indentation, each tab should be replaced with 4 spaces.

easybuild/easyblocks/s/simpack.py
```diff
@@ -0,0 +1,72 @@
...
+from easybuild.framework.easyconfig import CUSTOM
+from easybuild.tools.build_log import EasyBuildError
+from easybuild.tools.filetools import read_file, write_file
+from easybuild.tools.run import run_cmd
+
+
+class EB_Simpack(PackedBinary):
+    """Support for installing Simpack."""
+
+    def __init__(self, *args, **kwargs):
```
boegel: @onkelmonokel please add a docstring for each method, see existing easyblocks

Here something like `Constructor for Simpack easyblock` would suffice.

easybuild/easyblocks/s/simpack.py
```diff
@@ -0,0 +1,72 @@
...
+        super(EB_Simpack, self).__init__(*args, **kwargs)
+	self.license = self.cfg['license_server'] + ':' + self.cfg['license_server_port']
+	self.defaultconf = os.path.join(self.installdir, 'defaults/settings/default.ini')
+
+    def install_step(self):
+        run_cmd("./spck-2017-build53-linux64-installer.bin --mode unattended --prefix %s" % self.installdir, log_all=True, simple=True)
```
boegel: lines should be shorter than 120 characters

so, I'd suggest something like:

```python
cmd = "..."
run_cmd(cmd, log_all=True, ...)

File: easybuild/easyblocks/s/simpack.py [hunk 4]
```diff
@@ -0,0 +1,72 @@
...
+
+    def post_install_step(self):
+	if self.license != "license.example.com:000000":
```
boegel: I wouldn't skip the configuration, but maybe print a warning instead that the license server should be customised in the easyconfig file?

easybuild/easyblocks/s/simpack.py
```diff
@@ -0,0 +1,72 @@
...
+	    c = open(self.defaultconf, "a")
+	    print(self.defaultconf)
+	    c.write('[dsls]' + '\n')
+	    c.write('licServer=' + self.license + '\n')
+	    c.close()
```
boegel: please use the `write_file` function available in `easybuild.tools.filetools`

easybuild/easyblocks/s/simpack.py
```diff
@@ -0,0 +1,72 @@
...
+
+    def make_module_extra(self):
+        txt = super(EB_Simpack, self).make_module_extra()
+        txt += self.module_generator.prepend_paths("PATH", ['run/bin/linux64'])
+        return txt
+
+    def sanity_check_step(self):
+        custom_paths = {
+            'files': [],
+            'dirs': ['run/bin','defaults'],
```
boegel: maybe make this a bit more specific, i.e. check for the actual main binary?

easybuild/easyblocks/s/simpack.py
boegel: since these are only used in one place, I wouldn't introduce class variable, but just use local variables in `post_install_step`?

easybuild/easyblocks/s/simpack.py
```diff
@@ -0,0 +1,79 @@
...
+from easybuild.tools.filetools import write_file
+from easybuild.tools.run import run_cmd
+
+
+class EB_Simpack(PackedBinary):
+    """Support for installing Simpack."""
+
+    def __init__(self, *args, **kwargs):
+        """Constructor for Simpack easyblock."""
+        super(EB_Simpack, self).__init__(*args, **kwargs)
```
boegel: Since this doesn't initialise any class variables anymore, you can just drop the custom definition for `__init__` entirely, it doesn't do anything...

easybuild/easyblocks/s/simpack.py
```diff
@@ -0,0 +1,79 @@
...
+
+    def install_step(self):
+        """Configure Simpack installation."""
+        # Installer changes name:
+        installer = glob.glob('spck-*-linux64-installer.bin')
+        if len(installer) == 0:
+            raise EasyBuildError("Didn't find installer")
+        elif len(installer) > 1:
+            raise EasyBuildError("Found to many installers: %s", installer)
```
boegel:
```python
installers = glob.glob('spck-*-linux64-installer.bin')
if len(installers) == 1:
    cmd = "./%s --mode unattended --prefix %s" % (installers[0], self.installdir)
else:
    raise EasyBuildError("Failed to isolate installer: %s", installers)

File: easybuild/easyblocks/s/simpack.py [hunk 9]
```diff
@@ -0,0 +1,79 @@
...
+        cmd = "./%s --mode unattended --prefix %s"
+        run_cmd(cmd % (installer[0], self.installdir), log_all=True, simple=True)
+
+    def post_install_step(self):
+        """Post install (license configuration) for Simpack."""
+        license = os.getenv('EB_SIMPACK_LICENSE_SERVER')
+        if license is not None:
```
boegel: Please use `find_flexlm_license` function, see for example #1317 (in comparison with #1296)?

easybuild/easyblocks/s/simpack.py
```diff
@@ -0,0 +1,79 @@
...
+            defaultconf = os.path.join(self.installdir, 'defaults/settings/default.ini')
+            txt = '[dsls]\nlicserver=%s\n' % license
+            write_file(defaultconf, txt, append=True)
+
+    def make_module_extra(self):
+        """Extra module entries for Simpack."""
+        txt = super(EB_Simpack, self).make_module_extra()
+        txt += self.module_generator.prepend_paths("PATH", ['run/bin/linux64'])
+        return txt
```
boegel: Do this by customising `make_module_guesses`:

```python
def make_module_guesses(self):
    """Specify custom location for Simpack binaries."""
    guesses = super(EB_Simpack, self).make_module_guesses()
    guesses.update({'PATH': ['run/bin/linux64']})
    return guesses
```

PR #1192, opened 2017-06-14
Fixed bug in alias definition in IMPI

damianam:
The alias definition in `impi.py` is wrong. It uses `-cc` for the C++ and Fortran wrappers, which need `-cxx` and `[-f90|-fc]` instead. Example:

```
~]$ $EBROOTIMPI/bin64/mpiicpc -cxx=not_existing_compiler -show
not_existing_compiler [...]

~]$ $EBROOTIMPI/bin64/mpiicpc -cc=not_existing_compiler -show
icpc '-cc=not_existing_compiler' [...]
```


boegel:
Thanks for the fix @damianam!

PR #1195, opened 2017-06-21
consider $EB_*_LICENSE_SERVER(_PORT) in MATLAB and ANSYS easyblocks

boegel:
Currently the easyconfigs for ANSYS and MATLAB pick up on the `$EB_*_LICENSE_SERVER(_PORT)` environment variables, see https://github.com/hpcugent/easybuild-easyconfigs/pull/4762.

That's useful as a fallback mechanism so you can set these environment variables without touching the easyconfig files (mainly useful during testing), but since it requires an `import` statement in the easyconfig files, this should be handled by the easyblocks instead.

### DIFF
--- a/easybuild/easyblocks/a/ansys.py
+++ b/easybuild/easyblocks/a/ansys.py
@@ -47,7 +47,11 @@ class EB_ANSYS(PackedBinary):
     def install_step(self):
         """Custom install procedure for ANSYS."""
         licserv = self.cfg['license_server']
+        if licserv is None:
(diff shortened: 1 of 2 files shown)


boegel:
@vanzod up for reviewing this?

vanzod:
lgtm

boegel:
Tested with easyconfigs in https://github.com/hpcugent/easybuild-easyconfigs/pull/4762, thanks for the review @vanzod!

PR #1196, opened 2017-06-21
add omp_num_threads custom parameter in CP2K easyblock to allow defining $OMP_NUM_THREADS during testing

boegel:
required to avoid use of `import` in CP2K easyconfig, see https://github.com/hpcugent/easybuild-easyconfigs/pull/4763


boegel:
tested with `eb -x`, works as intended

@vanzod please review?

vanzod:
lgtm

boegel:
Thanks for the review @vanzod!

PR #1197, opened 2017-06-23
correct version check in MATLAB easyblock to check for 2016b rather than 2017a

boegel:
@hajgato I helped out someone fixing problems with the installation of MATLAB 2016b, it seems like just updating the version check was sufficient to resolve the issue.


ocaisa:
Changes are trivial so if it works it looks good to me

boegel:

PR #1204, opened 2017-06-30
EasyBlock supporting OpenKIM library.

schiotz:
The custom easyblock is necessary because the source files need
to be unpacked into a specific folder structure.

### DIFF
--- a/easybuild/easyblocks/o/openkim.py
+++ b/easybuild/easyblocks/o/openkim.py
@@ -1,0 +1,69 @@
+##
+# Copyright 2009-2017 Ghent University
+#
+# This file is part of EasyBuild,
+# originally created by the HPC team of Ghent University (http://ugent.be/hpc/en),
+# with support of Ghent University (http://ugent.be/hpc),
+# the Flemish Supercomputer Centre (VSC) (https://www.vscentrum.be),
+# Flemish Research Foundation (FWO) (http://www.fwo.be/en)
+# and the Department of Economy, Science and Innovation (EWI) (http://www.ewi-vlaanderen.be/en).
+#
+# http://github.com/hpcugent/easybuild
+#
+# EasyBuild is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation v2.
+#
+# EasyBuild is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with EasyBuild.  If not, see <http://www.gnu.org/licenses/>.
+##
+"""
+EasyBuild support for the Open Knowledgbase of Interatomic Models
+
+See OpenKIM.org
+
+@author: Jakob Schiotz (Tech. Univ. Denmark)
+"""
+
+from easybuild.easyblocks.generic.configuremake import ConfigureMake
+from easybuild.tools.filetools import extract_file
+from easybuild.tools.build_log import EasyBuildError
+import os
+
+class EB_OpenKIM(ConfigureMake):
+    def extract_step(self):
+        """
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/o/openkim.py
```diff
@@ -0,0 +1,69 @@
...
+            self.log.info("Unpacking source %s to %s" % (src['name'],
+                                                         targetdir))
+            srcdir = extract_file(src['path'], targetdir, cmd=src['cmd'],
+                                  extra_options=self.cfg['unpack_options'])
+            if srcdir:
+                self.src[self.src.index(src)]['finalpath'] = srcdir
+                if main_unpack_dir is None:
+                    main_unpack_dir = srcdir
+                    self.log.info("Detected main unpacking path: %s" 
+                                  % (main_unpack_dir,))
```
boegel: @schiotz `main_unpack_dir` is getting set using the 1st listed source, is that right?

If so, please make that explicit (and avoid using `assert`), something like this (note: untested):

```python
self.log.info("Unpacking main source %s to %s", self.src[0]['name'], self.builddir)
main_unpack_dir = extract_file(src.src[0]['path'], self.builddir, cmd=self.src[0]['cmd'], extra_options(self.cfg['unpack_options'])
if not main_unpack_dir:
    raise EasyBuildError("Failed to unpack %s", self.src[0]['name'])

for idx, src in enumerate(self.src[1:]):
    if src['name'].startswith('MD_'):
        targetdir = os.path.join(main_unpack_dir, 'src', 'model_drivers')
    elif src['name'].startswith('MO_'):
        targetdir = os.path.join(main_unpack_dir, 'src', 'models')
    else:
        raise EasyBuildError("Don't know which target directory to use when unpacking %s", src['name'])

    self.log.info("Unpacking source %s to %s", src['name'], targetdir)

    srcdir = extract_file(src['path'], targetdir, cmd=src['cmd'], extra_options=self.cfg['unpack_options'])
    if srcdir:
        self.src[idx+1]['finalpath'] = srcdir
    else:
        raise EasyBuildError("Unpacking source %s failed", src['name'])

**Comment by schiotz** [comment]
@boegel 
Thanks for your comments.  This is indeed more readable.

File: easybuild/easyblocks/o/openkim.py [hunk 2]
```diff
@@ -0,0 +1,71 @@
...
+class EB_OpenKIM(ConfigureMake):
+    def extract_step(self):
+        """
+        Unpack the source files.
+
+        The main file is unpacked as usual, all model files are
+        unpacked into the right subdirectories.
+        """
+        driverdir = ('src', 'model_drivers')
+        modeldir = ('src', 'models')
```
boegel: since both `driverdir` and `modeldir` are only used in one place, I don't see the reason to define variables for these, you can just use these values directly when defining `targetdir`...

easybuild/easyblocks/o/openkim.py
```diff
@@ -0,0 +1,71 @@
...
+        self.log.info("Unpacking main source %s to %s",
+                      self.src[0]['name'], self.builddir)
+        main_unpack_dir = extract_file(self.src[0]['path'], self.builddir, 
+            cmd=self.src[0]['cmd'], extra_options=self.cfg['unpack_options'])
+        if not main_unpack_dir:
+            raise EasyBuildError("Failed to unpack %s", self.src[0]['name'])
+        self.src[0]['finalpath'] = main_unpack_dir
```
boegel: making this `if...else` would be more readable imho

schiotz:
@boegel Thanks for your style changes - I have merged your merge rq into my merge rq - this is getting kind of convoluted :-)

schiotz:
@boegel 
OpenKIM is kind of a moving target.  With version 1.9, it is required that the API is installed, before the models are compiled.  Version 1.8 allowed it to be done simultaneously.  The change was presumably made to introduce a tool that makes it easy for the user to install models himself, reducing the need for installing both simultaneously.  

Version 1.9 then requires a much simpler EasyBlock, so I am closing this PR and have opened PR #1254 replacing it.

PR #1205, opened 2017-07-05
Update sanity tests for OpenFOAM, add logging

olesenm:
- this allows redirection of the Allwmake output to a log file so that
  it is possible to debug any errors with the build process and have a
  record of the configured build options.

### DIFF
--- a/easybuild/easyblocks/o/openfoam.py
+++ b/easybuild/easyblocks/o/openfoam.py
@@ -329,12 +329,12 @@ class EB_OpenFOAM(EasyBlock):
 
         # some randomly selected binaries
         # if one of these is missing, it's very likely something went wrong
-        bins = [os.path.join(self.openfoamdir, "bin", x) for x in ["foamExec", "paraFoam"]] + \
+        bins = [os.path.join(self.openfoamdir, "bin", x) for x in ["paraFoam"]] + \
                [os.path.join(toolsdir, "buoyant%sSimpleFoam" % x) for x in ["", "Boussinesq"]] + \
                [os.path.join(toolsdir, "%sFoam" % x) for x in ["boundary", "engine", "sonic"]] + \
                [os.path.join(toolsdir, "surface%s" % x) for x in ["Add", "Find", "Smooth"]] + \
-               [os.path.join(toolsdir, x) for x in ["deformedGeom", "engineSwirl", "modifyMesh",
-                                                    "refineMesh", "wdot"]]
+               [os.path.join(toolsdir, x) for x in ["checkMesh", "deformedGeom", "engineSwirl",
+                                                    "modifyMesh", "refineMesh"]]
         # check for the Pstream and scotchDecomp libraries, there must be a dummy one and an mpi one
         if 'extend' in self.name.lower():
             libs = [os.path.join(libsdir, "libscotchDecomp.%s" % shlib_ext),
--- a/easybuild/easyblocks/o/openfoam.py
+++ b/easybuild/easyblocks/o/openfoam.py
@@ -281,6 +281,9 @@ class EB_OpenFOAM(EasyBlock):
                 "Cleaning .*",
             ]
             run_cmd_qa(cmd_tmpl % 'Allwmake.firstInstall', qa, no_qa=noqa, log_all=True, simple=True)
+        elif LooseVersion(self.version) >= LooseVersion('1600'):
+            # Use Allwmake -log option if possible since this can be useful during builds, but also afterwards
+            run_cmd(cmd_tmpl % 'Allwmake -log', log_all=True, simple=True, log_output=True)
         else:
             run_cmd(cmd_tmpl % 'Allwmake', log_all=True, simple=True, log_output=True)
 


boegel:
problem with the tests should be fixed now that https://github.com/easybuilders/easybuild-framework/pull/2258 is merged; I have retriggered the tests...

olesenm:
Hello @easybuilders/easybuild-easyblocks-maintainers - please review

easybuild/easyblocks/o/openfoam.py
```diff
@@ -281,6 +281,9 @@ def build_step(self):
                 "Cleaning .*",
             ]
             run_cmd_qa(cmd_tmpl % 'Allwmake.firstInstall', qa, no_qa=noqa, log_all=True, simple=True)
+        elif LooseVersion(self.version) >= LooseVersion('1600'):
```
boegel: Hmm, this version check looks wrong? OpenFOAM versions are like `2.3.1`, `3.0.1`, etc.

So, the condition here would never be met?

easybuild/easyblocks/o/openfoam.py
```diff
@@ -326,12 +329,12 @@ def sanity_check_step(self):
...
         # if one of these is missing, it's very likely something went wrong
-        bins = [os.path.join(self.openfoamdir, "bin", x) for x in ["foamExec", "paraFoam"]] + \
+        bins = [os.path.join(self.openfoamdir, "bin", x) for x in ["paraFoam"]] + \
                [os.path.join(toolsdir, "buoyant%sSimpleFoam" % x) for x in ["", "Boussinesq"]] + \
                [os.path.join(toolsdir, "%sFoam" % x) for x in ["boundary", "engine", "sonic"]] + \
                [os.path.join(toolsdir, "surface%s" % x) for x in ["Add", "Find", "Smooth"]] + \
-               [os.path.join(toolsdir, x) for x in ["deformedGeom", "engineSwirl", "modifyMesh",
-                                                    "refineMesh", "wdot"]]
+               [os.path.join(toolsdir, x) for x in ["checkMesh", "deformedGeom", "engineSwirl",
+                                                    "modifyMesh", "refineMesh"]]
```
boegel: Why these changes to the sanity check, they are unrelated to the `-log` change?

easybuild/easyblocks/o/openfoam.py
olesenm: Actually we now have versions following YYMM, so 1606, 1612, 1706 thus far with 1706 having been released at the end of June-2017.
http://openfoam.com/releases/openfoam-v1706/

easybuild/easyblocks/o/openfoam.py
olesenm: Among other things, the `wdot` utility has been deprecated. There is a shell script placeholder (in a different location) that now only produces a message about it being redundant.
Since the comments in the sanity check claim to have some arbitrarily chosen applications, I simply picked some newer ones that are more likely to continue existing for a while.

easybuild/easyblocks/o/openfoam.py
boegel: Ah, OK, I overlooked that, my apologies.

So, the `-log` is only supported since the `1600` version?

I'd also suggest to cleaning this up a bit, feels too copy-paste as it is now:

```python
else:
    cmd = 'Allwmake'
    if LooseVersion(self.version) >= LooseVersion('1600'):
        # instruct to redirect output of Allwmake to a log file, for easier debugging & future reference
        cmd += ' -log'
    run_cmd(cmd_tmpl % cmd, log_all=True, simple=True, log_output=True)

easybuild/easyblocks/o/openfoam.py
boegel: OK, thanks for clarifying the removal of `wdot`, makes sense.

I'd rather keep the others around next to `blockMesh` & `checkMesh`, since they may help in catching problems with the build, unless you have a good reason for getting rid of those too?

What about the removal of `foamExec`, has that been deprecated/removed in recent versions too?

olesenm: Sure, no harm keeping around others. If foamExec hasn't been removed, then I agree that it makes a bad sanity test. Probably the only really safe script to check is paraFoam, the others may be too much of a moving target.

easybuild/easyblocks/o/openfoam.py
olesenm: It was added in 1606. How would you clean this up? (Too much of the package looks like cut-n-paste already).

easybuild/easyblocks/o/openfoam.py
boegel: Hmm, not sure what you are saying w.r.t. foamExec... Was it removed in recent OpenFOAM versions?

easybuild/easyblocks/o/openfoam.py
boegel: Then you should modify the version check to `>= LooseVersion('1606')`?

See code snippet above for a suggestion on how to avoid copy-pasting.

boegel:
@olesenm ping on following up on this? It would be nice to get this merged in soon...

akesandgren:
Just a note on the version thingie. Remember that there are versions of openfoam from both openfoam.com and openfoam.org. The openfoam.org versions are still at 5.x which is approx the same as the 17xx version from openfoam.com. Some users want the openform.org version and the easyblock needs to handle both.

boegel:
For now, I don't think there's a problem, but that may change in the future if the `.org` and `.com` versions go further apart...

@olesenm Can we think of a way to discriminate between the `.org` vs the `.com` versions (preferably without relying on checking the format of the version number)?

If needed, we can consider renaming the `.com` versions to something else than `OpenFOAM`, like we did with `OpenFOAM-Extend`.

olesenm:
@boegel : first day back from holidays (thus unresponsive for a period).

For spack, we have `openfoam` as a virtual package and `openfoam-com`, `openfoam-org` and `foam-extend` as individual packages that are each a provider of the `openfoam` virtual package.
I think that this removes all ambiguity and doesn't try to take any position about what is considered *upstream* or *fork* etc.

The configuration files for `openfoam-com` and `openfoam-org` are *very* similar, but probably not enough to use the same diffs for them. In spack, we generate the new content directly (instead of patching), but don't know how well that fits into the easybuild philosophy, nor if it works well with the rest of the easybuild setup. The `foam-extend` configuration files look quite different, so need some separate treatment. Most of the commonly used core routines I tucked into the `openfoam-com` package and just import them as required for the others.

easybuild/easyblocks/o/openfoam.py
olesenm: The original design of foamExec was to help starting up with identical paths, versions on various machines when not using a scheduler. While this generally still works, the upgrades to OpenFOAM configuration files means that the foamExec provision for starting an alternative OpenFOAM version and/or an alternative prefix location is quite broken. So I think that it (and foamJob) are in need of an overhaul or removal. In either case, a perhaps slightly dubious future.

olesenm:
For `openfoam-com`, the `wmake/rules/General/general` file defines the following line:

    WM_VERSION = OPENFOAM_PLUS=1706

where `OPENFOAM_PLUS` comes from the internal name for the development line, but is also used to define the release as an integer (YYMM) for use as a compiler define. In retrospect, it might have been cleaner to have defined `OPENFOAM_COM`, but not important enough to go change it now.

The following snippet should work to catch openfoam-com, but does not provide any distinction between development and release branches (just the major API level).

    # openfoam.com (development or release branch)
    #
    # parse wmake/rules/General/general/rules for this type of content
    #    WM_VERSION = OPENFOAM_PLUS=1706
    # can be used instead of parsing $WM_PROJECT_VERSION at all
    foamType  = "openfoam-com"
    reVersion = re.compile(r'OPENFOAM_PLUS\s*=\s*([\d.]+)')
    try:
        for line in open(environ["WM_DIR"] + '/rules/General/general', 'r'):
            m = re.search(reVersion, line)
            if m:
                foamType  = "openfoam-com"
                foamVer   = m.group(1)
                break
    except (KeyError, IOError):
        pass

    # OR a fallback:
    m = re.match(r'v?(\d{4})\+?', verString)
    # openfoam.com (release version) - eg, v1606+, v1612, 1706 etc
    ...


For `foam-extend` a `WM_FORK=extend` variable is defined in the `etc/bashrc` file.

For `openfoam-org` there is nothing particular to help users distinguish the source.

boegel:
@olesenm Please take a look at the updates in https://github.com/olesenm/easybuild-easyblocks/pull/1, which should make this good to go, sorry about the holdup...

boegel:
Merged via https://github.com/easybuilders/easybuild-easyblocks/pull/1272, thanks @olesenm!

PR #1207, opened 2017-07-06
make Boost easyblock fully aware of (pre)configopts, (pre)buildopts and (pre)installopts

boegel:
fix for issue reported in https://github.com/easybuilders/easybuild-easyconfigs/issues/4819


boegel:
Tested with existing `Boost` easyconfigs, no problem.

Should await confirmation from @SethosII (either here or in https://github.com/easybuilders/easybuild-easyconfigs/issues/4819) that this indeed fixes the problem...

SethosII:
This fixes it for me, thanks!

PR #1210, opened 2017-07-09
drop check for 'ipython' in sanity check of Anacdona easyblock, to also support Miniconda

pescobar:
I have removed `ipython` from the sanity check so the existing anaconda easyblock supports installing both Anaconda and Miniconda (https://conda.io/miniconda.html)

IMHO it makes sense that when installing conda as dependency of another tool we use Miniconda as it only includes the basics. Anaconda also includes other libraries which maybe are not needed. This would also help to reduce the size of the installdir which can also be helpful in case we generate containers in the future

I added a test easyconfig here https://github.com/easybuilders/easybuild-easyconfigs/pull/4841


easybuild/easyblocks/a/anaconda.py
```diff
@@ -67,10 +67,10 @@ def make_module_req_guess(self):
 
     def sanity_check_step(self):
         """
-        Custom sanity check for Anaconda
+        Custom sanity check for Anaconda and Miniconda
         """
         custom_paths = {
-            'files': [os.path.join('bin', x) for x in ['2to3', 'conda', 'ipython', 'pydoc', 'python', 'sqlite3']],
+            'files': [os.path.join('bin', x) for x in ['2to3', 'conda', , 'pydoc', 'python', 'sqlite3']],
```
verdurin: Is this dangling ',' a typo?

boegel:
Makes perfect sense to me, I doubt we're going to miss any installation problems with `Anaconda` by only no longer checking for `ipython`...

Issue #1209, opened 2017-07-09
Make Bundle of Rpackage extensions install the R packages in a library subdir as opposed to in the root of the installdir


pneerincx:
In a complete R installation (plain vanilla R + extra packages) the R packages go into ${R_HOME}/library/, where ${R_HOME} is a subdir of ${EBROOTR}. When an R installation is devided into two easyconfigs:
* a plain vanilla R and
* an R bundle with extensions/packages
the extra R packages from the bundle end up in the root installdir of the bundle. E.g. in case the bundle would be named R-bundle with one R package named SomePackage it would be installed in  ${EBROOTRMINBUNDLE}/SomePackage. This is sub optimal, because the root installdir will also contain the easybuild subdir with installtion logs use easyconfig etc. and we don't want R to try to parse that as an R package. In addition you'd have to specify at least one package's subdir in the sanitycheck step:
```
sanity_check_paths = {
    'files': [],
    'dirs': ['SomePackage'],
}
```
This is a bit silly as the sanity of the Package is already checked by trying to load it in R, which is a more robust test. (`exts_filter = ("R -q --no-save", "library(%(ext_name)s)")`).
It would be better if the RPackage would always install the packages in a _library_ subdir both when used in a complete R easyconfig as well as in a Bundle easyconfig. We can then list:
```
sanity_check_paths = {
    'files': [],
    'dirs': ['library'],
}
```
as opposed to one or more random package sub dirs to satisfy sanity_check_paths.

boegel:
follow-up in #1211

PR #1211, opened 2017-07-09
Improved rpackage easyblock 

pneerincx:
Patch to make Bundle of Rpackage extensions install the R packages in a library subdir as opposed to in the root of the installdir. See issue #1209


boegel:
@pneerincx Hmm, this makes sense, but it's a backwards-incompatible change...
I'm not sure whether that's an issue though, since it would only affects future installations of R extensions as separate installations.

Thoughts @easybuilders/easybuild-easyblocks-maintainers?

Also, you should customise `sanity_check_step` here to automatically check on a non-empty `library` dir, so that doesn't need to be done in all easyconfigs?

easybuild/easyblocks/generic/rpackage.py
```diff
@@ -187,7 +187,14 @@ def run(self):
...
             # extension is being installed in a separate installation prefix
-            lib_install_prefix = self.installdir
+            lib_install_prefix = os.path.join(self.installdir, 'library')
+            try:
+                os.mkdir(lib_install_prefix)
+            except OSError, err:
+                if os.path.isdir(lib_install_prefix):
+                    pass
+                else:
+                    raise EasyBuildError("Failed to create library directory for R packages: %s", err)
```
boegel: the whole `try`/`except` block can be replaced by calling the `mkdir` function from `easybuild.tools.filetools`?

boegel:
awaiting feedback from @easybuilders/easybuild-easyblocks-maintainers for this one...

easybuild/easyblocks/generic/rpackage.py
boegel: @pneerincx ping on this remark?

PR #1212, opened 2017-07-10
use plumed-patch command rather than 'plumed patch' in GROMACS easyblock

victorusu:
Instead of calling the plumed program, we call the plumed-patch script, which does not require to have mpi on the system.
    
This is important because in some centers the login nodes do not have mpi enabled.


boegel:
Tested with existing GROMACS easyconfig that includes PLUMED as a dependency, works like a charm.

Thanks @victorusu!

PR #1217, opened 2017-07-12
remove 'provides' line from setup.py

boegel:
I had to remove the `provides` line from `setup.py` in order to be able to release EasyBuild v3.3.1 with a recent version of `setuptools`, to work around this problem:

```
Upload failed (400): provides: Must be a valid Python identifier.
error: Upload failed (400): provides: Must be a valid Python identifier.
```

As far as I can tell, no information is lost by dropping this line.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -94,7 +94,6 @@ setup(
         "Topic :: Software Development :: Build Tools",
     ],
     platforms = "Linux",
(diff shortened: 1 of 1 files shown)


boegel:
@damianam This is good to go too, cfr. https://github.com/easybuilders/easybuild-framework/pull/2275?

damianam:
Yep

PR #1218, opened 2017-07-13
update siesta easyblock for versions 4.0.1 and 4.1-b3

migueldiascosta:
in the recently released versions, 

- `LIBS=`, in `arch.make`, is followed by a newline, which is matched by `\s` in the regex substitution, resulting in a broken line
- two utils are cleaned after being built, resulting in failed sanity checks

this fixes both issues, will submit corresponding easyconfigs

@akesandgren please review?


boegel:
Tested both with existing Siesta easyconfigs and https://github.com/easybuilders/easybuild-easyconfigs/pull/4888, going in, thanks @migueldiascosta!

PR #1219, opened 2017-07-13
fix PYTHONPATH for OpenBabel Python bindings

ostueker:
From OpenBabel 2.4.0 on, the Python bindings are no longer installed directly in `${PREFIX}/lib` but rather under `${PREFIX}/lib/python2.7/site-packages` (for 2.7 Python).

This commit sets the correct PYTHONPATH in the generated module.

### DIFF
--- a/easybuild/easyblocks/o/openbabel.py
+++ b/easybuild/easyblocks/o/openbabel.py
@@ -8,7 +8,7 @@
 # Flemish Research Foundation (FWO) (http://www.fwo.be/en)
 # and the Department of Economy, Science and Innovation (EWI) (http://www.ewi-vlaanderen.be/en).
 #
-# http://github.com/hpcugent/easybuild
+# https://github.com/easybuilders/easybuild
 #
 # EasyBuild is free software: you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
@@ -30,6 +30,7 @@ EasyBuild support for OpenBabel, implemented as an easyblock
 """
 import os
 from easybuild.easyblocks.generic.cmakemake import CMakeMake
+from easybuild.easyblocks.generic.pythonpackage import det_pylibdir
 from easybuild.framework.easyconfig import CUSTOM
 from easybuild.tools.modules import get_software_root, get_software_version
 from easybuild.tools.systemtools import get_shared_lib_ext
@@ -41,7 +42,7 @@ class EB_OpenBabel(CMakeMake):
     @staticmethod
     def extra_options():
         extra_vars = {
-            'try_python': [True, "Try to build Open Babel's Python bindings. (-DPYTHON_BINDINGS=ON)", CUSTOM],
+            'with_python_bindings': [True, "Try to build Open Babel's Python bindings. (-DPYTHON_BINDINGS=ON)", CUSTOM],
         }
         return CMakeMake.extra_options(extra_vars)
 
@@ -60,7 +61,7 @@ class EB_OpenBabel(CMakeMake):
         self.cfg['configopts'] += "-DBUILD_GUI=OFF "
 
         root_python = get_software_root('Python')
-        if root_python and self.cfg['try_python']:
+        if root_python and self.cfg['with_python_bindings']:
             self.log.info("Enabling Python bindings")
             self.with_python = True
             shortpyver = '.'.join(get_software_version('Python').split('.')[:2])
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/o/openbabel.py
```diff
@@ -8,7 +8,7 @@
 # Flemish Research Foundation (FWO) (http://www.fwo.be/en)
 # and the Department of Economy, Science and Innovation (EWI) (http://www.ewi-vlaanderen.be/en).
 #
-# https://github.com/easybuilders/easybuild
+# http://github.com/hpcugent/easybuild
```
boegel: @ostueker This is wrong after the move to the `easybuilders` organisation, please change it back?

easybuild/easyblocks/o/openbabel.py
```diff
@@ -26,16 +26,30 @@
...
-
+from distutils.version import LooseVersion
 
 class EB_OpenBabel(CMakeMake):
     """Support for installing the OpenBabel package."""
 
+    @staticmethod
+    def extra_options():
+        extra_vars = {
+            'try_python': [True, "Try to build Open Babel's Python bindings. (-DPYTHON_BINDINGS=ON)", CUSTOM],
```
boegel: What's the added benefit of introducing this?

If you don't want the Python bindings, it suffices to just exclude the use of Python as a dependency?

If there's a good use case for it, I'd rename it to `with_python_bindings`, since the `try_` makes it sound like it's likely (& acceptable) to fail..

easybuild/easyblocks/o/openbabel.py
```diff
@@ -46,15 +60,17 @@ def configure_step(self):
...
             self.log.info("Enabling Python bindings")
+            self.with_python = True
             shortpyver = '.'.join(get_software_version('Python').split('.')[:2])
             self.cfg['configopts'] += "-DPYTHON_BINDINGS=ON "
             shlib_ext = get_shared_lib_ext()
             self.cfg['configopts'] += "-DPYTHON_LIBRARY=%s/lib/libpython%s.%s " % (root_python, shortpyver, shlib_ext)
             self.cfg['configopts'] += "-DPYTHON_INCLUDE_DIR=%s/include/python%s " % (root_python, shortpyver)
         else:
             self.log.info("Not enabling Python bindings")
+            self.with_python = False
```
boegel: Already initialised as `False`, no point in setting it to `False` again?

easybuild/easyblocks/o/openbabel.py
```diff
@@ -76,7 +92,15 @@ def sanity_check_step(self):
     def make_module_extra(self):
         """Custom variables for OpenBabel module."""
         txt = super(EB_OpenBabel, self).make_module_extra()
-        txt += self.module_generator.prepend_paths('PYTHONPATH', ['lib'])
+        if self.with_python:
+            if LooseVersion(self.version) >= LooseVersion('2.4'):
+                # since OpenBabel 2.4.0 the Python bindings under 
+                # ${PREFIX}/lib/python2.7/site-packages  rather than ${PREFIX}/lib
+                shortpyver = '.'.join(get_software_version('Python').split('.')[:2])
+                ob_pythonpath = 'lib/python%s/site-packages/' % (shortpyver)
```
boegel: we have a dedicated function for this in the `PythonPackage` generic easyblock, so this can simply become:

```python
ob_pythonpath = det_pylibdir()
```

In some cases it may be `lib64` rather than `lib`, etc., which is taken into account by `det_pylibdir()`

easybuild/easyblocks/o/openbabel.py
ostueker: > it suffices to just exclude the use of Python as a dependency?
Unfortunately that does not work if `Eigen` has been built with a Python dependency as OpenBabel will inherit it indirectly.

boegel:
This looks good to go, tested it on all existing `OpenBabel` easyconfigs, works as expected.

I tried building OpenBabel 2.4.1 in order to have a test case for this, but the installation failed because most tests failed.

@ostueker Have you been able to come up with a working easyconfig for `OpenBabel` 2.4.1? And if so, can you share it?

ostueker:
@boegel:  Our OpenBabel easyconfigs are located [here](https://github.com/ComputeCanada/easybuild-easyconfigs/tree/computecanada-master/easybuild/easyconfigs/o/OpenBabel)
However our setup is quite customized (names of toolchains and certain prereqs are installed via NIX). But basically I resort to ignoring the tests as well.  :-(

boegel:
I figured out the problem I was running into with the tests in OpenBabel 2.4.1, was caused by commenting out the patch that reinstates `$LD_LIBRARY_PATH` for the tests, which is required.

You weren't hitting this issue @ostueker because I guess you're using RPATH?

PR for easyconfig in https://github.com/easybuilders/easybuild-easyconfigs/pull/5533, this is good to go, thanks @ostueker!

ostueker:
@boegel : Yes, we're using RPATH.

PR #1220, opened 2017-07-19
Fixed wrong use of build_type in self.cfg that resulted in an exception raised

damianam:
The correct easyconfig parameter is `buildtype` not `build_type`


boegel:
Introduced in #993, only a problem if EasyBuild is configured with `--disable-mpi-tests`.

Thanks @damianam!

PR #1221, opened 2017-07-19
Updates for GAMESS-US 20170420R1 release

ostueker:
Also added codepath for sockets build.

See also my commit e77ec00 "GAMESS-US Apr 20 2017 R1 configs, patches"
in easybuild-easyconfigs.


boegel:
tested with existing GAMESS-US easyconfigs, looks good to go, thanks @ostueker!

Issue #1222, opened 2017-07-25
using prebuildopts when installing R packages


jdonners:
in this case, I need to set an environment variable for Rmpi to pass the load test. Pull request is on the way.

boegel:
@jdonners No need to open an issue if there's a pull request (in the works).

Let's follow up in #1223

PR #1223, opened 2017-07-25
#1222 Added prebuildopts to generic/rpackage.py

jdonners:
Enable prebuildopts for an extension to R

### DIFF
--- a/easybuild/easyblocks/generic/rpackage.py
+++ b/easybuild/easyblocks/generic/rpackage.py
@@ -155,7 +155,8 @@ class RPackage(ExtensionEasyBlock):
     def install_R_package(self, cmd, inp=None):
         """Install R package as specified, and check for errors."""
 
-        cmdttdouterr, _ = run_cmd(cmd, log_all=True, simple=False, inp=inp, regexp=False)
+        cmd_full = "%s %s" % (self.cfg['prebuildopts'], cmd)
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/generic/rpackage.py
```diff
@@ -155,7 +155,8 @@ def build_step(self):
     def install_R_package(self, cmd, inp=None):
         """Install R package as specified, and check for errors."""
 
-        cmdttdouterr, _ = run_cmd(cmd, log_all=True, simple=False, inp=inp, regexp=False)
+        cmd_full = "%s %s" % (self.cfg['prebuildopts'], cmd)
```
boegel: @jdonners Shouldn't this be `preinstallopts` (since the command being executed here is `R CMD INSTALL`?

Also, make it aware of `installopts` too while you're at it?

boegel: @jdonners ping?

boegel: @jdonners re-ping?

boegel: @jdonners ping? :)

jdonners: sorry, I should better check the issues where I'm mentioned..

is there even a separate build step for an R package? I could add preinstallopts like:
```python
cmd_full = "%s %s %s" % (self.cfg['prebuildopts'], self.cfg['preinstallopts'], cmd)
```
would that work?

PR #1224, opened 2017-07-25
Fix a break inside loading gdb-ia when using easybuild

CJCShadowsan:
Found an issue loading gdb - fixed it by undoing the break that the for loop introduces as a workaround.

Ideally probably need to look at this in more detail and fix the for loop itself, but for now this allows both the compiler and gdb to work.


boegel:
@CJCShadowsan Is this for a particular version of `icc`?

We should indeed tried to fix the logic that breaks this, can you provide some more details (error messages, etc.)?

boegel:
@CJCShadowsan ping?

CJCShadowsan:
Sorry, been on holiday for 2 weeks.

This was for icc v17.0.4:

[root@mgmt01 ~]# icc -v
icc version 17.0.4 (gcc version 6.3.0 compatibility)

Without the workaround I put in place, gdb goes missing from the path (even
though the installer puts it in the right place).

I can poke around a bit more if necessary

Chris

boegel:
I don't see how this change can fix an issue where `gdb` is no longer available via `$PATH`, since you're only fixing stuff related to `$LD_LIBRARY_PATH`... Please clarify @CJCShadowsan?

CJCShadowsan:
The libipt library for gdb-ia is required, and without updating $LD_LIBRARY_PATH that library doesn't get loaded, and gdb-ia is broken.

boegel:
@CJCShadowsan Sorry for dragging this so long, just gave this a quick test, works as intended, so going in, thanks!

CJCShadowsan:
No problem, i've got a few more I should get around to raising!

Chris

PR #1228, opened 2017-08-15
reinstalling MCR with --module-only results in the wrong value for MC…

mboisson:
…RROOT because it discovers the subdir only during the install step. This patch fixes this

### DIFF
--- a/easybuild/easyblocks/m/mcr.py
+++ b/easybuild/easyblocks/m/mcr.py
@@ -33,8 +33,9 @@ EasyBuild support for installing MCR, implemented as an easyblock
 @author: Fotis Georgatos (Uni.Lu, NTUA)
 @author: Balazs Hajgato (Vrije Universiteit Brussel)
 """
-import re
+import glob
 import os
+import re
 import shutil
 import stat
 from distutils.version import LooseVersion
@@ -54,7 +55,7 @@ class EB_MCR(PackedBinary):
         super(EB_MCR, self).__init__(*args, **kwargs)
         self.comp_fam = None
         self.configfilename = "my_installer_input.txt"
-        self.subdir = ''
+        self.subdir = None
 
     @staticmethod
     def extra_options():
@@ -114,7 +115,9 @@ class EB_MCR(PackedBinary):
 
     def sanity_check_step(self):
         """Custom sanity check for MCR."""
-        if not self.subdir: self.find_subdir()
+        self.set_subdir()
+        if not isinstance(self.subdir,basestring):
+            raise EasyBuildError("Could not identify which subdirectory to pick: %s" % self.subdir)
         custom_paths = {
             'files': [],
             'dirs': [os.path.join(self.subdir, 'bin', 'glnxa64')],
@@ -133,7 +136,7 @@ class EB_MCR(PackedBinary):
         """Extend PATH and set proper _JAVA_OPTIONS (e.g., -Xmx)."""
         txt = super(EB_MCR, self).make_module_extra()
 
-        if not self.subdir: self.find_subdir()
+        self.set_subdir()
 
         xapplresdir = os.path.join(self.installdir, self.subdir, 'X11', 'app-defaults')
         txt += self.module_generator.set_environment('XAPPLRESDIR', xapplresdir)
@@ -146,12 +149,11 @@ class EB_MCR(PackedBinary):
 
         return txt
 
-    def find_subdir(self):
+    def set_subdir(self):
+        if self.subdir: return
         # determine subdirectory (e.g. v84 (2014a, 2014b), v85 (2015a), ...)
-        import glob
         subdirs = glob.glob(os.path.join(self.installdir,'v[0-9][0-9]*'))
         if len(subdirs) == 1:
             self.subdir = subdirs[0]
         else:
-            raise EasyBuildError("Found multiple subdirectories, don't know which one to pick: %s", subdirs)
-
+            self.subdir = subdirs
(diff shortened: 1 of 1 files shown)


boegel:
@mboisson The tests try running `--module-only --force` in a situation where there is no installation, so no subdirectories are found at all, the code needs to account for that somehow...

This wasn't a problem before when this block of code was a part of `install_step`, since that step is skipped entirely under `--module-only`, but that's not the correct behaviour as you point out here.

easybuild/easyblocks/m/mcr.py
```diff
@@ -137,6 +131,17 @@ def make_module_extra(self):
         """Extend PATH and set proper _JAVA_OPTIONS (e.g., -Xmx)."""
         txt = super(EB_MCR, self).make_module_extra()
 
+        # determine subdirectory (e.g. v84 (2014a, 2014b), v85 (2015a), ...)
+        subdirs = os.listdir(self.installdir)
+        # skip the easybuild folder if it exists
+        if "easybuild" in subdirs:
+            i = subdirs.index("easybuild")
```
boegel: This shouldn't be hardcoded, since this can be configured (via `--logfile-format`).

Either determine the directory name to filter out via `log_path`, or use a regex to grab the directory named `v[0-9]*`?

With `log_path`:

```python
from easybuild.tools.config import log_path
...
    eb_logdir_name = os.path.basename(log_path())
    if eb_logdir_name in subdirs:
        subdirs.remove(eb_logdir_name)
```

mboisson:
Ok, I think I fixed the code. I'm using glob to find the directory based on patterns instead of deleting the log path.

mboisson:
> The tests try running --module-only --force in a situation where there is no installation, so no subdirectories are found at all, the code needs to account for that somehow...

@boegel, do you mean that running --module-only --force is actually expected to succeed when there is no installation at all ? The sanity_checks are expected to fail when there is no installation, and I thought the sanity_checks were expected to always be run, even with --module-only --force ?

boegel:
@mboisson Yes, using `--module-only --force` skips the sanity check, to allow you to preview the module file without actually going through the installation. It's possible you'll end up with a partial module file in that case though, since some parts of the module file are only included when the installation directory is there and populated. See also the note at the end of http://easybuild.readthedocs.io/en/latest/Partial_installations.html#module-only-only-regenerate.

So, ideally there is some kind of fallback mechanism to allow that things work out without having the installation directory populated...

easybuild/easyblocks/m/mcr.py
```diff
@@ -147,3 +145,13 @@ def make_module_extra(self):
         txt += self.module_generator.set_environment('MCRROOT', os.path.join(self.installdir, self.subdir))
 
         return txt
+
+    def find_subdir(self):
+        # determine subdirectory (e.g. v84 (2014a, 2014b), v85 (2015a), ...)
+        import glob
```
boegel: please move this `import` statement to the top of the easyblock (and keep `import` lines sorted alphabetically)

easybuild/easyblocks/m/mcr.py
```diff
@@ -137,6 +133,8 @@ def make_module_extra(self):
         """Extend PATH and set proper _JAVA_OPTIONS (e.g., -Xmx)."""
         txt = super(EB_MCR, self).make_module_extra()
 
+        if not self.subdir: self.find_subdir()
```
boegel: please change this to:

```python
if not self.subdir:
    self.find_subdir()
```

(it's correct, but doesn't match the usual code style)

alternatively: just call `self.set_subdir()` and bump out early if `self.subdir` is already defined

easybuild/easyblocks/m/mcr.py
```diff
@@ -147,3 +145,13 @@ def make_module_extra(self):
...
         return txt
+
+    def find_subdir(self):
+        # determine subdirectory (e.g. v84 (2014a, 2014b), v85 (2015a), ...)
+        import glob
+        subdirs = glob.glob(os.path.join(self.installdir,'v[0-9][0-9]*'))
+        if len(subdirs) == 1:
+            self.subdir = subdirs[0]
+        else:
+            raise EasyBuildError("Found multiple subdirectories, don't know which one to pick: %s", subdirs)
```
boegel: Strictly speaking this error message isn't 100% accurate, since it's also possible no subdirectories are found (for example under `--module-only --force` without actually having the installation in place)

My suggestion would be to make `None` the default value for `self.subdir`, set it to `'NOTFOUND'` in case no matching subdirectories are found.

For a full installation, the sanity check should take care of this not passing through during a 'proper' installation.

(also, drop the empty line at the end)

mboisson:
Ok, I think I fixed these. I did not use 'NOTFOUND' because I wanted the exception to still be meaningful and show the list. Instead, I set subdir to the list of directories, and in the sanity_check I raise the exception. If the sanity check is not run and the subdir is not found, I set it to NOTFOUND.

easybuild/easyblocks/m/mcr.py
```diff
@@ -147,3 +151,12 @@ def make_module_extra(self):
         txt += self.module_generator.set_environment('MCRROOT', os.path.join(self.installdir, self.subdir))
 
         return txt
+
+    def set_subdir(self):
+        if self.subdir: return
+        # determine subdirectory (e.g. v84 (2014a, 2014b), v85 (2015a), ...)
+        subdirs = glob.glob(os.path.join(self.installdir,'v[0-9][0-9]*'))
+        if len(subdirs) == 1:
+            self.subdir = subdirs[0]
```
boegel: This should be `os.path.basename(subdirs[0])`, since otherwise `module_generator.prepend_paths` will complain about using an absolute path...

Fixed in https://github.com/ComputeCanada/easybuild-easyblocks/pull/1

mboisson:
Merged your PR.

boegel:
Tested with existing `MCR` easyconfigs, good to go, thanks @mboisson!

PR #1229, opened 2017-08-15
Updated VTune easyblock to behave like sourcing the amplxe-vars.sh script

damianam:
(ie: do not add LD_LIBRARY_PATH, LIBRARY_PATH and CPATH)

Otherwise error like this appear:

``$PREFIX/VTune/2017_update4/vtune_amplifier_xe/lib64/libstdc++.so.6: version `GLIBCXX_3.4.21' not found (required by XXX)``

### DIFF
--- a/easybuild/easyblocks/a/advisor.py
+++ b/easybuild/easyblocks/a/advisor.py
@@ -42,16 +42,16 @@ class EB_Advisor(IntelBase):
         """Constructor, initialize class variables."""
         super(EB_Advisor, self).__init__(*args, **kwargs)
         if LooseVersion(self.version) < LooseVersion('2017'):
-            self.base_path = 'advisor_xe'
+            self.subdir = 'advisor_xe'
         else:
-            self.base_path = 'advisor'
+            self.subdir = 'advisor'
 
     def make_module_req_guess(self):
         """Find reasonable paths for Advisor"""
-        return _make_module_req_guess_tools(self)
+        return self.get_guesses_tools()
 
     def sanity_check_step(self):
         """Custom sanity check paths for Advisor"""
         binaries = ['advixe-cl', 'advixe-feedback', 'advixe-gui', 'advixe-runss', 'advixe-runtrc', 'advixe-runtc']
-        custom_paths = _get_custom_paths_tools(self,binaries)
+        custom_paths = self.get_custom_paths_tools(binaries)
         super(EB_Advisor, self).sanity_check_step(custom_paths=custom_paths)
--- a/easybuild/easyblocks/a/advisor.py
+++ b/easybuild/easyblocks/a/advisor.py
@@ -48,29 +48,12 @@ class EB_Advisor(IntelBase):
         else:
             self.base_path = 'advisor'
 
+    def make_module_req_guess(self):
+        """Find reasonable paths for Advisor"""
+        return _make_module_req_guess_tools(self)
+
     def sanity_check_step(self):
         """Custom sanity check paths for Advisor"""
-
-        custom_paths = {
-            'files': [],
-            'dirs': ['%s/bin64' % self.base_path, '%s/lib64' % self.base_path]
-        }
-
+        binaries = ['advixe-cl', 'advixe-feedback', 'advixe-gui', 'advixe-runss', 'advixe-runtrc', 'advixe-runtc']
+        custom_paths = _get_custom_paths_tools(self,binaries)
         super(EB_Advisor, self).sanity_check_step(custom_paths=custom_paths)
-
-    def make_module_req_guess(self):
-        """
-        A dictionary of possible directories to look for
-        """
-        guesses = super(EB_Advisor, self).make_module_req_guess()
-
-        guesses['PATH'] = [os.path.join(self.subdir, 'bin64')]
-
-        # make sure $CPATH, $LD_LIBRARY_PATH and $LIBRARY_PATH are not updated in generated module file,
-        # because that leads to problem when the libraries included with Advisor are being picked up
-        for key in ['CPATH', 'LD_LIBRARY_PATH', 'LIBRARY_PATH']:
-            if key in guesses:
-                self.log.debug("Purposely not updating $%s in Advisor module file", key)
-                del guesses[key]
-
-        return guesses
--- a/easybuild/easyblocks/a/advisor.py
+++ b/easybuild/easyblocks/a/advisor.py
@@ -29,8 +29,6 @@ EasyBuild support for installing the Intel Advisor XE, implemented as an easyblo
 @author: Damian Alvarez (Forschungszentrum Juelich GmbH)
 """
 
-import os
-
 from distutils.version import LooseVersion
 
 from easybuild.easyblocks.generic.intelbase import IntelBase
--- a/easybuild/easyblocks/a/advisor.py
+++ b/easybuild/easyblocks/a/advisor.py
@@ -61,16 +61,18 @@ class EB_Advisor(IntelBase):
         A dictionary of possible directories to look for
         """
         guesses = super(EB_Advisor, self).make_module_req_guess()
+        # List of environment variables that the module shouldn't set
+        do_not_set = ['CPATH', 'LD_LIBRARY_PATH', 'LIBRARY_PATH']
 
         lib_path = '%s/lib64' % self.base_path
         include_path = '%s/include' % self.base_path
  
         guesses.update({
-            'CPATH': [include_path],
-            'INCLUDE': [include_path],
-            'LD_LIBRARY_PATH': [lib_path],
-            'LIBRARY_PATH': [lib_path],
             'PATH': ['%s/bin64' % self.base_path],
         })
+
+        for ev in do_not_set:
+            if ev in guesses:
+                del guesses[ev]
 
         return guesses
--- a/easybuild/easyblocks/a/advisor.py
+++ b/easybuild/easyblocks/a/advisor.py
@@ -29,6 +29,8 @@ EasyBuild support for installing the Intel Advisor XE, implemented as an easyblo
 @author: Damian Alvarez (Forschungszentrum Juelich GmbH)
 """
 
+import os
+
 from distutils.version import LooseVersion
 
 from easybuild.easyblocks.generic.intelbase import IntelBase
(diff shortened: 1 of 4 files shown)


easybuild/easyblocks/v/vtune.py
```diff
@@ -73,25 +74,26 @@ def make_module_req_guess(self):
...
         guesses = super(EB_VTune, self).make_module_req_guess()
+        # List of environment variables that the module shouldn't set
+        do_not_set = ['CPATH', 'LD_LIBRARY_PATH', 'LIBRARY_PATH']
 
         if self.cfg['m32']:
             guesses.update({
                 'PATH': [os.path.join(self.subdir, 'bin32')],
-                'LD_LIBRARY_PATH': [os.path.join(self.subdir, 'lib32')],
-                'LIBRARY_PATH': [os.path.join(self.subdir, 'lib32')],
             })
```
boegel: since this is now a single update, let's change this to:

```python
guesses['PATH'] = [os.path.join(self.subdir, 'bin32')]
```

same below

easybuild/easyblocks/v/vtune.py
```diff
@@ -73,25 +74,26 @@ def make_module_req_guess(self):
...
-                'LD_LIBRARY_PATH': [os.path.join(self.subdir, 'lib64')],
-                'LIBRARY_PATH': [os.path.join(self.subdir, 'lib64')],
             })
 
         guesses.update({
-            'CPATH': [os.path.join(self.subdir, 'include')],
             'MANPATH': [os.path.join(self.subdir, 'man')],
         })
 
+        for ev in do_not_set:
```
boegel: `do_not_set` is not used anywhere else, so no need to introduce a variable

also, clarify with a comment why this is being done, and include (debug) log messages for kicking those out:

```python
# make sure $CPATH, $LD_LIBRARY_PATH and $LIBRARY_PATH are not updated in generated module file,
# because that leads to problem when the libraries included with VTune are being picked up
for key in ['CPATH', 'LD_LIBRARY_PATH', 'LIBRARY_PATH']:
    if key in guesses:
        self.log.debug("Purposely ot updating $%s in VTune module file", key)
        del guesses[key]

Comment (boegel) [reply easybuild/easyblocks/v/vtune.py]: @damianam ping?

Comment (damianam) [reply easybuild/easyblocks/v/vtune.py]: @boegel pong back at ya'!

**Comment by damianam** [comment]
I think it makes sense to define an `IntelTools` that inherits from `IntelBase`, and used as a base for `EB_Advisor`, `EB_Vtune` and `EB_Inspector`. Thoughts?

File: easybuild/easyblocks/a/advisor.py [hunk 3]
```diff
@@ -46,31 +46,12 @@ def __init__(self, *args, **kwargs):
         else:
             self.base_path = 'advisor'
 
+    def make_module_req_guess(self):
+        """Find reasonable paths for Advisor"""
+        return _make_module_req_guess_tools(self)
```
boegel: `self.make_module_req_guess_tools()`

easybuild/easyblocks/a/advisor.py
```diff
@@ -46,31 +46,12 @@ def __init__(self, *args, **kwargs):
...
     def sanity_check_step(self):
         """Custom sanity check paths for Advisor"""
-
-        custom_paths = {
-            'files': [],
-            'dirs': ['%s/bin64' % self.base_path, '%s/lib64' % self.base_path]
-        }
-
+        binaries = ['advixe-cl', 'advixe-feedback', 'advixe-gui', 'advixe-runss', 'advixe-runtrc', 'advixe-runtc']
+        custom_paths = _get_custom_paths_tools(self,binaries)
```
boegel: `custom_paths = self.get_custom_paths_tools(binaries)`

easybuild/easyblocks/i/inspector.py
```diff
@@ -44,7 +44,8 @@ def __init__(self, *args, **kwargs):
 
         # recent versions of Inspector are installed to a subdirectory
         self.subdir = ''
-        if LooseVersion(self.version) >= LooseVersion('2013_update7') and LooseVersion(self.version) < LooseVersion('2017'):
+        if LooseVersion(self.version) >= LooseVersion('2013_update7') and \
+           LooseVersion(self.version) < LooseVersion('2017'):
```
boegel: Ugh, not a big fan of the break via `\` (we don't do this anywhere else either).

If you want to shorten this, I'd suggest:

```python
loosever = LooseVersion(self.version)
if loosever >= LooseVersion('2013_update7') and loosever < LooseVersion('2017'):

File: easybuild/easyblocks/i/inspector.py [hunk 6]
```diff
@@ -70,45 +71,11 @@ def install_step(self):
...
-            })
-
-        guesses.update({
-            'CPATH': [os.path.join(self.subdir, 'include')],
-            'MANPATH': [os.path.join(self.subdir, 'man')],
-        })
-
-        return guesses
+        """Find reasonable paths for Inspector"""
+        return _get_guesses_tools(self)
```
boegel: `self.get_guesses_tools()`

easybuild/easyblocks/i/inspector.py
```diff
@@ -70,45 +71,11 @@ def install_step(self):
...
-            dirs = ['lib32', 'include']
-        else:
-            files = [os.path.join('bin64', b) for b in binaries]
-            dirs = ['lib64', 'include']
-
-        custom_paths = {
-            'files': [os.path.join(self.subdir, f) for f in files],
-            'dirs': [os.path.join(self.subdir, d) for d in dirs],
-        }
+        custom_paths = _get_custom_paths_tools(self, binaries)
```
boegel: `custom_paths = self.get_custom_paths_tools(binaries)`

easybuild/easyblocks/v/vtune.py
```diff
@@ -42,10 +42,13 @@ def __init__(self, *args, **kwargs):
         """Easyblock constructor; define class variables."""
         super(EB_VTune, self).__init__(*args, **kwargs)
 
-        # recent versions of Inspector are installed to a subdirectory
+        # recent versions of VTune are installed to a subdirectory
         self.subdir = ''
-        if LooseVersion(self.version) >= LooseVersion('2013_update12'):
+        if LooseVersion(self.version) >= LooseVersion('2013_update12') and \
+           LooseVersion(self.version) < LooseVersion('2018'):
```
boegel:
```python
loosever = LooseVersion(self.version)
if loosever >= LooseVersion('2013_update12') and loosever < LooseVersion('2018'):
    self.subdir = 'vtune_amplifier_xe'
elif loosever >= LooseVersion('2018'):
    self.subdir = 'vtune_amplifier'

File: easybuild/easyblocks/v/vtune.py [hunk 9]
```diff
@@ -68,45 +71,11 @@ def install_step(self):
...
-            })
-
-        guesses.update({
-            'CPATH': [os.path.join(self.subdir, 'include')],
-            'MANPATH': [os.path.join(self.subdir, 'man')],
-        })
-
-        return guesses
+        """Find reasonable paths for VTune"""
+        return _get_guesses_tools(self)
```
boegel: `self.get_guesses_tools()`

easybuild/easyblocks/v/vtune.py
```diff
@@ -68,45 +71,11 @@ def install_step(self):
...
-            dirs = ['lib32', 'include']
-        else:
-            files = ['bin64/%s' % x for x in binaries]
-            dirs = ['lib64', 'include']
-
-        custom_paths = {
-            'files': [os.path.join(self.subdir, f) for f in files],
-            'dirs': [os.path.join(self.subdir, d) for d in dirs],
-        }
+        custom_paths = _get_custom_paths_tools(self, binaries)
```
boegel: `custom_paths = self.get_custom_paths_tools(binaries)`

easybuild/easyblocks/generic/intelbase.py
```diff
@@ -111,6 +112,42 @@ def __init__(self, *args, **kwargs):
 
         self.install_components = None
 
+    def _get_guesses_tools(self):
```
boegel: @damianam I wouldn't make these hidden/private, since that is an indication these are/should only be used inside `IntelBase`, which is the opposite of what you're doing...

So, use `def get_guesses_tools(self):`

easybuild/easyblocks/generic/intelbase.py
```diff
@@ -111,6 +112,42 @@ def __init__(self, *args, **kwargs):
...
+        # make sure $CPATH, $LD_LIBRARY_PATH and $LIBRARY_PATH are not updated in generated module file,
+        # because that leads to problem when the libraries included with VTune/Advisor/Inspector are being picked up
+        for key in ['CPATH', 'LD_LIBRARY_PATH', 'LIBRARY_PATH']:
+            if key in guesses:
+                self.log.debug("Purposely not updating $%s in %s module file", key, self.name)
+                del guesses[key]
+
+        return guesses
+
+    def _get_custom_paths_tools(self, binaries):
```
boegel: same here, use `def _get_custom_paths_tools(self, binaries):`

easybuild/easyblocks/i/inspector.py
damianam: Ok. I've just done it like that because the vtune easyblock had that break in that point.

easybuild/easyblocks/generic/intelbase.py
damianam: Yep, I was a bit confused on the exact meaning of the underscore. I mean, I want these methods to be called exclusively by classes that inherited from `IntelBase`. But I think python don't have that kind of distinction.

boegel: The leading underscore is only a convention too, it doesn't actually prevent you to do anything, it's just an indication that you shouldn't.

I don't think there's a way to restrict these methods to only be called by specific easyblocks, at least not without making an `IntelTools` that sits on top of `IntelBase`...

boegel:
Tested with all existing `VTune`, `Advisor` and `Inspector` easyconfigs, lgtm so going in...

Thanks @damianam!

PR #1234, opened 2017-08-29
also define CUDA_ROOT env var pointing to install dir

pescobar:


boegel:
@pescobar For which software were you hitting this missing `$CUDA_ROOT`? A quick Google search seems to suggest this is software-specific (e.g. PyCUDA)?

pescobar:
@boegel I noticed that one of my users was hardcoding this env var in his bashrc and I thought it would be useful to define it in the module . 

I thought it was a widely used env var. but you are right that this is not a standard env var. Only few applications use it. I didn't notice it. 

I think this PR can be discarded but it's also true that having this env var doesn't hurt

boegel:
@easybuilders/easybuild-easyblocks-maintainers Thoughts on this?

It indeed doesn't hurt to let `CUDA` modules define `$CUDA_ROOT`, even though it's not standard?

boegel:
awaiting feedback from @easybuilders/easybuild-easyblocks-maintainers ...

boegel:
I'll go ahead and include this, under the assumption that things won't blow up.

If they do, we'll hear about it soon enough...

PR #1235, opened 2017-08-29
BUG: use pip3 for PythonPackage when Python>=3

### LINKED ISSUE #1232: pip command in PythonPackage should reflect Python version
The `PythonPackage`easyblock doesn't check if it needs  to use `pip3` command  instead of `pip`, see https://lists.ugent.be/wws/arc/easybuild/2017-08/msg00051.html.

@boegel already gave a hint:

> I think is your actual problem...
> When you specify "use_pip = True", EasyBuild only tries to use 'pip', it doesn't take into account that it may need to use 'pip3' when Python 3 is being used... 
> 
> The logic for this is in the generic PythonPackage easyblock [1].
> 
> Are you up for taking a stab to fix that?
> Basically you need to do something like this when Python 3 is being used:
> 
>     self.install_cmd = self.install_cmd.replace('pip', 'pip3')
> 
> This should probably be done in the 'prepare_python' method, since only there the Python module used as a dependency will be loaded.

andreas-h:
I'm not sure how to test this and am away from cluster for a week (and EB on my laptop doesn't work ...), so I'm pushing this now and see what the CI tests say ...

closes #1232

### DIFF
--- a/easybuild/easyblocks/generic/pythonpackage.py
+++ b/easybuild/easyblocks/generic/pythonpackage.py
@@ -45,7 +45,7 @@ from easybuild.framework.easyconfig import CUSTOM
 from easybuild.framework.extensioneasyblock import ExtensionEasyBlock
 from easybuild.tools.build_log import EasyBuildError
 from easybuild.tools.filetools import mkdir, rmtree2, which
-from easybuild.tools.modules import get_software_root
+from easybuild.tools.modules import get_software_root, get_software_version
 from easybuild.tools.run import run_cmd
 
 
@@ -293,6 +293,24 @@ class PythonPackage(ExtensionEasyBlock):
         else:
             raise EasyBuildError("Failed to pick Python command to use")
 
+        if self.cfg.get('use_pip'):
+            self.pip_cmd = 'pip'
+            python_ver = get_software_version('Python')
+            if python_ver:
+                if LooseVersion(python_ver) >= LooseVersion('3'):
+                    self.log.info("Python3 detected.  Will use pip3.")
+                    self.install_cmd = self.install_cmd.replace('pip', 'pip3')
+                    self.pip_cmd = 'pip3'
+                    self.log.debug("Setting 'install_cmd' to '%s'", self.install_cmd)
+            else:
+                # Apparently we're using system Python.  Find out version and modify pip command if needed
+                out, _ = run_cmd("%s -V" % (python), simple=False)
+                if LooseVersion(out.split()[1]) >= LooseVersion('3'):
+                    self.log.info("Using system Python, which is Python3.  Will use pip3.")
+                    self.install_cmd = self.install_cmd.replace('pip', 'pip3')
+                    self.pip_cmd = 'pip3'
+                    self.log.debug("Setting 'install_cmd' to '%s'", self.install_cmd)
+
         # set Python lib directories
         self.set_pylibdirs()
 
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/generic/pythonpackage.py
```diff
@@ -293,6 +293,24 @@ def prepare_python(self):
...
+            python_ver = get_software_version('Python')
+            if python_ver:
+                if LooseVersion(python_ver) >= LooseVersion('3'):
+                    self.log.info("Python3 detected.  Will use pip3.")
+                    self.install_cmd = self.install_cmd.replace('pip', 'pip3')
+                    self.pip_cmd = 'pip3'
+                    self.log.debug("Setting 'install_cmd' to '%s'", self.install_cmd)
+            else:
+                # Apparently we're using system Python.  Find out version and modify pip command if needed
+                out, _ = run_cmd("%s -V" % (python), simple=False)
```
boegel: @andreas-h What I had in mind was using `python -V` as a replacement of using `get_software_version`, not a mix of both (less code to maintain is better).

Also, there's a better/easier approach, via `sys.version`:

```
$ python -V
Python 3.6.1
$ python -c "import sys; print(sys.version_info[0])"
3
```

```
$ python -V
Python 2.7.5
$ python -c "import sys; print(sys.version_info[0])"
2
```

That said, I'm starting to have 2nd doubt about this change, it feels like we're working around a problem that shouldn't be there, see also my question at https://lists.ugent.be/wws/arc/easybuild/2017-08/msg00065.html

andreas-h:
If the expected behavior of Python3.6 is to have the `pip` executable instead of `pip3` then of course I agree that this should not be handled by EB.

I just checked and in my `eb/software/Python/3.6.1-intel-2017a/bin` directory, I have `pip3` and `pip3.6`, but no `pip`.

This might be an issue with Python3.6, because in my `eb/software/Python/3.5.1-intel-2016a/bin` directory, I have `pip`, `pip3`, and `pip3.5`.

andreas-h:
Here is the relevant part of a diff of the `.eb` files in `ebfiles_repo`:

    1c1
    < # Built with EasyBuild version 3.0.1 on 2016-12-27_17-18-29
    ---
    > # Built with EasyBuild version 3.3.1 on 2017-08-24_19-32-47
    3c3
    < version = '3.5.1'
    ---
    > version = '3.6.1'
    6c6,7
    < description = "Python is a programming language that lets you work more quickly and integrate your systems more effectively."
    ---
    > description = """Python is a programming language that lets you work more quickly and integrate your systems
    >  more effectively."""
    8,12c9,10
    < toolchain = {'name': 'intel', 'version': '2016a'}
    < toolchainopts = {'pic': True, 'opt': True, 'optarch': True}
    < 
    < numpyversion = '1.10.4'
    < scipyversion = '0.17.0'
    ---
    > toolchain = {'name': 'intel', 'version': '2017a'}
    > toolchainopts = {'pic': True}

boegel:
@andreas-h What's the motivation for excluding `pip` & `setuptools` as extensions, that's probably the reason why you don't have `pip` command in your Python 3.6 installation...

Why prefer running `pip install -U ...` as postinstall command?

boegel:
Clear motivation for this PR is missing, and there's been no activity for quite a while here, so I'll go ahead and close this.

If this is still relevant, please open a new PR

PR #1236, opened 2017-09-01
update ROOT easyblock to support recent versions that require using CMake, add sanity check, clean up/enhance make_module*

boegel:
required for https://github.com/easybuilders/easybuild-easyconfigs/pull/5081

### DIFF
--- a/easybuild/easyblocks/i/icc.py
+++ b/easybuild/easyblocks/i/icc.py
@@ -44,9 +44,6 @@ from easybuild.easyblocks.t.tbb import get_tbb_gccprefix
 from easybuild.tools.run import run_cmd
 from easybuild.tools.systemtools import get_shared_lib_ext
 
-# location system header files on Debian/Ubuntu that Intel compilers must consider
-USR_INC_X86_64 = '/usr/include/x86_64-linux-gnu'
-
 
 def get_icc_version():
     """Obtain icc version string via 'icc --version'."""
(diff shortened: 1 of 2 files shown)


easybuild/easyblocks/r/root.py
```diff
@@ -28,43 +28,106 @@
...
+                self.cfg.update('configopts', '-Dbuiltin_fftw3=OFF -DFFTW_DIR=%s' % fftw_root)
+
+            gsl_root = get_software_root('GSL')
+            if gsl_root:
+                self.cfg.update('configopts', '-DGSL_DIR=%s' % gsl_root)
+
+            mesa_root = get_software_root('Mesa')
+            if mesa_root:
+                self.cfg.update('configopts', '-DDOPENGL_INCLUDE_DIR=%s' % os.path.join(mesa_root, 'include'))
+                self.cfg.update('configopts', '-DDOPENGL_INCLUDE_DIR=%s' % os.path.join(mesa_root, 'lib', 'libGL.so'))
```
wpoely86: LIB?

boegel: Should be `-DOPENGL_gl_LIBRARY`... Apparently ROOT is able to find the library regardless, but this is clearly wrong, I'll fix it.

easybuild/easyblocks/i/icc.py
```diff
@@ -44,6 +44,9 @@
 from easybuild.tools.run import run_cmd
 from easybuild.tools.systemtools import get_shared_lib_ext
 
+# location system header files on Debian/Ubuntu that Intel compilers must consider
```
wpoely86: only intel, not gcc?

boegel: this is not supposed to be in here, urgh, will remove...

PR #1239, opened 2017-09-07
{chem} GAMESS-US: fix build hangs on mathlib selection.

ostueker:
@boegel found and issue that without MKL the build hangs at:

> Enter your choice of 'mkl' or 'atlas' or 'acml' or 'pgiblas' or 'none':

See:
 https://github.com/easybuilders/easybuild-easyconfigs/pull/4912#issuecomment-327832302


boegel:
@ostueker I gave this a quick try, still hangs, it keeps asking the same question mentioning `You didn't provide a valid math library name, try again`.

See also https://github.com/easybuilders/easybuild-easyconfigs/pull/4912#issuecomment-327865503.

PR #1241, opened 2017-09-08
add custom easyblock for Stata

boegel:


easybuild/easyblocks/s/stata.py
```diff
@@ -0,0 +1,79 @@
+##
+# Copyright 2009-2017 Ghent University
```
wpoely86: you seems to live in the past? 😉

easybuild/easyblocks/s/stata.py
```diff
@@ -0,0 +1,79 @@
...
+
+    def sanity_check_step(self):
+        """Custom sanity check for Stata."""
+        custom_paths = {
+            'files': ['stata', 'xstata'],
+            'dirs': [],
+        }
+        super(EB_Stata, self).sanity_check_step(custom_paths=custom_paths)
+
+        # make sure required libpng library is there for stata
```
wpoely86: why do this? It's not a dependency?

boegel: Stata requires a very old version of `libpng`, see https://github.com/easybuilders/easybuild-easyconfigs/pull/5102

easybuild/easyblocks/s/stata.py
```diff
@@ -0,0 +1,79 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with EasyBuild.  If not, see <http://www.gnu.org/licenses/>.
+##
+"""
+EasyBuild support for building and installing Stata, implemented as an easyblock
+"""
```
wpoely86: an author tag?

easybuild/easyblocks/s/stata.py
wpoely86: Add as comment to the code?

PR #1246, opened 2017-09-14
add support to SystemCompiler generic easyblock to specify custom compiler command rather than always using lowercase 'name'

boegel:


ocaisa:
This will get some love when #1106 does :P

boegel:
@ocaisa Planning to get to that...

ocaisa:
One issue, it's not really made clear that this information is only used to figure out what the compiler version is. It has no impact on what compiler will be used by the generated module.

What is the use case exactly?

ocaisa:
Is it for the case where you providing a full path to the compiler? If so, that is actually not handled by the easyblock, you would need to strip the path from the incoming command to get the (current) pattern matching to work

ocaisa:
Handling that case would be nice, and this seems like a good PR to do it in

boegel:
I was using this to generate a `GCCcore` module with the system `gcc`, but you handled that different & perhaps better in https://github.com/easybuilders/easybuild-easyblocks/pull/1106, so let's go with that.

PR #1247, opened 2017-09-15
Make RubyGem extension EB use GEM_ envs except if installed with Ruby itself

hajgato:
If I understood correctly, `GEM_PATH` and `GEM_HOME` do not have to be set if extensions installed from `Ruby`. At the old EB, GEM envs were set only if individual gems were installed, not as a bundle. For some reason `make_module_extra(self)` part does not work (also in the previous version)

### DIFF
--- a/easybuild/easyblocks/generic/rubygem.py
+++ b/easybuild/easyblocks/generic/rubygem.py
@@ -29,7 +29,6 @@ EasyBuild support for Ruby Gems, implemented as an easyblock
 @author: Kenneth Hoste (Ghent University)
 """
 import os
-import re
 import shutil
 
 import easybuild.tools.environment as env
--- a/easybuild/easyblocks/generic/rubygem.py
+++ b/easybuild/easyblocks/generic/rubygem.py
@@ -90,7 +90,7 @@ class RubyGem(ExtensionEasyBlock):
             raise EasyBuildError("Ruby module not loaded?")
 
         # this is the 'proper' way to specify a custom installation prefix: set $GEM_HOME
-        if self.cfg.__dict__.get('software_name') != 'Ruby':
+        if not self.is_extension or self.master.name != 'Ruby':
             env.setvar('GEM_HOME', self.installdir)
 
         bindir = os.path.join(self.installdir, 'bin')
@@ -100,6 +100,6 @@ class RubyGem(ExtensionEasyBlock):
         """Extend $GEM_PATH in module file."""
         txt = super(RubyGem, self).make_module_extra()
         # for stand-alone Ruby gem installs, $GEM_PATH needs to be updated
-        if self.cfg.__dict__.get('software_name') != 'Ruby':
+        if not self.is_extension or self.master.name != 'Ruby':
             txt += self.module_generator.prepend_paths('GEM_PATH', [''])
-        return super(RubyGem, self).make_module_extra(txt)
+        return txt
--- a/easybuild/easyblocks/generic/rubygem.py
+++ b/easybuild/easyblocks/generic/rubygem.py
@@ -29,6 +29,7 @@ EasyBuild support for Ruby Gems, implemented as an easyblock
 @author: Kenneth Hoste (Ghent University)
 """
 import os
+import re
 import shutil
 
 import easybuild.tools.environment as env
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/generic/rubygem.py
```diff
@@ -90,7 +90,7 @@ def install_step(self):
             raise EasyBuildError("Ruby module not loaded?")
 
         # this is the 'proper' way to specify a custom installation prefix: set $GEM_HOME
-        if not self.is_extension:
+        if self.cfg.__dict__.get('software_name') != 'Ruby':
```
boegel: This is a bit dirty, but I think the real question we should ask is why the existing check based on `self.is_extension` doesn't, it should, see https://github.com/easybuilders/easybuild-framework/blob/master/easybuild/framework/extensioneasyblock.py#L88

easybuild/easyblocks/generic/rubygem.py
```diff
@@ -100,6 +100,6 @@ def make_module_extra(self):
         """Extend $GEM_PATH in module file."""
         txt = super(RubyGem, self).make_module_extra()
         # for stand-alone Ruby gem installs, $GEM_PATH needs to be updated
-        if not self.is_extension:
+        if self.cfg.__dict__.get('software_name') != 'Ruby':
             txt += self.module_generator.prepend_paths('GEM_PATH', [''])
-        return txt
+        return super(RubyGem, self).make_module_extra(txt)
```
boegel: Does this actually change anything? What's there should work...

easybuild/easyblocks/generic/rubygem.py
hajgato: I think it does not work, because it is a Bundle. So the gems are installed as extension not as an EasyBlock (rubygem). But I might misunderstand the code.

easybuild/easyblocks/generic/rubygem.py
hajgato: I think the original code was simply not executed. This change, indeed does not make any change, I just simply copied this part from the rpackage EasyBlock.

easybuild/easyblocks/generic/rubygem.py
hajgato: @boegel So how to proceed here?

hajgato:
@boegel  Ping? Can we do something, or this PR is completely dead?

boegel:
@hajgato Sorry, I've been meaning to get back to this but keep getting distracted. My plan was to get this merged for EasyBuild v3.4.1, but other stuff got in between...

easybuild/easyblocks/generic/rubygem.py
boegel: Well, it's wrong as it is now, since we already call out to `super(RubyGem, self).make_module_extra()` when defining `txt`, see above.

It should be reinstated to just `return txt` imho.

easybuild/easyblocks/generic/rubygem.py
boegel: OK, I think I see the problem...

We need to have `$GEM_HOME` set both when installing individual Ruby gems as stand-alone modules, and also when installing a bundle of gems together, but not when installing extensions as a part of a `Ruby` installation (i.e. a "batteries included" Ruby installation).

So, I think we need to keep the `not self.is_extension` condition, but combine it with another condition via `or`, like this:

```python
# define $GEM_HOME for both stand-alone installations and bundles of gems,
# to ensure that they are being installed in the right location (i.e. not in parent the Ruby installation)
if not self.is_extension or self.master.name != 'Ruby':
    env.setvar('GEM_HOME', self.installdir)
```

To clarify: `self.master` is made available so extensions can access the parent software (in this case, the bundle that is being installed), see https://github.com/easybuilders/easybuild-framework/blob/master/easybuild/framework/extension.py#L57.

Can you give that a try and see if it works as expected?

boegel: @hajgato Ping on this?

hajgato: @boegel We have a maintenance period right now, so I will test it when we are online again.

hajgato:
@boegel  This works perfect.

boegel:
Thanks for the fix @hajgato!

PR #1237, opened 2017-09-03
enhance icc easyblock to inlude multipath include dir in $CPATH

boegel:
requires https://github.com/easybuilders/easybuild-framework/pull/2294

@akesandgren, @besserox, @vanzod, @geimer: feedback welcome, let me know if this works as expected...

PR #1249, opened 2017-09-15
allow that 'gcc -print-multiarch' fails in icc easyblock

boegel:
bugfix for #1237, `gcc -print-multiarch` may fail for old versions of (system) GCC


boegel:
@JackPerdue This should fix the problem you were hitting with EasyBuild v3.4.0 with an old system GCC...

JackPerdue:
Works much better here.  icc/ifort rebuilt on RHEL6.

JackPerdue:
As mentioned on Slack.  I don't think this is the proper solution.  The proper solution is to load the GCCcore module when that step is run.

boegel:
@JackPerdue Making sure `GCCcore` is loaded when `gcc -print-multiarch` is being run is done in https://github.com/easybuilders/easybuild-easyconfigs/pull/5133 and https://github.com/easybuilders/easybuild-easyconfigs/pull/5134

Nevertheless, this change  should be merged too, the whole point was that this was allowed to fail, and it clearly doesn't...

PR #1251, opened 2017-09-18
Apply os.remove() also to links, instead of shutil.rmtree

damianam:
I cancelled an installation and when trying again EB failed trying to pass a symlink to `shutil.rmtree`. This should fix it.

### DIFF
--- a/easybuild/easyblocks/generic/intelbase.py
+++ b/easybuild/easyblocks/generic/intelbase.py
@@ -46,7 +46,7 @@ from easybuild.framework.easyblock import EasyBlock
 from easybuild.framework.easyconfig import CUSTOM
 from easybuild.framework.easyconfig.types import ensure_iterable_license_specs
 from easybuild.tools.build_log import EasyBuildError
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/generic/intelbase.py
```diff
@@ -164,7 +164,7 @@ def clean_home_subdir(self):
             for tree in os.listdir(self.home_subdir_local):
                 self.log.debug("... removing %s subtree" % tree)
                 path = os.path.join(self.home_subdir_local, tree)
-                if os.path.isfile(path):
+                if os.path.isfile(path) or os.path.islink(path):
                     os.remove(path)
```
boegel: use `remove_file` instead from `filetools` (we don't have a `remove_dir` yet, and if we do, we should also add a `remove` that wraps around both `remove_file` and `remove_dir`)

boegel:
Did a quick test with this in place, looks good, going in, thanks @damianam!

PR #1253, opened 2017-09-19
Fixed impi.py to allow 2018 versions

damianam:


boegel:
@damianam So, there's no separate `mpivars.sh` anymore in the 2018 version?

easybuild/easyblocks/i/impi.py
```diff
@@ -139,12 +139,16 @@ def post_install_step(self):
...
+            if impiver >= LooseVersion('2018.0.128'):
+                script_paths = [os.path.join('intel64', 'bin')]
+            else:
+                script_paths = [os.path.join('intel64', 'bin'), os.path.join('mic', 'bin')]
             # fix broken env scripts after the move
-            for script in [os.path.join('intel64', 'bin', 'mpivars.csh'), os.path.join('mic', 'bin', 'mpivars.csh')]:
+            for script in [os.path.join(script_path,'mpivars.csh') for script_path in script_paths]:
                 for line in fileinput.input(os.path.join(self.installdir, script), inplace=1, backup='.orig.easybuild'):
                     line = re.sub(r"^setenv I_MPI_ROOT.*", "setenv I_MPI_ROOT %s" % self.installdir, line)
                     sys.stdout.write(line)
```
boegel: @damianam While you're touching this, do you mind using `apply_regex_substitutions` instead like we already do in other easyblocks?

damianam:
@boegel What do you mean by "separate `mpivars.sh`"? There is no `mic` directory.

boegel:
@damianam Well, in previous versions, there was a separate `mpivars.sh` for `mic`, seems like they merged it now with the other one?

bartoldeman:
@boegel just happening to look at this. Intel MPI 2018 dropped support for Knights Corner (1st generation Xeon Phi), which means mic.

boegel:
I guess that explains it, thanks for clarifying @bartoldeman!

easybuild/easyblocks/i/impi.py
damianam: Done. Hopefully without a type, I haven't tested it.

easybuild/easyblocks/i/impi.py
```diff
@@ -139,15 +140,17 @@ def post_install_step(self):
...
             # fix broken env scripts after the move
-            for script in [os.path.join('intel64', 'bin', 'mpivars.csh'), os.path.join('mic', 'bin', 'mpivars.csh')]:
-                for line in fileinput.input(os.path.join(self.installdir, script), inplace=1, backup='.orig.easybuild'):
-                    line = re.sub(r"^setenv I_MPI_ROOT.*", "setenv I_MPI_ROOT %s" % self.installdir, line)
-                    sys.stdout.write(line)
-            for script in [os.path.join('intel64', 'bin', 'mpivars.sh'), os.path.join('mic', 'bin', 'mpivars.sh')]:
-                for line in fileinput.input(os.path.join(self.installdir, script), inplace=1, backup='.orig.easybuild'):
-                    line = re.sub(r"^I_MPI_ROOT=.*", "I_MPI_ROOT=%s; export I_MPI_ROOT" % self.installdir, line)
-                    sys.stdout.write(line)
+            regex_subs = (r"^setenv I_MPI_ROOT.*", r"setenv I_MPI_ROOT %s" % self.installdir)
```
boegel: @damianam This has to be a list:

```python
regex_subs = [(..., ...)]
```

same below

Otherwise:

```
  File "/easybuild/easyblocks/i/impi.py", line 150, in post_install_step
    apply_regex_substitutions(os.path.join(self.installdir, script), regex_subs)
  File "/easybuild/tools/filetools.py", line 906, in apply_regex_substitutions
    for i, (regex, subtxt) in enumerate(regex_subs):
ValueError: too many values to unpack
```

boegel:
tested with a bunch of existing `impi` easyconfigs, works fine now, thanks @damianam!

PR #1254, opened 2017-09-19
EasyBlock supporting OpenKIM library (Take 3).

schiotz:
This PR replaces PR #1204.

Version 1.9 of the OpenKIM library requires that the API is installed first, and the models later.  It also makes it easy for the users to install one or more models manually (or all of them), making it less important to include the models.

### DIFF
--- a/easybuild/easyblocks/o/openkim_api.py
+++ b/easybuild/easyblocks/o/openkim_api.py
@@ -1,0 +1,59 @@
+##
+# Copyright 2009-2017 Ghent University
+#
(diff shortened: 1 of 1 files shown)


schiotz:
I would like to withdraw this merge request.

The OpenKIM developers have in the meantime released a new version.  The installation procedure of the new version is a completely standard ConfigureMake, and therefore does not need a special block.

boegel:
@schiotz Hah, ok, that's a lot easier.

Sorry for not following up on this, total lack of time on my end... :(

PR #1255, opened 2017-09-19
Add support for Intel MKL and Intel MPI to scalapack.

akesandgren:
Needed for giolf toolchain among others.


boegel:
tested with https://github.com/easybuilders/easybuild-easyconfigs/pull/5140, looks good to go, will await successful test reports first before merging

Issue #507, opened 2014-11-07
Loading GCC does not override cc system compiler...


ocaisa:
which can lead to the situation where (for example)
cc: 4.4
c++: 4.9
gcc: 4.9
g++: 4.9
which gives a nasty surprise for software that tries cc and c++ before gcc and g++ in its configuration phase, e.g. gromacs (git master). 

Gromacs uses cmake so this can be solved by setting the appropriate env vars when loading the compiler to override this behaviour (though this approach would need pushenv and a little thought before implementation)

boegel:
@ocaisa: ping on this; should we symlink `cc` to `gcc` when installing `GCC`?

ocaisa:
I think this is probably a site specific choice because it really is a matter of taste. At our site, I would probably say yes but I'd have to run it by our group first

boegel:
I'd say in general, you'd expect `cc` to be the same as `gcc`?

We can certainly support this so it can be disabled easily, but I would enable it by default?

boegel:
thoughts on this?

ocaisa:
I'm with you, I am implementing it at JSC today

boegel:
\o/

ocaisa:
For us I put the alias in the easyconfig but I'm guessing you would like it added to the EasyBlock?

boegel:
It's ok to 'break' backwards compatibility when fixing bugs, no?
I consider `cc` being picked up from the OS a bug.

Defining an alias is probably not the best solution, since aliases don't travel down to subshells (and `eb` runs commands like `make` in subshells).

So, I think we need to make `cc` a wrapper around `gcc` in GCC's `bin` subdir?

fgeorgatos:
fyi. 
I also thought that aliases are not inheritable (and they aren't in the generic case), 
yet `module refresh` under `/etc/profile.d` is an interesting way to handle this (look at the very end):
http://lmod.readthedocs.org/en/latest/070_standard_modules.html

boegel:
related to https://github.com/hpcugent/easybuild-framework/issues/1993

@ocaisa did you end up implementing a workaround for this?

ocaisa:
No workaround just
```
modaliases = {'cc': 'gcc'}
```

boegel:
But that's not enough to catch the use of `cc` in EasyBuild, since commands like `make` are run in a subshell (and aliases don't travel down subshells)?

ocaisa:
[...]

PR #1256, opened 2017-09-19
enhance GCC easyblock to also put symlinks for cc/c++/f77/f95 in place

boegel:
fix for #507

### DIFF
--- a/easybuild/easyblocks/g/gcc.py
+++ b/easybuild/easyblocks/g/gcc.py
@@ -52,6 +52,14 @@ from easybuild.tools.systemtools import check_os_dependency, get_os_name, get_os
 from easybuild.tools.systemtools import get_shared_lib_ext
 
 
+COMP_CMD_SYMLINKS = {
+    'cc': 'gcc',
+    'c++': 'g++',
+    'f77': 'gfortran',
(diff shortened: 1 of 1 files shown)


akesandgren:
I would personally not do the chdir in and out, and instead change the symlink to:
symlink(comp_cmd_symlinks[key], target, ... instead. To avoid the basename call and still keep the symlink short-relative

boegel:
Thanks for the suggestion @akesandgren!

Retested, works like a charm:

```
$ ls -l $EBROOTGCCCORE/bin | egrep ' cc| c\+\+| f77 | f95'
-rwxr-xr-x 4 vsc40023 vsc40023 4566920 Sep 20 15:36 c++
lrwxrwxrwx 1 vsc40023 vsc40023       3 Sep 20 15:36 cc -> gcc
lrwxrwxrwx 1 vsc40023 vsc40023       8 Sep 20 15:36 f77 -> gfortran
lrwxrwxrwx 1 vsc40023 vsc40023       8 Sep 20 15:36 f95 -> gfortran
```

akesandgren:
old_cmds should be taken from comp_cmd_symlinks and not written twice in the code.

PR #1261, opened 2017-10-11
Do not overwrite option in easyconfig…

ocaisa:
…instead ignore it and print an explanatory warning

### DIFF
--- a/easybuild/easyblocks/generic/systemcompiler.py
+++ b/easybuild/easyblocks/generic/systemcompiler.py
@@ -167,13 +167,6 @@ class SystemCompiler(Bundle, EB_GCC, EB_ifort):
         self.log.debug("Derived version/install prefix for system compiler %s: %s, %s",
                        compiler_name, self.compiler_version, self.compiler_prefix)
 
-        if self.compiler_prefix in ['/usr', '/usr/local']:
-            if self.cfg['generate_standalone_module']:
-                # Force off adding paths to module since unloading such a module would be a potential shell killer
-                self.cfg['generate_standalone_module'] = False
-                self.log.warning("Disabling option 'generate_standalone_module' since installation prefix is %s",
-                                 self.compiler_prefix)
-
         # If EasyConfig specified "real" version (not 'system' which means 'derive automatically'), check it
         if self.cfg['version'] == 'system':
             self.log.info("Found specified version '%s', going with derived compiler version '%s'",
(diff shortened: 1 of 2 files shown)


ocaisa:
Required for tests to pass in https://github.com/easybuilders/easybuild-easyconfigs/pull/413

easybuild/easyblocks/generic/systemcompiler.py
```diff
@@ -216,14 +209,19 @@ def make_module_req_guess(self):
...
         """
         if self.cfg['generate_standalone_module']:
-            if self.cfg['name'] in ['GCC','GCCcore']:
-                guesses = EB_GCC.make_module_req_guess(self)
-            elif self.cfg['name'] in ['icc']:
-                guesses = EB_icc.make_module_req_guess(self)
-            elif self.cfg['name'] in ['ifort']:
-                guesses = EB_ifort.make_module_req_guess(self)
+            if self.compiler_prefix in ['/usr', '/usr/local']:
+                # Force off adding paths to module since unloading such a module would be a potential shell killer
```
boegel: comment is incorrect

easybuild/easyblocks/generic/systemcompiler.py
```diff
@@ -216,14 +209,19 @@ def make_module_req_guess(self):
...
+                self.log.warning("Ignoring option 'generate_standalone_module' since installation prefix is %s",
```
boegel: maybe use `print_warning` rather than `log.warning`?

easybuild/easyblocks/generic/systemcompiler.py
```diff
@@ -216,14 +209,19 @@ def make_module_req_guess(self):
...
+                                 self.compiler_prefix)
             else:
-                raise EasyBuildError("I don't know how to generate module var guesses for %s", self.cfg['name'])
+                if self.cfg['name'] in ['GCC','GCCcore']:
```
boegel: This should not be indented under `else`, should always be done regardless of `compiler_prefix` in `/usr`?

easybuild/easyblocks/generic/systemcompiler.py
boegel: nvm

easybuild/easyblocks/generic/systemcompiler.py
boegel: nvm, that's the whole point of the `/usr` check :)

boegel:
lgtm, thanks @ocaisa!

PR #1263, opened 2017-10-12
add custom easyblock for SAS

boegel:


ocaisa:
Test report in https://github.com/easybuilders/easybuild-easyconfigs/pull/5208#issuecomment-336079680 so good to go

PR #1274, opened 2017-10-19
 Add exceptions for FFTW/3.3.6 on POWER with GCC 5/6/7

ocaisa:


easybuild/easyblocks/f/fftw.py
```diff
@@ -181,8 +183,20 @@ def run_all_steps(self, *args, **kwargs):
                 if (prec == 'single' and (self.asimd or self.neon)) or (prec == 'double' and self.asimd):
                     prec_configopts.append('--enable-neon')
 
+                # For POWER with GCC 5/6/7 and FFTW/3.3.6 we need to disable some settings for tests to pass
+                # (we do it last so as not to affect previous logic)
+                cpu_arch = get_cpu_architecture()
+                fftw_version = self.version
+                if cpu_arch in [POWER] and self.toolchain.comp_family() in [TC_CONSTANT_GCC] and \
+                        LooseVersion(fftw_version) == LooseVersion("3.3.6"):
+                    # See https://github.com/FFTW/fftw3/issues/59 which applies to GCC 5/6/7
```
ocaisa: @boegel There is no easy way to check the GCC version to be used in this easyblock, at the time this list of options is created the dependencies are not loaded.

boegel: please use `==` rather than `in` if there's only one thing to compare with, and try to avoid line wrapping

Also, shouldn't you use `<=` rather than `==` for the FFTW version? In addition, are you sure the problem is going to be fixed in the next FFTW release?

```python
cpu_arch = get_cpu_architecture()
fftw_ver = LooseVersion(self.version)
if cpu_arch == POWER and comp_fam == TC_CONSTANT_GCC and fftw_ver <= LooseVersion('3.3.6'):

Comment (boegel) [reply easybuild/easyblocks/f/fftw.py]: W.r.t. to the GCC version, that would be tricky to check at this stage, indeed.

Not sure if we should worry about it too much though, GCC 4.x is quite old now, and unlikely to be used on POWER anyway?

What kind of problems would you run into when using GCC 4.x?

You could add a check for the GCC version being used in the `prepare` method instead, and error out if it's too old for POWER, at that point the toolchain *is* loaded.

File: easybuild/easyblocks/f/fftw.py [hunk 2]
```diff
@@ -181,8 +183,20 @@ def run_all_steps(self, *args, **kwargs):
...
+                    if (prec == 'single'):
```
boegel: drop the `(...)`, not needed:

```python
if prec == 'single':

File: easybuild/easyblocks/f/fftw.py [hunk 3]
```diff
@@ -181,8 +183,20 @@ def run_all_steps(self, *args, **kwargs):
...
+                        prec_configopts.append('--disable-altivec')
+                    if (prec == 'double'):
+                        prec_configopts.append('--disable-vsx')
+
                 # append additional configure options (may be empty string, but that's OK)
-                self.cfg.update('configopts', [' '.join(prec_configopts) + common_config_opts])
+                self.cfg.update('configopts', [' '.join(prec_configopts) + ' ' + common_config_opts])
```
boegel: Since this is a bug fix, maybe this should be done in a separate PR? Trivial to merge...

ocaisa: Done in #1277

easybuild/easyblocks/f/fftw.py
ocaisa: So the system GCC on the Power system I'm using is `4.8.5` and my understanding is that these flags work fine with GCC4.8 and GCC4.9 (see https://github.com/FFTW/fftw3/issues/59#issue-143078832)...but I haven't checked myself and I'm not going to.
Leaving a blanket GCC check like the above is fine since they will do harm for GCC4 (apart from maybe make it a little slower) and we are not expecting people to willingly chose an old compiler on Power.

ocaisa: I don't have a guarantee the problem will be fixed in the next release but the bug is reported ( https://github.com/FFTW/fftw3/issues/59#issue-143078832 ) and I would prefer not to turn off these optimisations by default but rather run into the problem and then see that I need to include 3.3.7 or whatever. For earlier versions than 3.3.6, I don't know if there is a problem or not so I err on the side of caution.

boegel: OK on both not checking GCC version & only doing this for FFTW 3.3.6 for now.

easybuild/easyblocks/f/fftw.py
boegel: merged #1277, thanks

boegel:
Tested with existing FFTW easyconfig, and confirmation that it works as intended via test report in https://github.com/easybuilders/easybuild-easyconfigs/pull/5242

PR #1275, opened 2017-10-19
Add support for Spectrum MPI to the systemmpi easyblock

ocaisa:
requires https://github.com/easybuilders/easybuild-framework/pull/2329


boegel:
Good to go, see test report in https://github.com/easybuilders/easybuild-easyconfigs/pull/5242

PR #1282, opened 2017-10-27
Move initialisation in system compiler/mpi easyblocks to the prepare step

ocaisa:
Some of the initialisation in the `systemcompiler` and `systemmpi` easyblocks was done unnecessarily early and is causing test failures in the easyconfigs repo (see https://github.com/easybuilders/easybuild-easyconfigs/pull/5242#issuecomment-339981393 ).
Moving these initialisations to the `prepare_step` avoids these failures and also makes more sense.


easybuild/easyblocks/generic/systemcompiler.py
```diff
@@ -115,9 +115,33 @@ def extra_options():
         return extra_vars
 
     def __init__(self, *args, **kwargs):
-        """Extra initialization: determine system compiler version and prefix."""
+        """Extra initialization: keep track of values that may change due to modifications to the version."""
         super(SystemCompiler, self).__init__(*args, **kwargs)
 
+        # fix installdir and module names (may differ because of changes to version)
```
boegel: @ocaisa This doesn't make sense (anymore?), since we didn't touch `self.cfg['version']` at all? Neither did we before you reshuffled things (we only do in `make_module_step` and `make_module_extend_modpath`, but neither of those affect the module name or the install directory?

ocaisa: I think I was just preserving what was originally there, I need to dig a bit to understand the issue. Do you mean that all of this can be moved to `make_module_step`?

ocaisa: Ok, I think you are right, I can't see any any reason for the mns related stuff, we only need `self.orig_version` and `self.orig_installdir`

ocaisa: Done

easybuild/easyblocks/generic/systemcompiler.py
```diff
@@ -115,9 +115,29 @@ def extra_options():
         return extra_vars
 
     def __init__(self, *args, **kwargs):
-        """Extra initialization: determine system compiler version and prefix."""
+        """Extra initialization: keep track of values that may change due to modifications to the version."""
         super(SystemCompiler, self).__init__(*args, **kwargs)
 
+        # Keep track of original values of vars that are subject to change, for restoring later.
+        # The version is determined/matched from the installation and the installdir is determined from the system (the original is used
```
boegel: line too long, fixed in https://github.com/ocaisa/easybuild-easyblocks/pull/13, should then be good to go in

PR #1283, opened 2017-10-27
Fixed wrong sanity check in boost when using python3

damianam:


easybuild/easyblocks/b/boost.py
```diff
@@ -240,7 +240,11 @@ def sanity_check_step(self):
         if self.cfg['boost_mpi']:
             custom_paths["files"].append('lib/libboost_mpi.%s' % shlib_ext)
         if get_software_root('Python'):
-            custom_paths["files"].append('lib/libboost_python.%s' % shlib_ext)
+            pymajorver = get_software_version('Python').split('.')[0]
+            if int(pymajorver) >= 3:
+                custom_paths["files"].append('lib/libboost_python%s.%s' % (pymajorver, shlib_ext))
+            else:
+                custom_paths["files"].append('lib/libboost_python.%s' % shlib_ext)
```
boegel: To avoid copy-pasting, I'd prefer this:

```python
pymajorver = get_software_version('Python').split('.')[0]
if int(pymajorver) >= 3:
    suff = pymajorver
else:
    suff = ''
custom_paths["files"].append('lib/libboost_python%s.%s' % (suff, shlib_ext))

boegel:
Strictly speaking, this fix is only needed for recent version of Boost (it's not needed for Boost 1.63.0 and earlier, it seems), but since `libboost_python3.so` is also there for older versions when building on top of Python 3, I don't think there's a need to make this conditional on recent versions.

So, going in, thanks @damianam!

PR #1284, opened 2017-11-04
Changed HPCG configure syntax for v3.0 support 

vanzod:
In `HPCG-3.0` the configure command syntax is different from the one implemented for previous versions. This solves the issue and keeps compatibility with older versions.

### DIFF
--- a/easybuild/easyblocks/h/hpcg.py
+++ b/easybuild/easyblocks/h/hpcg.py
@@ -50,10 +50,10 @@ class EB_HPCG(ConfigureMake):
         # configure with most generic configuration available, i.e. hybrid
         # this is not specific to GCC or OpenMP, we take full control over that via $CXX and $CXXFLAGS
         if LooseVersion(self.version) >= LooseVersion('3.0'):
-            cmd = "../configure MPI_GCC_OMP"
+            arg = "MPI_GCC_OMP"
         else:
-            cmd = "../configure ../setup/Make.MPI_GCC_OMP"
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/h/hpcg.py
```diff
@@ -48,7 +49,10 @@ def configure_step(self):
         mkdir("obj")
         # configure with most generic configuration available, i.e. hybrid
         # this is not specific to GCC or OpenMP, we take full control over that via $CXX and $CXXFLAGS
-        cmd = "../configure ../setup/Make.MPI_GCC_OMP" 
+        if LooseVersion(self.version) >= LooseVersion('3.0'):
+            cmd = "../configure MPI_GCC_OMP"
+        else:
+            cmd = "../configure ../setup/Make.MPI_GCC_OMP"
```
boegel: @vanzod I'd only change the argument to `../configure`?

```python
if LooseVersion(self.version) >= LooseVersion('3.0'):
    arg = 'MPI_GCC_OMP'
else:
    arg = '../setup/Make.MPI_GCC_OMP'

run_cmd("../configure %s" % arg, log_all=True, simple=True, log_ok=True, path='obj')

boegel:
@vanzod Can you look into an easyconfig PR to accompany this?

boegel:
Tested with https://github.com/easybuilders/easybuild-easyconfigs/pull/5298 and existing HPCG easyconfigs, looking good, thanks @vanzod!

PR #1287, opened 2017-11-09
add easyblock for TensorFlow (REVIEW)

boegel:
With this easyblock, I'm able to build & install TensorFlow 1.4.0 from source using `foss/2017b` and Python 3.6.3 (cfr. https://github.com/easybuilders/easybuild-easyconfigs/pull/5318).

Further testing is required, especially w.r.t. building with Intel compilers on top of Intel MKL & GPU support (i.e., including CUDA & cuDNN as dependencies).

One thing not taking care of now is avoiding that the installation procedure pulls in a whole bunch of dependencies itself, it seems like Bazel doesn't have good support for providing the dependencies via some other way (meaning that a lot of patching is probably required to avoid that they're automagically installed).

Also, we can avoid the need for `run_cmd_qa` by setting `$TF_*` environment variables that are picked up by `configure`, it may be worth while switching to that approach.


pforai:
+1 on the env vars!

pforai:
The env vars based block doesn't hang any more since its no long based on the qanda block :thumbsup:

easybuild/easyblocks/t/tensorflow.py
```diff
@@ -0,0 +1,260 @@
...
+            adjust_permissions(icc_wrapper, stat.S_IXUSR)
+            env.setvar('PATH', ':'.join([os.path.dirname(icc_wrapper), os.getenv('PATH')]))
+            self.log.info("Using wrapper script for 'icc': %s", which('icc'))
+
+        self.prepare_python()
+
+        cuda_root = get_software_root('CUDA')
+        cudnn_root = get_software_root('cuDNN')
+        jemalloc_root = get_software_root('jemalloc')
+        opencl_root = get_software_root('OpenCL')
```
damianam: There isn't any `OpenCL` easyconfig as far as I can see. Besides, the `OpenCL` runtime for intel processors needs root permissions, so it won't be made available any time soon through EB.

easybuild/easyblocks/t/tensorflow.py
```diff
@@ -0,0 +1,260 @@
...
+        # https://docs.bazel.build/versions/master/user-manual.html#flag--verbose_failures
+        cmd.extend(['--subcommands', '--verbose_failures'])
+
+        if self.toolchain.options.get('pic', None):
+            cmd.append('--copt="-fPIC"')
+
+        cmd.append(self.cfg['buildopts'])
+
+        # pass through environment variables that may specify location of license file for Intel compilers
+        for key in []:
```
damianam: I have the feeling that this shouldn't be an empty list....

easybuild/easyblocks/t/tensorflow.py
```diff
@@ -0,0 +1,260 @@
...
+
+        if cuda_root:
+            cmd.append('--config=cuda')
+
+        if self.cfg['with_mkl_dnn']:
+            # this makes TensorFlow download & use mkl-dnn (cfr. https://github.com/01org/mkl-dnn)
+            # using the full Intel MKL doesn't work without additional effort...
+            cmd.extend(['--config=mkl'])
+
+        cmd.append('//tensorflow/tools/pip_package:build_pip_package')
```
damianam: I am not sure what this does, but I going to guess that the double slash is a typo.

easybuild/easyblocks/t/tensorflow.py
```diff
@@ -0,0 +1,260 @@
...
+            'with_mkl_dnn': [True, "Make TensorFlow use Intel MKL-DNN", CUSTOM],
+        }
+        return PythonPackage.extra_options(extra_vars)
+
+    def configure_step(self):
+        """Custom configuration procedure for TensorFlow."""
+
+        tmpdir = tempfile.mkdtemp(suffix='-bazel-configure')
+
+        # create wrapper for icc to make sure location of license server is available...
```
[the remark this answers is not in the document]
boegel: ok, will do

easybuild/easyblocks/t/tensorflow.py
boegel: Well, ok, but I'll leave this in anyway, to avoid hardcoding building without `OpenCL` support.

At least this way there is an option to built TF with OpenCL support (for example via http://easybuild.readthedocs.io/en/latest/Using_external_modules.html).

easybuild/easyblocks/t/tensorflow.py
boegel: Woops, yes, I did the for for `$INTEL_LICENSE_FILE` through this initially, but it turns out that doesn't actually work...

easybuild/easyblocks/t/tensorflow.py
boegel: No, it's not, that's how Bazel works... See https://www.tensorflow.org/install/install_sources#build_the_pip_package .

boegel:
I made this `WIP` again, since it needs more testing by others, and there's two known issues:

* CUDA builds are not working yet
* if you use the TensorFlow easyblock with `intel` toolchain and have either `--use-ccache` or `--rpath` enabled, then the wrapper around `icc` that is being used results in a fork bomb (`icc` wrapper calls `ccache` wrapper which calls `icc` wrapper, etc.); that should be fixed

JackPerdue:
For whatever reason, it kept failing here because Bazel was trying to use system gcc includes (4.8.5 on RHEL7) instead of the EB installed GCCcore 6.3.0.  This was with our variant of intel/2017a (which uses the system binutils which happened to cause another problem but not near as difficult as the Bazel problem).   If I run xxd (hex dump) on the bazel binary I can see reference to both GCCs..... unfortunately it grabs the 48.5 when building TF.

My workaround for now is "rpm -e gcc --nodeps" (remove system gcc and includes) while buiding TF... works much better but isn't really a solution I would recommend.

JackPerdue:
OK... figured out the hang when loading CUDA.  Enabling CUDA seems to automatically require cuDNN in configure.py.   When checking for the library it looks for libcudnn.%(version)s.  However, at present, that symlink is not created (e.g. for 6.0-CUDA-8.0.44 I get libcudnn.so  libcudnn.so.6  libcudnn.so.6.0.21 (from the tar ball).  Note, there is no libcudnn.so.6.0.  Also note that the tarball provides 3 duplicate versions of the library instead of creating sensible symlinks.

Due to the missing library, EB goes into la la land (infinite loop) since configure keeps asking for the version and checking for the library.  My solution was to change cuDNN-6.0 to cuDNN-6.0.21.  A better workaround might be to remove libcudnn.so and libcudnn.so.6 and then create symlinks for those and libcudnn.so.6.0 all pointing at libcudnn.so.6.0.21.  

Note, I would change the logic for checking for CUDA/cuDNN seperately.  As currently coded the easyblock allows them to be enabled seperately (via dependencies)..  I think it would be more appropriate that if CUDA is detected then cuDNN must be loaded as well (and error if it isn't).  The TF configure.py requires cuDNN if CUDA is there so might as well make the easyblock do that as well.

That got to the build stage when it failed on cuda_compute_capabilities.   The current default is is an empty list.  I'd suggest change the default to something like:

cuda_compute_capabilities = ['3.0', '3.5', '3.7', '5.2', '6.0', '6.1', '7.0']   # current Data Center Teslas

and then let the user tweak in the easyconfig if they only need a certain subset.

Other notes....

If binutils_root = get_software_root('binutils') returns null then binutils_root should be set to /usr, not cause an error.

Wouldn't hurt (for troubleshooting) to add preconfigopts... e.g.

run_cmd('%(preconfigopts)s ./configure' % {'preconfigopts': self.cfg['preconfigopts'],}, log_all=True, simple=True)

Anyway... that got me further.  Currently fails because it wants something compiled with -fPIC.... just need to figure out what that is.

JackPerdue:
-fPIC  was easily solved with toolchainopt pic=True.  Now getting an error while building stuff in external/eigen_archive/unsupported/Eigen/CXX11/../../../Eigen/src/Core/util/XprHelper.h .  Tried toolchain opt cstd=c++ but didn't help.

JackPerdue:
[TensorFlow-1.4.0-intel-2017A-Python-3.5.2-CUDA-8.0.44.eb.txt](https://github.com/easybuilders/easybuild-easyblocks/files/1640047/TensorFlow-1.4.0-intel-2017A-Python-3.5.2-CUDA-8.0.44.eb.txt)

easybuild/easyblocks/t/tensorflow.py
```diff
@@ -0,0 +1,264 @@
...
+
+class EB_TensorFlow(PythonPackage):
+    """Support for building/installing TensorFlow."""
+
+    @staticmethod
+    def extra_options():
+        extra_vars = {
+            # see https://developer.nvidia.com/cuda-gpus
+            'cuda_compute_capabilities': [[], "List of CUDA compute capabilities to build with", CUSTOM],
+            'with_mkl_dnn': [True, "Make TensorFlow use Intel MKL-DNN", CUSTOM],
```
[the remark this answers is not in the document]
damianam: `mkl-dnn` and `imkl` are two different things. `mkl-dnn` is not part of the toolchain, and is totally independent from anything else (compiler, MPI, other math libraries). It can be automatically downloaded by Tensorflow during installation, which is what this option does.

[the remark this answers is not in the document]
akesandgren: Yes, but you don't want it unless you are using mkl and especially not if you are using Cuda and cudnn.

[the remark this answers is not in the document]
damianam: I am not sure what you mean. For sure you don't want it when using CUDA and cuDNN, but why wouldn't you want to use mkl-dnn without the full mkl?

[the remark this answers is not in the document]
akesandgren: Misread the instructions a bit. No, you don't need mkl, but the instructions from TensorFlow and mkl-ddn is a bit unclear.

So, at the very least it should be turned off when CUDA and cuDNN are enabled.

ocaisa:
I just read those instructions and I too was confused!

easybuild/easyblocks/t/tensorflow.py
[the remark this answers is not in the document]
damianam: Agreed.

easybuild/easyblocks/t/tensorflow.py
```diff
@@ -0,0 +1,264 @@
...
+        if self.toolchain.comp_family() == toolchain.INTELCOMP:
+            icc_wrapper_txt = INTEL_COMPILER_WRAPPER % {
+                'compiler_path': which('icc'),
+                'intel_license_file': os.getenv('INTEL_LICENSE_FILE', os.getenv('LM_LICENSE_FILE')),
+            }
+            icc_wrapper = os.path.join(tmpdir, 'bin', 'icc')
+            write_file(icc_wrapper, icc_wrapper_txt)
+            adjust_permissions(icc_wrapper, stat.S_IXUSR)
+            env.setvar('PATH', ':'.join([os.path.dirname(icc_wrapper), os.getenv('PATH')]))
+            self.log.info("Using wrapper script for 'icc': %s", which('icc'))
```
damianam: The wrapper creation is not listed in the extended dry run. I think it should be there too.

easybuild/easyblocks/t/tensorflow.py
```diff
@@ -0,0 +1,264 @@
...
+
+        cmd.append(self.cfg['buildopts'])
+
+        if cuda_root:
+            cmd.append('--config=cuda')
+
+        if self.cfg['with_mkl_dnn']:
+            # this makes TensorFlow download & use mkl-dnn (cfr. https://github.com/01org/mkl-dnn)
+            # using the full Intel MKL doesn't work without additional effort...
+            cmd.extend(['--config=mkl'])
```
damianam: Add support to use an external `mkl-dnn`

easybuild/easyblocks/t/tensorflow.py
```diff
@@ -0,0 +1,264 @@
...
+                'TF_CUDA_VERSION': get_software_version('CUDA'),
+            })
+        if cudnn_root:
+            config_env_vars.update({
+                'CUDNN_INSTALL_PATH': cudnn_root,
+                'TF_CUDNN_VERSION': get_software_version('cuDNN'),
+            })
+
+        for (key, val) in sorted(config_env_vars.items()):
+            env.setvar(key, val)
```
damianam: All these should be listed in the extended dry run too.

akesandgren:
I get a different failure.
========
ERROR: /scratch/eb-bWcAFk/tmpieLI4v-bazel-build/external/nccl_archive/BUILD:33:1
: undeclared inclusion(s) in rule '@nccl_archive//:nccl':
this rule is missing dependency declarations for the following files included by
 'external/nccl_archive/src/libwrap.cu.cc':
========

It gets this when building
>>>>>>>>> # @pcre//:pcre [action 'Compiling external/pcre/pcre_newline.c [for host]']

boegel:
@akesandgren Try again with this patch (will probably still fail, but differently, see also https://github.com/tensorflow/tensorflow/issues/10665)

```diff
--- tensorflow/contrib/BUILD.orig       2017-11-11 20:39:11.034107205 +0100
+++ tensorflow/contrib/BUILD    2017-11-11 20:39:30.134085519 +0100
@@ -56,7 +56,6 @@
         "//tensorflow/contrib/memory_stats:memory_stats_py",
         "//tensorflow/contrib/meta_graph_transform",
         "//tensorflow/contrib/metrics:metrics_py",
-        "//tensorflow/contrib/nccl:nccl_py",
         "//tensorflow/contrib/ndlstm",
         "//tensorflow/contrib/nearest_neighbor:nearest_neighbor_py",
         "//tensorflow/contrib/nn:nn_py",
@@ -102,7 +101,6 @@
         "//tensorflow/contrib/factorization/kernels:all_kernels",
         "//tensorflow/contrib/input_pipeline:input_pipeline_ops_kernels",
         "//tensorflow/contrib/layers:sparse_feature_cross_op_kernel",
-        "//tensorflow/contrib/nccl:nccl_kernels",
         "//tensorflow/contrib/nearest_neighbor:nearest_neighbor_ops_kernels",
         "//tensorflow/contrib/seq2seq:beam_search_ops_kernels",
         "//tensorflow/contrib/tensor_forest:model_ops_kernels",
@@ -123,7 +121,6 @@
         "//tensorflow/contrib/framework:all_ops",
         "//tensorflow/contrib/input_pipeline:input_pipeline_ops_op_lib",
         "//tensorflow/contrib/layers:sparse_feature_cross_op_op_lib",
-        "//tensorflow/contrib/nccl:nccl_ops_op_lib",
         "//tensorflow/contrib/nearest_neighbor:nearest_neighbor_ops_op_lib",
         "//tensorflow/contrib/seq2seq:beam_search_ops_op_lib",
         "//tensorflow/contrib/tensor_forest:model_ops_op_lib",
```

akesandgren:
That gets me further but it just fails the same way on another file.

akesandgren:
It gets one line in .tf_configure.bazelrc wrong.
Our top level path for EB is /hpc2n/eb which is a symlink pointing into /pfs/software/eb/<arch>
the .tf_configure.bazelrc has this:

```
build --action_env CUDNN_INSTALL_PATH="/pfs/software/eb/amd64_ubuntu1604_bdw/software/MPI/GCC-CUDA/5.4.0-2.26-8.0.61_375.26/OpenMPI/2.1.1/cuDNN/6.0.21"
```

Also in all of the CROSSTOOL files i can see:
```
toolchain {
...
  cxx_builtin_include_directory: "/pfs/software/eb/amd64_ubuntu1604_bdw/software/Core/GCCcore/5.4.0/lib/gcc/x86_64-unknown-linux-gnu/5.4.0/include"
  cxx_builtin_include_directory: "/pfs/software/eb/amd64_ubuntu1604_bdw/software/Core/GCCcore/5.4.0/lib/gcc/x86_64-unknown-linux-gnu/5.4.0/include-fixed"
  cxx_builtin_include_directory: "/pfs/software/eb/amd64_ubuntu1604_bdw/software/Core/GCCcore/5.4.0/include/c++/5.4.0"
  cxx_builtin_include_directory: "/pfs/software/eb/amd64_ubuntu1604_bdw/software/Compiler/GCC/5.4.0-2.26/CUDA/8.0.61_375.26/include"
...
```

boegel:
@akesandgren So what's wrong there exactly? It has resolved the symlink (on purpose), are you saying it shouldn't?

easybuild/easyblocks/t/tensorflow.py
boegel: We can do that later imho, as an enhancement (and I think I tried that initially, but it was less trivial than you would assume).

I don't consider this a blocker since a lot of other crap is being downloaded during the TF installation anyway, and it's not going to be easy to fix...

akesandgren:
@JackPerdue Found the problem with the GPU build (well at least the one I'm doing).

We just need to decide on the best way to add the necessary patching.

boegel:
@akesandgren So when are you going to share the modifications you had to make to the TF easyblock? ;)

akesandgren:
When i know that TF 1.5.0 can be built. Perhaps sometime tomorrow during travel breaks.

akesandgren:
Added a PR https://github.com/boegel/easybuild-easyblocks/pull/19 to this PR.
That fixes GPU builds and problems with Intel compiler on Ubuntu.

JackPerdue:
I still don't understand why y'all are being such cruel draconians by insisting on an EB binutil module..... what's wrong with this:

``142c142
<             raise EasyBuildError("Failed to determine installation prefix for binutils")
---
>             binutils_bin = '/usr/bin'
``

If you don't like that, try something like dirname `which ld` or something..... but raising an error is unnecessary and needlessly cruel.

easybuild/easyblocks/t/tensorflow.py
boegel: They are already, `setvar` will spit out an `export VAR=value` line under `-x`, no need to do anything special.

boegel:
@damianam @akesandgren I gave this another, hopefully final, review, I think this is good to go in now...
If the both of you agree (please take a couple of minutes to give the whole thing another view), this is good to merge by one of the @easybuilders/easybuild-easyblocks-maintainers, to be included with the upcoming EasyBuild v3.5.2 (ETA end of Feb'18).

easybuild/easyblocks/t/tensorflow.py
```diff
@@ -0,0 +1,308 @@
...
+
+class EB_TensorFlow(PythonPackage):
+    """Support for building/installing TensorFlow."""
+
+    @staticmethod
+    def extra_options():
+        extra_vars = {
+            # see https://developer.nvidia.com/cuda-gpus
+            'cuda_compute_capabilities': [[], "List of CUDA compute capabilities to build with", CUSTOM],
+            'with_jemalloc': [True, "Make TensorFlow use jemalloc", CUSTOM],
```
akesandgren: Wouldn't it be better to use jemalloc only when it is present as a dependency and remove this with_jemalloc thingie?

boegel:
@JackPerdue I guess we can relax the `binutils` requirement, sure, it's a bit strict right now... It's just that I was making the assumption here that a 'modern' EasyBuild toolchain is used, where we include our own `binutils`.
Is this a blocker for you?

easybuild/easyblocks/t/tensorflow.py
boegel: Currently we have no way to let Bazel pick up on self-provided dependencies, but we needed a way to disable it pulling in and building `jemalloc` itself (which is required on CentOS 6).

This is something to improve on in the future, but I think we'll have to wait for improvements in Bazel itself before this is feasible without heavy patching, it's a huge PITA right now to make that work.

This easyblock is a starting point, definitely not the end of the road...

akesandgren: Ok, good enough for me then.
Awaiting merge so i can redo the builds...

akesandgren:
Looks good enough to go for me.

boegel:
(re)tested with:

* TensorFlow 1.4.0 with `foss/2017b` (see #5318) and `intel/2017b` (see #5499)
* TensorFlow 1.4.1 with `foss/2017b` (see #5814)
* TensorFlow 1.5.0 with `foss/2017b` (see #5819)

pending: TensorFlow 1.5.0 with `goolfc/2017b` (by @akesandgren, see #5769)

JackPerdue:
@boegel no blocker here....digging further I see I have similar problems with Bazel.   So, for now, I've hacked together a binutils-system using bundle.py which only does:  

modextravars = {'EBROOTBINUTILS': '/usr'}

Seems  to work since it gets listed last in module.  That causes bazel.py to do a bunch of other stufff like testing for dwp who's absence on RHEL6 didn't show up until building TensorFlow.  Unfortunately, that hasn't solved the other problem with Bazel in that it wants to use the system GCC instead of the EB one.  Running "strings" on bazel still shows:

GCC: (GNU) 4.8.5 20150623 (Red Hat 4.8.5-16)
GCC: (GNU) 6.3.0

and 4.8.5 is what gets used by Bazel when building TF which breaks it (e.g. trying to use /usr/lib/gcc/x86_64-redhat-linux/4.8.5/include/stdint.h).     I tried rebuilding Bazel with all gcc RPMs removed but somehow it still picks it up.  So to build TF, I still have to remove system gcc (which would not be my first choice).  [note, still use Bazel block and config from EB 3.5.1]

Anyway.... not a blocker for me... lets get something out there and see who else has problems/fixes.

boegel:
@wpoely86 Do you see any reason not to merge this?

wpoely86:
No, I forgot to click the shiny green button :wink:

boegel:
🎉

PR #1288, opened 2017-11-10
Make the scalapack easyblock capable of building in parallel.

akesandgren:
Just make the library first in parallel, delete it and restart the build
in serial.

### DIFF
--- a/easybuild/easyblocks/s/scalapack.py
+++ b/easybuild/easyblocks/s/scalapack.py
@@ -47,7 +47,7 @@ from easybuild.toolchains.linalg.lapack import Lapack
 from easybuild.toolchains.linalg.openblas import OpenBLAS
 from easybuild.toolchains.linalg.intelmkl import IntelMKL
 from easybuild.tools.build_log import EasyBuildError
-from easybuild.tools.filetools import copy_file
+from easybuild.tools.filetools import copy_file, remove_file
 from easybuild.tools.modules import get_software_root
 
 
@@ -190,8 +190,21 @@ class EB_ScaLAPACK(ConfigureMake):
             extra_makeopts.append('CDEFS="-D%s"' % interface)
 
         # update make opts, and build_step
+        saved_buildopts = self.cfg['buildopts']
+
+        # Only build the library first, that can be done in parallel
+        # But the library itself will be broken
+        self.cfg.update('buildopts', 'lib')
         self.cfg.update('buildopts', ' '.join(extra_makeopts))
 
+        super(EB_ScaLAPACK, self).build_step()
+
+        # Now remake libscalapack.a serially and the tests.
+        self.cfg['buildopts'] = saved_buildopts
+        self.cfg.update('buildopts', ' '.join(extra_makeopts))
+
+        remove_file('libscalapack.a')
+        self.cfg['parallel'] = 1
         super(EB_ScaLAPACK, self).build_step()
 
     def install_step(self):
--- a/easybuild/easyblocks/s/scalapack.py
+++ b/easybuild/easyblocks/s/scalapack.py
@@ -49,6 +49,7 @@ from easybuild.toolchains.linalg.intelmkl import IntelMKL
 from easybuild.tools.build_log import EasyBuildError
 from easybuild.tools.filetools import copy_file, remove_file
 from easybuild.tools.modules import get_software_root
+from easybuild.tools.run import run_cmd
 
 
 class EB_ScaLAPACK(ConfigureMake):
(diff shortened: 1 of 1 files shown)


boegel:
@akesandgren Can you clarify why you need to remove the library after building it in parallel? Your comments mentions it's broken after building in parallel, but I'm not sure what you mean there.

Also, it this really worth the trouble, does it significantly speed up the build process?

akesandgren:
When building in parallel there is no guard on the actual $(AR) libscalapack.a xxx.o commands in the various sub parts of scvalapack, i.e., blacs, pblas, scalapack. So libscalapack.a gets garbled.
So, removing it and redoing the make command serially makes it create libscalapack.a correctly. The executables can then be linked to a correctly built libscalapack.

And yes, it reduces build time with up to 80% on my dual-core laptop, and when using intel compiler the reduced build time is significant.

boegel:
Tested with one of the existing `ScaLAPACK` 2.0.2 easyconfigs, works fine.

The speedup of the build process is quite significant too; with `ScaLAPACK-2.0.2-gompi-2017b-OpenBLAS-0.2.20.eb` I'm seeing the total required time to install it go down from ~6m30s to ~3m20s on a 16-core Intel Sandy Bridge system.

Requires changes to the ScaLAPACK easyconfigs (i.e. dropping the hardcoded `parallel = 1`) is done in https://github.com/easybuilders/easybuild-easyconfigs/pull/5331.

Thanks a lot for the effort on this @akesandgren, I'm sure figuring this out wasn't as easy as it looks. :)

akesandgren:
Oh but it was easy, I've known this for years :-)

easybuild/easyblocks/s/scalapack.py
```diff
@@ -190,8 +190,21 @@ def build_step(self):
...
 
         # update make opts, and build_step
+        saved_buildopts = self.cfg['buildopts']
+
+        # Only build the library first, that can be done in parallel
+        # But the library itself will be broken
+        self.cfg.update('buildopts', 'lib')
+        self.cfg.update('buildopts', ' '.join(extra_makeopts))
+
+        super(EB_ScaLAPACK, self).build_step()
```
boegel: @akesandgren Every now and then, a `ScaLAPACK` build fails here with `ranlib: ../../../libscalapack.a: Malformed archive`, see test report in https://github.com/easybuilders/easybuild-easyconfigs/pull/5331.

To remedy this, we should run the parallel `make` such that we just ignore the exit code, rather than just calling out to `build_step` of the parent here?

akesandgren: Ahh, I was looking at the failed reports but couldn't see any reason for the fail.
Ignoring the exit code might be bad in case some compile actually fails (in the future).
But one could patch the makefiles to have the ar/ranlib be ignored. And if so i should probably be done in the easyblock.

Thoughts on that?

boegel: Well, even if the parallel `make` fails for a different reason, that's OK, since we'll run a `make -j 1` anyway (in which case it'll fail again)?

akesandgren: Yes, that's true. So we could indeed ignore the first parallel make. How?

boegel: rather than doing the `super` call, you'll need to run `make` yourself, as follows:

```python
# deliberately ignore exit code using log_ok=False
# creating libscalapack.a may fail in parallel, but should work fine with non-parallel make afterwards
run_cmd("make -j %s" % self.cfg['parallel'], log_ok=True)
```

Note: this ignores `prebuidopts` & `buildopts`, if you want to be able to control the `make` command via the easyconfig, you should make this a bit smarter, see the `ConfigureMake.build_step` implementation.

Strictly speaking, not doing so would be a regression...

akesandgren: :-( I was hoping there would be a ignore_fails flag to that build step ...
Oh well, I'll see what i can come up with...

boegel: Well, you can enhance the existing `ConfigureMake.build_step` implementation to allow ignoring failures of the `make` command it runs, optionally of course.

It's a pretty specific use case though, you usually really want to check the exit code. :)

akesandgren:
@boegel Changes done. I hope i got it right. Your comment above on how to do it is self-contradictory.

boegel:
looks good, retested with all existing ScaLAPACK easyconfigs with the `parallel = 1` removed (see https://github.com/easybuilders/easybuild-easyconfigs/pull/5331), works great, so going in.

Thanks for the effort @akesandgren!

Issue #202, opened 2013-05-23
R produced with any ICTCE toolchain doesn't use multi-threaded BLAS


pforai:
As discussed on IRC today, all R builds produced with any Intel based toolchain (hence MKL) don't use the underlying BLAS library's thread "wayness" regardless of the value set in OMP_NUM_THREADS and friends.

This is different to the behaviour that can be seen on GOOLF builds, here a few benchmarks (kindly provided by @boegel) using the following code 

```
m=matrix(rnorm(4e6),2000)
system.time(m %*% m)
```

Output from a GOOLF build:

```
user  system elapsed
 2.300  0.100  0.346
```

This shows that clearly the underlying BLAS is performing multithread since the elapsed wall time is much lower than the user time.

Output from an ICTCE build:

```
user  system elapsed
1.780  0.010  1.822
```

In this case the elapsed wall time is higher then user time and shows that nothing is being done in parallel  (as is the case with GOOLF and probably other GNU toolchains and BLAS implementations).

At this point in time it is not exactly clear what is causing this to happen and if it can be altered at run time or if this is a consequence of compile time setup.

boegel:
I ran into this old issue while looking into an update of R (cfr. https://github.com/easybuilders/easybuild-easyconfigs/pull/5360).

The fix for this is pretty trivial, just use `BLAS_LIBS="$LIBBLAS_MT" LAPACK_LIBS="$LIBLAPACK_MT"` rather than `BLAS_LIBS="$LIBBLAS" LAPACK_LIBS="$LIBLAPACK"` in `prebuildopts` in the easyconfig file.

Currently testing this.

Are users going to be surprised by this, e.g. when they're already using multiple cores in parallel some other way?

verdurin:
Looks like a sensible improvement to me.

Maybe include a `modloadmsg` if you're concerned about the impact of the change?

boegel:
@verdurin I'm not sure we can explain this via a one-liner `modloadmsg` (and anything longer will not get read imho).

Maybe we should make it optional to link with a multithreaded MKL, and then use `-mt` as versionsuffix?

verdurin:
@boegel - yes, a different configuration with a suffix would indeed make it clearer/more explicit.

boegel:
I suggest we follow up on this in https://github.com/easybuilders/easybuild-easyblocks/pull/1292/files#r152269244

boegel:
[...]

PR #1292, opened 2017-11-21
correctly configure R for BLAS/LAPACK & check configure output

boegel:
R is currently not being configured correctly, see https://github.com/easybuilders/easybuild-easyconfigs/issues/1435 and also https://github.com/easybuilders/easybuild-easyconfigs/issues/4773.

If `LAPACK_LIBS` is defined in `preconfigopts`, R reports that LAPACK is being built `generic`, which seems wrong (although it *may* be OK for LAPACK, since there are no tuned LAPACK libraries, all the tuning is done in the BLAS library afaik).

This change only defines `$BLAS_LIBS` in the environment, and checks the output of the configure command to make sure that BLAS/LAPACK is i) used, ii) correctly configured (i.e. no `generic`).

The R easyconfigs (in particular the recent https://github.com/easybuilders/easybuild-easyconfigs/pull/5090 and https://github.com/easybuilders/easybuild-easyconfigs/pull/5360)  will need to be modified as well, since they now hard set `LAPACK_LIBS` via `preconfigopts`, which leads to `LAPACK(generic)` being reported.

That's a bit of a catch-22 situation though, since removing `preconfigpts` part and not including `--with-blas --with-lapack` in the easyconfig files requires that this updated easyblock is used, otherwise R gets (sliently) built without BLAS/LAPACK...

### DIFF
--- a/easybuild/easyblocks/r/r.py
+++ b/easybuild/easyblocks/r/r.py
@@ -26,12 +26,15 @@
 EasyBuild support for building and installing R, implemented as an easyblock
 
 @author: Jens Timmerman (Ghent University)
+@author: Kenneth Hoste (Ghent University)
 """
 import os
+import re
 from distutils.version import LooseVersion
 
+import easybuild.tools.environment as env
 from easybuild.easyblocks.generic.configuremake import ConfigureMake
-from easybuild.tools import environment
+from easybuild.tools.build_log import print_warning
 from easybuild.tools.modules import get_software_root
 from easybuild.tools.systemtools import get_shared_lib_ext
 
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/r/r.py
```diff
@@ -55,19 +58,39 @@ def prepare_for_extensions(self):
...
+        env.setvar('BLAS_LIBS', os.getenv('LIBBLAS_MT'))
+        self.cfg.update('configopts', "--with-blas --with-lapack")
 
         # make sure correct config script is used for Tcl/Tk
         for dep in ['Tcl', 'Tk']:
             root = get_software_root(dep)
             if root:
                 dep_config = os.path.join(root, 'lib', '%sConfig.sh' % dep.lower())
-                self.cfg.update('configopts', '-with-%s-config=%s' % (dep.lower(), dep_config))
+                self.cfg.update('configopts', '--with-%s-config=%s' % (dep.lower(), dep_config))
```
boegel: Nasty typo here, this went undetected because the existing R easyconfigs already include `--with-tcl-config` and `--with-tk-config` in `configopts`...

easybuild/easyblocks/r/r.py
```diff
@@ -55,19 +58,39 @@ def prepare_for_extensions(self):
...
     def configure_step(self):
-        """Configuration step, we set FC, F77 is already set by EasyBuild to the right compiler,
-        FC is used for Fortan90"""
-        environment.setvar('FC', self.toolchain.get_variable('F90'))
+        """Custom configuration for R."""
+
+        # define $BLAS_LIBS to build R correctly against BLAS/LAPACK library
+        # $LAPACK_LIBS should *not* be specified since that may lead to using generic LAPACK
+        # see https://github.com/easybuilders/easybuild-easyconfigs/issues/1435
+        env.setvar('BLAS_LIBS', os.getenv('LIBBLAS_MT'))
```
boegel: I am using `$LIBBLAS_MT` here, while existing R easyconfigs use `$LIBBLAS`.

This matter mainly when using Intel MKL, i.e. multithreaded vs sequential, see also the discussion in #202.

We could/should make this configurable, and stick to linking with the sequential BLAS library by default?

PR #1294, opened 2017-11-21
add support for custom easyconfig parameter in R easyblock to build on top of multi-threaded BLAS/LAPACK

boegel:
follow-up for https://github.com/easybuilders/easybuild-easyblocks/pull/1292/files#r152269244


boegel:
After discussing this with @migueldiascosta, the idea now is to have a general mechanism for this via the EasyBuild framework, without having to tweak easyblocks, see discussion in https://github.com/easybuilders/easybuild-framework/issues/2350

PR #1296, opened 2017-11-22
add custom easyblock for COMSOL

Micket:
Automated installs of COMSOL requires writing a config file.
The license information is required, and exposed via environment variable EB_COMSOL_LICENSE (as it can be either file, port@server, key).
Coupled with easybuilders/easybuild-easyconfigs#5390

### DIFF
--- a/easybuild/easyblocks/c/comsol.py
+++ b/easybuild/easyblocks/c/comsol.py
@@ -1,0 +1,78 @@
+##
+# This file is an EasyBuild reciPY as per https://github.com/easybuilders/easybuild
+##
+"""
+EasyBuild support for installing COMSOL, implemented as an easyblock
+
+@author: Mikael OEhman (Chalmers University of Technology)
+"""
+
+import re
+import shutil
+import os
+import stat
+
+from easybuild.easyblocks.generic.packedbinary import PackedBinary
+from easybuild.framework.easyconfig import CUSTOM
+from easybuild.tools.build_log import EasyBuildError
+from easybuild.tools.filetools import read_file, write_file
+from easybuild.tools.run import run_cmd
+
+
+class EB_COMSOL(PackedBinary):
+    """Support for installing COMSOL."""
+
+    def __init__(self, *args, **kwargs):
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/c/comsol.py
```diff
@@ -0,0 +1,78 @@
...
+    """Support for installing COMSOL."""
+
+    def __init__(self, *args, **kwargs):
+        super(EB_COMSOL, self).__init__(*args, **kwargs)
+        self.configfile = os.path.join(self.builddir, 'my_setupconfig.ini')
+
+    def configure_step(self):
+        license = os.getenv('EB_COMSOL_LICENSE')
+        if license is None:
+            raise EasyBuildError("EB_COMSOL_LICENSE not set (required)")
```
boegel: I like the approach in #1317 better, since that leverages the available `find_flexlm_license` function

easybuild/easyblocks/c/comsol.py
```diff
@@ -0,0 +1,78 @@
...
+        except IOError, err:
+            raise EasyBuildError("Failed to create installation config file %s: %s", self.configfile, err)
+
+    def install_step(self):
+        run_cmd("./setup -s %s" % self.configfile, log_all=True, simple=True)
+
+    def make_module_extra(self):
+        txt = super(EB_COMSOL, self).make_module_extra()
+        txt += self.module_generator.prepend_paths("PATH", ['bin'])
+        return txt
```
boegel: @Micket This should be done automatically already, why did you have to add this?

boegel:
@Micket #1317 looks like a revised version of this, let's go forward with that?

easybuild/easyblocks/c/comsol.py
Micket: @boegel but it would also limit the license to just be flex licenses, which is not the only option for COMSOL, e.g. passcode, but I suppose that's basically unheard of in HPC environments.

Micket:
@boegel Sure.

PR #1299, opened 2017-11-27
Allow tuning of build command in PythonPackage

wpoely86:
Most often, you need to run `setup.py build` but some software needs
`setup.py build_ext`. We now have an option to do this. The default is
still `build`

### DIFF
--- a/easybuild/easyblocks/generic/pythonpackage.py
+++ b/easybuild/easyblocks/generic/pythonpackage.py
@@ -403,7 +403,7 @@ class PythonPackage(ExtensionEasyBlock):
     def build_step(self):
         """Build Python package using setup.py"""
         if self.use_setup_py:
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/generic/pythonpackage.py
```diff
@@ -402,7 +403,7 @@ def configure_step(self):
     def build_step(self):
         """Build Python package using setup.py"""
         if self.use_setup_py:
-            cmd = "%s %s setup.py build %s" % (self.cfg['prebuildopts'], self.python_cmd, self.cfg['buildopts'])
+            cmd = "%s %s setup.py %s %s" % (self.cfg['prebuildopts'], self.python_cmd, self.cfg['buildcmd'], self.cfg['buildopts'])
```
boegel: line too long now...

I'd suggest this:

```python
cmd = ' '.join([self.['prebuildopts'], self.python_cmd, 'setup.py', self.cfg['buildcmd'], self.cfg['buildopts']])

PR #1302, opened 2017-12-04
set default java encoding to utf8 for Trinity

smoors:
This is needed to avoid compilation failure due to non-standard characters in java comments (see the Butterfly module).

### DIFF
--- a/easybuild/easyblocks/t/trinity.py
+++ b/easybuild/easyblocks/t/trinity.py
@@ -73,7 +73,7 @@ class EB_Trinity(EasyBlock):
 
         self.log.info("Begin Butterfly")
 
-        setvar("-Dfile.encoding", "UTF8")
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/t/trinity.py
```diff
@@ -72,6 +73,8 @@ def butterfly(self):
 
         self.log.info("Begin Butterfly")
 
+        setvar("-Dfile.encoding", "UTF8")
```
boegel: This looks wrong to me...

`setvar` is for defining environment variables, you are defining `$-Dfile.encoding` to `UTF8` here, which is not what you want to do I'm guessing?

smoors: oops..

boegel:
Tested with existing `Trinity` easyconfigs, lgtm, so going in, thanks @smoors!

PR #1304, opened 2017-12-07
add easyblock for installing Octave + generic easyblock for installing Octave packages as extensions

boegel:
fixes https://github.com/easybuilders/easybuild-easyconfigs/issues/4108

This is based on the work provided by @lekshmideepu during the EasyBuild User Meeting in Jülich (Feb'17).

### DIFF
--- a/easybuild/easyblocks/o/octave.py
+++ b/easybuild/easyblocks/o/octave.py
@@ -1,0 +1,88 @@
+##
+# Copyright 2009-2017 Ghent University
+#
+# This file is part of EasyBuild,
+# originally created by the HPC team of Ghent University (http://ugent.be/hpc/en),
+# with support of Ghent University (http://ugent.be/hpc),
+# the Flemish Supercomputer Centre (VSC) (https://www.vscentrum.be),
+# Flemish Research Foundation (FWO) (http://www.fwo.be/en)
+# and the Department of Economy, Science and Innovation (EWI) (http://www.ewi-vlaanderen.be/en).
+#
+# https://github.com/easybuilders/easybuild
+#
+# EasyBuild is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation v2.
+#
+# EasyBuild is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with EasyBuild.  If not, see <http://www.gnu.org/licenses/>.
+##
+"""
+EasyBuild support for building and installing Octave, implemented as an easyblock
+
+@author: Lekshmi Deepu (Juelich Supercomputing Centre)
+@author: Kenneth Hoste (Ghent University)
+"""
+import os
+
+import easybuild.tools.environment as env
+from easybuild.easyblocks.generic.configuremake import ConfigureMake
+from easybuild.tools.build_log import EasyBuildError
+from easybuild.tools.modules import get_software_root
+from easybuild.tools.run import run_cmd
+
+
+EXTS_FILTER_OCTAVE_PACKAGES = ("octave --eval 'pkg list' | grep %(ext_name)s", '')
+
+
+class EB_Octave(ConfigureMake):
+    """Support for building/installing Octave."""
+
+    def configure_step(self):
+        """Custom configuration procedure for Octave."""
+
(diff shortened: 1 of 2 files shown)


easybuild/easyblocks/o/octave.py
```diff
@@ -0,0 +1,88 @@
...
+
+        qt_root = get_software_root('Qt5') or get_software_root('Qt')
+        if qt_root:
+            qt_vars = {
+                'LRELEASE': os.path.join(qt_root, 'bin', 'lrelease'),
+                'MOC': os.path.join(qt_root, 'bin', 'moc'),
+                'RCC': os.path.join(qt_root, 'bin', 'rcc'),
+                'UIC': os.path.join(qt_root, 'bin', 'uic'),
+            }
+            for key, val in sorted(qt_vars.items()):
```
damianam: Mmmh why `sorted`?

easybuild/easyblocks/o/octave.py
```diff
@@ -0,0 +1,88 @@
...
+        super(EB_Octave, self).prepare_for_extensions()
+
+    def sanity_check_step(self):
+        """Custom sanity check for Octave."""
+        custom_paths = {
+            'files': ['bin/octave'],
+            'dirs': [],
+        }
+        if self.cfg['exts_list']:
+            custom_paths['dirs'].extend(['share/octave/packages', 'share/octave/packages-arch-dep'])
```
damianam: Use `os.path.join` instead of a hardcoded string?

easybuild/easyblocks/o/octave.py
```diff
@@ -0,0 +1,88 @@
...
+
+        custom_commands = ["octave --eval '1+2'"]
+
+        super(EB_Octave, self).sanity_check_step(custom_paths=custom_paths, custom_commands=custom_commands)
+
+    def make_module_req_guess(self):
+        """Custom extra paths/variables to define in generated module for Octave."""
+        guesses = super(EB_Octave, self).make_module_req_guess()
+        guesses.update({'OCTAVE_LD_LIBRARY_PATH': ['lib']})
```
damianam: :-? What is `OCTAVE_LD_LIBRARY_PATH`? I haven't  seen any reference to it anywhere else on internet.

easybuild/easyblocks/o/octave.py
```diff
@@ -0,0 +1,88 @@
...
+EXTS_FILTER_OCTAVE_PACKAGES = ("octave --eval 'pkg list' | grep %(ext_name)s", '')
+
+
+class EB_Octave(ConfigureMake):
+    """Support for building/installing Octave."""
+
+    def configure_step(self):
+        """Custom configuration procedure for Octave."""
+
+        self.cfg.update('configopts', '--with-blas="$LIBBLAS" --with-lapack="$LIBLAPACK"')
```
damianam: Mmmmh, I think we should check first if `$LIBBLAS` and `$LIBLAPACK` are defined, to don't pass inadvertently `--with-blas= --with-lapack=` to `configopts`, in case someone messes up and tries to install it in a toolchain without BLAS and LAPACK

easybuild/easyblocks/o/octave.py
boegel: Easier to debug in case it's needed, no technical reason.

easybuild/easyblocks/o/octave.py
boegel: OK, will do.

easybuild/easyblocks/o/octave.py
```diff
@@ -0,0 +1,88 @@
...
+
+        qt_root = get_software_root('Qt5') or get_software_root('Qt')
+        if qt_root:
```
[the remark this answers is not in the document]
boegel: OK, I'll include a debug log message.

easybuild/easyblocks/o/octave.py
boegel: will do

easybuild/easyblocks/o/octave.py
boegel: This was there in Lekshmi's version, but now I realise why: she used the `EB_OCaml` easyblock as a base, where `CAML_LD_LIBRARY_PATH` is defined (it should be for `OCaml`).

Doesn't make sense for `Octave` though, you're right, thanks for bringing this up!

damianam:
lgtm. Going in @boegel, thanks!

easybuild/easyblocks/o/octave.py
lekshmideepu: boegel: yes, right. I had made a note about this variable, but forgot to update in the code.

lekshmideepu:
@boegel Great! :+1:

PR #1305, opened 2017-12-08
Enhance amber.py with support for OpenBLAS and better Intel MPI support + fix support for only installing AmberTools

akesandgren:
Amber differs between -mpi and -intelmpi

Also make NAB openmp capable.

### DIFF
--- a/easybuild/easyblocks/a/amber.py
+++ b/easybuild/easyblocks/a/amber.py
@@ -121,13 +121,20 @@ class EB_Amber(ConfigureMake):
 
         # define environment variables for MPI, BLAS/LAPACK & dependencies
         mklroot = get_software_root('imkl')
+        openblasroot = get_software_root('OpenBLAS')
         if mklroot:
             env.setvar('MKL_HOME', mklroot)
+        elif openblasroot:
+            env.setvar('GOTO', os.environ['LIBLAPACK'])
 
         mpiroot = get_software_root(self.toolchain.MPI_MODULE_NAME[0])
         if mpiroot and self.toolchain.options.get('usempi', None):
             env.setvar('MPI_HOME', mpiroot)
             self.with_mpi = True
+            if self.toolchain.mpi_family() == toolchain.INTELMPI:
+                self.mpi_option = '-intelmpi'
+            else:
+                self.mpi_option = '-mpi'
 
         common_configopts = [self.cfg['configopts'], '--no-updates', '-static', '-noX11']
 
@@ -157,11 +164,15 @@ class EB_Amber(ConfigureMake):
         else:
             raise EasyBuildError("Don't know how to compile with compiler family '%s' -- check EasyBlock?", comp_fam)
 
+        # The NAB compiles need openmp flag
+        if self.toolchain.options.get('openmp', None):
+            env.setvar('CUSTOMBUILDFLAGS', self.toolchain.get_flag('openmp'))
+
         # compose list of build targets
         build_targets = [('', 'test')]
 
         if self.with_mpi:
-            build_targets.append(('-mpi', 'test.parallel'))
+            build_targets.append(('%s' % self.mpi_option, 'test.parallel'))
             # hardcode to 4 MPI processes, minimal required to run all tests
             env.setvar('DO_PARALLEL', 'mpirun -np 4')
 
@@ -171,7 +182,7 @@ class EB_Amber(ConfigureMake):
             self.with_cuda = True
             build_targets.append(('-cuda', 'test.cuda'))
             if self.with_mpi:
-                build_targets.append(("-cuda -mpi", 'test.cuda_parallel'))
+                build_targets.append(("-cuda %s" % self.mpi_option, 'test.cuda_parallel'))
 
         ld_lib_path = os.environ.get('LD_LIBRARY_PATH', '')
         env.setvar('LD_LIBRARY_PATH', os.pathsep.join([os.path.join(self.installdir, 'lib'), ld_lib_path]))
--- a/easybuild/easyblocks/a/amber.py
+++ b/easybuild/easyblocks/a/amber.py
@@ -125,7 +125,11 @@ class EB_Amber(ConfigureMake):
         if mklroot:
             env.setvar('MKL_HOME', mklroot)
         elif openblasroot:
-            env.setvar('GOTO', os.environ['LIBLAPACK'])
+	    lapack = os.getenv('LIBLAPACK', 'LIBLAPACK_NOT_SET')
+	    if lapack == 'LIBLAPACK_NOT_SET':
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/a/amber.py
```diff
@@ -157,11 +164,15 @@ def install_step(self):
...
+        # The NAB compiles need openmp flag
+        if self.toolchain.options.get('openmp', None):
+            env.setvar('CUSTOMBUILDFLAGS', self.toolchain.get_flag('openmp'))
+
         # compose list of build targets
         build_targets = [('', 'test')]
 
         if self.with_mpi:
-            build_targets.append(('-mpi', 'test.parallel'))
+            build_targets.append(('%s' % self.mpi_option, 'test.parallel'))
```
boegel: no need for the `'%s' %` bit, just use `self.mpi_option`:

```python
build_targets.append((self.mpi_option, 'test.parallel'))

File: easybuild/easyblocks/a/amber.py [hunk 2]
```diff
@@ -157,11 +164,15 @@ def install_step(self):
         else:
             raise EasyBuildError("Don't know how to compile with compiler family '%s' -- check EasyBlock?", comp_fam)
 
+        # The NAB compiles need openmp flag
+        if self.toolchain.options.get('openmp', None):
+            env.setvar('CUSTOMBUILDFLAGS', self.toolchain.get_flag('openmp'))
```
boegel: Hmm, shouldn't we pass the full `$CFLAGS` here? If OpenMP is enabled via `toolchainopts`, it will include `-fopenmp`.

easybuild/easyblocks/a/amber.py
```diff
@@ -121,13 +121,20 @@ def install_step(self):
 
         # define environment variables for MPI, BLAS/LAPACK & dependencies
         mklroot = get_software_root('imkl')
+        openblasroot = get_software_root('OpenBLAS')
         if mklroot:
             env.setvar('MKL_HOME', mklroot)
+        elif openblasroot:
+            env.setvar('GOTO', os.environ['LIBLAPACK'])
```
boegel: `os.getenv('LIBLAPACK', 'LIBLAPACK_NOT_SET')` may make more sense

In theory, you could list `OpenBLAS` as a dependency and build with an MPI toolchain like `gompi`, which would make the current code crash hard because `$LIBLAPACK` is not defined.

boegel: Also, should we complain in an `else` that no LAPACK library was found?

Building without it is probably going to end in tears performance-wise?

akesandgren: Can't do "-lopenblas -lgfortran" (which LIBLAPACK is with foss) since the openblas could have been built by another compiler, like intel or pgi and in that case -lgfortran is wrong.
And doing a list of all the possible combos is just wrong.
But technically speaking, openblas does, or at least used to , not depend on fortran libs, so LIBLAPACK being -lopenblas -lgfortran is wrong i think. And one could then do just -lopenblas here.

So, at the moment, we really have to rely on LIBLAPACK giving us the correct set of libraries. (And that applies to all other easyblocks/easyconfigs really). But complaining if LIBLAPACK isn't set is a good thing to do.

easybuild/easyblocks/a/amber.py
akesandgren: Amber doesn't quite use EB's CFLAGS at all, yet. So at the moment adding just the openmp flag is what should be done for consistency until that is dealt with.

easybuild/easyblocks/a/amber.py
boegel: @akesandgren Wait, `os.getenv('LIBLAPACK')` and `os.environ['LIBLAPACK']` are both going to give you the value for `$LIBLAPACK`...

Or are you just pointing out an issue in case `OpenBLAS` is combined with non-GCC compilers?

If so, please open an issue in the framework repository to track that.

That's orthogonal to this though?

easybuild/easyblocks/a/amber.py
boegel: Well, isn't this an opportunity to do so? What would be the correct way to let Amber honor the compiler flags put in place by EasyBuild? Wouldn't it be defining `$CUSTOMBUILDFLAGS` equal to `$CFLAGS`?

akesandgren: Fixing that requires more patching of makefiles that i haven't had time to dive into. The amber build is somewhat convoluted... There are places where CUSTOMBUILDFLAGS isn't used for instance.
Getting this right is not trivial.

easybuild/easyblocks/a/amber.py
akesandgren: The issue is OpenBLAS built with non-GCC toolchains and probably that GCC-toolchain should define LIBLAPACK as just -lopenblas in this case, but i have too take a second look at how openblas is being built at the moment. It used to be the case that openblas was fortran-free. This might have changed with later versions and must be verified.

So orthogonal to the matter at hand yes.

easybuild/easyblocks/a/amber.py
```diff
@@ -121,13 +121,24 @@ def install_step(self):
 
         # define environment variables for MPI, BLAS/LAPACK & dependencies
         mklroot = get_software_root('imkl')
+        openblasroot = get_software_root('OpenBLAS')
         if mklroot:
             env.setvar('MKL_HOME', mklroot)
+        elif openblasroot:
+	    lapack = os.getenv('LIBLAPACK', 'LIBLAPACK_NOT_SET')
+	    if lapack == 'LIBLAPACK_NOT_SET':
```
boegel: If you'll check it immediately like this, this should be:

```python
lapack = os.getenv('LIBLAPACK')
if lapack is None:
    raise ...
else:
    ...

File: easybuild/easyblocks/a/amber.py [hunk 5]
```diff
@@ -121,13 +121,24 @@ def install_step(self):
 
         # define environment variables for MPI, BLAS/LAPACK & dependencies
         mklroot = get_software_root('imkl')
+        openblasroot = get_software_root('OpenBLAS')
         if mklroot:
             env.setvar('MKL_HOME', mklroot)
+        elif openblasroot:
+	    lapack = os.getenv('LIBLAPACK')
+	    if lapack is None:
+		raise EasyBuildError("LIBLAPACK (from OpenBLAS) not found in environement. Not building with a lapack capable toolchain?")
```
boegel: @akesandgren This line is too long (> 120 chars). I would just drop the 2nd sentence, since a situation where `OpenBLAS` is there but `$LIBLAPACK` is not defined is quite strange...

boegel:
Tested with `Amber-14-intel-2016a-AmberTools-15-patchlevel-13-13.eb` (the only one I can test) + `AmberTools` (see https://github.com/easybuilders/easybuild-easyconfigs/pull/5632), looks good, so going in. Thanks @akesandgren!

PR #1306, opened 2017-12-08
Make cuda.py install the CUPTI support too.

akesandgren:
Users want the tracing interface installed.

### DIFF
--- a/easybuild/easyblocks/c/cuda.py
+++ b/easybuild/easyblocks/c/cuda.py
@@ -125,18 +125,19 @@ class EB_CUDA(Binary):
         if LooseVersion(self.version) < LooseVersion("6"):
             chk_libdir += ["lib"]
 
-        extra_files = []
-        if LooseVersion(self.version) < LooseVersion('7'):
-            extra_files.append('open64/bin/nvopencc')
-        if LooseVersion(self.version) >= LooseVersion('7'):
-            extra_files.append("extras/CUPTI/lib64/libcupti.%s" % shlib_ext)
-
         custom_paths = {
-            'files': ["bin/%s" % x for x in ["fatbinary", "nvcc", "nvlink", "ptxas"]] + extra_files +
+            'files': ["bin/%s" % x for x in ["fatbinary", "nvcc", "nvlink", "ptxas"]] +
                      ["%s/lib%s.%s" % (x, y, shlib_ext) for x in chk_libdir for y in ["cublas", "cudart", "cufft",
                                                                                       "curand", "cusparse"]],
-            'dirs': ["include", "extras/CUPTI/include"],
+            'dirs': ["include"],
         }
+
+        if LooseVersion(self.version) < LooseVersion('7'):
+            custom_paths['files'].append('open64/bin/nvopencc')
+        if LooseVersion(self.version) >= LooseVersion('7'):
+            custom_paths['files'].append("extras/CUPTI/lib64/libcupti.%s" % shlib_ext)
+            custom_paths['dirs'].append("extras/CUPTI/include")
+
 
         super(EB_CUDA, self).sanity_check_step(custom_paths=custom_paths)
 
--- a/easybuild/easyblocks/c/cuda.py
+++ b/easybuild/easyblocks/c/cuda.py
@@ -128,12 +128,13 @@ class EB_CUDA(Binary):
         extra_files = []
         if LooseVersion(self.version) < LooseVersion('7'):
             extra_files.append('open64/bin/nvopencc')
+        if LooseVersion(self.version) >= LooseVersion('7'):
+            extra_files.append("extras/CUPTI/lib64/libcupti.%s" % shlib_ext)
 
         custom_paths = {
             'files': ["bin/%s" % x for x in ["fatbinary", "nvcc", "nvlink", "ptxas"]] + extra_files +
                      ["%s/lib%s.%s" % (x, y, shlib_ext) for x in chk_libdir for y in ["cublas", "cudart", "cufft",
-                                                                                      "curand", "cusparse"]] +
-                     ["extras/CUPTI/%s/lib%s.%s" % (x, y, shlib_ext) for x in chk_libdir for y in ["cupti"]],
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/c/cuda.py
```diff
@@ -132,8 +132,9 @@ def sanity_check_step(self):
         custom_paths = {
             'files': ["bin/%s" % x for x in ["fatbinary", "nvcc", "nvlink", "ptxas"]] + extra_files +
                      ["%s/lib%s.%s" % (x, y, shlib_ext) for x in chk_libdir for y in ["cublas", "cudart", "cufft",
-                                                                                      "curand", "cusparse"]],
-            'dirs': ["include"],
+                                                                                      "curand", "cusparse"]] +
+                     ["extras/CUPTI/%s/lib%s.%s" % (x, y, shlib_ext) for x in chk_libdir for y in ["cupti"]],
```
boegel: Since there's only one library in the `y` list, you can simplify this to:

```python
["extras/CUPTI/%s/libcupti.%s" % (x, shlib_ext) for x in chk_libdir],

File: easybuild/easyblocks/c/cuda.py [hunk 2]
```diff
@@ -145,8 +146,8 @@ def make_module_req_guess(self):
 
         guesses.update({
             'PATH': ['open64/bin', 'bin'],
-            'LD_LIBRARY_PATH': ['lib64'],
-            'CPATH': ['include'],
+            'LD_LIBRARY_PATH': ['lib64', 'extras/CUPTI/lib64'],
```
boegel: Are you sure it's `lib64`? If so, don't use `chk_libdir` above in the sanity check, but check for `"extras/CUPTI/lib64/libcupti.%s" % shlib_ext` instead.

easybuild/easyblocks/c/cuda.py
boegel: Just took a quick look, this needs to be done conditionally based on the CUDA version, it's only there since CUDA 7.0 it seems, and we still have easyconfigs around for CUDA 5.x and 6x.

easybuild/easyblocks/c/cuda.py
akesandgren: From the original cuda.py
"# Versions higher than 6 do not provide 32 bit libraries"

So chk_libdir uses in the sanity check is technically correct but if cupti doesn't exist prior to v7 then it can be skipped.

boegel: Even in recent version only `extras/CUPTI/lib64` is there, there never is a `extras/CUPTI/lib` in my installations, so it doesn't make sense to use `chk_libdir` for the `extras/CUPTI/lib*` locations?

akesandgren: Yeah, fixed that. But by main question is if the LD_LIBRARY_PATH should be dealt with to have 'lib' if version < 6?

boegel: There is no `CUPTI` directory anywhere in CUDA5.x or 6.x, as far as I can tell.

akesandgren: I was talking about "lib" and "lib64" in LD_LIBRARY_PATH. I.e. should "lib" be added if version < 6?
(Nothing to do with cupti here)

boegel: We could consider including `lib` in case 32-bit builds are being done (we have a toolchain option for that), but including both `lib` and `lib64` is not a good idea imho...

akesandgren: Ok. Skipping that part then.

easybuild/easyblocks/c/cuda.py
```diff
@@ -128,12 +128,14 @@ def sanity_check_step(self):
...
             extra_files.append('open64/bin/nvopencc')
+        if LooseVersion(self.version) >= LooseVersion('7'):
+            extra_files.append("extras/CUPTI/lib64/libcupti.%s" % shlib_ext)
 
         custom_paths = {
             'files': ["bin/%s" % x for x in ["fatbinary", "nvcc", "nvlink", "ptxas"]] + extra_files +
                      ["%s/lib%s.%s" % (x, y, shlib_ext) for x in chk_libdir for y in ["cublas", "cudart", "cufft",
                                                                                       "curand", "cusparse"]],
-            'dirs': ["include"],
+            'dirs': ["include", "extras/CUPTI/include"],
```
boegel: This should also be dependent on the CUDA version, so I'd propose an approach like:

```
custom_paths = {
    'files': ...,
    'dirs': ...,
}

if LooseVersion(self.version) < LooseVersion('7'):
    custom_paths['files'].append('open64/bin/nvopencc')
if LooseVersion(self.version) >= LooseVersion('7'):
    custom_paths['files'].append("extras/CUPTI/lib64/libcupti.%s" % shlib_ext)
    custom_paths['dirs'].append('extras/CUPTI/include')

File: easybuild/easyblocks/c/cuda.py [hunk 4]
```diff
@@ -143,10 +145,22 @@ def make_module_req_guess(self):
 
         guesses = super(EB_CUDA, self).make_module_req_guess()
 
+        # The dirs should be in the order ['open64/bin', 'bin']
+        bin_path = []
+        if LooseVersion(self.version) < LooseVersion('7'):
+            bin_path.append('open64/bin')
+        bin_path.append('bin')
```
boegel: It's OK to include non-existing directories, the existence of the directories will be checked before they're used in the generated module file

akesandgren: So you want me to revert to the original non-version-checking code here? For all of the vars?

boegel: Well, not sure... I guess there's some value in not making EB look for subdirectories that we know are not there.

So let's leave it as is.

You'll need to fix the issue in `sanity_check_step` though.

boegel:
testing with all existing `CUDA` easyconfigs + also #5534, all is well, going in

PR #1307, opened 2017-12-08
Make FDTD_Solutions easyblock do the install with copytree instead of rpm rebuild

akesandgren:
rpm installation is not kosher on Debian based systems.


mboisson:
FWIW, I have used this easyblock successfully to install FDTD Solutions on Compute Canada servers.

easybuild/easyblocks/f/fdtd_solutions.py
```diff
@@ -33,64 +33,45 @@
...
+        pass
 
+    def install_step(self):
+        """Install FDTD Solutions using copy tree."""
         try:
-            os.rename(rpms[0], '%s.bk' % rpms[0])
-            shutil.copy2(rebuilt_rpms[0], rpms[0])
+            # shutil.copytree doesn't allow the target directory to exist already
+            rmtree2(self.installdir)
+            shutil.copytree(os.path.join(self.cfg['start_dir'], 'opt', 'lumerical', 'fdtd'),
```
boegel: @akesandgren This should be updated to use `copy_dir` from `filetools` (no need for the `try`/`except` anymore then)

Also, please define a variable with the path you're copying, since you're using it in two places (here + the error message)

```python
ftdt_dir = os.path.join(self.cfg['start_dir'], 'opt', 'lumerical', 'fdtd')
copy_dir(ftdt_dir, self.installdir, symlinks=self.cfg['keepsymlinks'])

File: easybuild/easyblocks/f/fdtd_solutions.py [hunk 2]
```diff
@@ -33,64 +33,45 @@
...
-        return guesses
+            raise EasyBuildError("Failed to copy %s to %s: %s",
+        os.path.join(self.cfg['start_dir'], 'opt', 'lumerical', 'fdtd'), self.installdir, err)
 
     def sanity_check_step(self):
         """Custom sanity check for FDTD Solutions."""
         custom_paths = {
             'files': [],
-            'dirs': ['opt/lumerical/fdtd/bin', 'opt/lumerical/fdtd/lib'],
+            'dirs': ['bin', 'lib'],
```
boegel: @akesandgren It's probably worth checking for some specific files at least?

easybuild/easyblocks/f/fdtd_solutions.py
```diff
@@ -33,64 +33,39 @@
...
-        """Adjust values for PATH, LD_LIBRARY_PATH, etc."""
-        guesses = super(EB_FDTD_underscore_Solutions, self).make_module_req_guess()
-        guesses.update({
-            'PATH': ['opt/lumerical/fdtd/bin'],
-            'LD_LIBRARY_PATH': ['opt/lumerical/fdtd/lib'],
-        })
-        return guesses
+        """Install FDTD Solutions using copy tree."""
+        fdtd_dir = os.path.join(self.cfg['start_dir'], 'opt', 'lumerical', 'fdtd')
+        copy_dir(fdtd_dir, self.installdir, symlinks=self.cfg['keepsymlinks'])
```
boegel: @akesandgren Installation fails now because the install directory is pre-created, and `copy_dir` doesn't like that.

You'll need to either call `remove_dir(self.installdir)s` first, or override the `make_installdir` method to make it a no-op (just log a message like `"Not pre-creating installation directory %s" % self.installdir` instead)

akesandgren: Something like that then?

boegel: I overlooked that `make_installdir` is also responsible for removing an existing installation directory in case of a forced reinstall, so you should (also) set `self.cfg['dontcreateinstalldir'] = True` before calling the parent:

```python
def make_installdir(self):
    """Override installdir creation"""
    self.log.warning("Not pre-creating installation directory %s" % self.installdir)
    self.cfg['dontcreateinstalldir'] = True
    super(EB_EB_FDTD_underscore_Solutions, self).make_installdir()

File: easybuild/easyblocks/f/fdtd_solutions.py [hunk 4]
```diff
@@ -33,64 +33,45 @@
...
 
     def sanity_check_step(self):
         """Custom sanity check for FDTD Solutions."""
         custom_paths = {
-            'files': [],
-            'dirs': ['opt/lumerical/fdtd/bin', 'opt/lumerical/fdtd/lib'],
+            'files': ['bin/fdtd-solutions', 'lib/fdtd-engine-mpichp4_libFNP.so'],
+            'dirs': ['api'],
```
boegel: @akesandgren There no `api` installed for the existing `FDTD_Solutions` easyconfigs (`8.6.2` and `8.11.337`).

And for `8.6.2`, there's no `lib/fdtd-engine-mpichp4_libFNP.so` either (not even close).

akesandgren: Gah.... I only have 8.16.982 to look at...
Any suggestions?

boegel: How about just checking for `bin/fdtd-solutions` and a non-empty `lib` dir?

That's still a little bit stricter than what was there (which was basically just checking for `bin` & `lib`)...

akesandgren: checking for a non-empty lib? what do you mean? I can check for an existing lib dir like it was done before, but how can i check for a non-empty one without checking for a specific file?

Or did you just mean an existing lib dir?

mboisson: Correct me if I'm wrong, but `'dirs': ['lib']` already fails if the `lib` directory is empty.

akesandgren: It does? Didn't really know that, but if that's the case it's good enough for me :-)

mboisson: In `easyblock.py` : 
```
2140         path_keys_and_check = {
2141             # files must exist and not be a directory
2142             'files': ('file', lambda fp: os.path.exists(fp) and not os.path.isdir(fp)),
2143             # directories must exist and be non-empty
2144             'dirs': ("(non-empty) directory", lambda dp: os.path.isdir(dp) and os.listdir(dp)),
2145         }
```

boegel: Exactly, the `dirs` entries need to exist & be non-empty. Checking for an empty dir is kind of useless. ;)

PR #1308, opened 2017-12-08
Add missing arch lib path to ipp module.

akesandgren:
Also do sanity check on the actual (selected) arch lib path.


boegel:
@akesandgren Sorry for not picking up on this... Can you clarify why you closed this?

akesandgren:
There is another PR that does the same thing that has been merged already.
Well, except that mine used self.arch at some places instead of hardcoding intel64.

boegel:
@akesandgren Is it worth opening a follow-up PR to fix the hardcoding?

akesandgren:
That's my intention but there's no need to stress about it.

PR #1310, opened 2017-12-08
Enhance MPI support in molpro.py.

akesandgren:
Don't touch the LAUNCHER, the default works as is and the old
replacement code broke the code since it changed multiple LAUNCHER
instances.
Also reduce optimization to -O2, intel coredumps on some files.
Finally turn on VERBOSE builds.

### DIFF
--- a/easybuild/easyblocks/m/molpro.py
+++ b/easybuild/easyblocks/m/molpro.py
@@ -32,6 +32,7 @@ import shutil
 import re
 from distutils.version import LooseVersion
 
+import easybuild.tools.toolchain as toolchain
 from easybuild.easyblocks.generic.binary import Binary
 from easybuild.easyblocks.generic.configuremake import ConfigureMake
 from easybuild.framework.easyblock import EasyBlock
@@ -62,7 +63,6 @@ class EB_Molpro(ConfigureMake, Binary):
         super(EB_Molpro, self).__init__(*args, **kwargs)
 
         self.full_prefix = ''  # no None, to make easyblock compatible with --module-only
-        self.orig_launcher = None
 
         self.cleanup_token_symlink = False
         self.license_token = os.path.join(os.path.expanduser('~'), '.molpro', 'token')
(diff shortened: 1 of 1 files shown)


ocaisa:
What is the LAUNCHER default? Unless it is `srun` on JURECA it won't work.

akesandgren:
The default is mpirun which normally works on slurm based systems.

But if that is an actual problem then that part of the code has to be rewritten a lot more carefully anyway.
The code before this PR breaks things.

akesandgren:
@ocaisa Do you have a possibility of testing this on JURECA?

easybuild/easyblocks/m/molpro.py
```diff
@@ -106,6 +106,13 @@ def configure_step(self):
                     self.cfg.update('configopts', "-mpp -mppbase %s" % os.environ['MPI_INC_DIR'])
                 else:
                     raise EasyBuildError("$MPI_INC_DIR not defined")
+                if self.toolchain.mpi_family() == toolchain.OPENMPI:
+                    self.cfg.update('configopts', "-openmpi")
+                elif self.toolchain.mpi_family() == toolchain.INTELMPI:
+                    self.cfg.update('configopts', "-intel-mpi")
+                else:
+                    raise EasyBuildError("Failed to find out which MPI toolchain is in use: %s"
+                                         % self.toolchain.mpi_family())
```
boegel: @akesandgren We have several `Molpro` easyconfig that use `dummy` as toolchain, this change will break the installation of each of those...

akesandgren: Why? It already below a

akesandgren: And moreover, all those taht use dummy toolchain also use precompiled_binaries = True and won't be in that part of the code anyway.

easybuild/easyblocks/m/molpro.py
```diff
@@ -106,6 +106,13 @@ def configure_step(self):
                     self.cfg.update('configopts', "-mpp -mppbase %s" % os.environ['MPI_INC_DIR'])
                 else:
                     raise EasyBuildError("$MPI_INC_DIR not defined")
+                if self.toolchain.mpi_family() == toolchain.OPENMPI:
+                    self.cfg.update('configopts', "-openmpi")
+                elif self.toolchain.mpi_family() == toolchain.INTELMPI:
+                    self.cfg.update('configopts', "-intel-mpi")
```
boegel: This break the existing `Molpro-2012.1.27-intel-2015a.eb` easyconfig:

```
unknown Molpro option: -intel-mpi
```

easybuild/easyblocks/m/molpro.py
boegel: OK, sorry, I overlooked that part...

easybuild/easyblocks/m/molpro.py
akesandgren: Wheren't you going to deprecate intel <= 2016a?? :-)

akesandgren: @boegel Repeating: Do you have 2012.1.27 available? I only have 2012.1.52
I.e., can i have a copy of that source to look at.

akesandgren:
@boegel @ocaisa This could use a retest from you

akesandgren:
@ocaisa still need a re-test from JURECA on this...

akesandgren:
In that case we need someone to review this, @boegel ??

PR #1311, opened 2017-12-08
Pre v3 of OpenFOAM needs extra flag to build with newer g++

akesandgren:
-fno-delete-null-pointer-checks is needed to build older OpenFOAM
versions with newer GCC versions.


boegel:
(close/reopen to trigger HoundCI)

boegel:
tested and approved, going in, thanks @akesandgren

PR #1312, opened 2017-12-08
Update quantumespresso.py for newer versions.

akesandgren:
Main change is not doing the build in the installation dir.
That is something that should be avoided whenever possible...

Adds following support:
 - HDF5
 - Adaptively Compressed Exchange
 - QE >v6


migueldiascosta:
@akesandgren w.r.t. `ELPA`, I have no experience with it so that this with a grain of salt, but for me the current easyblock doesn't actually use it because the define added by configure is erased when dflags is applied in the easyblock (and the same thing happens with the elpa includes and libs)

the only way I could get it to build once the define is included was with the changes below and using `ELPA` 2017.11.001 (2016.05.004 fails with wrong number of arguments to `solve_evp_real_1stage`...)

```diff
@@ -95,11 +95,15 @@
             elpa_v = get_software_version("ELPA")
             if LooseVersion(elpa_v) >= LooseVersion("2016"):
                 if self.toolchain.options.get('openmp', False):
-                    self.cfg.update('configopts', '--with-elpa-include=%s/include' % elpa)
-                    self.cfg.update('configopts', '--with-elpa-lib=%s/lib/libelpa_openmp.a' % elpa)
+                    elpa_lib = '%s/lib/libelpa_openmp.a' % elpa
+                    elpa_include = '%s/include/elpa_openmp-%s' % (elpa, elpa_v)
+                    self.cfg.update('configopts', '--with-elpa-include=%s' % elpa_include)
+                    self.cfg.update('configopts', '--with-elpa-lib=%s' % elpa_lib)
                 else:
-                    self.cfg.update('configopts', '--with-elpa-include=%s/include' % elpa)
-                    self.cfg.update('configopts', '--with-elpa-lib=%s/lib/libelpa.a' % elpa)
+                    elpa_lib = '%s/lib/libelpa.a' % elpa
+                    elpa_include = '%s/include/elpa-%s' % (elpa, elpa_v)
+                    self.cfg.update('configopts', '--with-elpa-include=%s' % elpa_include)
+                    self.cfg.update('configopts', '--with-elpa-lib=%s' % elpa_lib)
             else:
                 EasyBuildError("Quantum ESPRESSO needs ELPA to be version 2016 or later")
 
@@ -160,6 +164,11 @@
             hdf5_lib_repl = '-L%s/lib -lhdf5hl_fortran -lhdf5_hl -lhdf5_fortran -lhdf5 -lsz -lz -ldl -lm' % hdf5
             repls.append(('HDF5_LIB', hdf5_lib_repl, False))
 
+        if elpa:
+            dflags.append(" -D__ELPA_2016")
+            repls.append(('IFLAGS', '-I%s' % os.path.join(elpa_include, 'modules'), True))
+            repls.append(('LIBOBJS', elpa_lib, True))
+
         if self.cfg['with_ace']:
             dflags.append(" -D__EXX_ACE")
```

akesandgren:
I didn't get around to verify ELPA functionality yesterday. It's on my todo list...

akesandgren:
QM-E needs ELPA 2015 or 2016.11.001_pre or newer but not 2017

easybuild/easyblocks/q/quantumespresso.py
```diff
@@ -140,28 +192,30 @@ def configure_step(self):
...
+        # Ake Sandgren, 20180712
         if get_software_root('FoX'):
-            self.log.debug("Found FoX external module, disabling libfox target in Makefile")
-            regex_subs = [(r"(libfox: touch-dummy)\n.*", r"\1\n\techo 'libfox: external module used' #")]
-            apply_regex_substitutions('Makefile', regex_subs)
-            # Make configure and make look in the correct directory
-            regex_subs = [(r"\(TOPDIR\)/FoX", r"(EBROOTFOX)")]
-            apply_regex_substitutions('install/configure', regex_subs)
-            apply_regex_substitutions('make.inc', regex_subs)
+            raise EasyBuildError("Found FoX external module, QuantumESPRESSO must use the version they include with the source.")
```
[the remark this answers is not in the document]
akesandgren: Yes, we should change the easyconfig for all QEs that have FoX deps, and add the drop_LIBS_before_configure_of_fox patch.
I'll take a look at that...

easybuild/easyblocks/q/quantumespresso.py
```diff
@@ -338,16 +438,20 @@ def sanity_check_step(self):
                         "virtual.x"]
             if LooseVersion(self.version) > LooseVersion("5"):
                 upftools.extend(["interpolate.x", "upf2casino.x"])
+            if LooseVersion(self.version) >= LooseVersion("6"):
```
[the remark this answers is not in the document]
akesandgren: Oops, sorry, though I checked older versions for that one but apparently not. Will fix.

easybuild/easyblocks/q/quantumespresso.py
```diff
@@ -270,8 +325,50 @@ def configure_step(self):
...
+            shutil.copytree(os.path.join(self.cfg['start_dir'], 'bin'), bindir)
+
+            upftools = []
+            if 'upf' in self.cfg['buildopts'] or 'all' in self.cfg['buildopts']:
+                upftools = ["casino2upf.x", "cpmd2upf.x", "fhi2upf.x", "fpmd2upf.x", "ncpp2upf.x",
+                            "oldcp2upf.x", "read_upf_tofile.x", "rrkj2upf.x", "uspp2upf.x", "vdb2upf.x",
+                            "virtual.x"]
+                if LooseVersion(self.version) > LooseVersion("5"):
+                    upftools.extend(["interpolate.x", "upf2casino.x"])
+                if LooseVersion(self.version) >= LooseVersion("6"):
```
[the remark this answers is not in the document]
akesandgren: Same as above. Fixed.

easybuild/easyblocks/q/quantumespresso.py
```diff
@@ -338,16 +438,20 @@ def sanity_check_step(self):
...
                 upftools.extend(["interpolate.x", "upf2casino.x"])
+            if LooseVersion(self.version) >= LooseVersion("6"):
+                upftools.extend(["fix_upf.x"])
 
         if 'vdw' in self.cfg['buildopts']:  # only for v4.x, not in v5.0 anymore
             bins.extend(["vdw.x"])
 
         if 'w90' in self.cfg['buildopts']:
-            bins.extend(["wannier90.x"])
+            bins.extend(["wannier90.x", "postw90.x"])
```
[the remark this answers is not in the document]
akesandgren: bummer...

easybuild/easyblocks/q/quantumespresso.py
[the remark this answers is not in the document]
akesandgren: QE 6.2 easyconfig updated in https://github.com/easybuilders/easybuild-easyconfigs/pull/6638

easybuild/easyblocks/q/quantumespresso.py
```diff
@@ -270,8 +326,51 @@ def configure_step(self):
...
     def install_step(self):
-        """Skip install step, since we're building in the install directory."""
-        pass
+        """Custom install step for Quantum ESPRESSO."""
+
+        # Copy all binaries
+        try:
+            # binary
+            bindir = os.path.join(self.installdir, 'bin')
+            shutil.copytree(os.path.join(self.cfg['start_dir'], 'bin'), bindir)
```
boegel: @akesandgren Please use `copy_dir` and `cope_file` from `filetools` rather than using `shutil.*` directly.

if you do, there's no need anymore to wrap this in a `try`/`except`...

easybuild/easyblocks/q/quantumespresso.py
```diff
@@ -270,8 +326,51 @@ def configure_step(self):
...
+            if 'upf' in self.cfg['buildopts'] or 'all' in self.cfg['buildopts']:
+                upftools = ["casino2upf.x", "cpmd2upf.x", "fhi2upf.x", "fpmd2upf.x", "ncpp2upf.x",
+                            "oldcp2upf.x", "read_upf_tofile.x", "rrkj2upf.x", "uspp2upf.x", "vdb2upf.x",
+                            "virtual.x"]
+                if LooseVersion(self.version) > LooseVersion("5"):
+                    upftools.extend(["interpolate.x", "upf2casino.x"])
+                if LooseVersion(self.version) >= LooseVersion("6.3"):
+                    upftools.extend(["fix_upf.x"])
+            upf_bins = [os.path.join('upftools', x) for x in upftools]
+            for x in upf_bins:
```
boegel: please avoid using single-letter variable names outside of list comprehensions, use a more meaningful name like `upf_bin`

easybuild/easyblocks/q/quantumespresso.py
```diff
@@ -73,16 +73,78 @@ def patch_step(self):
...
+
+        repls = []
+
+        extra_libs = []
+
+        comp_fam_dflags = {
+            toolchain.INTELCOMP: '-D__INTEL',
+            toolchain.GCC: '-D__GFORTRAN -D__STD_F95',
+        }
+        dflags.append(comp_fam_dflags[self.toolchain.comp_family()])
```
boegel: This will result in a hard crash when any other compiler family is used (`KeyError`), that should be handled better to produce a clean error message mentioning the compiler family?

easybuild/easyblocks/q/quantumespresso.py
```diff
@@ -73,16 +73,78 @@ def patch_step(self):
...
+            dflags.append(" -D__SCALAPACK")
+            if self.toolchain.options.get('usempi', None):
+                if get_software_root("impi") and get_software_root("imkl"):
+                    self.cfg.update('configopts', '--with-scalapack=intel')
+
+        libxc = get_software_root("libxc")
+        if libxc:
+            libxc_v = get_software_version("libxc")
+            if LooseVersion(libxc_v) < LooseVersion("3.0.1"):
+                EasyBuildError("Must use libxc >= 3.0.1")
```
boegel: missing `raise`

easybuild/easyblocks/q/quantumespresso.py
```diff
@@ -73,16 +73,78 @@ def patch_step(self):
...
+            toolchain.INTELCOMP: '-D__INTEL',
+            toolchain.GCC: '-D__GFORTRAN -D__STD_F95',
+        }
+        dflags.append(comp_fam_dflags[self.toolchain.comp_family()])
+
         if self.toolchain.options.get('openmp', False) or self.cfg['hybrid']:
             self.cfg.update('configopts', '--enable-openmp')
+            dflags.append(" -D__OPENMP")
 
         if not self.toolchain.options.get('usempi', None):
```
boegel: @akesandgren Positive logic is easier to follow, so please change and switch if/else bodies?

```python
    ...
else:
    ...
```

Same suggestion for other places you're adding an `else`

easybuild/easyblocks/q/quantumespresso.py
```diff
@@ -73,16 +73,78 @@ def patch_step(self):
...
+            if LooseVersion(self.version) >= LooseVersion("6"):
+                elpa_min_ver = "2016.11.001.pre"
+                dflags.append('-D__ELPA_2016')
+            else:
+                elpa_min_ver = "2015"
+                dflags.append('-D__ELPA_2015')
 
-        repls = []
+            if LooseVersion(elpa_v) < LooseVersion(elpa_min_ver):
+                EasyBuildError("QuantumESPRESSO %s needs ELPA to be version %s or newer" % (self-version, elpa_min_ver))
```
boegel: typo: `self-version` should be `self.version`?

easybuild/easyblocks/q/quantumespresso.py
```diff
@@ -73,16 +73,78 @@ def patch_step(self):
...
+
+            elpa_include = os.path.join(elpa, 'include')
+            repls.append(('IFLAGS', '-I%s' % os.path.join(elpa_include, 'modules'), True))
+            self.cfg.update('configopts', '--with-elpa-include=%s' % elpa_include)
+            if self.toolchain.options.get('openmp', False):
+                elpa_lib = os.path.join(elpa, 'lib', 'libelpa_openmp.a')
+            else:
+                elpa_lib = os.path.join(elpa, 'lib', 'libelpa.a')
+
+            self.cfg.update('configopts', '--with-elpa-lib=%s' % elpa_lib)
```
boegel: flesh out common part, easier to digest?

```python
if self.toolchain.options.get('openmp', False):
    elpa_lib = 'libelpa_openmp.a'
else:
    elpa_lib = 'libelpa.a'
self.cfg.update('configopts', '--with-elpa-lib=%s' % os.path.join(elpa, 'lib', elpa_lib))

File: easybuild/easyblocks/q/quantumespresso.py [hunk 12]
```diff
@@ -93,17 +155,13 @@ def configure_step(self):
...
             env.setvar('FCCPP', "%s -E" % os.getenv('CC'))
 
-        super(EB_QuantumESPRESSO, self).configure_step()
+        if self.toolchain.comp_family() in [toolchain.INTELCOMP]:
+            repls.append(('F90FLAGS', '-fpp', True))
 
-        # compose list of DFLAGS (flag, value, keep_stuff)
-        # for guidelines, see include/defs.h.README in sources
-        dflags = []
+        if self.toolchain.comp_family() in [toolchain.GCC]:
```
boegel: `elif`?

easybuild/easyblocks/q/quantumespresso.py
```diff
@@ -93,17 +155,13 @@ def configure_step(self):
             # also define $FCCPP, but do *not* include -C (comments should not be preserved when preprocessing Fortran)
             env.setvar('FCCPP', "%s -E" % os.getenv('CC'))
 
-        super(EB_QuantumESPRESSO, self).configure_step()
+        if self.toolchain.comp_family() in [toolchain.INTELCOMP]:
```
boegel: if there's only a single one, use `==` rather than `in`?

easybuild/easyblocks/q/quantumespresso.py
```diff
@@ -246,7 +302,7 @@ def configure_step(self):
 
         # move non-espresso directories to where they're expected and create symlinks
         try:
-            dirnames = [d for d in os.listdir(self.builddir) if not d == self.install_subdir]
+            dirnames = [d for d in os.listdir(self.builddir) if not d == self.install_subdir and not d == 'd3q-latest']
```
boegel: `and d != 'd3q-latest'`

easybuild/easyblocks/q/quantumespresso.py
akesandgren: Probably, not something i changed, it just moved from line 102
We can target that later i think

akesandgren:
It shouldn't affect older builds, the changes are basically related to things in v6 and later.
And QuantumESPRESSO-5.4.0-foss-2016b-hybrid.eb built correctly. Haven't tried anything older since i don't have those toolchains.

akesandgren:
I seem to have missed one step (again) for the ELPA part. Retesting with fix, new commit coming soon.

easybuild/easyblocks/q/quantumespresso.py
akesandgren: would a code like:
if not self.toolchain.comp_family() in comp_fam_dflags:
  raise EBerror('text')

work?

akesandgren: For example:
        if not self.toolchain.comp_family() in comp_fam_dflags:
            raise EasyBuildError("Toolchain %s is not yet supported in QuantumESPRESSO")

akesandgren: perhaps "supported for" is better?

boegel: @akesandgren Yes, that should work and result in a decent error rather than a `KeyError` + traceback.

```python
comp_fam =  self.toolchain.comp_family()
if comp_fam in comp_fam_dflags:
    dflags.append(comp_fam_dflags[comp_fam])
else:
    raise EasyBuildError("...")

File: easybuild/easyblocks/q/quantumespresso.py [hunk 15]
```diff
@@ -246,7 +303,7 @@ def configure_step(self):
 
         # move non-espresso directories to where they're expected and create symlinks
         try:
-            dirnames = [d for d in os.listdir(self.builddir) if not d == self.install_subdir]
+            dirnames = [d for d in os.listdir(self.builddir) if d != self.install_subdir and d != 'd3q-latest']
```
boegel: @akesandgren Style nitpicking: this is slightly cleaner (I overlooked that earlier):

```python
dirnames = [d for d in os.listdir(self.builddir) if d not in [self.install_subdir, 'd3q-latest']]

File: easybuild/easyblocks/q/quantumespresso.py [hunk 16]
```diff
@@ -361,28 +464,19 @@ def sanity_check_step(self):
...
         super(EB_QuantumESPRESSO, self).sanity_check_step(custom_paths=custom_paths)
-
-    def make_module_req_guess(self):
-        """Custom path suggestions for Quantum ESPRESSO."""
-        guesses = super(EB_QuantumESPRESSO, self).make_module_req_guess()
-
-        # order matters here, 'bin' should be *last* in this list to ensure it gets prepended to $PATH last,
-        # so it gets preference over the others
-        # this is important since some binaries are available in two places (e.g. dos.x in both bin and WANT/bin)
-        bindirs = ['upftools', 'WANT/bin', 'YAMBO/bin', 'bin']
```
boegel: @akesandgren This can be removed thanks to the custom `install_step`, but I don't see the `YAMBO/bin` binaries being copied?

I see these for older versions of QE (5.2.1 & older) in `YAMBO/bin`:
```
a2y  p2y  yambo  ypp
```

easybuild/easyblocks/q/quantumespresso.py
```diff
@@ -361,28 +464,19 @@ def sanity_check_step(self):
...
-                        'files': [os.path.join(pref, 'bin', x) for x in bins] +
-                                 [os.path.join(pref, 'upftools', x) for x in upftools] +
-                                 [os.path.join(pref, 'WANT', 'bin', x) for x in want_bins] +
-                                 [os.path.join(pref, 'YAMBO', 'bin', x) for x in yambo_bins],
-                        'dirs': [os.path.join(pref, 'include')]
+                        'files': [os.path.join('bin', x) for x in bins] +
+                                 [os.path.join('bin', x) for x in upftools] +
+                                 [os.path.join('bin', x) for x in want_bins] +
+                                 [os.path.join('bin', x) for x in yambo_bins] +
+                                 [os.path.join('bin', x) for x in d3q_bins],
```
boegel: @akesandgren Since everything is located in the same directory now, you can trim this down (and why not fix the indent while you're at it):

```python
custom_paths = {
    'files': [os.path.join('bin', x) for x in bins + upftools + want_bins + yambo_bins + d3q_bins],
    'dirs': ['bin'],
}
```

Also, checking for a non-empty `bin` directory is kind of silly when we're also checking for a bunch of binaries in that dir, so might as well make `dirs` use `[]`.

easybuild/easyblocks/q/quantumespresso.py
```diff
@@ -270,8 +327,46 @@ def configure_step(self):
...
+                wanttools.extend(["cmplx_bands.x", "decay.x", "sax2qexml.x", "sum_sgm.x"])
+        want_bins = [os.path.join('WANT', 'bin', x) for x in wanttools]
+        for want_bin in want_bins:
+            copy_file(os.path.join(self.cfg['start_dir'], want_bin), bindir)
+
+        # Pick up files not installed in bin
+        w90tools = []
+        if 'w90' in self.cfg['buildopts']:
+            if LooseVersion(self.version) < LooseVersion("6"):
+                w90tools = ["w90chk2chk.x"]
```
boegel: This doesn't seem to be there with QE 4.2...

akesandgren: You mean w90chk2chk.x or ?

akesandgren: Anyway, after a bit of looking back and forth I think i found the correct version checks for W90.

boegel: Yes, I did mean `w90chk2chk.x`

easybuild/easyblocks/q/quantumespresso.py
```diff
@@ -270,8 +331,52 @@ def configure_step(self):
...
+                w90tools = ["postw90.x"]
+                if LooseVersion(self.version) < LooseVersion("6.1"):
+                    w90tools = ["w90chk2chk.x"]
+        w90_bins = [os.path.join('W90', x) for x in w90tools]
+        for w90_bin in w90_bins:
+            copy_file(os.path.join(self.cfg['start_dir'], w90_bin), bindir)
+
+        yambo_bins = []
+        if 'yambo' in self.cfg['buildopts']:
+            yambo_bins = ["a2y", "p2y", "yambo", "ypp"]
```
boegel: @akesandgren These binaries are in `YAMBO/bin`, so you'll need this above this `for` loop:

```python
yambo_bins = [os.path.join('YAMBO', 'bin', x) for x in yambo_bins]
```

easybuild/easyblocks/q/quantumespresso.py
```diff
@@ -338,16 +447,21 @@ def sanity_check_step(self):
...
+            if LooseVersion(self.version) >= LooseVersion("6.3"):
+                upftools.extend(["fix_upf.x"])
 
         if 'vdw' in self.cfg['buildopts']:  # only for v4.x, not in v5.0 anymore
             bins.extend(["vdw.x"])
 
         if 'w90' in self.cfg['buildopts']:
-            bins.extend(["wannier90.x"])
+            if LooseVersion(self.version) < LooseVersion("6"):
+                bins.extend(["w90chk2chk.x"])
```
boegel: @akesandgren Version checking doesn't match what's in `install_step`, making the sanity check fail for QE 4.2 for example...

akesandgren: Doh.... will fix.

akesandgren:
@boegel Retest with existing easyconfigs?

easybuild/easyblocks/q/quantumespresso.py
```diff
@@ -73,18 +73,86 @@ def patch_step(self):
...
+            if self.toolchain.options.get('openmp', False):
+                elpa_lib = 'libelpa_openmp.a'
+            else:
+                elpa_lib = 'libelpa.a'
 
-        if self.toolchain.comp_family() in [toolchain.INTELCOMP]:
+            elpa_lib = os.path.join(elpa, 'lib', elpa_lib)
+            self.cfg.update('configopts', '--with-elpa-lib=%s' % elpa_lib)
+
+        if self.toolchain.comp_family() == toolchain.INTELCOMP:
```
boegel: @akesandgren Please re-use `comp_fam` here

easybuild/easyblocks/q/quantumespresso.py
```diff
@@ -93,17 +161,12 @@ def configure_step(self):
             # also define $FCCPP, but do *not* include -C (comments should not be preserved when preprocessing Fortran)
             env.setvar('FCCPP', "%s -E" % os.getenv('CC'))
 
-        super(EB_QuantumESPRESSO, self).configure_step()
+        if self.toolchain.comp_family() == toolchain.INTELCOMP:
```
boegel: @akesandgren Please re-use `comp_fam` here

easybuild/easyblocks/q/quantumespresso.py
```diff
@@ -93,17 +161,12 @@ def configure_step(self):
...
+            repls.append(('F90FLAGS', '-fpp', True))
+        elif self.toolchain.comp_family() == toolchain.GCC:
```
boegel: @akesandgren Please re-use `comp_fam` here

easybuild/easyblocks/q/quantumespresso.py
```diff
@@ -183,11 +243,12 @@ def configure_step(self):
                         line = re.sub(r"^(%s\s*=[ \t]*).*$" % k, r"\1%s" % v, line)
 
                 # fix preprocessing directives for .f90 files in make.sys if required
-                if self.toolchain.comp_family() in [toolchain.GCC]:
-                    line = re.sub(r"\$\(MPIF90\) \$\(F90FLAGS\) -c \$<",
-                                  "$(CPP) -C $(CPPFLAGS) $< -o $*.F90\n" +
-                                  "\t$(MPIF90) $(F90FLAGS) -c $*.F90 -o $*.o",
-                                  line)
+                if LooseVersion(self.version) < LooseVersion("6.0"):
+                    if self.toolchain.comp_family() == toolchain.GCC:
```
boegel: @akesandgren Please re-use `comp_fam` here

easybuild/easyblocks/q/quantumespresso.py
```diff
@@ -234,7 +295,7 @@ def configure_step(self):
             try:
                 for line in fileinput.input(make_sys_in_path, inplace=1, backup='.orig.eb'):
                     # fix preprocessing directives for .f90 files in make.sys if required
-                    if self.toolchain.comp_family() in [toolchain.GCC]:
+                    if self.toolchain.comp_family() == toolchain.GCC:
```
boegel: @akesandgren Please re-use `comp_fam` here

boegel:
Thanks @akesandgren!

PR #1314, opened 2017-12-08
enhance VMD easyblock to also build Surf & Stride + pick up netCDF for VMD >= 1.9.3

akesandgren:
Add support for:
 - netCDF
 - Surf
 - Stride


easybuild/easyblocks/v/vmd.py
```diff
@@ -45,6 +46,15 @@
...
     """Easyblock for building and installing VMD"""
 
+    def extract_step(self):
+        """Custom extract step for VMD."""
+        super(EB_VMD, self).extract_step()
+
+        vmddir = os.path.join(self.src[0]['finalpath'], '%s-%s' % (self.name.lower(), self.version))
+        if LooseVersion(self.version) >= LooseVersion("1.9.3"):
+            change_dir(os.path.join(vmddir, 'lib', 'surf'))
+            extract_file('surf.tar.Z', os.getcwd())
```
boegel: @akesandgren Any reference to what this is exactly? Will it always be there?

easybuild/easyblocks/v/vmd.py
```diff
@@ -150,6 +165,40 @@ def configure_step(self):
...
+
+        vmddir = os.path.join(self.cfg['start_dir'], '%s-%s' % (self.name.lower(), self.version))
+        self.have_stride = False
+        # Build Surf, which is part of VMD as of VMD version 1.9.3
+        if LooseVersion(self.version) >= LooseVersion("1.9.3"):
+            change_dir(os.path.join(vmddir, 'lib', 'surf'))
+            surf_cmd = 'make CC="%s" OPT="%s"' % (os.environ['CC'], os.environ['CFLAGS'])
+            run_cmd(surf_cmd)
+            # Build Stride if it was downloaded
+            change_dir(os.path.join(vmddir, 'lib', 'stride'))
```
boegel: What if this directory doesn't exist?

easybuild/easyblocks/v/vmd.py
```diff
@@ -150,6 +165,40 @@ def configure_step(self):
...
+    def build_step(self):
+        """Custom build step for VMD."""
+        super(EB_VMD, self).build_step()
+
+        vmddir = os.path.join(self.cfg['start_dir'], '%s-%s' % (self.name.lower(), self.version))
+        self.have_stride = False
+        # Build Surf, which is part of VMD as of VMD version 1.9.3
+        if LooseVersion(self.version) >= LooseVersion("1.9.3"):
+            change_dir(os.path.join(vmddir, 'lib', 'surf'))
+            surf_cmd = 'make CC="%s" OPT="%s"' % (os.environ['CC'], os.environ['CFLAGS'])
```
boegel: `s/surf_cmd/surf_build_cmd/g`?

easybuild/easyblocks/v/vmd.py
```diff
@@ -95,7 +109,8 @@ def configure_step(self):
 
         # plugins need to be built first (see http://www.ks.uiuc.edu/Research/vmd/doxygen/compiling.html)
         change_dir(os.path.join(self.cfg['start_dir'], 'plugins'))
-        cmd = "make LINUXAMD64 TCLLIB='-F%s' TCLINC='-I%s' %s" % (tcllib, tclinc, self.cfg['buildopts'])
+        cmd = "make LINUXAMD64 TCLLIB='-L%s' TCLINC='-I%s' " % (tcllib, tclinc)
+        cmd += "NETCDFLIB='-L%s' NETCDFINC='-I%s' %s" % (netcdflib, netcdfinc, self.cfg['buildopts'])
```
boegel: Maybe we should clean this up as follows:

```python
cmd = ' '.join([
    'make',
    'LINUXAMD64',
    "TCLLIB='-L%s'" % tcllib,
    "TCLINC='-I%s'" % tclinc,
    "NETCDFLIB='-L%s'" % netcdflib,
    "NETCDFINC='-I%s'" % netcdfinc,
    self.cfg['buildopts'],
])

File: easybuild/easyblocks/v/vmd.py [hunk 5]
```diff
@@ -45,6 +46,15 @@
 class EB_VMD(ConfigureMake):
     """Easyblock for building and installing VMD"""
 
+    def extract_step(self):
+        """Custom extract step for VMD."""
+        super(EB_VMD, self).extract_step()
+
+        vmddir = os.path.join(self.src[0]['finalpath'], '%s-%s' % (self.name.lower(), self.version))
```
boegel: @akesandgren We should determine this once, and save it in `self.vmd_dir`, since we're using the same thing also in other places. You can probably also use:

```python
self.vmd_dir = os.path.join(self.builddir, '%s-%s' % (self.name.lower(), self.version))
```

We should set this in `__init__`, to make sure it's *always* defined (also under `--module-only` for example)

easybuild/easyblocks/v/vmd.py
akesandgren: There is something in the docco about it but i can't remember, it's something users sometimes need.
As of that version they moved it into VMD so should always be there. Unless they change their mind...

akesandgren: page 4 of the install guide says that since surf is free it is provided as is in the distro of VMD.
As for what it really is.... no idea.

easybuild/easyblocks/v/vmd.py
akesandgren: Ahhh, need to deal with that. best approach being??

easybuild/easyblocks/v/vmd.py
akesandgren: Can i use self.builddir in the extract step? I thought it didn't setup until later.

akesandgren: Found an example of using self.builddir in __init__ at least (Matlab) so should be doable.
But start_dir vs builddir and the above self.src[0]['finalpath'], I can never remember which is what...

easybuild/easyblocks/v/vmd.py
boegel: `os.path.exists` before `change_dir` + proper error reporting

easybuild/easyblocks/v/vmd.py
boegel: Can we include a URL in a comment here that provides some context?

If it's not there at some point, we'll notice quickly enough.

easybuild/easyblocks/v/vmd.py
boegel: You can use `self.builddir` early on, as soon as the easyblock is initialized (it's defined in `EasyBlock.__init__`).

`self.builddir` is the top-level build directory.
`self.start_dir` is the build directory in which the build will be performed by default. This is typically the unpack source directory (so it's under `self.builddir`). But if you have multiple source files, it may be equal to `self.builddir`.
`self.src[0]['finalpath']` is the location where the first source file was unpacked. If there's only one source file, it's the same as `self.start_dir`

easybuild/easyblocks/v/vmd.py
akesandgren: since stride isn't mandatory, should we just issue a warning or raise an error do you think?

boegel:
@akesandgren Ran into this issue when rebuilding `VMD-1.9.3-intel-2018b-Python-2.7.15.eb` using this modified `VMD` easyblock:

```
== 2018-12-14 12:04:33,963 run.py:173 DEBUG run_cmd: running cmd make CC="icc" OPT="-O2 -xHost -ftz -fp-speculation=safe -fp-model source" (in /tmp/easybuild_build/VMD/1.9.3/intel-2018b-Python-2.7.15/vmd-1.9.3/lib/surf)
== 2018-12-14 12:04:33,963 run.py:192 INFO running cmd: make CC="icc" OPT="-O2 -xHost -ftz -fp-speculation=safe -fp-model source"
== 2018-12-14 12:04:34,225 build_log.py:162 ERROR EasyBuild crashed with an error (at ?:124 in __init__): cmd "make CC="icc" OPT="-O2 -xHost -ftz -fp-speculation=safe -fp-model source"" exited with exit code 2 and output:
make: *** No rule to make target `/usr/include/sgidefs.h', needed by `surf.o'.  Stop.
```

akesandgren:
Oh, unexpected...

boegel:
Changes tested with https://github.com/easybuilders/easybuild-easyconfigs/pull/7305, lgtm, so going in, thanks @akesandgren!

PR #1316, opened 2017-12-08
Add new block for libgpuarray

akesandgren:


boegel:
@akesandgren Hmm, this may be overkill, see https://github.com/easybuilders/easybuild-easyconfigs/pull/5429 (which can even be simplified, I think)...

akesandgren:
Perhaps. Didn't think of doing it that way.
But if it produces the same results then i have no objections.

boegel:
@akesandgren This can be closed now that https://github.com/easybuilders/easybuild-easyconfigs/pull/5429 is merged?

akesandgren:
Not needed any longer, solved by a better easyconfig

PR #1317, opened 2017-12-08
Add new easyblock for COMSOL

akesandgren:
Has support for MATLAB as a dependency.

### DIFF
--- a/easybuild/easyblocks/c/comsol.py
+++ b/easybuild/easyblocks/c/comsol.py
@@ -25,6 +25,7 @@
 """
 EasyBuild support for installing Comsol, implemented as an easyblock
 
+@author: Mikael OEhman (Chalmers University of Technology)
 @author: Ake Sandgren (HPC2N, Umea University)
 """
 import re
@@ -36,7 +37,7 @@ import easybuild.tools.environment as env
 from easybuild.easyblocks.generic.packedbinary import PackedBinary
 from easybuild.framework.easyconfig import CUSTOM
 from easybuild.tools.build_log import EasyBuildError
-from easybuild.tools.filetools import adjust_permissions, find_flexlm_license, read_file, write_file
+from easybuild.tools.filetools import adjust_permissions, copy_file, find_flexlm_license, read_file, write_file
 from easybuild.tools.modules import get_software_root
 from easybuild.tools.run import run_cmd
 from easybuild.tools.systemtools import get_shared_lib_ext
@@ -56,11 +57,11 @@ class EB_Comsol(PackedBinary):
 
         # The tar file comes from the DVD and has 0444 as permission at
         # the top dir.
-        adjust_permissions(self.cfg['start_dir'], stat.S_IWUSR)
+        adjust_permissions(self.start_dir, stat.S_IWUSR)
 
         default_lic_env_var = 'LM_LICENSE_FILE'
         lic_specs, self.license_env_var = find_flexlm_license(custom_env_vars=[default_lic_env_var],
-            lic_specs=[self.cfg['license_file']])
+                                                              lic_specs=[self.cfg['license_file']])
 
         if lic_specs:
             if self.license_env_var is None:
@@ -76,52 +77,48 @@ class EB_Comsol(PackedBinary):
             msg += "specify 'license_file', or define $LM_LICENSE_FILE"
             raise EasyBuildError(msg)
 
-        try:
-            shutil.copyfile(os.path.join(self.cfg['start_dir'], 'setupconfig.ini'), self.configfile)
-            config = read_file(self.configfile)
+        copy_file(os.path.join(self.start_dir, 'setupconfig.ini'), self.configfile)
+        config = read_file(self.configfile)
 
-            regdest = re.compile(r"^installdir\s*=.*", re.M)
-            reggui = re.compile(r"^showgui\s*=.*", re.M)
-            regagree = re.compile(r"^agree\s*=.*", re.M)
-            reglicpath = re.compile(r"^license\s*=.*", re.M)
-            reglinuxlauncher = re.compile(r"^linuxlauncher\s*=.*", re.M)
-            regsymlinks = re.compile(r"^symlinks\s*=.*", re.M)
-            regfileassoc = re.compile(r"^fileassoc\s*=.*", re.M)
+        config_vars = {
+            'agree': '1',
+            'desktopshortcuts': '0',
+            'fileassoc': '0',
+            'firewall': '0',
+            'installdirs': self.installdir,
+            'license': self.license_file,
+            'licmanager': '0',
+            'linuxlauncher': '0',
+            'showgui': '0',
+            'startmenushortcuts': '0',
+            'symlinks': '0',
+        }
 
-            config = regdest.sub("installdir=%s" % self.installdir, config)
-            config = reggui.sub("showgui=0", config)
-            config = regagree.sub("agree=1", config)
-            config = reglicpath.sub("license=%s" % self.license_file, config)
-            config = reglinuxlauncher.sub("linuxlauncher=0", config)
-            config = regsymlinks.sub("symlinks=0", config)
-            config = regfileassoc.sub("fileassoc=0", config)
+        matlab_root = get_software_root("MATLAB")
+        if matlab_root:
+            config_vars.update('matlabdir': matlab_root)
 
-            matlab_root = get_software_root("MATLAB")
-            if matlab_root:
-                regmatlab = re.compile(r"^matlabdir\s*=.*", re.M)
-                config = regmatlab.sub("matlabdir=%s" % matlabroot, config)
+        for key, val in config_vars.items():
+            regex = re.compile(r"^%s\s*=.*" % key, re.M)
+            config = regex.sub("%s=%s" % (key, val), config)
 
-            write_file(self.configfile, config)
-
-        except IOError, err:
-            raise EasyBuildError("Failed to create installation config file %s: %s", self.configfile, err)
+        write_file(self.configfile, config)
 
         self.log.debug('configuration file written to %s:\n %s', self.configfile, config)
 
     def install_step(self):
         """COMSOL install procedure using 'install' command."""
 
-        src = os.path.join(self.cfg['start_dir'], 'setup')
+        src = os.path.join(self.start_dir, 'setup')
 
         # make sure setup script is executable
         adjust_permissions(src, stat.S_IXUSR)
 
         # make sure $DISPLAY is not defined, which may lead to (hard to trace) problems
         # this is a workaround for not being able to specify --nodisplay to the install scripts
-        if 'DISPLAY' in os.environ:
-            os.environ.pop('DISPLAY')
+        env.unset_env_vars(['DISPLAY'])
 
-        cmd = "%s ./setup -s %s %s" % (self.cfg['preinstallopts'], self.configfile, self.cfg['installopts'])
+        cmd = ' '.join(self.cfg['preinstallopts'], src, '-s', self.configfile, self.cfg['installopts'])
         run_cmd(cmd, log_all=True, simple=True)
 
     def sanity_check_step(self):
--- a/easybuild/easyblocks/c/comsol.py
+++ b/easybuild/easyblocks/c/comsol.py
@@ -96,7 +96,7 @@ class EB_Comsol(PackedBinary):
 
         matlab_root = get_software_root("MATLAB")
         if matlab_root:
-            config_vars.update('matlabdir': matlab_root)
+            config_vars.update({'matlabdir': matlab_root})
 
         for key, val in config_vars.items():
             regex = re.compile(r"^%s\s*=.*" % key, re.M)
--- a/easybuild/easyblocks/c/comsol.py
+++ b/easybuild/easyblocks/c/comsol.py
@@ -59,7 +59,7 @@ class EB_COMSOL(PackedBinary):
         # the top dir.
         adjust_permissions(self.start_dir, stat.S_IWUSR)
 
-        default_lic_env_var = 'LM_LICENSE_FILE'
+        default_lic_env_var = 'LMCOMSOL_LICENSE_FILE'
         lic_specs, self.license_env_var = find_flexlm_license(custom_env_vars=[default_lic_env_var],
                                                               lic_specs=[self.cfg['license_file']])
 
--- a/easybuild/easyblocks/c/comsol.py
+++ b/easybuild/easyblocks/c/comsol.py
@@ -1,0 +1,136 @@
+##
+# Copyright 2009-2017 Ghent University
+#
+# This file is part of EasyBuild,
+# originally created by the HPC team of Ghent University (http://ugent.be/hpc/en),
+# with support of Ghent University (http://ugent.be/hpc),
+# the Flemish Supercomputer Centre (VSC) (https://www.vscentrum.be),
+# Flemish Research Foundation (FWO) (http://www.fwo.be/en)
+# and the Department of Economy, Science and Innovation (EWI) (http://www.ewi-vlaanderen.be/en).
+#
+# https://github.com/easybuilders/easybuild
+#
+# EasyBuild is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation v2.
+#
+# EasyBuild is distributed in the hope that it will be useful,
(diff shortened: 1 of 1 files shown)


boegel:
@akesandgren Is this a modified version of #1296?

easybuild/easyblocks/c/comsol.py
```diff
@@ -0,0 +1,136 @@
...
+
+            self.license_file = os.pathsep.join(lic_specs)
+            env.setvar(self.license_env_var, self.license_file)
+        else:
+            msg = "No viable license specifications found; "
+            msg += "specify 'license_file', or define $LM_LICENSE_FILE"
+            raise EasyBuildError(msg)
+
+        try:
+            shutil.copyfile(os.path.join(self.cfg['start_dir'], 'setupconfig.ini'), self.configfile)
```
boegel: Please use `copy_file` function from `filetools` module, see http://easybuild.readthedocs.io/en/latest/api/easybuild.tools.filetools.html#easybuild.tools.filetools.copy_file.

easybuild/easyblocks/c/comsol.py
```diff
@@ -0,0 +1,136 @@
...
+    def configure_step(self):
+        """Configure COMSOL installation: create license file."""
+
+        # The tar file comes from the DVD and has 0444 as permission at
+        # the top dir.
+        adjust_permissions(self.cfg['start_dir'], stat.S_IWUSR)
+
+        default_lic_env_var = 'LM_LICENSE_FILE'
+        lic_specs, self.license_env_var = find_flexlm_license(custom_env_vars=[default_lic_env_var],
+            lic_specs=[self.cfg['license_file']])
```
boegel: please align `lic_specs` with `custom_env_vars` on line above

easybuild/easyblocks/c/comsol.py
```diff
@@ -0,0 +1,136 @@
...
+        super(EB_Comsol, self).__init__(*args, **kwargs)
+        self.comp_fam = None
+        self.configfile = os.path.join(self.builddir, 'my_setupconfig.ini')
+
+    def configure_step(self):
+        """Configure COMSOL installation: create license file."""
+
+        # The tar file comes from the DVD and has 0444 as permission at
+        # the top dir.
+        adjust_permissions(self.cfg['start_dir'], stat.S_IWUSR)
```
boegel: you can use `self.start_dir` now (since EasyBuild v3.5.0)

easybuild/easyblocks/c/comsol.py
```diff
@@ -0,0 +1,136 @@
...
+            regsymlinks = re.compile(r"^symlinks\s*=.*", re.M)
+            regfileassoc = re.compile(r"^fileassoc\s*=.*", re.M)
+
+            config = regdest.sub("installdir=%s" % self.installdir, config)
+            config = reggui.sub("showgui=0", config)
+            config = regagree.sub("agree=1", config)
+            config = reglicpath.sub("license=%s" % self.license_file, config)
+            config = reglinuxlauncher.sub("linuxlauncher=0", config)
+            config = regsymlinks.sub("symlinks=0", config)
+            config = regfileassoc.sub("fileassoc=0", config)
```
boegel: This should really be a loop... Also, #1296 is tweaking more entries?

```python
config_vars = {
    'agree': '1',
    'fileassoc': '0',
    'license': self.license_file,
    'linuxlauncher': '0',
    'showgui': '0',
    'symlinks': '0',
}
for key, val in config_vars.items():
    regex = re.compile(r"^%s\s*=.*" % key, re.M)
    config = regex.sub("%s=%s" % (key, val), config)
```

easybuild/easyblocks/c/comsol.py
```diff
@@ -0,0 +1,136 @@
...
+
+            matlab_root = get_software_root("MATLAB")
+            if matlab_root:
+                regmatlab = re.compile(r"^matlabdir\s*=.*", re.M)
+                config = regmatlab.sub("matlabdir=%s" % matlabroot, config)
```
boegel: Please use this above, before the `re.compile`, `regex.sub` loop:

```python
matlab_root = get_software_root("MATLAB")
if matlab_root:
    config_vars.update('matlabdir': matlab_root)
```

(note the typo `matlab_root` vs `matlabroot`...)

easybuild/easyblocks/c/comsol.py
```diff
@@ -0,0 +1,136 @@
...
+
+            write_file(self.configfile, config)
+
+        except IOError, err:
```
boegel: Any `IOError` that occur are handled by `read_file`, `write_file` or `copy_file` already, so no need to wrap this in a `try` `except`.

easybuild/easyblocks/c/comsol.py
```diff
@@ -0,0 +1,136 @@
...
+
+        src = os.path.join(self.cfg['start_dir'], 'setup')
+
+        # make sure setup script is executable
+        adjust_permissions(src, stat.S_IXUSR)
+
+        # make sure $DISPLAY is not defined, which may lead to (hard to trace) problems
+        # this is a workaround for not being able to specify --nodisplay to the install scripts
+        if 'DISPLAY' in os.environ:
+            os.environ.pop('DISPLAY')
```
boegel: `pop` should only be used if you need the value, if you just want to wipe it, use `del os.environ['DISPLAY']`, or even better `unset_env_vars` from `easybuild.tools.environment` (since that also properly logs)

easybuild/easyblocks/c/comsol.py
```diff
@@ -0,0 +1,136 @@
...
+
+        cmd = "%s ./setup -s %s %s" % (self.cfg['preinstallopts'], self.configfile, self.cfg['installopts'])
```
boegel: Why not use `src` here rather than hardcoding `setup` again?

```python
cmd = ' '.join([self.cfg['preinstallopts'], src, '-s', self.configfile, self.cfg['installopts']])

File: easybuild/easyblocks/c/comsol.py [hunk 9]
```diff
@@ -0,0 +1,136 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with EasyBuild.  If not, see <http://www.gnu.org/licenses/>.
+##
+"""
+EasyBuild support for installing Comsol, implemented as an easyblock
+
+@author: Ake Sandgren (HPC2N, Umea University)
```
boegel: This seems to be based on #1296, so please also include `@author: Mikael OEhman (Chalmers University of Technology)`

akesandgren:
I don't think it is. Our initial Comsol install was done Nov 2:nd

boegel:
@akesandgren Hah, well ok, it's just that they look suspiciously similar...

Since @Micket also opened a PR for a COMSOL easyblock and what will end up merged will be a joined version of both, it still seems appropriate to also include him in the `authors` part?

akesandgren:
Yeah, can do that.

akesandgren:
I think i took mine from the matlab easyblock. Hence the unsetting of DISPLAY which i'm not sure is really needed for comsol.

fgeorgatos:
Hi,

I think I'm the one to blame for that: at some point I figured out that
DISPLAY affected the way MATLAB was setting up itself and it was triggering
failures.

In fact, as a matter of reproducibility, that as well as some LC_* and
friend  variables should rather have some default (but overridable)
settings.

F.


On Sun, 17 Dec 2017 at 1:03 PM Åke Sandgren <notifications@github.com>
wrote:

> I think i took mine from the matlab easyblock. Hence the unsetting of
> DISPLAY which i'm not sure is really needed ...
>
-- 
echo "sysadmin know better bash than english"|sed s/min/mins/ \
  | sed 's/better bash/bash better/' # signal detected in a CERN forum

easybuild/easyblocks/c/comsol.py
```diff
@@ -0,0 +1,133 @@
...
+        self.configfile = os.path.join(self.builddir, 'my_setupconfig.ini')
+
+    def configure_step(self):
+        """Configure COMSOL installation: create license file."""
+
+        # The tar file comes from the DVD and has 0444 as permission at
+        # the top dir.
+        adjust_permissions(self.start_dir, stat.S_IWUSR)
+
+        default_lic_env_var = 'LM_LICENSE_FILE'
```
Micket: This seems to be set by find_flexlm_license automatically so it wouldn't be necessary, and custom_env_vars need not be set to achieve this.
But shouldn't this be set to something COMSOL-specific to allow multiple flex softwares to be installed in one swoop?

mboisson:
@akesandgren, you have a PR to your PR, https://github.com/akesandgren/easybuild-easyblocks/pull/8  to fix the capitalization of COMSOL

mboisson:
If you merge akesandgren#8, this PR works for me (I just installed COMSOL 5.3a successfully with it).

easybuild/easyblocks/c/comsol.py
```diff
@@ -0,0 +1,133 @@
...
+        self.configfile = os.path.join(self.builddir, 'my_setupconfig.ini')
+
+    def configure_step(self):
+        """Configure COMSOL installation: create license file."""
+
+        # The tar file comes from the DVD and has 0444 as permission at
+        # the top dir.
+        adjust_permissions(self.start_dir, stat.S_IWUSR)
+
+        default_lic_env_var = 'LMCOMSOL_LICENSE_FILE'
```
boegel: @akesandgren Is this the standard env var to use for `COMSOL`, mentioned it the documentation?

easybuild/easyblocks/c/comsol.py
```diff
@@ -0,0 +1,133 @@
+##
+# Copyright 2009-2017 Ghent University
```
boegel: 2018 please :)

easybuild/easyblocks/c/comsol.py
```diff
@@ -0,0 +1,133 @@
...
+# along with EasyBuild.  If not, see <http://www.gnu.org/licenses/>.
+##
+"""
+EasyBuild support for installing COMSOL, implemented as an easyblock
+
+@author: Mikael OEhman (Chalmers University of Technology)
+@author: Ake Sandgren (HPC2N, Umea University)
+"""
+import re
+import os
```
boegel: please keep these sorted alphabetically (`o` before `r`)

easybuild/easyblocks/c/comsol.py
```diff
@@ -0,0 +1,133 @@
...
+import shutil
+import stat
+
+import easybuild.tools.environment as env
+from easybuild.easyblocks.generic.packedbinary import PackedBinary
+from easybuild.framework.easyconfig import CUSTOM
```
boegel: please remove unused `import`s

easybuild/easyblocks/c/comsol.py
```diff
@@ -0,0 +1,133 @@
...
+from easybuild.tools.systemtools import get_shared_lib_ext
+
+
+class EB_COMSOL(PackedBinary):
+    """Support for installing COMSOL."""
+
+    def __init__(self, *args, **kwargs):
+        """Add extra config options specific to COMSOL."""
+        super(EB_COMSOL, self).__init__(*args, **kwargs)
+        self.comp_fam = None
```
boegel: `self.comp_fam` is unused, so please remove

easybuild/easyblocks/c/comsol.py
```diff
@@ -0,0 +1,133 @@
...
+        self.configfile = os.path.join(self.builddir, 'my_setupconfig.ini')
+
+    def configure_step(self):
+        """Configure COMSOL installation: create license file."""
+
+        # The tar file comes from the DVD and has 0444 as permission at
+        # the top dir.
```
boegel: no need to wrap comment across two lines?

easybuild/easyblocks/c/comsol.py
```diff
@@ -0,0 +1,133 @@
...
+                                                              lic_specs=[self.cfg['license_file']])
+
+        if lic_specs:
+            if self.license_env_var is None:
+                self.log.info("Using COMSOL license specifications from 'license_file': %s", lic_specs)
+                self.license_env_var = default_lic_env_var
+            else:
+                self.log.info("Using COMSOL license specifications from $%s: %s", self.license_env_var, lic_specs)
+
+            self.license_file = os.pathsep.join(lic_specs)
```
boegel: should be `os.path.join`?

easybuild/easyblocks/c/comsol.py
```diff
@@ -0,0 +1,133 @@
...
+            env.setvar(self.license_env_var, self.license_file)
+        else:
+            msg = "No viable license specifications found; "
+            msg += "specify 'license_file', or define $LM_LICENSE_FILE"
```
boegel: should use `default_lic_env_var` rather than hardcoded `$LM_LICENSE_FILE`?

easybuild/easyblocks/c/comsol.py
```diff
@@ -0,0 +1,133 @@
...
+            config = regex.sub("%s=%s" % (key, val), config)
+
+        write_file(self.configfile, config)
+
+        self.log.debug('configuration file written to %s:\n %s', self.configfile, config)
+
+    def install_step(self):
+        """COMSOL install procedure using 'install' command."""
+
+        src = os.path.join(self.start_dir, 'setup')
```
boegel: rename `src` to `setup_script`?

easybuild/easyblocks/c/comsol.py
akesandgren: Well, don't have any documentation, but it is used in the setup script.

easybuild/easyblocks/c/comsol.py
akesandgren: bah, picky...

easybuild/easyblocks/c/comsol.py
akesandgren: Keep forgetting how long you want the lines :-)

easybuild/easyblocks/c/comsol.py
akesandgren: yes clearly. probably C&P bug

akesandgren: Or maybe not? intelbase.py also uses pathsep.join for the same purpose. And so are all the other blocks that uses find_flexlm_license. So I'll leave it as is to avoid breaking anything. path.join and join doesn't work exactly the same.

easybuild/easyblocks/c/comsol.py
akesandgren: fixed

easybuild/easyblocks/c/comsol.py
akesandgren: Done

easybuild/easyblocks/c/comsol.py
boegel: OK then, good enough.

easybuild/easyblocks/c/comsol.py
boegel: Ah, I see why, with `os.path.join` you would have to 'unpack' `lic_specs` since `os.path.join` requires multiple arguments rather than one single list argument, which is uglier:

```python
os.path.join(*lic_specs)
```

So do leave as is, indeed.

easybuild/easyblocks/c/comsol.py
boegel: 120 chars max. see http://easybuild.readthedocs.io/en/latest/Code_style.html

boegel:
@mboisson I assume that the easyconfig file you used for COMSOL 5.3a is as trivial as the one in https://github.com/easybuilders/easybuild-easyconfigs/pull/5390?

mboisson:
With the usual Compute Canada caveats for rpath and nix, yes. Here is our easyconfig: 

https://github.com/ComputeCanada/easybuild-easyconfigs/blob/computecanada-master/easybuild/easyconfigs/c/COMSOL/COMSOL-5.3a-dummy-dummy.eb

boegel:
@mboisson So do you just provide a dummy license during the installation (which is then removed via `postinstallcmds`?

mboisson:
I provide a user-provided license for installation, but it is not in the recipe, it is in the environment variables.

boegel:
I tested this with COMSOL 5.3a (cfr. https://github.com/easybuilders/easybuild-easyconfigs/pull/5882), works like a charm, so this is ready to go imho.

Thanks @akesandgren and @mboisson, and @Micket for the feedback!

PR #1320, opened 2017-12-12
Fix copy&paste error when adding run_cmd to scalapack.py

akesandgren:


boegel:
Duplicate of #1321.

@akesandgren Same problem applies to `verbose` (which is fixed in #1321 too).

Issue #1319, opened 2017-12-11
undefined variable in scalapack


olesenm:
In develop:

    File "/home/mol/Software/easybuild/easybuild-easyblocks/easybuild/easyblocks/s/scalapack.py", line 210, in build_step
    (out, _) = run_cmd(cmd, path=path, log_ok=False, log_all=True, simple=False, log_output=verbose)
    NameError: global name 'path' is not defined

https://github.com/easybuilders/easybuild-easyblocks/blob/5a3cd4918d18f131714431cacbab96f8caf16f92/easybuild/easyblocks/s/scalapack.py#L210

boegel:
Hmm, this slipped in with #1288.

The strange thing is that I haven't been hit by this, despite it being a clear problem and testing with that updated easyblock... 😕 

cc @akesandgren

akesandgren:
Should be fixed in PR #1320 

Made a C&P error when copying things from configuremake.py

boegel:
OK, now I am seeing it, I probably tested #1288 with an earlier version of the updated easyblock somehow, ugh. Fixed in #1321.

boegel:
Thanks for reporting this @olesenm, we shouldn't have missed this...

PR #1321, opened 2017-12-12
fix undefined variables used when running of parallel make in ScaLAPACK easyblock

boegel:
Fix for issue reported by @olesenm in #1319.
This somehow slipped in #1288, despite my testing (I probably accidentally tested with an earlier version of that updated ScaLAPACK easyblock somehow...).


boegel:
This is properly tested now, fixes the issue reported in #1319 (after reproducing it first without the fix in place).

boegel:
@damianam Frankly, so am I, I'm usually quite careful, but I messed up somehow...

damianam:
It happens to the best.

olesenm:
Good - thanks for the quick resolution.
Now I just need to figure out why the build (foss-2016b) is bailing out with `complex` error. (Oh I do hate getting stopped by fortran, when I don't even use it...)

PR #1337, opened 2018-01-05
BamTools 2.5.0 libraries can be located in either 'lib' or 'lib64'

boegel:
follow-up for #1332

cc @SethosII

### DIFF
--- a/easybuild/easyblocks/b/bamtools.py
+++ b/easybuild/easyblocks/b/bamtools.py
@@ -87,7 +87,8 @@ class EB_BamTools(MakeCp, CMakeMake):
                                           'lib/libjsoncpp.a'])
             custom_paths['dirs'].extend(['include/api', 'docs'])
         else:
-            custom_paths['files'].extend(['include/bamtools/shared/bamtools_global.h', 'lib64/libbamtools.a'])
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/b/bamtools.py
```diff
@@ -87,7 +87,8 @@ def sanity_check_step(self):
                                           'lib/libjsoncpp.a'])
             custom_paths['dirs'].extend(['include/api', 'docs'])
         else:
-            custom_paths['files'].extend(['include/bamtools/shared/bamtools_global.h', 'lib64/libbamtools.a'])
-            custom_paths['dirs'].extend(['include/bamtools/api', 'lib64/pkgconfig'])
+            custom_paths['files'].extend(['include/bamtools/shared/bamtools_global.h',
+                                          ('lib/libbamtools.a', 'lib64/libbamtools.a')])
+            custom_paths['dirs'].extend(['include/bamtools/api', ('lib.pkgconfig', 'lib64/pkgconfig')])
```
SethosII: @boegel Here is a typo. It should be `lib/pkgconfig` instead of `lib.pkgconfig`.

PR #1338, opened 2018-01-08
Fix quotes when using $ORIGIN in the RPATH

jdonners:
When compiling DOLFIN with RPATH support, the use of `$ORIGIN` in the RPATH should be preserved. However, the lack of quotes now causes expansion of `$ORIGIN`, which is basically empty when `patchelf` is called in the dolfin easyblock. Quotes are added to fix the RPATH support for DOLFIN.


boegel:
Looks trivial enough, so going in...

@jdonners Do you have easyconfigs for a recent version of DOLFIN? The ones we have in the central repository are quite old...

damianam:
I have some for `2017.1.0`, but they are tailored to our setup and not the most elegant ones to be honest. They work fine though.

ocaisa:
https://github.com/easybuilders/JSC/tree/master/Golden_Repo/2017b/d/DOLFIN

PR #1339, opened 2018-01-09
Added support for shared libs to trilinos easyblock

jdonners:
The sanity check only searched for static libraries previously.

### DIFF
--- a/easybuild/easyblocks/t/trilinos.py
+++ b/easybuild/easyblocks/t/trilinos.py
@@ -37,6 +37,7 @@ from easybuild.easyblocks.generic.cmakemake import CMakeMake
 from easybuild.framework.easyconfig import CUSTOM
 from easybuild.tools.build_log import EasyBuildError
 from easybuild.tools.modules import get_software_root
+from easybuild.tools.systemtools import get_shared_lib_ext
 
 
 class EB_Trilinos(CMakeMake):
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/t/trilinos.py
```diff
@@ -238,8 +239,14 @@ def sanity_check_step(self):
...
 
+        # Get the library extension
+        if self.cfg['shared_libs']:
+            lib_ext = "." + get_shared_lib_ext()
+        else:
+            lib_ext = ".a"
+
         custom_paths = {
-            'files': [os.path.join('lib', 'lib%s.a' % x.lower()) for x in libs],
+            'files': [os.path.join('lib', 'lib%s' % x.lower()+lib_ext) for x in libs],
```
boegel: @jdonners I'd slightly change this to:

```python
'files': [os.path.join('lib', 'lib%s.%s' % (x.lower(), lib_ext)) for x in libs],
```

easybuild/easyblocks/t/trilinos.py
```diff
@@ -238,8 +239,14 @@ def sanity_check_step(self):
             libs.remove('Kokkos')
             libs.append('kokkoscore')
 
+        # Get the library extension
+        if self.cfg['shared_libs']:
+            lib_ext = "." + get_shared_lib_ext()
+        else:
+            lib_ext = ".a"
```
boegel: A one-liner here could be:

```python
lib_ext = ('a', get_shared_lib_ext())[self.cfg['shared_libs']]
```

But that's probably a bit hard to read. ;-)

I'd exclude the `.` here, that makes the below easier to read:

```python
if self.cfg['shared_libs']:
    lib_ext = get_shared_lib_ext()
else:
    lib_ext = 'a'

PR #1341, opened 2018-01-11
support install_target in PythonPackage + deprecate use_easy_install & use_setup_py_develop

SethosII:
This makes the option passed to `setup.py` configurable. Therefore `use_setup_py_develop` and `use_easy_install` are obsolete. `install_target = 'develop'` and `install_target = 'easy_install'` can be used instead.

See also https://github.com/easybuilders/easybuild-easyconfigs/pull/5610#issuecomment-356604042.

### DIFF
--- a/easybuild/easyblocks/generic/pythonpackage.py
+++ b/easybuild/easyblocks/generic/pythonpackage.py
@@ -247,8 +247,13 @@ class PythonPackage(ExtensionEasyBlock):
             else:
                 self.install_cmd = SETUP_PY_INSTALL_CMD
 
+            if self.cfg['install_target'] == 'easy_install':
+                self.cfg.update('installopts', '--no-deps')
             if self.cfg.get('zipped_egg', False):
-                raise EasyBuildError("Installing zipped eggs requires using easy_install or pip")
+                if self.cfg['install_target'] == 'easy_install':
+                    self.cfg.update('installopts', '--zip-ok')
+                else:
+                    raise EasyBuildError("Installing zipped eggs requires using easy_install or pip")
 
         self.log.debug("Using '%s' as install command", self.install_cmd)
 
--- a/easybuild/easyblocks/generic/pythonpackage.py
+++ b/easybuild/easyblocks/generic/pythonpackage.py
@@ -51,7 +51,8 @@ from easybuild.tools.run import run_cmd
 
 # not 'easy_install' deliberately, to avoid that pkg installations listed in easy-install.pth get preference
 # '.' is required at the end when using easy_install/pip in unpacked source dir
-EASY_INSTALL_INSTALL_CMD = "%(python)s setup.py easy_install --prefix=%(prefix)s %(installopts)s %(loc)s"
+EASY_INSTALL_TARGET = "easy_install"
+EASY_INSTALL_INSTALL_CMD = "%(python)s setup.py " + EASY_INSTALL_TARGET + " --prefix=%(prefix)s %(installopts)s %(loc)s"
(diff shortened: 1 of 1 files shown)


boegel:
@SethosII Can you add a clear deprecation log message when `use_setup_py_develop` or `use_easy_install` are used?

Something like:

```python
self.log.deprecated("Use 'install_target' rather than 'use_setup_py_develop'.", '4.0')
```

The `4.0` here indicates that this will result in an error with EasyBuild 4.0 (until then, it will result in a deprecation message).

Ideally, we should also update easyconfigs in the central repo that use either `use_setup_py_develop` or `use_easy_install` to use `install_target` instead, i.e. these:

* `tvb-data`
* `tvb-framework`
* `tvb-library`
* `hanythingondemand`
* `vsc-base-2.4.2.eb`, `vsc-base-2.5.1.eb`, `vsc-base-2.5.8.eb`
* `vsc-mympirun-3.4.2-vsc-base-2.4.2.eb`

SethosII:
@boegel  I added a deprecation message and also another commit to add the functionality from `use_easy_install` when using `install_target = 'easy_install'` (add `--no-deps` to `install_opts` (https://github.com/easybuilders/easybuild-easyblocks/blob/master/easybuild/easyblocks/generic/pythonpackage.py#L222) and allow `zipped_egg` (https://github.com/easybuilders/easybuild-easyblocks/blob/master/easybuild/easyblocks/generic/pythonpackage.py#L225 and https://github.com/easybuilders/easybuild-easyblocks/blob/master/easybuild/easyblocks/generic/pythonpackage.py#L248). I created a pull request to use `install_target` in existing EasyConfigs: https://github.com/easybuilders/easybuild-easyconfigs/pull/5625.

easybuild/easyblocks/generic/pythonpackage.py
```diff
@@ -186,6 +186,7 @@ def extra_options(extra_vars=None):
             'use_setup_py_develop': [False, "Install using '%s'" % SETUP_PY_DEVELOP_CMD, CUSTOM],
```
boegel: please add ` (deprecated)` to the help message, same for `use_easy_install`

easybuild/easyblocks/generic/pythonpackage.py
```diff
@@ -240,12 +242,18 @@ def __init__(self, *args, **kwargs):
             self.use_setup_py = True
 
             if self.cfg.get('use_setup_py_develop', False):
+                self.log.deprecated("Use 'install_target' rather than 'use_setup_py_develop'.", '4.0')
                 self.install_cmd = SETUP_PY_DEVELOP_CMD
             else:
                 self.install_cmd = SETUP_PY_INSTALL_CMD
 
+            if self.cfg['install_target'] == 'easy_install':
```
boegel: introduce a `EASY_INSTALL_TARGET` constant on top, and use that (also use it to define `EASY_INSTALL_INSTALL_CMD`)

easybuild/easyblocks/generic/pythonpackage.py
```diff
@@ -240,12 +242,18 @@ def __init__(self, *args, **kwargs):
...
+                self.cfg.update('installopts', '--no-deps')
             if self.cfg.get('zipped_egg', False):
-                raise EasyBuildError("Installing zipped eggs requires using easy_install or pip")
+                if self.cfg['install_target'] == 'easy_install':
```
boegel: use `EASY_INSTALL_TARGET` here too

SethosII:
@boegel I added the requested changes.

boegel:
Thanks a lot for your effort @SethosII!

PR #1343, opened 2018-01-15
Explicitly allow static builds to be disabled and add support for non standard zlib and bzip2 in Amber easyblock

valtandor:
…standard zlib and bzip2

### DIFF
--- a/easybuild/easyblocks/a/amber.py
+++ b/easybuild/easyblocks/a/amber.py
@@ -1,6 +1,7 @@
 ##
 # Copyright 2009-2018 Ghent University
 # Copyright 2015-2018 Stanford University
+# Copyright 2018 Landcare Research New Zealand Ltd
 #
 # This file is part of EasyBuild,
 # originally created by the HPC team of Ghent University (http://ugent.be/hpc/en),
@@ -59,6 +60,7 @@ class EB_Amber(ConfigureMake):
             'patchruns': [1, "Number of times to run Amber's update script before building", CUSTOM],
             # enable testing by default
             'runtest': [True, "Run tests after each build", CUSTOM],
+            'static': [True, "Build statically linked executables", CUSTOM],
         })
         return ConfigureMake.extra_options(extra_vars)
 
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/a/amber.py
```diff
@@ -141,9 +143,17 @@ def install_step(self):
                 self.mpi_option = '-mpi'
 
         common_configopts = [self.cfg['configopts'], '--no-updates', '-noX11']
-        if self.name == 'Amber':
+        if self.name == 'Amber' and self.cfg['static']:
             common_configopts.append('-static')
 
+        envval = lambda ev: os.getenv(ev) or ''
```
boegel: @valtandor You don't need this, you can use `os.getenv('AMBERBUILDFLAGS', '')` instead

easybuild/easyblocks/a/amber.py
```diff
@@ -141,9 +143,17 @@ def install_step(self):
...
+        zlibroot = get_software_root('zlib')
+        if zlibroot:
+            env.setvar('AMBERBUILDFLAGS', envval('AMBERBUILDFLAGS') + " -L" + os.path.join(zlibroot, "lib"))
+        bzip2root = get_software_root('bzip2')
+        if bzip2root:
+            env.setvar('AMBERBUILDFLAGS', envval('AMBERBUILDFLAGS') + " -L" + os.path.join(bzip2root, "lib"))
```
boegel: You can simplify this a little bit by making it a loop:

```python
for dep in ['bzip2', 'zlib']:
    root = get_software_root(dep)
    amberbuildflags = os.getenv('AMBERBUILDFLAGS', '')
    env.setvar('AMBERBUILDFLAGS', amberbuildflags + " -L" + os.path.join(root, 'lib'))

akesandgren:
The "static" flag functionality has been merged in PR #1468 (copied from this PR)
There should be no need to touch AMBERBUILDFLAGS, bzip2 and zlib will be picked up from LD_LIBRARY_PATH and when having the Amber module loaded cpptraj and ambpdb will use the correct search path.

boegel:
This has bitrotted, and it seems like the relevant stuff has been merged already via #1468, so closing...

PR #1347, opened 2018-01-17
Fix Tkinter so it works with both Python 2 and 3.

schiotz:
The current Tkinter easyblock only works for Python 2, as the names of some of the files that should be installed are different in Python 3.

This module is needed since Tkinter support has been removed from the default Python installations.


boegel:
Looks great, tested with both existing `Tkinter-2.7.13-intel-2017a-Python-2.7.13.eb` and the new `Tkinter` easyconfigs from https://github.com/easybuilders/easybuild-easyconfigs/pull/5658, so going in. Thanks @schiotz!

PR #1353, opened 2018-02-01
Add new easyblock for Theano

smoors:
include random string in compiledir_format and make sure theano uses BLAS libraries

### DIFF
--- a/easybuild/easyblocks/t/theano.py
+++ b/easybuild/easyblocks/t/theano.py
@@ -1,0 +1,21 @@
+from easybuild.easyblocks.generic.pythonpackage import PythonPackage
+import random
+import string
+
+
+class EB_Theano(PythonPackage):
+    """ include a random string in compiledir_format to fix problems with architecture-specific compilation
+    when running the software on a heterogeneous compute cluster. """
+
+    def make_module_extra(self):
+
+        txt = super(EB_Theano, self).make_module_extra()
+
+        rand_string = ''.join(random.choice(string.letters) for i in range(10))
+
+        theano_flags = ('compiledir_format=compiledir_%s-%%(short_platform)s-%%(processor)s-'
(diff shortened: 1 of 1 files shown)


smoors:
adding BLAS libraries is recommended for performance: if you don't, Theano will show a warning like this:
fix warning: WARNING (theano.tensor.blas): Using NumPy C-API based implementation for BLAS functions.

boegel:
@smoors Companion PR to update `Theano` easyconfigs to leverage this is still missing?

Also, please customize `sanity_check_step` as well.

smoors:
@boegel I was actually waiting until this easyblock is accepted :)
what do want to customize in `sanity_check_step` exactly? the updated compilation directory is created only when the user starts using the module.

boegel:
@smoors the sanity check should be customized to check for files & directories that should be there when Theano is correctly installed (rather than relying in the generic `bin` and `lib` or `lib64` check), it's not related to the compilation directory at all (sorry for not being more specific there), basically what we already have in the easyconfigs (but then letting the easyblock handle it):

```python
sanity_check_paths = {
    'files': ['bin/theano-cache', 'bin/theano-nose'],
    'dirs': ['lib/python%(pyshortver)s/site-packages'],
}
```

Having a PR for updated Theano easyconfigs already is useful so we can easily test this easyblock (we can't merge it without testing it with all existing Theano easyconfigs).

smoors:
@boegel see #6378

damianam:
Given that Theano is practically abandonware, does it make sense to merge this now? It seems simple enough to implement that in the existing easyconfigs?

smoors:
@damianam why do you say it is abandonware? they still have regular releases

boegel:
@smoors https://groups.google.com/forum/#!topic/theano-users/7Poq8BZutbY

smoors:
ok its clear that support will not continue much longer, will close this.

PR #1361, opened 2018-02-20
fix icc wrapper used in TensorFlow easyblock to avoid other wrappers from calling it

boegel:
…


wpoely86:
Tested https://github.com/easybuilders/easybuild-easyconfigs/pull/5499 with this.

PR #1362, opened 2018-02-20
don't use PERL_MM_OPT and PERL_MB_OPT

RvDijk:
Something like this should do the trick, however I was not able to test anything. Using e.g. `eb ~/Perl-5.26.1-test-foss-2018a.eb --include-easyblocks=./perlmodule.py,../p/perl.py ` did not use my custom `perlmodule.py`. It would be helpful to know how I can use a custom easyblock such as perlmodule.py for testing.

### DIFF
--- a/easybuild/easyblocks/generic/perlmodule.py
+++ b/easybuild/easyblocks/generic/perlmodule.py
@@ -58,9 +58,8 @@ class PerlModule(ExtensionEasyBlock, ConfigureMake):
         """Install procedure for Perl modules: using either Makefile.Pl or Build.PL."""
         # Perl modules have two possible installation procedures: using Makefile.PL and Build.PL
         # configure, build, test, install
-        unset_cmd = "unset PERL_MM_OPT PERL_MB_OPT"
         if os.path.exists('Makefile.PL'):
-            run_cmd('%s %s perl Makefile.PL PREFIX=%s %s' % (unset_cmd, self.cfg['preconfigopts'], self.installdir, self.cfg['configopts']))
+            run_cmd('%s perl Makefile.PL PREFIX=%s %s' % (self.cfg['preconfigopts'], self.installdir, self.cfg['configopts']))
             ConfigureMake.build_step(self)
             ConfigureMake.test_step(self)
             ConfigureMake.install_step(self)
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/generic/perlmodule.py
```diff
@@ -58,16 +58,17 @@ def install_perl_module(self):
         """Install procedure for Perl modules: using either Makefile.Pl or Build.PL."""
         # Perl modules have two possible installation procedures: using Makefile.PL and Build.PL
         # configure, build, test, install
+        unset_cmd = "unset PERL_MM_OPT PERL_MB_OPT"
```
boegel: You'll probably need to make this `unset ... && ` to chain it with the Perl command? `preconfigopts` complicates this though...

boegel: A better approach would be to use `unset_env_vars(['PERL_MB_OPT', 'PERL_MM_OPT'])`, where `unset_env_vars` can be imported from `easybuild.tools.environment`.

This will result in proper logging for example, and since the environment in which commands are run via `run_cmd` is inherited, that should work.

easybuild/easyblocks/generic/perlmodule.py
```diff
@@ -58,16 +58,17 @@ def install_perl_module(self):
...
         if os.path.exists('Makefile.PL'):
-            run_cmd('%s perl Makefile.PL PREFIX=%s %s' % (self.cfg['preconfigopts'], self.installdir, self.cfg['configopts']))
+            run_cmd('%s %s perl Makefile.PL PREFIX=%s %s' % (unset_cmd, self.cfg['preconfigopts'], self.installdir, self.cfg['configopts']))
```
boegel: Line becomes too long, and this won't work especially in combination with `preconfigopts`, but you can just revert this change if you use `unset_env_vars` instead.

easybuild/easyblocks/generic/perlmodule.py
boegel: Do include a comment to explain why these environment variables need to be unset.

boegel:
@RvDijk To test with a custom `perlmodule.py`, you currently need to place it in a subdirectory named `generic`, and specify its location via `--include-easyblocks=/path/to/generic/perlmodule.py`, see https://github.com/easybuilders/easybuild-framework/issues/1964 .

RvDijk:
@boegel I agree with you a lot, I'll look into using `unset_env_vars`. I already wanted to do something like that but was not able to find how to change environment vars within these Python scripts. I think `unset_env_vars` is one of the best solutions.

RvDijk:
@boegel I was able to build a simplified version of Perl, where one extension was used. The build succeeded succesfully, where the previous version of perlmodule.py would have failed.

easybuild/easyblocks/generic/perlmodule.py
```diff
@@ -53,6 +53,10 @@ def __init__(self, *args, **kwargs):
         """Initialize custom class variables."""
         super(PerlModule, self).__init__(*args, **kwargs)
         self.testcmd = None
+        # Environment variables PERL_MM_OPT and PERL_MB_OPT cause installations to fail.
```
boegel: @RvDijk Let's move the extra line you added below this block to above this block? :)

Also, drop the `WIP` from the PR title, it looks good to go.

boegel:
Thanks @RvDijk!

PR #1363, opened 2018-02-21
added easyblock for oasyslsdyna

mboisson:
Work in progress, but I would like comments anyway.


mboisson:
No, I have not had time to come back to this, especially changing to use run_qmd_qa. The other changes are trivial, but this one requires rerunning the recipe and fiddling with the right parameters to use.

PR #1365, opened 2018-02-22
Add make-makeinstall easyblock without configure

rubendibattista:
Did not find a better way without modifying `ConfigureMake`. Maybe adding a `bool` option to make the configure step optional?

(This module needed for Lua install. Check new PR for `Lua`)


boegel:
@rubendibattista You're looking for `skipsteps`, you can easily skip the `configure` step of `ConfigureMake` using this in your easyconfig file:

```python
skipsteps = ['configure']
```

No need for a custom easyblock for this...

rubendibattista:
Ok. Sorry for wasting time. :)

PR #1366, opened 2018-02-23
added option use_pip_for_dependencies for pythonpackages

pescobar:
with this option a easyconfig like this one can be installed. The only change is `use_pip_for_dependencies = True`

I know this is not the recommended approach and it doesn't help to reproduce the same installation in the future but in this specific example deeptools requires 41 different python libraries so creating a proper easyconfig with all the required libraries require quite some manual work. And this is a tool that probably only 1 or 2 of my users will need so it's not worth the time IMHO and for me it's ok to take the risk of not writting a fully complete easyconfig with the complete list of deps and rely on pip for the installation but save a lot of time of manual work

```
easyblock = 'PythonPackage'

name = 'deepTools'
version = '3.0.0'
versionsuffix = '-Python-%(pyver)s'

homepage = 'https://github.com/fidelram/deepTools'
description = """Tools to process and analyze deep sequencing data"""

toolchain = {'name': 'goolf', 'version': '1.7.20'}

source_urls = ['https://github.com/fidelram/deepTools/archive/']
sources = ['%(version)s.tar.gz']

dependencies = [
    ('Python', '2.7.11'),
    ('cURL', '7.40.0', '', True),
]

use_pip = True
use_pip_for_dependencies = True
unpack_sources = False

sanity_check_paths = {
    'files': [],
    'dirs': ['lib/python%(pyshortver)s/site-packages/deeptools'],
}

moduleclass = 'bio'
```

### DIFF
--- a/easybuild/easyblocks/generic/pythonpackage.py
+++ b/easybuild/easyblocks/generic/pythonpackage.py
@@ -234,7 +234,10 @@ class PythonPackage(ExtensionEasyBlock):
             # don't auto-install dependencies with pip unless use_pip_for_dependencies=True
             # the default is use_pip_for_dependencies=False
             if self.cfg.get('use_pip_for_dependencies') == False:
+                self.log.info("Using pip with --no-deps option")
                 self.cfg.update('installopts', '--no-deps')
+            else:
+                self.log.info("Using pip to also install the dependencies")
 
             # don't (try to) uninstall already availale versions of the package being installed
             self.cfg.update('installopts', '--ignore-installed')
--- a/easybuild/easyblocks/generic/pythonpackage.py
+++ b/easybuild/easyblocks/generic/pythonpackage.py
@@ -184,6 +184,7 @@ class PythonPackage(ExtensionEasyBlock):
             'runtest': [True, "Run unit tests.", CUSTOM],  # overrides default
             'use_easy_install': [False, "Install using '%s' (deprecated)" % EASY_INSTALL_INSTALL_CMD, CUSTOM],
             'use_pip': [False, "Install using '%s'" % PIP_INSTALL_CMD, CUSTOM],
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/generic/pythonpackage.py
```diff
@@ -230,8 +231,13 @@ def __init__(self, *args, **kwargs):
...
 
-            # don't auto-install dependencies
-            self.cfg.update('installopts', '--no-deps')
+            # don't auto-install dependencies with pip unless use_pip_for_dependencies=True
+            # the default is use_pip_for_dependencies=False
+            if self.cfg.get('use_pip_for_dependencies') == False:
+                self.log.info("Using pip with --no-deps option")
+                self.cfg.update('installopts', '--no-deps')
+            else:
+                self.log.info("Using pip to also install the dependencies")
```
boegel: Checking `== False` is a bit weird @pescobar, since it's a boolean... Should be:

```python
# don't auto-install dependencies with pip, unless use_pip_for_deps is enabled
if self.cfg.get('use_pip_for_deps'):
    self.log.info("Using pip to also install the dependencies")
else:
    self.log.info("Using pip with --no-deps option")
    self.cfg.update('installopts', '--no-deps')

File: easybuild/easyblocks/generic/pythonpackage.py [hunk 2]
```diff
@@ -184,6 +184,7 @@ def extra_options(extra_vars=None):
             'runtest': [True, "Run unit tests.", CUSTOM],  # overrides default
             'use_easy_install': [False, "Install using '%s' (deprecated)" % EASY_INSTALL_INSTALL_CMD, CUSTOM],
             'use_pip': [False, "Install using '%s'" % PIP_INSTALL_CMD, CUSTOM],
+            'use_pip_for_dependencies': [False, "Install dependencies using '%s'" % PIP_INSTALL_CMD, CUSTOM],
```
boegel: @pescobar I would rename this to `use_pip_for_deps`?

PR #1367, opened 2018-02-26
make CUDA easyblock aware of preinstallopts

vanzod:
The `CUDA` easyblock was missing support for `preinstallopts`.
This is necessary to be able to define `PERL5LIB` if needed in order for the installer script to correctly locate `InstallUtils.pm`.

### DIFF
--- a/easybuild/easyblocks/c/cuda.py
+++ b/easybuild/easyblocks/c/cuda.py
@@ -67,7 +67,6 @@ class EB_CUDA(Binary):
         # script has /usr/bin/perl hardcoded, but we want to have control over which perl is being used
         if LooseVersion(self.version) <= LooseVersion("5"):
             install_script = "install-linux.pl"
-            install_script_path = os.path.join(self.builddir, install_script)
             cmd = "%(preinstallopts)s perl ./%(script)s --prefix=%(installdir)s %(installopts)s" % {
                 'preinstallopts': self.cfg['preinstallopts'],
                 'script': install_script,
--- a/easybuild/easyblocks/c/cuda.py
+++ b/easybuild/easyblocks/c/cuda.py
@@ -67,23 +67,17 @@ class EB_CUDA(Binary):
         # script has /usr/bin/perl hardcoded, but we want to have control over which perl is being used
         if LooseVersion(self.version) <= LooseVersion("5"):
             install_script = "install-linux.pl"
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/c/cuda.py
```diff
@@ -68,13 +68,23 @@ def install_step(self):
         if LooseVersion(self.version) <= LooseVersion("5"):
             install_script = "install-linux.pl"
             install_script_path = os.path.join(self.builddir, install_script)
```
boegel: @vanzod As far as I can tell, `install_script_path` is totally unused in the easyblock, so please remove it

easybuild/easyblocks/c/cuda.py
```diff
@@ -68,13 +68,23 @@ def install_step(self):
...
             # installparams = "-samplespath=%(x)s/samples/ -toolkitpath=%(x)s -samples -toolkit" % {'x': self.installdir}
             install_script = "cuda-installer.pl"
             installparams = "-toolkitpath=%s -toolkit" % self.installdir
-            cmd = "perl ./%s -verbose -silent %s %s" % (install_script, installparams, self.cfg['installopts'])
+            cmd = "%(preinstallopts)s perl ./%(script)s -verbose -silent %(installparams)s %(installopts)s" % {
+                'preinstallopts': self.cfg['preinstallopts'],
+                'script': install_script,
+                'installparams': installparams,
+                'installopts': self.cfg['installopts']
+            }
```
boegel: Hmm, this is becoming very copy-pasty, I think we can do better:

```python
if LooseVersion(self.version) <= LooseVersion("5"):
    install_script = "install-linux.pl"
    self.cfg.update('installopts', '--prefix=%s' % self.installdir)
else:
    install_script = "cuda-installer.pl"
    # note: also including samples (via "-samplespath=%(installdir)s -samples") would require libglut
    self.cfg.update('installopts', "-verbose -script -toolkitpath=%s -toolkit" % self.installdir)

cmd = "%(preinstallopts)s perl ./%(script)s %(installopts)s" % {
    'preinstallopts': self.cfg['preinstallopts'],
    'script': install_script,
    'installopts': self.cfg['installopts'],
}

boegel:
lgtm, tested with existing `CUDA` easyconfigs, thanks @vanzod!

PR #1369, opened 2018-02-27
add support for 'default_component_specs' easyconfig parameter in Bundle easyblock + fix handling of per-component (checksums for) patches

boegel:
* add support for `default_component_specs` easyconfig parameter
* fix applying of per-component patches (they were silently ignored)
* fix handling of checksums for per-component patches (were incorrectly mixed with checksums for sources)


boegel:
tested with X11 and FALCON easyconfigs (the only easyconfigs we have that use the `components` easyconfig parameter supported by `Bundle`), so good to go? @wpoely86

PR #1377, opened 2018-03-06
add support for detecting auto-downloaded dependencies in PythonPackage easyblock

boegel:
edit: https://github.com/easybuilders/easybuild-framework/pull/2493 + depends on https://github.com/easybuilders/easybuild-framework/pull/2494

### DIFF
--- a/easybuild/easyblocks/generic/pythonpackage.py
+++ b/easybuild/easyblocks/generic/pythonpackage.py
@@ -178,17 +178,18 @@ class PythonPackage(ExtensionEasyBlock):
         if extra_vars is None:
             extra_vars = {}
         extra_vars.update({
-            'unpack_sources': [True, "Unpack sources prior to build/install", CUSTOM],
+            'buildcmd': ['build', "Command to pass to setup.py to build the extension", CUSTOM],
+            'download_dep_fail': [False, "Fail if downloaded dependencies are detected", CUSTOM],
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/generic/pythonpackage.py
```diff
@@ -178,17 +178,18 @@ def extra_options(extra_vars=None):
         if extra_vars is None:
             extra_vars = {}
         extra_vars.update({
-            'unpack_sources': [True, "Unpack sources prior to build/install", CUSTOM],
+            'buildcmd': ['build', "Command to pass to setup.py to build the extension", CUSTOM],
+            'download_dep_fail': [False, "Fail if downloaded dependencies are detected", CUSTOM],
```
boegel: Set to `False` for backward-compatibility, but we should consider switching this to `True` by default for EasyBuild v4.0.0.

Problem is that we'll need to fix quite a bit of easyconfigs...

We could try to enforce a policy where new easyconfigs for Python packages include `download_dep_fail = True`?

boegel:
@wpoely86 please review?

boegel:
@akesandgren suggestion is now implemented, but it depends on two changes in EasyBuild framework (see updated description)

boegel:
The framework PRs on which this depends have been merged, so this is good to go?

vanzod:
My tests did not show any issue. Merging

PR #1378, opened 2018-03-06
enhance Gurobi easyblock to support installing Python bindings

boegel:


easybuild/easyblocks/g/gurobi.py
```diff
@@ -48,17 +51,29 @@ def install_step(self):
 
         copy_file(self.cfg['license_file'], os.path.join(self.installdir, 'gurobi.lic'))
 
+        if get_software_root('Python'):
+            run_cmd("python setup.py install --prefix=%s" % self.installdir)
```
wpoely86: any reason why we cannot use the existing python extension stuff?

boegel: It's just not worth the trouble here...

The `PythonPackage` easyblock does a lot more than what we need here, and we do leverage what we can use from there (see `det_pylibdir`).

This is not a regular `setup.py` script (note that I don't need to fiddle with `$PYTHONPATH` for example), it has specific logic for installing the Python bindings in the right place.

It would be significantly more complicated to leverage `PythonPackage` here, simply not worth the trouble in this particular case imho.

PR #1385, opened 2018-03-08
Added $EBROOTIFORT/include in CPATH

damianam:
This enables the creation of fortran 2008 bindings in MPI, for MPICH flavored MPIs

### DIFF
--- a/easybuild/easyblocks/i/ifort.py
+++ b/easybuild/easyblocks/i/ifort.py
@@ -81,9 +81,9 @@ class EB_ifort(EB_icc, IntelBase):
         """
         Additional paths to consider for prepend-paths statements in module file
         """
-        guesses = super(EB_icc, self).make_module_req_guess()
+        guesses = super(EB_ifort, self).make_module_req_guess()
         if LooseVersion(self.version) >= LooseVersion('2016'):
             # This enables the creation of fortran 2008 bindings in MPI
-            guesses['CPATH'].extend(['include'])
+            guesses['CPATH'].append('include')
 
         return guesses
--- a/easybuild/easyblocks/i/ifort.py
+++ b/easybuild/easyblocks/i/ifort.py
@@ -76,3 +76,14 @@ class EB_ifort(EB_icc, IntelBase):
         custom_commands = ["which ifort"]
 
         IntelBase.sanity_check_step(self, custom_paths=custom_paths, custom_commands=custom_commands)
+
+    def make_module_req_guess(self):
+        """
+        Additional paths to consider for prepend-paths statements in module file
+        """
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/i/ifort.py
```diff
@@ -76,3 +76,14 @@ def sanity_check_step(self):
...
         IntelBase.sanity_check_step(self, custom_paths=custom_paths, custom_commands=custom_commands)
+
+    def make_module_req_guess(self):
+        """
+        Additional paths to consider for prepend-paths statements in module file
+        """
+        guesses = super(EB_icc, self).make_module_req_guess()
+        if LooseVersion(self.version) >= LooseVersion('2016'):
+            # This enables the creation of fortran 2008 bindings in MPI
+            guesses['CPATH'].extend(['include'])
```
boegel: @damianam: Would it make sense to just add `include` in the initial `CPATH` list in `EB_icc.make_module_req_guess` instead?

`<prefix>/include` will only be added to `$CPATH` if it exists, so it's OK if it's also considered for `icc`

Also, style nitpicking: to add a single element, you should use `append(x)` rather than `extend([x])` ;-)

damianam: No, in `$EBROOTICC/include` there are some other headers, like `complex.h`, that apparently cause problems. Besides that, there is something fishy, the resulting module now is missing some paths in other variables. I'll try to take a look at it today.

boegel: OK, thanks for clarifying, makes sense to only do this for `ifort` then.

I see no reason why other paths would suddenly go missing though...

easybuild/easyblocks/i/ifort.py
```diff
@@ -76,3 +76,14 @@ def sanity_check_step(self):
         custom_commands = ["which ifort"]
 
         IntelBase.sanity_check_step(self, custom_paths=custom_paths, custom_commands=custom_commands)
+
+    def make_module_req_guess(self):
+        """
+        Additional paths to consider for prepend-paths statements in module file
+        """
+        guesses = super(EB_icc, self).make_module_req_guess()
+        if LooseVersion(self.version) >= LooseVersion('2016'):
```
boegel: @damianam Why exactly is this version check here? If the `include` dir was not there for older versions, don't bother, the existence of the directory is checked before the path is added to `$CPATH`

damianam: It existed, but the headers needed for fortran 2008 were not there, so there is no point in adding that path, I guess.

boegel:
Thanks @damianam!

PR #1386, opened 2018-03-14
TensorFlow 1.6.0 fix: use the absolute path for the C compiler when compiling with GPU support

smoors:
This is needed for the `bazel build` command with the option `--config=cuda`
Inspiration for this fix was found here:
https://github.com/tensorflow/tensorflow/issues/14380

### DIFF
--- a/easybuild/easyblocks/t/tensorflow.py
+++ b/easybuild/easyblocks/t/tensorflow.py
@@ -125,7 +125,7 @@ class EB_TensorFlow(PythonPackage):
         if cuda_root:
             config_env_vars.update({
                 'CUDA_TOOLKIT_PATH': cuda_root,
-                'GCC_HOST_COMPILER_PATH': which(os.getenv('CC')),
+                'GCC_HOST_COMPILER_PATH': os.path.realpath(which(os.getenv('CC'))),
                 'TF_CUDA_COMPUTE_CAPABILITIES': ','.join(self.cfg['cuda_compute_capabilities']),
                 'TF_CUDA_VERSION': get_software_version('CUDA'),
             })
--- a/easybuild/easyblocks/t/tensorflow.py
+++ b/easybuild/easyblocks/t/tensorflow.py
@@ -125,7 +125,7 @@ class EB_TensorFlow(PythonPackage):
         if cuda_root:
             config_env_vars.update({
                 'CUDA_TOOLKIT_PATH': cuda_root,
-                'GCC_HOST_COMPILER_PATH': os.path.realpath(which(os.getenv('CC'))),
+                'GCC_HOST_COMPILER_PATH': which(os.getenv('CC')),
                 'TF_CUDA_COMPUTE_CAPABILITIES': ','.join(self.cfg['cuda_compute_capabilities']),
                 'TF_CUDA_VERSION': get_software_version('CUDA'),
             })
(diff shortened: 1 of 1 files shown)


boegel:
Hmm, I think this reverts something that @akesandgren did as a part of the sprint during the EasyBuild User Meeting in Amsterdam (see commit history in #1287)...

Why didn't this pop up earlier? Is this specific to TF 1.6.0?

boegel:
Also, what happens when `$CC` actually points to a wrapper for the compiler, for example when `--use-ccache` or `--rpath` is used?

akesandgren:
It doesn't touch the things i added as far as i can see.

I don't really see how doing this change would make a difference.

What is the actual problem or error output from not doing this?

smoors:
@akesandgren this is the error I get.
it complains about some missing header files, but they are present.
however, the /apps/brussel dir is a symlink.
according to a commenter on the link I posted above, "It usually boils down to crosstool referencing files via symlink, while compilers resolve the symlinks when outputting the .d files and vice versa."
this fix was not necessary with TensorFlow 1.5.0. it could be due to a different Bazel version (0.11.0 vs. 0.7.0).
this is all a great mystery to me but as Ward said, "with Tensorflow, you can expect anything".

ERROR: /tmp/smoors/eb-8JIy9O/tmp0ArazE-bazel-build/external/flatbuffers/BUILD:52:1: undeclared inclusion(s) in rule '@flatbuffers//:flatc_library':
this rule is missing dependency declarations for the following files included by 'external/flatbuffers/src/util.cpp':
  '/apps/brussel/CO7/ivybridge-ib/software/GCCcore/6.4.0/lib/gcc/x86_64-pc-linux-gnu/6.4.0/include/stdarg.h'
  '/apps/brussel/CO7/ivybridge-ib/software/GCCcore/6.4.0/lib/gcc/x86_64-pc-linux-gnu/6.4.0/include/stddef.h'
  '/apps/brussel/CO7/ivybridge-ib/software/GCCcore/6.4.0/lib/gcc/x86_64-pc-linux-gnu/6.4.0/include/stdint.h'
  '/apps/brussel/CO7/ivybridge-ib/software/GCCcore/6.4.0/lib/gcc/x86_64-pc-linux-gnu/6.4.0/include-fixed/limits.h'
  '/apps/brussel/CO7/ivybridge-ib/software/GCCcore/6.4.0/lib/gcc/x86_64-pc-linux-gnu/6.4.0/include-fixed/syslimits.h'^M
Target //tensorflow/tools/pip_package:build_pip_package failed to build

smoors:
@boegel  concerning compiler wrappers, it would indeed be good if someone tested that.

akesandgren:
There might be another way to solve this, but i have to look at that later today.

boegel:
@smoors I ran into similar issues while developing the TensorFlow easyblock, but I didn't see these problems anymore with the easyblock we ended up with in #1287...

easybuild/easyblocks/t/tensorflow.py
```diff
@@ -125,7 +125,7 @@ def configure_step(self):
         if cuda_root:
             config_env_vars.update({
                 'CUDA_TOOLKIT_PATH': cuda_root,
-                'GCC_HOST_COMPILER_PATH': which(os.getenv('CC')),
+                'GCC_HOST_COMPILER_PATH': os.path.realpath(which(os.getenv('CC'))),
```
boegel: @smoors Please use the `resolve_path` function from `filetools` (which has error handling)

akesandgren:
Doing a build of 1.6.0 with goolfc, will report back.

boegel:
@akesandgren You should probably make that `cxx_inc_dir_lines + '\n' + cxx_inc_dir_lines_no_resolv_path` to avoid glueing the last & first entry together?

`'\n'.join` will only add newlines in between element, not at the ned

smoors:
@akesandgren ok, will test.
do you think it is really necessary to have both the "resolved" and "unresolved" paths?

akesandgren:
@boegel forgot that (again...)
@smoors Yes it is. The resolved paths was what we added to get 1.5.0 building. Apparently something changed so we now need both.

smoors:
the solution of @akesandgren seems to work and is probably safer wrt wrappers, so let's use it.

boegel:
Thanks @smoors and @akesandgren!

PR #1392, opened 2018-03-28
imkl: fix sanity check overwriting base libs with interface libs.

bartoldeman:


boegel:
Testing revealed no problems, so good to go, thanks @bartoldeman!

PR #1393, opened 2018-03-28
Make cdft lib compilation optional for Intel MKL, by detecting MPI availability

bartoldeman:
This allows installing MKL using a compiler-only toolchain with interfaces enabled.

### DIFF
--- a/easybuild/easyblocks/i/imkl.py
+++ b/easybuild/easyblocks/i/imkl.py
@@ -77,6 +77,8 @@ class EB_imkl(IntelBase):
             super(EB_imkl, self).prepare_step(*args, **kwargs)
         else:
             super(EB_imkl, self).prepare_step(*args, **kwargs)
+
+        # build the mkl interfaces, if desired
         if self.cfg['interfaces']:
             cdftlibs = ['fftw2x_cdft']
             if LooseVersion(self.version) >= LooseVersion('10.3'):
@@ -92,7 +94,7 @@ class EB_imkl(IntelBase):
                 }
                 mpi_fam = self.toolchain.mpi_family()
                 mpi_spec = mpi_spec_by_fam.get(mpi_fam)
-                self.log.debug("Determined MPI specification based on MPI toolchain component: %s" % mpi_spec)
+                self.log.debug("Determined MPI specification based on MPI toolchain component: %s", mpi_spec)
             else:
                 # can't use toolchain.mpi_family, because of dummy toolchain
                 if get_software_root('MPICH2') or get_software_root('MVAPICH2'):
@@ -102,7 +104,7 @@ class EB_imkl(IntelBase):
                 elif not get_software_root('impi'):
                     # no compatible MPI found: do not build cdft
                     cdftlibs = []
-                self.log.debug("Determined MPI specification based on loaded MPI module: %s" % mpi_spec)
+                self.log.debug("Determined MPI specification based on loaded MPI module: %s", mpi_spec)
             self.cdftlibs = cdftlibs
             self.mpi_spec = mpi_spec
 
@@ -246,8 +248,6 @@ class EB_imkl(IntelBase):
             # blas95 and lapack also need include/.mod to be processed
             fftw2libs = ['fftw2xc', 'fftw2xf']
             fftw3libs = ['fftw3xc', 'fftw3xf']
-            # cdftlibs set in prepare_step
-            cdftlibs = self.cdftlibs
 
             interfacedir = os.path.join(self.installdir, intsubdir)
             try:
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/i/imkl.py
```diff
@@ -218,9 +246,8 @@ def post_install_step(self):
             # blas95 and lapack also need include/.mod to be processed
             fftw2libs = ['fftw2xc', 'fftw2xf']
             fftw3libs = ['fftw3xc', 'fftw3xf']
-            cdftlibs = ['fftw2x_cdft']
-            if LooseVersion(self.version) >= LooseVersion('10.3'):
-                cdftlibs.append('fftw3x_cdft')
+            # cdftlibs set in prepare_step
+            cdftlibs = self.cdftlibs
```
boegel: @bartoldeman There's no point in doing this, should use `self.cdftlibs` everywhere and just stop using `cdftlibs`?

easybuild/easyblocks/i/imkl.py
```diff
@@ -77,6 +77,34 @@ def prepare_step(self, *args, **kwargs):
...
+                if get_software_root('MPICH2') or get_software_root('MVAPICH2'):
+                    mpi_spec = 'mpich2'
+                elif get_software_root('OpenMPI'):
+                    mpi_spec = 'openmpi'
+                elif not get_software_root('impi'):
+                    # no compatible MPI found: do not build cdft
+                    cdftlibs = []
+                self.log.debug("Determined MPI specification based on loaded MPI module: %s" % mpi_spec)
+            self.cdftlibs = cdftlibs
+            self.mpi_spec = mpi_spec
```
boegel: @bartoldeman Please initialize both of these in the constructor of the easyblock (`__init__`)

easybuild/easyblocks/i/imkl.py
```diff
@@ -77,6 +77,34 @@ def prepare_step(self, *args, **kwargs):
...
+            if hasattr(self.toolchain, 'MPI_FAMILY') and self.toolchain.MPI_FAMILY is not None:
+                mpi_spec_by_fam = {
+                    toolchain.MPICH: 'mpich2',  # MPICH is MPICH v3.x, which is MPICH2 compatible
+                    toolchain.MPICH2: 'mpich2',
+                    toolchain.MVAPICH2: 'mpich2',
+                    toolchain.OPENMPI: 'openmpi',
+                }
+                mpi_fam = self.toolchain.mpi_family()
+                mpi_spec = mpi_spec_by_fam.get(mpi_fam)
+                self.log.debug("Determined MPI specification based on MPI toolchain component: %s" % mpi_spec)
```
boegel: @bartoldeman While touching this, please change this to:

```python
self.log.debug("...: %s", mpi_spec)
```

(`, ` rather than ` % `; same below)

easybuild/easyblocks/i/imkl.py
```diff
@@ -77,6 +77,34 @@ def prepare_step(self, *args, **kwargs):
             super(EB_imkl, self).prepare_step(*args, **kwargs)
         else:
             super(EB_imkl, self).prepare_step(*args, **kwargs)
+        if self.cfg['interfaces']:
```
boegel: @bartoldeman One more minor style issues: add an empty line above (and maybe also a comment explaining what's going on in the block below)

easybuild/easyblocks/i/imkl.py
```diff
@@ -265,28 +292,8 @@ def post_install_step(self):
                 if lib in fftw3libs:
                     buildopts.append('install_to=$INSTALL_DIR')
                 elif lib in cdftlibs:
```
boegel: should be `self.cdftlibs`, same above

bartoldeman:
Changes applied and tested with imkl easyconfigs with and without MPI toolchains.

bartoldeman:
Ping?

easybuild/easyblocks/i/imkl.py
```diff
@@ -71,6 +71,38 @@ def __init__(self, *args, **kwargs):
...
+                mpi_fam = self.toolchain.mpi_family()
+                self.mpi_spec = mpi_spec_by_fam.get(mpi_fam)
+                debugstr = "MPI toolchain component"
+            else:
+                # can't use toolchain.mpi_family, because of dummy toolchain
+                if get_software_root('MPICH2') or get_software_root('MVAPICH2'):
+                    self.mpi_spec = 'mpich2'
+                elif get_software_root('OpenMPI'):
+                    self.mpi_spec = 'openmpi'
+                elif not get_software_root('impi'):
```
boegel: @bartoldeman At this point, the modules for the dependencies are not loaded yet (that's done in `prepare_step`), so each of these `get_software_root` calls will return `None`...

We should initialize `self.cdftlibs` and `self.mpi_spec` in `__init__`, and move this whole block to `prepare_step`, after the `super(EB_imkl, self).prepare_step` call

PR #1396, opened 2018-04-04
Add Lua easyblock to patch `luaconf.h` to correctly insert path to Lua install

rubendibattista:
The `luaconf.h` file needs to be patched in order for `lua` to be able to find modules inside the correct install dir.

### DIFF
--- a/easybuild/easyblocks/l/lua.py
+++ b/easybuild/easyblocks/l/lua.py
@@ -36,12 +36,11 @@ i.e. configure/make/make install, implemented as an easyblock.
 import fileinput
 import os
 import re
+import sys
 
-from easybuild.framework.easyblock import EasyBlock
 from easybuild.easyblocks.generic.configuremake import ConfigureMake
-from easybuild.framework.easyconfig import CUSTOM
 from easybuild.tools.run import run_cmd
-from easybuild.tools.modules import get_software_root
+from easybuild.tools.build_log import EasyBuildError
 
 
 class EB_Lua(ConfigureMake):
@@ -59,15 +58,29 @@ class EB_Lua(ConfigureMake):
 
         src_dir = self.src[0]['finalpath']
 
-        luaconf_h = os.path.join(src_dir, 'luaconf.h')
+        luaconf_h = os.path.join(src_dir, 'src', 'luaconf.h')
 
         # Open file
         self.log.debug("Patching luaconf.h at {}".format(luaconf_h))
         for line in fileinput.input(luaconf_h, inplace=True):
             if line.startswith('#define LUA_ROOT'):
-                re.replace(r'/usr/local', self.install_subdir, line)
-                print(line)
+                installdir = os.path.join(self.installdir, '')  # Add trailing slash
+                line = re.sub(r'/usr/local/', installdir, line)
+            sys.stdout.write(line)
 
+    def test_step(self):
+        out = super(EB_Lua, self).test_step()
+
+        # Check if the path where lua looks for packages is correctly pointing
+        # to installdir
+        cmd = "src/lua -e 'io.write(package.path)'"
+
+        stdout, exitcode = run_cmd(cmd, log_all=True, simple=False)
+
+        if not(self.installdir in stdout):
+            raise EasyBuildError("package.path seems to not point in the right path")
+
+        return out
 
     def install_step(self):
         """
--- a/easybuild/easyblocks/l/lua.py
+++ b/easybuild/easyblocks/l/lua.py
@@ -1,0 +1,83 @@
+##
+# Copyright 2009-2018 Ghent University
+#
+# This file is part of EasyBuild,
+# originally created by the HPC team of Ghent University (http://ugent.be/hpc/en),
+# with support of Ghent University (http://ugent.be/hpc),
+# the Flemish Supercomputer Centre (VSC) (https://www.vscentrum.be),
+# Flemish Research Foundation (FWO) (http://www.fwo.be/en)
+# and the Department of Economy, Science and Innovation (EWI) (http://www.ewi-vlaanderen.be/en).
+#
+# https://github.com/easybuilders/easybuild
+#
+# EasyBuild is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation v2.
+#
+# EasyBuild is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with EasyBuild.  If not, see <http://www.gnu.org/licenses/>.
+##
+"""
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/l/lua.py
```diff
@@ -0,0 +1,96 @@
...
+        `luaconf.h` file in the sources to point to the correct Lua Root
+        """
+
+        src_dir = self.src[0]['finalpath']
+
+        luaconf_h = os.path.join(src_dir, 'src', 'luaconf.h')
+
+        # Open file
+        self.log.debug("Patching luaconf.h at {}".format(luaconf_h))
+        for line in fileinput.input(luaconf_h, inplace=True):
```
boegel: @rubendibattista Please use `apply_regex_substitutions` from `easybuild.tools.filetools` instead

easybuild/easyblocks/l/lua.py
```diff
@@ -0,0 +1,96 @@
...
+        """
+        Configure step
+
+        Lua does not need a configure step. In this step we just patch the
+        `luaconf.h` file in the sources to point to the correct Lua Root
+        """
+
+        src_dir = self.src[0]['finalpath']
+
+        luaconf_h = os.path.join(src_dir, 'src', 'luaconf.h')
```
boegel: Can't use `self.builddir`?

easybuild/easyblocks/l/lua.py
```diff
@@ -0,0 +1,96 @@
...
+    def test_step(self):
+        out = super(EB_Lua, self).test_step()
+
+        # Check if the path where lua looks for packages is correctly pointing
+        # to installdir
+        cmd = "src/lua -e 'io.write(package.path)'"
+
+        stdout, exitcode = run_cmd(cmd, log_all=True, simple=False)
+
+        if not(self.installdir in stdout):
```
boegel: should be `if self.installdir not in stdout:` (which is not the same thing)

easybuild/easyblocks/l/lua.py
```diff
@@ -0,0 +1,96 @@
...
+        # Open file
+        self.log.debug("Patching luaconf.h at {}".format(luaconf_h))
+        for line in fileinput.input(luaconf_h, inplace=True):
+            if line.startswith('#define LUA_ROOT'):
+                installdir = os.path.join(self.installdir, '')  # Add trailing slash
+                line = re.sub(r'/usr/local/', installdir, line)
+            sys.stdout.write(line)
+
+    def test_step(self):
+        out = super(EB_Lua, self).test_step()
```
boegel: missing docstring

easybuild/easyblocks/l/lua.py
```diff
@@ -0,0 +1,96 @@
...
+    def install_step(self):
+        """
+        Create the installation in correct location
+        - typical: make install
+        """
+
+        cmd = "%s make install INSTALL_TOP=%s %s" % (self.cfg['preinstallopts'],
+                                                     self.install_subdir, self.cfg['installopts'])
+
+        (out, _) = run_cmd(cmd, log_all=True, simple=False)
```
boegel: Why not call out to parent class, like this:

```python
self.cfg.update('installopts', 'INSTALL_TOP=%s' % self.install_subdir)
super(EB_Lua, self).install_step()
```

Are you sure `install_subdir` is right, shouldn't you pass the full path `self.installdir`?

easybuild/easyblocks/l/lua.py
```diff
@@ -0,0 +1,85 @@
+##
+# Copyright 2009-2018 Ghent University
```
boegel: @rubendibattista New easyblock, so should be `2018-2018`?

easybuild/easyblocks/l/lua.py
```diff
@@ -0,0 +1,85 @@
...
+"""
+EasyBuild support for software that uses the GNU installation procedure,
+i.e. configure/make/make install, implemented as an easyblock.
+
+@author: Stijn De Weirdt (Ghent University)
+@author: Dries Verdegem (Ghent University)
+@author: Kenneth Hoste (Ghent University)
+@author: Pieter De Baets (Ghent University)
+@author: Jens Timmerman (Ghent University)
+@author: Toon Willems (Ghent University)
```
boegel: @rubendibattista You should be the (only) author?

easybuild/easyblocks/l/lua.py
```diff
@@ -0,0 +1,85 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with EasyBuild.  If not, see <http://www.gnu.org/licenses/>.
+##
+"""
+EasyBuild support for software that uses the GNU installation procedure,
+i.e. configure/make/make install, implemented as an easyblock.
```
boegel: Please change to: `Custom easyblock for building and installing Lua.`

easybuild/easyblocks/l/lua.py
```diff
@@ -0,0 +1,85 @@
...
+        """
+        Configure step
+
+        Lua does not need a configure step. In this step we just patch the
+        `luaconf.h` file in the sources to point to the correct Lua Root
+        """
+        src_dir = self.src[0]['finalpath']
+        luaconf_h = os.path.join(src_dir, 'src', 'luaconf.h')
+
+        # Open file
```
boegel: comment is incorrect/useless?

easybuild/easyblocks/l/lua.py
```diff
@@ -0,0 +1,85 @@
...
+        self.log.debug("Patching luaconf.h at {}".format(luaconf_h))
+        installdir = os.path.join(self.installdir, '')  # Add trailing slash
```
boegel: Why not do that explicitly in the line below?

```python
apply_regex_substitutions(luaconf_h, [(r'/usr/local/', '%s/' % self.installdir)])

File: easybuild/easyblocks/l/lua.py [hunk 11]
```diff
@@ -0,0 +1,85 @@
...
+        apply_regex_substitutions(luaconf_h, [(r'/usr/local/', installdir)])
+
+    def test_step(self):
+        out = super(EB_Lua, self).test_step()
+
+        # Check if the path where lua looks for packages is correctly pointing
+        # to installdir
+        cmd = "src/lua -e 'io.write(package.path)'"
```
boegel: This should be a sanity check command, so we check the `lua` binary that actually gets installed?

When writing a custom easyblock, you should include a custom implementation of `sanity_check_step` anyway...

easybuild/easyblocks/l/lua.py
```diff
@@ -0,0 +1,85 @@
...
+
+        if self.installdir not in stdout:
+            raise EasyBuildError("package.path seems to not point in the right path")
+
+        return out
+
+    def install_step(self):
+        """
+        Create the installation in correct location
+        - typical: make install
```
boegel: Please customize the docstring to what is actually going on...

PR #1397, opened 2018-04-11
add support to enable integration of pscom in psmpi easyblock

damianam:
This enables integration of pscom in psmpi build

### DIFF
--- a/easybuild/easyblocks/p/psmpi.py
+++ b/easybuild/easyblocks/p/psmpi.py
@@ -57,6 +57,7 @@ class EB_psmpi(EB_MPICH):
         extra_vars.update({
             'mpich_opts': [None, "Optional options to configure MPICH", CUSTOM],
             'threaded': [False, "Enable multithreaded build (which is slower)", CUSTOM],
+            'pscom_allin_path': [None, "Enable pscom integration by giving its source path", CUSTOM],
         })
         return extra_vars
 
@@ -92,8 +93,11 @@ class EB_psmpi(EB_MPICH):
             self.cfg.update('configopts', ' --with-mpichconf="%s"' % self.cfg['mpich_opts'])
 
         # Lastly, set pscom related variables
-        self.cfg.update('preconfigopts', ('PSCOM_LDFLAGS=-L{0}/lib '+
-                'PSCOM_CPPFLAGS=-I{0}/include').format(get_software_root('pscom')))
+        if self.cfg['pscom_allin_path'] is None:
+            self.cfg.update('preconfigopts', ('PSCOM_LDFLAGS=-L{0}/lib '+
+                    'PSCOM_CPPFLAGS=-I{0}/include').format(get_software_root('pscom')))
+        else:
+            self.cfg.update('configopts', ' --with-pscom-allin="%s"' % self.cfg['pscom_allin_path'])
 
         super(EB_psmpi, self).configure_step(add_mpich_configopts=False)
 
--- a/easybuild/easyblocks/p/psmpi.py
+++ b/easybuild/easyblocks/p/psmpi.py
@@ -97,7 +97,10 @@ class EB_psmpi(EB_MPICH):
             self.cfg.update('preconfigopts', ('PSCOM_LDFLAGS=-L{0}/lib '+
                     'PSCOM_CPPFLAGS=-I{0}/include').format(get_software_root('pscom')))
         else:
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/p/psmpi.py
```diff
@@ -92,8 +93,14 @@ def configure_step(self):
...
-        self.cfg.update('preconfigopts', ('PSCOM_LDFLAGS=-L{0}/lib '+
-                'PSCOM_CPPFLAGS=-I{0}/include').format(get_software_root('pscom')))
+        if self.cfg['pscom_allin_path'] is None:
+            self.cfg.update('preconfigopts', ('PSCOM_LDFLAGS=-L{0}/lib '+
+                    'PSCOM_CPPFLAGS=-I{0}/include').format(get_software_root('pscom')))
+        else:
+            pscom_allin_path = self.cfg['pscom_allin_path'].strip()
+            self.cfg.update('configopts', ' --with-pscom-allin="%s"' % pscom_allin_path)
+            self.cfg.update('preconfigopts', ('PSCOM_LDFLAGS=-L{0}/lib '+
+                    'PSCOM_CPPFLAGS=-I{0}/include').format(pscom_allin_path))
```
boegel: I would reformat this a bit for readability, and to avoid duplicate code:

```python
if self.cfg['pscom_allin_path'] is None:
    pscom_allin_path = get_software_root('pscom')
else:
    pscom_allin_path = self.cfg['pscom_allin_path'].strip()
    self.cfg.update('configopts', ' --with-pscom-allin="%s"' % pscom_allin_path)

pscom_flags = "PSCOM_LDFLAGS=-L{0}/lib PSCOM_CPPFLAGS=-I{0}/include".format(pscom_allin_path)
self.cfg.update('preconfigopts', pscom_flags)
```

easybuild/easyblocks/p/psmpi.py
```diff
@@ -92,8 +93,14 @@ def configure_step(self):
             self.cfg.update('configopts', ' --with-mpichconf="%s"' % self.cfg['mpich_opts'])
 
         # Lastly, set pscom related variables
-        self.cfg.update('preconfigopts', ('PSCOM_LDFLAGS=-L{0}/lib '+
-                'PSCOM_CPPFLAGS=-I{0}/include').format(get_software_root('pscom')))
+        if self.cfg['pscom_allin_path'] is None:
+            self.cfg.update('preconfigopts', ('PSCOM_LDFLAGS=-L{0}/lib '+
+                    'PSCOM_CPPFLAGS=-I{0}/include').format(get_software_root('pscom')))
+        else:
+            pscom_allin_path = self.cfg['pscom_allin_path'].strip()
```
boegel: Why do you need the `.strip()` here, that's a bit strange?

easybuild/easyblocks/p/psmpi.py
```diff
@@ -57,6 +57,7 @@ def extra_options(extra_vars=None):
         extra_vars.update({
             'mpich_opts': [None, "Optional options to configure MPICH", CUSTOM],
             'threaded': [False, "Enable multithreaded build (which is slower)", CUSTOM],
+            'pscom_allin_path': [None, "Enable pscom integration by giving its source path", CUSTOM],
```
boegel: `s/source path/installation prefix/` ?

ocaisa:
I'm not sure I fully understand what this means? Is it that `pscom` becomes a build dependency only? In the past we've been able to switch out pscom for a new version, but this change would prevent that so what is the advantage?

easybuild/easyblocks/p/psmpi.py
damianam: Otherwise there was an empty space that messed up the definition of the variable

easybuild/easyblocks/p/psmpi.py
damianam: Done in the next commit

easybuild/easyblocks/p/psmpi.py
damianam: Well, it is an installation prefix just if it is installed, right? But with the `allin` option it is not really installed so....... I still think source path is "correcter"

damianam:
@ocaisa with this change `pscom` is integrated directly in `psmpi`. It is not a dependency of any kind. The advantage is enabling the compiler to use IPO, and to avoid the use of procedure lookup tables when calling library functions. This has a big impact on KNL, and that's the whole reason behind this change. In other processors it doesn't really make sense.

PR #1398, opened 2018-04-13
Set CMake patch when it's loaded in PythonPackage

wpoely86:
It's is not a great nor a final solution but it already helps.

We should think how to handle this in the generic case? Wrapper for CMake in the PythonPackage block?


wpoely86:
It's different (I think) in the sense that you never directly call cmake. For the outside, it's an ordinary python package but internally it calls cmake. The `CMakePythonPackage` actually runs cmake itself directly.

boegel:
Makes sense, thanks for clarifying @wpoely86.

I gave this a quick test with easyconfigs that could potentially be affected by this change (i.e. ones that combine `PythonPackage` with `CMake` as (build) dependency), but didn't see any problems, so good to go, thanks!

PR #1402, opened 2018-04-17
add custom easyblock for installing Nim

boegel:


vanzod:
Tested with `Nim-0.18.0-GCCcore-6.4.0.eb`. Passed

PR #1404, opened 2018-04-18
Lammps

mboisson:
Compute Canada's LAMMPS EasyBlock

### DIFF
--- a/easybuild/easyblocks/l/lammps.py
+++ b/easybuild/easyblocks/l/lammps.py
@@ -1,0 +1,186 @@
+##
+# Copyright 2009-2015 Ghent University
+#
+# This file is part of EasyBuild,
+# originally created by the HPC team of Ghent University (http://ugent.be/hpc/en),
+# with support of Ghent University (http://ugent.be/hpc),
+# the Flemish Supercomputer Centre (VSC) (https://vscentrum.be/nl/en),
+# the Hercules foundation (http://www.herculesstichting.be/in_English)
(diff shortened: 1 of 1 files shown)


boegel:
@mboisson This got integrated in #594, so this PR can be closed?

mboisson:
It seems so, but #594 still is incompatible with multi-python support as we do in CC.

damianam:
@mboisson maybe you can issue a PR against @boegel's branch? I think that's better than having 2 separate PR's tackling partly the same issues.

mboisson:
The two PRs are incompatible I believe because of explicit checks for python in the other PR

ocaisa:
@mboisson Are you very familiar with LAMMPS? I've gotten a working build using the new LAMMPS cmake procedure (see https://github.com/easybuilders/easybuild-easyconfigs/pull/6917) and I've been increasingly adding capabilities. I'm willing to work on a cmake-based easyblock but I need a collaborator that knows LAMMPS to know if the end result actually makes sense.

mboisson:
No. Our resident expert on LAMMPS would be @aliker

mboisson:
No. Our resident expert on LAMMPS would be @alikerr

boegel:
@ocaisa I think @vanzod is quite familiar with LAMMPS (although maybe not the recent CMake-based build procedure).

boegel:
closing this, superseded by https://github.com/easybuilders/easybuild-easyblocks/pull/1964

PR #1405, opened 2018-04-20
 update wien2k easyblock for v17

migueldiascosta:


easybuild/easyblocks/w/wien2k.py
```diff
@@ -101,21 +101,27 @@ def configure_step(self):
...
+            else:
+                comp_answer = 'LG'  # Linux (gfortran compiler + OpenBlas)
         else:
             raise EasyBuildError("Failed to determine toolchain-dependent answers.")
 
         # libraries
-        rlibs = "%s %s" % (os.getenv('LIBLAPACK_MT'), self.toolchain.get_flag('openmp'))
-        rplibs = [os.getenv('LIBSCALAPACK_MT'), os.getenv('LIBLAPACK_MT')]
+        liblapack = os.getenv('LIBLAPACK_MT').replace('static', 'dynamic')
+        libscalapack = os.getenv('LIBSCALAPACK_MT').replace('static', 'dynamic')
```
boegel: @migueldiascosta Can you explain this change? Maybe include a comment above these two lines to explain why you're doing this?

easybuild/easyblocks/w/wien2k.py
```diff
@@ -128,24 +134,30 @@ def configure_step(self):
...
-             'RP_LIBS' : rplibs,  # libraries for 'real' parallel binary
+             'RP_LIBS': rplibs,  # libraries for 'real' parallel binary
              'MPIRUN': '',
             }
 
         for line in fileinput.input(self.cfgscript, inplace=1, backup='.orig'):
             # set config parameters
-            for (k,v) in d.items():
-                regexp = re.compile('^([a-z0-9]+):%s:.*' % k)
+            for (k, v) in d.items():
```
boegel: @migueldiascosta While you're refactoring this, please avoid using single letter variables outside of list comprehensions, so please replace `k` with `key`, `v` with `val` and `d` with `vars`?

easybuild/easyblocks/w/wien2k.py
```diff
@@ -157,10 +169,16 @@ def configure_step(self):
             'COMPILERC': os.getenv('CC'),
             'COMPILER': os.getenv('F90'),
             'COMPILERP': os.getenv('MPIF90'),
-        }
+            }
```
boegel: Hmm, why change this indentation? Does `flake8` complain about this, I don't think it does?

easybuild/easyblocks/w/wien2k.py
```diff
@@ -157,10 +169,16 @@ def configure_step(self):
...
 
-        for (k, v) in dc.items():
-            write_file(k, v)
+        if LooseVersion(self.version) < LooseVersion("17"):
+            for (k, v) in dc.items():
```
boegel: same here, avoid single letter variables outside of list comprehensions

easybuild/easyblocks/w/wien2k.py
```diff
@@ -186,23 +204,57 @@ def configure_step(self):
...
                  'Shared Memory Architecture? (y/N):': 'N',
                  'Set MPI_REMOTE to  0 / 1:': '0',
                  'You need to KNOW details about your installed  MPI and FFTW ) (y/n)': 'y',
-                 'Please specify whether you want to use FFTW3 (default) or FFTW2  (FFTW3 / FFTW2):' : fftw_spec,
-                 'Please specify the ROOT-path of your FFTW installation (like /opt/fftw3):' : fftw_root,
-                 'is this correct? enter Y (default) or n:' : 'Y',
+                 'Please specify whether you want to use FFTW3 (default) or FFTW2  (FFTW3 / FFTW2):': fftw_spec,
+                 'Please specify the ROOT-path of your FFTW installation (like /opt/fftw3):': fftw_root,
+                 'is this correct? enter Y (default) or n:': 'Y',
+                 'LIBXC (that you have installed before)? (y,N):': '',
```
boegel: We actually have `libxc`, so you could add it as a dependency?

easybuild/easyblocks/w/wien2k.py
```diff
@@ -186,23 +204,57 @@ def configure_step(self):
...
+                        'Do you want to automatically search for FFTW installations? (Y,n):': 'n',
+                        'Please specify the ROOT-path of your FFTW installation (like /opt/fftw3/) '
+                        'or accept present choice (enter):': fftw_root,
+                        'Is this correct? enter Y (default) or n:': 'Y',
+                        'Do you want to use ELPA? (y,N):': 'N',
+                        'Please specify the name of your FFTW library or accept present choice (enter):': '',
+                        'Please specify your parallel compiler options or accept the recommendations '
+                        '(Enter - default)!:': '',
+                        'Please specify your MPIRUN command or accept the recommendations (Enter - default)!:': '',
+                        'Please enter the full path to your temporary directory:': '/tmp',
```
boegel: @migueldiascosta please don't hardcode `/tmp` here, use `tempfile.gettempdir()` (which was set to a unique directory for this `eb` session)

easybuild/easyblocks/w/wien2k.py
```diff
@@ -186,23 +204,57 @@ def configure_step(self):
...
+                        'Do you want to use a present FFTW installation? (Y,n):': 'y',
+                        'Please specify the path of your FFTW installation (like /opt/fftw3/) '
+                        'or accept present choice (enter):': fftw_root,
+                        'Please specify the target achitecture of your FFTW library (e.g. lib64) '
+                        'or accept present choice (enter):': '',
+                        'Do you want to automatically search for FFTW installations? (Y,n):': 'n',
+                        'Please specify the ROOT-path of your FFTW installation (like /opt/fftw3/) '
+                        'or accept present choice (enter):': fftw_root,
+                        'Is this correct? enter Y (default) or n:': 'Y',
+                        'Do you want to use ELPA? (y,N):': 'N',
```
boegel: same here, we have `ELPA` supported, so you could add it as a dep?

easybuild/easyblocks/w/wien2k.py
```diff
@@ -186,23 +204,57 @@ def configure_step(self):
...
+                        'Do you want to use the MKL version of ScaLAPACK? (Y,n):': 'n',  # we set it ourselves below
+                        'Do you use Intel MPI? (Y,n):': 'y',
+                        'Is this correct? (Y,n):': 'y',
+                        'Please specify the target architecture of your ScaLAPACK libraries (e.g. intel64)!:': '',
+                        'ScaLAPACK root:': os.getenv('MKLROOT') or os.getenv('EBROOTSCALAPACK'),
+                        'ScaLAPACK library:': next((lib[2:] for lib in os.getenv('LIBSCALAPACK').split()
+                                                    if 'scalapack' in lib), 'scalapack'),
+                        'BLACS root:': os.getenv('MKLROOT') or os.getenv('EBROOTOPENBLAS'),
+                        'BLACS library:': next((lib[2:] for lib in os.getenv('LIBSCALAPACK').split()
+                                                if 'blacs' in lib), 'openblas'),
```
boegel: please construct the values above the `qanda.update`

easybuild/easyblocks/w/wien2k.py
```diff
@@ -259,11 +313,21 @@ def build_step(self):
...
-                }
+            'Press RETURN to continue': '\nQ',  # also answer on first qanda pattern with 'Q' to quit
+            ' Please enter the full path of the perl program: ': '',
+            }
+
+        if LooseVersion(self.version) < LooseVersion("17"):
+            qanda.update({
+                    'L Perl path (if not in /usr/bin/perl) Q Quit Selection:': 'R',
+                    'A Compile all programs S Select program Q Quit Selection:': 'A',
+                    })
```
boegel: please fix indent for this line?

easybuild/easyblocks/w/wien2k.py
```diff
@@ -259,11 +313,21 @@ def build_step(self):
...
+        else:
+            qanda.update({
+                    'program Q Quit Selection:': 'A',
+                    'Path Q Quit Selection:': 'R',
+                    })
```
boegel: please fix indent for this line?

easybuild/easyblocks/w/wien2k.py
```diff
@@ -128,24 +134,30 @@ def configure_step(self):
         rplibs = ' '.join(rplibs)
 
         d = {
-             'FC': '%s %s' % (os.getenv('F90'), os.getenv('FFLAGS')),
-             'MPF': "%s %s" % (os.getenv('MPIF90'), os.getenv('FFLAGS')),
+             'FC': '%s' % os.getenv('F90'),
+             'FOPT': '%s' % os.getenv('OPTFLAGS'),
```
boegel: use `FFLAGS` rather than `OPTFLAGS`?

easybuild/easyblocks/w/wien2k.py
```diff
@@ -128,24 +134,30 @@ def configure_step(self):
...
+             'MPF': '%s' % os.getenv('MPIF90'),
+             'FPOPT': '%s' % os.getenv('OPTFLAGS'),
```
boegel: use `FFLAGS` rather than `OPTFLAGS`?

easybuild/easyblocks/w/wien2k.py
migueldiascosta: well, the new `siteconfig_lapw` script insists on adding the dynamic `mkl_scalapack` instead of using `RP_LIBS`. at least in `lapw2_mpi`, switching everything to dynamic solved the problem, but I suppose we can also patch the script...

easybuild/easyblocks/w/wien2k.py
migueldiascosta: will do

easybuild/easyblocks/w/wien2k.py
migueldiascosta: it wasn't on purpose (or maybe `emacs` was upset ;)), will change it back

easybuild/easyblocks/w/wien2k.py
boegel: They probably changed that for a reason?

Just wondering whether we should also do this for old WIEN2k versions...

easybuild/easyblocks/w/wien2k.py
migueldiascosta: this will be hardcoded into the `lapw` scripts, so it cannot be an `eb` temporary directory

I think the `siteconfig_lapw` requires a real path here... but maybe be can patch the scripts afterwards to use a variable...

boegel: OK, please mention that in a comment, that this is a directory that must exist at runtime...

easybuild/easyblocks/w/wien2k.py
migueldiascosta: I don't think they had a specific reason. Also, I tested the block with `14.2` and `16.1` and it still works.

But if we really want to use static libraries for some other reason, I can try harder to make it work

easybuild/easyblocks/w/wien2k.py
migueldiascosta: done

easybuild/easyblocks/w/wien2k.py
migueldiascosta: I can get it to build but only with `ELPA` version `2015.02.002`, `WIEN2k` seems to expect modules such as `mod_blacs_infrastructure` that don't seem to be present in more recent versions, did anything change? (I'm afraid I'm not familiar with `ELPA`...)

PR #1413, opened 2018-04-30
Install Chimera in a subdirectory to avoid its dependencies being added to the environment

sebth:
I noticed that when loading Chimera after loading Python, Python starts picking up the Tcl/Tk version that comes bundled with Chimera instead of the correct one. This PR fixes that by installing Chimera in a subdirectory so that the bundled dependencies are not added to PATH, LD_LIBRARY_PATH, etc. A symlink is added from 'bin/chimera' in the root directory to 'bin/chimera' in subdirectory, so that a 'chimera' command will exist in PATH. Chimera will still find its bundled dependencies, because it uses a wrapper startup script that adds them to the environment.

This method is similar to what they describe in the installation instructions at <http://www.cgl.ucsf.edu/chimera/data/downloads/1.11.2/linux.html>.

### DIFF
--- a/easybuild/easyblocks/c/chimera.py
+++ b/easybuild/easyblocks/c/chimera.py
@@ -10,6 +10,7 @@ import os
 
 from easybuild.easyblocks.generic.packedbinary import PackedBinary
 from easybuild.tools.build_log import EasyBuildError
+from easybuild.tools.filetools import mkdir, symlink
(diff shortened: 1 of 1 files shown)


boegel:
Thanks a lot for your contribution @sebth, sorry about the long wait for feedback...

This looks perfect & well documented, I tested this myself and it works as expected, so good to go.

PR #1417, opened 2018-05-08
New EasyBlock for the GDAL-GRASS extension

valtandor:


boegel:
@valtandor Do you have an example easyconfig that requires this?

easybuild/easyblocks/g/gdal_grass.py
```diff
@@ -0,0 +1,55 @@
...
+
+class EB_GDAL_minus_GRASS(ConfigureMake):
+    """Support for building/installing the GDAL-GRASS plugin."""
+
+    # Only the sanity check should need to be different
+    def sanity_check_step(self):
+        """
+        Custom sanity check for GDAL-GRASS
+        """
+        if os.getenv('EBROOTGDAL') is None:
```
damianam: Use `get_software_root ` from `easybuild.tools.modules`?

easybuild/easyblocks/g/gdal_grass.py
```diff
@@ -0,0 +1,55 @@
...
+            raise EasyBuildError("$EBROOTGDAL is not defined for some reason -- check environment")
```
damianam: `raise EasyBuildError("Dependency module GDAL not loaded?"` (for consistency with other parts of the code)

easybuild/easyblocks/g/gdal_grass.py
```diff
@@ -0,0 +1,55 @@
...
+        else:
+            gdalplugindir = os.path.join(os.getenv('EBROOTGDAL'), 'lib', 'gdalplugins')
```
damianam: The result of `get_software_root` used above can be stored in a variable instead of querying the environment multiple times.

akesandgren:
And it is wrong since gdal-grass should not write files into the gdal installation tree!
That's what the sanity_check is looking for at least.

PR #1420, opened 2018-05-09
Make configure and make look for FoX in EBROOTFOX in QuantumESPRESSO easyblock

damianam:


easybuild/easyblocks/q/quantumespresso.py
```diff
@@ -161,6 +161,13 @@ def configure_step(self):
                  r"\1\n\techo 'libfox: external module used' #"),
             ]
             apply_regex_substitutions('Makefile', regex_subs)
+            # Make configure and make look in the correct directory
+            regex_subs = [
+                (r"\(TOPDIR\)/FoX/finclude",
+                 r"(EBROOTFOX)/finclude"),
+            ]
```
boegel: Isn't it more clear to collapse this to a single line, since it's only a single substitution?

```python
regex_subs = [(r"\(TOPDIR\)/FoX/finclude", r"(EBROOTFOX)/finclude")]

File: easybuild/easyblocks/q/quantumespresso.py [hunk 2]
```diff
@@ -161,6 +161,13 @@ def configure_step(self):
...
+            apply_regex_substitutions('install/configure', regex_subs)
+            apply_regex_substitutions('make.inc', regex_subs)
```
boegel: Do you also need to patch `make.inc`? Doesn't seem to make any difference w.r.t. `FoX` include dir?

easybuild/easyblocks/q/quantumespresso.py
damianam: I am just following the same style than above.

easybuild/easyblocks/q/quantumespresso.py
boegel: The reference is to `-L$(TOPDIR)/FoX/lib`, and it's not getting patched currently:

```
$ grep TOPDIR make.inc* | grep FoX
make.inc:FOX_LIB  = -L$(TOPDIR)/FoX/lib  -lFoX_dom -lFoX_sax -lFoX_wxml -lFoX_common\
make.inc.orig.eb:FOX_LIB  = -L$(TOPDIR)/FoX/lib  -lFoX_dom -lFoX_sax -lFoX_wxml -lFoX_common\
```

Maybe simplify the regex to make it just replace `\(TOPDIR\)/FoX/` with `(EBROOTFOX)/` ?

easybuild/easyblocks/q/quantumespresso.py
boegel: Well, ok, the one above should be changed too ;)

easybuild/easyblocks/q/quantumespresso.py
boegel: @damianam Ping on fixing this?

boegel:
Works like a charm now, so going in, thanks @damianam!

PR #1424, opened 2018-05-14
fix suffix for Boost Python library in Boost 1.67.0 & newer

rubendibattista:
Newer Boost for `python2`  generates (at least on my machines) libraries with the version suffix `27`.

(easyconfig coming...)


boegel:
@rubendibattista I can confirm I'm seeing the same thing... The changes look good to me, but it would be nice to have a test case for this (i.e. an easyconfig for a recent version of Boost with Python bindings enabled).

rubendibattista:
It's coming ;)

boegel:
Tested with both easyconfig for Boost 1.67.0 w/ Python 2.7.14 in https://github.com/easybuilders/easybuild-easyconfigs/pull/6357 and a couple of existing Boost easyconfigs, so good to go.

Thanks @rubendibattista!

PR #1425, opened 2018-05-14
Use configParser to manage site.cfg. Handle OpenBLAS accordingly to documentation

rubendibattista:
I reversed the order the things that are done because I needed to compute the various `blas`, `lapack` etc variables before putting them into the `site.cfg` since the `site.cfg` of numpy does not allow interpolation

### DIFF
--- a/easybuild/easyblocks/n/numpy.py
+++ b/easybuild/easyblocks/n/numpy.py
@@ -161,7 +161,6 @@ class EB_numpy(FortranPythonPackage):
                 self.sitecfg.set('mkl', 'lapack_libs', lapack)
                 self.sitecfg.set('mkl', 'mk_libs', blas)
 
-
         else:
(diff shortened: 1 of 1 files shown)


rubendibattista:
A Note: since the `site.cfg` is a `ConfigParser` file, maybe it would be useful to handle it like that in the `pythonpackage` easyblock directly, instead of translating it into a string.

boegel:
@rubendibattista Before we deep-dive into this: is this backwards compatible?

It seems like you're mixing two (major?) changes here: the use of `ConfigParser` (which makes sense at first sight) + changes w.r.t. how OpenBLAS is handled... I'm wondering whether we should tackle that in separate PRs?

rubendibattista:
@boegel I just checked for numpy 1.9.3 and the specification of the `site.cfg` already contained the `[openblas]` section. Did not check by the way. How should we handle this to ensure it's backward compatible?

Issue #801, opened 2016-01-11
Binary unconditionaly prepends installdir to PATH


nathanhaigh:
I've recently had a problem with writing an easyconfig for hub (https://github.com/hpcugent/easybuild-easyconfigs/pull/2249).

It boiled down to the fact that the program executables were in a `bin` subdirectory and the module file created by EasyBuild had prepended both the `bin` subdirectory and the parent directory to the `PATH`. Normally, this might not cause issues. However, with `hub` there was an executable in the parent directory, which is distributed to help installing `hub` to the relevant location. This file was called `install`.

As a result, on subsequent calls to `eb` for installing tools, it was `hub`'s `install` file which was being picked up and not the expected system's. To make matters worse, `hub`'s `install` made a call to `install` itself. Thus resulting in recursive calls to `hub`'s `install` file.

It would be good to have a way to conditionally add the `installdir` to `PATH`

Flamefire:
I do have a very similar problem: I have some software that comes with an `install.sh` script with non-standard parameters. This script serves as a wrapper for some custom `configure` and also executes `make` and copies files. So although the software is not in binary form, it has a binary installer which makes it suitable for the `Binary` EasyBlock.

However as it installs a directory structure (containing `bin`) I don't want `PATH` to include the INSTALL_DIR.

boegel:
This can be closed once @Flamefire's PR #1426 gets merged?

boegel:
ok to close now that #1426 is merged (to be included in EasyBuild v3.6.1), which paths are added to `$PATH` can now be controlled via the custom easyconfig parameter `prepend_to_path`

PR #1426, opened 2018-05-17
support adding specific paths to $PATH for generic Binary easyblock via 'prepend_to_path' custom easyconfig parameter

Flamefire:
Also allows to not add a path at all or set it to a specific one.
See e.g. #801 for problems caused by current behavior.

Basically: This EasyBlock is for binary packages or binary installers. So assuming the User/EasyConfig meant to add `installdir` to `PATH` is questionable at least as we do know nothing about the package to be installed.

I thought about making it a simple True/False switch but I think this is more flexible: You can either be fine with `installdir` being added (kept as default for backwards compatibility), change it to a custom binary path or disable it altogether.

### DIFF
--- a/easybuild/easyblocks/generic/binary.py
+++ b/easybuild/easyblocks/generic/binary.py
@@ -58,9 +58,8 @@ class Binary(EasyBlock):
             'install_cmd': [None, "Install command to be used.", CUSTOM],
             # staged installation can help with the hard (potentially faulty) check on available disk space
             'staged_install': [False, "Perform staged installation via subdirectory of build directory", CUSTOM],
-            'prepend_to_path': ['',
-                                "Prepend the given path to environment PATH in the module file. Set to None to skip",
-                                CUSTOM],
+            'prepend_to_path': [[''], "Prepend the given directories (relative to install-dir) to the environment "
+                                      "variable PATH in the module file.", CUSTOM],
         })
         return extra_vars
 
@@ -134,10 +133,10 @@ class Binary(EasyBlock):
                       self.__class__.__name__)
 
     def make_module_extra(self):
-        """Add the specified directory to the PATH."""
+        """Add the specified directories to the PATH."""
 
         txt = super(Binary, self).make_module_extra()
-        if not self.cfg['prepend_to_path'] is None:
-            txt += self.module_generator.prepend_paths("PATH", [self.cfg['prepend_to_path']])
+        if self.cfg['prepend_to_path'] is not None:
+            txt += self.module_generator.prepend_paths("PATH", self.cfg['prepend_to_path'])
         self.log.debug("make_module_extra added this: %s" % txt)
         return txt
--- a/easybuild/easyblocks/generic/binary.py
+++ b/easybuild/easyblocks/generic/binary.py
@@ -71,7 +71,7 @@ class Binary(EasyBlock):
         super(Binary, self).__init__(*args, **kwargs)
 
         self.actual_installdir = None
-        if self.cfg['staged_install']:
+        if self.cfg.get('staged_install', False):
             self.actual_installdir = self.installdir
             self.installdir = os.path.join(self.builddir, 'staged')
             mkdir(self.installdir, parents=True)
@@ -102,7 +102,8 @@ class Binary(EasyBlock):
 
     def install_step(self):
         """Copy all files in build directory to the install directory"""
-        if self.cfg['install_cmd'] is None:
+        install_cmd = self.cfg.get('install_cmd', None)
+        if install_cmd is None:
             try:
                 # shutil.copytree doesn't allow the target directory to exist already
                 rmtree2(self.installdir)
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/generic/binary.py
```diff
@@ -58,6 +58,9 @@ def extra_options(extra_vars=None):
             'install_cmd': [None, "Install command to be used.", CUSTOM],
             # staged installation can help with the hard (potentially faulty) check on available disk space
             'staged_install': [False, "Perform staged installation via subdirectory of build directory", CUSTOM],
+            'prepend_to_path': ['',
+                                "Prepend the given path to environment PATH in the module file. Set to None to skip",
```
boegel: Please clarify that the specified path is relative to the installation prefix? Also, putting each part on a separate line is overkill?

```python
'prepend_to_path': ['', "Prepend given directory (relative to installation prefix) to $PATH (use None to skip)",
                    CUSTOM],

File: easybuild/easyblocks/generic/binary.py [hunk 2]
```diff
@@ -131,9 +134,10 @@ def sanity_check_rpath(self):
                       self.__class__.__name__)
 
     def make_module_extra(self):
-        """Add the install directory to the PATH."""
+        """Add the specified directory to the PATH."""
 
         txt = super(Binary, self).make_module_extra()
-        txt += self.module_generator.prepend_paths("PATH", [''])
+        if not self.cfg['prepend_to_path'] is None:
```
boegel: @Flamefire Please use `if ... is not None`, see https://docs.python.org/2/reference/expressions.html#not-in

easybuild/easyblocks/generic/binary.py
boegel: Also, maybe we should make the default value `['']` (and adjust accordingly below), to allow for adding multiple location to `$PATH`?

And then you don't need to use `None` as a magic value, an empty list will do too to skip adding anything to `$PATH`...

Flamefire: The list idea is good, I'm just not sure if it is helpful. Can there be any situation where you would want to add multiple paths to PATH? Usually  it is "bin" and this already lets you add one additional path. Should be enough but I'll change that if you like it better. And I don't consider `None` to be a magic value. IMO it is very descriptive.

About the separate lines: The Code-review bot was complaining about line length. If I add the relative part then even this would not fit. Where and how shall I split it? I could not find convention.

boegel: I admit that having to add multiple locations to `$PATH` is rare, but it does happen (e.g. both `bin` & `scripts`). You are right that not being able to provide multiple locations via `prepend_to_path` is not a big issue though, you can always get additional locations included via `modextrapaths = {'PATH': [...]}` in the easyconfig file.
I was only referring to `None` as magical value since an empty string `''` can't be used to skip adding a location to `$PATH`, and with a list this is no longer necessary (since an empty list `[]` also does that).
With that in mind, let's indeed keep the default value as is (`''`).

W.r.t. line length:

```python
'prepend_to_path': ['', "Prepend given directory (relative to installation prefix) to $PATH"
                    " (use None to skip)", CUSTOM],
```

(not using a `,` after the first line results in the two strings to be concatenated for the help message)

boegel:
@Flamefire Any idea if you'll be able to make the (trivial) requested changes soon? It would be nice to squeeze this into EasyBuild v3.6.1, which I'd like to get released this week...

Flamefire:
I was just doing them :)

easybuild/easyblocks/generic/binary.py
```diff
@@ -131,9 +133,10 @@ def sanity_check_rpath(self):
                       self.__class__.__name__)
 
     def make_module_extra(self):
-        """Add the install directory to the PATH."""
+        """Add the specified directories to the PATH."""
 
         txt = super(Binary, self).make_module_extra()
-        txt += self.module_generator.prepend_paths("PATH", [''])
+        if self.cfg['prepend_to_path'] is not None:
+            txt += self.module_generator.prepend_paths("PATH", self.cfg['prepend_to_path'])
```
boegel: @Flamefire I think it's a good idea to be a bit more careful here w.r.t. accessing the custom `prepend_to_path` easyconfig parameter. Easyblocks that derive from `Binary` may not pick up the custom easyconfig parameters defined for `Binary` (because they don't properly define the `extra_options` static method, or because they don't care), which would result in `self.cfg['prepend_to_path']` failing hard with ` KeyError`...

We can be robust against this by using `self.cfg.get('prepend_to_path')`, but to fully retain backward compatibility we also need to make sure `['']` is used as a default value, so we should introduce a constant to avoid hardcoding `['']` in multiple places:

```python
PREPEND_TO_PATH_DEFAULT = ['']
...
    def make_module_extra(self):
        ...
        prepend_to_path = self.cfg.get('prepend_to_path', PREPEND_TO_PATH_DEFAULT)
        if prepend_to_path:
            txt += self.module_generator.prepend_paths("PATH", prepend_to_path)

Flamefire: Sure. I was just following existing usages. Example: `if self.cfg['install_cmd'] is None:`

Shall I change those usages too while I'm on it?

boegel: @Flamefire Makes sense, but better do that in a separate commit?

boegel: Those actually show that this may be overkill (if there would be a problem with easyblock deriving from `Binary` we would've known already), but it doesn't hurt in the long run to make this a bit more robust...

boegel:
Reviewed & tested this with a couple of easyconfigs using `Binary` as well as with easyblocks that derive from `Binary`, good to go, thanks a lot for tackling this @Flamefire!

PR #1427, opened 2018-05-19
assume PGI Community edition is used when no license file is specified

boegel:


boegel:
Even worse:

```
$ ml show PGI
...
prepend_path("None","UNKNOWN")

$ ml PGI
$ echo $None
UNKNOWN
```

I'll get that fixed, thanks for pointing this out @wpoely86!

Issue #1407, opened 2018-04-25
Doris easyblock needs to be tweaked to accept some variations in questions for run_cmd_qa


migueldiascosta:
e.g. the easyblock expects

```python
"===> Do you have the LAPACK library (y/n)? [n]": 'y',
"===> What is the path to the LAPACK library liblapack.a? []": os.getenv('LAPACK_LIB_DIR'),
```

but the default values, e.g. `[n]` and `[]`, will be different if the script detects the libraries somewhere

migueldiascosta:
fixed by #1428

PR #1428, opened 2018-05-22
avoid hardcoding defaults in question patterns in Doris easyblock

boegel:
fix for #1407


boegel:
tested with existing `Doris` easyconfigs, still works like charm

PR #1434, opened 2018-06-04
Boost: add option use_glibcxx11_abi

smoors:
Needed for building EMAN2

### DIFF
--- a/easybuild/easyblocks/b/boost.py
+++ b/easybuild/easyblocks/b/boost.py
@@ -176,13 +176,14 @@ class EB_Boost(EasyBlock):
         bjamoptions = " --prefix=%s" % self.objdir
 
         cxxflags = os.getenv('CXXFLAGS')
-        cxxflags += ' -D_GLIBCXX_USE_CXX11_ABI='
-        if self.cfg['use_glibcxx11_abi']:
-            cxxflags += '1'
-        elif self.cfg['use_glibcxx11_abi'] is False:
-            # only disable if use_glibcxx11_abi was explicitly set to False
-            # None value is the default, which corresponds to default setting (=1 since GCC 5.x)
-            cxxflags += '0'
+        # only disable -D_GLIBCXX_USE_CXX11_ABI if use_glibcxx11_abi was explicitly set to False
+        # None value is the default, which corresponds to default setting (=1 since GCC 5.x)
+        if self.cfg['use_glibcxx11_abi'] is not None:
+            cxxflags += ' -D_GLIBCXX_USE_CXX11_ABI='
+            if self.cfg['use_glibcxx11_abi']:
+                cxxflags += '1'
+            else:
+                cxxflags += '0'
         if cxxflags is not None:
             bjamoptions += " cxxflags='%s'" % cxxflags
         ldflags = os.getenv('LDFLAGS')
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/b/boost.py
```diff
@@ -175,6 +176,10 @@ def build_step(self):
         bjamoptions = " --prefix=%s" % self.objdir
 
         cxxflags = os.getenv('CXXFLAGS')
+        if self.cfg['use_glibcxx11_abi'] is False:
+            cxxflags += ' -D_GLIBCXX_USE_CXX11_ABI=0'
+        elif self.cfg['use_glibcxx11_abi']:
+            cxxflags += ' -D_GLIBCXX_USE_CXX11_ABI=1'
```
boegel: @smoors This changes the default to always include `-D_GLIBCXX_USE_CXX11_ABI=1`, is that a good idea?

smoors: That's not the idea. If the flag is not set, the default setting is used, whatever that is (in that case `self.cfg['use_glibcxx11_abi'] is None`).
If it is set to True/False, it is used/not used regardless of the default setting.

smoors: Or maybe I misunderstood how this works in easybuild?

boegel: No, you're right, but the logic is a bit confusing, and the duplicate code doesn't help.

How about this as alternative:

```python
cxxflags += ' -D_GLIBCXX_USE_CXX11_ABI='
if self.cfg['use_glibcxx11_abi']:
    cxxflags += '1'
else:
    cxxflags += '0'
```

smoors: @boegel Your alternative is clearer but does not allow for the default GCC setting to be used. Maybe that's ok because `-D_GLIBCXX_USE_CXX11_ABI=1` has been the default since GCC 5 ?
I suggest another even simpler alternative (with comment added):
```
# If 'use_glibcxx11_abi' is not set or is True, use the default which is '-D_GLIBCXX_USE_CXX11_ABI=1' since GCC 5
if self.cfg['use_glibcxx11_abi'] is False:
    cxxflags += ' -D_GLIBCXX_USE_CXX11_ABI=0'
```

boegel: You're right w.r.t. the default. Your alternative makes sense, but doesn't give the option to explicitly enable it, which may be useful when working with older GCC versions.

So:

```python
cxxflags += ' -D_GLIBCXX_USE_CXX11_ABI='
if self.cfg['use_glibcxx11_abi']:
    cxxflags += '1'
elif self.cfg['use_glibcxx11_abi'] is False:
    # only disable if use_glibcxx11_abi was explicitly set to False
    # None value is the default, which corresponds to default setting (=1 since GCC 5.x)
    cxxflags += '0'

smoors: I agree. Note that the logic is now the same as my original code ;)

akesandgren: If it is None then wouldn't it ignore both cases and end up with ...ABI= without a 0 or 1 at the end?

(Python is still not my friend)

smoors: Ah, yes :(

boegel: Good catch @akesandgren, so maybe wrap the whole thing under `if self.cfg['use_glibcxx11_abi'] is not None:`?

smoors:
@boegel @akesandgren this version should make everybody happy ?

easybuild/easyblocks/b/boost.py
smoors: too late, I already did it ;)

boegel:
Tested with https://github.com/easybuilders/easybuild-easyconfigs/pull/6420 and existing Boost easyconfigs, good to go, thanks @smoors!

PR #1436, opened 2018-06-13
Enabled filtering of paths in the CPATH variable generated for the wrapper

damianam:
As we know, bazel pulls a lot of stuff on its own. In the existing setup, these versions might conflict with packages already installed and included in the CPATH variable generated for the `icc` wrapper. This PR enables package filtering there. What this does not do is filtering CPATH when using `gcc`-based toolchains. Input? Should we extend the filter to work also under other toolchains where a wrapper is not generated?

### DIFF
--- a/easybuild/easyblocks/t/tensorflow.py
+++ b/easybuild/easyblocks/t/tensorflow.py
@@ -110,7 +110,7 @@ class EB_TensorFlow(PythonPackage):
                 self.log.info("$%s old value was %s" % (var, path))
                 filtered_path = os.pathsep.join([p for fil in path_filter for p in path if fil not in p])
                 self.log.info("$%s new value is %s" % (var, filtered_path))
-                setvar(var, filtered_path)
+                env.setvar(var, filtered_path)
 
         # put wrapper for Intel C compiler in place (required to make sure license server is found)
         # cfr. https://github.com/bazelbuild/bazel/issues/663
@@ -125,7 +125,7 @@ class EB_TensorFlow(PythonPackage):
             }
             icc_wrapper = os.path.join(wrapper_dir, 'icc')
             write_file(icc_wrapper, icc_wrapper_txt)
-            env.setvar('PATH', ':'.join([os.path.dirname(icc_wrapper), os.getenv('PATH')]))
+            env.setvar('PATH', os.pathsep.join([os.path.dirname(icc_wrapper), os.getenv('PATH')]))
             if self.dry_run:
                 self.dry_run_msg("Wrapper for 'icc' was put in place: %s", icc_wrapper)
             else:
--- a/easybuild/easyblocks/t/tensorflow.py
+++ b/easybuild/easyblocks/t/tensorflow.py
@@ -103,13 +103,12 @@ class EB_TensorFlow(PythonPackage):
         # filter out paths from CPATH and LIBRARY_PATH. This is needed since bazel will pull some dependencies that
         # might conflict with dependencies on the system and/or installed with EB. For example: protobuf
         path_filter = self.cfg['path_filter']
-        if len(path_filter > 0):
-            self.log.info("Filtering $CPATH and $LIBRARY_PATH")
+        if path_filter:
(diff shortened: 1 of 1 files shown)


boegel:
@damianam Do you have an example of how you use this, and which exact problem/errors it fixes?

damianam:
We have `protobuf` pulled in as a dependency in our `Python` module. The version that `tensorflow` pulls in is incompatible with ours (slightly different API). So via `CPATH` our protobuf leaks into the `bazel` setup and compilation fails. Setting `wrapper_filter = ['protobuf']` in the easyconfig fixes it.

boegel:
@akesandgren Thoughts on this?

akesandgren:
It should do the same regardless of toolchain used or people will get confused
But wouldn't you need to filter LD_LIBRARY_PATHS too then?

damianam:
Agreed. We should filter `CPATH` in `gcc` toolchains too. Seems logical.

Regarding `LD_LIBRARY_PATH`, I don't think so. That is a runtime environment variable. `LIBRARY_PATH` on the other hand......

easybuild/easyblocks/t/tensorflow.py
```diff
@@ -99,10 +100,18 @@ def configure_step(self):
 
         tmpdir = tempfile.mkdtemp(suffix='-bazel-configure')
 
+        # filter out paths from CPATH and LIBRARY_PATH
+        package_filter = self.cfg['package_filter']
+        for var in ['CPATH', 'LIBRARY_PATH']:
+            path = os.getenv(var).split(':')
+            filtered_path = [p for fil in package_filter for p in path if fil not in p]
```
boegel: @damianam Please log the value of the environment variable before and after filtering.

easybuild/easyblocks/t/tensorflow.py
```diff
@@ -69,6 +69,7 @@ def extra_options():
         extra_vars = {
             # see https://developer.nvidia.com/cuda-gpus
             'cuda_compute_capabilities': [[], "List of CUDA compute capabilities to build with", CUSTOM],
+            'package_filter': [[], "List of paths to filter out", CUSTOM],
```
boegel: It's actually a list of filter *patterns* for paths  (since you're checking with `if fil not in p`), please clarify that and maybe also rename to `cpath_libpath_filter_patterns` to make it more explicit?

easybuild/easyblocks/t/tensorflow.py
```diff
@@ -99,10 +100,18 @@ def configure_step(self):
 
         tmpdir = tempfile.mkdtemp(suffix='-bazel-configure')
 
+        # filter out paths from CPATH and LIBRARY_PATH
```
boegel: please add an extra comment line or two explaining why this may be needed

easybuild/easyblocks/t/tensorflow.py
```diff
@@ -99,10 +100,18 @@ def configure_step(self):
...
+        package_filter = self.cfg['package_filter']
+        for var in ['CPATH', 'LIBRARY_PATH']:
+            path = os.getenv(var).split(':')
+            filtered_path = [p for fil in package_filter for p in path if fil not in p]
+            os.environ[var] = ':'.join(filtered_path)
```
boegel: please use `setvar` from `easybuild.tools.environment` (which logs), and use `os.pathsep`

```python
setvar(var, os.pathsep.join(filtered_path))

damianam:
@boegel @akesandgren please review again.

akesandgren:
Code looks sane, though I don't know what one would put in the filter. (Literally that is)

damianam:
Our easyconfig has this: `package_filter = ['protobuf']`. Without it the protobuf package we have included as part of the other stuff that gets pulled in by `Python` leaks in and breaks the compilation.

boegel:
Reviewed & tested with an existing `TensorFlow` easyconfig, lgtm, thanks @damianam!

PR #1437, opened 2018-06-15
make sure xmlpatterns always gets built with Qt

boegel:
fix for https://github.com/easybuilders/easybuild-easyconfigs/issues/6302

For some reason, `xmlpatterns` isn't always installed. Explicitly configuring for it seems to help though... This problem only seems to occur with `Qt` 4.x, the `Qt5` installations I have all include `bin/xmlpatterns`

WIP since all existing `Qt` easyconfigs need to be retested to ensure they all still work...

### DIFF
--- a/easybuild/easyblocks/q/qt.py
+++ b/easybuild/easyblocks/q/qt.py
@@ -81,6 +81,8 @@ class EB_Qt(ConfigureMake):
         else:
             raise EasyBuildError("Don't know which platform to set based on compiler family.")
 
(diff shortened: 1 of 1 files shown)


boegel:
no surprises after testing with existing `Qt` easyconfigs, good to go @damianam

damianam:
@boegel do you have the build reports for the existing `Qt` easyconfigs with this `easyblock`? Maybe it makes sense to start storing that for verification? It might be important as the codebase and number of contributors grow. Just food for thought, not that it is important now.

PR #1438, opened 2018-06-15
fix symlink check in NWChem easyblock + use change_dir/remove_file/symlink functions

boegel:
This is basically two fixes combined:

* i) Don't use `tempfile.gettempdir()` to determine location for temporary directory that is local to the system but can be a common target for the `$HOME/.nwchemrc` symlink across workernodes on a cluster. Since `tempfile.gettempdir()` follows `$TMPDIR` which is set to a job-unique directory by resource managers like Torque, the location obtained that way isn't common at all...

* ii) Fix the check to see whether an existing `$HOME/.nwchemrc` points to the right location. We can't just use `os.path.samefile`, since the symlink may be broken (in fact, it usually is). Resolving the symlink and directly comparing the locations fixes that.

(WIP since it still requires thorough testing)

### DIFF
--- a/easybuild/easyblocks/n/nwchem.py
+++ b/easybuild/easyblocks/n/nwchem.py
@@ -40,7 +40,7 @@ from distutils.version import LooseVersion
 from easybuild.easyblocks.generic.configuremake import ConfigureMake
 from easybuild.framework.easyconfig import CUSTOM
 from easybuild.tools.build_log import EasyBuildError
-from easybuild.tools.filetools import adjust_permissions, mkdir, write_file
+from easybuild.tools.filetools import adjust_permissions, change_dir, mkdir, remove_file, symlink, write_file
 from easybuild.tools.modules import get_software_libdir, get_software_root, get_software_version
 from easybuild.tools.run import run_cmd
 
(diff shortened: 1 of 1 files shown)


boegel:
tested with https://github.com/easybuilders/easybuild-easyconfigs/pull/6430 and existing `NWChem` easyconfigs, good to go imho

cc @damianam

easybuild/easyblocks/n/nwchem.py
```diff
@@ -98,12 +102,14 @@ def configure_step(self):
...
                                os.listdir(os.path.dirname(self.local_nwchemrc)))
 
-                if os.path.islink(self.home_nwchemrc) and not os.path.samefile(self.home_nwchemrc, self.local_nwchemrc):
-                    raise EasyBuildError("Found %s, but it's not a symlink to %s. "
-                                         "Please (re)move %s while installing NWChem; it can be restored later",
-                                         self.home_nwchemrc, self.local_nwchemrc, self.home_nwchemrc)
+                if os.path.islink(self.home_nwchemrc):
+                    home_nwchemrc_target = os.readlink(self.home_nwchemrc)
+                    if home_nwchemrc_target != self.local_nwchemrc:
+                        raise EasyBuildError("Found %s, but it's not a symlink to %s. "
```
damianam: Doesn't it make sense to move it and restore it later automatically?

boegel: I'm not sure... What if multiple `NWChem` builds are being done in parallel on the system for example?

Also, if there's a symlink in place that's different from what EB expects, it was most likely created by the user for a particular version of `NWChem`.

It's safer to be explicit about this imho.

damianam: Valid points.

PR #1439, opened 2018-06-15
add easyblock for CFDEMcoupling

boegel:


easybuild/easyblocks/c/cfdemcoupling.py
```diff
@@ -0,0 +1,159 @@
...
+            else:
+                raise EasyBuildError("Failed to find subdirectory for %s in %s %s", pkgname, self.builddir, top_dirs)
+
+        # always use env.setvar instead of os.putenv or os.environ for defining environment variables
+        env.setvar('CFDEM_VERSION', self.version)
+        env.setvar('CFDEM_PROJECT_DIR', self.cfdem_project_dir)
+
+        # define $CFDEM_PROJECT_USER_DIR to an empty existing directory
+        env.setvar('CFDEM_PROJECT_USER_DIR', os.path.join(self.builddir, 'CFDEM_PROJECT_USER_DIR'))
+        mkdir(os.getenv('CFDEM_PROJECT_USER_DIR'), parents=True)
```
damianam: Seems odd to set a variable and read it right after. Not that it matters a lot, but it is certainly not efficient. Doesn't it make more sense to define a python variable and use it in both places?

easybuild/easyblocks/c/cfdemcoupling.py
```diff
@@ -0,0 +1,159 @@
...
+        else:
+            raise EasyBuildError("Failed to isolate LPP-* directory in %s", self.builddir)
+
+        # build in parallel
+        env.setvar("WM_NCOMPPROCS", str(self.cfg['parallel']))
+
+        vtk_root = get_software_root('VTK')
+        if vtk_root:
+            vtk_ver_maj_min = '.'.join(get_software_version('VTK').split('.')[:2])
+            env.setvar('VTK_INC_USR', '-I%s' % os.path.join(vtk_root, 'include', 'vtk-%s' % vtk_ver_maj_min))
```
damianam: I think it makes sense to check if these directories exists and are not empty.

easybuild/easyblocks/c/cfdemcoupling.py
```diff
@@ -0,0 +1,159 @@
...
+            env.setvar('VTK_LIB_USR', '-L%s' % os.path.join(vtk_root, 'lib'))
+        else:
+            raise EasyBuildError("VTK not included as dependency")
+
+        # can't seem to use defined 'cfdemSysTest' alias...
```
damianam: I am not sure I understand the meaning of this comment.......

damianam:
@boegel ping

akesandgren:
@boegel ping since this blocks EC PR 6465 which is in the sprint...

easybuild/easyblocks/c/cfdemcoupling.py
```diff
@@ -0,0 +1,169 @@
...
+        else:
+            raise EasyBuildError("OpenFOAM not included as dependency")
+
+        # make sure expected additionalLibs_* file is available
+        addlibs_subdir = os.path.join('src', 'lagrangian', 'cfdemParticle', 'etc', 'addLibs_universal')
+        src_addlibs = os.path.join(self.cfdem_project_dir, addlibs_subdir, 'additionalLibs_%s.x' % openfoam_maj_ver)
+        target_addlibs = os.path.join(self.cfdem_project_dir, addlibs_subdir, 'additionalLibs_%s' % openfoam_ver)
+        copy_file(src_addlibs, target_addlibs)
+
+        # can't seem to use defined 'cfdemCompCFDEMall' alias...
```
akesandgren, line 128: Should this comment be similar as the one above or ?
I'm fine with it as it is, just checking if you missed it or left it on purpose.

easybuild/easyblocks/c/cfdemcoupling.py
```diff
@@ -0,0 +1,169 @@
...
+        """Custom sanity check for CFDEMcoupling."""
+        comp_fam = self.toolchain.comp_family()
+        if comp_fam == toolchain.GCC:
+            wm_compiler = 'Gcc'
+        elif comp_fam == toolchain.INTELCOMP:
+            wm_compiler = 'Icc'
+        else:
+            raise EasyBuildError("Unknown compiler family, don't know how to set WM_COMPILER")
+
+        psubdir = "linux64%sDPInt32Opt" % wm_compiler
```
akesandgren, line 146: It is preferable to handle debug being set and use Debug (or if it is Dbg) instead of Opt if it is.
Or is this not really built?

akesandgren: See the OpenFOAM easyblock for reference

easybuild/easyblocks/c/cfdemcoupling.py
boegel: I missed it, but do we need to clarify it in both places? I'd say it's fine as is

easybuild/easyblocks/c/cfdemcoupling.py
boegel: I wouldn't switch to `Dbg` if `eb --debug` is used (which is about debug logging, not debug builds).

akesandgren: I meant toolchainopt debug

boegel: I made the necessary changes to set `$WM_COMPILE_OPTION=Debug` if the `debug` toolchain option is set, but when using that the installation fails with:

```
software/binutils/2.28-GCCcore-6.4.0/bin/ld.gold: error: cannot find -lfiniteVolume
/software/binutils/2.28-GCCcore-6.4.0/bin/ld.gold: error: cannot find -lturbulenceModels
/software/binutils/2.28-GCCcore-6.4.0/bin/ld.gold: error: cannot find -lincompressibleTurbulenceModels
/software/binutils/2.28-GCCcore-6.4.0/bin/ld.gold: error: cannot find -lfvOptions
/software/binutils/2.28-GCCcore-6.4.0/bin/ld.gold: error: cannot find -lmeshTools
/software/binutils/2.28-GCCcore-6.4.0/bin/ld.gold: error: cannot find -llagrangian
/software/binutils/2.28-GCCcore-6.4.0/bin/ld.gold: error: --add-needed is not supported but is required for libvtkCommonCore-8.1.so.1 in /software/CFDEMcoupling/3.8.0-foss-2018a/CFDEMcoupling-3.8.0/platforms/linux64GccDPInt32Debug/lib/liblmp_auto-fulldebug.so                                                                                                                                                                                           collect2: error: ld returned 1 exit status
```

Also, I don't it's worth the effort at all to try and add support for debug builds in this case...

akesandgren: so lets ignore it...

akesandgren (approved):
LGTM

akesandgren:
Going in, thanks @boegel!

PR #1440, opened 2018-06-20
Adding support for --optarch=GENERIC for GROMACS

victorusu:
This PR allows the use of EASYBUILD_OPTARCH=GENERIC for GROMACS. It sets
the GMX_SIMD to 'SSE2' if the underlying software is a X86_64. This is a
good approximation since SSE2 were introduced in 2001 for intel CPUs and
in 2003 for AMD ones. This build would only break if one is compiling
for a >18 year old intel or >15 years old AMD.

For other architectures it sets the GMX_SIMD to 'None'

### DIFF
--- a/easybuild/easyblocks/g/gromacs.py
+++ b/easybuild/easyblocks/g/gromacs.py
@@ -103,7 +103,7 @@ class EB_GROMACS(CMakeMake):
             # According to [2] the performance difference between SSE2 and SSE4.1 is minor on x86
             # and SSE4.1 is not supported by AMD Magny-Cours[1].
             res = 'SSE2'
-        elif optarch == OPTARCH_GENERIC:
+        elif optarch == OPTARCH_GENERIC and LooseVersion(self.version) >= LooseVersion('5.0'):
             cpu_arch = get_cpu_architecture()
             if cpu_arch == X86_64:
                 res = 'SSE2'
--- a/easybuild/easyblocks/g/gromacs.py
+++ b/easybuild/easyblocks/g/gromacs.py
@@ -112,8 +112,10 @@ class EB_GROMACS(CMakeMake):
             else:
                 res = 'None'
         elif optarch:
-            print_warning("OPTARCH set to %s but not used."
-                          "Compiling GROMACS for the current host architecture", optarch)
+            warn_msg = "--optarch configuration setting set to %s but not taken into account; " % optarch
+            warn_msg += "compiling GROMACS for the current host architecture (i.e. the default behavior)"
+            self.log.warning(warn_msg)
+            print_warning(warn_msg)
 
         if res:
             self.log.info("Target architecture based on optarch configuration option ('%s'): %s", optarch, res)
--- a/easybuild/easyblocks/g/gromacs.py
+++ b/easybuild/easyblocks/g/gromacs.py
@@ -49,8 +49,7 @@ from easybuild.tools.filetools import download_file, extract_file, which
 from easybuild.tools.modules import get_software_libdir, get_software_root, get_software_version
 from easybuild.tools.run import run_cmd
 from easybuild.tools.toolchain.compiler import OPTARCH_GENERIC
-from easybuild.tools.systemtools import X86_64
-from easybuild.tools.systemtools import get_cpu_architecture, get_shared_lib_ext
+from easybuild.tools.systemtools import X86_64, get_cpu_architecture, get_shared_lib_ext
 
 
 class EB_GROMACS(CMakeMake):
--- a/easybuild/easyblocks/g/gromacs.py
+++ b/easybuild/easyblocks/g/gromacs.py
@@ -113,7 +113,6 @@ class EB_GROMACS(CMakeMake):
                 res = 'None'
         elif optarch:
             self.log.info("OPTARCH set to %s but not used. Compiling GROMACS for the current host architecture")
-            
 
         if res:
             self.log.info("Target architecture based on optarch configuration option ('%s'): %s", optarch, res)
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/g/gromacs.py
```diff
@@ -48,7 +48,9 @@
 from easybuild.tools.filetools import download_file, extract_file, which
 from easybuild.tools.modules import get_software_libdir, get_software_root, get_software_version
 from easybuild.tools.run import run_cmd
-from easybuild.tools.systemtools import get_platform_name , get_shared_lib_ext
+from easybuild.tools.toolchain.compiler import OPTARCH_GENERIC
+from easybuild.tools.systemtools import X86_64
+from easybuild.tools.systemtools import get_cpu_architecture, get_shared_lib_ext
```
boegel: @victorusu Please collapse this to a single line:

```python
from easybuild.tools.systemtools import X86_64, get_cpu_architecture, get_shared_lib_ext

File: easybuild/easyblocks/g/gromacs.py [hunk 2]
```diff
@@ -102,6 +104,12 @@ def get_gromacs_arch(self):
             # According to [2] the performance difference between SSE2 and SSE4.1 is minor on x86
             # and SSE4.1 is not supported by AMD Magny-Cours[1].
             res = 'SSE2'
+        elif optarch == OPTARCH_GENERIC:
+            cpu_arch = get_cpu_architecture()
+            if cpu_arch == X86_64:
+                res = 'SSE2'
+            else:
+                res = 'None'
```
boegel: @victorusu Is `-DGMX_SIMD=None` a correct setting? Maybe we should include a reference to the documentation for this in a comment above this line?

victorusu: @boegel, this is a link to the 'None' option http://manual.gromacs.org/documentation/2018/install-guide/index.html#simd-support

damianam:
If I understand the code properly, this easyblock is ignoring the case where `-xHost` or `-march=native` is used. In fact, it is ignoring any setting for GCC. Or am I missing something?

easybuild/easyblocks/g/gromacs.py
boegel: @victorusu Please include that link as a comment *in* the easyblock, for future reference

boegel:
@damianam We're checking the value for the value passed to `--optarch` here, where the default is empty. By default, `-xHost` is indeed used, but not via `--optarch`, `-xHost` gets included in `$CXXFLAGS` & co when the `optarch` toolchain option is set to True (unless it's overridden by `--optarch`, then the flags specified there are used instead).

As far as I know, with `--optarch` unspecfied, there's nothing to do: the `-xHost` included in `$CXXFLAGS` will ensure that GROMACS is properly optimized I think?

W.r.t. not taking into account the GCC equivalents: not sure what the best way would be to deal with that...

easybuild/easyblocks/g/gromacs.py
```diff
@@ -102,6 +103,12 @@ def get_gromacs_arch(self):
             # According to [2] the performance difference between SSE2 and SSE4.1 is minor on x86
             # and SSE4.1 is not supported by AMD Magny-Cours[1].
             res = 'SSE2'
+        elif optarch == OPTARCH_GENERIC and LooseVersion(self.version) >= LooseVersion('5.0'):
```
boegel: @victorusu This version check is overkill, this is already handled where `get_gromacs_arch` is called?

damianam:
You are right:

> By default, the GROMACS build system will detect the SIMD instruction set supported by the CPU architecture (on which the configuring is done), and thus pick the best available SIMD parallelization supported by GROMACS.

Regarding GCC: I guess we need an if statement to check which compiler is being used, and set `GMX_SIMD` depending on the `optarch` value.

boegel:
@damianam Let's open an issue for the latter, that's out of scope for this PR imho...

boegel:
@victorusu Maybe we do need to add a fallback `elif optarch` that emits a clear warning that a value for `optarch` was specified but wasn't dealt with properly, to make it explicit?

easybuild/easyblocks/g/gromacs.py
```diff
@@ -102,6 +105,14 @@ def get_gromacs_arch(self):
...
             # and SSE4.1 is not supported by AMD Magny-Cours[1].
             res = 'SSE2'
+        elif optarch == OPTARCH_GENERIC:
+            cpu_arch = get_cpu_architecture()
+            if cpu_arch == X86_64:
+                res = 'SSE2'
+            else:
+                res = 'None'
+        elif optarch:
+            self.log.info("OPTARCH set to %s but not used. Compiling GROMACS for the current host architecture")
```
boegel: @victorusu Maybe using `print_warning` makes more sense here, to make it pop up in the output of `eb` (rather than only in the log file)?

Also, the `%s` isn't being templated... And I would use `--optarch configuration option` rather than `OPTARCH`

easybuild/easyblocks/g/gromacs.py
```diff
@@ -102,6 +105,15 @@ def get_gromacs_arch(self):
...
             res = 'SSE2'
+        elif optarch == OPTARCH_GENERIC:
+            cpu_arch = get_cpu_architecture()
+            if cpu_arch == X86_64:
+                res = 'SSE2'
+            else:
+                res = 'None'
+        elif optarch:
+            print_warning("OPTARCH set to %s but not used."
+                          "Compiling GROMACS for the current host architecture", optarch)
```
boegel: @victorusu You need to use `%` to format the string when using `print_warning` (while just passing the variables as argument does work with `self.log`...).

We can enhance `print_warning` to also handle this, but as it is now you'll run into a crash, so please change.

Also, I'd both print & log (since the printed message will otherwise not be included in the log...)

```python
warn_msg = "--optarch configuration setting set to %s but not taken into account; " % optarch
warn_msg += "compiling GROMACS for the current host architecture (i.e. the default behavior)"
self.log.warning(warn_msg)
print_warning(warn_msg)

PR #1441, opened 2018-06-20
enhance Python easyblock to add support for checking stack limit and setting it to 'unlimited' if possible

boegel:
fix for https://github.com/easybuilders/easybuild-easyconfigs/issues/6484

In most environment where the Intel compiler issues occur, this will enable dancing around it by setting `ulimit_unlimited = True` in the `Python` 3.x easyconfig file...

If `ulimit -s unlimited` can't be set, a clear warning will be printed to inform the `eb` user that this will likely result in compilation errors (which will be caught by the enhanced sanity check).

Disabled by default since this is only relevant to Python 3.6.x and a certain range of Intel compiler versions.

### DIFF
--- a/easybuild/easyblocks/p/python.py
+++ b/easybuild/easyblocks/p/python.py
@@ -170,8 +170,10 @@ class EB_Python(ConfigureMake):
                 self.cfg.update('prebuildopts', "ulimit -s %s && " % UNLIMITED)
             else:
                 msg = "Current stack size limit is %s, and can not be set to %s due to hard limit of %s;"
+                msg += " setting stack size limit to %s instead, "
                 msg += " this may break part of the compilation (e.g. hashlib)..."
-                print_warning(msg % (curr_ulimit_s, UNLIMITED, max_ulimit_s))
+                print_warning(msg % (curr_ulimit_s, UNLIMITED, max_ulimit_s, max_ulimit_s))
+                self.cfg.update('prebuildopts', "ulimit -s %s && " % max_ulimit_s)
 
         super(EB_Python, self).build_step(*args, **kwargs)
 
--- a/easybuild/easyblocks/p/python.py
+++ b/easybuild/easyblocks/p/python.py
@@ -166,7 +166,7 @@ class EB_Python(ConfigureMake):
                 self.log.info("Current stack size limit is %s: OK", curr_ulimit_s)
             elif max_ulimit_s == UNLIMITED:
                 self.log.info("Current stack size limit is %s, setting it to %s for build...",
-                              UNLIMITED, curr_ulimit_s)
+                              curr_ulimit_s, UNLIMITED)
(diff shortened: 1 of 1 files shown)


vanzod:
Tested with `Python-3.6.4-intel-2018a.eb` and fails with:
```
== FAILED: Installation ended unsuccessfully (build directory: /mnt/ramdisk/Python/3.6.4/intel-2018a): build failed (first 300 chars): Found one or more errors in output of python -c 'import hashlib': ERROR:root:code for hash sha3_224 was not found.
  File "/opt/easybuild/software/MPI/intel/2018.1.163-GCC-6.4.0-2.28/impi/2018.1.163/Python/3.6.4/lib/python3.6/hashlib.py", line 121, in __get_openssl
== Results of the build can be found in the log file(s) /tmp/eb-X4LBW3/easybuild-Python-3.6.4-20180621.090409.LqdIB.log
ERROR: Build of /home/davide/Python-3.6.4-intel-2018a.eb failed (err: 'build failed (first 300 chars): Found one or more errors in output of python -c \'import hashlib\': ERROR:root:code for hash sha3_224 was not found.\nTraceback (most recent call last):\n  File "/opt/easybuild/software/MPI/intel/2018.1.163-GCC-6.4.0-2.28/impi/2018.1.163/Python/3.6.4/lib/python3.6/hashlib.py", line 121, in __get_openssl')```

File: easybuild/easyblocks/p/python.py [hunk 1]
```diff
@@ -137,6 +149,34 @@ def configure_step(self):
...
+            # figure out hard limit for stack size limit;
+            # this determines whether or not we can use "ulimit -s unlimited"
+            (out, _) = run_cmd("ulimit -s -H")
+            max_ulimit_s = out.strip()
+
+            if curr_ulimit_s == UNLIMITED:
+                self.log.info("Current stack size limit is %s: OK", curr_ulimit_s)
+            elif max_ulimit_s == UNLIMITED:
+                self.log.info("Current stack size limit is %s, setting it to %s for build...",
+                              UNLIMITED, curr_ulimit_s)
```
vanzod: The order of the two variables should be inverted:
`
self.log.info("Current stack size limit is %s, setting it to %s for build...",
                    curr_ulimit_s, UNLIMITED)
`

vanzod:
The easyblock works correctly with `Python-3.6.4-intel-2018a.eb` and a non-unlimited stack size.
Also it correctly catches the `hashlib` error.

easybuild/easyblocks/p/python.py
boegel: Nice catch, fixed now...

boegel:
Thanks a lot for testing & confirming the fix @vanzod!

This is good to go imho (once Travis gives the green light after the tiny bug fix)

PR #1442, opened 2018-06-22
Ipp

damianam:
`NLSPATH` with idb has been an useless variable for a long time. At least since 2015. More importantly, the `ipp` easyblock has never set correctly the library paths, as the ipp libraries are in `ipp/lib/intel64`, not in `lib/intel64` (which also exists and has some extra libraries)

### DIFF
--- a/easybuild/easyblocks/generic/intelbase.py
+++ b/easybuild/easyblocks/generic/intelbase.py
@@ -434,12 +434,6 @@ class IntelBase(EasyBlock):
             txt += self.module_generator.prepend_paths(self.license_env_var, [self.license_file],
                                                        allow_abs=True, expand_relpaths=False)
 
-        if self.cfg['m32']:
-            nlspath = os.path.join('idb', '32', 'locale', '%l_%t', '%N')
-        else:
-            nlspath = os.path.join('idb', 'intel64', 'locale', '%l_%t', '%N')
-        txt += self.module_generator.prepend_paths('NLSPATH', nlspath)
-
         return txt
 
     def cleanup_step(self):
--- a/easybuild/easyblocks/i/ipp.py
+++ b/easybuild/easyblocks/i/ipp.py
@@ -112,7 +112,7 @@ class EB_ipp(IntelBase):
         custom_paths = {
             'files': [
                 os.path.join('ipp', 'lib', 'intel64', 'libipp%s') % y for x in ipp_libs
-                    for y in ['%s.a' % x, '%s.%s' % (x, shlib_ext)]
+                for y in ['%s.a' % x, '%s.%s' % (x, shlib_ext)]
             ],
             'dirs': dirs,
         }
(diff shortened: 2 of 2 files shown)


easybuild/easyblocks/generic/intelbase.py
```diff
@@ -434,12 +434,6 @@ def make_module_extra(self, *args, **kwargs):
             txt += self.module_generator.prepend_paths(self.license_env_var, [self.license_file],
                                                        allow_abs=True, expand_relpaths=False)
 
-        if self.cfg['m32']:
-            nlspath = os.path.join('idb', '32', 'locale', '%l_%t', '%N')
-        else:
-            nlspath = os.path.join('idb', 'intel64', 'locale', '%l_%t', '%N')
-        txt += self.module_generator.prepend_paths('NLSPATH', nlspath)
-
```
akesandgren: The debugger is nowadays a version of gdb and the locale is under
debugger_2018/gdb/intel64/share/locale/

But I think we can scrap the NLSPATH completely as above for debugger.

@boegel @damianam Thoughts?

easybuild/easyblocks/i/ipp.py
```diff
@@ -109,12 +109,12 @@ def make_module_req_guess(self):
         guesses = super(EB_ipp, self).make_module_req_guess()
 
         if LooseVersion(self.version) >= LooseVersion('9.0'):
-            lib_path = os.path.join('lib', self.arch)
+            lib_path = [os.path.join('ipp/lib', self.arch), os.path.join('lib', self.arch)]
             include_path = 'ipp/include'
```
akesandgren: While we're at it. Use os.path.join all the way, i.e.
lib_path = [os.path.join('ipp', 'lib', self.arch), os.path.join('lib', self.arch)]
include_path = os.path.join('ipp', 'include')

easybuild/easyblocks/generic/intelbase.py
damianam: I agree

easybuild/easyblocks/i/ipp.py
```diff
@@ -83,20 +84,36 @@ def sanity_check_step(self):
...
+            ]
         else:
-            dirs = ['composerxe/lib/intel64', 'ipp/bin', 'ipp/include',
-                    'ipp/tools/intel64']
+            dirs = [
+                os.path.join('composerxe', 'lib', 'intel64'),
+                os.path.join('ipp', 'bin'),
+                os.path.join('ipp', 'include'),
+                os.path.join('ipp', 'tools', 'intel64')
+            ]
```
akesandgren: I would write this slightly different. Something like
dirs = []
ipp_dir_list = ['bin', 'include', os.path.join('tools', 'intel64')]
if LooseVersion ...
  add the version specific ones to dirs and ipp_dir_list
dirs.extend(os.path.join('ipp', d)) for d in ipp_dir_list
(At least i think it is extend one should use, but i usually never get that right)

PR #1447, opened 2018-07-01
Scipion easyblock (WIP)

akesandgren:
An initial easyblock to build Scipion.

### DIFF
--- a/easybuild/easyblocks/s/scipion.py
+++ b/easybuild/easyblocks/s/scipion.py
@@ -1,0 +1,167 @@
+##
+# Copyright 2018 Ghent University
+#
+# This file is part of EasyBuild,
+# originally created by the HPC team of Ghent University (http://ugent.be/hpc/en),
+# with support of Ghent University (http://ugent.be/hpc),
+# the Flemish Supercomputer Centre (VSC) (https://www.vscentrum.be),
+# Flemish Research Foundation (FWO) (http://www.fwo.be/en)
+# and the Department of Economy, Science and Innovation (EWI) (http://www.ewi-vlaanderen.be/en).
+#
+# https://github.com/easybuilders/easybuild
+#
+# EasyBuild is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation v2.
+#
+# EasyBuild is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with EasyBuild.  If not, see <http://www.gnu.org/licenses/>.
+##
+"""
+EasyBuild support for building and installing Scipion, implemented as an easyblock
+
+@author: Ake Sandgren (HPC2N, Umea University)
+"""
+import glob
+import os
+import re
+import shutil
+
+import easybuild.tools.environment as env
+from easybuild.easyblocks.generic.scons import SCons
+from easybuild.tools.filetools import apply_regex_substitutions, change_dir, mkdir, symlink
+from easybuild.tools.modules import get_software_root
+from easybuild.tools.run import run_cmd
+
+class EB_Scipion(SCons):
+    """Support for building/installing Scipion."""
+
+    def __init__(self, *args, **kwargs):
+        """Initialize Scipion-specific variables."""
+        super(EB_Scipion, self).__init__(*args, **kwargs)
+
+    def configure_step(self):
+        """Custom configuration procedure for Scipion."""
+        # Setup scipion.conf, protocols.conf, and host.conf
+        # scipion.conf needs to be configured with paths to all the
+        # dependencies, stupid but that's the way it works at the
+        # moment.
+
+        # Create the config files and then make the subtitutions
+        cmd = './scipion --config %s/config/scipion.conf config' % self.installdir
+        run_cmd(cmd, log_all=True, simple=True)
+
+        # Mapping of EB software names and variables to patch in
+        # scipion.conf
+        map_sw2var_name = {
+            'ctffind': 'CTFFIND4_HOME',
+            'EMAN2': 'EMAN2DIR',
+            'frealign': 'FREALIGN_HOME',
+            'MotionCor2': 'MOTIONCOR2_HOME',
+            'RELION': 'RELION_HOME',
+            'Java': 'JAVA_HOME',
+            'VMD': 'VMD_HOME',
+            'MATLAB': 'MATLAB_DIR',
+        }
+
+        # Some other variables we need to change
+        regex_subs = [
+            (r'^CC = .*$', r'CC = %s' % os.getenv('CC')),
+            (r'^CXX = .*$', r'CXX = %s' % os.getenv('CXX')),
+            (r'^LINKERFORPROGRAMS = .*$', r'LINKERFORPROGRAMS = %s' % os.getenv('CXX')),
+            (r'^MPI_CC = .*$', r'MPI_CC = %s' % os.getenv('MPICC')),
+            (r'^MPI_CXX = .*$', r'MPI_CXX = %s' % os.getenv('MPICXX')),
+            (r'^MPI_LINKERFORPROGRAMS = .*$', r'MPI_LINKERFORPROGRAMS = %s' % os.getenv('MPICXX')),
+            (r'^MPI_LIBDIR = .*$', r'MPI_LIBDIR = /'),
+            (r'^MPI_INCLUDE = .*$', r'MPI_INCLUDE = /'),
+            (r'^MPI_BINDIR = .*$', r'MPI_BINDIR = /'),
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/s/scipion.py
```diff
@@ -0,0 +1,165 @@
...
+from easybuild.tools.modules import get_software_root
+from easybuild.tools.run import run_cmd
+
+
+class EB_Scipion(SCons):
+    """Support for building/installing Scipion."""
+
+    def __init__(self, *args, **kwargs):
+        """Initialize Scipion-specific variables."""
+        super(EB_Scipion, self).__init__(*args, **kwargs)
```
boegel: if nothing custom is done in `__init__`, you can drop this

easybuild/easyblocks/s/scipion.py
```diff
@@ -0,0 +1,165 @@
...
+        regex_subs = [
+            (r'^CC = .*$', r'CC = %s' % os.getenv('CC')),
+            (r'^CXX = .*$', r'CXX = %s' % os.getenv('CXX')),
+            (r'^LINKERFORPROGRAMS = .*$', r'LINKERFORPROGRAMS = %s' % os.getenv('CXX')),
+            (r'^MPI_CC = .*$', r'MPI_CC = %s' % os.getenv('MPICC')),
+            (r'^MPI_CXX = .*$', r'MPI_CXX = %s' % os.getenv('MPICXX')),
+            (r'^MPI_LINKERFORPROGRAMS = .*$', r'MPI_LINKERFORPROGRAMS = %s' % os.getenv('MPICXX')),
+            (r'^MPI_LIBDIR = .*$', r'MPI_LIBDIR = /'),
+            (r'^MPI_INCLUDE = .*$', r'MPI_INCLUDE = /'),
+            (r'^MPI_BINDIR = .*$', r'MPI_BINDIR = /'),
```
boegel: these three look a bit strange, can you clarify with a comment above?

easybuild/easyblocks/s/scipion.py
```diff
@@ -0,0 +1,165 @@
...
+            # The EB motioncor2 build renames that binary to an explicit name
+            (r'^MOTIONCOR2_BIN = .*$', r'MOTIONCOR2_BIN = motioncor2'),
+        ]
+        for (k, v) in map_sw2var_name.items():
```
boegel: please avoid single-letter variables outside of list comprehensions, use `(key, val)` instead

easybuild/easyblocks/s/scipion.py
```diff
@@ -0,0 +1,165 @@
...
+            swroot = get_software_root(k)
+            if swroot:
+                regex_subs.extend([(r'^%s = .*$' % v, r'%s = %s' % (v, swroot))])
+        cuda_root = get_software_root('CUDA')
+        if cuda_root:
+            regex_subs.extend([(r'^(.*CUDA_LIB) = .*$', r'\1 = %s/lib64' % (cuda_root))])
+            regex_subs.extend([(r'^CUDA_BIN = .*$', r'CUDA_BIN = %s/bin' % (cuda_root))])
+            regex_subs.extend([(r'^CUDA = .*$', r'CUDA = True')])
```
boegel: no `else` to set `CUDA = False`?

easybuild/easyblocks/s/scipion.py
```diff
@@ -0,0 +1,165 @@
...
+            opencv = 'True'
+        else:
+            opencv = 'False'
+        regex_subs.extend([(r'^OPENCV = .*$', r'OPENCV = %s' % opencv)])
+
+        if get_software_root('MATLAB'):
+            matlab = 'True'
+        else:
+            matlab = 'False'
+        regex_subs.extend([(r'^MATLAB = .*$', r'MATLAB = %s' % matlab)])
```
boegel: please avoid duplicate code, and you can make this quite a bit shorter using `bool(...)`:

```python
for dep in ['MATLAB', 'OpenCV']:
    use_dep = bool(get_software_root(dep))
    regex_subs.append((r'^%s = .*$' % dep.upper(), r'%s = %s' % (dep.upper(), use_dep)))

File: easybuild/easyblocks/s/scipion.py [hunk 6]
```diff
@@ -0,0 +1,165 @@
...
+
+        apply_regex_substitutions('%s/config/scipion.conf' % self.installdir, regex_subs)
+
+    def setup_env(self):
+        # Need to setup some environemt variables before running SCons
```
boegel: missing docstring for `setup_env`

maybe rename it to something more specific like `setup_scipion_env` to avoid a potential clash with a general method defined in `EasyBlock` class

easybuild/easyblocks/s/scipion.py
```diff
@@ -0,0 +1,165 @@
...
+        env.setvar('SCIPION_HOME', self.installdir)
+        env.setvar('SCIPION_CWD', self.builddir)
+        env.setvar('SCIPION_VERSION', self.version)
+        env.setvar('SCIPION_CONFIG', '%s/config/scipion.conf' % self.installdir)
```
boegel: please use `os.path.join` when composing paths

```python
env.setvar('SCIPION_CONFIG', os.path.join(self.installdir, 'config', 'scipion.conf'))

File: easybuild/easyblocks/s/scipion.py [hunk 8]
```diff
@@ -0,0 +1,165 @@
...
+        env.setvar('SCIPION_LOCAL_CONFIG', '%s/config/scipion.conf' % self.installdir)
+        env.setvar('SCIPION_PROTOCOLS', '%s/config/protocols.conf' % self.installdir)
+        env.setvar('SCIPION_HOSTS', '%s/config/hosts.conf' % self.installdir)
+        # Read this from scipion.conf
```
boegel: is this a TODO?

easybuild/easyblocks/s/scipion.py
```diff
@@ -0,0 +1,165 @@
...
+        env.setvar('MPI_CXX', os.getenv('MPICXX'))
+        env.setvar('MPI_LINKERFORPROGRAMS', os.getenv('MPICXX'))
+        java_root = get_software_root('Java')
+        env.setvar('JNI_CPPPATH', '%s/include:%s/include/linux' % (java_root, java_root))
+        env.setvar('JAVAC', 'javac')
+        env.setvar('JAR', 'jar')
+
+    def build_step(self):
+        """Custom build step for Scipion."""
+        self.setup_env()
```
boegel: Why not call this once in `configure_step`?

Then you don't need to customize `build_step` at all (the environment isn't reset between steps).

easybuild/easyblocks/s/scipion.py
```diff
@@ -0,0 +1,165 @@
...
+
+        super(EB_Scipion, self).build_step()
+
+    def install_step(self):
+        """Custom install step for Scipion."""
+        self.setup_env()
```
boegel: no need for this if already called in `configure_step`?

easybuild/easyblocks/s/scipion.py
```diff
@@ -0,0 +1,165 @@
...
+        mkdir(bindir)
+        cwd = change_dir(bindir)
+        symlink(os.path.join('..', 'scipion'), 'scipion', use_abspath_source=False)
+        change_dir(cwd)
+
+    def sanity_check_step(self):
+        """Custom sanity check for Scipion."""
+
+        custom_paths = {
+            'files': ['software/em/xmipp/bin/xmipp_volume_validate_pca'],
```
boegel: also check for `bin/scipion`?

akesandgren:
Doh, didn't search for a pre-existing easyblock...

easybuild/easyblocks/s/scipion.py
akesandgren: They should go away in the final version probably. I used them to detect when scipion was doing bad MPI stuff.

easybuild/easyblocks/s/scipion.py
akesandgren: yes.

boegel: OK, I marked this PR as `WIP` then to reflect that...

easybuild/easyblocks/s/scipion.py
boegel: Ah, ok, then please trash them :)

easybuild/easyblocks/s/scipion.py
akesandgren: Oops, forgot that, did remember to mark the easyconfig as WIP though :-)

akesandgren:
A new easyblock that will handle Scipion V2 is in the works.

akesandgren:
Closing this is favor of easyblock that deals with V2

runiq:
> Closing this is favor of easyblock that deals with V2

@akesandgren Just for the record, is #1837 the easyblock you are referring to?

akesandgren:
No, 1837 is a new easyblock for the now standalone Xmipp. Scipion v2 is not PR:ed yet but will be later today I hope.

akesandgren:
The new easyblock for scipion is https://github.com/easybuilders/easybuild-easyblocks/pull/1847

runiq:
Thank you!

PR #1451, opened 2018-07-05
allow skipping tests when installing Perl extension, by setting 'runtest' to False

boegel:


akesandgren:
So self.cfg['runtest'] is "test" by default?

boegel:
@akesandgren Yes, see line 48 in `perlmodule.py`, or the output of `eb -a -e PerlModule | grep runtest`

PR #1452, opened 2018-07-05
Added support for installing intel products using serial numbers

damianam:


easybuild/easyblocks/generic/intelbase.py
```diff
@@ -253,35 +255,40 @@ def prepare_step(self, *args, **kwargs):
...
-                                                                  lic_specs=license_specs)
-
-            if lic_specs:
-                if self.license_env_var is None:
-                    self.log.info("Using Intel license specifications from 'license_file': %s", lic_specs)
-                    self.license_env_var = default_lic_env_var
+        if self.serial_number:
+            self.log.info("Using provided serial number (%s) and ignoring other licenses", self.serial_number)
+        else:
+            if self.requires_runtime_license:
```
boegel: @damianam this could be an `elif` (which also avoids all the extra indentation)?

```python
elif self.requires_runtime_license:
    ...

boegel:
Thanks @damianam!

Issue #1445, opened 2018-06-29
The TensorFlow easyblock should require cuDNN when CUDA is enabled


verdurin:
At the moment the TF easyblock does not enforce the presence of `cuDNN` when `CUDA` is enabled.

boegel:
@akesandgren Up for this?

akesandgren:
Would a EasyBuildError("TensorFlow have a strict dependency on cuDNN if CUDA is enabled")
message be ok, or do you want another text.

boegel:
@akesandgren with `s/have/has/g`, ok for me

akesandgren:
Solved, together with a couple of other minor things, in PR #1453

PR #1453, opened 2018-07-06
Multiple updates to Tensorflow easyblock

akesandgren:
- Make cuDNN a strict requirement when CUDA is enabled.
- Handle building -rc versions of TernsorFlow.
- mkl-dnn can be built with cuDNN enabled without problems.
- Enable mkl-dnn by default.
- Add support for v1.10
- Add TensorRT support (WIP)
- Add NCCL support (WIP)

Solves issue #1445

### DIFF
--- a/easybuild/easyblocks/t/tensorflow.py
+++ b/easybuild/easyblocks/t/tensorflow.py
@@ -139,6 +139,7 @@ class EB_TensorFlow(PythonPackage):
         cudnn_root = get_software_root('cuDNN')
         opencl_root = get_software_root('OpenCL')
         tensorrt_root = get_software_root('TensorRT')
+        nccl_root = get_software_root('NCCL')
 
         use_mpi = self.toolchain.options.get('usempi', False)
 
@@ -177,6 +178,16 @@ class EB_TensorFlow(PythonPackage):
                 })
             else:
                 raise EasyBuildError("TensorFlow has a strict dependency on cuDNN if CUDA is enabled")
+            if nccl_root:
+                nccl_version = get_software_version('NCCL')
+                config_env_vars.update({
+                    'NCCL_INSTALL_PATH': nccl_root,
+                })
+            else:
+                nccl_version = '1.3'  # Use simple downloadable version
+            config_env_vars.update({
+                'TF_NCCL_VERSION': nccl_version,
+            })
 
         for (key, val) in sorted(config_env_vars.items()):
             env.setvar(key, val)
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/t/tensorflow.py
```diff
@@ -281,11 +283,6 @@ def build_step(self):
         if cuda_root:
             cmd.append('--config=cuda')
 
-        # enable mkl-dnn by default, but only if cuDNN is not listed as dependency
-        if self.cfg['with_mkl_dnn'] is None and get_software_root('cuDNN') is None:
-            self.log.info("Enabling use of mkl-dnn since cuDNN is not listed as dependency")
-            self.cfg['with_mkl_dnn'] = True
```
boegel: @akesandgren Can you clarify this? Can both be used together? There must have been a reason why did this?

akesandgren: We did this because i originally thought there would be a problem, but I later tried and they can be compiled in at the same time.
So it's basically removing a bit of code that shouldn't have been there in the first place.

akesandgren:
Verified to work with old TensorFlow-1.5.0-foss-2017b-Python-3.6.3.eb

akesandgren:
Tested with TensorFlow-1.5.0-foss-2017b-Python-3.6.3.eb and https://github.com/easybuilders/easybuild-easyconfigs/pull/6677 (TensorFlow-1.10.0-fosscuda-2018b-Python-2.7.15.eb)

easybuild/easyblocks/t/tensorflow.py
```diff
@@ -165,10 +171,22 @@ def configure_step(self):
...
+                })
+            else:
+                raise EasyBuildError("TensorFlow has a strict dependency on cuDNN if CUDA is enabled")
+            if nccl_root:
+                nccl_version = get_software_version('NCCL')
+                config_env_vars.update({
+                    'NCCL_INSTALL_PATH': nccl_root,
+                })
+            else:
+                nccl_version = '1.3'  # Use simple downloadable version
```
boegel: @akesandgren Where did you get this `1.3` from? Isn't this something that will vary across TF versions?

akesandgren: 1.3 is the fallback version if there is no external NCCL v2.x installed (requires manual download).
1.3 can be auto-downloaded and is the version they mention in all TF versions from 1.4, i think, as the default if no external NCCL is available.

But, from TF 1.10 they have changed the Q&A stuff so one now must specify it.

akesandgren:
No more changes in sight, gfood to go from my side.

PR #1455, opened 2018-07-08
define $LDSHARED when installing Python packages when Python's value doesn't use toolchain compiler ($CC)

Micket:
Fix for setuptools at GCCcore level for easybuilders/easybuild-easyconfigs#6537

### DIFF
--- a/easybuild/easyblocks/generic/pythonpackage.py
+++ b/easybuild/easyblocks/generic/pythonpackage.py
@@ -453,7 +453,10 @@ class PythonPackage(ExtensionEasyBlock):
             curr_cc = os.getenv('CC')
             python_ldshared = get_config_vars('LDSHARED')[0]
             if python_ldshared:
-                if python_ldshared.split(' ')[0] != curr_cc:
+                if python_ldshared.split(' ')[0] == curr_cc:
+                    self.log.info("Python's value for $LDSHARED ('%s') uses current $CCi value ('%s'), so not touching it",
+                                  python_ldshared, curr_cc)
+                else:
                     self.log.info("Python's value for $LDSHARED ('%s') doesn't use current $CC value ('%s'), fixing",
                                   python_ldshared, curr_cc)
                     env.setvar("LDSHARED", curr_cc + " -shared")
--- a/easybuild/easyblocks/generic/pythonpackage.py
+++ b/easybuild/easyblocks/generic/pythonpackage.py
@@ -431,6 +431,24 @@ class PythonPackage(ExtensionEasyBlock):
             except IOError:
                 raise EasyBuildError("Creating %s failed", self.sitecfgfn)
 
+        # ensure that LDSHARED uses CC
+        curr_cc = os.getenv('CC')
+        python_ldshared = get_config_vars('LDSHARED')[0]
+        if python_ldshared and curr_cc:
+            if python_ldshared.split(' ')[0] == curr_cc:
+                self.log.info("Python's value for $LDSHARED ('%s') uses current $CC value ('%s'), not touching it",
+                              python_ldshared, curr_cc)
+            else:
+                self.log.info("Python's value for $LDSHARED ('%s') doesn't use current $CC value ('%s'), fixing",
+                              python_ldshared, curr_cc)
+                env.setvar("LDSHARED", curr_cc + " -shared")
+        else:
+            if curr_cc:
+                self.log.info("No $LDSHARED found for Python, setting to '%s -shared'", curr_cc)
+                env.setvar("LDSHARED", curr_cc + " -shared")
+            else:
+                self.log.info("No value set for $CC, so not touching $LDSHARED either")
+
         # creates log entries for python being used, for debugging
         run_cmd("%s -V" % self.python_cmd, verbose=False, trace=False)
         run_cmd("%s -c 'import sys; print(sys.executable)'" % self.python_cmd, verbose=False, trace=False)
@@ -449,23 +467,6 @@ class PythonPackage(ExtensionEasyBlock):
                 library_paths = os.pathsep.join(self.toolchain.get_variable("LDFLAGS", list))
                 env.setvar("CMAKE_INCLUDE_PATH", include_paths)
                 env.setvar("CMAKE_LIBRARY_PATH", library_paths)
-
-            curr_cc = os.getenv('CC')
-            python_ldshared = get_config_vars('LDSHARED')[0]
-            if python_ldshared and curr_cc:
-                if python_ldshared.split(' ')[0] == curr_cc:
-                    self.log.info("Python's value for $LDSHARED ('%s') uses current $CC value ('%s'), not touching it",
-                                  python_ldshared, curr_cc)
-                else:
-                    self.log.info("Python's value for $LDSHARED ('%s') doesn't use current $CC value ('%s'), fixing",
-                                  python_ldshared, curr_cc)
-                    env.setvar("LDSHARED", curr_cc + " -shared")
-            else:
-                if curr_cc:
-                    self.log.info("No $LDSHARED found for Python, setting to '%s -shared'", curr_cc)
-                    env.setvar("LDSHARED", curr_cc + " -shared")
-                else:
-                    self.log.info("No value set for $CC, so not touching $LDSHARED either")
 
             cmd = ' '.join([self.cfg['prebuildopts'], self.python_cmd, 'setup.py', self.cfg['buildcmd'],
                             self.cfg['buildopts']])
--- a/easybuild/easyblocks/generic/pythonpackage.py
+++ b/easybuild/easyblocks/generic/pythonpackage.py
@@ -452,7 +452,8 @@ class PythonPackage(ExtensionEasyBlock):
             curr_cc = os.getenv('CC')
             python_ldshared = get_config_vars('LDSHARED')[0].split(' ')
             if python_ldshared[0] != curr_cc:
-                self.log.info("Python's value for $LDSHARED ('%s') doesn't use current $CC value ('%s'), fixing that...", python_ldshared, curr_cc)
+                self.log.info("Python's value for $LDSHARED ('%s') doesn't use current $CC value ('%s'), fixing",
+                              python_ldshared, curr_cc)
                 env.setvar("LDSHARED", curr_cc + " -shared")
 
             cmd = ' '.join([self.cfg['prebuildopts'], self.python_cmd, 'setup.py', self.cfg['buildcmd'],
--- a/easybuild/easyblocks/generic/pythonpackage.py
+++ b/easybuild/easyblocks/generic/pythonpackage.py
@@ -454,7 +454,7 @@ class PythonPackage(ExtensionEasyBlock):
             python_ldshared = get_config_vars('LDSHARED')[0]
             if python_ldshared:
                 if python_ldshared.split(' ')[0] == curr_cc:
-                    self.log.info("Python's value for $LDSHARED ('%s') uses current $CCi value ('%s'), so not touching it",
+                    self.log.info("Python's value for $LDSHARED ('%s') uses current $CC value ('%s'), not touching it",
                                   python_ldshared, curr_cc)
                 else:
                     self.log.info("Python's value for $LDSHARED ('%s') doesn't use current $CC value ('%s'), fixing",
--- a/easybuild/easyblocks/generic/pythonpackage.py
+++ b/easybuild/easyblocks/generic/pythonpackage.py
@@ -36,6 +36,7 @@ import re
 import sys
 import tempfile
 from distutils.version import LooseVersion
+from distutils.sysconfig import get_config_vars
 from vsc.utils import fancylogger
 from vsc.utils.missing import nub
 
@@ -450,10 +451,14 @@ class PythonPackage(ExtensionEasyBlock):
                 env.setvar("CMAKE_LIBRARY_PATH", library_paths)
 
             curr_cc = os.getenv('CC')
-            python_ldshared = get_config_vars('LDSHARED')[0].split(' ')
-            if python_ldshared[0] != curr_cc:
-                self.log.info("Python's value for $LDSHARED ('%s') doesn't use current $CC value ('%s'), fixing",
-                              python_ldshared, curr_cc)
+            python_ldshared = get_config_vars('LDSHARED')[0]
+            if python_ldshared:
+                if python_ldshared.split(' ')[0] != curr_cc:
+                    self.log.info("Python's value for $LDSHARED ('%s') doesn't use current $CC value ('%s'), fixing",
+                                  python_ldshared, curr_cc)
+                    env.setvar("LDSHARED", curr_cc + " -shared")
+            else:
+                self.log.info("No LDSHARED found for Python, setting to '%s -shared'", curr_cc)
                 env.setvar("LDSHARED", curr_cc + " -shared")
 
             cmd = ' '.join([self.cfg['prebuildopts'], self.python_cmd, 'setup.py', self.cfg['buildcmd'],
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/generic/pythonpackage.py
```diff
@@ -446,6 +446,8 @@ def build_step(self):
                 env.setvar("CMAKE_INCLUDE_PATH", include_paths)
                 env.setvar("CMAKE_LIBRARY_PATH", library_paths)
 
+            env.setvar("LDSHARED", "%s -shared" % self.toolchain.get_variable("CC", str))
```
damianam: AFAIK this is necessary just when compiling python packages with `icc`, when python has been compiled in `GCCcore`. Possibly the same is true when compiling with PGI, but I've never tried it. I am basically worried about introducing side effects when applying this to every situation. I would suggest to include a check that verifies the compiler used in the toolchain of `Python` or `Python-core` and the compiler used in the toolchain being used to compile this particular package, and add `LDSHARED` just when both differ.

Micket: Well, yes. By default, Pythons distutils will take the linker name from when the interpreter was compiled, regardless of what CC is set to in the package currently being built.

From what I can see in distutils source code, I couldn't spot any potential harm in always setting LDSHARED. If anything, it leads to less guesswork for distutils.
Here is a link to the relevant code;
https://github.com/python/cpython/blob/e42b705188271da108de42b55d9344642170aa2b/Lib/distutils/sysconfig.py#L172

I don't mind adding a conditional check though, but I don't do much hacking on blocks, so I would appreciate a hint on what's the cleanest way to check this would be.

boegel: I sort of share @damianam's concern on always setting `$LDSHARED` here.
On the other hand, if both compilers are the same, setting `$LDSHARED` like this shouldn't have any impact. In theory...

I checedk what this spits out for a couple of different `Python` installations:

```
python -c "from distutils.sysconfig import get_config_vars; print(get_config_vars('LDSHARED'))[0]"
```

Results:

* system Python 2 (`/usr/bin/python`) in CentOS 7:
  ```
  gcc -pthread -shared -Wl,-z,relro
  ```
  This already disagrees with the hardcoded `$CC -shared` that is being added here.

* `Python/2.7.14-foss-2018a`
  ```
  gcc -shared -L/prefix/software/GCCcore/6.4.0/lib64 -L/prefix/software/GCCcore/6.4.0/lib -L/prefix/software/OpenBLAS/0.2.20-GCC-6.4.0-2.28/lib -L/prefix/software/ScaLAPACK/2.0.2-gompi-2018a-OpenBLAS-0.2.20/lib -L/prefix/software/FFTW/3.3.7-gompi-2018a/lib -L/prefix/software/bzip2/1.0.6-GCCcore-6.4.0/lib -L/prefix/software/zlib/1.2.11-GCCcore-6.4.0/lib -L/prefix/software/libreadline/7.0-GCCcore-6.4.0/lib -L/prefix/software/ncurses/6.0-GCCcore-6.4.0/lib -L/prefix/software/SQLite/3.21.0-GCCcore-6.4.0/lib -L/prefix/software/GMP/6.1.2-GCCcore-6.4.0/lib -L/prefix/software/libffi/3.2.1-GCCcore-6.4.0/lib64 -L/prefix/software/libffi/3.2.1-GCCcore-6.4.0/lib
  ```
  This is more than just `$CC -shared`, it's also a bunch of `-L` specs, seems like one for each Python dependency (probably using `$LDFLAGS` that was set when `Python` was built).

* `Python/2.7.14-intel-2018a`
  ```
  icc -shared -L/prefix/software/icc/2018.1.163-GCC-6.4.0-2.28/lib/intel64 -L/prefix/software/imkl/2018.1.163-iimpi-2018a/lib -L/prefix/software/imkl/2018.1.163-iimpi-2018a/mkl/lib/intel64 -L/prefix/software/imkl/2018.1.163-iimpi-2018a/lib -L/prefix/software/bzip2/1.0.6-GCCcore-6.4.0/lib -L/prefix/software/zlib/1.2.11-GCCcore-6.4.0/lib -L/prefix/software/libreadline/7.0-GCCcore-6.4.0/lib -L/prefix/software/ncurses/6.0-GCCcore-6.4.0/lib -L/prefix/software/SQLite/3.21.0-GCCcore-6.4.0/lib -L/prefix/software/GMP/6.1.2-GCCcore-6.4.0/lib -L/prefix/software/libffi/3.2.1-GCCcore-6.4.0/lib64 -L/prefix/software/libffi/3.2.1-GCCcore-6.4.0/lib
  ```
  Again `$CC -shared` + `-L` specs determined by `$LDFLAGS` that was set when building this `Python`.

* `Python/3.6.4-foss-2018a`
  ```
  gcc -shared -L/prefix/software/GCCcore/6.4.0/lib64 -L/prefix/software/GCCcore/6.4.0/lib -L/prefix/software/OpenBLAS/0.2.20-GCC-6.4.0-2.28/lib -L/prefix/software/ScaLAPACK/2.0.2-gompi-2018a-OpenBLAS-0.2.20/lib -L/prefix/software/FFTW/3.3.7-gompi-2018a/lib -L/prefix/software/bzip2/1.0.6-GCCcore-6.4.0/lib -L/prefix/software/zlib/1.2.11-GCCcore-6.4.0/lib -L/prefix/software/libreadline/7.0-GCCcore-6.4.0/lib -L/prefix/software/ncurses/6.0-GCCcore-6.4.0/lib -L/prefix/software/SQLite/3.21.0-GCCcore-6.4.0/lib -L/prefix/software/GMP/6.1.2-GCCcore-6.4.0/lib -L/prefix/software/libffi/3.2.1-GCCcore-6.4.0/lib64 -L/prefix/software/libffi/3.2.1-GCCcore-6.4.0/lib -L/prefix/software/GCCcore/6.4.0/lib64 -L/prefix/software/GCCcore/6.4.0/lib -L/prefix/software/OpenBLAS/0.2.20-GCC-6.4.0-2.28/lib -L/prefix/software/ScaLAPACK/2.0.2-gompi-2018a-OpenBLAS-0.2.20/lib -L/prefix/software/FFTW/3.3.7-gompi-2018a/lib -L/prefix/software/bzip2/1.0.6-GCCcore-6.4.0/lib -L/prefix/software/zlib/1.2.11-GCCcore-6.4.0/lib -L/prefix/software/libreadline/7.0-GCCcore-6.4.0/lib -L/prefix/software/ncurses/6.0-GCCcore-6.4.0/lib -L/prefix/software/SQLite/3.21.0-GCCcore-6.4.0/lib -L/prefix/software/GMP/6.1.2-GCCcore-6.4.0/lib -L/prefix/software/libffi/3.2.1-GCCcore-6.4.0/lib64 -L/prefix/software/libffi/3.2.1-GCCcore-6.4.0/lib -lssp
  ```
  Same as for `Python/2.7.14-foss-2018a`, except for the additional `-lssp` at the end (which is self-inflicted, we add `-lssp` to `$LDFLAGS` in easyconfigs for recent Python 3.x verisons).

* `Python/3.6.4-foss-2018a`
  ```
  icc -shared -L/prefix/software/icc/2018.1.163-GCC-6.4.0-2.28/lib/intel64 -L/prefix/software/imkl/2018.1.163-iimpi-2018a/lib -L/prefix/software/imkl/2018.1.163-iimpi-2018a/mkl/lib/intel64 -L/prefix/software/imkl/2018.1.163-iimpi-2018a/lib -L/prefix/software/bzip2/1.0.6-GCCcore-6.4.0/lib -L/prefix/software/zlib/1.2.11-GCCcore-6.4.0/lib -L/prefix/software/libreadline/7.0-GCCcore-6.4.0/lib -L/prefix/software/ncurses/6.0-GCCcore-6.4.0/lib -L/prefix/software/SQLite/3.21.0-GCCcore-6.4.0/lib -L/prefix/software/GMP/6.1.2-GCCcore-6.4.0/lib -L/prefix/software/libffi/3.2.1-GCCcore-6.4.0/lib64 -L/prefix/software/libffi/3.2.1-GCCcore-6.4.0/lib -L/prefix/software/icc/2018.1.163-GCC-6.4.0-2.28/lib/intel64 -L/prefix/software/imkl/2018.1.163-iimpi-2018a/lib -L/prefix/software/imkl/2018.1.163-iimpi-2018a/mkl/lib/intel64 -L/prefix/software/imkl/2018.1.163-iimpi-2018a/lib -L/prefix/software/bzip2/1.0.6-GCCcore-6.4.0/lib -L/prefix/software/zlib/1.2.11-GCCcore-6.4.0/lib -L/prefix/software/libreadline/7.0-GCCcore-6.4.0/lib -L/prefix/software/ncurses/6.0-GCCcore-6.4.0/lib -L/prefix/software/SQLite/3.21.0-GCCcore-6.4.0/lib -L/prefix/software/GMP/6.1.2-GCCcore-6.4.0/lib -L/prefix/software/libffi/3.2.1-GCCcore-6.4.0/lib64 -L/prefix/software/libffi/3.2.1-GCCcore-6.4.0/lib -lssp
  ```
  Basically the same as for `Python/3.6.4-foss-2018a`, but with `gcc` replaced by `icc` (also extra `-lssp` compared to `Python/2.7.14-intel-2018a`).

Long story short: always setting `$LDSHARED` to `$CC -shared` is probably not a good idea, since then we'd effectively be ignoring the `$LDFLAGS` part of it. But we can ask `Python` what it thinks `$LDSHARED` should be, and then replace the compiler command? (see below)

To make the setting of `$LDSHARED` conditional, we can check the value that `get_config_vars('LDSHARED')` spits out, and see whether it starts with the current value for `$CC`. Something like this (untested, but should work I think):

```python
# take into account that $LDSHARED may already be set...
curr_ldshared = os.getenv('LDSHARED')
python_ldshared = get_config_vars('LDSHARED')[0].split(' ')
curr_cc = os.getenv('CC')
if python_ldshared[0] != curr_cc:
    self.log.info("Python's value for $LDSHARED ('%s') doesn't use current $CC value ('%s'), fixing that...", python_ldshared, curr_cc)
    if curr_ldshared:
        # not sure what to do here... print/log a warning that $LDSHARED will be redefined? raise an error?
    env.setvar('LDSHARED', ' '.join([curr_cc] + python_ldshared[1:])
```

Not sure if the search-and-replace is wise. Maybe just setting it to `$CC -shared` conditionally is better/good enough?

Micket: If so, my vote is to conditionally set it to "CC - shared".
Keeping the flags from another linker seems fragile, as there wouldn't really be any guarantee that the other compiler uses the same names (not the case here, but it was actually link flags compatibility differences which caused me to find this problem to start with when I started compiling for Python-core. (in that case, was just some `icc` compile flag that leaked into the link stage if i recall correctly, but it was enough to break the link step)

If there are no objections, I will update with
```python
python_ldshared = get_config_vars('LDSHARED')[0].split(' ')
curr_cc = os.getenv('CC')
if python_ldshared[0] != curr_cc:
    env.setvar("LDSHARED", curr_cc + " -shared")
```


Should i prefer using `os.getenv` or `toolchain.get_variable` here?

boegel: @Micket I don't think it matters much which you use, `LDSHARED` will be set by the toolchain mechanism, so there's no point to try and grab it via `self.toolchain` imho...

easybuild/easyblocks/generic/pythonpackage.py
```diff
@@ -449,6 +449,13 @@ def build_step(self):
                 env.setvar("CMAKE_INCLUDE_PATH", include_paths)
                 env.setvar("CMAKE_LIBRARY_PATH", library_paths)
 
+            curr_cc = os.getenv('CC')
+            python_ldshared = get_config_vars('LDSHARED')[0].split(' ')
+            if python_ldshared[0] != curr_cc:
```
boegel: @Micket Missing import statement for `get_config_vars`?

Also, I would like to be a bit more careful here. We're assuming that there's at least one entry in the return value of `get_config_vars`. Will that always be the case, on all platforms?

To avoid introducing problems when installing any Python package, I would prefer a bit more robust code, like:

```python
curr_cc = os.getenv('CC')
python_ldshared = get_config_vars('LDSHARED')
if python_ldshared:
    if python_ldshared[0].split(' ')[0] != curr_cc:
        ...
else:
    self.log.warning("Failed to determine value for Python's $LDSHARED, so can't check it against $CC ('%s')", curr_cc)
```

Micket: Or set it to `$CC -shared` in the else condition?
I have to note that distutils will append both LDFLAGS and CFLAGS to this command.

I.M.H.O. distutils is a bit of a hodgepodge of conditional code so it's hard to say what's correct. 
As far as I can see, it always expects get_config_vars to return something, else it would append LDFLAGS to an empty string?

If there truly is a case where it's not set at all, what will it use to link? Surely `$CC -shared` would be as good as any choice then.

boegel: @Micket That sounds reasonable, let's go with that.

easybuild/easyblocks/generic/pythonpackage.py
```diff
@@ -449,6 +450,17 @@ def build_step(self):
                 env.setvar("CMAKE_INCLUDE_PATH", include_paths)
                 env.setvar("CMAKE_LIBRARY_PATH", library_paths)
 
+            curr_cc = os.getenv('CC')
+            python_ldshared = get_config_vars('LDSHARED')[0]
+            if python_ldshared:
+                if python_ldshared.split(' ')[0] != curr_cc:
+                    self.log.info("Python's value for $LDSHARED ('%s') doesn't use current $CC value ('%s'), fixing",
+                                  python_ldshared, curr_cc)
+                    env.setvar("LDSHARED", curr_cc + " -shared")
```
boegel: @Micket Let's squeeze in another log message to get "full coverage" in the log w.r.t. `$LDSHARED` (and let's switch to positive logic while we're at it, easier to parse for humans):

```python
if python_ldshared:
    if python_ldshared.split(' ')[0] == curr_cc:
        self.log.info("Python's value of $LDSHARED ('%s') uses current $CC value ('%s'), so not touching it",
                      python_ldshared, curr_cc)
    else:
        ...
else:
    ...
```

boegel:
While testing this a bit, I ran into a problem because an assumption is being made that `$CC` is always set:

```
  File "/tmp/eb-RG7KYO/included-easyblocks/easybuild/easyblocks/generic/pythonpackage.py", line 462, in build_step
    env.setvar("LDSHARED", curr_cc + " -shared")
TypeError: unsupported operand type(s) for +: 'NoneType' and 'str'
```

The problem is that `$CC` is not defined (by EasyBuild) for Python packages that get installed on top of the system Python, with a `dummy` toolchain.
That includes EasyBuild itself for example, but we also have easyconfigs for `PyYAML`, `vsc-*` tools, etc. like that.

Proposed fix in https://github.com/Micket/easybuild-easyblocks/pull/1 .

Another thing I mentioned is that we're only checking/setting `$LDSHARED` here when `setup.py` is being used. @Micket Is that intentional?
If not, we should move this block of code to the `configure_step` (outside of an `if` scope)?

boegel:
@Micket What about if `pip` is used? I'm not sure that rules out that `distutils` is involved...

I would always set `$LDSHARED` when needed, regardless of whether `setup.py` is used or not. If that causes problems we can revisit this when we have examples where it's problematic?

Micket:
@boegel Sorry, apparently i missed pushing the last update.

Micket:
Just to keep a public record from our discussion: I vote for opt-in, at least for 3.8.1, as it at least allows conveniently testing out alternatives.

boegel:
Spent a bit more time testing this, both the (unchanged) default behavior and the use of `check_ldshared` via:

```python
exts_default_options = {'check_ldshared': True}
```

in a `Python` easyconfig, works as expected, so good to go.

Thanks a lot for your work on this @Micket!

PR #1460, opened 2018-07-16
update wien2k easyblock to 18.1

migueldiascosta:


ocaisa:
Ah, you already have...merging!

PR #1468, opened 2018-07-27
Amber: Handle X11 better and make static a build flag.

akesandgren:
Also fix some lingering tabs.

Static handling taken from PR#1343

The zlib/bzip2 part is not needed from that PR.

### DIFF
--- a/easybuild/easyblocks/a/amber.py
+++ b/easybuild/easyblocks/a/amber.py
@@ -128,7 +128,7 @@ class EB_Amber(ConfigureMake):
         elif openblasroot:
             lapack = os.getenv('LIBLAPACK')
             if lapack is None:
-                raise EasyBuildError("LIBLAPACK (from OpenBLAS) not found in environement.")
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/a/amber.py
```diff
@@ -125,11 +126,11 @@ def install_step(self):
...
             env.setvar('MKL_HOME', mklroot)
         elif openblasroot:
-	    lapack = os.getenv('LIBLAPACK')
-	    if lapack is None:
-		raise EasyBuildError("LIBLAPACK (from OpenBLAS) not found in environement.")
-	    else:
-		env.setvar('GOTO', lapack)
+            lapack = os.getenv('LIBLAPACK')
+            if lapack is None:
+                raise EasyBuildError("LIBLAPACK (from OpenBLAS) not found in environement.")
```
damianam: Since you are at it, while not fix typos like `environement`?

akesandgren: as you wish :-)

PR #1473, opened 2018-08-16
Used os.path.join in cuda.py, and added ldconfig to post_install_step

damianam:
This PR makes the easyblock use `os.path.join` instead of `/`, but more importantly, runs `ldconfig -N` in the stubs directory, to create the proper symlinks. This is necessary to deal with binaries created outside of EB, which might be linked against `libcuda.so.1` instead of `libcuda.so`, like, for instance, `MVAPICH2-GDR`

### DIFF
--- a/easybuild/easyblocks/c/cuda.py
+++ b/easybuild/easyblocks/c/cuda.py
@@ -132,11 +132,10 @@ class EB_CUDA(Binary):
         if LooseVersion(self.version) < LooseVersion("6"):
             chk_libdir += ["lib"]
 
+        culibs = ["cublas", "cudart", "cufft", "curand", "cusparse"]
         custom_paths = {
-            'files': [os.path.join("bin", "%s") % x for x in ["fatbinary", "nvcc", "nvlink", "ptxas"]] +
-                     [os.path.join("%s", "lib%s.%s") % (x, y, shlib_ext) for x in chk_libdir for y in ["cublas",
-                                                                                      "cudart", "cufft", "curand",
-                                                                                      "cusparse"]],
+            'files': [os.path.join("bin", x) for x in ["fatbinary", "nvcc", "nvlink", "ptxas"]] +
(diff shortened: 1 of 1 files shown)


damianam:
@akesandgren could you review this PR?

easybuild/easyblocks/c/cuda.py
```diff
@@ -129,17 +133,18 @@ def sanity_check_step(self):
             chk_libdir += ["lib"]
 
         custom_paths = {
-            'files': ["bin/%s" % x for x in ["fatbinary", "nvcc", "nvlink", "ptxas"]] +
-                     ["%s/lib%s.%s" % (x, y, shlib_ext) for x in chk_libdir for y in ["cublas", "cudart", "cufft",
-                                                                                      "curand", "cusparse"]],
+            'files': [os.path.join("bin", "%s") % x for x in ["fatbinary", "nvcc", "nvlink", "ptxas"]] +
+                     [os.path.join("%s", "lib%s.%s") % (x, y, shlib_ext) for x in chk_libdir for y in ["cublas",
```
akesandgren: A variant of writing this is:
[os.path.join("bin", x) for x in [...]]
[os.path.join(x, "lib%s.%s") % (y, shlib_ext) for x in ...

At least the first one is slightly easier to read

boegel: probably easier to lift the `["cublas", ...]` list into a variable:

```python
culibs = ["cublas", "cudart", "cufft", "curand", "cusparse"]
...
   [os.path.join("%s", "lib%s.%s") % (d, l, shlib_ext) for d in chk_libdir for l in culibs]

boegel:
other than the minor style issues that @houndci-bot is tripping over: lgtm

damianam:
@boegel @akesandgren any other comments?

boegel:
lgtm, tested with a bunch of existing `CUDA` easyconfigs, (still) works like charm...

Going in, thanks @damianam!

Issue #1475, opened 2018-08-17
ANSYS >= 19.0 needs CEI/bin added to the list of paths to support Ensight (new)


JackPerdue:
I think the topic covers it... here's a diff for someone with more git wizardry than I.
[ansysCEI.patch.txt](https://github.com/easybuilders/easybuild-easyblocks/files/2295620/ansysCEI.patch.txt)

boegel:
guts of the patch:

```diff
--- /sw/eb/sw/EasyBuild/3.6.2/lib/python2.7/site-packages/easybuild_easyblocks-3.6.2-py2.7.egg/easybuild/easyblocks/a/ansys.py	2018-07-12 12:38:41.743295000 -0500
+++ easyblocks/ansys.py	2018-08-16 15:48:12.000000000 -0500
@@ -34,6 +34,7 @@
 from easybuild.easyblocks.generic.packedbinary import PackedBinary
 from easybuild.tools.run import run_cmd
 from easybuild.tools.filetools import adjust_permissions
+from distutils.version import LooseVersion
 
 
 class EB_ANSYS(PackedBinary):
@@ -76,6 +77,8 @@
             "Icepak/bin",
             "icemcfd/linux64_amd/bin"
         ]
+        if LooseVersion(self.version) >= LooseVersion('19.0'):
+            dirs.append("CEI/bin")
         guesses.update({"PATH": [os.path.join(self.ansysver, dir) for dir in dirs]})
         return guesses

```

boegel:
fixed in #1476

PR #1476, opened 2018-08-17
add CEI/bin to $PATH for ANSYS 19 & newer

boegel:
fix for #1475


JackPerdue:
Note, I'm not sure if LooseVersion() is required or not.  I basically copied code from other easyblocks.  I guess if ANSYS releases 19.1.1 then LooseVersion might be a good idea, but I don't know.

boegel:
@damianam merge away, that was just because I requested @verdurin's review last week (since he was MotW then)

PR #1477, opened 2018-08-19
Make scons accept parallel setting.

gatagat:
Make scons accept parallel setting as is usual with all other blocks.


boegel:
Tested with `Serf-1.3.8-foss-2015a.eb` and `Cantera-2.3.0-intel-2018a-Python-2.7.14.eb`, which both use the `SCons` generic easyblock, works like a charm, so good to go into `develop` (to be included with the next EasyBuild release).

Thanks a lot for your contribution @gatagat!

PR #1479, opened 2018-08-22
SuperLU handle both lib and lib64 subdirectories

hajgato:


hajgato:
@damianam Done plus let the automatic fallback to handle lib/lib64 in the sanity check.
@boegel fixed the commit-hell.

damianam:
@hajgato I wasn't aware of any automatic fallback to check that. @boegel since when do we have that? :-?

boegel:
@damianam Since EasyBuild v3.6.1, see https://github.com/easybuilders/easybuild-framework/pull/2477

Not sure if it's a good idea to rely on that here though, since we *know* where stuff is installed...

The fallback can be disabled by configuring EasyBuild with `--disable-lib64-fallback-sanity-check`, in which case the `SuperLU` installation would fail during the sanity check if stuff is installed in `lib64`, while we actually know it's being installed there.

The fallback is sort of intended as a stop gap to avoid having to go hunt for `lib` hardcoding everywhere and replace it with `lib or lib64`...

easybuild/easyblocks/s/superlu.py
```diff
@@ -123,16 +123,19 @@ def install_step(self):
         """
         super(EB_SuperLU, self).install_step()
 
-        expected_libpath = os.path.join(self.installdir, "lib", "libsuperlu.%s" % self.lib_ext)
-        actual_libpath   = os.path.join(self.installdir, "lib", "libsuperlu_%s.%s" % (self.cfg['version'], self.lib_ext))
+        libbit = "lib"
+        if not os.path.exists(os.path.join(self.installdir, libbit)):
+                libbit = "lib64"
```
boegel: @hajgato please fix indent here (4 spaces)

Also, what if the `lib64` dir doesn't exist either? Don't we want a clean error message then?

easybuild/easyblocks/s/superlu.py
```diff
@@ -123,22 +123,30 @@ def install_step(self):
...
 
-        expected_libpath = os.path.join(self.installdir, "lib", "libsuperlu.%s" % self.lib_ext)
-        actual_libpath   = os.path.join(self.installdir, "lib", "libsuperlu_%s.%s" % (self.cfg['version'], self.lib_ext))
+        self.libbits = ""
+        for libnames in ["lib", "lib64"]:
+            if os.path.exists(os.path.join(self.installdir, libnames)):
+                self.libbits = libnames
+
+        if not self.libbits:
+            raise EasyBuildError("No lib or lib64 directory exist in installdir '%s'" % self.installdir)
```
boegel: @hajgato Using a `for` loop is a bit weird here imho, since there are only two options:

```python
self.libbits = 'lib'
if not os.path.exists(os.path.join(self.installdir, self.libbits)):
    self.libbits = 'lib64'

if not os.path.exists(os.path.join(self.installdir, self.libbits)):
    raise EasyBuildError("No lib or lib64 subdirectory exist in %s", self.installdir)

File: easybuild/easyblocks/s/superlu.py [hunk 3]
```diff
@@ -123,22 +123,30 @@ def install_step(self):
...
                 raise EasyBuildError("Failed to create symlink '%s' -> '%s: %s", expected_libpath, actual_libpath, err)
 
-
     def sanity_check_step(self):
         """
         Check for main library files for SuperLU
         """
         custom_paths = {
-            'files': ["include/supermatrix.h", "lib/libsuperlu.%s" % self.lib_ext],
+            'files': ["include/supermatrix.h", "%s/libsuperlu.%s" % (self.libbits, self.lib_ext)],
```
boegel: @hajgato Please use `os.path.join(self.libbits, 'libsuperlu.%s' % self.lib_ext)`

easybuild/easyblocks/s/superlu.py
hajgato: @boegel Done.

boegel:
Good to go for me, still works like a charm with all existing `SuperLU` easyconfigs.

@damianam Yours to merge?

PR #1480, opened 2018-08-22
Added CCOLAMD and CAMD to Trilinos

vanzod:
For `Trilinos-12.12.1` the paths for `SuiteSparse` `CCOLAMD` and `CAMD` libraries must be added to the linker paths.

### DIFF
--- a/easybuild/easyblocks/t/trilinos.py
+++ b/easybuild/easyblocks/t/trilinos.py
@@ -125,7 +125,7 @@ class EB_Trilinos(CMakeMake):
         if suitesparse:
             self.cfg.update('configopts', "-DTPL_ENABLE_UMFPACK:BOOL=ON")
             incdirs, libdirs, libnames = [], [], []
-            for lib in ["UMFPACK", "CHOLMOD", "COLAMD", "AMD"]:
+            for lib in ["UMFPACK", "CHOLMOD", "COLAMD", "AMD", "CCOLAMD", "CAMD"]:
                 incdirs.append(os.path.join(suitesparse, lib, "Include"))
                 libdirs.append(os.path.join(suitesparse, lib, "Lib"))
                 libnames.append(lib.lower())


damianam:
@vanzod are there side effects when this is applied to older versions of trilinos? Potentially yes, but in reality I don't know. The answer would decide if an extra version check is needed or not.

boegel:
@damianam I was wondering the same thing. I agree though, it would be quite surprising if it does break previous Trilinos versions.

I'll try and verify, I should be able to get the old Trilinos easyconfigs built (but it takes *forever* to build Trilinos, so it'll take a while...).

damianam:
Thanks @boegel! I'll wait for your feedback before merging.

boegel:
@vanzod Do you happen to still have the error messages that are fixed by this change?

Always good to mention those in the PR description...

boegel:
Here's what I'm seeing for `Trilinos` 12.12.4 with `intel/2018a`, when using this updated easyblock:

```
/prefix/software/SuiteSparse/5.1.2-intel-2018a-METIS-5.1.0/CHOLMOD/Lib/libcholmod.a(cholmod_metis.o): In function `cholmod_metis_bisector':
cholmod_metis.c:(.text+0x5c8): undefined reference to `METIS_ComputeVertexSeparator'
/prefix/software/SuiteSparse/5.1.2-intel-2018a-METIS-5.1.0/CHOLMOD/Lib/libcholmod.a(cholmod_metis.o): In function `cholmod_metis':
cholmod_metis.c:(.text+0x14b5): undefined reference to `METIS_NodeND'
make[2]: *** [packages/ifpack/test/CompareWithAztecOO/Ifpack_CompareWithAztecOO.exe] Error 1
```

Is this change supposed to fix that issue @vanzod? Or did you fix another issue with this change?

boegel:
https://answers.launchpad.net/dorsal/+question/223167 suggests that my problem is caused by not linking `libmetis.a` as well.

Meanwhile, I've confirmed that the changes in 86bbe63 do not break older Trilinos versions; `Trilinos-12.4.2-intel-2015b-Python-2.7.11.eb` still builds fine.

Not sure if @vanzod is anticipating further changes to resolve the issue he's hitting in https://github.com/easybuilders/easybuild-easyconfigs/pull/6752 (but it seems like a patch will be required to fix that, rather than more changes to the easyblock).

boegel:
I'll go ahead and merge this, since I've confirmed that this change fixes the problem that Davide mentioned for `Trilinos` 12.12.1, and it doesn't break older `Trilinos` easyconfigs.

It's likely more changes will be needed for `Trilinos` 12.12.1 however...

PR #1481, opened 2018-08-23
Add support to apply (binary) patches after main CUDA install

wpoely86:
The make binary patches instead of updating the version: https://developer.nvidia.com/cuda-downloads?target_os=Linux&target_arch=x86_64&target_distro=CentOS&target_version=7&target_type=runfilelocal

### DIFF
--- a/easybuild/easyblocks/c/cuda.py
+++ b/easybuild/easyblocks/c/cuda.py
@@ -18,6 +18,7 @@ Ref: https://speakerdeck.com/ajdecon/introduction-to-the-cuda-toolkit-for-buildi
 @author: Fotis Georgatos (Uni.lu)
 @author: Kenneth Hoste (Ghent University)
 @author: Damian Alvarez (Forschungszentrum Juelich)
+@author: Ward Poelmans (Free University of Brussels)
 """
 import os
 import stat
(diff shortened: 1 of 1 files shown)


boegel:
@wpoely86 Do you have an example easyconfig for this?

boegel:
Ah, nevermind, https://github.com/easybuilders/easybuild-easyconfigs/pull/6754 needs it.

damianam:
lgtm!

easybuild/easyblocks/c/cuda.py
```diff
@@ -124,6 +133,11 @@ def create_wrapper(wrapper_name, wrapper_comp):
 
     def sanity_check_step(self):
         """Custom sanity check for CUDA."""
+
+        versionfile = read_file(os.path.join(self.installdir, "version.txt"))
```
boegel: @wpoely86 At the very least, this should be conditional based on the CUDA version. This `version.txt` file is not there for CUDA 7.0.28 for example, so this change is not backward-compatible...

Also, it really only makes sense to check when patches are involved, i.e. when `len(self.src) > 1`?

boegel: fixed in #1483

PR #1482, opened 2018-08-23
switch to using CMake install procedure for Eigen 3.3.4 & newer

boegel:
If the CMake-based installation procedure is not using, the `eigen3.pc` file for `pkg-config` does not get generated & installed...

### DIFF
--- a/easybuild/easyblocks/e/eigen.py
+++ b/easybuild/easyblocks/e/eigen.py
@@ -33,6 +33,9 @@ class EB_Eigen(CMakeMake):
 
     def configure_step(self):
         """Custom configuration procedure for Eigen."""
+        # start using CMake for Eigen 3.3.4 and newer versions
+        # not done for older versions, since this implies using (a dummy-built) CMake as a build dependency,
+        # which is a bit strange for a header-only library like Eigen...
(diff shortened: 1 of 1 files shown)


damianam:
@boegel it lgtm, but I wonder where the version check comes from. I haven't seen anything in Eigen's changelog pointing to any change in cmake files or pkg-config in `3.3.4`.

boegel:
@damianam I don't want to make this change for old `Eigen` versions/easyconfigs because i) the problem hasn't popped up for older versions (I suspect because not many software packages require `eigen3.pc`, and ii) mainly because the change requires using `CMake` as a build dependency for something that is installed with `dummy` toolchain (since nothing is actually being compiled).
That implies that `CMake` itself should also be built with `dummy`, which leaves us at the mercy of the OS compiler for building `CMake` 😱

damianam:
I understand........ But I would add a comment making clear where the version restriction comes from. This is an EB issue, not an eigen requirement. Normally version checks are due to changes from one version onwards, which is not the case here. Looks to me like sort of "abusing" version checks for something they are not intended for (not exactly that, but you know what I mean)

boegel:
@damianam I understand the potential confusion, clarifying comment added in 175326cc9

PR #1483, opened 2018-08-24
make CUDA version check version dependent (only for v9.x & newer)

wpoely86:
Follow up fix from #1481

### DIFF
--- a/easybuild/easyblocks/c/cuda.py
+++ b/easybuild/easyblocks/c/cuda.py
@@ -134,9 +134,10 @@ class EB_CUDA(Binary):
     def sanity_check_step(self):
         """Custom sanity check for CUDA."""
 
-        versionfile = read_file(os.path.join(self.installdir, "version.txt"))
-        if not re.search("Version %s$" % self.version, versionfile):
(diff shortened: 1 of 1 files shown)


boegel:
@wpoely86 Turns out only checking for the CUDA version to enable the check is a bit optimistic...

Installation of `CUDA-8.0.61_375.26-GCC-5.4.0-2.26.eb` now fails with:

```
== FAILED: Installation ended unsuccessfully (build directory: /tmp/easybuild_build/CUDA/8.0.61_375.26/GCC-5.4.0-2.26): build failed (first 300 chars): Unable to find the correct version (8.0.61_375.26) in the version.txt file
```

because of

```
$ cat $EASYBUILD_PREFIX/software/CUDA/8.0.61_375.26-GCC-5.4.0-2.26/version.txt
CUDA Version 8.0.61
```

while the version is `8.0.61_375.26`

So better only enable the check when more than one source file is provided?

wpoely86:
That still will not help. If some easyconfigs use a different version scheme, there isn't much I can do? I can limit it to version 9 and newer which all use the same version convention?

boegel:
@wpoely86 Fine by me, I don't expect we'll be installing CUDA 8.x with patches anyway...

PR #1484, opened 2018-08-24
add support for only building Python bindings (+ code cleanup) in Boost easyblock

boegel:


boegel:
@damianam I've verified that the changes don't break existing `Boost` easyconfigs.

PR #1485, opened 2018-08-24
provide control over subdirectory in which R packages are installed

boegel:
If `exts_subdir` is left undefined, this doesn't change anything, so backward compatible.

Use case for this is https://github.com/egaffo/CirComPara, which expects R extensions to be installed in `tools/R_libs`, see https://github.com/egaffo/CirComPara/blob/circompara/install_circompara#L24


damianam:
lgtm. Do you have accompanying easyconfigs you can reference here?

boegel:
@damianam It's required for https://github.com/easybuilders/easybuild-easyconfigs/pull/6764, but that's still very much a WIP.

Doesn't mean this PR should be blocked though, it just adds an exrtra knob to turn but doesn't make any functional changes.

damianam:
Going in!

PR #1488, opened 2018-08-27
FreeFem++ EasyBlock

hajgato:


easybuild/easyblocks/f/freefem.py
```diff
@@ -0,0 +1,60 @@
...
+
+class EB_FreeFem(ConfigureMake):
+    """Support for building and installing FreeFem++."""
+
+    def configure_step(self):
+        """FreeFem++ configure should run twice. First PETSc configured, then PETSc have to be build,
+        then configure FreeFem++ with the builded PETSc."""
+
+        # first Autoreconf has to be run
+        if not get_software_root('Autotools'):
```
boegel: @hajgato Check for `Autoconf` rather than `Autotools`?

easybuild/easyblocks/f/freefem.py
```diff
@@ -0,0 +1,60 @@
...
+            raise EasyBuildError("Autoconfig is required to build FreeFem++. Please add it as build dependency")
```
boegel: Should be `Autoconf`, not `Autoconfig`?

easybuild/easyblocks/f/freefem.py
```diff
@@ -0,0 +1,60 @@
...
+        run_cmd("autoreconf -i", log_all=True, simple=False)
+
+        # delete old installation, then set keeppreviousinstall to True (do not delete PETsc install)
+        self.make_installdir()
+        self.cfg['keeppreviousinstall'] = True
+
+        # configure and make petsc-slepc
+        cmd = "./configure --prefix=%s &&" % self.installdir
+        cmd += "cd download/ff-petsc &&"
+        cmd += "make petsc-slepc &&"
```
boegel: @hajgato Can you clarify why you prefer using the bundled/downloaded `PETSc` & `SLEPc` rather than including them as dependencies?

easybuild/easyblocks/f/freefem.py
```diff
@@ -0,0 +1,60 @@
...
+
+@author: Balazs Hajgato (Free University Brussels - VUB)
+"""
+from easybuild.easyblocks.generic.configuremake import ConfigureMake
+from easybuild.tools.run import run_cmd
+from easybuild.tools.build_log import EasyBuildError
+from easybuild.tools.modules import get_software_root
+
+
+class EB_FreeFem(ConfigureMake):
```
boegel: @hajgato Please use the correct auto-derived class name:

```
$ python -c "from easybuild.tools.filetools import encode_class_name; print encode_class_name('FreeFem++')"
EB_FreeFem_plus__plus_
```

easybuild/easyblocks/f/freefem.py
hajgato: Done

easybuild/easyblocks/f/freefem.py
hajgato: Done

easybuild/easyblocks/f/freefem.py
hajgato: The whole `FreeFem++` install system is an annoying one-command-installs-all-but-give-the-root-password style, so it would take too much time to look deep into the install scripts. (basically for the same reason why we use downloaded/bundled deps for `PETSc`. Additionally `FreeFrem++` needs bot real and complex `PETSc` (and `SLEPc` built on the top of both versions).

akesandgren:
@hajgato ping? Anything left to do here?

hajgato:
@akesandgren I think I fixed all the requested changes.

easybuild/easyblocks/f/freefem.py
```diff
@@ -0,0 +1,109 @@
...
+from easybuild.tools.build_log import EasyBuildError
+from easybuild.tools.modules import get_software_root
+
+
+class EB_FreeFem_plus__plus_(ConfigureMake):
+    """Support for building and installing FreeFem++."""
+
+    def configure_step(self):
+        """FreeFem++ configure should run twice. First PETSc configured, then PETSc have to be build,
+        then configure FreeFem++ with the builded PETSc."""
```
akesandgren: s/builded/built/

easybuild/easyblocks/f/freefem.py
```diff
@@ -0,0 +1,109 @@
...
+        run_cmd("autoreconf -i", log_all=True, simple=False)
+
+        # delete old installation, then set keeppreviousinstall to True (do not delete PETsc install)
+        self.make_installdir()
+        self.cfg['keeppreviousinstall'] = True
+
+        # configure and make petsc-slepc
+        cmd = "./configure --prefix=%s &&" % self.installdir
+        cmd += "cd download/ff-petsc &&"
+        cmd += "make petsc-slepc &&"
```
akesandgren: So the configure shouldn't be done in ff-petsc directory?

easybuild/easyblocks/f/freefem.py
```diff
@@ -0,0 +1,109 @@
...
+        # PaStiX does not work parallel, so disable it.
+        self.cfg.update('configopts', '--disable-pastix ')
+
+        super(EB_FreeFem_plus__plus_, self).configure_step()
+
+    def build_step(self):
+        # dependencies are heavily patched by FreeFem++, we therefore let it download and install them itself
+        self.cfg.update('prebuildopts', 'download/getall -a -o ScaLAPACK,ARPACK,freeYams,Gmm++,Hips,Ipopt,METIS,'
+                        'ParMETIS,MMG3D,mshmet,MUMPS,NLopt,pARMS,PaStiX,Scotch,SuiteSparse,SuperLU_DIST,SuperLU,'
+                        'TetGen,PETSc,SLEPc,hpddm &&')
```
akesandgren: Hmmm, the first thing you do in configure_step is build ff_petsc, but here you ask it to download petsc and slepc again.

easybuild/easyblocks/f/freefem.py
```diff
@@ -0,0 +1,109 @@
...
+        # define FreeFem++ sanity_check_paths here
+
+        custom_paths = {
+            'files': ['bin/%s' % x for x in ['bamg', 'cvmsh2', 'ffglut', 'ffmedit']] +
+            ['bin/ff-%s' % x for x in ['c++', 'get-dep', 'mpirun', 'pkg-download']] +
+            ['bin/FreeFem++%s' % x for x in ['', '-mpi', '-nw']],
+            'dirs': ['share/freefem++/%s' % self.version] +
+            ['lib/ff++/%s/%s' % (self.version, x) for x in ['bin', 'etc', 'idp', 'include', 'lib']]
+        }
+        super(EB_FreeFem_plus__plus_, self).sanity_check_step(custom_paths=custom_paths)
```
akesandgren: I think sanity check step should return the result. @boegel I that true in all cases or only when dealing with extensions?

easybuild/easyblocks/f/freefem.py
hajgato: no. here we just build part of the ff-petsc. (not all modules)

hajgato: We use the "root" configure for all the downloaded modules.

easybuild/easyblocks/f/freefem.py
hajgato: Indeed, but anyway, it checks whether things are already downloaded or not, so `PETSc` and `SLEPc` is not downloaded again. I do not remember why I did build the `PETSc` and `SLEPc` in advance, but maybe some of the downloaded modules needs it at compilation time, and this is the way how I made sure that they are already built.

If you want, I can remove `PETSc` and `SLEPc`, but anyway, they would not be downloaded again.

easybuild/easyblocks/f/freefem.py
hajgato: I do not understand this comment. I thought that I just hard-coded in the easyblock the sanity-check files and dirs, nothing else. (So we do not have to put in the easyconfig)

lexming:
@hajgato do you agree to close this PR in favour of https://github.com/easybuilders/easybuild-easyblocks/pull/1969?

hajgato:
@lexming Sure.

PR #1489, opened 2018-08-29
added regex to fix WM_PROJECT_VERSION correctly

mboisson:
This fixes WM_PROJECT_VERSION in OpenFOAM's bashrc files to use whatever is used by the easyconfig. It allows to use custom version numbers since OpenFOAM is so bad at versioning their software. .

### DIFF
--- a/easybuild/easyblocks/o/openfoam.py
+++ b/easybuild/easyblocks/o/openfoam.py
@@ -155,7 +155,12 @@ class EB_OpenFOAM(EasyBlock):
             self.log.debug("Patching out hardcoded $WM_* env vars in %s", script)
             # disable any third party stuff, we use EB controlled builds
             regex_subs = [(r"^(setenv|export) WM_THIRD_PARTY_USE_.*[ =].*$", r"# \g<0>")]
-            regex_subs += [(r"^(setenv|export) WM_PROJECT_VERSION=.*$", r"export WM_PROJECT_VERSION=%s #\g<0>" % self.version)]
+
+            # replacement dictionary of key-values
+            replacement_dict = {"WM_PROJECT_VERSION": self.version}
+            for key,val in replacement_dict.items():
+                regex_subs += [(r"^(setenv|export) %s=.*$" % key, r"export %s=%s #\g<0>" % (key, val))]
+
             WM_env_var = ['WM_COMPILER', 'WM_MPLIB', 'WM_THIRD_PARTY_DIR']
             # OpenFOAM >= 3.0.0 can use 64 bit integers
             if 'extend' not in self.name.lower() and LooseVersion(self.version) >= LooseVersion('3.0'):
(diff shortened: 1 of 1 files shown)


akesandgren:
@mboisson move the PR to develop instead of master please.

boegel:
@akesandgren You can change the target branch (as maintainer or PR author) via the `Edit` button (I've done that just now).

boegel:
@mboisson This looks good to me, but it will need to be tested thoroughly on all existing `OpenFOAM(-Extend)` easyconfigs (I can look into that).

mboisson:
oops, my bad for master instead of develop @akesandgren

boegel:
@mboisson One problematic case is an old version of OpenFOAM-Extend:

```
$ grep "exportWM_PROJECT_VERSION" $EASYBUILD_PREFIX/software/OpenFOAM-*/*/etc/bashrc
...
/prefix/software/OpenFOAM-Extend/1.6-goolf-1.4.10-20130711/OpenFOAM-1.6-ext/etc/bashrc:export WM_PROJECT_VERSION=1.6-ext
```

You can dance around this and not touch `WM_PROJECT_VERSION` for `OpenFOAM-Extend` versions < 2.0 .

mboisson:
Ok, added a test case for OpenFOAM Extend < 2.0

easybuild/easyblocks/o/openfoam.py
```diff
@@ -155,14 +155,21 @@ def configure_step(self):
             self.log.debug("Patching out hardcoded $WM_* env vars in %s", script)
             # disable any third party stuff, we use EB controlled builds
             regex_subs = [(r"^(setenv|export) WM_THIRD_PARTY_USE_.*[ =].*$", r"# \g<0>")]
+
+            # this does not work for OpenFOAM Extend lower than 2.0
+            if 'extend' not in self.name.lower() or LooseVersion(self.version) >= LooseVersion('2.0'):
+                # replacement dictionary of key-values
+                replacement_dict = {"WM_PROJECT_VERSION": self.version}
+                for key, val in replacement_dict.items():
+                    regex_subs += [(r"^(setenv|export) %s=.*$" % key, r"export %s=%s #\g<0>" % (key, val))]
```
boegel: @mboisson Since `replacement_dict` only has a single entry, this can be collapsed to:

```python
if 'extend' not in self.name.lower() or LooseVersion(self.version) >= LooseVersion('2.0'):
    key = 'WM_PROJECT_VERSION'
    regex_subs.append((r"^(setenv|export) %s=.*$" % key, r"export %s=%s #\g<0>" % (key, self.version)))

boegel: If this makes the line too long, split it on the `.`:

```python
regex_subs.append((r"^(setenv|export) %s=.*$" % key,
                   r"export %s=%s #\g<0>" % (key, self.version)))

boegel:
Tested with existing `OpenFOAM` easyconfigs, (still) works like a charm, thanks @mboisson!

PR #1490, opened 2018-08-30
Trinity 2.8 sanity check

ebirn:
Trinity 2.8 uses CMake for Chrysalis build, resultin in bin/ subdirectory.
see https://github.com/trinityrnaseq/trinityrnaseq/blob/484b3c9668b79b402df3da04117157c0ef5fea25/Changelog.txt#L2

When this is ok, I can bring in a PR for Trinity 2.8

### DIFF
--- a/easybuild/easyblocks/t/trinity.py
+++ b/easybuild/easyblocks/t/trinity.py
@@ -319,11 +319,16 @@ class EB_Trinity(EasyBlock):
         else:
             sep = '_r'
 
+        if version >= LooseVersion('2.8'):
+            chrysalis_bin = 'Chrysalis/bin/Chrysalis'
+        else:
+            chrysalis_bin = 'Chrysalis/Chrysalis'
+
         path = 'trinityrnaseq%s%s' % (sep, self.version)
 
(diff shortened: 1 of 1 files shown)


boegel:
@ebirn Looks good, but isn't there an additional change needed to make the `Chrysalis` binary available via `$PATH`?
Not sure if that's relevant for `Trinity`, maybe `Chrysalis` is only used internally and not directly by end users of `Trinity`...

damianam:
@ebirn ping? I have the feeling that the binaries are symlinked somewhere else? Why this https://github.com/easybuilders/easybuild-easyblocks/blob/6fb52e8a09509aee292f4cd2aabe47f6cc1fc4f1/easybuild/easyblocks/t/trinity.py#L343 if not?

pescobar:
I tried this easyblock and works fine for me . Can we get it merged? should I send another PR if @ebirn doesn't react?

pescobar:
@ebirn asked me to close this one and create a new PR with the small remarks fixed so I close this one

please review here https://github.com/easybuilders/easybuild-easyblocks/pull/1579

PR #1494, opened 2018-08-31
IMPROVE PGI siterc so it allows -pthread switch

### LINKED ISSUE #1493: PGI doesn't accept "-pthread" switch
I'm trying to compile OpenMPI 2.1.2 with PGI 18.4 and I stumbled upon a pgi error saying that it doen't accept -pthread switch

After few research I saw this thread: https://www.pgroup.com/userforum/viewtopic.php?t=5253&sid=93356f57a47eee097b569b79bc744676
basicaly PGI accept -lpthread instead
and one solution for PGI to accept -pthread would be to modify the siterc in $PGI_INSTALL/bin/siterc

MisterFruits:
fix #1493

### DIFF
--- a/easybuild/easyblocks/p/pgi.py
+++ b/easybuild/easyblocks/p/pgi.py
@@ -64,6 +64,14 @@ append LDLIBARGS=$library_path;
 append LDLIBARGS=-L/usr/lib/x86_64-linux-gnu;
 """
 
+# contents for siterc file to make PGI accept the -pthread switch
+# cf. https://www.pgroup.com/userforum/viewtopic.php?t=5253&sid=93356f
+SITERC_PTHREAD_SWITCH = """
+# allow -pthread switch stands for -lpthread
+switch -pthread is
+append(LDLIB1=-lpthread);
+"""
+
 
 class EB_PGI(PackedBinary):
     """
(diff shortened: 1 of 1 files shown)


damianam:
We have something similar in our setup, but it does this:

`switch -pthread is replace(-lpthread) positional(linker); `

Looks to me like a bit better, since it will replace it just when linking.

boegel:
@MisterFruits Thoughts on @damianam's suggestion? Looks cleaner to me too...

MisterFruits:
I have to say that I'm not a PGI expert ;) eventually looks cleaner yes ! I'll update the PR but I won't be able to test

easybuild/easyblocks/p/pgi.py
```diff
@@ -64,6 +64,13 @@
 append LDLIBARGS=-L/usr/lib/x86_64-linux-gnu;
 """
 
+# contents for siterc file to make PGI accept the -pthread switch
+# cf. https://www.pgroup.com/userforum/viewtopic.php?t=5253&sid=93356f
+SITERC_PTHREAD_SWITCH = """
+# allow -pthread switch stands for -lpthread
```
akesandgren: I think "replace unknown switch -pthread with -lpthread" is a clearer comment.

easybuild/easyblocks/p/pgi.py
```diff
@@ -144,11 +151,14 @@ def install_step(self):
...
                     os.remove(path)
 
-        # install (or update) siterc file to make PGI consider $LIBRARY_PATH
+        # install (or update) siterc file to make PGI consider $LIBRARY_PATH and accept -pthread
         siterc_path = os.path.join(self.installdir, self.pgi_install_subdir, 'bin', 'siterc')
         write_file(siterc_path, SITERC_LIBRARY_PATH, append=True)
         self.log.info("Appended instructions to pick up $LIBRARY_PATH to siterc file at %s: %s",
                       siterc_path, SITERC_LIBRARY_PATH)
+        write_file(siterc_path, SITERC_PTHREAD_SWITCH, append=True)
+        self.log.info("Appended instructions to allow -pthread switch to siterc file at %s: %s",
```
akesandgren: And "Append instructions to replace -pthread with -lpthread to siterc file at ..." feels better to me.

akesandgren:
Have tested with and without this change, result with change is excellent.

boegel:
Thanks @MisterFruits!

akesandgren:
Going in, thanks @MisterFruits

PR #1497, opened 2018-09-01
Add easyblock for BWISE

wpoely86:


easybuild/easyblocks/b/bwise.py
```diff
@@ -0,0 +1,107 @@
...
+    def build_step(self, verbose=False, path=None):
+        """Run multiple times for different sources"""
+
+        makefiles_fixes = [
+            (r'^CC=.*$', 'CC=$(CXX)'),
+            (r'^CFLAGS=.*$', 'CFLAGS:=$(CFLAGS)'),
+            (r'^LDFLAGS=.*$', 'LDFLAGS:=$(LDFLAGS) -fopenmp')
+        ]
+
+        os.chdir(self.builddir)
```
boegel: @wpoely86 please use the `change_dir` function (everywhere, instead of `os.chdir`)

easybuild/easyblocks/b/bwise.py
```diff
@@ -0,0 +1,107 @@
...
+        extra['files_to_copy'][2] = CUSTOM
+        return extra
+
+    def build_step(self, verbose=False, path=None):
+        """Run multiple times for different sources"""
+
+        makefiles_fixes = [
+            (r'^CC=.*$', 'CC=$(CXX)'),
+            (r'^CFLAGS=.*$', 'CFLAGS:=$(CFLAGS)'),
+            (r'^LDFLAGS=.*$', 'LDFLAGS:=$(LDFLAGS) -fopenmp')
```
boegel: Hmm, maybe it's better to set the `openmp` toolchain option to `True` rather than hardcoding this? Hardcoding `-fopenmp` will work for GCC & (recent enough) Intel compilers, but not for others...

easybuild/easyblocks/b/bwise.py
```diff
@@ -0,0 +1,107 @@
...
+        ]
+
+        os.chdir(self.builddir)
+        # first BWISE itself
+        bwisepath = glob.glob('BWISE-*')
```
boegel: Why not `glob.glob(os.path.join(self.builddir, 'BWISE-*', 'src'))` and then complain if it gives an empty result?
That's only one `change_dir`, etc.

Also, what's wrong with `self.start_dir`? (I can imagine what the problem is, but please clarify that in a comment here).

easybuild/easyblocks/b/bwise.py
```diff
@@ -0,0 +1,107 @@
...
+
+        # Onwards to BGREAT
+        os.chdir(self.builddir)
+        bgreatpath = glob.glob('BGREAT2-*')
+        if bgreatpath:
+            os.chdir(bgreatpath[0])
+        else:
+            raise EasyBuildError("Could not find BGREAT path")
+        apply_regex_substitutions('makefile', makefiles_fixes)
+        super(EB_BWISE, self).build_step()
```
boegel: This is way too similar to the above, please flesh this out in a helper method that you can call out to. Something like:

```python
# first BWISE itself
subdir = self.build_in_subdir(os.path.join('BWISE-*', 'src'))
bwise_py = os.path.join(subdir, '..', 'Bwise.py')
apply_regex_substitutions(bwise_py, [(r'^BWISE_MAIN = .*$', 'BWISE_MAIN = os.environ[\'EBROOTBWISE\']')])

# onwards to BGREAT
subdir = self.build_in_subdir('BGREAT-*')
copy_file(os.path.join(subdir, 'bgreat'), self.cfg['start_dir'])

# Finally, BTRIM
subdir = self.build_in_subdir('BTRIM-*')
copy_file(os.path.join(subdir, 'btrim'), self.cfg['start_dir'])

File: easybuild/easyblocks/b/bwise.py [hunk 5]
```diff
@@ -0,0 +1,107 @@
...
+
+        binaries = ['sequencesToNumbers', 'numbersFilter', 'path_counter', 'maximal_sr', 'simulator',
+                    'path_to_kmer', 'K2000/*.py', 'K2000/*.sh']
+        self.cfg['files_to_copy'] = [(['bgreat', 'btrim', 'Bwise.py'] + ['src/%s' % x for x in binaries], 'bin'),
+                                     'data']
+
+    def install_step(self):
+        super(EB_BWISE, self).install_step()
+
+        # BWISE expects it to be at this location...
```
boegel: @wpoely86 Please clarify where `bcalm` comes from, i.e. needs to be available via a `BCALM` dependency? better make sure of that, via `which('bcalm')` (not here I guess, but earlier, `prepare_step` maybe)?

easybuild/easyblocks/b/bwise.py
wpoely86: The issue is that `-fopenmp` is not adding to `LDFLAGS`, only `CFLAGS`

easybuild/easyblocks/b/bwise.py
```diff
@@ -0,0 +1,120 @@
...
+                raise EasyBuildError("Could not find a subdirectory matching the pattern %s", pattern)
+
+        # BWISE has 3 independant parts, we build them one by one
+        # first BWISE itself
+        subdir = find_subdir(os.path.join('BWISE-*', 'src'))
+        change_dir(subdir)
+        apply_regex_substitutions('makefile', makefiles_fixes)
+        apply_regex_substitutions(os.path.join(subdir, '..', 'Bwise.py'),
+                                  [(r'^BWISE_MAIN = .*$', 'BWISE_MAIN = os.environ[\'EBROOTBWISE\']')])
+        super(EB_BWISE, self).build_step()
```
boegel: Why not include this in `find_subdir` (you'll need to rename it of course) too, since they are repeated below?

```
change_dir(subdir)
apply_regex_substitutions('makefile', makefiles_fixes)
super(EB_BWISE, self).build_step()
```

You can tweak `Bwise.py` after calling `build_step`?

easybuild/easyblocks/b/bwise.py
```diff
@@ -0,0 +1,120 @@
...
+                                     'data']
+
+    def install_step(self):
+        super(EB_BWISE, self).install_step()
+
+        # BWISE expects BCALM to be at exactly this location...
+        bcalm = which('bcalm')
+        txt = """#!/bin/sh
+%s "$@"
+        """ % bcalm
```
boegel: I don't think injecting the full path to `bcalm` is a good idea... Just leave this wrapper as it was?

easybuild/easyblocks/b/bwise.py
```diff
@@ -0,0 +1,120 @@
...
+                    'path_to_kmer', 'K2000/*.py', 'K2000/*.sh']
+        self.cfg['files_to_copy'] = [(['bgreat', 'btrim', 'Bwise.py'] + ['src/%s' % x for x in binaries], 'bin'),
+                                     'data']
+
+    def install_step(self):
+        super(EB_BWISE, self).install_step()
+
+        # BWISE expects BCALM to be at exactly this location...
+        bcalm = which('bcalm')
+        txt = """#!/bin/sh
```
boegel: rename `txt` to `bcalm_wrapper`?

easybuild/easyblocks/b/bwise.py
```diff
@@ -0,0 +1,120 @@
...
+        """Change default values of options"""
+        extra = MakeCp.extra_options()
+        # files_to_copy is not mandatory here
+        extra['files_to_copy'][2] = CUSTOM
+        return extra
+
+    def build_step(self):
+        """Run multiple times for different sources"""
+
+        # BCALM is a git submodule of BWISE but we use it as a dependency
```
boegel: Is it worth mentioning *why* BCALM is used as a dep, rather than just being included in the BWISE installation like you do with BGREAT & BTRIM?

PR #1498, opened 2018-09-04
FSL - Improve configuration choice

vanzod:
`FSL` build relies on an internal check and a script to identify the appropriate configuration to use. Since the available configurations are for relatively old compilers, the current easyblock uses `difflib.get_close_matches` to identify the closest configuration and copies the directory containing it into a new one named after the configuration targeted by `FSL`. In certain cases the guessed closest configuration is not the same as the one at which required patches are applied, resulting in a build failure.
This easyblock change uses the patch files as first source of information to guess the target directory and rolls back to `difflib.get_close_matches` only if no target configuration is found in the patches.

### DIFF
--- a/easybuild/easyblocks/f/fsl.py
+++ b/easybuild/easyblocks/f/fsl.py
@@ -33,6 +33,7 @@ import os
 import re
 import shutil
 from distutils.version import LooseVersion
+from vsc.utils.missing import nub
 
 import easybuild.tools.environment as env
 from easybuild.framework.easyblock import EasyBlock
@@ -71,16 +72,16 @@ class EB_FSL(EasyBlock):
             systype_regex = re.compile("^diff.*config\/(.*(apple|gnu|i686|linux|spark)(?:(?!\/).)*)", re.M)
 
             patched_cfgs = []
-        
+
             for patch in self.patches:
                 patchfile = read_file(patch['path'])
                 res = systype_regex.findall(patchfile)
                 patched_cfgs.extend([i[0] for i in res])
 
             # Check that at least one config has been found
-            if len(patched_cfgs) != 0:
+            if patched_cfgs:
                 # Check that a single config has been patched
-                if patched_cfgs[1:] == patched_cfgs[:-1]:
+                if len(nub(patched_cfgs)) == 1:
                     best_cfg = patched_cfgs[0]
                     self.log.debug("Found patched config dir: %s", best_cfg)
                 else:
@@ -95,8 +96,8 @@ class EB_FSL(EasyBlock):
                 cfgs = os.listdir(cfgdir)
                 best_cfg = difflib.get_close_matches(fslmachtype, cfgs)[0]
                 self.log.debug("Best matching config dir for %s is %s" % (fslmachtype, best_cfg))
-        except OSError, err:
-            raise EasyBuildError("Unable to access configurations directory: %s", err)
+        except OSError as err:
+            raise EasyBuildError("Unable to access configuration directory: %s", cfgdir, err)
 
         # Prepare config
         # Either use patched config or copy closest match
--- a/easybuild/easyblocks/f/fsl.py
+++ b/easybuild/easyblocks/f/fsl.py
@@ -32,6 +32,7 @@ import difflib
 import os
 import re
 import shutil
+from distutils.version import LooseVersion
 
 import easybuild.tools.environment as env
 from easybuild.framework.easyblock import EasyBlock
@@ -64,26 +65,28 @@ class EB_FSL(EasyBlock):
         fslmachtype = out.strip()
         self.log.debug("FSL machine type: %s" % fslmachtype)
 
-        # Check if a specific machine type directory is patched
-        systype_regex = re.compile("^diff.*config\/(.*(apple|gnu|i686|linux|spark)(?:(?!\/).)*)", re.M)
+        # Patch files for ver. < 5.0.10 patch multiple config directories
+        if LooseVersion(self.version) >= LooseVersion('5.0.10'):
+            # Check if a specific machine type directory is patched
+            systype_regex = re.compile("^diff.*config\/(.*(apple|gnu|i686|linux|spark)(?:(?!\/).)*)", re.M)
 
-        patched_cfgs = []
+            patched_cfgs = []
         
-        for patch in self.patches:
-            patchfile = read_file(patch['path'])
-            res = systype_regex.findall(patchfile)
-            patched_cfgs.extend([i[0] for i in res])
+            for patch in self.patches:
+                patchfile = read_file(patch['path'])
+                res = systype_regex.findall(patchfile)
+                patched_cfgs.extend([i[0] for i in res])
 
-        # Check that at least one config has been found
-        if len(patched_cfgs) != 0:
-            # Check that a single config has been patched
-            if patched_cfgs[1:] == patched_cfgs[:-1]:
-                best_cfg = patched_cfgs[0]
-                self.log.debug("Found patched config dir: %s", best_cfg)
+            # Check that at least one config has been found
+            if len(patched_cfgs) != 0:
+                # Check that a single config has been patched
+                if patched_cfgs[1:] == patched_cfgs[:-1]:
+                    best_cfg = patched_cfgs[0]
+                    self.log.debug("Found patched config dir: %s", best_cfg)
+                else:
+                    raise EasyBuildError("Patch files are editing multiple config dirs: %s", patched_cfgs)
             else:
-                raise EasyBuildError("Patch files are editing multiple config dirs: %s", patched_cfgs)
-        else:
-            self.log.debug("No config dir found in patch files")
+                self.log.debug("No config dir found in patch files")
 
         # If no patched config is found, pick best guess
         cfgdir = os.path.join(self.fsldir, "config")
--- a/easybuild/easyblocks/f/fsl.py
+++ b/easybuild/easyblocks/f/fsl.py
@@ -68,6 +68,7 @@ class EB_FSL(EasyBlock):
         diff_regex = re.compile("^diff.*config\/", re.M)
         
         patched_cfgs = []
+        best_cfg = None
 
         for patch in self.patches:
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/f/fsl.py
```diff
@@ -63,15 +63,43 @@ def configure_step(self):
...
+        systype_regex = re.compile(".*(apple|gnu|i686|linux|spark).*", re.M)
+        diff_regex = re.compile("^diff.*config\/", re.M)
+        
+        patched_cfgs = []
+        best_cfg = None
+
+        for patch in self.patches:
+            try:
+                for line in open(patch['path'], 'r'):
+                    if all(regex.match(line) for regex in [systype_regex, diff_regex]):
```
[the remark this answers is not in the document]
vanzod: The reason they are separate is because I use `systype_regex` twice and only once in conjunction with `diff_regex`. Hence I preferred to split them like that instead of having `systype_regex` being a subset of `diff_regex` (i.e. avoid copy/paste).

easybuild/easyblocks/f/fsl.py
```diff
@@ -63,15 +63,43 @@ def configure_step(self):
...
+                        matches = [m.group(0) for l in line.split('/') for m in [systype_regex.search(l)] if m]
+                        patched_cfgs.extend(matches)
+            except OSError, err:
+                raise EasyBuildError("Unable to open patch file: %s", patch['path'])
+            
+        if patched_cfgs:
+            if patched_cfgs[1:] == patched_cfgs[:-1]:
```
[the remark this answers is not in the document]
vanzod: That would work in the first `if` statement. In the second I am checking that all strings in `patched_cfgs` are the same, which means that the patch targets only one configuration.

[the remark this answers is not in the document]
boegel: OK, then please clarify with a comment above this line, that's not clear at all :)

boegel:
@vanzod It looks like this can only work with FSL 5.0.10 & newer (works with `FSL-5.0.10-intel-2017a.eb` and `FSL-5.0.11-foss-2018b.eb`

easybuild/easyblocks/f/fsl.py
```diff
@@ -63,22 +65,47 @@ def configure_step(self):
...
 
-        # prepare config
-        # either using matching config, or copy closest match
+        # Patch files for ver. < 5.0.10 patch multiple config directories
+        if LooseVersion(self.version) >= LooseVersion('5.0.10'):
+            # Check if a specific machine type directory is patched
+            systype_regex = re.compile("^diff.*config\/(.*(apple|gnu|i686|linux|spark)(?:(?!\/).)*)", re.M)
+
+            patched_cfgs = []
+
```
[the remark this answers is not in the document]
boegel: @vanzod The hound is merciless...

easybuild/easyblocks/f/fsl.py
```diff
@@ -63,22 +65,47 @@ def configure_step(self):
...
+        # If no patched config is found, pick best guess
         cfgdir = os.path.join(self.fsldir, "config")
         try:
-            cfgs = os.listdir(cfgdir)
-            best_cfg = difflib.get_close_matches(fslmachtype, cfgs)[0]
+            if not best_cfg:
+                cfgs = os.listdir(cfgdir)
+                best_cfg = difflib.get_close_matches(fslmachtype, cfgs)[0]
+                self.log.debug("Best matching config dir for %s is %s" % (fslmachtype, best_cfg))
+        except OSError, err:
```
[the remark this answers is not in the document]
boegel: This should probably be `except OSError as err:`

easybuild/easyblocks/f/fsl.py
```diff
@@ -63,22 +65,47 @@ def configure_step(self):
...
+
+            patched_cfgs = []
+        
+            for patch in self.patches:
+                patchfile = read_file(patch['path'])
+                res = systype_regex.findall(patchfile)
+                patched_cfgs.extend([i[0] for i in res])
+
+            # Check that at least one config has been found
+            if len(patched_cfgs) != 0:
```
boegel: `if patched_cfgs:` is fine here

easybuild/easyblocks/f/fsl.py
```diff
@@ -63,22 +65,47 @@ def configure_step(self):
...
+                # Check that a single config has been patched
+                if patched_cfgs[1:] == patched_cfgs[:-1]:
```
boegel: @vanzod This hurts my brain... How about using `nub` from `vsc.utils.missing` and change this to:

```python
if len(nub(patched_cfgs)) == 1:
```

easybuild/easyblocks/f/fsl.py
```diff
@@ -63,22 +65,47 @@ def configure_step(self):
...
         cfgdir = os.path.join(self.fsldir, "config")
         try:
-            cfgs = os.listdir(cfgdir)
-            best_cfg = difflib.get_close_matches(fslmachtype, cfgs)[0]
+            if not best_cfg:
+                cfgs = os.listdir(cfgdir)
+                best_cfg = difflib.get_close_matches(fslmachtype, cfgs)[0]
+                self.log.debug("Best matching config dir for %s is %s" % (fslmachtype, best_cfg))
+        except OSError, err:
+            raise EasyBuildError("Unable to access configurations directory: %s", err)
```
boegel: include value of `cfgdir` in error message please

easybuild/easyblocks/f/fsl.py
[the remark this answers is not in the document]
vanzod: I like that

easybuild/easyblocks/f/fsl.py
vanzod: I like that. I did not know about `nub`

easybuild/easyblocks/f/fsl.py
```diff
@@ -63,22 +66,47 @@ def configure_step(self):
...
-                shutil.copytree(srcdir, tgtdir)
-                self.log.debug("Copied %s to %s" % (srcdir, tgtdir))
-        except OSError, err:
-            raise EasyBuildError("Failed to copy closest matching config dir: %s", err)
+            if not best_cfg:
+                cfgs = os.listdir(cfgdir)
+                best_cfg = difflib.get_close_matches(fslmachtype, cfgs)[0]
+                self.log.debug("Best matching config dir for %s is %s" % (fslmachtype, best_cfg))
+        except OSError as err:
+            raise EasyBuildError("Unable to access configuration directory: %s", cfgdir, err)
```
boegel: @vanzod You have a `%s` missing here, this is going to result in a crash rather than an error message, should be:

```python
raise EasyBuildError("Unable to access configuration directory %s: %s", cfgdir, err)

File: easybuild/easyblocks/f/fsl.py [hunk 9]
```diff
@@ -63,22 +66,47 @@ def configure_step(self):
...
-            self.log.debug("Best matching config dir for %s is %s" % (fslmachtype, best_cfg))
-
-            if fslmachtype != best_cfg:
-                srcdir = os.path.join(cfgdir, best_cfg)
-                tgtdir = os.path.join(cfgdir, fslmachtype)
-                shutil.copytree(srcdir, tgtdir)
-                self.log.debug("Copied %s to %s" % (srcdir, tgtdir))
-        except OSError, err:
-            raise EasyBuildError("Failed to copy closest matching config dir: %s", err)
+            if not best_cfg:
```
boegel: I'm hitting this now:

```
  File "/tmp/eb-EdQY9x/included-easyblocks/easybuild/easyblocks/fsl.py", line 95, in configure_step
    if not best_cfg:
UnboundLocalError: local variable 'best_cfg' referenced before assignment
```

vanzod: It looks like I have accidentally deleted the definition of `best_cfg` during the refactoring

easybuild/easyblocks/f/fsl.py
```diff
@@ -63,22 +66,48 @@ def configure_step(self):
...
 
-        # prepare config
-        # either using matching config, or copy closest match
+        # Patch files for ver. < 5.0.10 patch multiple config directories
+        if LooseVersion(self.version) >= LooseVersion('5.0.10'):
+            # Check if a specific machine type directory is patched
+            systype_regex = re.compile("^diff.*config\/(.*(apple|gnu|i686|linux|spark)(?:(?!\/).)*)", re.M)
+
+            patched_cfgs = []
+            best_cfg = None
```
boegel: @vanzod This needs to go outside of this `if` clause to ensure it's always defined for old FSL versions?

vanzod: True. Sloppy brain

boegel:
Tests with existing easyconfigs are under way...

easybuild/easyblocks/f/fsl.py
```diff
@@ -63,22 +66,49 @@ def configure_step(self):
...
-            if fslmachtype != best_cfg:
-                srcdir = os.path.join(cfgdir, best_cfg)
-                tgtdir = os.path.join(cfgdir, fslmachtype)
-                shutil.copytree(srcdir, tgtdir)
-                self.log.debug("Copied %s to %s" % (srcdir, tgtdir))
-        except OSError, err:
-            raise EasyBuildError("Failed to copy closest matching config dir: %s", err)
+            if not best_cfg:
+                cfgs = os.listdir(cfgdir)
+                best_cfg = difflib.get_close_matches(fslmachtype, cfgs)[0]
```
boegel: @vanzod This whole block is now below the `if` for FSL 5.0.10 or newer, while it wasn't before. That makes this change backward incompatible, resulting in breaking the installation of `FSL-5.0.9-intel-2016a.eb` for example...

boegel: Wait, no, it's not... But there still seems to be a problem with `FSL-5.0.9-intel-2016a.eb` (will re-check).

boegel:
Good to go, thanks @vanzod!

PR #1503, opened 2018-09-07
add generic ModulesRC easyblock

boegel:
This is a new generic easyblock that generates a `.modulerc` file to define a module version alias rather than installing an actual module file (cfr. suggestion by @mboisson in https://github.com/easybuilders/easybuild-easyconfigs/pull/5203#issuecomment-415827214).

This can be used to for example install a `Java/1.8` wrapper (see https://github.com/easybuilders/easybuild-easyconfigs/pull/6712) for another `Java` module, without causing trouble with for example Lmod's module collection concept...

edit: now requires ~~https://github.com/easybuilders/easybuild-framework/pull/2571~~ and ~~https://github.com/easybuilders/easybuild-framework/pull/2575~~

### DIFF
--- a/easybuild/easyblocks/generic/modulerc.py
+++ b/easybuild/easyblocks/generic/modulerc.py
@@ -68,7 +68,7 @@ class ModuleRC(EasyBlock):
         if self.name != deps[0]['name']:
             raise EasyBuildError("Name does not match dependency name: %s vs %s", self.name, deps[0]['name'])
 
-        # version to prefix of version to alias to
+        # ensure version to alias to is a prefix of the version of the dependency
         if not deps[0]['version'].startswith(self.version):
             raise EasyBuildError("Version is not a prefix of dependency version: %s vs %s",
                                  self.version, deps[0]['version'])
--- a/easybuild/easyblocks/generic/modulerc.py
+++ b/easybuild/easyblocks/generic/modulerc.py
@@ -36,7 +36,7 @@ from easybuild.tools.filetools import write_file
 from easybuild.tools.module_generator import ModuleGeneratorTcl
 
 
-class ModuleAlias(EasyBlock):
+class ModuleRC(EasyBlock):
     """
     Generic easyblock to create a software-specific .modulerc file
     """
@@ -77,10 +77,8 @@ class ModuleAlias(EasyBlock):
         alias_modname = deps[0]['full_mod_name']
         self.log.info("Adding module version alias for %s to %s", alias_modname, modulerc)
 
-        modulerc_txt = '\n'.join([
-            ModuleGeneratorTcl.MODULE_SHEBANG,
-            "module-version %s %s" % (alias_modname, self.version),
-        ])
+        module_version_specs = {'modname': alias_modname, 'sym_version': self.version, 'version': deps[0]['version']}
+        modulerc_txt = self.module_generator.modulerc(module_version=module_version_specs)
         write_file(modulerc, modulerc_txt)
 
         modpath = self.module_generator.get_modules_path(fake=fake)
--- a/easybuild/easyblocks/generic/modulerc.py
+++ b/easybuild/easyblocks/generic/modulerc.py
@@ -62,7 +62,7 @@ class ModuleRC(EasyBlock):
 
         deps = self.cfg['dependencies']
         if len(deps) != 1:
-            raise EasyBuildError("There should be only one single dependency specified, found %d", len(deps))
+            raise EasyBuildError("There should be exactly one dependency specified, found %d", len(deps))
 
         # names should match
         if self.name != deps[0]['name']:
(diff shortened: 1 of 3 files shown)


boegel:
The tests only fail with EnvironmentModulesC 3.2.10, looks like there was a weird bug with `module-version` there, see https://sourceforge.net/p/modules/mailman/message/31144974 + https://sourceforge.net/p/modules/mailman/message/33399425...

@xdelaruelle Any ideas what's going on here? In this case, there's only a singe `module-version` statement in the `.modulerc` file, like this:

```
#%Module
module-version Java/1.8.0_181 1.8
```

Then how can it trigger a `Duplicate version symbol` error?!
```

**Comment by boegel** [comment]
OK, the problem was probably that there was an actual module file `foo/1.3.2` together with a module alias for `foo` defined via `module-version` in `.modulerc` that was also using `1.3.2` as version, which results in a clash...

That should be fixed now by user different versions for the `ModulesAlias` test.

**Comment by xdelaruelle** [comment]
@boegel It seems that `.modulerc` files are parsed twice on version 3.2 which leads to these `Duplicate version symbol` messages. The symbolic version is however correctly set, so I believe this message could be ignored on your test cases.

File: easybuild/easyblocks/generic/modulerc.py [hunk 1]
```diff
@@ -0,0 +1,94 @@
...
+    def install_step(self):
+        """Do nothing."""
+        pass
+
+    def make_module_step(self, fake=False):
+        """Install .modulerc file."""
+        modfile_path = self.module_generator.get_module_filepath(fake=fake)
+        modulerc = os.path.join(os.path.dirname(modfile_path), '.modulerc')
+
+        if os.path.exists(modulerc) and not build_option('force'):
```
ocaisa: This implies that you can only have one entry in the `.modulerc` per software package. Probably ok, but worth noting the limitation

easybuild/easyblocks/generic/modulerc.py
```diff
@@ -0,0 +1,94 @@
...
+            raise EasyBuildError("Found existing .modulerc at %s, not overwriting without --force", modulerc)
+
+        deps = self.cfg['dependencies']
+        if len(deps) != 1:
+            raise EasyBuildError("There should be exactly one dependency specified, found %d", len(deps))
+
+        # names should match
+        if self.name != deps[0]['name']:
```
ocaisa: Given the discussion about have a `Pythoncore` (or similar) in https://github.com/easybuilders/easybuild-easyconfigs/pull/6537 I wonder if we should consider having exceptions here? In that case I could imagine using `.modulrerc` to alias back to the underlying installation (at least for `GCC`)

ocaisa: Nah, forget that

easybuild/easyblocks/generic/modulerc.py
```diff
@@ -0,0 +1,94 @@
...
+            raise EasyBuildError("Name does not match dependency name: %s vs %s", self.name, deps[0]['name'])
+
+        # version to prefix of version to alias to
```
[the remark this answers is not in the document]
boegel: Yes, exactly that, and you're right, the comment that was there didn't make any sense, it should have been something like `# version is prefix of version to alias to`, but yours is better :)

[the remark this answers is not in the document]
boegel: Yes, exactly that, and you're right, the comment that was there didn't make any sense, it should have been something like `# version is prefix of version to alias to`, but yours is better :)

easybuild/easyblocks/generic/modulerc.py
boegel: Indeed, that's a rather strict limitation for now, which we can review later when the need for it pops up?

A whole bunch of questions come into play to relax this:

* What do we do with an existing `.modulerc` that was not created by EasyBuild?
* How do we allow both appending to an existing `.modulerc` and overwriting the existing one (from the `eb` command line, I don't think specifying this in the easyconfig makes much sense)?
* Should we detect potentially conflicting lines in an existing `.modulerc`?

boegel: Indeed, that's a rather strict limitation for now, which we can review later when the need for it pops up?

A whole bunch of questions come into play to relax this:

* What do we do with an existing `.modulerc` that was not created by EasyBuild?
* How do we allow both appending to an existing `.modulerc` and overwriting the existing one (from the `eb` command line, I don't think specifying this in the easyconfig makes much sense)?
* Should we detect potentially conflicting lines in an existing `.modulerc`?

easybuild/easyblocks/generic/modulerc.py
boegel: Regardless of whether what popped into your head makes sense or not: starting out strict and relaxing it later is a lot easier than the other way around... :)

Issue #1450, opened 2018-07-04
HDF5 lacks pkgconfig files


akesandgren:
At least Ubuntu packaging of HDF5 contains pkgconfig files for HDF5. The EB build doesn't add any which causes some codes to find the OS-installed hdf5 instead.
The Ubuntu pkgconfig files are an addition to HDF5, but they should be fairly easy to create on the fly.

boegel:
@akesandgren Any reference to how this is done in Ubuntu?

akesandgren:
They just add prewritten (well a sed 's/@MULTIARCH@/$(DEB_HOST_MULTIARCH)/g' is used on the template) hdf5.pc files to /usr/lib/pkgconfig.
Nothing fancy, should be trivial to do inside the easyblock.

boegel:
fixed in #1504

PR #1504, opened 2018-09-11
Add a pkgconfig file to HDF5 when installing.

akesandgren:
Add a simple pkgconfig file to HDF5 during install, and add corresponding environment variable to the module file.

Should solve https://github.com/easybuilders/easybuild-easyblocks/issues/1450

### DIFF
--- a/easybuild/easyblocks/h/hdf5.py
+++ b/easybuild/easyblocks/h/hdf5.py
@@ -38,8 +38,18 @@ import easybuild.tools.environment as env
 import easybuild.tools.toolchain as toolchain
 from easybuild.easyblocks.generic.configuremake import ConfigureMake
 from easybuild.tools.build_log import EasyBuildError
+from easybuild.tools.filetools import write_file
 from easybuild.tools.modules import get_software_root
 from easybuild.tools.systemtools import get_shared_lib_ext
+
+# pkgconfig data template
+HDF5_PKG_CONFIG = """Name: HDF5
+Description: Hierarchical Data Format 5 (HDF5)
+Version: %s
+Requires:
+Cflags: -I%s
+Libs: -L%s -lhdf5
+"""
 
 class EB_HDF5(ConfigureMake):
     """Support for building/installing HDF5"""
@@ -81,7 +91,14 @@ class EB_HDF5(ConfigureMake):
 
         super(EB_HDF5, self).configure_step()
 
-    # default make and make install are ok
+    # default make and make install are ok but add a pkconfig file
+    def install_step(self):
+        """Custom install step for HDF5"""
+        super(EB_HDF5, self).install_step()
+        write_file(os.path.join(self.installdir, "lib", "pkgconfig", "hdf5.pc"),
+                   HDF5_PKG_CONFIG % (self.version, os.path.join(self.installdir, "include"),
+                                      os.path.join(self.installdir, "lib")),
+                   forced=True)
 
     def sanity_check_step(self):
         """
@@ -106,6 +123,15 @@ class EB_HDF5(ConfigureMake):
             'dirs': ['include'],
         }
         super(EB_HDF5, self).sanity_check_step(custom_paths=custom_paths)
+
+    def make_module_req_guess(self):
+        """Specify pkgconfig path for HDF5."""
+        guesses = super(EB_HDF5, self).make_module_req_guess()
+        guesses.update({
+            'PKG_CONFIG_PATH': [os.path.join('lib', 'pkgconfig')],
+        })
+
+        return guesses
 
     def make_module_extra(self):
         """Also define $HDF5_DIR to installation directory."""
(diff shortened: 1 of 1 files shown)


akesandgren:
Just ran into a package that got hit by the lack of a pkgconfig file from HDF5. With this patch and a rebuild of the HDF5 version in the dependency made the problem go away.

So it does work.

akesandgren:
Any chance of this making it into 3.7.0?

easybuild/easyblocks/h/hdf5.py
```diff
@@ -81,7 +91,14 @@ def configure_step(self):
...
 
-    # default make and make install are ok
+    # default make and make install are ok but add a pkconfig file
+    def install_step(self):
+        """Custom install step for HDF5"""
+        super(EB_HDF5, self).install_step()
+        write_file(os.path.join(self.installdir, "lib", "pkgconfig", "hdf5.pc"),
+                   HDF5_PKG_CONFIG % (self.version, os.path.join(self.installdir, "include"),
+                                      os.path.join(self.installdir, "lib")),
+                   forced=True)
```
boegel: @akesandgren This is a bit... messy. How about:

```python
inc_dir = os.path.join(self.installdir, 'include')
lib_dir = os.path.join(self.installdir, 'lib')
hdf5_pc_txt = HDF5_PKG_CONFIG % (self.version, inc_dir, lib_dir)
write_file(os.path.join(lib_dir, 'pkgconfig', 'hdf5.pc'), hdf5_pc_txt, forced=True)
```

Can you clarify the `forced=True`?

easybuild/easyblocks/h/hdf5.py
```diff
@@ -107,6 +124,15 @@ def sanity_check_step(self):
         }
         super(EB_HDF5, self).sanity_check_step(custom_paths=custom_paths)
 
+    def make_module_req_guess(self):
+        """Specify pkgconfig path for HDF5."""
+        guesses = super(EB_HDF5, self).make_module_req_guess()
+        guesses.update({
+            'PKG_CONFIG_PATH': [os.path.join('lib', 'pkgconfig')],
+        })
```
boegel: No need to wrap across multiple lines:

```python
guesses.update({'PKG_CONFIG_PATH': [os.path.join('lib', 'pkgconfig')]})

File: easybuild/easyblocks/h/hdf5.py [hunk 3]
```diff
@@ -38,9 +38,19 @@
...
 from easybuild.tools.modules import get_software_root
 from easybuild.tools.systemtools import get_shared_lib_ext
 
+# pkgconfig data template
+HDF5_PKG_CONFIG = """Name: HDF5
+Description: Hierarchical Data Format 5 (HDF5)
+Version: %s
+Requires:
+Cflags: -I%s
+Libs: -L%s -lhdf5
```
boegel: @akesandgren Did you come up with this yourself, or is there an upstream reference for it somewhere?

It looks OK, but I do wonder if this will cause problems somewhere because we have a `.pc` in place all of a sudden...

What about dependencies like `Szip` etc., should we take those into account here too?

easybuild/easyblocks/h/hdf5.py
akesandgren: Just picked forced=True because i felt like it. Want me to remove it?

easybuild/easyblocks/h/hdf5.py
akesandgren: Based on Debian packaging file(s). (Added to the comment)
Yes, it should most likely contain linking with zlib and Szip, but also libs from any other dependency a specific easyconfig uses, like libxml2.

Didn't want to make it too complicated to start with though.

boegel: I understand, but won't this possible impact stuff that depends on `HDF5`?

I guess testing will tell...

easybuild/easyblocks/h/hdf5.py
boegel: If there's no reason to use `forced=True`, then yes, remove it (but that's already done I see). :)

easybuild/easyblocks/h/hdf5.py
akesandgren: It should be done properly before merging, but I haven't gotten around to test it fully yet.
Don't really know what to test it with to see if -lz -lszip is needed.

boegel: The current definition of only including `-lhdf5` worked fine with `h5py` (but I'm not sure the `.pc` is even picked up there).

Making this more complete shouldn't make things worse?

We should dynamically complete this though, something:

```python
known_deps = {
    'libxml2': '-lxml2',
    'Szip': '-lsz',  # not sure what's correct here
    'zlib': '-lz',
}
for dep in known_deps:
    # compose value for 'Libs' using known_deps[dep]
```

or even:

```python
for dep in self.cfg.dependencies:
  if dep in known_deps:
    # compose value for 'Libs' using known_deps[dep]
  else:
    raise EasyBuildError("Unknown dependency for HDF5, don't know how to update pkgconfig file: %s", dep)

akesandgren:
How about this? Didn't do the "unknown dependency" part.

boegel:
Looks good, I see no reason to hold this back, we'll learn soon enough if this causes trouble somewhere...

Thanks @akesandgren!

PR #1506, opened 2018-09-12
sidestep using the config.guess that comes with a package in ConfigureMake generic easyblock

ocaisa:


micaeljtoliveira:
@ocaisa After fixing the typo in the script invocation (`config.guess` instead of `config_guess`), I tested this on my powerpc by building a package with an outdated `config.gues` an it worked like a charm.

easybuild/easyblocks/generic/configuremake.py
```diff
@@ -83,11 +83,16 @@ def configure_step(self, cmd_prefix=''):
         if prefix_opt is None:
             prefix_opt = '--prefix='
 
-        cmd = "%(preconfigopts)s %(cmd_prefix)s./configure %(prefix_opt)s%(installdir)s %(configopts)s" % {
+        # Avoid using config.guess from the package as it is frequently out of date, use the version shipped with EB
+        build_type, _ = run_cmd('config_guess')
+        build_type_option = '--build=' + build_type.strip()
```
boegel: @ocaisa As indicated by @micaeljtoliveira, this should be `config.guess`.
Also, please log what the result of `config_guess` is, and maybe also which `config.guess` is being (result of `which('config.guess')`).

Can we detect when `--build` needs to be used?
I'm a bit worried about the impact this may have on systems where this isn't needed currently...

easybuild/easyblocks/generic/configuremake.py
```diff
@@ -83,11 +83,16 @@ def configure_step(self, cmd_prefix=''):
...
+
+        cmd = "%(preconfigopts)s %(cmd_prefix)s./configure %(prefix_opt)s%(installdir)s %(build_type_option)s %(configopts)s" % {
             'preconfigopts': self.cfg['preconfigopts'],
             'cmd_prefix': cmd_prefix,
             'prefix_opt': prefix_opt,
             'installdir': self.installdir,
+            'build_type_option': build_type_option,
             'configopts': self.cfg['configopts'],
         }
```
boegel: To fix this long line, we should compose `cmd` differently, for example with:

```python
cmd = ' '.join([
    self.cfg['preconfigopts'],
    cmd_prefix + './configure',
    prefix_opt + self.installdir,
    build_type_option,
    self.cfg['configopts'],
])

File: setup.py [hunk 3]
```diff
@@ -83,6 +83,10 @@ def read(fname):
     packages = ["easybuild", "easybuild.easyblocks", "easybuild.easyblocks.generic"],
     package_dir = {"easybuild.easyblocks": "easybuild/easyblocks"},
     package_data = {'easybuild.easyblocks': ["[a-z0-9]/*.py"]},
+    scripts=[
+        # updated config.guess that is leveraged by the configuremake easyblock
+        'easybuild/scripts/config.guess',
```
boegel: Maybe we should rename this script, or put it in a subdirectory, to clarify what it is intended for?

easybuild/easyblocks/generic/configuremake.py
```diff
@@ -83,11 +85,23 @@ def configure_step(self, cmd_prefix=''):
         if prefix_opt is None:
             prefix_opt = '--prefix='
 
-        cmd = "%(preconfigopts)s %(cmd_prefix)s./configure %(prefix_opt)s%(installdir)s %(configopts)s" % {
+        # Avoid using config.guess from the package as it is frequently out of date, use the version shipped with EB
+        build_type = self.cfg.get('build_type')
+        if build_type is None:
+            build_type, exit_code = run_cmd('config.guess')
+            if exit_code != 0:
+                raise EasyBuildError("config.guess shipped with EB could not determine build type: '%s'",
```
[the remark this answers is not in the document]
boegel: @ocaisa No need for this, `run_cmd` check the exit code of the command that was run already by default, and will produce a clear error message when `!= 0`.

So, you can use:

```python
build_type, _ = run_cmd('config.guess')
```

(`_` is often used to signal you're ignoring part of the return value)

easybuild/easyblocks/generic/configuremake.py
ocaisa: I don't know enough about Autotools to answer this... @geimer ?

I would expect that it does no harm, `config.guess` is intended to be updated and so backwards compatible, it's output is the default setting for `--build` (see https://www.gnu.org/software/autoconf/manual/autoconf-2.61/html_node/Specifying-Names.html).

easybuild/easyblocks/generic/configuremake.py
```diff
@@ -83,11 +85,23 @@ def configure_step(self, cmd_prefix=''):
...
+                                     build_type)
+            build_type = build_type.strip()
+        build_type_option = '--build=' + build_type
+
+        cmd = ("%(preconfigopts)s %(cmd_prefix)s./configure %(prefix_opt)s%(installdir)s %(build_type_option)s "
```
boegel: @ocaisa I like the `cmd = ' '.join([...])` option better, there's little point in using a dict here to format the command...

ocaisa:
Tested locally and works as expected on this system:
```
== 2018-09-13 11:18:59,528 run.py:192 INFO running cmd: which config.guess 
== 2018-09-13 11:18:59,536 run.py:192 INFO running cmd: config.guess 
== 2018-09-13 11:18:59,599 run.py:506 INFO cmd "config.guess" exited with exit code 0 and output:
x86_64-pc-linux-gnu

== 2018-09-13 11:18:59,600 configuremake.py:95 INFO /home/alanc/EasyBuild_Git/easybuild-easyblocks/easybuild/scripts/configuremake/config.guess returned a build type x86_64-pc-linux-gnu
```

micaeljtoliveira:
I'm testing this by rebuilding a few toolchains, both on powerpc and x86. 

Found the first problem: zlib uses the ConfigureMake easyblock, but its build system is not autotools based. I guess the `--build` option should only be passed to packages that have the autotools as a dependency.

ocaisa:
Thanks @micaeljtoliveira , I made a fix for that but it also scares me a little. Since this easyblock is inherited all over the place it  could easily have an unintended knock-on effect. I wonder is the better option to (also) look in the unpacked sources directory for an existing `config.guess` and add an additional guard for that?

geimer:
Correctly generated tarballs of Autotools-based projects do **not** need Autotools as a (build)dependency.  Everything should be included in the tarball to run out-of-the-box (unless you need to regenerate the build system due to some change, or require an updated `config.guess` ;-)).  There are of course broken tarballs out there that require the user to run Autotools before `configure && make && make install`...

W.r.t. looking into the unpacked sources: You know the usual "build systems" of scientific software, right?  What if a package uses a homegrown `configure` script (like zlib), but the tarball also includes all third-party dependencies of which one or more use Autotools?

ocaisa:
@geimer Then the only fix I see for this is the same as the one used by Slack: go in and update every instance of `config.guess` you find

micaeljtoliveira:
A simpler option could be to inspect the configure script itself. All autotools generated configure scripts should include a line like this one:

> Generated by GNU Autoconf 2.64.

I think it is highly unlikely that some package uses autoconf to generate the script and then removes or changes the logic of the `--build` option. If this happens, then there should be a special place in hell for that package developers...

ocaisa:
@boegel Thoughts? Both are viable options (check the configure script, or just hard replace every instance)

ocaisa:
I've gone with the solution from @micaeljtoliveira for the moment since it was quite simple and is likely to work almost everywhere

ocaisa:
I would've gone with the other solution but I don't like the idea that we modify sources that have been checksummed (except via a patch). I see a scenario (like third party deps being included) where this will fall over but I think I am ok with catching just 99% of cases.

easybuild/easyblocks/generic/configuremake.py
```diff
@@ -83,13 +85,27 @@ def configure_step(self, cmd_prefix=''):
...
-            'cmd_prefix': cmd_prefix,
-            'prefix_opt': prefix_opt,
-            'installdir': self.installdir,
-            'configopts': self.cfg['configopts'],
-        }
+        configure_command = cmd_prefix + './configure'
+        # Avoid using config.guess from an Autoconf generated package as it is frequently out of date, use the version
+        # shipped with EB instead and provide the result to the configure command
+        build_type_option = ''
+        if 'Generated by GNU Autoconf' in open(configure_command).read():
```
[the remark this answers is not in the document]
ocaisa: Hmm, in that case is it safe to assume that with `autotools` will insert a recent config.guess? Otherwise this looks like another reason to take the alternative route.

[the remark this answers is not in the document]
geimer: It will use the one provided with the Autotools package (I believe it is the one shipped with Automake).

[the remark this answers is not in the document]
ocaisa: Hmm, seems ok then. In the worst case, we need to patch `Automake` with the latest `config.guess` when adding it to a new toolchain (if they are out of sync).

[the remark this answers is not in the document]
micaeljtoliveira: Yes, @geimer is correct, it will use the one provided with the autotools. I'm quite certain it is not the one from Autoconf, so it must be the one from Automake.

easybuild/easyblocks/generic/configuremake.py
```diff
@@ -52,6 +54,8 @@ def extra_options(extra_vars=None):
             'configure_cmd_prefix': ['', "Prefix to be glued before ./configure", CUSTOM],
             'prefix_opt': [None, "Prefix command line option for configure script ('--prefix=' if None)", CUSTOM],
             'tar_config_opts': [False, "Override tar settings as determined by configure.", CUSTOM],
+            'build_type': [None, "Type of system package is being configured for, e.g., x86_64-pc-linux-gnu (determined"
+                                 " by config.guess shipped with EasyBuild if None)", CUSTOM],
```
boegel: @ocaisa Please start the `(determined` on this line

easybuild/easyblocks/generic/configuremake.py
```diff
@@ -83,13 +87,30 @@ def configure_step(self, cmd_prefix=''):
...
-            'preconfigopts': self.cfg['preconfigopts'],
-            'cmd_prefix': cmd_prefix,
-            'prefix_opt': prefix_opt,
-            'installdir': self.installdir,
-            'configopts': self.cfg['configopts'],
-        }
+        configure_command = cmd_prefix + './configure'
+
+        # Avoid using config.guess from an Autoconf generated package as it is frequently out of date, use the version
+        # shipped with EB instead and provide the result to the configure command
```
boegel: wrap 2nd part of sentence on 2nd line:

```python
# use the version shipped ...

File: easybuild/easyblocks/generic/configuremake.py [hunk 9]
```diff
@@ -83,13 +87,30 @@ def configure_step(self, cmd_prefix=''):
...
+        build_type_option = ''
+        # possible that the configure_command is generated using preconfigopts...we're at the mercy of the gods then
+        if os.path.isfile(configure_command):
+            if 'Generated by GNU Autoconf' in open(configure_command).read():
```
boegel: use `read_file` please

easybuild/easyblocks/generic/configuremake.py
```diff
@@ -83,13 +87,30 @@ def configure_step(self, cmd_prefix=''):
...
+                build_type = self.cfg.get('build_type')
+                if build_type is None:
+                    config_guess_path, _ = run_cmd('which config.guess')
+                    config_guess_path = config_guess_path.strip()
```
boegel: there's a dedicated `which` function in `filetools`, please use that (on some stripped down systems `which` is not installed)

```python
config_guess_path = which('config.guess')
```

(no need to `.strip()` either then)

easybuild/easyblocks/generic/configuremake.py
```diff
@@ -83,13 +87,30 @@ def configure_step(self, cmd_prefix=''):
...
+                    build_type, _ = run_cmd('config.guess', log_all=True)
+                    build_type = build_type.strip()
+                    self.log.info("%s returned a build type %s" % (config_guess_path, build_type))
```
boegel:
```python
self.log.info("%s returned a build type %s", config_guess_path, build_type)

File: easybuild/scripts/configuremake/README.md [hunk 12]
```diff
@@ -0,0 +1,8 @@
+The script here is intended to be used by the configuremake generic easyblock. It is an update for `config.guess` used
```
boegel: `ConfigureMake generic easyblock`

easybuild/scripts/configuremake/README.md
```diff
@@ -0,0 +1,8 @@
...
+by `Autotools` which is frequently not up to date in package releases.
+
+You can obtain the newest version of `config.guess` from the 'config' project at
+http://savannah.gnu.org/ . The command to fetch it is
+```
+$ wget -O config.guess 'http://git.savannah.gnu.org/gitweb/?p=config.git;a=blob_plain;f=config.guess;hb=HEAD'
+```
```
boegel: Add a datestamp of last update, and include corresponding SHA256 checksum?

setup.py
```diff
@@ -83,6 +83,10 @@ def read(fname):
     packages = ["easybuild", "easybuild.easyblocks", "easybuild.easyblocks.generic"],
     package_dir = {"easybuild.easyblocks": "easybuild/easyblocks"},
     package_data = {'easybuild.easyblocks': ["[a-z0-9]/*.py"]},
+    scripts=[
+        # updated config.guess that is leveraged by the configuremake easyblock
```
boegel: `ConfigureMake`

boegel:
@ocaisa This looks reasonable after visual inspection (other than the minor suggestions I made).

I share your nervousness about changing the default behavior of this commonly used easyblock though (535 different software packages use it directly, and several other easyblocks are derived from `ConfigureMake`), so I would like to see this tested by several people and on platforms other than `x86_64` once we're done tweaking it, just to weed out any serious problems we're not anticipating.

The regression test that's coming up for EasyBuild v3.7.0 is a good time to battle-test this (on `x86_64`), and if any serious problems pop up we can always roll this back before the actual release...

micaeljtoliveira:
@boegel @ocaisa I've rebuilt the foss-2017b and foss-2018a/b toochains on a x86 using this branch and found no further issues. I will now proceed to more extensive tests on a powerpc. My plan is to test as many packages from the foss toolchains as possible.

easybuild/easyblocks/generic/configuremake.py
```diff
@@ -83,13 +88,35 @@ def configure_step(self, cmd_prefix=''):
...
+        build_type_option = ''
+        # possible that the configure_command is generated using preconfigopts...we're at the mercy of the gods then
+        if os.path.isfile(configure_command):
+            if 'Generated by GNU Autoconf' in read_file(configure_command):
+                build_type = self.cfg.get('build_type')
+
+                if build_type is None:
+                    config_guess_path = which('config.guess')
+                    if config_guess_path is None:
+                        self.log.info("No config.guess in $PATH, not setting --build")
```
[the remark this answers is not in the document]
ocaisa: Fatal is too much I think since it is (in most cases) not really required. The whole problem being solved here only really appears for (at present) ARM/Power9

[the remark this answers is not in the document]
ocaisa: I'll upgrade it to a `print_warning`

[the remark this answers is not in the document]
boegel: I understand that, but if `config.guess` can't be found something is wrong with how the easyblocks were installed, and people should know about it?

For people that simply `git clone` the repo and use the easyblocks from there, they should be aware that we're shipping a script too now that should be added to `$PATH`?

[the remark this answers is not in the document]
ocaisa: I've made a more explicit `print_warning` there now

boegel:
I tested this on CentOS 7 / x86_64 with a bunch of pretty much randomly picked easyconfigs (both old & new) that use `ConfigureMake`, i.e.:

* `parallel-20180422-intel-2018a.eb`
* `PROJ-5.0.0-intel-2018a.eb`
* `worker-1.6.8-intel-2018a.eb`
* `libxc-3.0.1-foss-2018a.eb`
* `LZO-2.10-foss-2018a.eb`
* `FFmpeg-4.0.1-intel-2018a.eb`
* `Stacks-2.2-foss-2018a.eb`
* `UDUNITS-2.2.26-foss-2018a.eb`
* `pkg-config-0.29.2-GCCcore-7.3.0.eb`
* `Bison-3.0.4-GCCcore-7.3.0.eb`
* `PhyML-20120412-goolf-1.4.10.eb`
* `libmatheval-1.1.8-goolf-1.4.10.eb`
* `Lua-5.1.4-5-goolf-1.4.10.eb`

Works like a charm (I see `--build=x86_64-pc-linux-gnu` being added to every `./configure` command), and the warning for the missing `config.guess` (which I triggered by manually removing the `config.guess`) works like a charm too (and is pretty clear).

So, I think this is good to go as is. 
If any problems pop up, we can follow up in other PRs.

Any famous last words before we merge this in @ocaisa?

boegel:
Apparently the `config.guess` script is GPLv3 licensed, with an exception for software that includes " a configuration script generated by Autoconf" (which is not the case for EasyBuild).

GPLv3 is not compatible with GPLv2, so strictly speaking we're not allowed to ship a copy of `config.guess` with EasyBuild (unless we relicense EasyBuild under GPLv3).

But I guess we could ask the project that delivers `config.guess` whether it's OK? I doubt they would strongly object against this since `config.guess` is meant to be copied regardless of software license...

ocaisa:
I've emailed the project lead to ask for permission, I guess we wait and see what they say.

geimer:
@ocaisa IANAL (I am not a lawyer), but at first glance this seems to comply to the license exception.  Maybe it's worth checking how the GPLv3 defines the term 'contains', though.  Having said that, I believe the exception was added for `configure` scripts that actually use `config.guess`, though this isn't explicitly stated.  Thus, maybe adding `AC_CANONICAL_BUILD` or similar would make sense to be on the safe side?

ocaisa:
"contains" is not defined there. I tried out adding `AC_CANONICAL_BUILD` but that just leads to more requirements, since it's not explicitly stated I'd prefer not to add more bloat.

micaeljtoliveira:
To add to the discussion about the license issue, it would seem to me that there might actually be no problem at all, as one could argue that the `config.guess` script and the rest of EasyBuild are separate programs that form an aggregate, instead of being a single program with two parts (see the sections about aggregation and plug-ins in the [GPL FAQ](https://www.gnu.org/licenses/gpl-faq.html#MereAggregation)). If that is the case, then there is no need for the licenses to be compatible.

boegel:
I'm inclined to agree, but I really don't like this part: `Where's the line between two separate programs, and one program with two parts? This is a legal question, which ultimately judges will decide.`

We really can't be sure whether this is OK or not without a clear answer from the FSF...

We briefly discussed this during the EasyBuild conf call yesterday (see notes at https://github.com/easybuilders/easybuild/wiki/Conference-call-notes-20180919), where a couple of potential alternatives were suggested:

* a) try to download an updated `config.guess` on the fly rather than packaging it
  * could be saved in source path to avoid re-downloading
  * failure to download could be ignored, either verbosely (warning message) or silently (log warning), and fallback could be to current behavior
* b) add an option in framework to specify the location of an updated `config.guess` script, and let `ConfigureMake` use it if and when it's available

I think a) could be a good middle ground... It's pretty much automagic (as long as you're online), and the updated `config.guess` can easily be "staged in" if needed (e.g. if building software is usually done offline), just like source tarballs can be. And it alleviates the licensing concerns all together...

My approach on it would be to define a constant in `ConfigureMake` that points to the download location, and it would download the `config.guess` versioned/checksummed somehow, and then symlink `config.guess` in `$PATH` automatically. This allows updating the `config.guess` in place if the needed occurs.

boegel:
@ocaisa That sounds like what I had in mind, indeed.

The location should be a `generic/ConfigureMake/` subdirectory in the 1st directory listed in the source path. You can use the `self.obtain_file` method for that in an easyblok, but it would have to be enhanced to allow specifying a particular subdirectory, since now it'll use `self.name` to determine to location(s) to search + where to download to in case it's missing...

The other option is to use `source_paths()` from `easybuild.tools.config` to determine the locations to consider for `generic/ConfigureMake/config.guess`, and to use `download_file` from `filetools` to download to the first entry returned by `source_paths()` if it's nowhere to be found yet. But then you may end up re-implementing a big chunk of the logic in `self.obtain_file`, not sure...

I'm not sure how strict we should be w.r.t. the checksum of the `config.guess` that we pick up, since we probably want to allow using an updated or tweaked `config.guess`?
So I guess thoroughly logging (location + version + SHA256 checksum of used `config.guess`) is enough?
But we should download a specific version rather than latest though, otherwise it'll get confusing fast...

In addition, we should consider only running `config.guess` on platforms other than x86_64 (see `get_cpu_architecture()` function from `easybuild.tools.systemtools`...

Running directly rather that updating `$PATH` makes sense of course.

All of this should probably be handled in a dedicated separate method that just returns the value produced by `config.guess`, that can then be passed to the `--build` option?

ocaisa:
I think the option of `source_paths()` + `download_file()` + `verify_checksum` should be easy enough to add to the `fetch_step`. I would have to add an EB version to the path, so that this gets updated when the framework does. Then I just set a value for self.config_guess and keep the existing logic.

ocaisa:
@boegel Done and seems to work.

boegel:
@ocaisa That looks a lt better, it's a good middle ground imho.

This way we avoid the licensing concerns, and we allow to easily use an alternate/updated `config.guess` for experimenting.

I did some additional updates in https://github.com/ocaisa/easybuild-easyblocks/pull/14, mainly to have better logging of the `config.guess` that is being used.

boegel:
@ocaisa Problem with the broken test is fixed with https://github.com/ocaisa/easybuild-easyblocks/pull/15

boegel:
Thanks a lot for your efforts on this @ocaisa, and to @geimer and @micaeljtoliveira for the feedback!

Going in to be included in the upcoming EasyBuild v3.7.0...

PR #1507, opened 2018-09-13
Add support for IntelMPI in TensorFlow.

akesandgren:
We need a separate wrappet for mpiicc (and probably for mpicxx if
using non-Intel compilers) because IntelMPI needs I_MPI_ROOT to be set.

This patch only adds support for IntelMPI with Intel compilers.

Also simplify PATH setting for the wrappers.

### DIFF
--- a/easybuild/easyblocks/t/tensorflow.py
+++ b/easybuild/easyblocks/t/tensorflow.py
@@ -105,6 +105,23 @@ class EB_TensorFlow(PythonPackage):
             # if with_jemalloc was specified, stick to that
             self.log.info("with_jemalloc was specified as %s, so sticking to it", self.cfg['with_jemalloc'])
 
+    def write_wrapper(self, wrapper_dir, compiler, i_mpi_root):
+        """Helper function to write a compiler wrapper."""
+        wrapper_txt = INTEL_COMPILER_WRAPPER % {
+            'compiler_path': which(compiler),
+            'intel_mpi_root': i_mpi_root,
+            'cpath': os.getenv('CPATH'),
+            'intel_license_file': os.getenv('INTEL_LICENSE_FILE', os.getenv('LM_LICENSE_FILE')),
+            'wrapper_dir': wrapper_dir,
+        }
+        wrapper = os.path.join(wrapper_dir, compiler)
+        write_file(wrapper, wrapper_txt)
+        if self.dry_run:
+            self.dry_run_msg("Wrapper for '%s' was put in place: %s", compiler, wrapper)
+        else:
+            adjust_permissions(wrapper, stat.S_IXUSR)
+            self.log.info("Using wrapper script for '%s': %s", compiler, which(compiler))
+
     def configure_step(self):
         """Custom configuration procedure for TensorFlow."""
 
@@ -124,49 +141,20 @@ class EB_TensorFlow(PythonPackage):
         use_mpi = self.toolchain.options.get('usempi', False)
 
         wrapper_dir = os.path.join(tmpdir, 'bin')
-        wrapper_cpath = os.getenv('CPATH')
-        wrapper_intel_lic = os.getenv('INTEL_LICENSE_FILE', os.getenv('LM_LICENSE_FILE'))
         use_wrapper = False
 
         if self.toolchain.comp_family() == toolchain.INTELCOMP:
             # put wrappers for Intel C/C++ compilers in place (required to make sure license server is found)
             # cfr. https://github.com/bazelbuild/bazel/issues/663
             for compiler in ('icc', 'icpc'):
-                wrapper_txt = INTEL_COMPILER_WRAPPER % {
-                    'compiler_path': which(compiler),
-                    'intel_mpi_root': 'NOT-USED-WITH-ICC',
-                    'cpath': wrapper_cpath,
-                    'intel_license_file': wrapper_intel_lic,
-                    'wrapper_dir': wrapper_dir,
-                }
-                wrapper = os.path.join(wrapper_dir, compiler)
-                write_file(wrapper, wrapper_txt)
-                if self.dry_run:
-                    self.dry_run_msg("Wrapper for '%s' was put in place: %s", compiler, wrapper)
-                else:
-                    adjust_permissions(wrapper, stat.S_IXUSR)
-                    self.log.info("Using wrapper script for '%s': %s", compiler, which(compiler))
+                self.write_wrapper(wrapper_dir, compiler, 'NOT-USED-WITH-ICC')
             use_wrapper = True
 
         if use_mpi and get_software_root('impi'):
             # put wrappers for Intel MPI compiler wrappers in place
             # (required to make sure license server and I_MPI_ROOT are found)
             for compiler in (os.getenv('MPICC'), os.getenv('MPICXX')):
-                wrapper_txt = INTEL_COMPILER_WRAPPER % {
-                    'compiler_path': which(compiler),
-                    'intel_mpi_root': os.getenv('I_MPI_ROOT'),
-                    'cpath': wrapper_cpath,
-                    'intel_license_file': wrapper_intel_lic,
-                    'wrapper_dir': wrapper_dir,
-                }
-                wrapper = os.path.join(wrapper_dir, compiler)
-                write_file(wrapper, wrapper_txt)
-
-                if self.dry_run:
-                    self.dry_run_msg("Wrapper for '%s' was put in place: %s", compiler, wrapper)
-                else:
-                    adjust_permissions(wrapper, stat.S_IXUSR)
-                    self.log.info("Using wrapper script for '%s': %s", compiler, which(compiler))
+                self.write_wrapper(wrapper_dir, compiler, os.getenv('I_MPI_ROOT'))
             use_wrapper = True
 
         if use_wrapper:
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/t/tensorflow.py
```diff
@@ -194,6 +221,8 @@ def configure_step(self):
             env.setvar(key, val)
 
         # patch configure.py (called by configure script) to avoid that Bazel abuses $HOME/.cache/bazel
+        # Should perhaps set --install_base  and --output_user_root too
+        # bazel still uses $HOME/.cache/bazel for some things.
```
boegel: @akesandgren open an issue for this, no point in keeping track of TODOs in comments (especially without mentioning `TODO` or `FIXME` ;-))

easybuild/easyblocks/t/tensorflow.py
```diff
@@ -112,25 +113,47 @@ def configure_step(self):
...
+                'wrapper_dir': icc_wrapper_dir,
             }
-            icc_wrapper = os.path.join(wrapper_dir, 'icc')
+            icc_wrapper = os.path.join(icc_wrapper_dir, 'icc')
             write_file(icc_wrapper, icc_wrapper_txt)
-            env.setvar('PATH', os.pathsep.join([os.path.dirname(icc_wrapper), os.getenv('PATH')]))
+            env.setvar('PATH', os.pathsep.join([icc_wrapper_dir, os.getenv('PATH')]))
+
+            if use_mpi:
+                mpiicc_wrapper_dir = os.path.join(tmpdir, 'bin2')
```
boegel: @akesandgren Why create a separate wrapper dir?

I don't see the point of that, the `icc` and `mpiicc` wrappers can live in the same place?
Added benefit is that you only need to update `$PATH` once.

easybuild/easyblocks/t/tensorflow.py
```diff
@@ -112,25 +113,47 @@ def configure_step(self):
...
+
         # put wrapper for Intel C compiler in place (required to make sure license server is found)
         # cfr. https://github.com/bazelbuild/bazel/issues/663
         if self.toolchain.comp_family() == toolchain.INTELCOMP:
-            wrapper_dir = os.path.join(tmpdir, 'bin')
+            icc_wrapper_dir = os.path.join(tmpdir, 'bin')
 
             icc_wrapper_txt = INTEL_COMPILER_WRAPPER % {
                 'compiler_path': which('icc'),
+                'intel_mpi_root': '',
```
boegel: @akesandgren I would use a more meaningful value here, like `UNKNOWN` or `IRRELEVANT`, in case someone inspects the generated wrapper?

easybuild/easyblocks/t/tensorflow.py
```diff
@@ -53,6 +53,7 @@
 
 export CPATH='%(cpath)s'
 export INTEL_LICENSE_FILE='%(intel_license_file)s'
+export I_MPI_ROOT='%(intel_mpi_root)s'
```
boegel: maybe also include a comment line above:

```python
# only relevant for MPI compiler wrapper (mpiicc), not for regular compiler (icc)
export I_MPI_ROOT='%(intel_mpi_root)s'

File: easybuild/easyblocks/t/tensorflow.py [hunk 5]
```diff
@@ -112,25 +113,47 @@ def configure_step(self):
...
-            env.setvar('PATH', os.pathsep.join([os.path.dirname(icc_wrapper), os.getenv('PATH')]))
+            env.setvar('PATH', os.pathsep.join([icc_wrapper_dir, os.getenv('PATH')]))
+
+            if use_mpi:
+                mpiicc_wrapper_dir = os.path.join(tmpdir, 'bin2')
+                mpiicc_wrapper_txt = INTEL_COMPILER_WRAPPER % {
+                    'compiler_path': which('mpiicc'),
+                    'intel_mpi_root': os.getenv('I_MPI_ROOT'),
+                    'cpath': os.getenv('CPATH'),
+                    'intel_license_file': os.getenv('INTEL_LICENSE_FILE', os.getenv('LM_LICENSE_FILE')),
```
boegel: Since we use these in two places, we should grab them only once from the environment above:

```python
wrapper_cpath = os.getenv('CPATH')
wrapper_intel_licfile = os.getenv('INTEL_LICENSE_FILE', os.getenv('LM_LICENSE_FILE'))

File: easybuild/easyblocks/t/tensorflow.py [hunk 6]
```diff
@@ -112,25 +113,47 @@ def configure_step(self):
...
+                    'wrapper_dir': mpiicc_wrapper_dir,
+                }
+                mpiicc_wrapper = os.path.join(mpiicc_wrapper_dir, 'mpiicc')
+                write_file(mpiicc_wrapper, mpiicc_wrapper_txt)
+                env.setvar('PATH', os.pathsep.join([mpiicc_wrapper_dir, os.getenv('PATH')]))
```
boegel: this can be removed if we let the wrapper share the same dir

easybuild/easyblocks/t/tensorflow.py
```diff
@@ -142,11 +165,15 @@ def configure_step(self):
         tensorrt_root = get_software_root('TensorRT')
         nccl_root = get_software_root('NCCL')
 
-        use_mpi = self.toolchain.options.get('usempi', False)
+        mpi_home = ''
+        if use_mpi:
+            impi_root = get_software_root('impi')
+            if impi_root:
+                mpi_home = os.path.join(impi_root, 'intel64')
```
boegel: probably worth adding some debug logging here (only under `use_mpi`):

```python
self.log.debug("Derived value for MPI_HOME: %s", mpi_home)

**Comment by boegel** [comment]
@akesandgren How does the problem that you're fixing here manifest itself when building TF with `intel` toolchain?

File: easybuild/easyblocks/t/tensorflow.py (hunk 2 above)
Comment (akesandgren) [reply easybuild/easyblocks/t/tensorflow.py]: The wrapper removes its own path from the environment so mpiicc won't be able to find the icc wrapper if they live in the same wrapper dir i think. Haven't properly tested that though.
But on the other hand it should inherit the env that mpiicc is using....

Need to test this...

**Comment by akesandgren** [comment]
WIll take a look at the comments tomorrow.

File: easybuild/easyblocks/t/tensorflow.py (hunk 2 above)
Comment (boegel) [reply easybuild/easyblocks/t/tensorflow.py]: Hmm, right, I forgot about that part... Better make that clear with a comment, and use a better name than `bin2` ;)

**Comment by akesandgren** [comment]
Without the mpiicc wrapper and setting of I_MPI_ROOT I get this when linking:
/hpc2n/eb/software/Compiler/GCCcore/7.3.0/binutils/2.30/bin/ld.gold: error: cannot find -lmpifort
/hpc2n/eb/software/Compiler/GCCcore/7.3.0/binutils/2.30/bin/ld.gold: error: cannot find -lmpigi

The reason for that is, that without I_MPI_ROOT defined mpiicc passes the wrong -L args down to icc.

**Comment by akesandgren** [comment]
Think about a helper func looking like this:
```
    def write_wrapper(self, tmpdir, compiler, i_mpi_root)
        """Helper function to write a compiler wrapper."""
        wrapper_dir = os.path.join(tmpdir, 'bin')
        wrapper_txt = INTEL_COMPILER_WRAPPER % {
            'compiler_path': which(compiler),
            'intel_mpi_root': i_mpi_root,
            'cpath': os.getenv('CPATH'),
            'intel_license_file': os.getenv('INTEL_LICENSE_FILE', os.getenv('LM_LICENSE_FILE'))
            'wrapper_dir': wrapper_dir,
        }
        wrapper = os.path.join(wrapper_dir, compiler)
        write_file(wrapper, wrapper_txt)
        if self.dry_run:
            self.dry_run_msg("Wrapper for '%s' was put in place: %s", compiler, wrapper)
        else:
            adjust_permissions(wrapper, stat.S_IXUSR)
            self.log.info("Using wrapper script for '%s': %s", compiler, which(compiler))
```

Should that be called as:
```
self.write_wrapper(tmpdir, compiler, 'NOT-USED-WITH-ICC')
```
or some other way? (Still learning python :-)

And should i make the configure_step tmpdir variable into self.tmpdir instead?
and/or the other ones too?

File: easybuild/easyblocks/t/tensorflow.py [hunk 8]
```diff
@@ -142,11 +170,16 @@ def configure_step(self):
         tensorrt_root = get_software_root('TensorRT')
         nccl_root = get_software_root('NCCL')
 
-        use_mpi = self.toolchain.options.get('usempi', False)
+        mpi_home = ''
+        if use_mpi:
+            impi_root = get_software_root('impi')
```
boegel: @akesandgren Since you're using this in two places in `configure_step`, why not combine this with the exact same check above to put the `mpiicc` wrapper in place?

```python
use_mpi = self.toolchain.options.get('usempi', False)
impi_root = get_software_root('impi')
mpi_home = ''
if use_mpi and impi_root:
    # put wrappers for Intel MPI compiler wrappers in place
    ...
    # set correct value for MPI_HOME
    mpi_home = os.path.join(impi_root, 'intel64')

akesandgren: Done

boegel:
Looks good, tested with several existing `TensorFlow` easyconfigs (none of which are using both `intel` toolchain & have `usempi` enabled though...), no regressions, so good to go.

PR #1513, opened 2018-09-18
take (pre)installopts into account in RPackage generic easyblock

boegel:


boegel:
@bartoldeman It's actually out of pure need, see `hdfr5` in https://github.com/easybuilders/easybuild-easyconfigs/pull/6858... ;)

Issue #1515, opened 2018-09-19
Tensorflow easyblock should probably use --install_base  and --output_user_root


akesandgren:
Should perhaps set --install_base  and --output_user_root too, bazel still uses $HOME/.cache/bazel for some things.

Needs investigation.

akesandgren:
Ok, seems to work, but bazel still creates $HOME/.cache/bazel/_bazel_username/
Nothing is stored there (after adding the above args to bazel), but it is still annoying.

boegel:
@akesandgren Can we consider this fixed via #1519? I don't think that Bazel create the empty `$HOME/.cache/bazel/_bazel_$USER` directory is too big of a problem...

akesandgren:
Yeah, the main problem is solved. Making sure it doesn't create the dir can probably only be handled by explicitly setting $HOME before calling bazel...

boegel:
@akesandgren Well, that's worth considering...

PR #1519, opened 2018-09-20
Make sure bazel doesn't write files in $HOME/.cache when building tensorflow

akesandgren:
It still creates an empty directory, $HOME/.cache/bazel/_bazel_username,
but at least it doesn't write files in there.

This fixes https://github.com/easybuilders/easybuild-easyblocks/issues/1515


akesandgren:
That's clearly one of the quickest PR -> merge I've seen :-)

PR #1520, opened 2018-09-20
Enabling VSX on POWER for FFTW >= 3.3.7.

micaeljtoliveira:
Apparently the issue with VSX on POWER has been fixed in FFTW 3.3.7 (see https://github.com/FFTW/fftw3/commit/04590cb11baa11bbfdebe101fa90186bbf48423c). My tests indicate that this seems to be the case.


boegel:
Going in, thanks @micaeljtoliveira!

PR #1525, opened 2018-09-24
Trilinos: Add librt as dependency when linking suitesparse.

bartoldeman:
This change avoids this error, at least on CentOS 6:
SuiteSparse_config.c:function SuiteSparse_tic: error: undefined reference to 'clock_gettime'"


boegel:
re-tested with `Trilinos-12.12.1-foss-2018a-Python-3.6.4.eb`, lgtm, thanks @bartoldeman!

PR #1530, opened 2018-09-27
fix checking for downloaded dependencies for stand-alone installations in PythonPackage generic easyblock

boegel:
I noticed that downloaded dependencies for stand-alone installations using `PythonPackage` (i.e. non-extensions) were not being detected even though `download_dep_fail = True` is set in the easyconfig file.

The `return` value of `sanity_check_step` is only taken into account for Python packages installed as extensions; for stand-alone installations, `self.sanity_check_fail_msgs` is checked in the `sanity_check_step` implementation in `EasyBlock`.

So, to fix the issue I had to:
* move down the `super` call, and check for downloaded dependencies *first*
* append the error message to `self.sanity_check_fail_msgs` (and then let the `super` call take care of actually reporting the error, after performing the rest of the sanity check)
* combine the result of the `super` call with the "local findings" (which is only relevant for extensions)

### DIFF
--- a/easybuild/easyblocks/generic/pythonpackage.py
+++ b/easybuild/easyblocks/generic/pythonpackage.py
@@ -535,7 +535,7 @@ class PythonPackage(ExtensionEasyBlock):
             exts_filter = (orig_exts_filter[0].replace('python', self.python_cmd), orig_exts_filter[1])
             kwargs.update({'exts_filter': exts_filter})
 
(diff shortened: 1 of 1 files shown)


boegel:
This looks/sounds worse than it actually is: since all easyconfigs that enable `download_dep_fail` also enable `use_pip`, there are actually *no* "casualties" here, i.e. none of the central easyconfigs will be broken because of this fix...

The reason is that we run we `pip` with `--no-deps` (unless `use_pip_for_deps = True` is used, which we never do in the central easyconfigs).

PR #1532, opened 2018-09-27
also specify --host option to configure script based on config.guess result in ConfigureMake easyblock

boegel:
see discussion in https://github.com/easybuilders/easybuild-easyconfigs/issues/6922

cc @ocaisa, @vanzod

### DIFF
--- a/easybuild/easyblocks/generic/configuremake.py
+++ b/easybuild/easyblocks/generic/configuremake.py
@@ -72,9 +72,12 @@ class ConfigureMake(EasyBlock):
             'configure_cmd_prefix': ['', "Prefix to be glued before ./configure", CUSTOM],
             'prefix_opt': [None, "Prefix command line option for configure script ('--prefix=' if None)", CUSTOM],
             'tar_config_opts': [False, "Override tar settings as determined by configure.", CUSTOM],
-            'build_type': [None, "Type of system package is being configured for, e.g., x86_64-pc-linux-gnu "
-                                 "(determined by config.guess shipped with EasyBuild if None, "
-                                 "False implies to leave it up to the configure script)", CUSTOM],
+            'build_type': [None, "Value to provide to --build option of configure script, e.g., x86_64-pc-linux-gnu "
+                                 "(determined by config.guess shipped with EasyBuild if None,"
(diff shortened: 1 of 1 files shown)


ocaisa:
All this does is enforce the default behaviour, I see no problem.

boegel:
@micaeljtoliveira Given your feedback in https://github.com/easybuilders/easybuild-easyconfigs/issues/6922, thoughts on this change, do you agree it makes sense to set both `--build` & `--host`, regardless of the buggy way in which the `Hypre` `configure` script is "written"?

micaeljtoliveira:
@boegel  I'm no expert in cross-compilation or this specific part of the autotools, but if their documentation is to be trusted, setting both `--build`  and `--host` should be fine. Personally I would avoid doing this systematically as a workaround for a bug in a specific package and I would patch `Hypre` instead, but this is obviously up to you.

In case you are wondering, this is how the patch should look like for `Hypre` 2.11.2 (assuming I understood the problem correctly):
```
2330a2331,2332
> if test "x$build_alias" = "x"
> then
2367a2370,2374
> 
> fi
> 
> if test "x$host_alias" = "x"
> then
2399a2407,2409
> 
> fi
> 
7230c7240
<    if test $host = $build
---
>    if test $host_alias = $build_alias
```

boegel:
@micaeljtoliveira Thanks forthe feedback, much appreciated!

This problem may not be specific to `Hypre` at all, who knows... I'm pretty sure there's software out there we haven't played with yet. ;)

Also, since there should be no harm in setting both (especially since it doesn't prevent you from actually cross-compiling thanks to the `build_type` and `host_type` easyconfig parameters now supported by `ConfigureMake`), I think this is fine.

This approach has the added benefit that we don't need to touch the `Hypre` easyconfigs (which worked fine before we started setting (only) `--build`), and I couldn't find a clear reference to people reporting the `invalid value of canonical host` error for `Hypre`, so it seems to me we're in uncharted territory when only setting `--build`, which scares me a bit...

I agree with you that it's not entirely clear what the best way forward is here. But my gut feeling (which may be wrong, time will tell) suggests that also setting `--host` in `ConfigureMake` will result in less headaches going forward.

PR #1533, opened 2018-09-28
use DOT_MODULERC constant in ModuleRC easyblock rather than hardcoding '.modulerc'

boegel:
This is required because of the changes in https://github.com/easybuilders/easybuild-framework/pull/2597, since the `.modulerc` file has to named `.modulerc.lua` in case it's generated in Lua syntax for Lmod >= 7.8


boegel:
@vanzod This is now synced with `develop` to include the fix from #1535

vanzod:
Tested and working as expected.

Issue #1534, opened 2018-09-29
Issue with ModuleRC easyblock when using HierarchicalMNS


guest1604:
Hello,

I notice an issue when installing Java-1.8.eb with HierarchicalMNS.

```
$ eb Java-1.8.eb --robot --installpath=/home/user/.local/easybuild/hmns --robot --module-naming-scheme=HierarchicalMNS --debug
== temporary log file in case of crash /tmp/eb-6SEbHt/easybuild-bVMHYn.log
== resolving dependencies ...
== processing EasyBuild easyconfig /home/user/.local/easybuild/software/EasyBuild/3.7.0/lib/python2.7/site-packages/easybuild_easyconfigs-3.7.0-py2.7.egg/easybuild/easyconfigs/j/Java/Java-1.8.0_181.eb
== building and installing Core/Java/1.8.0_181...
== fetching files...
== creating build dir, resetting environment...
== unpacking...
== patching...
== preparing...
== configuring...
== building...
== testing...
== installing...
== taking care of extensions...
== postprocessing...
== sanity checking...
== cleaning up...
== creating module...
== permissions...
== packaging...
== COMPLETED: Installation ended successfully
== Results of the build can be found in the log file(s) /home/user/.local/easybuild/hmns/software/Core/Java/1.8.0_181/easybuild/easybuild-Java-1.8.0_181-20180929.190754.log
== processing EasyBuild easyconfig /home/user/.local/easybuild/software/EasyBuild/3.7.0/lib/python2.7/site-packages/easybuild_easyconfigs-3.7.0-py2.7.egg/easybuild/easyconfigs/j/Java/Java-1.8.eb
== building and installing Core/Java/1.8...
== fetching files...
== creating build dir, resetting environment...
== unpacking...
== patching...
== preparing...
== configuring...
== building...
== testing...
== installing...
== taking care of extensions...
== postprocessing...
== sanity checking...
== FAILED: Installation ended unsuccessfully (build directory: /home/user/.local/easybuild/build/Java/1.8/dummy-dummy): build failed (first 300 chars): Module command 'module load Java/1.8' failed with exit code 1; stderr: Lmod has detected the following error: The following module(s) are unknown: "Java/1.8"

Please check the spelling or version number. Also try "module spider ..."
It is also possible your cache file is out-of-date; it may help to 
[...]

PR #1535, opened 2018-09-29
use short module name when creating module wrapper in ModuleRC generic easyblock

boegel:


boegel:
fix confirmed by @guest1604 in #1534

Issue #1536, opened 2018-10-01
Use of --ignore-installed in TensorFlow's step_install


agijsberts:
TensorFlow's EasyBlock installs the produced wheel package via
``pip install --ignore-installed``

https://github.com/easybuilders/easybuild-easyblocks/blob/4f816688b92748d0fcd40ae553fb2dcd369c965a/easybuild/easyblocks/t/tensorflow.py#L372

As a consequence, it pulls in binary wheels of all dependencies, including for example NumPy. So loading the TensorFlow module will replace the NumPy as part of the Python package (linked against MKL in my case) with a precompiled NumPy linked against a pre-shipped OpenBLAS. Is this intended behavior?

At least on our systems, removing ``--ignore-installed`` leads to desired behavior, i.e. pip does not install those dependencies that were already satisfied via dependent modules.

boegel:
@agijsberts That's by no means intended, we overlooked this completely... Thanks a lot for reporting this!

The `--ignore-installed` was added in the original implementation of the `TensorFlow` easyblock to try and ensure the specified `*.whl` is installed, cfr. https://github.com/easybuilders/easybuild-easyblocks/pull/1287 .

We probably also need to use `--no-deps` to avoid that `pip` pulls in all dependencies, there should be no need for that, everything should be in place already (and if not, that should be fixed).

cc @akesandgren

agijsberts:
Thank you for confirming my suspicion.

Mind that the pip command does pull in tensorboard, which is obviously an important dependency. I guess you mean that tensorboard (and maybe a couple of other libraries) should be pulled in via the extensions mechanism (**exts_list**)?

boegel:
@agijsberts Indeed, we'll need to fix this in the existing `TensorFlow` easyconfigs, but that should be just a matter of including everything that is needed via `exts_list`, indeed...

boegel:
Fixed in #1537, where `--no-deps` is used if extensions are included in the `TensorFlow` easyconfig file (i.e. if `exts_list` is not empty).

See also updated easyconfigs for TensorFlow 1.10.x in https://github.com/easybuilders/easybuild-easyconfigs/pull/6940

PR #1537, opened 2018-10-02
use --no-deps when installing .whl in TensorFlow easyblock if extension are being installed, move test run to sanity check

boegel:
This fixes the problem reported by @agijsberts in #1536, by also using `--no-deps` in the `pip install` command.

This change implies that all additional Python packages required by `TensorFlow` that are not available in the `Python` installation (e.g. `numpy` is) are included in the easyconfig file as extensions, so all `TensorFlow` easyconfig files that use this easyblock will need to be fixed...

Two other changes are also required:

* running the MNIST tutorial example as a test now *must* be done in the sanity check, since not all extension are available yet after the `pip install` of the generated `.whl`; the comment claiming that this couldn't be done in the sanity check step was incorrect, since the build dir isn't cleaned up yet

* creating of `google/__init__.py` now doesn't make sense anymore since the `protobuf` Python package will not be installed automagically anymore (instead, it should be listed as a dependency in the `TensorFlow` easyconfig file)

### DIFF
--- a/easybuild/easyblocks/t/tensorflow.py
+++ b/easybuild/easyblocks/t/tensorflow.py
@@ -38,6 +38,7 @@ from distutils.version import LooseVersion
 import easybuild.tools.environment as env
 import easybuild.tools.toolchain as toolchain
 from easybuild.easyblocks.generic.pythonpackage import PythonPackage
+from easybuild.easyblocks.python import EXTS_FILTER_PYTHON_PACKAGES
 from easybuild.framework.easyconfig import CUSTOM
 from easybuild.tools.build_log import EasyBuildError
 from easybuild.tools.filetools import adjust_permissions, apply_regex_substitutions, mkdir, resolve_path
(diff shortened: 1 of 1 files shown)


boegel:
@akesandgren After some further testing (& thinking), I've made a couple of changes:

* prepare for installing Python packages as extensions

* only use `--no-deps` when the `TensorFlow` easyconfig also installs extensions, which should effectively restore backward compatibility (i.e. we can only fix the auto-downloading of Python packages for the latest `TensorFlow` version without breaking older existing easyconfigs)

* restore the `touch` on `google/__init__.py`, but only when it's really needed (i.e. when the `google/protobuf` dir exists)

Are you up for reviewing this & retesting https://github.com/easybuilders/easybuild-easyconfigs/pull/6940 (which I also changed to use `protobuf` 3.6.0, based on the feedback by @agijsberts)?

PR #1539, opened 2018-10-04
add symlink to wrapped module file when creating .modulerc in temporary location

boegel:
fix for problem reported by @zao in https://github.com/easybuilders/easybuild/issues/462 and by @verdurin in https://github.com/easybuilders/easybuild-easyconfigs/pull/6906

requires https://github.com/easybuilders/easybuild-framework/pull/2611

### DIFF
--- a/easybuild/easyblocks/generic/modulerc.py
+++ b/easybuild/easyblocks/generic/modulerc.py
@@ -30,10 +30,8 @@ EasyBuild support for installing a software-specific .modulerc file
 import os
 
 from easybuild.framework.easyblock import EasyBlock
-from easybuild.tools.build_log import EasyBuildError
-from easybuild.tools.build_log import print_msg
-from easybuild.tools.config import build_option
-from easybuild.tools.filetools import write_file
+from easybuild.tools.build_log import EasyBuildError, print_msg
+from easybuild.tools.filetools import mkdir, symlink
 
 
 class ModuleRC(EasyBlock):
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/generic/modulerc.py
```diff
@@ -58,9 +56,6 @@ def make_module_step(self, fake=False):
         modfile_path = self.module_generator.get_module_filepath(fake=fake)
         modulerc = os.path.join(os.path.dirname(modfile_path), self.module_generator.DOT_MODULERC)
 
-        if os.path.exists(modulerc) and not build_option('force'):
-            raise EasyBuildError("Found existing file at %s, not overwriting without --force", modulerc)
```
boegel: This check is now being done in the `ModuleGenerator.modulerc` method (see https://github.com/easybuilders/easybuild-framework/pull/2611), so no point in doing it again here.

easybuild/easyblocks/generic/modulerc.py
```diff
@@ -77,9 +72,21 @@ def make_module_step(self, fake=False):
         alias_modname = deps[0]['short_mod_name']
         self.log.info("Adding module version alias for %s to %s", alias_modname, modulerc)
 
-        module_version_specs = {'modname': alias_modname, 'sym_version': self.version, 'version': deps[0]['version']}
-        modulerc_txt = self.module_generator.modulerc(module_version=module_version_specs)
-        write_file(modulerc, modulerc_txt, backup=True)
+        # add symlink to wrapped module file when generating .modulerc in temporary directory (done during sanity check)
+        # this is strictly required for Lmod 6.x, for which .modulerc and wrapped module file must be in same location
+        if fake:
```
damianam: Doesn't it make sense to add this symlink after the file has been created, instead of creating a broken symlink that is "fixed" afterwards?

boegel: This is not a broken symlink at all. It links to the `Java/1.8.0_181.lua` module file (for example), which must already exist at this point (since it's a dependency for `Java/1.8`)

PR #1545, opened 2018-10-12
Fixed incorrect LibInt reference, and added support for CP2K 6.1

damianam:


boegel:
@damianam Do you have an easyconfig PR to go with this? Not strictly needed, but it's good to be able to verify the changes for CP2K 6.1...

damianam:
I do have it, but I am testing it myself right now (one of the dependencies is failing right now.....). However, the file is for our stack, which doesn't match upstream.

akesandgren:
@damianam how did your testing go?

easybuild/easyblocks/c/cp2k.py
```diff
@@ -394,32 +400,40 @@ def configure_common(self):
...
-            self.log.warning("LibInt module not loaded, so building without LibInt support")
+            # throw a warning, since CP2K without Libint doesn't make much sense
+            self.log.warning("Libint module not loaded, so building without Libint support")
 
 
         libxc = get_software_root('libxc')
         if libxc:
             cur_libxc_version = get_software_version('libxc')
+            if LooseVersion(self.version) >= LooseVersion('6.1'):
+                LIBXC_MIN_VERSION = '4.0.3'
```
boegel: @damianam This is basically changing what is supposed to be a constant, which is kind of weird.

Since the minimal `libxc` version depends on the CP2K version now, we should simply drop the constant, and go with something like:

```python
# CP2K requires a particular version of libxc
if LooseVersion(self.version) >= LooseVersion('6.1'):
    libxc_min_version = '4.0.3'
else:
    libxc_min_version = '2.0.1'

if LooseVersion(cur_libxc_version) < LooseVersion(libxc_min_version):
    ...
```

and then drop the `LIBXC_MIN_VERSION` definition at the top

easybuild/easyblocks/c/cp2k.py
```diff
@@ -394,32 +400,40 @@ def configure_common(self):
...
+            else:
+                dflags = ' -D__LIBXC2'
+
             if LooseVersion(cur_libxc_version) < LooseVersion(LIBXC_MIN_VERSION):
-                raise EasyBuildError("CP2K only works with libxc v%s (or later)", LIBXC_MIN_VERSION)
+                raise EasyBuildError("This version of CP2K is not compatible with libxc < %s" % LIBXC_MIN_VERSION)
 
-            options['DFLAGS'] += ' -D__LIBXC2'
-            if LooseVersion(cur_libxc_version) >= LooseVersion('2.2'):
+            options['DFLAGS'] += dflags
```
boegel: @damianam This line should be moved up, above the version check for `libxc` (to keep things together better)

boegel:
@damianam Ping on tackling the remarks?

surak:
ping

easybuild/easyblocks/c/cp2k.py
```diff
@@ -394,32 +397,40 @@ def configure_common(self):
...
             options['LIBS'] += ' %s -lstdc++ %s' % (libint_libs, libint_wrapper)
 
         else:
-            # throw a warning, since CP2K without LibInt doesn't make much sense
-            self.log.warning("LibInt module not loaded, so building without LibInt support")
+            # throw a warning, since CP2K without Libint doesn't make much sense
+            self.log.warning("Libint module not loaded, so building without Libint support")
 
 
         libxc = get_software_root('libxc')
```
boegel: @damianam please include https://www.cp2k.org/howto:compile#k_libxc_optional_wider_choice_of_xc_functionals as a comment here, to explain what we do here

boegel:
lgtm, tested, going in, thanks @damianam!

PR #1546, opened 2018-10-15
add support for impi version 2019

damianam:


boegel:
@damianam It looks like most of this is just prefixing `intel64`?

We should handle that more cleanly, it's too much copy-pasting right now imho, which will get messy if we would need to make further tweaks for future versions...

easybuild/easyblocks/i/impi.py
```diff
@@ -186,18 +196,22 @@ def make_module_req_guess(self):
...
-                'MIC_LD_LIBRARY_PATH' : ['mic/lib'],
+                'MIC_LD_LIBRARY_PATH': ['mic/lib'],
             }
         else:
-            lib_dirs = ['lib/em64t', 'lib64']
-            include_dirs = ['include64']
+            # Keep release_mt and release in front, to give priority to the possible symlinks in intel64/lib.
+            # IntelMPI 2019 changed the default library to be the non-mt version.
+            lib_dirs = ['intel64/%s' % x for x in ['lib/release_mt', 'lib/release', 'lib', 'libfabric/lib']] +
+                    ['lib/em64t', 'lib64']
```
boegel: @damianam As discussed, please clearly discriminate between the paths that should be consider for `impi` 2019.x and the older versions.

boegel:
Looking good & tested, so going in, thanks @damianam!

PR #1547, opened 2018-10-15
Also install cp2k as a library

ocaisa:


ocaisa:
@hajgato Maybe you want to take a look at this since your the last one to make a change here? Turns out `make all` doesn't build `libcp2k.a` so I had to trigger it in addition. Currently I copy over all the libraries I find but probably only `libcp2k.a` is really necessary

hajgato:
@ocaisa We have never used libcp2k. Its probably only needed if you want to call `CP2K` from a C/C++/Fortran code. Anyway, it does not hurt to have it.

damianam:
I had the same PR about to be made :-P

easybuild/easyblocks/c/cp2k.py
```diff
@@ -803,6 +806,20 @@ def install_step(self):
         except OSError, err:
             raise EasyBuildError("Copying executables from %s to bin dir %s failed: %s", exedir, targetdir, err)
 
+        # copy libraries
+        if self.cfg['library']:
+            targetdir = os.path.join(self.installdir, 'lib')
+            libdir = os.path.join(self.cfg['start_dir'], 'lib/%s' % self.typearch, self.cfg['type'])
```
boegel: should use `... , 'lib', self.typearch, ...`

easybuild/easyblocks/c/cp2k.py
```diff
@@ -803,6 +806,20 @@ def install_step(self):
...
+            try:
+                if not os.path.exists(targetdir):
+                    os.makedirs(targetdir)
+                os.chdir(libdir)
+                for library in os.listdir(libdir):
+                    if os.path.isfile(library):
+                        shutil.copy2(library, targetdir)
```
boegel: @ocaisa Please use `change_dir` and `copy_file` from `filetools` module here.
Also no need to pre-create the `targetdir`, `copy_file` already does that.

```python
try:
    change_dir(libdir)
    for lib in os.listdir(libdir):
        if os.path.isfile(lib):
            copy_file(lib, os.path.join(targetdir, lib))

except OSError as err:
    raise EasyBuildError("...")

File: easybuild/easyblocks/c/cp2k.py [hunk 3]
```diff
@@ -97,6 +97,7 @@ def extra_options():
             'extradflags': ['', "Extra DFLAGS to be added", CUSTOM],
             'ignore_regtest_fails': [False, ("Ignore failures in regression test "
                                              "(should be used with care)"), CUSTOM],
+            'library': [True, "Also build CP2K as a library", CUSTOM],
```
boegel: @ocaisa By enabling this by default, this has to work on all existing `CP2K` easyconfigs, which includes some very old ones...

Maybe `False` is a better default, and people can just build it when they need it? We can start enabling it for recent `CP2K` easyconfigs, of course...

easybuild/easyblocks/c/cp2k.py
```diff
@@ -792,25 +794,29 @@ def install_step(self):
...
             raise EasyBuildError("Copying executables from %s to bin dir %s failed: %s", exedir, targetdir, err)
 
+        # copy libraries
+        if self.cfg['library']:
+            targetdir = os.path.join(self.installdir, 'lib')
+            libdir = os.path.join(self.cfg['start_dir'], 'lib', self.typearch, self.cfg['type'])
+            try:
+                copy_dir(libdir, targetdir)
+            except OSError as err:
+                raise EasyBuildError("Copying libraries from %s to lib dir %s failed: %s", libdir, targetdir, err)
```
boegel: @ocaisa No need for the `try`/`except` anymore here, that's already handled by `copy_dir`; same below

ocaisa:
Successful test report available at https://gist.github.com/ocaisa/fde481bbee42937b0a56a992600fbfb1

PR #1553, opened 2018-10-19
add generic easyblock for installing a bundle of Python packages

boegel:
I got sick & tired of adding the same boilerplate stuff to easyconfigs that are just a bundle of Python packages, so let's add a generic easyblock specifically for that purpose...


bartoldeman:
lgtm going in @boegel !

PR #1554, opened 2018-10-19
Also populate the include dir for CP2K

ocaisa:
No includes were in #1547 , this fixes that problem for both C and fortran users.

At the moment this is quite brutish, I've opened an [issue with CP2K](https://github.com/cp2k/cp2k/issues/53) to see if they can help us trim things back to what is strictly necessary.

### DIFF
--- a/easybuild/easyblocks/c/cp2k.py
+++ b/easybuild/easyblocks/c/cp2k.py
@@ -805,7 +805,7 @@ class EB_CP2K(EasyBlock):
             # Also need to populate the include directory
             targetdir = os.path.join(self.installdir, 'include')
             libcp2k_header = os.path.join(self.cfg['start_dir'], 'src', 'start', 'libcp2k.h')
-            target_header = os.path.join(targetdir,os.path.basename(libcp2k_header))
+            target_header = os.path.join(targetdir, os.path.basename(libcp2k_header))
             copy_file(libcp2k_header, target_header)
             # include all .mod files for fortran users (don't know the exact list so take everything)
             mod_path = os.path.join(self.cfg['start_dir'], 'obj', self.typearch, self.cfg['type'])
--- a/easybuild/easyblocks/c/cp2k.py
+++ b/easybuild/easyblocks/c/cp2k.py
@@ -50,7 +50,7 @@ from easybuild.framework.easyblock import EasyBlock
 from easybuild.framework.easyconfig import CUSTOM
 from easybuild.tools.build_log import EasyBuildError
 from easybuild.tools.environment import setvar
-from easybuild.tools.filetools import copy_dir, write_file
+from easybuild.tools.filetools import copy_dir, copy_file, mkdir, search_file, write_file
 from easybuild.tools.config import build_option
 from easybuild.tools.modules import get_software_root, get_software_version
 from easybuild.tools.run import run_cmd
@@ -797,11 +797,21 @@ class EB_CP2K(EasyBlock):
         targetdir = os.path.join(self.installdir, 'bin')
         copy_dir(exedir, targetdir)
 
-        # copy libraries
+        # copy libraries and include files, not sure what is strictly required so we take everything
         if self.cfg['library']:
             libdir = os.path.join(self.cfg['start_dir'], 'lib', self.typearch, self.cfg['type'])
             targetdir = os.path.join(self.installdir, 'lib')
             copy_dir(libdir, targetdir)
+            # Also need to populate the include directory
+            targetdir = os.path.join(self.installdir, 'include')
+            mkdir(targetdir)  # Needs to exist for copy below to work correctly
+            include_file = os.path.join(self.cfg['start_dir'], 'src', 'start', 'libcp2k.h')
+            copy_file(include_file, targetdir)
+            # include all .mod files for fortran users
+            mod_path = os.path.join(self.cfg['start_dir'], 'obj', self.typearch, self.cfg['type'])
+            _, mod_files = search_file([mod_path], '.*\.mod', silent=True)
+            for mod_file in mod_files:
+                copy_file(mod_file, targetdir)
 
         # copy data dir
         datadir = os.path.join(self.cfg['start_dir'], 'data')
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/c/cp2k.py
```diff
@@ -797,11 +797,21 @@ def install_step(self):
...
+        # copy libraries and include files, not sure what is strictly required so we take everything
         if self.cfg['library']:
             libdir = os.path.join(self.cfg['start_dir'], 'lib', self.typearch, self.cfg['type'])
             targetdir = os.path.join(self.installdir, 'lib')
             copy_dir(libdir, targetdir)
+            # Also need to populate the include directory
+            targetdir = os.path.join(self.installdir, 'include')
+            mkdir(targetdir)  # Needs to exist for copy below to work correctly
+            include_file = os.path.join(self.cfg['start_dir'], 'src', 'start', 'libcp2k.h')
+            copy_file(include_file, targetdir)
```
boegel: @ocaisa `copy_file` will make sure the target dir exists, so you can shorten this to:

```python
target_inc_dir = os.path.join(self.installdir, 'include')
libcp2k_header = os.path.join(self.cfg['start_dir'], 'src', 'start', 'libcp2k.h')
copy_file(libcp2k_header, target_inc_dir)

File: easybuild/easyblocks/c/cp2k.py [hunk 2]
```diff
@@ -797,11 +797,21 @@ def install_step(self):
...
+            # include all .mod files for fortran users
+            mod_path = os.path.join(self.cfg['start_dir'], 'obj', self.typearch, self.cfg['type'])
+            _, mod_files = search_file([mod_path], '.*\.mod', silent=True)
```
boegel: hmm, `glob` probably makes more sense here?

```python
for mod_file in glob.glob(os.path.join(mod_path, '*.mod')):
    copy_file(mod_file, target_inc_dir)

File: easybuild/easyblocks/c/cp2k.py (hunk 1 above)
Comment (ocaisa) [reply easybuild/easyblocks/c/cp2k.py]: Yeah, pretty sure I am right here. `copy_file` creates the parent directory of the target but it is working under the assumption that the target is a file. I'm abusing it here since I know it uses `copy2` under the hood and `copy2` allows me to use a directory as a target but I need to make sure that directory exists first (otherwise it copies it as a file with the directory name).

Comment (ocaisa) [reply easybuild/easyblocks/c/cp2k.py]: Yeah, pretty sure I am right here. `copy_file` creates the parent directory of the target but it is working under the assumption that the target is a file. I'm abusing it here since I know it uses `copy2` under the hood and `copy2` allows me to use a directory as a target but I need to make sure that directory exists first (otherwise it copies it as a file with the directory name).

Comment (ocaisa) [reply easybuild/easyblocks/c/cp2k.py]: Yeah, pretty sure I am right here. `copy_file` creates the parent directory of the target but it is working under the assumption that the target is a file. I'm abusing it here since I know it uses `copy2` under the hood and `copy2` allows me to use a directory as a target but I need to make sure that directory exists first (otherwise it copies it as a file with the directory name).

Comment (ocaisa) [reply easybuild/easyblocks/c/cp2k.py]: Yeah, pretty sure I am right here. `copy_file` creates the parent directory of the target but it is working under the assumption that the target is a file. I'm abusing it here since I know it uses `copy2` under the hood and `copy2` allows me to use a directory as a target but I need to make sure that directory exists first (otherwise it copies it as a file with the directory name).

Comment (ocaisa) [reply easybuild/easyblocks/c/cp2k.py]: Yeah, pretty sure I am right here. `copy_file` creates the parent directory of the target but it is working under the assumption that the target is a file. I'm abusing it here since I know it uses `copy2` under the hood and `copy2` allows me to use a directory as a target but I need to make sure that directory exists first (otherwise it copies it as a file with the directory name).

Comment (ocaisa) [reply easybuild/easyblocks/c/cp2k.py]: Yeah, pretty sure I am right here. `copy_file` creates the parent directory of the target but it is working under the assumption that the target is a file. I'm abusing it here since I know it uses `copy2` under the hood and `copy2` allows me to use a directory as a target but I need to make sure that directory exists first (otherwise it copies it as a file with the directory name).

Comment (ocaisa) [reply easybuild/easyblocks/c/cp2k.py]: Yeah, pretty sure I am right here. `copy_file` creates the parent directory of the target but it is working under the assumption that the target is a file. I'm abusing it here since I know it uses `copy2` under the hood and `copy2` allows me to use a directory as a target but I need to make sure that directory exists first (otherwise it copies it as a file with the directory name).

Comment (ocaisa) [reply easybuild/easyblocks/c/cp2k.py]: Yeah, pretty sure I am right here. `copy_file` creates the parent directory of the target but it is working under the assumption that the target is a file. I'm abusing it here since I know it uses `copy2` under the hood and `copy2` allows me to use a directory as a target but I need to make sure that directory exists first (otherwise it copies it as a file with the directory name).

Comment (ocaisa) [reply easybuild/easyblocks/c/cp2k.py]: Yeah, pretty sure I am right here. `copy_file` creates the parent directory of the target but it is working under the assumption that the target is a file. I'm abusing it here since I know it uses `copy2` under the hood and `copy2` allows me to use a directory as a target but I need to make sure that directory exists first (otherwise it copies it as a file with the directory name).

Comment (ocaisa) [reply easybuild/easyblocks/c/cp2k.py]: Yeah, pretty sure I am right here. `copy_file` creates the parent directory of the target but it is working under the assumption that the target is a file. I'm abusing it here since I know it uses `copy2` under the hood and `copy2` allows me to use a directory as a target but I need to make sure that directory exists first (otherwise it copies it as a file with the directory name).

Comment (ocaisa) [reply easybuild/easyblocks/c/cp2k.py]: Yeah, pretty sure I am right here. `copy_file` creates the parent directory of the target but it is working under the assumption that the target is a file. I'm abusing it here since I know it uses `copy2` under the hood and `copy2` allows me to use a directory as a target but I need to make sure that directory exists first (otherwise it copies it as a file with the directory name).

Comment (boegel) [reply easybuild/easyblocks/c/cp2k.py]: @ocaisa Then just copy to `os.path.join(targetdir, os.path.basename(include_file))` instead?

File: easybuild/easyblocks/c/cp2k.py [hunk 3]
```diff
@@ -845,6 +855,8 @@ def sanity_check_step(self):
         }
         if self.cfg['library']:
             custom_paths['files'].append(os.path.join('lib', 'libcp2k.a'))
+            custom_paths['files'].append(os.path.join('include', 'libcp2k.h'))
+            custom_paths['files'].append(os.path.join('include', 'libcp2k.mod'))
```
ocaisa: In the [CP2K ticket](https://github.com/cp2k/cp2k/issues/53) they said it's enough to have these files, for my Fortran user who asked for this this wasn't the case however. I think a few more mod files are required but rather than track them down I'm sticking with the naive approach of putting them all in the include dir. If someone else knows better I'm open to suggestions

boegel:
lgtm, tested it and work like a charm

PR #1555, opened 2018-10-19
add custom easyconfig parameter in GCC easyblock to control use of gold linker

boegel:
cfr. https://github.com/easybuilders/easybuild-easyconfigs/issues/6097

### DIFF
--- a/easybuild/easyblocks/g/gcc.py
+++ b/easybuild/easyblocks/g/gcc.py
@@ -71,18 +71,19 @@ class EB_GCC(ConfigureMake):
     @staticmethod
     def extra_options():
         extra_vars = {
-            'languages': [[], "List of languages to build GCC for (--enable-languages)", CUSTOM],
-            'withlibiberty': [False, "Enable installing of libiberty", CUSTOM],
-            'withlto': [True, "Enable LTO support", CUSTOM],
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/g/gcc.py
```diff
@@ -434,7 +442,7 @@ def build_step(self):
 
                         # ensure generic build when 'generic' is set to True or when --optarch=GENERIC is used
                         # non-generic build can be enforced with generic=False if --optarch=GENERIC is used
-                        if build_option('optarch') == OPTARCH_GENERIC and self.cfg['generic'] != False:
+                        if build_option('optarch') == OPTARCH_GENERIC and self.cfg['generic'] is not False:
```
akesandgren: is not False....
why not just just "self.cfg['generic']"
or at least "is True"

akesandgren: Ahh, "use False to enforce non-generic regardless".... didn't notice that one.

boegel: Indeed, it's easy to overlook that one, `False` is a special value here

akesandgren:
@vanzod want to take a look too?

PR #1556, opened 2018-10-24
quantumespresso.py: Ifort compiler needs -assume byterecl

akesandgren:
See install/configure for the default settings.

Should fix Intel builds of https://github.com/easybuilders/easybuild-easyconfigs/pull/6593

### DIFF
--- a/easybuild/easyblocks/q/quantumespresso.py
+++ b/easybuild/easyblocks/q/quantumespresso.py
@@ -162,7 +162,9 @@ class EB_QuantumESPRESSO(ConfigureMake):
             env.setvar('FCCPP', "%s -E" % os.getenv('CC'))
 
         if comp_fam == toolchain.INTELCOMP:
-            repls.append(('F90FLAGS', '-fpp', True))
+            # Intel compiler must have -assume byterecl (see install/configure)
+            repls.append(('F90FLAGS', '-fpp -assume byterecl', True))
+            repls.append(('FFLAGS', '-assume byterecl', True))
         elif comp_fam == toolchain.GCC:
             repls.append(('F90FLAGS', '-cpp', True))
 


boegel:
@akesandgren Can you clarify why this hasn't been a problem up until now? We have a bunch of easyconfigs for QE with `intel` already...

bartoldeman:
This may actually solve a longstanding ticket we have here about the use of EPW
They tried the examples downloaded from  https://github.com/QEF/q-e/releases/tag/qe-6.3 .
Both  "/test-suite/epw_metal"  and  "epw_base".  crashed with the error in epw.out

 ```

  %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%      task #         8     from readwfc : error #        20
     error opening wfc file
 %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
```

migueldiascosta:
@bartoldeman yes, this should fix the epw test-suite issue, at least it did for me

migueldiascosta:
@boegel the inconsistency within each build comes from epw defining the record length, in `EPW/src/readwfc.f90`, differently from the rest of QE, in `Modules/io_files.f90`. In particular, QE gets the `DIRECT_IO_FACTOR` with `INQUIRE` while epw uses a define that assumes the record length is in bytes.

But I think this does mean that the direct access unformatted files created from builds with and without `-assume byterecl` are incompatible. The reason why it's usually not an issue is probably that these direct access files, the "distributed" wavefunctions, are typically used in intermediate steps of the same workflow. The "collected" wavefunctions are formatted and not affected by this...

migueldiascosta:
actually, where `INQUIRE` is used it should be compatible regardless of `-assume byterecl`, e.g. for double precision it will return 2 (4-byte words) without the flag and 8 (bytes) with it, so the flag is only needed where `DIRECT_IO_FACTOR` is defined instead of inquired. That seems to be only in epw...

migueldiascosta:
https://github.com/QEF/q-e/commit/1b50a031097d8a259fe0e7972029d4a607be9681

 "Preprocessed variable DIRECT_IO_FACTOR replaced by a standard fortran call to "inquire". Obsolete for PWscf but still used by postprocessing and phonons. To be tested. To be done: EPW, containing several such preprocessed variables."

akesandgren:
The shorter summary is that it has always been a problem but nobody noticed...
QME itself always used byterecl, we (I) just forgot to make sure we did the same thing.

bartoldeman:
`make run-custom-test testdir=epw_base`
now gives me:
`All done. 13 out of 13 tests passed.`
whereas before, `All done. ERROR: only 6 out of 13 tests passed.`.

PR #1559, opened 2018-10-25
pass $PYTHONPATH while building TensorFlow, disable cross-compilation mode if optarch is set, fix sanity check for installing TensorFlow as extension

boegel:
Passing down `$PYTHONPATH` to Bazel is required to be able to build `TensorFlow` 1.11.0, which requires that particular Python packages are pre-installed (`keras_applications` and `keras_preprocessing`), see also https://github.com/tensorflow/tensorflow/issues/22395 .

Also using `--distinct_host_configuration=false` is required for TensorFlow 1.11.0 because of a bug in Bazel (see https://github.com/bazelbuild/bazel/issues/6473).

But it's probably a good idea to use `--distinct_host_configuration=false` when not cross compiling, , since it can significantly speed up the build process (see https://docs.bazel.build/versions/master/guide.html#configurations).

Both these changes should not affect existing easyconfigs (already verified with `TensorFlow-1.10.1-foss-2018b-Python-3.6.6.eb`).

### DIFF
--- a/easybuild/easyblocks/t/tensorflow.py
+++ b/easybuild/easyblocks/t/tensorflow.py
@@ -439,7 +439,7 @@ class EB_TensorFlow(PythonPackage):
         # test installation using MNIST tutorial examples
         if self.cfg['runtest']:
             pythonpath = os.getenv('PYTHONPATH', '')
-            env.setvar('PYTHONPATH', '%s:%s' % (os.path.join(self.installdir, self.pylibdir), pythonpath))
+            env.setvar('PYTHONPATH', os.pathsep.join([os.path.join(self.installdir, self.pylibdir), pythonpath]))
 
             for mnist_py in ['mnist_softmax.py', 'mnist_with_summaries.py']:
                 tmpdir = tempfile.mkdtemp(suffix='-tf-%s-test' % os.path.splitext(mnist_py)[0])
--- a/easybuild/easyblocks/t/tensorflow.py
+++ b/easybuild/easyblocks/t/tensorflow.py
@@ -432,6 +432,10 @@ class EB_TensorFlow(PythonPackage):
         ]
         res = super(EB_TensorFlow, self).sanity_check_step(custom_paths=custom_paths, custom_commands=custom_commands)
 
+        # determine top-level directory
+        # start_dir is not set when TensorFlow is installed as an extension, then fall back to ext_dir
+        topdir = self.start_dir or self.ext_dir
+
         # test installation using MNIST tutorial examples
         if self.cfg['runtest']:
             pythonpath = os.getenv('PYTHONPATH', '')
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/t/tensorflow.py
```diff
@@ -342,6 +342,21 @@ def build_step(self):
         if self.toolchain.options.get('pic', None):
             cmd.append('--copt="-fPIC"')
 
+        # include install location of Python packages in $PYTHONPATH,
+        # and specify that value of $PYTHONPATH should be passed down into Bazel build environment;
+        # this is required to make sure that Python packages included as extensions are found at build time;
+        # see also https://github.com/tensorflow/tensorflow/issues/22395
+        pythonpath = os.getenv('PYTHONPATH', '')
+        env.setvar('PYTHONPATH', '%s:%s' % (os.path.join(self.installdir, self.pylibdir), pythonpath))
```
akesandgren: Use os.pathjoin instead of %s:%s ?

easybuild/easyblocks/t/tensorflow.py
```diff
@@ -427,3 +442,5 @@ def sanity_check_step(self):
                 mnist_py = os.path.join(self.start_dir, 'tensorflow', 'examples', 'tutorials', 'mnist', mnist_py)
                 cmd = "%s %s --data_dir %s" % (self.python_cmd, mnist_py, tmpdir)
                 run_cmd(cmd, log_all=True, simple=True, log_ok=True)
+
+        return res
```
akesandgren: Should this res be additive with the mnist test result?

boegel:
There's a problem popping up I wasn't seeing before:

```
  File "/user/data/gent/vsc400/vsc40023/easybuild_easy_installed/lib/python2.7/site-packages/easybuild_framework-3.7.1.dev0-py2.7.egg/easybuild/framework/easyblock.py", line 2237, in _sanity_check_step_extensions
    res = ext.sanity_check_step()
  File "/tmp/eb-anfye_/included-easyblocks/easybuild/easyblocks/tensorflow.py", line 442, in sanity_check_step
    mnist_py = os.path.join(self.start_dir, 'tensorflow', 'examples', 'tutorials', 'mnist', mnist_py)
  File "/usr/lib64/python2.7/posixpath.py", line 77, in join
    elif path == '' or path.endswith('/'):
AttributeError: 'NoneType' object has no attribute 'endswith'
```

This may have something to do with using `PythonBundle` to install `TensorFlow`, see also https://github.com/easybuilders/easybuild-easyblocks/pull/1553 .

easybuild/easyblocks/t/tensorflow.py
boegel: You probably mean `os.pathsep.join`? If so, done

easybuild/easyblocks/t/tensorflow.py
boegel: As it is now, if the mnist test fails, it will raise an error straight away rather than postponing it until the result of the sanity check for extensions is reported.

That's probably good enough, since `TensorFlow` will typically be the last extension being installed.

Changing this in a way that works both for `TensorFlow` installed as an extension and stand-alone is a bit messy...

boegel:
`AttributeError` problem is fixed with changes in 92885f7

easybuild/easyblocks/t/tensorflow.py
akesandgren: Yeah, can never remember the exact name of all these functions :-)

easybuild/easyblocks/t/tensorflow.py
akesandgren: Ok, good enough

PR #1561, opened 2018-10-26
add new generic easyblock that implements installation procedure using Meson & Ninja

boegel:


easybuild/easyblocks/generic/mesonninja.py
```diff
@@ -0,0 +1,103 @@
...
+    def configure_step(self, cmd_prefix=''):
+        """
+        Configure with Meson.
+        """
+        if self.cfg.get('separate_build_dir', True):
+            builddir = os.path.join(self.builddir, 'easybuild_obj')
+            mkdir(builddir)
+            change_dir(builddir)
+
+        cmd = "%(preconfigopts)s meson --prefix %(installdir)s %(configopts)s %(sourcedir)s" % {
```
akesandgren: Wouldn't it be better to have a check that meson and ninja is actually in the builddeps list first?
Makes for better error messages.

easybuild/easyblocks/generic/mesonninja.py
```diff
@@ -0,0 +1,103 @@
...
+        (out, _) = run_cmd(cmd, log_all=True, simple=False)
+        return out
+
+    def build_step(self, verbose=False, path=None):
+        """
+        Build with Ninja.
+        """
+        parallel = ''
+        if self.cfg['parallel']:
+            parallel = "-j %s" % self.cfg['parallel']
```
akesandgren: Aren't we doing -j by default? At least for configuremake and cmakemake?

easybuild/easyblocks/generic/mesonninja.py
```diff
@@ -0,0 +1,103 @@
...
+        """
+        if self.cfg['runtest']:
+            (out, _) = run_cmd(self.cfg['runtest'], log_all=True, simple=False)
+            return out
+
+    def install_step(self):
+        """
+        Install with 'ninja install'.
+        """
+        cmd = "%(preinstallopts)s ninja %(installopts)s install" % {
```
akesandgren: And there is usually no reason to not use -j when doing install, i.e. if the build step can do -j then so can the install ...

ocaisa:
@boegel That's funny, I was just thinking we needed this yesterday. Shouldn't there be a check that we have the necessary commands (`meson`, `ninja`) available, whether via an EB dep or otherwise?

easybuild/easyblocks/generic/mesonninja.py
boegel: We are, `self.cfg['parallel']` is set to the available cores by default.

easybuild/easyblocks/generic/mesonninja.py
boegel: Well, I can add it, but we generally don't do it for `install`...

easybuild/easyblocks/generic/mesonninja.py
akesandgren: oops, reading the code wrong (again)

easybuild/easyblocks/generic/mesonninja.py
akesandgren: Any specific reason why not?

boegel: If we'd start doing it for `ConfigureMake`, my guess is that we'd hit many failing installations...

And to escape we'd need to add a new `parallel_install` eaysyconfig parameter to avoid having to build sequentially...

easybuild/easyblocks/generic/mesonninja.py
boegel: done

boegel:
@ocaisa Seems like this is being picked up by many projects: `FUSE`, `ATK`, `Gdk-Pixbuf`, ...

PR #1562, opened 2018-10-26
Added a test to disable compiling FFTW with MPI if the toolchain does…

mboisson:
… not support MPI


mboisson:
Do you have a suggestion on how to do that ? Looking for « MPI » in the name of each dependencies doesn’t seem like a good way to do it (mvapich, mpi4py, compiler, would be counterexamples of how it would not work.)
  

  
  

  
  
>

boegel:
@akesandgren I'm not sure that makes much sense... Also, as @mboisson pointed out, detecting an MPI library in the list of dependencies in a clean way is hard.

Do you have a particular example use case in mind?

akesandgren:
No, it was just an idle thought. Ignore it.

PR #1563, opened 2018-10-28
ENH: update WRF and WPS to support version 4

andreas-h:
Both WRF and WPS saw quite some changes in version 4.0.  This commit
adapts the existing WPS and WRF easyblocks to accomodate them.

### DIFF
--- a/easybuild/easyblocks/w/wps.py
+++ b/easybuild/easyblocks/w/wps.py
@@ -65,8 +65,10 @@ class EB_WPS(EasyBlock):
         testdata_urls = ["http://www2.mmm.ucar.edu/wrf/src/data/avn_data.tar.gz"]
         if LooseVersion(self.version) < LooseVersion('3.8'):
             testdata_urls.append("http://www2.mmm.ucar.edu/wrf/src/wps_files/geog.tar.gz")  # 697MB download, 16GB unpacked!
-        else:
+        elif LooseVersion(self.version) < LooseVersion('4.0'):
             testdata_urls.append("http://www2.mmm.ucar.edu/wrf/src/wps_files/geog_complete.tar.bz2")  # 2.3GB download!
+        else:
+            testdata_urls.append("http://www2.mmm.ucar.edu/wrf/src/wps_files/geog_high_res_mandatory.tar.gz")  # 2.6GB download, 29GB unpacked!!
         if self.cfg.get('testdata') is None:
             self.cfg['testdata'] = testdata_urls
 
@@ -93,8 +95,14 @@ class EB_WPS(EasyBlock):
         # WRF dependency check
         wrf = get_software_root('WRF')
         if wrf:
-            majver = get_software_version('WRF').split('.')[0]
-            self.wrfdir = os.path.join(wrf, "WRFV%s" % majver)
+            wrfver = get_software_version('WRF')
+            majver = wrfver.split('.')[0]
+            if majver == '3':
+                self.wrfdir = os.path.join(wrf, "WRFV%s" % majver)
+            elif majver == '4':
+                self.wrfdir = os.path.join(wrf, "WRF-%s" % wrfver)
+            else:
+                raise EasyBuildError("Unknown WRF version: %s" % wrfver)
         else:
             raise EasyBuildError("WRF module not loaded?")
 
@@ -251,7 +259,12 @@ class EB_WPS(EasyBlock):
             if not self.cfg['testdata']:
                 raise EasyBuildError("List of URLs for testdata not provided.")
 
-            wpsdir = os.path.join(self.builddir, "WPS")
+
+            mainver = self.version.split('.')[0]
+            if mainver < 4:
+                wpsdir = os.path.join(self.builddir, "WPS")
+            else:
+                wpsdir = os.path.join(self.builddir, "WPS-%s" % self.version)
 
             try:
                 # create temporary directory
@@ -275,8 +288,12 @@ class EB_WPS(EasyBlock):
                 # GEOGRID
 
                 # setup directories and files
-                for d in os.listdir(os.path.join(tmpdir, "geog")):
-                    os.symlink(os.path.join(tmpdir, "geog", d), os.path.join(tmpdir, d))
+                if mainver < 4:
+                    geog_data_dir = "geog"
+                else:
+                    geog_data_dir = "WPS_GEOG"
+                for d in os.listdir(os.path.join(tmpdir, geog_data_dir)):
+                    os.symlink(os.path.join(tmpdir, geog_data_dir, d), os.path.join(tmpdir, d))
 
                 # copy namelist.wps file and patch it for geogrid
                 copy_file(os.path.join(wpsdir, 'namelist.wps'), namelist_file)
@@ -356,8 +373,13 @@ class EB_WPS(EasyBlock):
 
     def sanity_check_step(self):
         """Custom sanity check for WPS."""
+        mainver = self.version.split('.')[0]
+        if mainver < 4:
+            wpsdir = 'WPS'
+        else:
+            wpsdir = 'WPS-%s' % self.version
         custom_paths = {
-            'files': ['WPS/%s' % x for x in ['geogrid.exe', 'metgrid.exe', 'ungrib.exe']],
+            'files': ['%s/%s' % (wpsdir, x) for x in ['geogrid.exe', 'metgrid.exe', 'ungrib.exe']],
             'dirs': [],
         }
         super(EB_WPS, self).sanity_check_step(custom_paths=custom_paths)
--- a/easybuild/easyblocks/w/wps.py
+++ b/easybuild/easyblocks/w/wps.py
@@ -339,9 +339,8 @@ class EB_WPS(EasyBlock):
                 run_wps_cmd('metgrid')
 
                 # clean up
-                remove_dir(tmpdir)
-
                 change_dir(self.builddir)
+                remove_dir(tmpdir)
 
             except OSError as err:
                 raise EasyBuildError("Failed to run WPS test: %s", err)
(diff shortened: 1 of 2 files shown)


easybuild/easyblocks/w/wps.py
```diff
@@ -93,8 +95,14 @@ def configure_step(self):
         # WRF dependency check
         wrf = get_software_root('WRF')
         if wrf:
-            majver = get_software_version('WRF').split('.')[0]
-            self.wrfdir = os.path.join(wrf, "WRFV%s" % majver)
+            wrfver = get_software_version('WRF')
+            majver = wrfver.split('.')[0]
+            if majver == '3':
+                self.wrfdir = os.path.join(wrf, "WRFV%s" % majver)
+            elif majver == '4':
```
boegel: @andreas-h Please use LooseVersion here rather than doing string comparison on the major version

easybuild/easyblocks/w/wps.py
```diff
@@ -93,8 +95,14 @@ def configure_step(self):
...
+                self.wrfdir = os.path.join(wrf, "WRF-%s" % wrfver)
+            else:
+                raise EasyBuildError("Unknown WRF version: %s" % wrfver)
```
boegel: I would make this less strict, and assume that ‘WRF-‘ is used from version 4.x onwards.

Even better would be to employ a slightly more dynamic approach to finding the subdir, e.g. go with the subdirectory that starts with ‘WRF’ if there’s only one, and error out otherwise.

easybuild/easyblocks/w/wps.py
```diff
@@ -251,7 +259,12 @@ def run_wps_cmd(cmdname, mpi_cmd=True):
             if not self.cfg['testdata']:
                 raise EasyBuildError("List of URLs for testdata not provided.")
 
-            wpsdir = os.path.join(self.builddir, "WPS")
+
+            mainver = self.version.split('.')[0]
+            if mainver < 4:
```
boegel: This does not do what you think it does :)

You’re comparing a string with an integer here. That doesn’t cause any type errors in Python 2.x, but you really don’t want to do that since this will always be False regardless of the value of mainver...

Please also use LooseVersion here, and drop the ‘split’ above, compare using self.version

easybuild/easyblocks/w/wps.py
```diff
@@ -275,8 +288,12 @@ def run_wps_cmd(cmdname, mpi_cmd=True):
                 # GEOGRID
 
                 # setup directories and files
-                for d in os.listdir(os.path.join(tmpdir, "geog")):
-                    os.symlink(os.path.join(tmpdir, "geog", d), os.path.join(tmpdir, d))
+                if mainver < 4:
```
boegel: Same here, please use LooseVersion to check versions

easybuild/easyblocks/w/wps.py
```diff
@@ -275,8 +288,12 @@ def run_wps_cmd(cmdname, mpi_cmd=True):
...
+                    geog_data_dir = "geog"
+                else:
+                    geog_data_dir = "WPS_GEOG"
+                for d in os.listdir(os.path.join(tmpdir, geog_data_dir)):
+                    os.symlink(os.path.join(tmpdir, geog_data_dir, d), os.path.join(tmpdir, d))
```
boegel: While touching this, please switch to using the symlink function that is available in easybuild.tools.filetools, and avoid using single-letter variable names outside of list comprehensions (so use ‘dir_name’ rather than ‘d’).

easybuild/easyblocks/w/wrf.py
```diff
@@ -362,8 +369,10 @@ def install_step(self):
     def sanity_check_step(self):
         """Custom sanity check for WRF."""
 
-        mainver = self.version.split('.')[0]
-        self.wrfsubdir = "WRFV%s" % mainver
+        if LooseVersion(self.version) < LooseVersion('4.0'):
+            self.wrfsubdir = "WRFV3"
+        else:
+            self.wrfsubdir = "WRF-%s" % self.version
```
akesandgren: Move the self.wrfsubdir setting to "init" instead of repeating the code. And use it in the setting of wrfdir above. Cleaner and less code.

boegel: @andreas-h I took a look at this myself + some more cleanup for these easyblocks, please take a look at https://github.com/andreas-h/easybuild-easyblocks/pull/4

easybuild/easyblocks/w/wps.py
andreas-h: addressed by https://github.com/andreas-h/easybuild-easyblocks/pull/4

easybuild/easyblocks/w/wps.py
andreas-h: not sure if going the dynamic route is worth it, given potential future changes in WRF layout.

easybuild/easyblocks/w/wps.py
andreas-h: addressed by https://github.com/andreas-h/easybuild-easyblocks/pull/4

easybuild/easyblocks/w/wps.py
andreas-h: addressed by https://github.com/andreas-h/easybuild-easyblocks/pull/4

easybuild/easyblocks/w/wrf.py
andreas-h: Thanks a lot :)

easybuild/easyblocks/w/wps.py
```diff
@@ -340,20 +330,20 @@ def run_wps_cmd(cmdname, mpi_cmd=True):
...
 
                 # run metgrid.exe
                 run_wps_cmd('metgrid')
 
                 # clean up
-                rmtree2(tmpdir)
+                remove_dir(tmpdir)
 
-                os.chdir(self.builddir)
+                change_dir(self.builddir)
```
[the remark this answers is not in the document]
andreas-h: Agreed. 

That's what `tempfile.mkdtemp()` is doing according to Python 2.7 and 3.6 docs.  Should I just reverse the order (first change out of `tmpdir`, then remove it), or should I not trust Python and first make sure that `tmpdir` is actually an absolute path?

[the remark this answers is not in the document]
akesandgren: If python doccos says it's an absdir then trust it. So just reverse the order.

easybuild/easyblocks/w/wps.py
boegel: At least then there's a chance that this will work, so it's worth a shot imho.

easybuild/easyblocks/w/wrf.py
```diff
@@ -251,7 +268,7 @@ def test_step(self):
 
             # exclude 2d testcases in non-parallel WRF builds
             if self.cfg['buildtype'] in self.parallel_build_types:
-                self.testcases = [test for test in self.testcases if not "2d_" in test]
+                self.testcases = [test for test in self.testcases if '2d_' not in test]
```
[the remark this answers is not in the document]
andreas-h: It seems to me that the code is right, but the comment is wrong.  When looking at the 2-D testcases, those are really 2-D (x times z or y times z).  Since WRF paralellizes over x/y, it does kind of make sense that parallelization doesn't work in those cases. At least on my machine, I cannot build the 2-D testcases with parallelization.

So my way forward would be to fix the comment to read "exclude 2d testcases in parallel WRF builds".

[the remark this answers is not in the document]
akesandgren: Ok, as long as it's consistent :-)

easybuild/easyblocks/w/wps.py
andreas-h: The logical place to put this would be the new `det_wrf_subdir` function.  However, then, we would need to check for existence of the directory in that subdir.  In the wrf easyblock, the function gets called from `__init__()`.

I'm not sure about the internals of EasyBuild; however I suspect that by the time the easyblock's `__init__` is being run, the sources have not been unpacked yet, so we cannot implement the dynamic route there.

Would it be safe to call the `det_wrf_subdir` from the `configure_step()`? That way, we could rely on the sources being unpacked, and could implement the dynamic check without code duplication.

boegel: It depends where the result from `det_wrf_subdir` is used. We should try and make sure that things work properly when `--module-only` is used (which skips the `configure` step entirely, but doesn't skip `sanity_check_step`).

One downside of checking for a particular subdir is that it'll interfere with `-x` (i.e. it may result in an error because no files are present whatsoever under `-x`). That's something we should try and avoid...

So, with all that in mind, I'd keep it the way it is now in `det_wrf_subdir`: use `WRF-<version>` for newer versions. If needed, it'll be easier to update that for future versions.

boegel:
@akesandgren Are you OK with current status on this?

easybuild/easyblocks/w/wps.py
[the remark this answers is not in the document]
akesandgren: This still needs to be fixed.

easybuild/easyblocks/w/wrf.py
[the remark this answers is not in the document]
akesandgren: Comment still wrong.

akesandgren:
Still 2 things to change, chdir/remove_dir order and the comment about 2D tests

easybuild/easyblocks/w/wps.py
andreas-h: agreed.

akesandgren:
@boegel I'm ok with this. It's your change requests that are blocking.

boegel:
Thanks a lot for the updates @andreas-h!

Issue #1564, opened 2018-11-01
ANSYS 19.2 install fails due to failure to adjust_permissions() on a broken symlink


chrissamuel:
Trying to install ANSYS 19.2 for a user results in the following error:
```
[csamuel@john99 ANSYS]$ eb ANSYS-19.2.eb
== temporary log file in case of crash /tmp/eb-KRPVhr/easybuild-oHgIW0.log
== processing EasyBuild easyconfig /apps/generic/recipes/a/ANSYS/ANSYS-19.2.eb
== building and installing core/ansys/19.2...
== fetching files...
== creating build dir, resetting environment...
== unpacking...
== patching...
== preparing...
== configuring...
== building...
== testing...
== installing...
== FAILED: Installation ended unsuccessfully (build directory: /dev/shm/easybuild/ANSYS/19.2/dummy-dummy): build failed (first 300 chars): Failed to chmod/chown several paths: ['/apps/skylake/software/core/ansys/19.2/v192/ansys/bin/.change_ansysdir192'] (last error: [Errno 2] No such file or directory: '/apps/skylake/software/core/ansys/19.2/v192/ansys/bin/.change_ansysdir192')
== Results of the build can be found in the log file(s) /tmp/eb-KRPVhr/easybuild-ANSYS-19.2-20181101.185240.rpbIc.log
```
The symlink `/apps/skylake/software/core/ansys/19.2/v192/ansys/bin/.change_ansysdir192` is indeed broken:

```
[csamuel@john99 ANSYS]$ ls -l /apps/skylake/software/core/ansys/19.2/v192/ansys/bin/.change_ansysdir192
lrwxrwxrwx 1 csamuel hpcadmin 19 Aug  9 06:10 /apps/skylake/software/core/ansys/19.2/v192/ansys/bin/.change_ansysdir192 -> .change_ansysdir.sh
```

The easybuild log mentioned above only has this appearing relevant (no errors during build :
```
== 2018-11-01 18:54:22,170 easyblock.py:2515 INFO Starting install step
== 2018-11-01 18:54:22,170 easyblock.py:2521 INFO Running method stage_install_step part of step install
== 2018-11-01 18:54:22,170 easyblock.py:2521 INFO Running method make_installdir part of step install
== 2018-11-01 18:54:22,171 easyblock.py:879 INFO Found old directory /apps/skylake/software/core/ansys/19.2
[...]

PR #1566, opened 2018-11-01
ignore errors when adjusting permissions in ANSYS easyblock, to avoid failure over a broken symlink

boegel:
fix for problem reported in #1564 by @chrissamuel


boegel:
This change doesn't make much sense, since `adjust_permissions` skip symlinks by default, see discussion in #1564

PR #1567, opened 2018-11-01
enhance PythonPackage/PythonBundle to run 'pip check' when 'pip install --no-deps' is used (WIP)

### LINKED ISSUE #1565: download_dep_fail doesn't catch downloaded dependencies when use_pip=True
This is because `use_pip` adds `--no-deps` to pip's `installopts` and doesn't try to download any dependencies

It would only make sense to use `download_dep_fail` together with `use_pip_for_deps`, but since `--ignore-installed` is also added to `installopts`, this then tries to download all dependencies even when already present

As it is, it seems `download_dep_fail` is only useful with `easy_install`

boegel:
Enhance `PythonPackage` to run '`pip check`' when '`pip install --no-deps`' is used, to ensure all required dependencies are available.

This (sort of) fixes #1565, in the sense that it'll make sure that all required Python packages are indeed in place, although it's not done during the sanity check, but before the actual installation of the Python package...

When testing this, this already signaled that `cffi` was missing in a bunch of `Python` easyconfigs (which is being fixed in https://github.com/easybuilders/easybuild-easyconfigs/pull/7105).

(WIP because still testing the impact this has on easyconfigs that set `use_pip = True`...)

cc @migueldiascosta

### DIFF
--- a/easybuild/easyblocks/generic/pythonpackage.py
+++ b/easybuild/easyblocks/generic/pythonpackage.py
@@ -207,6 +207,7 @@ class PythonPackage(ExtensionEasyBlock):
         self.unpack_options = self.cfg['unpack_options']
 
         self.python_cmd = None
+        self.pip_version = None
         self.pylibdir = UNKNOWN
         self.all_pylibdirs = [UNKNOWN]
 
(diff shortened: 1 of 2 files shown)


easybuild/easyblocks/generic/pythonpackage.py
```diff
@@ -502,9 +502,15 @@ def install_step(self):
         new_pythonpath = os.pathsep.join([x for x in abs_pylibdirs + [pythonpath] if x is not None])
         env.setvar('PYTHONPATH', new_pythonpath, verbose=False)
 
+        install_cmd = self.compose_install_command(self.installdir)
+        self.log.info("Installing %s v%s with: %s", self.name, self.version, install_cmd)
+
+        # run 'pip check' before trying to install package, if pip is used with --no-deps
+        if "pip install" in install_cmd and '--no-deps' in install_cmd:
+            run_cmd("pip check", log_all=True, log_ok=True, simple=False)
```
migueldiascosta: this is checking the dependencies of installed packages, but the current one isn't installed yet

(it does work for the ones that are themselves used as dependencies later, but the "last" one isn't checked)

boegel: This is checking whether all required dependencies of the package to-be-installed are in place; if not, `pip check` will fail.

So rather than checking for dependencies being downloaded during the installation, we're simply checking before the installation...

boegel: After discussing this with @migueldiascosta it's clear that I misunderstood what `pip check` does.

It indeed checks whether all dependencies of *installed* packages are available, not of the package to-be-installed (which is how I understood this).

I'll rework this, and see if it's still useful.

Thanks for pointing this out @migueldiascosta!

boegel:
closing this, superseded by #1853...

PR #1568, opened 2018-11-02
adapt PSI easyblock for PSI4 > 1.2

hajgato:


easybuild/easyblocks/p/psi.py
```diff
@@ -139,7 +140,12 @@ def configure_step(self):
         else:
             self.log.info("Using CMake based build")
             self.cfg.update('configopts', ' -DPYTHON_INTERPRETER=%s' % os.path.join(pythonroot, 'bin', 'python'))
-            self.cfg.update('configopts', ' -DCMAKE_BUILD_TYPE=Release')
+            if (self.name == 'PSI4') and (LooseVersion(self.version) >= LooseVersion("1.2")):
+                self.log.info("You have to remove CMAKE_BUILD_TYPE test in PSI4 source and the downoaded dependencies!")
+                self.log.info("Use PATCH_COMMAND in the corresponding CMakeLists.txt")
+                self.cfg.update('configopts', ' -DCMAKE_BUILD_TYPE=None')
```
boegel: @hajgato Can you clarify this? Is this supposed to abort the build?

easybuild/easyblocks/p/psi.py
```diff
@@ -139,7 +140,12 @@ def configure_step(self):
         else:
             self.log.info("Using CMake based build")
             self.cfg.update('configopts', ' -DPYTHON_INTERPRETER=%s' % os.path.join(pythonroot, 'bin', 'python'))
-            self.cfg.update('configopts', ' -DCMAKE_BUILD_TYPE=Release')
+            if (self.name == 'PSI4') and (LooseVersion(self.version) >= LooseVersion("1.2")):
```
boegel: @hajgato Style nitpicking: no need for the `(...)`:

```python
if self.name == 'PSI4' and LooseVersion(self.version) >= LooseVersion("1.2"):

File: easybuild/easyblocks/p/psi.py [hunk 3]
```diff
@@ -150,11 +156,30 @@ def configure_step(self):
...
+                    self.cfg.update('configopts', " -DENABLE_CHEMPS2=ON")
+                    if LooseVersion(self.version) < LooseVersion("1.2"):
+                        self.cfg.update('configopts', " -DCHEMPS2_ROOT=%s" % chempsroot)
+                    else:
+                        self.cfg.update('configopts', " -DCMAKE_INSIST_FIND_PACKAGE_CheMPS2=ON "
+                                        "-DCheMPS2_DIR=%s/share/cmake/CheMPS2" % chempsroot)
+
+                #  Be aware, PSI4 wants exact versions of the following deps! built with CMake!!
+                if LooseVersion(self.version) >= LooseVersion("1.2"):
+                    depsdict = {'libxc': 'Libxc', 'Libint': 'Libint', 'pybind11': 'pybind11', 'gau2grid': 'gau2grid'}
```
boegel: @hajgato Neither `get_software_root` nor CMake cares about casing, so just use a list of dep names here?

easybuild/easyblocks/p/psi.py
```diff
@@ -150,11 +156,30 @@ def configure_step(self):
...
+                    for deps in depsdict:
```
boegel: `dep` makes more sense here (singular)

easybuild/easyblocks/p/psi.py
```diff
@@ -150,11 +156,30 @@ def configure_step(self):
...
                 if chempsroot:
-                    self.cfg.update('configopts', " -DENABLE_CHEMPS2=ON -DCHEMPS2_ROOT=%s" % chempsroot)
+                    self.cfg.update('configopts', " -DENABLE_CHEMPS2=ON")
+                    if LooseVersion(self.version) < LooseVersion("1.2"):
+                        self.cfg.update('configopts', " -DCHEMPS2_ROOT=%s" % chempsroot)
+                    else:
+                        self.cfg.update('configopts', " -DCMAKE_INSIST_FIND_PACKAGE_CheMPS2=ON "
+                                        "-DCheMPS2_DIR=%s/share/cmake/CheMPS2" % chempsroot)
+
+                #  Be aware, PSI4 wants exact versions of the following deps! built with CMake!!
```
boegel: What happens if you use the wrong version? How can one figure out which versions PSI4 wants?

easybuild/easyblocks/p/psi.py
```diff
@@ -177,7 +202,19 @@ def test_step(self):
         """
         testdir = tempfile.mkdtemp()
         env.setvar('PSI_SCRATCH', testdir)
-        super(EB_PSI, self).test_step()
+        if (self.name == 'PSI4') and (LooseVersion(self.version) >= LooseVersion("1.2")):
```
boegel: same here, no need for the wrapping `(...)`

easybuild/easyblocks/p/psi.py
```diff
@@ -150,11 +156,30 @@ def configure_step(self):
...
+                                        "-DCheMPS2_DIR=%s/share/cmake/CheMPS2" % chempsroot)
+
+                #  Be aware, PSI4 wants exact versions of the following deps! built with CMake!!
+                if LooseVersion(self.version) >= LooseVersion("1.2"):
+                    depsdict = {'libxc': 'Libxc', 'Libint': 'Libint', 'pybind11': 'pybind11', 'gau2grid': 'gau2grid'}
+                    for deps in depsdict:
+                        depsroot = get_software_root(deps)
+                        if depsroot:
+                            self.cfg.update('configopts', " -DCMAKE_INSIST_FIND_PACKAGE_{0}=ON "
+                                            "-D{0}_DIR={1}/share/cmake/{2}" .format(depsdict[deps], depsroot, deps))
```
boegel:
```python
dep_config_opt = " -DCMAKE_INSIST_FIND_PACKAGE_{0}=ON " % dep
dep_config_opt += "-D%s_DIR=%s % (dep, os.path.join(deproot, 'share', 'cmake', dep))
self.cfg.update('configopts', dep_config_opt)
```

easybuild/easyblocks/p/psi.py
hajgato: @boegel 
Building `Release` version involves adding `-O3 -DNDEBUG` flags (after EB defined flags). 
There are 2 ways to avoid that, build in non-release version (then we have to patch the downloaded cmake/autocmake_safeguards.cmake)
or patch the downloaded CMakeLists.txt files to disobey `CMake` defined CMAKE_${lang}_FLAGS_RELEASE variables (Which is hardcoded in `CMake` as CMAKE_${lang}_FLAGS_RELEASE_INIT). 
Maybe there are another way, but I do not know about it or I was not thinking about it.

easybuild/easyblocks/p/psi.py
hajgato: If we use the wrong version then PSI4 will not find them. Otherwise you have to patch the corresponding `CMakeLists.txt` files in the source.

easybuild/easyblocks/p/psi.py
boegel: @hajgato Can't we specify `-DCMAKE_CXX_FLAGS_RELEASE="$CXXFLAGS"` via `configopts`?

easybuild/easyblocks/p/psi.py
boegel: @hajgato Please mention the latter in the comment in the easyblock?

easybuild/easyblocks/p/psi.py
```diff
@@ -150,11 +156,30 @@ def configure_step(self):
...
+                                        "-DCheMPS2_DIR=%s/share/cmake/CheMPS2" % chempsroot)
+
+                #  Be aware, PSI4 wants exact versions of the following deps! built with CMake!!
+                if LooseVersion(self.version) >= LooseVersion("1.2"):
+                    for dep in ['libxc', 'Libint', 'pybind11', 'gau2grid']:
+                        deproot = get_software_root(dep)
+                        if deproot:
+                            self.cfg.update('configopts', " -DCMAKE_INSIST_FIND_PACKAGE_%s=ON" % dep)
+                            self.cfg.update('configopts', " -D%s_DIR%s "
+                                            % (dep, os.path.join(deproot, 'share', 'cmake', dep)))
```
boegel: @hajgato This is missing a `=` now?

```python
dep_dir = os.path.join(deproot, 'share', 'cmake', dep)
self.cfg.update('configopts', " -D%s_DIR=%s " % (dep, dep_dir))

easybuild/easyblocks/p/psi.py
hajgato: @boegel 
Yes we could. But then that would apply only for `PSI4`, and not for the external projects (libint, libxc, etc, and a as a matter of fact a part of psi4 which is build as an external project), so we should patch the external project `CMakeList.txt` files for `CMAKE_CXX_FLAGS_RELEASE`. So whatever you do, you cannot solve it without patching.

boegel: OK, but that seems better to me than using `-DCMAKE_BUILD_TYPE=None` which looks quite fishy...

How much work would it be to make sure `libxc` & co pick up on `CMAKE_CXX_FLAGS_RELEASE`?

hajgato: @boegel The problem that you have to let `CMake`  get patched the downloaded dependencies. So you have to patch the "source" CMakeLists.txt file (for each corresponding deps), then you have to make a patch for the software that will be downloaded. So in principle it is possible to do but you have to download everything to be able to do the patches (and again when you upgrade the version(s)). Maybe we can do one script that inserts the right variables to the right place in CMakeList.txt of the downloaded deps. What I feel as an advantage of the current solution that it is a simple `sed`. I know that it is an ugly hack, but in principle I could let CMake to use `-O3` everywhere, and hoping all compilation would be fine.

Just a note:
Here, I do feel that we have to find a permanent solution for `CMake`'s Release-build optimization flag hijack, as it was/will cause similar issues. I was looking here and there, but I did not find any out-of-the-box solution. It can be very annoying when developers use multiple external-projects in CMake. (like here, because in every external project you have to override the Release optimization flags, of course as long as we want to be controlled them by EasyBuild)

hajgato: @hajgato I could rename the build type from `None` to `NoAgressiveOptimizationsRelease` ...

boegel: @hajgato Or maybe `EasyBuildRelease` or something like that, to specify that it's a custom solution for EB?

I wonder if @akesandgren has some thoughts on the CMake mess here...

akesandgren: I would have to look at the actual CMake code.... Haven't had time yet though.
There are a couple of other things I need to finish up before I can spend time on this one.

hajgato:
@boegel Changed CMake build type from `None` to `EasyBuildRelease`

boegel:
Tested with existing `PSI` and `PSI4` easyconfigs + https://github.com/easybuilders/easybuild-easyconfigs/pull/7109, good to go, thanks @hajgato!

PR #1569, opened 2018-11-04
Update pgi for llvm

akesandgren:
PGI 18 ships with a LLVM version and broken permissions in the cuda nvvp tar files.

This fixes both.

### DIFF
--- a/easybuild/easyblocks/p/pgi.py
+++ b/easybuild/easyblocks/p/pgi.py
@@ -39,6 +39,7 @@ import stat
 import sys
 
 import easybuild.tools.environment as env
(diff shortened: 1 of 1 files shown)


boegel:
@akesandgren Does this also fix the `numactl` issue you mentioned in #7113?

akesandgren:
The numactl problem was caused my me accidentally having
toolchain = {'name': 'dummy', 'version': 'dummy'}

I was investigating a problem with my new pgicuda easyblock and forgot to remove it. So that was never a real problem, just me shooting myself in the foot.

PR #1570, opened 2018-11-07
delete the temp dir where test's logs are saved

pescobar:
I tested the tensorflow installation with userA and then rerun the same installation with userB and I found that there was a folder `/tmp/tensorflow` owned by userA. 

My `$TMPDIR=/scratch` so it's ignored

### DIFF
--- a/easybuild/easyblocks/t/tensorflow.py
+++ b/easybuild/easyblocks/t/tensorflow.py
@@ -447,4 +447,6 @@ class EB_TensorFlow(PythonPackage):
                 cmd = "%s %s --data_dir %s" % (self.python_cmd, mnist_py, tmpdir)
                 run_cmd(cmd, log_all=True, simple=True, log_ok=True)
 
+            os.rmdir('/tmp/tensorflow')
+
         return res


boegel:
@pescobar We should probably fix this in another way, i.e. make sure that `/tmp/tensorflow` is never used...

EasyBuild sets up a session-specific temporary directory, we should instruct TensorFlow to use that somehow (since that dir is cleaned up automatically).

@akesandgren Any idea where this comes from?

akesandgren:
I thought i had fixed that already, but i'll take another loook tomorrow.

pescobar:
that folder is used for tests . These are the leftovers after installing tensorflow

```
$ ls -1 /tmp/tensorflow/mnist/logs/mnist_with_summaries/
test
train
```

akesandgren:
Found the problem, checking if the fix is good or not...

akesandgren:
Better solution in https://github.com/easybuilders/easybuild-easyblocks/pull/1572

boegel:
agreed, #1572 is merged, so this can be closed

Thanks for the effort on this regardless, @pescobar!

PR #1572, opened 2018-11-08
Fix TensorFlow test tempdir problem

akesandgren:
mnist_with_summaries.py uses a log dir that by default is in /tmp/tensorflow

Make sure it ends up in an EB controlled tempdir that gets cleaned up when done.
The --log_dir arg to mnist_softmax.py is ignored since it doesn't use it.


boegel:
Works like a charm, so going in, thanks @akesandgren!

Issue #1574, opened 2018-11-12
VMD's reprod easyconfig  does not work


mboisson:
I am rebuilding a lot of software packages using the `.eb` that is contained in the `reprod` folder. VMD fails to rebuild with the error : 
```
== 2018-11-12 12:38:14,171 run.py:192 INFO running cmd:  ./configure  LINUXAMD64 LP64 IMD PTHREADS FLTK TK COLVARS NOSILENT TCL OPENGL MESA NETCDF  PYTHON  ICC  LINUXAMD64 LP64 IMD PTHREADS FLTK TK COLVARS NOSILENT TCL OPENGL MESA NETCDF  PYTHON  ICC
== 2018-11-12 12:38:14,336 build_log.py:158 ERROR EasyBuild crashed with an error (at ?:124 in __init__): cmd " ./configure  LINUXAMD64 LP64 IMD PTHREADS FLTK TK COLVARS NOSILENT TCL OPENGL MESA NETCDF  PYTHON  ICC  LINUXAMD64 LP64 IMD PTHREADS FLTK TK COLVARS NOSILENT TCL OPENGL MESA NETCDF  PYTHON  ICC " exited with exit code 255 and output:
Unknown option 'LINUXAMD64'
``` 

note that the list of arguments appears twice. The culprit is that somehow, the `.eb` file in the `reprod` folder contains the list in its `configopts` and it gets duplicated instead. The `.eb` file stored in the `ebfiles_repo` does not contain this variable : 
```
-bash-4.2$ grep configopts /cvmfs/soft.computecanada.ca/easybuild/ebfiles_repo/VMD/VMD-1.9.3-iimkl-2018.3-Python-2.7.14.eb
-bash-4.2$ grep configopts /cvmfs/soft.computecanada.ca/easybuild/software/2017/avx2/Compiler/intel2018.3/vmd/1.9.3/easybuild/reprod/VMD-1.9.3-iimkl-2018.3-Python-2.7.14.eb
configopts = " LINUXAMD64 LP64 IMD PTHREADS FLTK TK COLVARS NOSILENT TCL OPENGL MESA NETCDF  PYTHON  ICC "
```

mboisson:
Mmm, I think this is actually a problem in the EasyConfig.update method. It appends to the values without checking if what it's appending is already present.

mboisson:
There are other recipes for which the `reprod` recipe does not actually work. 

[...]

PR #1575, opened 2018-11-12
modified vmd.py so that configopts don't contain duplicate values (otherwise it fails to rebuild)

mboisson:
This PR requires https://github.com/easybuilders/easybuild-framework/pull/2657  and is to fix https://github.com/easybuilders/easybuild-easyblocks/issues/1574


boegel:
Tested and approved, thanks @mboisson!

PR #1576, opened 2018-11-13
enhance Mothur easyblock to support use of Boost and HDF5 as dependencies

hajgato:


boegel:
all existing `Mothur` easyconfigs working fine, so good to go, thanks @hajgato!

PR #1577, opened 2018-11-13
enhance sanity check for icc & ifort: also check for compilers_and_libraries_*/linux subdirectory

boegel:
The `icc` and `ifort` easyblocks heavily rely on the `compilers_and_libraries_%(version)s/linux` subdirectory being there when generating the module file.

If the directory is not there as expected, for example because the wrong version was used, then the generated module file is crippled.

We already have a check in place that catches this issue (the `which icc` being run in the sanity check), but it's worthwhile to also have an explicit check for this subdirectory, since that results in a clearer error message, and you're likely to quickly figure out what's wrong.

### DIFF
--- a/easybuild/easyblocks/i/icc.py
+++ b/easybuild/easyblocks/i/icc.py
@@ -69,11 +69,18 @@ class EB_icc(IntelBase):
 
         self.debuggerpath = None
 
-        if LooseVersion(self.version) >= LooseVersion('2016') and self.cfg['components'] is None:
-            # we need to use 'ALL' by default, using 'DEFAULTS' results in key things not being installed (e.g. bin/icc)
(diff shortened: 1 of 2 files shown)


boegel:
Should be fully backwards compatible w.r.t. existing `icc` and `ifort` easyconfigs, tested with:

* `icc-2016.0.109.eb`
* `icc-2017.6.256-GCC-6.4.0-2.28.eb`
* `icc-2016.3.210-GCC-5.4.0-2.26.eb`
* `icc-2018.3.222-GCC-7.3.0-2.30.eb`
* `ifort-2016.0.109.eb`
* `ifort-2018.3.222-GCC-7.3.0-2.30.eb`
* `ifort-2017.4.196-GCC-6.4.0-2.28.eb`

PR #1580, opened 2018-11-14
add version check in sanity check step of SCOTCH easyblock (+ code cleanup)

boegel:
cfr. https://github.com/easybuilders/easybuild-easyconfigs/issues/7154

cc @smoors


smoors:
this seems to work as expected:
`libptscotchparmetis.a` and `libscotchmetis.a` are installed, but not `metis.h` and `parmetis.h`.
version check also works.

PR #1581, opened 2018-11-14
fixing fftw reprod by skipping early if configopts is a list

mboisson:
Related to https://github.com/easybuilders/easybuild-easyblocks/issues/1574
Building the "reprod" recipe fails with 
```
  File "/cvmfs/soft.computecanada.ca/easybuild/lib/python2.7/site-packages/easybuild-framework/easybuild/framework/easyblock.py", line 2860, in build_and_install_one
  File "/cvmfs/soft.computecanada.ca/easybuild/lib/python2.7/site-packages/easybuild-easyblocks/easybuild/easyblocks/f/fftw.py", line 212, in run_all_steps
    self.cfg.update('configopts', [' '.join(prec_configopts) + ' ' + common_config_opts])
TypeError: cannot concatenate 'str' and 'list' objects
``` 

if not for this PR.

### DIFF
--- a/easybuild/easyblocks/f/fftw.py
+++ b/easybuild/easyblocks/f/fftw.py
@@ -149,7 +149,7 @@ class EB_FFTW(ConfigureMake):
         common_config_opts = self.cfg['configopts']
 
         # early exit if common_config_opts is not a string
-        if not isinstance(common_config_opts,basestring):
+        if not isinstance(common_config_opts, basestring):
             self.log.warning("Found configopts that is a list. Skipping auto-configuration.")
             self.log.debug("List of configure options to iterate over: %s", self.cfg['configopts'])
 
--- a/easybuild/easyblocks/f/fftw.py
+++ b/easybuild/easyblocks/f/fftw.py
@@ -148,6 +148,13 @@ class EB_FFTW(ConfigureMake):
         # keep track of configopts specified in easyconfig file, so we can include them in each iteration later
         common_config_opts = self.cfg['configopts']
 
+        # early exit if common_config_opts is not a string
+        if not isinstance(common_config_opts,basestring):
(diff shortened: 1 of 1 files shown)


mboisson:
Presumably, one could by themselves have multiple configopts if they want to trigger multiple builds (like we do for python support in other packages). I don't know if this happens much with FFTW ? 

Also, even though  easybuilders/easybuild-framework#2664  possibly fix the reprod issue for future versions, it does not help for software packages that have already been installed.

akesandgren:
So, should we merge this or not? @boegel

boegel:
I'm not sure, just skipping all the stuff we do if `configopts` is not specified as a string makes me a bit nervous...

Maybe people naively try to use a list of values for `configopts` themselves in an `FFTW` easyconfig?
If so, it shouldn't (relatively silently, since `log.warning` only goes to the log file) skip the normal iterative confguration we do...

Instead, it should fail hard with a friendly error message imho...

mboisson:
If the reprove issue has been solved with the more recent refactoring, then that’s fine. This PR was to allow the reprod to go through without error, so issuing an error would not work in that case.
  

  
  

  
  
>

boegel:
@mboisson Well, we would still get a cleaner error in case someone tries to specify `configopts` as a list by mistake, which is better than producing a traceback...

Do you mind replacing the `log.warning` + `log.debug` with

```python
raise EasyBuildError("Value for configopts parameter must be a string value, got %s", self.cfg['configopts'])
```

damianam:
@mboisson ping

mboisson:
I don't think that we should fail if an easyconfig purposefully uses an array of configopts in order to create multiple builds (as used to be done by the reprod easyconfig).

mboisson:
This may no longer apply. Closing.

PR #1586, opened 2018-11-16
avoid hardcoding --with-rdma configure option in MVAPICH2 easyblock

boegel:
fix for problem reported by @peterinl via #7167

With this change included, you can avoid that `--with-rdma` is used by setting `rdma_type = None` or `rdma_type = ''` or `rdma_type = False` in the `MVAPICH2` easyconfig file.

### DIFF
--- a/easybuild/easyblocks/m/mvapich2.py
+++ b/easybuild/easyblocks/m/mvapich2.py
@@ -70,8 +70,6 @@ class EB_MVAPICH2(EB_MPICH):
 
         if self.cfg['rdma_type']:
             add_configopts.append('--with-rdma=%s' % self.cfg['rdma_type'])
-        else:
-            add_configopts.append('--without-rdma')
 
         # enable specific support options (if desired)
         if self.cfg['withmpe']:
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/m/mvapich2.py
```diff
@@ -67,7 +67,11 @@ def configure_step(self):
 
         # additional configuration options
         add_configopts = []
-        add_configopts.append('--with-rdma=%s' % self.cfg['rdma_type'])
+
+        if self.cfg['rdma_type']:
+            add_configopts.append('--with-rdma=%s' % self.cfg['rdma_type'])
+        else:
+            add_configopts.append('--without-rdma')
```
boegel: Hmm, when testing this with the existing `MVAPICH2-2.0.1-GCC-4.8.4.eb`, I ran into this:

```
configure: error: The specified RDMA type is not supported
```

@peterinl: Maybe `--without-rdma` is not sufficient? Should this be something different?

peterinl:
I would think "--without-rdma" is fine, or better yet, just don't add the "--with-rdma" option at all.


--
Peter Cebull
High Performance Computing and Data Analytics
Idaho National Laboratory
208-526-1909

________________________________

PR #1587, opened 2018-11-16
Added Code_Saturne support

mcakircali:
tested with 5.x easyconfigs for intel-2018a (that are added new too).

### DIFF
--- a/easybuild/easyblocks/c/code_saturne.py
+++ b/easybuild/easyblocks/c/code_saturne.py
@@ -45,52 +45,52 @@ class EB_Code_underscore_Saturne(EasyBlock):
     def extra_options(extra_vars=None):
         """Extra easyconfig parameters specific to Code_Saturne."""
         extra_vars = EasyBlock.extra_options(extra_vars)
-    
+
         # add more custom easyconfig parameters specific to OpenFOAM
         extra_vars.update({
             'debug': [False, "Build the debug version.", CUSTOM],
             'slurm': [False, "Build for the slurm resource manager.", CUSTOM],
         })
-        
+
         return extra_vars
 
 
     def prepare_step(self, *args, **kwargs):
         """Prepare step for Code_Saturne obtained from the repository."""
         super(EB_Code_underscore_Saturne, self).prepare_step(*args, **kwargs)
- 
+
         self.log.info("Running ./sbin/bootstrap ...")
 
         cmd = './sbin/bootstrap'
         (out, _) = run_cmd(cmd, log_all=True, simple=False, log_output=True)
-        
+
         return out
-        
+
 
     def configure_step(self):
         """Configure step for Code_Saturne."""
-        
+
         self.log.info("Configuration step is running...")
-        
+
         # only use the opt flags
         env.setvar("CFLAGS", os.environ['OPTFLAGS'])
         env.setvar("CXXFLAGS", os.environ['OPTFLAGS'])
         env.setvar("FCFLAGS", os.environ['OPTFLAGS'])
-        
+
         cmd = ' '.join([
             './configure',
             '--prefix=' + self.installdir,
             '--without-modules',
             self.cfg['configopts'], 
             ])
-        
+
         if self.cfg['debug']:
             cmd = ' '.join([ cmd, '--enable-debug' ]) 
-        
+
         (out, _) = run_cmd(cmd, log_all=True, simple=False, log_output=True)
 
         return out
-    
+
 
     def build_step(self):
         """ Build step for Code_Saturne."""
@@ -104,7 +104,7 @@ class EB_Code_underscore_Saturne(EasyBlock):
         (out, _) = run_cmd(cmd, log_all=True, simple=False, log_output=True)
 
         return out
-    
+
 
     def install_step(self):
         """ Build step for Code_Saturne."""
@@ -115,27 +115,27 @@ class EB_Code_underscore_Saturne(EasyBlock):
 
         return out
 
-    
+
     def post_install_step(self):
         """Custom post install step for Code_Saturne."""
         super(EB_Code_underscore_Saturne, self).post_install_step()
 
         # create a "etc/code_saturne.cfg" and modify it to match SLURM
         if self.cfg['slurm']:
-            
+
             self.log.info("Running the post-install SLURM step ...")
-            
+
             target_path = os.path.join(self.installdir,'etc/code_saturne.cfg')
             from_path = target_path + '.template'
-    
-            apply_regex_substitutions(from_path, [(r"# batch =", 
-                                                     r"batch = SLURM")])
-            apply_regex_substitutions(from_path, [(r"# mpiexec = mpiexec", 
-                                                     r"mpiexec = srun")])
-            apply_regex_substitutions(from_path, [(r"# mpiexec_n = ' -n '", 
-                                                     r"mpiexec_n = ' -n '")])
-            apply_regex_substitutions(from_path, [(r"# mpiexec_n_per_node =", 
-                                r"mpiexec_n_per_node = ' --ntasks-per-node '")])
+
+            apply_regex_substitutions(from_path,
+                                      [(r"# batch =", r"batch = SLURM")])
+            apply_regex_substitutions(from_path,
+                                      [(r"# mpiexec = mpiexec", r"mpiexec = srun")])
+            apply_regex_substitutions(from_path,
+                                      [(r"# mpiexec_n = ' -n '", r"mpiexec_n = ' -n '")])
+            apply_regex_substitutions(from_path,
+                                      [(r"# mpiexec_n_per_node =", r"mpiexec_n_per_node = ' --ntasks-per-node '")])
 
             os.rename(from_path, target_path)
 
@@ -144,23 +144,22 @@ class EB_Code_underscore_Saturne(EasyBlock):
         """Custom sanity check step for Code_Saturne."""
 
         shlib_ext = get_shared_lib_ext()
-        
+
         custom_paths = {
             'files': ['bin/code_saturne', 'lib/libsaturne.%s' % shlib_ext],
             'dirs': ['bin', 'lib', 'libexec', 'include', 'etc'],
         }
- 
+
         super(EB_Code_underscore_Saturne, self).sanity_check_step(
             custom_paths=custom_paths)
 
 
     def make_module_extra(self, altroot=None, altversion=None):
         """Extra environment variables for Code_Saturne."""
- 
+
         txt = super(EB_Code_underscore_Saturne, self).make_module_extra()
- 
-        cs_bashPath = os.path.join(self.installdir, 'etc', 
-                                   'bash_completion.d', 'code_saturne')
+
+        cs_bashPath = os.path.join(self.installdir, 'etc', 'bash_completion.d', 'code_saturne')
         txt += self.module_generator.set_environment('CS_BASH', cs_bashPath)
- 
+
         return txt
--- a/easybuild/easyblocks/c/code_saturne.py
+++ b/easybuild/easyblocks/c/code_saturne.py
@@ -23,15 +23,16 @@
 # along with EasyBuild. If not, see <http://www.gnu.org/licenses/>.
 ##
 """
-EasyBuild support for building and installing Code_Saturne, implemented as an easyblock
+EasyBuild support for building and installing Code_Saturne,
+implemented as an easyblock.
 
-@author: Metin Cakircali (Juelich Supercomputing Centre)
+@author: Metin Cakircali (Juelich Supercomputing Centre, FZJ GmbH)
 """
 
 import os
 
 from easybuild.framework.easyblock import EasyBlock
-from easybuild.framework.easyconfig import CUSTOM
+from easybuild.framework.easyconfig.default import CUSTOM
 from easybuild.tools.filetools import apply_regex_substitutions
 from easybuild.tools.run import run_cmd
 from easybuild.tools.systemtools import get_shared_lib_ext
@@ -47,10 +48,10 @@ class EB_Code_underscore_Saturne(EasyBlock):
         """Extra easyconfig parameters specific to Code_Saturne."""
         extra_vars = EasyBlock.extra_options(extra_vars)
 
-        # add more custom easyconfig parameters specific to OpenFOAM
+        # add some custom easyconfig parameters specific to Code_Saturne
         extra_vars.update({
             'debug': [False, "Build the debug version.", CUSTOM],
-            'SLURM': [False, "Build for the slurm resource manager.", CUSTOM],
+            'slurm': [False, "Build for the slurm workload manager.", CUSTOM],
         })
 
         return extra_vars
@@ -67,7 +68,7 @@ class EB_Code_underscore_Saturne(EasyBlock):
 
         self.log.info("Running ./sbin/bootstrap ...")
         cmd = './sbin/bootstrap'
-        (out, _) = run_cmd(cmd, log_all=True, simple=False, log_output=True)
+        run_cmd(cmd, log_all=True, simple=True, log_output=True)
 
         cmd = ' '.join([
             './configure',
@@ -111,22 +112,32 @@ class EB_Code_underscore_Saturne(EasyBlock):
         """Custom post install step for Code_Saturne."""
         super(EB_Code_underscore_Saturne, self).post_install_step()
 
-        # create a "etc/code_saturne.cfg" and modify it to match SLURM
-        if self.cfg['SLURM']:
+        # if we are using a SLURM system, we need to change the defaults;
+        # so, modify the template and rename it to "etc/code_saturne.cfg"
+        if self.cfg['slurm']:
 
             self.log.info("Running the post-install SLURM step ...")
 
             target_path = os.path.join(self.installdir, 'etc/code_saturne.cfg')
             from_path = target_path + '.template'
 
-            apply_regex_substitutions(from_path,
-                                      [(r"# batch =", r"batch = SLURM")])
-            apply_regex_substitutions(from_path,
-                                      [(r"# mpiexec = mpiexec", r"mpiexec = srun")])
-            apply_regex_substitutions(from_path,
-                                      [(r"# mpiexec_n = ' -n '", r"mpiexec_n = ' -n '")])
-            apply_regex_substitutions(from_path,
-                                      [(r"# mpiexec_n_per_node =", r"mpiexec_n_per_node = ' --ntasks-per-node '")])
+            apply_regex_substitutions(
+                from_path,
+                [(r"# batch =", r"batch = SLURM")]
+                )
+            apply_regex_substitutions(
+                from_path,
+                [(r"# mpiexec = mpiexec", r"mpiexec = srun")]
+                )
+            apply_regex_substitutions(
+                from_path,
+                [(r"# mpiexec_n = ' -n '", r"mpiexec_n = ' -n '")]
+                )
+            apply_regex_substitutions(
+                from_path,
+                [(r"# mpiexec_n_per_node =",
+                  r"mpiexec_n_per_node = ' --ntasks-per-node '")]
+                )
 
             os.rename(from_path, target_path)
 
@@ -143,12 +154,13 @@ class EB_Code_underscore_Saturne(EasyBlock):
         super(EB_Code_underscore_Saturne, self).sanity_check_step(
             custom_paths=custom_paths)
 
-    def make_module_extra(self, altroot=None, altversion=None):
+    def make_module_extra(self):
         """Extra environment variables for Code_Saturne."""
 
         txt = super(EB_Code_underscore_Saturne, self).make_module_extra()
 
-        cs_bash_path = os.path.join(self.installdir, 'etc', 'bash_completion.d', 'code_saturne')
+        cs_bash_path = os.path.join(self.installdir, 'etc',
+                                    'bash_completion.d', 'code_saturne')
         txt += self.module_generator.set_environment('CS_BASH', cs_bash_path)
 
         return txt
--- a/easybuild/easyblocks/c/code_saturne.py
+++ b/easybuild/easyblocks/c/code_saturne.py
@@ -54,7 +54,6 @@ class EB_Code_underscore_Saturne(EasyBlock):
 
         return extra_vars
 
-
     def prepare_step(self, *args, **kwargs):
         """Prepare step for Code_Saturne obtained from the repository."""
         super(EB_Code_underscore_Saturne, self).prepare_step(*args, **kwargs)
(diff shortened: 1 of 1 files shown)


mcakircali:
./sbin/bootstrap is part of the Code_Saturne source...

I can move this run_cmd to the beginning of configure step but it would be good to receive more info about the idea of prepare_step..

easybuild/easyblocks/c/code_saturne.py
```diff
@@ -0,0 +1,158 @@
...
+        return extra_vars
+
+    def prepare_step(self, *args, **kwargs):
+        """Prepare step for Code_Saturne obtained from the repository."""
+        super(EB_Code_underscore_Saturne, self).prepare_step(*args, **kwargs)
+
+        self.log.info("Running ./sbin/bootstrap ...")
+
+        cmd = './sbin/bootstrap'
+        (out, _) = run_cmd(cmd, log_all=True, simple=False, log_output=True)
```
boegel: @mcakircali The tests fail right now because `prepare_step` is also run when `eb --module-only` is used, but the build directory is not set up under `--module-only` (so there's no unpacking, etc.).

You can fix this in two ways, up to you which makes more sense to you:

* move running `./sbin/bootstrap` to `configure_step` (which is skipped under `--module-only`)
* only conditionally run `./sbin/bootstrap`, e.g. when the file is there, or when `--module-only` is not used (you can check the latter with `if build_option('module_only'):`)

easybuild/easyblocks/c/code_saturne.py
```diff
@@ -0,0 +1,158 @@
...
+
+    @staticmethod
+    def extra_options(extra_vars=None):
+        """Extra easyconfig parameters specific to Code_Saturne."""
+        extra_vars = EasyBlock.extra_options(extra_vars)
+
+        # add more custom easyconfig parameters specific to OpenFOAM
+        extra_vars.update({
+            'debug': [False, "Build the debug version.", CUSTOM],
+            'slurm': [False, "Build for the slurm resource manager.", CUSTOM],
```
damianam: The default doesn't create a valid `.cfg` file. Also, using slurm doesn't automatically warrant the use of `srun`. I suggest to remove this variable but replace it with 
```
            'batch': ['SLURM', "Build for the slurm resource manager/scheduler.", CUSTOM],
            'mpiexec': ['srun', "Command to run MPI jobs", CUSTOM],
            'mpiexec_n': ['-n', "Flag to specify number of MPI tasks", CUSTOM],
            'mpiexec_n_per_node': ['--ntasks-per-node', "Flag to specify number of MPI tasks per node.", CUSTOM],
```

easybuild/easyblocks/c/code_saturne.py
damianam: I like option 2

easybuild/easyblocks/c/code_saturne.py
```diff
@@ -0,0 +1,158 @@
...
+        """Configure step for Code_Saturne."""
+
+        self.log.info("Configuration step is running...")
+
+        # only use the opt flags
+        env.setvar("CFLAGS", os.environ['OPTFLAGS'])
+        env.setvar("CXXFLAGS", os.environ['OPTFLAGS'])
+        env.setvar("FCFLAGS", os.environ['OPTFLAGS'])
+
+        cmd = ' '.join([
```
damianam: This is missing `preconfigopts`

easybuild/easyblocks/c/code_saturne.py
```diff
@@ -0,0 +1,158 @@
...
+        return out
+
+    def build_step(self):
+        """ Build step for Code_Saturne."""
+
+        paracmd = ''
+        if self.cfg['parallel']:
+            paracmd = "-j %s" % self.cfg['parallel']
+
+        cmd = "make %s" % paracmd
```
damianam: This is missing `prebuildopts` and `buildopts`

easybuild/easyblocks/c/code_saturne.py
```diff
@@ -0,0 +1,158 @@
...
+
+        (out, _) = run_cmd(cmd, log_all=True, simple=False, log_output=True)
+
+        return out
+
+    def install_step(self):
+        """ Build step for Code_Saturne."""
+
+        cmd = "make install"
```
damianam: This is missing `preinstallopts` and `installopts`

easybuild/easyblocks/c/code_saturne.py
```diff
@@ -0,0 +1,158 @@
...
+        (out, _) = run_cmd(cmd, log_all=True, simple=False, log_output=True)
+
+        return out
+
+    def post_install_step(self):
+        """Custom post install step for Code_Saturne."""
+        super(EB_Code_underscore_Saturne, self).post_install_step()
+
+        # create a "etc/code_saturne.cfg" and modify it to match SLURM
+        if self.cfg['slurm']:
```
damianam: If the extra parameters are added, this should be adjusted accordingly

easybuild/easyblocks/c/code_saturne.py
```diff
@@ -0,0 +1,158 @@
...
+
+        super(EB_Code_underscore_Saturne, self).sanity_check_step(
+            custom_paths=custom_paths)
+
+    def make_module_extra(self, altroot=None, altversion=None):
+        """Extra environment variables for Code_Saturne."""
+
+        txt = super(EB_Code_underscore_Saturne, self).make_module_extra()
+
+        cs_bashPath = os.path.join(self.installdir, 'etc', 'bash_completion.d', 'code_saturne')
```
damianam: I am not sure this variable name follows any particular code style. Change it maybe to `cs_bash_path`?

easybuild/easyblocks/c/code_saturne.py
mcakircali: thx for the example. I want to keep it simple. Of course, should be flexible enough too.
Is there a situation/system that srun doesn't work with SLURM?

easybuild/easyblocks/c/code_saturne.py
mcakircali: @boegel thanks!
I like option 1, as it's more logical. there's no meaning to do bootstrap step for --module-only.. also option 2 sounds like a workaround that complicates things.

easybuild/easyblocks/c/code_saturne.py
mcakircali: sure, looks better.

easybuild/easyblocks/c/code_saturne.py
mcakircali: The preconfigopts for this case is bootstrap command, which will be here when I move it from prepare step to configure step (option 1 #discussion_r234549524). It's hardcoded because it's not optional. I don't see any other use for preconfigopts in this case.

easybuild/easyblocks/c/code_saturne.py
damianam: Well, I don't know every system in the world, but using `srun` is not a requirement to use SLURM. LRZ for instance has mpiexec in their documentation: https://www.lrz.de/services/compute/linux-cluster/batch_parallel/

easybuild/easyblocks/c/code_saturne.py
damianam: Hardcoding the bootstrap is fine IMO. But `preconfigopts` is a standard easyconfig parameter. See https://easybuild.readthedocs.io/en/latest/version-specific/easyconfig_parameters.html Without it you are breaking something that is expected to be there. This is frequently used to change environment or do other hacks before configuring, so I think it is positive to have it, even though using it would be rare.

easybuild/easyblocks/c/code_saturne.py
mcakircali: I know it's not requirement, but it doesn't mean that srun wouldn't work.. They also refer to srun in their documentation. I don't see a reason to complicate, that's all.

damianam:
lgtm. @boegel could you give it another look?

easybuild/easyblocks/c/code_saturne.py
```diff
@@ -0,0 +1,167 @@
...
+
+    def configure_step(self):
+        """Configure step for Code_Saturne."""
+
+        self.log.info("Configuration step is running...")
+
+        # only use the opt flags
+        env.setvar("CFLAGS", os.environ['OPTFLAGS'])
+        env.setvar("CXXFLAGS", os.environ['OPTFLAGS'])
+        env.setvar("FCFLAGS", os.environ['OPTFLAGS'])
```
boegel: @mcakircali Can you clarify what this is about? This is counterintuitive w.r.t. what EasyBuild usually does.

easybuild/easyblocks/c/code_saturne.py
```diff
@@ -0,0 +1,167 @@
...
+
+        self.log.info("Running ./sbin/bootstrap ...")
+        cmd = './sbin/bootstrap'
+        run_cmd(cmd, log_all=True, simple=True, log_output=True)
+
+        cmd = ' '.join([
+            self.cfg['preconfigopts'],
+            './configure',
+            '--prefix=' + self.installdir,
+            '--without-modules',
```
boegel: What does `--without-modules` imply exactly?

Can we make this a bit more flexible, by letting people use `with_modules = True` in the easyconfig for example (with the default being `False`)?

easybuild/easyblocks/c/code_saturne.py
```diff
@@ -0,0 +1,167 @@
...
+        paracmd = ''
+        if self.cfg['parallel']:
+            paracmd = "-j %s" % self.cfg['parallel']
+
+        cmd = "%s make %s %s" % (self.cfg['prebuildopts'],
+                                 paracmd, self.cfg['buildopts'])
+
+        (out, _) = run_cmd(cmd, log_all=True, simple=False, log_output=True)
+
+        return out
```
boegel: This is basically duplicating what `ConfigureMake` does, so why not derive `EB_Code_underscore_Saturne` from `ConfigureMake` instead?

easybuild/easyblocks/c/code_saturne.py
```diff
@@ -0,0 +1,167 @@
...
+
+    def install_step(self):
+        """ Install step for Code_Saturne."""
+
+        cmd = "%s make install %s" % (self.cfg['preinstallopts'],
+                                      self.cfg['installopts'])
+
+        (out, _) = run_cmd(cmd, log_all=True, simple=False, log_output=True)
+
+        return out
```
boegel: Same here, another reason to just derive from `ConfigureMake` and not define `install_step` here?

easybuild/easyblocks/c/code_saturne.py
```diff
@@ -0,0 +1,167 @@
...
+                )
+            apply_regex_substitutions(
+                from_path,
+                [(r"# mpiexec_n = ' -n '", r"mpiexec_n = ' -n '")]
+                )
+            apply_regex_substitutions(
+                from_path,
+                [(r"# mpiexec_n_per_node =",
+                  r"mpiexec_n_per_node = ' --ntasks-per-node '")]
+                )
```
boegel: The multi-line formatting is a bit over the top here imho...

Also, why not make this a single call to `apply_regex_substitutions`?

```python
regex_subs = [
    (r"# batch =", r"batch = SLURM")],
    (r"# mpiexec = mpiexec", r"mpiexec = srun"),
    (r"# mpiexec_n = ' -n '", r"mpiexec_n = ' -n '"),
    (r"# mpiexec_n_per_node =", r"mpiexec_n_per_node = ' --ntasks-per-node '"),
]
apply_regex_substitutions(from_path, regex_subs)

File: easybuild/easyblocks/c/code_saturne.py [hunk 13]
```diff
@@ -0,0 +1,167 @@
...
+
+        shlib_ext = get_shared_lib_ext()
+
+        custom_paths = {
+            'files': ['bin/code_saturne', 'lib/libsaturne.%s' % shlib_ext],
+            'dirs': ['bin', 'lib', 'libexec', 'include', 'etc'],
+        }
+
+        super(EB_Code_underscore_Saturne, self).sanity_check_step(
+            custom_paths=custom_paths)
```
boegel: No need to wrap at all here?

```python
super(EB_Code_underscore_Saturne, self).sanity_check_step(custom_paths=custom_paths)

File: easybuild/easyblocks/c/code_saturne.py [hunk 14]
```diff
@@ -0,0 +1,167 @@
...
+
+    def make_module_extra(self):
+        """Extra environment variables for Code_Saturne."""
+
+        txt = super(EB_Code_underscore_Saturne, self).make_module_extra()
+
+        cs_bash_path = os.path.join(self.installdir, 'etc',
+                                    'bash_completion.d', 'code_saturne')
```
boegel: Same here, no need to wrap (max line length is 120 columns in our code style, cfr. https://easybuild.readthedocs.io/en/latest/Code_style.html)

```python
cs_bash_path = os.path.join(self.installdir, 'etc', 'bash_completion.d', 'code_saturne')

File: easybuild/easyblocks/c/code_saturne.py [hunk 15]
```diff
@@ -0,0 +1,167 @@
...
+                from_path,
+                [(r"# mpiexec_n = ' -n '", r"mpiexec_n = ' -n '")]
+                )
+            apply_regex_substitutions(
+                from_path,
+                [(r"# mpiexec_n_per_node =",
+                  r"mpiexec_n_per_node = ' --ntasks-per-node '")]
+                )
+
+            os.rename(from_path, target_path)
```
boegel: Please use the `move_file` function which is available from `easybuild.tools.filetools` (cfr. https://easybuild.readthedocs.io/en/latest/api/easybuild.tools.filetools.html#easybuild.tools.filetools.move_file)

easybuild/easyblocks/c/code_saturne.py
mcakircali: In this case, EB flags should add on top of default Code_Saturne flags except the optarch.

mcakircali: * correction: "should not"

easybuild/easyblocks/c/code_saturne.py
mcakircali: Code_Saturne reads the "loaded" modules at the time of build and loads them later when you run the software. This behavior is same as EB. I don't see why collide 2 tools to manage module loads at the same time. This should stay as is.

easybuild/easyblocks/c/code_saturne.py
mcakircali: I based on ConfigureMake but simplified it. The reason for a new block is basically the EB CFLAGS etc. We don't need this block, if there is a way to suppress those via config?

easybuild/easyblocks/c/code_saturne.py
mcakircali: Good to know. I somehow missed the note on your style page.

easybuild/easyblocks/c/code_saturne.py
mcakircali: This is renaming. The file is kept in the same directory. Can you give an example how to use move_file to use as rename?

easybuild/easyblocks/c/code_saturne.py
mcakircali: Sorry, I don't get the idea. Can you please give an example how to derive?
just remove the def?

easybuild/easyblocks/c/code_saturne.py
mcakircali: Yes, this def is also same. Sorry, I need an example to know how to derive?

easybuild/easyblocks/c/code_saturne.py
mcakircali: Maybe there is a better way to control and suppress the EB flags. I admit that it's not flexible if you want to play with the flags, but that shouldn't be played around in these kind of solvers because they come with their own flags tested by developers.

damianam: What other flags are you concerned about? Maybe these can be changed in the `toolchainopts` of the easyconfig instead? If so, looks like a better approach to me. If not, we'll have to evaluate how to do this.

easybuild/easyblocks/c/code_saturne.py
damianam: What you say is true and makes sense if you think about compilers and MPIs in a hierarchy, but in a flat module naming scheme things are different. Loading the `Code_Saturne` module would load the dependencies, but one can swap the compiler and the MPI runtime without unloading `Code_Saturne`. Similarly, even in a hierarchy, the same can happen with other dependencies. If you have 2 versions of a dependency, you can load `Code_Saturne` and then swap the dependency, and potentially run into issues. Maybe it is not a bad idea to enable the use of module detection here.

easybuild/easyblocks/c/code_saturne.py
damianam: You need to swap this line:

`class EB_Code_underscore_Saturne(EasyBlock):`

With 

`class EB_Code_underscore_Saturne(ConfigureMake):`

and remove the unnecessary methods.

easybuild/easyblocks/c/code_saturne.py
mcakircali: What kind of issue?
how do you hide EB module and developers modules from Code_Saturne during build?

easybuild/easyblocks/c/code_saturne.py
damianam: `move_file(from_path, target_path)` should work

easybuild/easyblocks/c/code_saturne.py
damianam: You can't hide those modules in an easy way, but not everyone sets up the installation environment using modules like we do.

> What kind of issue?

The kind of issue that `Code_Saturne` tries to avoid by saving the list of modules and using them at run time I suppose. Incompatible libraries, options missing, etc.

easybuild/easyblocks/c/code_saturne.py
damianam: We can talk next week face-to-face for sure, but in any case we should report back here so there is a trace of why things are done in a particular way

easybuild/easyblocks/c/code_saturne.py
mcakircali: > You can't hide those modules in an easy way, but not everyone sets up the installation environment using modules like we do.

So, no solution for us?

> 
> > What kind of issue?
> 
> The kind of issue that `Code_Saturne` tries to avoid by saving the list of modules and using them at run time I suppose. Incompatible libraries, options missing, etc.

It comes to the same point; --with-modules is for setups without EB. IMO, If things mess up by swapping dependencies, then it should be addressed in EB not in software or any workaround in block files.

easybuild/easyblocks/c/code_saturne.py
mcakircali: of course we will document here

easybuild/easyblocks/c/code_saturne.py
damianam: > So, no solution for us?

Sure. As I see things, the solution for us is to use `--without-modules`. That would work fine in our setup.

> It comes to the same point; --with-modules is for setups without EB

No, it's not. `--with-modules` is for setups that:

1. Do not depend on modules to set up the general software installation environment
2. Have a flat module naming scheme

You can have both things in EB setups

mcakircali: @damianam maybe you don't know what Code_Saturne does. Again, "--without-modules works" for every setup. If you have flat naming setup, then you have a problem but I suggest a useful feature that EB does not allow one to swap dependencies on such setups. Removing this option and setting it in config is just unnecessary and cumbersome approach and you can't inline with mainstream or push such configs to mainstream. We already have such a diverse setup, better minimize it.

easybuild/easyblocks/c/code_saturne.py
```diff
@@ -0,0 +1,167 @@
+##
+# Copyright 2009-2018 Ghent University
```
wpoely86: we're at 2019 already :wink:

easybuild/easyblocks/c/code_saturne.py
```diff
@@ -0,0 +1,167 @@
...
+    """Support for building and installing Code_Saturne."""
+
+    @staticmethod
+    def extra_options(extra_vars=None):
+        """Extra easyconfig parameters specific to Code_Saturne."""
+        extra_vars = EasyBlock.extra_options(extra_vars)
+
+        # add some custom easyconfig parameters specific to Code_Saturne
+        extra_vars.update({
+            'debug': [False, "Build the debug version.", CUSTOM],
```
wpoely86: we have a `toolchainsopts = { 'debug': True }`. Can you use that?

easybuild/easyblocks/c/code_saturne.py
```diff
@@ -0,0 +1,167 @@
...
+            'slurm': [False, "Build for the slurm workload manager.", CUSTOM],
+        })
+
+        return extra_vars
+
+    def configure_step(self):
+        """Configure step for Code_Saturne."""
+
+        self.log.info("Configuration step is running...")
```
wpoely86: Don't we already have enough stuff printed? I'm not sure if this has any value?

easybuild/easyblocks/c/code_saturne.py
wpoely86: So what's the conclusion? Cause this indeed seems pretty weird.

easybuild/easyblocks/c/code_saturne.py
```diff
@@ -0,0 +1,167 @@
...
+
+        # only use the opt flags
+        env.setvar("CFLAGS", os.environ['OPTFLAGS'])
+        env.setvar("CXXFLAGS", os.environ['OPTFLAGS'])
+        env.setvar("FCFLAGS", os.environ['OPTFLAGS'])
+
+        self.log.info("Running ./sbin/bootstrap ...")
```
wpoely86: `run_cmd` will ready logs this?

easybuild/easyblocks/c/code_saturne.py
wpoely86: As far as I understand the discussion, `--without-modules` makes sense? EB handles this, Code_Saturne should not interfere?

easybuild/easyblocks/c/code_saturne.py
wpoely86: +1 on that

easybuild/easyblocks/c/code_saturne.py
akesandgren: Please don't confuse EB with the actual module system in use.
EB just builds software and creates modules in one of several module naming systems.
What the actual module system does is not EB's problem.
Lmod for instance can be setup to refuse swapping altogether or allow it.

boegel:
Looks like we never got to a point here where the easyblock was eligible for merging...

On top of that, the only `Code_Saturne` easyconfig file that's available (in https://github.com/easybuilders/easybuild-easyconfigs/pull/7170) is used a toolchain is now deprecated.

So, I'll go ahead and close this. If support for Code_Saturne would still be useful, I suggest we start afresh in a new PR.

PR #1592, opened 2018-11-21
ENH: some tweaks for WRFv4 and WPSv4

andreas-h:
* optionally build WRF with PnetCDF support
* fix netCDF-Fortran dependency problems in WPS

### DIFF
--- a/easybuild/easyblocks/w/wps.py
+++ b/easybuild/easyblocks/w/wps.py
@@ -100,10 +100,6 @@ class EB_WPS(EasyBlock):
         # netCDF dependency check + setting env vars (NETCDF, NETCDFF)
         set_netcdf_env_vars(self.log)
 
-        netcdff = get_software_root('netCDF-Fortran')
-        if netcdff:
-            env.setvar('NETCDFF', netcdff)
-
         # WRF dependency check
         wrf = get_software_root('WRF')
         if wrf:
--- a/easybuild/easyblocks/w/wps.py
+++ b/easybuild/easyblocks/w/wps.py
@@ -100,6 +100,10 @@ class EB_WPS(EasyBlock):
         # netCDF dependency check + setting env vars (NETCDF, NETCDFF)
         set_netcdf_env_vars(self.log)
 
+        netcdff = get_software_root('netCDF-Fortran')
+        if netcdff:
+            env.setvar('NETCDFF', netcdff)
+
         # WRF dependency check
         wrf = get_software_root('WRF')
         if wrf:
--- a/easybuild/easyblocks/w/wrf.py
+++ b/easybuild/easyblocks/w/wrf.py
@@ -106,10 +106,10 @@ class EB_WRF(EasyBlock):
                 if not os.path.exists(os.path.join(hdf5, 'bin', f)):
                     parallel_hdf5 = False
                     break
-            if not (hdf5 or parallel_hdf5):
-                raise EasyBuildError("Parallel HDF5 module not loaded?")
-            else:
+            if parallel_hdf5:
                 env.setvar('PHDF5', hdf5)
+            else:
+                self.log.info("Parallel HDF5 module not loaded, assuming that's OK...")
         else:
             self.log.info("HDF5 module not loaded, assuming that's OK...")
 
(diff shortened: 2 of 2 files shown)


damianam:
@andreas-h the PR lgtm, but I'd like to have in the comments a reference to why these variables are needed. Do you think you can add that?

easybuild/easyblocks/w/wrf.py
```diff
@@ -112,6 +113,11 @@ def configure_step(self):
         else:
```
andreas-h: @boegel I don't really understand this part of the code.  It seems to me the `if not (hdf5 or parallel_hdf5)` can never be triggered, since we're already inside the `if hdf5` which means that `(hdf5 or parallel_hdf5)` will always be `True` which means that the `if not` will never be triggered. I'm aware that this piece of the code is years old, but do you happen to remember why this was written that way?

andreas-h:
1. WPS: I wasn't able to figure out how to tell WPS the correct path to netCDF-Fortran otherwise. Build was working properly using intel-2018b, but with foss I was running into issues where netcdff wasn't found. This was the only way I could fix that.
2. the `HDF5` env variable probably isn't strictly needed, but `./configure` was complaining about it not being set:

      > HDF5 not set in environment. Will configure WRF for use without.

3. PnetCDF is an optional dependency. Setting the `PNETCDF` env variable enables output using ParallelNetCDF.

4. `EM_CORE` and `NMM_CORE` variables control which dynamical model core is being built.  You can only choose one of them.  If nothing is explicitly set, EM_CORE is being built.
I agree that optimally, one would make this configurable and add (yet) another version suffix to indicate the choice.However, I have no experience at all with NMM_CORE and so didn't want to do that.  I'm working on a PR to optionally build the WRF-Chem chemistry code inside the wrf EasyBlock and that one only works with EM_CORE anyways.

Now that I think about it, maybe it'd be best to *not* set the `_CORE` variables explicitly here.  I would then only do that in the wrf-chem stuff I'm developing, since there it's not optional any more. Apart from that I don't think that any additional comments help.  I can add some if you think it's needed.

easybuild/easyblocks/w/wrf.py
akesandgren: Yeah, that line is wrong, it should be (note the change from error to info)
```
if not parallel_hdf5:
  self.log.info("Parallel HDF5 module not loaded, assuming that's OK...")
else:
  ...
```
This is assuming that the PHDF5 variable is still relevant for WRF.

andreas-h: there's a library `external/io_phdf5` which is still part of WRF.  So I'm assuming it's still relevant.

I'll fix this one later whem I'm at work and update this PR.

easybuild/easyblocks/w/wps.py
```diff
@@ -100,6 +100,10 @@ def configure_step(self):
         # netCDF dependency check + setting env vars (NETCDF, NETCDFF)
         set_netcdf_env_vars(self.log)
 
+        netcdff = get_software_root('netCDF-Fortran')
+        if netcdff:
+            env.setvar('NETCDFF', netcdff)
```
damianam: Is this needed at all? It is supposed to be set in https://github.com/easybuilders/easybuild-easyblocks/blob/master/easybuild/easyblocks/n/netcdf.py#L132 already. I don't know why you need that in your case, the code looks the same to me

andreas-h: you're right, just tested this and updated the PR without it.

easybuild/easyblocks/w/wrf.py
```diff
@@ -98,20 +98,26 @@ def configure_step(self):
...
-            else:
+            if parallel_hdf5:
                 env.setvar('PHDF5', hdf5)
+            else:
+                self.log.info("Parallel HDF5 module not loaded, assuming that's OK...")
         else:
             self.log.info("HDF5 module not loaded, assuming that's OK...")
 
+        # Parallel netCDF (optional) dependency
+        pnetcdf = get_software_root('PnetCDF')
```
boegel: @andreas-h We don't have any easyconfigs for `PnetCDF` yet, I think... Is this an entirely separate package from `netCDF`, or does it just require `netCDF` to be installed with specific configure options?

andreas-h:
It is a separate package, and there are already easyconfigs in develop. I'll submit a PR for the latest version later today.

PR #1593, opened 2018-11-22
PR with the AVL EasyBlock submitted by chrissamuel in issue 1589

damianam:


chrissamuel:
Hi @damianam, thanks for doing this!

chrissamuel:
@damianam yeah, it's a commercial app that requires a license (FlexLM etc). :-(

All I can report is that it successfully installed it for us at Swinburne and our user is now using the application.

chrissamuel:
Added a :+1: review in case that helps.

easybuild/easyblocks/a/avl.py
```diff
@@ -0,0 +1,83 @@
...
+        # Get the port and server
+        licport, licserv = licserv.split('@')
+
+        # create license file
+        lictxt = '\n'.join([
+            "SERVER %s 000000000000 %s" % (licserv, licport),
+            "USE_SERVER",
+        ])
+
+        licfile = os.path.join(self.installdir, 'etc/lmx/license.dat')
```
akesandgren: This should be
```
licfile = os.path.join(self.installdir, 'etc', 'lmx', 'license.dat')
```

easybuild/easyblocks/a/avl.py
```diff
@@ -0,0 +1,83 @@
...
+        write_file(licfile, lictxt)
+
+    def sanity_check_step(self):
+        """Custom sanity check for AVL."""
+        custom_paths = {
+            'files': ['FIRE/bin/fire_launcher.py', 'bin/fire_wm', 'bin/diagnose'],
```
akesandgren: And strictly speaking, these should also use os.path.join, but it gets a bit cluttered if doing so.

PR #1594, opened 2018-11-23
avoid hardcoding 'PREFIX=<installdir>' in build/install options when using SCons easyblock

boegel:


bartoldeman:
lgtm, can be merged once travis is happy

PR #1595, opened 2018-11-27
add custom easyblock for Blender

smoors:


easybuild/easyblocks/b/blender.py
```diff
@@ -0,0 +1,97 @@
...
+"""
+EasyBuild support for Blender, implemented as an easyblock
+
+@author: Stijn De Weirdt (Ghent University)
+@author: Dries Verdegem (Ghent University)
+@author: Kenneth Hoste (Ghent University)
+@author: Pieter De Baets (Ghent University)
+@author: Jens Timmerman (Ghent University)
+@author: Ward Poelmans (Ghent University)
+@author: Samuel Moors (Vrije Universiteit Brussel)
```
boegel: @smoors You're the only author (for now), so please trim this down to just yourself.

easybuild/easyblocks/b/blender.py
```diff
@@ -0,0 +1,97 @@
...
+        self.cfg.update('configopts', '-DWITH_BUILDINFO=OFF')
+
+        # disable SSE detection to give EasyBuild full control over optimization compiler flags being used
+        self.cfg.update('configopts', '-DWITH_CPU_SSE=OFF')
+        self.cfg.update('configopts', '-DCMAKE_C_FLAGS_RELEASE="-DNDEBUG"')
+        self.cfg.update('configopts', '-DCMAKE_CXX_FLAGS_RELEASE="-DNDEBUG"')
+
+        # these are needed until extra dependencies are added for them to work
+        self.cfg.update('configopts', '-DWITH_GAMEENGINE=OFF')
+        self.cfg.update('configopts', '-DWITH_SYSTEM_GLEW=OFF')
```
boegel: @smoors Ideally, we avoid hardcoding these in the easyblock. It's OK to do this in an easyconfig file, since if you don't want these you can easily get around it by tweaking the easyconfig.

You could do this with constructions like this, for example:

```python
if '-DWITH_SYSTEM_GLEW=' not in self.cfg['configopts']:
    self.cfg.update('configopts', '-DWITH_SYSTEM_GLEW=OFF')
```

Even better, use a loop for all of these, over a map, something like this (untested):

```python
default_config_opts = {
    'WITH_INSTALL_PORTABLE': 'OFF',
    ...
}
for key in sorted(default_config_opts):
    opt = '-D%s=' % key
    if opt not in self.cfg['configopts']:
        self.cfg.update('configopts', opt + default_config_opts[key])

File: easybuild/easyblocks/b/blender.py [hunk 3]
```diff
@@ -0,0 +1,97 @@
...
+
+        # Python paths
+        python_root = get_software_root('Python')
```
boegel: @smoors Make sure `Python` is actually included as a dependency (and if not, skip this configuration part):

```python
if python_root:
    pyshortver = '.'.join(get_software_version('Python').split('.')[:2])
    ...

File: easybuild/easyblocks/b/blender.py [hunk 4]
```diff
@@ -0,0 +1,97 @@
...
+        pyshortver = '.'.join(get_software_version('Python').split('.')[:2])
+        site_packages = os.path.join(python_root, 'lib', 'python%s' % pyshortver, 'site-packages')
+        numpy_root = glob.glob(os.path.join(site_packages, 'numpy*'))[0]
+        requests_root = glob.glob(os.path.join(site_packages, 'requests*'))[0]
```
boegel: Make sure you only have 1 hit here, to avoid using the wrong thing?

easybuild/easyblocks/b/blender.py
```diff
@@ -0,0 +1,97 @@
...
+        shlib_ext = get_shared_lib_ext()
+        self.cfg.update('configopts', '-DPYTHON_VERSION=%s' % pyshortver)
+        self.cfg.update('configopts', '-DPYTHON_LIBRARY=%s/lib/libpython%sm.%s' % (python_root, pyshortver, shlib_ext))
```
boegel: use `os.path.join` here please

```python
python_lib = os.path.join(python_root, 'lib', 'libpython%sm.%s' % (pyshortver, shlib_ext)
self.cfg.update('configopts', '-DPYTHON_LIBRARY=' + python_lib)
```

The `m` here is only correct when Python 3 is used, so maybe a `glob` on `libpython*.so` is better?

easybuild/easyblocks/b/blender.py
```diff
@@ -0,0 +1,97 @@
...
+        self.cfg.update('configopts', '-DPYTHON_INCLUDE_DIR=%s/include/python%sm' % (python_root, pyshortver))
```
boegel: Same here, please use `os.path.join`

easybuild/easyblocks/b/blender.py
```diff
@@ -0,0 +1,97 @@
...
+        self.cfg.update('configopts', '-DPYTHON_NUMPY_PATH=%s' % numpy_root)
+        self.cfg.update('configopts', '-DPYTHON_REQUESTS_PATH=%s' % requests_root)
+
+        # OpenEXR
+        self.cfg.update('configopts', '-DOPENEXR_INCLUDE_DIR=$EBROOTOPENEXR/include')
```
boegel: Only if `OpenEXR` is included as a dependency? And avoid using `$EBROOT*` environment variables, grab the actual value using `get_software_root`:

```python
openexr_root = get_software_root('OpenEXR')
if openexr_root:
    self.cfg.update('configopts', '-DOPENEXR_INCLUDE_DIR=%s' % os.path.join(openexr_root, 'include'))

File: easybuild/easyblocks/b/blender.py [hunk 8]
```diff
@@ -0,0 +1,101 @@
...
+        # Python paths
+        python_root = get_software_root('Python')
+        if python_root:
+            shlib_ext = get_shared_lib_ext()
+            pyshortver = '.'.join(get_software_version('Python').split('.')[:2])
+            site_packages = os.path.join(python_root, 'lib', 'python%s' % pyshortver, 'site-packages')
+            numpy_root = glob.glob(os.path.join(site_packages, 'numpy-*-py%s-linux-x86_64.egg' % pyshortver))[0]
+            requests_root = glob.glob(os.path.join(site_packages, 'requests-*-py%s.egg' % pyshortver))[0]
+            python_lib = glob.glob(os.path.join(python_root, 'lib', 'libpython%s*.%s' % (pyshortver, shlib_ext)))[0]
+            python_include_dir = glob.glob(os.path.join(python_root, 'include', 'python%s*' % pyshortver))[0]
```
boegel: @smoors You're blindly using the 1st value here, but what if there are multiple results returned (or none)? Please make sure you only have a single result, and fail otherwise, something like:

```python
glob_pattern = os.path.join(python_root, 'include', 'python%s*' % pyshortver)
res = glob.glob(glob_pattern)
if len(res) == 1:
    python_include_dir = res[0]
else:
    raise EasyBuildError("Was expected exactly one match for '%s', found %d: %s", glob_pattern, len(res), res)
```

Ideally, introduce a function for this that takes the glob pattern and returns the result (or fails), since you need it in a bunch of places...

```python
python_include_dir = find_dir_for_glob_pattern(os.path.join(python_root, 'include', 'python%s*' % pyshortver))

Comment (smoors) [reply easybuild/easyblocks/b/blender.py]: ok, I understand what you want now. I (perhaps wrongly) assumed it is reasonable to expect there cannot be more than 1 value, and if there is none it will fail anyway.
your suggestion is indeed safer and will give a more clear error.

File: easybuild/easyblocks/b/blender.py [hunk 9]
```diff
@@ -0,0 +1,111 @@
+##
+# Copyright 2009-2018 Ghent University
```
wpoely86: bump to 2019

easybuild/easyblocks/b/blender.py
```diff
@@ -0,0 +1,111 @@
...
+from easybuild.easyblocks.generic.cmakemake import CMakeMake
+from easybuild.tools.build_log import EasyBuildError
+from easybuild.tools.modules import get_software_root, get_software_version
+from easybuild.tools.systemtools import get_shared_lib_ext
+
+
+def find_glob_pattern(glob_pattern):
+    res = glob.glob(glob_pattern)
+    if len(res) != 1:
+        raise EasyBuildError("Was expecting exactly one match for '%s', found %d: %s", glob_pattern, len(res), res)
```
wpoely86: you will need to add a rule for a dry-run here. Else the tests will fail.

```python
if self.dry_run:
  return glob_pattern
```

In the beginning? Then it will return something and dry-run results will still make sense?

easybuild/easyblocks/b/blender.py
```diff
@@ -0,0 +1,113 @@
...
+            raise EasyBuildError("Was expecting exactly one match for '%s', found %d: %s", glob_pattern, len(res), res)
+        return res[0]
+
+    def configure_step(self):
+        """Set CMake options for Blender"""
+        self.cfg['separate_build_dir'] = True
+
+        default_config_opts = {
+            'WITH_INSTALL_PORTABLE': 'OFF',
+            'WITH_BUILDINFO': 'OFF',
```
boegel: @smoors Please sort alphabetically by key?

easybuild/easyblocks/b/blender.py
```diff
@@ -0,0 +1,113 @@
...
+                self.cfg.update('configopts', opt + default_config_opts[key])
+
+        # Python paths
+        python_root = get_software_root('Python')
+        if python_root:
+            shlib_ext = get_shared_lib_ext()
+            pyshortver = '.'.join(get_software_version('Python').split('.')[:2])
+            site_packages = os.path.join(python_root, 'lib', 'python%s' % pyshortver, 'site-packages')
+
+            numpy_root = self.find_glob_pattern(
```
boegel: @smoors Maybe include a comment here that we assume that `numpy/requests` are included with the Python installation.

Note that this will need to be updated for the `2019a` generation where `numpy` lives in `SciPy-bundle` (but this can be tackled in a follow-up PR, not need to block this one for it imho).

easybuild/easyblocks/b/blender.py
```diff
@@ -0,0 +1,113 @@
...
+        super(EB_Blender, self).configure_step()
+
+    def sanity_check_step(self):
+        """Custom sanity check for Blender."""
+
+        custom_paths = {
+            'files': ['bin/blender'],
+            'dirs': [],
+        }
+
```
boegel: empty lines above & below can be dropped imho (but that's just personal preference)

smoors:
test succeeded in https://github.com/easybuilders/easybuild-easyconfigs/pull/7188

boegel:
Tested with https://github.com/easybuilders/easybuild-easyconfigs/pull/7188, good to go, thanks @smoors!

PR #1596, opened 2018-11-28
Fix Gromacs use of MKL for non-Intel compilers (e.g. gomkl toolchain).

bartoldeman:
Using only 'libmkl_lapack.a' (a text file with
GROUP (libmkl_intel_lp64.a libmkl_intel_thread.a libmkl_core.a)
) only works with the Intel compiler.
It can use $LAPACK_STATIC_LIBS but needs to use -Wl,--start-group
and -Wl,--end-group to avoid symbol errors.

There is no need to use the threaded MKL, so don't use MT here,
following the same method that was used for other BLAS/LAPACK.


migueldiascosta:
ah, this is targeting `master` instead of `develop`

bartoldeman:
@migueldiascosta oops yes, I have changed the base now.

migueldiascosta:

PR #1597, opened 2018-12-06
Enable TensorFlow to detect any MPI runtime supported in EB

damianam:


akesandgren:
Going in, thanks @damianam!

PR #1600, opened 2018-12-11
add custom easyblock for RepeatMasker

boegel:


vanzod:
Tested with https://github.com/easybuilders/easybuild-easyconfigs/pull/7281. Correctly built.
LGTM

PR #1601, opened 2018-12-11
Trilinos - Support build against MKL

vanzod:
Trilinos has a dedicated flag for building against MKL that should be used in conjunction with the BLAS and LAPACK flags. Set the flag to `True` if MKL is available.

### DIFF
--- a/easybuild/easyblocks/t/trilinos.py
+++ b/easybuild/easyblocks/t/trilinos.py
@@ -124,6 +124,12 @@ class EB_Trilinos(CMakeMake):
                 lib_names += ";libgfortran.a"
             self.cfg.update('configopts', '-D%s_LIBRARY_NAMES="%s"' % (dep, lib_names))
 
+        # MKL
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/t/trilinos.py
```diff
@@ -124,6 +124,12 @@ def configure_step(self):
                 lib_names += ";libgfortran.a"
             self.cfg.update('configopts', '-D%s_LIBRARY_NAMES="%s"' % (dep, lib_names))
 
+        # MKL
+        if get_software_root('imkl'):
+            self.cfg.update('configopts', "-DTPL_ENABLE_MKL:BOOL=ON")
```
boegel: @vanzod Maybe it's worth considering only doing this conditionally for sufficiently recent versions of Trilinos, e.g. `> 12.12` (which is already quite old, `12.12.1` dates from Sept'17)

boegel:
Going in, thanks @vanzod!

PR #1602, opened 2018-12-12
(WIP) Enable PGO builds for psmpi

damianam:
This is relevant just for KNL systems

### DIFF
--- a/easybuild/easyblocks/p/psmpi.py
+++ b/easybuild/easyblocks/p/psmpi.py
@@ -36,22 +36,23 @@ from distutils.version import LooseVersion
 from easybuild.easyblocks.mpich import EB_MPICH
 from easybuild.framework.easyconfig import CUSTOM
 from easybuild.tools.build_log import EasyBuildError, print_msg
-from easybuild.tools.filetools import remove_file, write_file
+from easybuild.tools.filetools import mkdir, remove_dir, write_file
 from easybuild.tools.modules import get_software_root
 from easybuild.tools.run import run_cmd
 
 
 PINGPONG_PGO_TEST = """
 /****************************************
- * Buggy, ugly and extremely simple MPI ping pong.
- * The goal is to trigger pscom internal routines
- * to generate profiles to be used by PGO-enabled
- * compilers.
+ * Potentially buggy, ugly and extremely simple MPI ping pong.
+ * The goal is to trigger pscom internal routines to generate
+ * profiles to be used by PGO-enabled compilers.
  *
  * Nothing more, nothing less.
  *
- * We try small and large messages to walk the path for
- * eager and rendezvous protocols
+ * We try small and large messages to walk the path for eager
+ * and rendezvous protocols
+ *
+ * author: Damian Alvarez (d.alvarez@fz-juelich.de)
  ****************************************/
 
 #include "mpi.h"
@@ -91,7 +92,12 @@ class EB_psmpi(EB_MPICH):
     * Determines the compiler to be used based on the toolchain
     * Enables threading if required by the easyconfig
     * Sets extra MPICH options if required by the easyconfig
+    * Generates a PGO profile and uses it if required
     """
+
+    def __init__(self, *args, **kwargs):
+        super(EB_psmpi, self).__init__(*args, **kwargs)
+        self.profdir = os.path.join(self.builddir, 'profile')
 
     @staticmethod
     def extra_options(extra_vars=None):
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/p/psmpi.py
```diff
@@ -29,12 +29,60 @@
...
+/****************************************
+ * Buggy, ugly and extremely simple MPI ping pong.
+ * The goal is to trigger pscom internal routines
+ * to generate profiles to be used by PGO-enabled
+ * compilers.
+ *
+ * Nothing more, nothing less.
+ *
+ * We try small and large messages to walk the path for
+ * eager and rendezvous protocols
```
boegel: @damianam Where did this come from? Should mention author or upstream source?

easybuild/easyblocks/p/psmpi.py
```diff
@@ -58,9 +106,56 @@ def extra_options(extra_vars=None):
...
+            'mpiexec_cmd': ['srun -n ', "Command to run benchmarks to generate PGO profile. With -n switch", CUSTOM],
         })
         return extra_vars
 
+    def pgo_steps(self):
+        """
+        Set of steps to be performed after the initial installation, if PGO were enabled
+        """
+        # Remove old profiles
+        prof_dir = '%s/profile/' % self.builddir
```
boegel: `os.path.join(self.builddir, 'profile')`

easybuild/easyblocks/p/psmpi.py
```diff
@@ -58,9 +106,56 @@ def extra_options(extra_vars=None):
...
+        old_profs = os.listdir(prof_dir)
+        for prof in [os.path.join(prof_dir, p) for p in old_profs]:
+            remove_file(prof)
```
boegel: `remove_dir(prof_dir)` followed by `mkdir(prof_dir)`?

easybuild/easyblocks/p/psmpi.py
```diff
@@ -58,9 +106,56 @@ def extra_options(extra_vars=None):
...
+        new_profs = os.listdir(prof_dir)
+        if new_profs:
+            raise EasyBuildError("The configure PGO profiles were not wiped (%s)" % new_profs)
```
boegel: How can this ever be triggered?

easybuild/easyblocks/p/psmpi.py
```diff
@@ -92,6 +187,10 @@ def configure_step(self):
         if self.cfg['mpich_opts'] is not None:
             self.cfg.update('configopts', ' --with-mpichconf="%s"' % self.cfg['mpich_opts'])
 
+        # Add PGO related options, if enabled
+        if self.cfg['pgo']:
+            self.cfg.update('configopts', ' --with-profile=gen --with-profdir=%(builddir)s/profile')
```
boegel: @damianam Avoid having the profdir location in two places, use a class variable that you define early on (in `__init__`)

easybuild/easyblocks/p/psmpi.py
damianam: There is no upstream source or author other than me. It is a very simple test. I based it on a hello world example, but I don't think it should matter too much......

easybuild/easyblocks/p/psmpi.py
```diff
@@ -29,12 +29,73 @@
 """
 
 import easybuild.tools.toolchain as toolchain
```
boegel: @damianam While you're at it, please move down this import among the other `easybuild` import statements

easybuild/easyblocks/p/psmpi.py
```diff
@@ -58,9 +124,59 @@ def extra_options(extra_vars=None):
             'mpich_opts': [None, "Optional options to configure MPICH", CUSTOM],
             'threaded': [False, "Enable multithreaded build (which is slower)", CUSTOM],
             'pscom_allin_path': [None, "Enable pscom integration by giving its source path", CUSTOM],
+            'pgo': [False, "Enable profiling guided optimizations", CUSTOM],
+            'mpiexec_cmd': ['srun -n ', "Command to run benchmarks to generate PGO profile. With -n switch", CUSTOM],
```
akesandgren: mpirun is a better default, that usually works even outside the batch system.

easybuild/easyblocks/p/psmpi.py
```diff
@@ -58,9 +124,59 @@ def extra_options(extra_vars=None):
...
+
+        # Compile and run example to generate profile
+        print_msg("generating PGO profile...")
+        (out, _) = run_cmd('%s 2 hostname' % self.cfg['mpiexec_cmd'])
+        nodes = out.split()
+        if nodes[0] == nodes[1]:
+            raise EasyBuildError("The profile is generated with 1 node! Use 2 nodes to generate a proper profile!")
+
+        write_file('pingpong.c', PINGPONG_PGO_TEST)
+        run_cmd('%s/bin/mpicc pingpong.c -o pingpong' % self.installdir)
```
akesandgren: os.path.join(self.installdir, 'bin', 'mpicc') or perhaps the correct ENV var for MPICC

easybuild/easyblocks/p/psmpi.py
ocaisa: `mpiexec -np` is the only one that is actually part of the standard, if you were to go for a generic default then it should really be that

akesandgren:
@damianam ping

damianam:
I really dropped the ball here, but it is not relevant anymore, so closing.

PR #1603, opened 2018-12-12
cleanup in Perl and PerlModule easyblocks

boegel:


akesandgren:
Going in, thanks @boegel!

PR #1604, opened 2018-12-12
inform Hound about Python 2 builtins

boegel:
cfr. HoundCI complaints in #1583


akesandgren:
Going in, thanks @boegel!

PR #1609, opened 2018-12-19
Make openfoam easyblock handle debug build.

akesandgren:
This makes a debug build of OpenFOAM possible.


boegel:
lgtm, tested with a whole bunch of existing `OpenFOAM` easyconfigs, so good to go, thanks @akesandgren!

PR #1610, opened 2018-12-19
Handle multiple installation keys for MATLAB

akesandgren:
One can have multiple installation keys to matlab, if so one has to do
one install per key.
Matlab installer ignores TMPDIR (always using /tmp), give use the -tmpdir option.
Don't add JAVA_OPTIONS to the module file if there are no java_options.

### DIFF
--- a/easybuild/easyblocks/m/matlab.py
+++ b/easybuild/easyblocks/m/matlab.py
@@ -60,7 +60,6 @@ class EB_MATLAB(PackedBinary):
     def extra_options():
         extra_vars = {
             'java_options': ['-Xmx256m', "$_JAVA_OPTIONS value set for install and in module file.", CUSTOM],
-            'tmpdir': [None, "Tempdir during installation, passed to matlab installer", CUSTOM],
             'key': [None, "Installation key(s), make one install for each key. Single key or a list of keys", CUSTOM],
         }
         return PackedBinary.extra_options(extra_vars)
--- a/easybuild/easyblocks/m/matlab.py
+++ b/easybuild/easyblocks/m/matlab.py
@@ -129,19 +129,13 @@ class EB_MATLAB(PackedBinary):
         if LooseVersion(self.version) >= LooseVersion('2016b'):
             change_dir(self.builddir)
 
-        # MATLAB installer ignores TMPDIR
-        tmpdir = ''
-        if self.cfg['tmpdir']:
-            tmpd = self.cfg['tmpdir']
-        else:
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/m/matlab.py
```diff
@@ -133,18 +129,45 @@ def install_step(self):
...
-                raise EasyBuildError("Found error pattern '%s' in output of installation command '%s': %s",
-                                     regex.pattern, cmd, out)
+        # MATLAB installer ignores TMPDIR
+        tmpdir = ''
+        if self.cfg['tmpdir']:
+            tmpd = self.cfg['tmpdir']
+        else:
+            tmpd = os.getenv('TMPDIR', '')
+        if tmpd:
+            tmpdir = '-tmpdir %s' % tmpd
```
boegel: @akesandgren EasyBuild always sets up a unique temporary directory per installation, so there's no need to introduce a custom `tmpdir` easyconfig parameter for MATLAB, you can simplify this to:

```python
tmpdir = "-tmpdir %s" % tempfile.mkdtemp()

File: easybuild/easyblocks/m/matlab.py [hunk 2]
```diff
@@ -133,18 +129,45 @@ def install_step(self):
...
+
+        keys = self.cfg['key']
+        if keys is None:
+            keys = os.getenv('EB_MATLAB_KEY', '00000-00000-00000-00000-00000-00000-00000-00000-00000-00000')
+        if isinstance(keys, str):
```
boegel: This should be

```python
if isinstance(keys, basestring):

File: easybuild/easyblocks/m/matlab.py (hunk 1 above)
Comment (akesandgren) [reply easybuild/easyblocks/m/matlab.py]: Ah yes, I forgot about that when i originally made this. The problem was that I needed a much larger tmpdir to get it to build/install...

File: easybuild/easyblocks/m/matlab.py [hunk 3]
```diff
@@ -60,6 +60,8 @@ def __init__(self, *args, **kwargs):
     def extra_options():
         extra_vars = {
             'java_options': ['-Xmx256m', "$_JAVA_OPTIONS value set for install and in module file.", CUSTOM],
+            'tmpdir': [None, "Tempdir during installation, passed to matlab installer", CUSTOM],
```
boegel: @akesandgren This one should be dropped now?

boegel:
lgtm, tested, so good to go, thanks @akesandgren!

PR #1611, opened 2019-01-03
run 'make check' in parallel for GROMACS since it involves more compilation

akesandgren:


akesandgren:
Nah, we want "make -j" if self.cfg['parallel'] isn't set.

easybuild/easyblocks/g/gromacs.py
```diff
@@ -314,7 +314,8 @@ def test_step(self):
             # make very sure OMP_NUM_THREADS is set to 1, to avoid hanging GROMACS regression test
             env.setvar('OMP_NUM_THREADS', '1')
 
-            self.cfg['runtest'] = 'check'
+            # run 'make check' in parallel since it involves more compilation
+            self.cfg['runtest'] = 'check -j %s' % self.cfg['parallel']
```
boegel: @akesandgren @ocaisa's suggestion is valid here: the `parallel` easyconfig parameter may not be set (at least in theory, it's pretty uncommon), so let's do this instead:

```python
self.cfg['runtest'] = 'check'
if self.cfg['parallel']:
    # run 'make check' in parallel since it involves more compilation
    self.cfg.update('runtest', "-j %s" % self.cfg['parallel'])
```

boegel:
Tested parallel `make check` with:

* `GROMACS-5.1.2-foss-2015b-hybrid.eb`
*  `GROMACS-5.1.2-intel-2016a-hybrid-dp.eb`
* `GROMACS-2016.2-foss-2017a.eb`
* `GROMACS-2016.4-intel-2017a.eb`
* `GROMACS-2016.4-foss-2017b.eb`
* `GROMACS-2016.5-intel-2018a.eb`
* `GROMACS-2018-foss-2018a.eb`
* `GROMACS-2018.3-foss-2018b.eb`

All worked fine, so good to go as soon as @ocaisa's remark is tackled.

PR #1612, opened 2019-01-05
avoid that (system) Intel compilers are always considered when building SuiteSparse

boegel:
cfr. https://github.com/easybuilders/easybuild-easyconfigs/pull/7348 by @casparvl


akesandgren:
Going in, thanks @boegel!

PR #1613, opened 2019-01-10
add custom easyblock for STAR-CCM+

boegel:


easybuild/easyblocks/s/star_ccm.py
```diff
@@ -0,0 +1,114 @@
...
+# EasyBuild is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with EasyBuild.  If not, see <http://www.gnu.org/licenses/>.
+##
+"""
+EasyBuild support for building and installing STAR-CCM+, implemented as an easyblock
```
wpoely86: author tag?

easybuild/easyblocks/s/star_ccm.py
```diff
@@ -0,0 +1,114 @@
...
+        """No build procedure for STAR-CCM+."""
+        pass
+
+    def install_step(self):
+        """Custom install procedure for STAR-CCM+."""
+        cands = glob.glob("./STAR-CCM+%s_*.sh" % self.version)
+        if len(cands) == 1:
+            install_script = cands[0]
+
+            env.setvar('CHECK_DISK_SPACE', 'OFF')
```
wpoely86: that's neat

easybuild/easyblocks/s/star_ccm.py
```diff
@@ -0,0 +1,114 @@
...
+            env.setvar('IATEMPDIR', tempfile.mkdtemp())
+
+            cmd = ' '.join([
+                install_script,
+                "-i silent",
+                "-DINSTALLDIR=%s" % self.installdir,
+                "-DINSTALLFLEX=false",
+                "-DADDSYSTEMPATH=false",
+            ])
```
wpoely86: add preinstallopts and installopts here too?

easybuild/easyblocks/s/star_ccm.py
```diff
@@ -0,0 +1,114 @@
...
+            run_cmd(cmd, log_all=True, simple=True)
+        else:
+            raise EasyBuildError("Found no or multiple possible matches for STAR-CCM+ installation script: %s", cands)
```
wpoely86: handle dry-run here too?

easybuild/easyblocks/s/star_ccm.py
```diff
@@ -0,0 +1,114 @@
...
+    def find_starccm_subdirs(self):
+        """Determine subdirectory of install directory in which STAR-CCM+ was installed."""
+        cwd = change_dir(self.installdir)
+        cands = glob.glob(os.path.join(self.version + '*', 'STAR-CCM+%s*' % self.version))
+        if len(cands) == 1:
+            self.starccm_subdir = cands[0]
+            self.log.info("Found STAR-CCM+ subdirectory: %s", self.starccm_subdir)
+            self.starview_subdir = os.path.join(os.path.dirname(self.starccm_subdir), 'STAR-View+%s' % self.version)
+        else:
+            raise EasyBuildError("Failed to determine the STAR-CCM+ subdirectory in %s: %s", self.installdir, cands)
```
wpoely86: dry-run case

wpoely86: maybe this flow should be put in the framework? I've seen it in multiple easyblock ?

easybuild/easyblocks/s/star_ccm.py
```diff
@@ -0,0 +1,114 @@
...
+            'files': [os.path.join(self.installdir, self.starccm_subdir, 'star', 'bin', 'starccm+'),
+                      os.path.join(self.installdir, self.starview_subdir, 'bin', 'starview+')],
+            'dirs': [],
+        }
+        super(EB_STAR_minus_CCM_plus_, self).sanity_check_step(custom_paths=custom_paths)
+
+    def make_module_extra(self):
+        """Extra statements specific to STAR-CCM+ to include in generated module file."""
+        if self.starccm_subdir is None or self.starview_subdir is None:
+            self.find_starccm_subdirs()
```
wpoely86: is this still needed? Sanity check already does it?

akesandgren:
@boegel status of this one?

boegel:
@akesandgren I don't have time to get back to this, for now.

akesandgren:
@boegel Would be useful if you could get back to this one soon...

easybuild/easyblocks/s/star_ccm.py
boegel: Yeah, and it gets better...

I just spent a couple of hours figuring out why the install script for STAR-CCM+ 16.06.008 was exiting with a non-zero exit code, with zero output.

After finding the log file (buried in `$EBROOTSTARMINCCMPLUS/16.06.008/STAR-CCM+16.06.008/STAR-CCM+16.06.008_InstallLog.log` (yes, the version subdir is there twice...), and mostly seeing `Status: SUCCESSFUL`, I was even more confused.

There was one `Status: ERROR` though, in the disk space check step....

So it seems that setting `$CHECK_DISK_SPACE` to `OFF` skips the disk space check, but also marks that step as failed, and hence the install script exists with a non-zero exit code. 🤦 

So to fix: either don't set `$CHECK_DISK_SPACE` (if that works, I think it's only required when installing to an NFS mounted filesystem), or totally ignore the exit code of the install script and fully rely on the sanity check done by EasyBuild...

boegel:
Hmm, looks like GitHub is a bit confused with the commit history... Those commits for the 4.5.5 are already in develop (see #2741).
Maybe another push (which is needed to fix the tests anyway) will resolve that.

easybuild/easyblocks/s/star_ccm.py
boegel: It actually already exists: [`find_glob_pattern` from `easybuild.tools.filetools`](https://docs.easybuild.io/en/latest/api/easybuild.tools.filetools.html#easybuild.tools.filetools.find_glob_pattern)

easybuild/easyblocks/s/star_ccm.py
boegel: It's needed when doing `--module-only --force`, because then the sanity check step is never run.

easybuild/easyblocks/s/star_ccm.py
boegel: fixed in b2dc80685

easybuild/easyblocks/s/star_ccm.py
boegel: I disabled this again, since it was causing trouble.

People can still resort to `CHECK_DISK_SPACE=OFF eb ...` when installing STAR-CCM+, if the disk space check is causing trouble.

easybuild/easyblocks/s/star_ccm.py
boegel: fixed in b2dc80685

easybuild/easyblocks/s/star_ccm.py
boegel: fixed in b2dc80685

easybuild/easyblocks/s/star_ccm.py
boegel: fixed in b2dc80685

wpoely86 (approved):

PR #1617, opened 2019-01-15
Mathematica: let's answer SELinux question with 'no'?

jose-d:
In case when `SELinux` is enabled (e.g. in `permissive` mode like I have it), Mathematica asks if the SELinux context of installed files should be updated by the installer:

```
The installer has detected that SELinux is enabled on this system. The
security context of the included libraries may need to be altered in order
for Wolfram Mathematica to function properly.

Should the installer attempt to make this change (y/n)?
```

Currently, this questions stays unanswered and the installation fails.
Having this question in Q/A dictionary will fix this "problem".

I assume "yes" is the right answer for most of the sites.

### DIFF
--- a/easybuild/easyblocks/m/mathematica.py
+++ b/easybuild/easyblocks/m/mathematica.py
@@ -68,6 +68,7 @@ class EB_Mathematica(Binary):
         qa = {
             r"Enter the installation directory, or press ENTER to select %s: >" % qa_install_path: self.installdir,
             r"Create directory (y/n)? >": 'y',
+            r"Should the installer attempt to make this change (y/n)? >": 'y',
             r"or press ENTER to select /usr/local/bin: >": os.path.join(self.installdir, "bin"), 
         }
         no_qa = [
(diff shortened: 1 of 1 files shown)


akesandgren:
The answer should in all likelyhood be "no", since easybuild installation is done but non-root users and thus can't update the SElinux config.

Don't have SElinux so I'm only guessing.

jose-d:
> The answer should in all likelyhood be "no", since easybuild installation is done but non-root users and thus can't update the SElinux config.

Right. It is working even with 'yes', because installer _attempts_ to do it, but after reading some manual the 'no' is seems be the better answer.

jose-d:
Travis - broken OS deps?
```
...
The following packages have unmet dependencies:
 libopenmpi-dev : Depends: libhwloc-dev but it is not going to be installed
...
The command "sudo -E apt-get -yq --no-install-suggests --no-install-recommends 
...
```

boegel:
@jose-d Failing test looks like a fluke in Travis, I've retriggered it.

boegel:
tested with existing `Mathematica` easyconfigs, works fine, so good to go

Thanks @jose-d!

Issue #1608, opened 2018-12-18
Make CMakeMake easyblock immune to system boost.


bartoldeman:
As mentioned in 
https://github.com/easybuilders/easybuild-easyconfigs/pull/7149
if an old boost-devel rpm installed on CentOS6:
```
boost-devel-1.41.0-25.el6.centos.x86_64

$ rpm -ql boost-devel-1.41.0-25.el6.centos.x86_64 | grep -i cmake
/usr/lib64/boost/Boost-relwithdebinfo.cmake
/usr/lib64/boost/Boost.cmake
/usr/lib64/boost/BoostConfig.cmake
/usr/lib64/boost/BoostConfigVersion.cmake

-- Boost  found.
Boost Include: /usr/include
Boost Linkdir: /usr/lib64
```
then anything that combines cmake and boost gets confused by this as the EB boost does not include any cmake files.

adding `-DBoost_NO_SYSTEM_PATHS=ON -DBoost_NO_BOOST_CMAKE=ON`
avoids this.

ref:
https://github.com/Kitware/CMake/blob/master/Modules/FindBoost.cmake
(which shows that if it finds a Boost-cmake config file, it's game over)
https://cmake.org/cmake/help/v3.12/module/FindBoost.html

vanzod:
Solved with #1618

PR #1618, opened 2019-01-19
make sure CMake doesn't pick up on system Boost in CMakeMake generic easyblock (

boegel:
fix for #1608 raised by @bartoldeman

I tested this with a couple of easyconfigs that use `CMakeMake` and have `Boost` as a dependency, seems to work as expected, but more extensive testing is needed.

I also wonder whether there is a more general approach to take here, why only do this for `Boost`?

### DIFF
--- a/easybuild/easyblocks/generic/cmakemake.py
+++ b/easybuild/easyblocks/generic/cmakemake.py
@@ -39,6 +39,7 @@ from easybuild.framework.easyconfig import CUSTOM
 from easybuild.tools.config import build_option
 from easybuild.tools.filetools import change_dir, mkdir
 from easybuild.tools.environment import setvar
+from easybuild.tools.modules import get_software_root
 from easybuild.tools.run import run_cmd
 from vsc.utils.missing import nub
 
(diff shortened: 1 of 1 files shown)


akesandgren:
This only works for CmakeFiles that uses CMakes builtin search for Boost.
Good enough for most Boost related cases but there's probably not a generic approach that will work.

And there might be situations where this is the wrong thing to do.
Controlling this is really up to each individual easyconfig.

ocaisa:
@akesandgren If it doesn't work then it is doing no harm so I think it should be on by default, but for individual easyconfigs I don't see an issue with being able to switch it off

easybuild/easyblocks/generic/cmakemake.py
```diff
@@ -107,6 +108,14 @@ def configure_step(self, srcdir=None, builddir=None):
         # show what CMake is doing by default
         options.append('-DCMAKE_VERBOSE_MAKEFILE=ON')
 
+        # don't pick up on system Boost if Boost is included as dependency
+        # - specify Boost location via -DBOOST_ROOT
+        # - instruct CMake to not search for Boost headers/libraries in other places
+        # - disable search for Boost CMake package configuration file
+        boost_root = get_software_root('Boost')
+        if boost_root:
```
ocaisa: Add an option `enable_system_boost` with a default of false so this can be switched off in a corner case.

Also should we care that this may only be relevant to versions of CMake that supported this functionality?

ocaisa: Looks like `Find_Boost` goes back to 2.8.4 (oldest in EB) so should be fine

boegel:
I've tested this quite extensively, haven't hit any problems at all, so this is good to go imho...

PR #1619, opened 2019-01-19
enhance Python easyblock to also include <prefix>/include/python* in $CPATH

boegel:
fix for #1446 reported by @akesandgren

Generated `Python` module files are as expected:

```diff
$ diff -u /prefix/modules/all/Python/3.6.6-intel-2018b.lua.bk /prefix/modules/all/Python/3.6.6-intel-2018b.lua
--- /prefix/modules/all/Python/3.6.6-intel-2018b.lua.bk	2018-12-15 09:59:14.938975000 +0100
+++ /prefix/modules/all/Python/3.6.6-intel-2018b.lua	2019-01-19 11:37:33.613860000 +0100
@@ -71,6 +71,7 @@
 end

 prepend_path("CPATH", pathJoin(root, "include"))
+prepend_path("CPATH", pathJoin(root, "include/python3.6m"))
 prepend_path("LD_LIBRARY_PATH", pathJoin(root, "lib"))
 prepend_path("LIBRARY_PATH", pathJoin(root, "lib"))
 prepend_path("MANPATH", pathJoin(root, "share/man"))

$ diff -u /prefix/modules/all/Python/2.7.15-intel-2018b.lua.bk /prefix/modules/all/Python/2.7.15-intel-2018b.lua
--- /prefix/modules/all/Python/2.7.15-intel-2018b.lua.bk	2019-01-04 22:06:17.270376000 +0100
+++ /prefix/modules/all/Python/2.7.15-intel-2018b.lua	2019-01-19 11:36:31.050536000 +0100
@@ -67,6 +67,7 @@
 end

 prepend_path("CPATH", pathJoin(root, "include"))
+prepend_path("CPATH", pathJoin(root, "include/python2.7"))
 prepend_path("LD_LIBRARY_PATH", pathJoin(root, "lib"))
 prepend_path("LIBRARY_PATH", pathJoin(root, "lib"))
 prepend_path("MANPATH", pathJoin(root, "share/man"))

```

Issue #1446, opened 2018-06-30
Python easyblock doesn't add the correct include dir to $CPATH


akesandgren:
The python easyblock only add $EBROOTPYTHON/include to CPATH, but the Python include files are in
$EROOTPYTHON/include/python%(pyver)s

This makes some codes fail in compilation.

boegel:
@akesandgren Do you have an example of something that fails because of this?

It's a bit strange this has never popped up before...

akesandgren:
Scipion (or rather Xmipp inside Scipion) fails to find Python.h.
Since that uses SCons it would normally use the system python and link stuff to its internal tree and thus bypass the problem.

I don't know of any other examples at the moment. But the current CPATH add from python.py just uses ConfigureMakes rule for CPATH additions.

Adding include/%(pyver)s may well break stuff though, for instance if something has a dep-chain that pulls in multiple python versions. But if it does I would say that the dependency chain, for whatever it is that gets broken, is then broken in itself.

I know that the preferred style is to either have explicit -Ipython-path/include/pythonx.y or have
#include <pythonx.y/something.h>
but I see lots of code that don't follow the rules (in general that is)

mstud:
Another example where this fails would be LAMMPS with the PYTHON package enabled (even though #594 is not officially in yet):
```
../pair_python.cpp:18:20: fatal error: Python.h: No such file or directory
 #include <Python.h>
                    ^
compilation terminated.

```

boegel:
@akesandgren I'm not sure we should risk breaking stuff by always adding `include/pythonX.Y` to `$CPATH`...

I'm more inclined to only do so for software where it's really needed, like `Xmipp` or `LAMMPS`, since there the fix is pretty straightforward. While if software starts failing to build because we have included it through the `Python` module, it may be a lot harder to resolve those issues?

@mstud Can you mention that in #594 explicitly?

mstud:
Also note that `include/%(pyver)s` won't be enough in case of Python 3, there you also may need to add the abiflags to the version, e.g. `include/python3.6m`

boegel:
@mstud Good point, that does probably mean we should provide a function in `Python.py` that other easyblocks can leverage to determine the (full) include path...

akesandgren:
Well, I don't really think there is a risk to add include/%(pyver)s (or similar for P3) to CPATH.
[...]

PR #1620, opened 2019-01-22
revert change that adds 'include/python<version>*' to $CPATH for Python

boegel:
This reverts the changes made in #1619, since it turns it causes problem with `ROOT` for example, leading to errors like:

```
In file included from /tmp/easybuild_build/ROOT/6.14.08/foss-2019a-Python-2.7.15/root-6-14-08/graf2d/asimage/src/TASPluginGS.cxx:40:
/prefix/software/Python/2.7.15-foss-2019a/include/python2.7/import.h:10:18: error: expected constructor, destructor, or type conversion before ‘PyImport_GetMagicNumber’
 PyAPI_FUNC(long) PyImport_GetMagicNumber(void);
                  ^~~~~~~~~~~~~~~~~~~~~~~
/prefix/software/Python/2.7.15-foss-2019a/include/python2.7/import.h:11:11: error: expected constructor, destructor, or type conversion before ‘(’ token     PyAPI_FUNC(PyObject *) PyImport_ExecCodeModule(char *name, PyObject *co);
           ^
```

The issue is that the Python header files have quite generic names like `import.h`:

```
abstract.h	   code.h	    funcobject.h    methodobject.h  pyconfig.h	    pystrcmp.h	    timefuncs.h
asdl.h		   compile.h	    genobject.h     modsupport.h    pyctype.h	    pystrtod.h	    token.h
ast.h		   complexobject.h  graminit.h	    moduleobject.h  py_curses.h     Python-ast.h    traceback.h
bitset.h	   cStringIO.h	    grammar.h	    node.h	    pydebug.h	    Python.h	    tupleobject.h
boolobject.h	   datetime.h	    import.h	    object.h	    pyerrors.h	    pythonrun.h     ucnhash.h
bufferobject.h	   descrobject.h    intobject.h     objimpl.h	    pyexpat.h	    pythread.h	    unicodeobject.h
bytearrayobject.h  dictobject.h     intrcheck.h     opcode.h	    pyfpe.h	    rangeobject.h   warnings.h
bytes_methods.h    dtoa.h	    iterobject.h    osdefs.h	    pygetopt.h	    setobject.h     weakrefobject.h
bytesobject.h	   enumobject.h     listobject.h    parsetok.h	    pymacconfig.h   sliceobject.h
cellobject.h	   errcode.h	    longintrepr.h   patchlevel.h    pymactoolbox.h  stringobject.h
ceval.h		   eval.h	    longobject.h    pgen.h	    pymath.h	    structmember.h
classobject.h	   fileobject.h     marshal.h	    pgenheaders.h   pymem.h	    structseq.h
cobject.h	   floatobject.h    memoryobject.h  pyarena.h	    pyport.h	    symtable.h
codecs.h	   frameobject.h    metagrammar.h   pycapsule.h     pystate.h	    sysmodule.h
```

So it's quite easy to introduce a clash on names of header files by always injecting `include/python2.7` to `$CPATH`...

### DIFF
--- a/easybuild/easyblocks/p/python.py
+++ b/easybuild/easyblocks/p/python.py
@@ -254,20 +254,3 @@ class EB_Python(ConfigureMake):
                 raise EasyBuildError("Expected to find exactly one _tkinter*.so: %s", tkinter_so_hits)
 
         super(EB_Python, self).sanity_check_step(custom_paths=custom_paths, custom_commands=custom_commands)
-
-    def make_module_req_guess(self):
-        """
-        Return dictionary with additional entries for path-like environment variables;
-        also include <prefix>/include/python* to $CPATH.
-        """
-        guesses = super(EB_Python, self).make_module_req_guess()
-
-        pyincdirs = glob.glob(os.path.join(self.installdir, 'include', 'python' + self.pyshortver + '*'))
-        if len(pyincdirs) == 1:
-            guesses['CPATH'].append(os.path.join('include', os.path.basename(pyincdirs[0])))
-        elif pyincdirs:
-            # only fail if *multiple* hits were found
-            # under --module-only --force or --extended-dry-run, nothing may actually be installed (yet)
-            raise EasyBuildError("Failed to isolate subdirectory with Python header files: %s", pyincdirs)
-
-        return guesses


akesandgren:
Since we have "CPATH specifies a list of directories to be searched as if specified with -I, but after any paths given with -I options on the command line."
ROOT should make sure to have -I for the dir where it's import.h lives. And then there shouldn't be a problem.
If ROOT isn't doing this then it is to blame. If it is then something else is wrong.
After all, we did put the include/python2.7 last in CPATH right?

So technically speaking this patch is still correct.

akesandgren:
What was the actual compiler invocation when that blew up?

boegel:
@akesandgren We put the `include/python<version>*` last in the list of guesses for `$CPATH`, but it will actually be *prepended* to `$CPATH` when the `Python` module gets loaded...

Even if we make sure we append, it still matters in which order dependencies are being loaded whether an `import.h` provided by something else is shadowed or not.

akesandgren:
Yes, deps that also does CPATH addition might cause problems for what you are building.
But if you are appending include/pythonx.y to the actual CPATH env-var, then any dep that adds to it should be prepending and thus be picked up first.

The problem with ROOT has to be something else since ROOT's own include files should not be subject to CPATH searching.

ocaisa:
Browsing the source, what it actually wants to import is (I think) https://github.com/cxx-hep/root-cern/blob/master/graf2d/asimage/src/libAfterImage/import.h
It should be more careful about this but tracking down the problem in their build system is probably not so easy.

ocaisa:
Ok, actually it wasn't too hard. The problem is 
https://github.com/root-project/root/blob/331efa4c00fefc38980eaaf7b41b8e95fcd1a23b/graf2d/asimage/CMakeLists.txt#L11
From https://cmake.org/cmake/help/v3.13/command/include_directories.html it looks like setting `CMAKE_INCLUDE_DIRECTORIES_BEFORE` to `ON` would solve the issue

ocaisa:
@boegel This sounds like a setting we may want on by default in the CMake easyblock?

akesandgren:
So a simple patch to change
include_directories(SYSTEM ${FREETYPE_INCLUDE_DIRS} ${AFTERIMAGE_INCLUDE_DIR} ${X11_INCLUDE_DIR})
to
include_directories(BEFORE ${FREETYPE_INCLUDE_DIRS} ${AFTERIMAGE_INCLUDE_DIR} ${X11_INCLUDE_DIR})

should solve this problem.

boegel:
@akesandgren I'm confident we can fix the issue with `ROOT`, but it starting to fail after merging #1619 really makes me wonder whether it's a good idea to through with that change (which is why I'm reverting it here). I suspect there may be other software that will start failing in a similar way because of this change, which is a regression that I'd like to avoid...

I'm actually not aware of builds failing because we haven't been including `include/python<version>*` in `$CPATH`. Do you have examples where that is actually a problem?
And why don't we include `-I$EBROOTPYTHON/include/python*` in `$CFLAGS` (& co) instead?

It seems like it may be less painful to fix those, rather than trying to have a general fix which causes problems elsewhere...

akesandgren:
Yes, I do agree that reverting this is the way forward due to the immediate fail.
The example I had is the one(s) it listed in the issue for this.

Lets merge the revert...

akesandgren:
Going in, thanks @boegel!

boegel:
Just for clarity, the original motivation for this (now reverted) change was `Scipion` failing to find `Python.h`, see https://github.com/easybuilders/easybuild-easyblocks/issues/1446#issuecomment-401758620

PR #1621, opened 2019-01-22
Add custom easyblock for ELPA

micaeljtoliveira:
Tested on both x86_64 and ppc, and seems to select the appropriate configuration flags depending on the type of CPU and available features. I also added a few other custom options.

The only thing I'm not really sure about is if we should compile with and without MPI and install both versions of the library, as this is supported by ELPA. This should not be very difficult to implement.

Also, once this is in the corresponding easyconfigs will have to be updated. What is the usual procedure for that?


ocaisa:
@micaeljtoliveira Can you open a merge request for an easyconfig that uses this?

Regarding updating existing easyconfigs, you don't necessarily have to update existing ones, you can just create a PR for the latest version with this

easybuild/easyblocks/e/elpa.py
```diff
@@ -0,0 +1,200 @@
...
+class EB_ELPA(ConfigureMake):
+    """Support for building/installing ELPA."""
+
+    @staticmethod
+    def extra_options():
+        """Custom easyconfig parameters for ELPA."""
+        extra_vars = {
+            'auto_detect_cpu_features': [True, "Auto-detect available CPU features, and configure accordingly", CUSTOM],
+            'with_mpi': [True, "Enable building of ELPA MPI library", CUSTOM],
+            'with_openmp': [True, "Enable building of ELPA OpenMP library", CUSTOM],
```
boegel: @micaeljtoliveira We have `openmp`, `usempi`, `shared` in `toolchainopts`, better rely on that rather than introducing custom easyconfig parameters specific to ELPA (where possible)? See https://easybuild.readthedocs.io/en/latest/version-specific/toolchain_opts.html#foss

micaeljtoliveira: @boegel I based my implementation on the FFTW easyblock, which provides these parameters, In any case, I don't see any problem in removing them.

ocaisa: The toolchainopt `usempi` may  have an unwanted side-effect (particuarly in the serial case), it switches `CC` to `mpicc` (from `gcc` for `foss`). Given that you are building all the possibilities in different directories, I'd be more inclined to keep these options as a feature of ELPA (they don't really have the same purpose as the `toolchainopts` which are about influencing a particular build, here you are selecting whether or not to *include* a particular build)

micaeljtoliveira: @ocaisa Thanks for the explanation on the purpose of the `toolchainopts`. That clarifies the logic behind some choices in the easyblocks that were not very clear to me.

micaeljtoliveira:
@boegel @ocaisa: As discussed last week, ideally the framework should be changed so that it is possible to iterate over a list of toolchain options just like we do for `configopts` and `buildopts`. This way we could set different values for the `usempi` toolchain option when building in serial or with mpi.

damianam:
I like very much the idea of having an easyblock for ELPA. I also think it makes sense to keep those options as ELPA options, and not rely on `toolchainopts`. The libraries have different names, so having them in a single prefix is a logical way to proceed.

@micaeljtoliveira what about GPU support?

micaeljtoliveira:
@damianam GPU support would be great, but I don't have the expertise nor time to implement it. But I would be very happy if some one wants to have a go at it.

@boegel Any chance of having this PR accepted soon? I know it's not ideal, having the serial version of ELPA compiled with MPI toolchain option, but it's still a clear improvement over the current situation.

boegel:
@micaeljtoliveira A couple of issues were reporting on the mailing list with the proposed ELPA easyblock, see https://lists.ugent.be/wws/arc/easybuild/2019-04/msg00018.html and https://lists.ugent.be/wws/arc/easybuild/2019-04/msg00014.html .

Can you look into those? Do shout if you need help...

From @ocaisa's remark, I understand that there's no longer a need to be able to iterate over a list of `toolchainopts`, the `with_mpi` and `with_openmp` approaches are fine?

micaeljtoliveira:
@boegel I've seen the emails. Fixing the AVX issue should be trivial if indeed it only requires changing `avx512` to `avx512f`.

> I understand that there's no longer a need to be able to iterate over a list of toolchainopts, the with_mpi and with_openmp approaches are fine?

Not quite, unfortunately. If I understood correctly, there might be a problem with some toolchains when building a serial application with the mpi compiler wrappers. This will be the case when using the  `usempi` toolchain option for both serial and mpi builds of the library.

akesandgren:
We should preferably get this either merged or rejected before 3.9.1.
What is the currecnt status of it? Does it work as is? Should I run some tests?

boegel:
@akesandgren I think this needs support for iterating of a list of toolchain options, so we can have `usempi` set to `False` when building the serial part, but have `usempi` set to `True` for the parallel part.

On the other hand, I'm not sure that's actually needed at all... `usempi` is a convenience mechanism available for MPI software that doesn't properly know how to get built, but I'm guessing that's not a problem for ELPA.
If so, then `usempi` should never be set to `True`.

@micaeljtoliveira Can you shed some light on this?

micaeljtoliveira:
@akesandgren @boegel The original idea was to build and install all the different flavors of the library (serial, MPI, OpenMP, MPI+OpenMP) with just one easyconfig, in a somehow similar way to what is done for FFTW. Unfortunately the way how ELPA is compiled would require to iterate over toolchain options, as mentioned by @boegel, which is currently not supported.

I think the best is to simplify things so that this can be merged. I will remove the iteration over `usempi = True/False`  Then users who want to build the serial version can simply tweak the corresponding easyconfig. I will also fix one issue reported sometime ago on the mailing list regarding the AVX512 support.

micaeljtoliveira:
@boegel I've updated the easyblock. Now it does not loop over MPI on/off anymore. This way there's no need to iterate over toolchain options and the choice of compiling with or without MPI support should be done in the corresponding easyconfig. I've also fixed a couple of other problems, including the AVX512 issue reported on the mailing list.

boegel:
@micaeljtoliveira Can you clarify what would need to be done in the easyconfig file to build ELPA with MPI support?

Also, why did you also drop the custom `sanity_check_step` implementation?
I feel it's important to have a custom sanity check for a software-specific easyblock like this one...

micaeljtoliveira:
@boegel 

> @micaeljtoliveira Can you clarify what would need to be done in the easyconfig file to build ELPA with MPI support?

You need to set the toolchain option `usempi` to `True`. For both serial and MPI, it is also necessary to specify the `LIBS` environment variable in order to link with the correct linear algebra library (LAPACK or SCALAPACK).
 
> Also, why did you also drop the custom `sanity_check_step` implementation?
> I feel it's important to have a custom sanity check for a software-specific easyblock like this one...

Since the folders and names of the libraries will depend on the value of `usempi`, I thought I was best that this was specified in the easyconfig.

micaeljtoliveira:
@boegel :

> > Since the folders and names of the libraries will depend on the value of usempi,
> > I thought I was best that this was specified in the easyconfig.
> 
> I would prefer having this in the easyblock (which is do once and forget), rather than having to remember to update the easyconfigs according to having `usempi` set or not, that'll be annoying...

No problem, I will add the check back. I'm still learning about the preferred/best way of doing things in EasyBuild, so you need to have a little patience ;)

easybuild/easyblocks/e/elpa.py
```diff
@@ -0,0 +1,227 @@
...
+        for with_omp in with_omp_opts:
+            if with_omp:
+                omp_configopt = '--enable-openmp'
+            else:
+                omp_configopt = '--disable-openmp'
+
+            # append additional configure and build options
+            self.cfg.update('configopts',
+                            [omp_configopt + ' ' + common_config_opts])
+            self.cfg.update('buildopts', [common_build_opts])
```
boegel: This can be simplified imho, since there can only be two values in `with_omp_opts`:

```python
self.cfg.update('configopts', ['--disable-openmp ' + common_config_opts])

if self.toolchain.options.get('openmp', False):
    self.cfg.update('configopts', ['--enable-openmp ' + common_config_opts])

File: easybuild/easyblocks/e/elpa.py [hunk 3]
```diff
@@ -0,0 +1,227 @@
...
+        self.cfg.update('buildopts', 'V=1')
+
+        # keep track of common configopts and of configopts specified
+        # in easyconfig file, so we can include them in each iteration
+        # later
+        common_config_opts = self.cfg['configopts']
+        common_build_opts = self.cfg['buildopts']
+
+        self.cfg['configopts'] = []
+        self.cfg['buildopts'] = []
```
boegel: There's no need to also make `buildopts` a list here, since the build options are always the same.

If `buildopts` is just a string, the same value will be used for every iteration by default.

Less code to worry about :)

easybuild/easyblocks/e/elpa.py
```diff
@@ -0,0 +1,227 @@
...
+        with_omp_opts = [False]
+        if self.cfg['with_openmp']:
+            with_omp_opts.append(True)
+
+        if self.toolchain.options.get('usempi', None):
+            mpi_suff = ''
+        else:
+            mpi_suff = '_onenode'
+
+        for with_omp in with_omp_opts:
```
boegel:
```python
for with_omp in [False, toolchain.options.get('openmp', False)]:

File: easybuild/easyblocks/e/elpa.py [hunk 5]
```diff
@@ -0,0 +1,227 @@
...
+            'dirs': ['lib/pkgconfig', 'bin'],
+        }
+
+        shlib_ext = get_shared_lib_ext()
+
+        extra_files = []
+
+        with_omp_opts = [False]
+        if self.cfg['with_openmp']:
+            with_omp_opts.append(True)
```
boegel: No need for this, see below

boegel:
@micaeljtoliveira One thing that confuses me a bit is that the current `ELPA` easyconfigs we have include `usempi`: True`, yet `sanity_check_paths` includes `lib/libelpa_openmp.a` rather than `lib/libelpa_onenode_openmp.a`... So were we actualy building `ELPA` without MPI support all along?

micaeljtoliveira:
> @micaeljtoliveira One thing that confuses me a bit is that the current `ELPA` easyconfigs we have include `usempi: True`, yet `sanity_check_paths` includes `lib/libelpa_openmp.a` rather than `lib/libelpa_onenode_openmp.a`... So were we actualy building `ELPA` without MPI support all along?

@boegel  The naming convention used by ELPA is somewhat confusing: the `openmp` suffix means that it was compiled with OpenMP and MPI, while the `onenode_openmp` suffix means it was compiled with OpenMP and **without** MPI.

easybuild/easyblocks/e/elpa.py
```diff
@@ -0,0 +1,227 @@
...
+
+        if self.toolchain.options.get('usempi', None):
+            mpi_suff = ''
+        else:
+            mpi_suff = '_onenode'
```
boegel: @micaeljtoliveira Please clarify what the `onenode` means exactly (i.e. no MPI support) with a comment above, since this is confusing. ;)

easybuild/easyblocks/e/elpa.py
```diff
@@ -0,0 +1,227 @@
...
+        # later
+        common_config_opts = self.cfg['configopts']
+        common_build_opts = self.cfg['buildopts']
+
+        self.cfg['configopts'] = []
+        self.cfg['buildopts'] = []
+
+        with_omp_opts = [False]
+        if self.cfg['with_openmp']:
+            with_omp_opts.append(True)
```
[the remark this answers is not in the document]
micaeljtoliveira: @boegel No, there's probably no reason. This basically came from the FFTW easyblock, as you and others suggested I use it as a starting point.

easybuild/easyblocks/e/elpa.py
micaeljtoliveira: Yes, you are right. Those are "leftovers" from the previous version where we were also looping over MPI on/off.

easybuild/easyblocks/e/elpa.py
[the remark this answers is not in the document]
micaeljtoliveira: @boegel Looking at this again, it's not clear to me what the behavior of the easyblock should be regarding OpenMP.

At the moment two versions of the library are built, with and without OpenMP support, unless  `with_openmp` is explicitly set to `False`, in which case only the version without OpenMP support is built.

Now, if I change the easyblock the way you suggest to use the `openmp` toolchain option, what should be the behavior? Should the easyblock build both versions unless `openmp` is set explicitly to `False`? This is not the usual behavior of the `openmp` toolchain option. Do you think this could be confusing for users?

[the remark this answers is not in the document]
boegel: @micaeljtoliveira I don't think you can easily tell whether or not the `openmp` toolchain option was explicitly set to `False` or not, since the default value is `False` (and the `openmp` toolchain option is always defined as soon as there's a compiler component in the toolchain , so `self.toolchain.get('openmp')` will always return the default value).

I guess that answer the question: only building with OpenMP support when the `openmp` toolchain option is set to `True` makes sense.

easybuild/easyblocks/e/elpa.py
```diff
@@ -0,0 +1,210 @@
...
+                omp_suff = ''
+
+            extra_files.append('include/elpa%s%s-%s/elpa/elpa.h' % (mpi_suff, omp_suff, self.version))
+            extra_files.append('include/elpa%s%s-%s/modules/elpa.mod' % (mpi_suff, omp_suff, self.version))
+
+            extra_files.append('lib/libelpa%s%s.a' % (mpi_suff, omp_suff))
+            if self.cfg['with_shared']:
+                extra_files.append('lib/libelpa%s%s.%s' % (mpi_suff, omp_suff, shlib_ext))
+
+        custom_paths['files'] = nub(extra_files)
```
boegel: @micaeljtoliveira The `nub` isn't actually needed anymore, is it? I don't see how there can be any duplicates (anymore)?

If so, please remove that (and drop the `import` as well).

It's not a big issue to have it there, but since `nub` has been moved to another place for EasyBuild 4.0, we should get rid of it if it's not actually needed...

micaeljtoliveira: @boegel Actually it's possible to have duplicates and that's because of a change you requested ;)

The culprit is this line:
`for with_omp in [False, self.toolchain.options.get('openmp', False)]:`

If `openmp = False`, the loop will iterate twice over the value `False`.

boegel: Oh, right... So we do need `nub`, but I've move it up here:

```python
for with_omp in nub([False, self.toolchain.options.get('openmp', False)]):

File: easybuild/easyblocks/e/elpa.py [hunk 9]
```diff
@@ -0,0 +1,211 @@
...
+                    self.log.info("Enabling use of %s (should be supported based on CPU features)", flag.upper())
+                    setattr(self, flag, True)
+
+    def run_all_steps(self, *args, **kwargs):
+        """
+        Put configure options in place for different builds (with and without openmp).
+        """
+
+        # make all builds verbose
+        self.cfg.update('preconfigopts', 'autoreconf &&')
```
boegel: @micaeljtoliveira How is this related to making the build verbose exactly?

Also, should we try and avoid picking up `autoreconf` from the OS here (i.e. requiring that `Autotools` is listed as a build dep)?

micaeljtoliveira: @boegel Ups, copy-paste strikes again. I'll fix it.

micaeljtoliveira: @boegel This prompted me to look if the call to `autoreconf` was really necessary. After a couple of tests, it seems not. Maybe it was at some point in some of the older easyconfigs, but AFAICS that's not the case anymore.

Issue #1466, opened 2018-07-25
Go packages installations


pelacables:
I needèd to install a go package and did not find any easyblock that support duch installation so I asked the mailing list for help and Alan O'Cais did provide a sample easyconfig (using the Binary easyblock) that installed the go package:

I leave it here for the record, in case someone else needs to install go package with EB and while the go easyblock is not available.

https://gist.github.com/ocaisa/1057ca936273db6715768fcb9b4b9191

```
easyblock = 'Binary'

name = 'Go-Example'
version = '20180724'
commit_id = '46695d8'

homepage = 'https://github.com/golang/example'
description = """This repository contains a collection of Go programs and libraries that demonstrate the language, standard libraries, and tools."""

toolchain = {'name': 'GCCcore', 'version': '7.3.0'}

source_urls = ['https://github.com/golang/example/archive/']
sources = ['%s.tar.gz' % commit_id]

buildininstalldir = True
extract_sources = True

builddependencies = [
    ('binutils', '2.30'),
]

dependencies = [
    ('Go', '1.10.3', '', True),
]

# The sources are expected in a very particular place
unpack_options = '&& mkdir -p %(installdir)s/src/github.com/golang&& mv example-* %(installdir)s/src/github.com/golang/example'


# Update GOPATH so install location is correct
install_cmd = "export GOPATH=%(installdir)s:$GOPATH &&"
# Update PATH to include new bin directory (in case it is needed)
install_cmd += "export PATH=%(installdir)s/bin:$PATH &&"

# Do the stringutils installation
install_cmd += "cd %(installdir)s/src/github.com/golang/example/stringutil &&"
install_cmd += "go build && go install &&"

# Do the hello installation
install_cmd += "cd %(installdir)s/src/github.com/golang/example/hello &&"
install_cmd += "go build && go install"

modextrapaths = {
    'GOPATH': ['.'],
}

sanity_check_paths = {
    'files': ['bin/hello'],
    'dirs': ['bin', 'src', 'pkg'],
}
``

PR #1629, opened 2019-02-06
Add generic easyblock for Go packages/applications

bedroge:
This is a generic easyblock that allows Go packages to be installed. It uses the "go get" command to get (the sources of) the dependencies, and to build and install everything.

I'm currently setting GOPATH to the build dir and GOBIN to the install dir: this means that only the compiled executables will be installed, while the sources will be wiped; since I'm not really a Go expert, does it make sense to do so? I think by doing so (and setting GOPATH in the modulefile), it would allow other modules to reuse code from this module. But since it's not really clear which dependencies the "go get" command installs, for now I preferred to just remove all sources.

Furthermore, it's basically now just compiling the Go files, and you don't need the Go module for running them. There's a check in the easyblock that raises an error if there is no Go module in the (build)dependencies. Is this the right approach, or should this for instance be added as a toolchain instead?

I've tested this easyblock with a SeqKit easyconfig (also see https://github.com/easybuilders/easybuild-easyconfigs/pull/6885):
```
easyblock = 'GoPackage'

name = 'SeqKit'
version = '0.10.0'

homepage = 'https://bioinf.shenwei.me/seqkit/'
description = """SeqKit - a cross-platform and ultrafast toolkit for FASTA/Q file manipulation"""

source_urls = ['https://github.com/shenwei356/%(namelower)s/archive/']
sources = ['v%(version)s.tar.gz']

builddependencies = [('Go', '1.11.5')]

sanity_check_paths = {
    'files': ['%(namelower)s'],
    'dirs': [],
}

moduleclass = 'bio'
```

And with a dummy easyconfig that compiles this golang-example project:
https://github.com/buildkite/golang-example

For both of them, this easyblock worked, though only with a recent version of Go (1.11.5). It did not work with Go 1.8.1; I suspect this is due to some changes in the "go get" command that were introduced in newer version.

### DIFF
--- a/easybuild/easyblocks/generic/gopackage.py
+++ b/easybuild/easyblocks/generic/gopackage.py
@@ -1,0 +1,74 @@
+# -*- coding: utf-8 -*-
+##
+# Copyright 2009-2018 Ghent University
+#
+# This file is part of EasyBuild,
+# originally created by the HPC team of Ghent University (http://ugent.be/hpc/en),
+# with support of Ghent University (http://ugent.be/hpc),
+# the Flemish Supercomputer Centre (VSC) (https://www.vscentrum.be),
+# Flemish Research Foundation (FWO) (http://www.fwo.be/en)
+# and the Department of Economy, Science and Innovation (EWI) (http://www.ewi-vlaanderen.be/en).
+#
+# https://github.com/easybuilders/easybuild
+#
+# EasyBuild is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation v2.
+#
+# EasyBuild is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with EasyBuild.  If not, see <http://www.gnu.org/licenses/>.
+##
+"""
+EasyBuild support for building and installing Go packages, implemented as an easyblock
+
+@author: Bob Dröge (University of Groningen)
+"""
+import os
+import shutil
+
+from easybuild.framework.easyconfig import CUSTOM
+from easybuild.framework.extensioneasyblock import ExtensionEasyBlock
+from easybuild.tools.build_log import EasyBuildError
+from easybuild.tools.filetools import mkdir
+from easybuild.tools.run import run_cmd, parse_log_for_error
+from easybuild.tools.modules import get_software_root
+
+class GoPackage(ExtensionEasyBlock):
+    """
+    Install a Go package as a separate module, or as an extension.
+    """
+
+    def configure_step(self):
+        """Custom configure: fetch the sources of the dependencies."""
+        cmd = "env GOPATH=%s go get -d ./..." % (self.builddir)
+        run_cmd(cmd)
+
+    def build_step(self):
+        """Build the Go packages."""
+        cmd = "env GOPATH=%s go build ./..." % (self.builddir)
+        run_cmd(cmd)
+
+    def install_step(self):
+        """Install procedure for Go packages."""
+        cmd = "env GOBIN=%s GOPATH=%s go install ./..." % (self.installdir, self.builddir)
+        run_cmd(cmd)
+
+    def make_module_req_guess(self):
+        """Add the installation directory to PATH."""
+        guesses = super(GoPackage, self).make_module_req_guess()
+        guesses['PATH'] = ['']
+        return guesses
+
+    def make_module_req_guess(self):
+        """Customized dictionary of paths to look for with PERL*LIB."""
+
+        guesses = super(GoPackage, self).make_module_req_guess()
+        guesses.update({
+            "PATH" : [''],
+        })
+        return guesses
--- a/easybuild/easyblocks/generic/gopackage.py
+++ b/easybuild/easyblocks/generic/gopackage.py
@@ -31,11 +31,9 @@ EasyBuild support for building and installing Go packages, implemented as an eas
 import os
 import shutil
 
-from easybuild.framework.easyconfig import CUSTOM
 from easybuild.framework.extensioneasyblock import ExtensionEasyBlock
 from easybuild.tools.build_log import EasyBuildError
-from easybuild.tools.filetools import mkdir
-from easybuild.tools.run import run_cmd, parse_log_for_error
+from easybuild.tools.run import run_cmd
 from easybuild.tools.modules import get_software_root
 
 class GoPackage(ExtensionEasyBlock):
@@ -45,6 +43,8 @@ class GoPackage(ExtensionEasyBlock):
 
     def configure_step(self):
         """Custom configure: fetch the sources of the dependencies."""
+        if not get_software_root("Go"):
+            raise EasyBuildError("Go packages require to have a Go module in the builddependencies.")
         cmd = "env GOPATH=%s go get -d ./..." % (self.builddir)
         run_cmd(cmd)
 
--- a/easybuild/easyblocks/generic/gopackage.py
+++ b/easybuild/easyblocks/generic/gopackage.py
@@ -44,7 +44,7 @@ class GoPackage(Tarball):
     def install_step(self):
         """Install procedure for Go packages."""
         if not get_software_root("Go"):
-            raise EasyBuildError("Go packages require to have a Go module in the builddependencies.")
+            raise EasyBuildError("Go packages require to have a Go module in the (build)dependencies.")
         cmd = "env GOBIN=%s GOPATH=%s go get ./..." % (self.installdir, self.builddir)
         run_cmd(cmd)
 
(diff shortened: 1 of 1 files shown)


wpoely86:
I've got no experience with installing Go stuff but this looks fine although al little magical? The `go get` does everything?

wpoely86:
Maybe add a default sanity check that the install dir should not be empty?

bedroge:
I'm not too familiar with installing Go packages either, but "go get" should do the trick. According to the Go documentation, this replaced the "goinstall" command that was used in earlier versions. I've created this easyblock in particular for Seqkit, which also has "go get" as one of the installation methods:
https://github.com/shenwei356/seqkit#method-4-for-go-developer-latest-stabledev-version

This easyblock downloads the source and runs go get in the source directory, but you could even do it directly by passing a github url as described on that web page.

Regarding the sanity check: how would I do this? For instance, SeqKit only installs a binary "seqkit", and I guess that most packages would do something similar. But I don't know if I can assume that this is always true (?). Or should I just check if there is something in the install dir besides the "easybuild" dir?

ocaisa:
Just for reference: #1466

easybuild/easyblocks/generic/gopackage.py
```diff
@@ -0,0 +1,53 @@
...
+from easybuild.tools.run import run_cmd
+from easybuild.tools.modules import get_software_root
+
+
+class GoPackage(Tarball):
+    """
+    Install a Go package as a separate module, or as an extension.
+    """
+
+    def install_step(self):
```
damianam: Since go compiles the packages, wouldn't it make sense to do this in the `build_step`?

easybuild/easyblocks/generic/gopackage.py
```diff
@@ -0,0 +1,53 @@
...
+
+@author: Bob Dröge (University of Groningen)
+"""
+from easybuild.easyblocks.generic.tarball import Tarball
+from easybuild.tools.build_log import EasyBuildError
+from easybuild.tools.run import run_cmd
+from easybuild.tools.modules import get_software_root
+
+
+class GoPackage(Tarball):
```
damianam: Why `Tarball`? Doesn't it make more sense to derive it from `EasyBlock`? After all, `Tarball` just defines 2 steps, one is here redefined and the other one is irrelevant (why skipping rpath checks here?)

easybuild/easyblocks/generic/gopackage.py
```diff
@@ -0,0 +1,53 @@
...
+        """Install procedure for Go packages."""
+        if not get_software_root("Go"):
+            raise EasyBuildError("Go packages require to have a Go module in the (build)dependencies.")
+        cmd = "env GOBIN=%s GOPATH=%s go get ./..." % (self.installdir, self.builddir)
+        run_cmd(cmd)
+
+    def make_module_req_guess(self):
+        """Add the installation directory to PATH."""
+        guesses = super(GoPackage, self).make_module_req_guess()
+        guesses['PATH'] = ['']
```
damianam: Doesn't it make more sense to install the binaries in `bin/`?

ocaisa: I thought the packages were installed in `bin` actually

ocaisa:
This could be split into 3 parts that match up to the EB steps:
```
go get -d ...    # Download packages only
go build         # Build
go install       # Install
```
which would allow us to tackle each part individually. The output of `go get -help` is pretty useful here, e.g.,
```
When checking out a new package, get creates the target directory
GOPATH/src/<import-path>. If the GOPATH contains multiple entries,
get uses the first one. For more details see: 'go help gopath'.
```
Which means we can probably manipulate GOPATH so that it checks out in one place and builds/installs in another.

ocaisa:
```
If the GOBIN environment variable is
set, commands are installed to the directory it names instead
of DIR/bin. GOBIN must be an absolute path.
```

easybuild/easyblocks/generic/gopackage.py
bedroge: I think I did this because the install_step somehow removed everything that was written to installdir by build_step, even if I override the install_step and let it do nothing. Is there a way to prevent that?

easybuild/easyblocks/generic/gopackage.py
bedroge: Probably makes more sense, indeed. I think I did this because I only needed an install_step and Tarball looked similar to what I needed.

easybuild/easyblocks/generic/gopackage.py
bedroge: Because I override $GOBIN, it does not make a bin directory (see comment from @ocaisa below). But I can easily add this and set $GOBIN to installdir/bin.

bedroge:
> This could be split into 3 parts that match up to the EB steps:
> 
> ```
> go get -d ...    # Download packages only
> go build         # Build
> go install       # Install
> ```

I think I initially did it like this (see older commit in this PR), but thought it would be easier to just run one command that does everything. I can change this back, though?

ocaisa:
Yeah, I see that now. Maybe you are right to keep it handled with `go get ...`, everything needs to be in `GOPATH` anyway. But if you do it like that you should be prepending the final installation directory to "GOPATH" so that anything that depends on this doesn't reinstall all the packages, something like:
```
    def make_module_extra(self):
        """Correctly prepend $GOPATH for Go packages."""
        txt = super(EB_Go, self).make_module_extra()
        txt += self.module_generator.prepend_paths("GOPATH", [''])
        return txt
```

I guess what I don't like is that any easyconfigs that use this are not reproducible, what packages go downloads depends on when you download.

ocaisa:
And your use of `GOPATH` in the easyblock should account for the fact that GOPATH may already have a value:
```
GOPATH=%s:$GOPATH
```

ocaisa:
Sorry, I'm spamming this thread, I see you only put the executables in the install for, so GOPATH is not relevant

easybuild/easyblocks/generic/gopackage.py
damianam: I think that'd make sense. It would be what anybody would expect I guess.

easybuild/easyblocks/generic/gopackage.py
damianam: I don't think the `install_step` does that if you derive from `EasyBlock`. But the cleanup might. Take a look at the `buildininstalldir` option, that might be what you need.

bedroge: It still seems to do this when deriving from EasyBlock. Adding buildinstalldir=True puts all the source files in the installdir, which is not what we want. The issue seems to be that the install_step cleans the installdir (because it might be an old installation), and hence removes the executable that the build_step installed. I guess we can prevent this with keeppreviousinstall=True, but I don't think that's a good idea. Any other ideas?

bedroge:
I found another issue: go get also seems to update the source of the thing we're actually installing, so it basically ignored whatever version you put in the easyconfig... :(

I'll try looking for a (easy) solution; it's probably possible to move the source to somewhere in the $GOPATH, but it has to have the right directory structure then (src/github.com/git_repo/project), which might be a bit tricky. If anyone knows a better/easier way to do this, please let me know.

ocaisa:
Yes, the GOPATH solution was what I did in https://gist.github.com/ocaisa/1057ca936273db6715768fcb9b4b9191 and the directory structure is indeed a pain.

bedroge:
I just tried to test that method manually for SeqKit (this is the package for which I started writing this easyblock) by doing a go get first to get the dependencies, and then overwriting the downloaded seqkit folder by the version that I want. However, that still does not work, because SeqKit has dependencies on, for instance:
github.com/shenwei356/bio
But this repo was changed for the latest SeqKit version, does not work with the older SeqKit anymore, and does not have versions, tags, whatever. So how does Go ever allow you to install a specific version of a package, if it's not clear at all which dependencies it needs? Apparently it's only possible to always download/install the most recent version of everything, unless you know how/where to fetch all the dependencies yourself and put them all at the right place? Then I don't think it makes sense to have an easyblock for this at all...

bedroge:
I've been thinking about other approaches to solve these issues: instead of using "go get" to determine and fetch all the dependencies, we could define all the dependencies (other Go libraries) in the easyconfig and add some functionality in the GoPackage easyblock that basically looks at the source url and uses that to determine the right directory structure. Most of the Go stuff seems to be on Github, and for unversioned dependencies we would need commit hashes or something in order to get all the right versions in place. This should probably work and would also make things reproducible (which is the whole point of an easyconfig anyway), but it might take a lot of time to write easyconfigs for such packages.

Do you think this is the right way to do it and is it worth looking into this, or should we just drop this whole GoPackage idea?

ocaisa:
Something to control dep versions is in the pipeline according to https://github.com/alco/gostart#16-how-do-i-freeze-packages-when-deploying

ocaisa:
It's interesting to look at how that linked `dep` package recommends  the install, https://golang.github.io/dep/docs/installation.html#install-from-source

bedroge:
Apparently Go itself also has a similar mechanism for this:
https://github.com/golang/go/wiki/Modules

But then the thing we want to compile in an easyconfig should have this additional information about supported versions of dependencies, which a lot do not seem to have (yet)...

akesandgren:
@bedroge Any conclusion on this?

bedroge:
> @bedroge Any conclusion on this?

I did not find a (good) way to install all dependencies of a Go package in a reproducible way, so I'll close this PR.

PR #1631, opened 2019-02-12
update ANSYS easyblock to deal with changed versioning scheme

boegel:


akesandgren:
Going in, thanks @boegel!

PR #1632, opened 2019-02-13
Add extra path to builddir for SNPhylo >= 20160204

vanzod:
The newest version of `SNPhylo` builds in a directory with the same name and must be added to the source path of the `shutil.copy2` and `shutil.copytree` commands to correctly find the files to install.

### DIFF
--- a/easybuild/easyblocks/s/snphylo.py
+++ b/easybuild/easyblocks/s/snphylo.py
@@ -36,7 +36,7 @@ from distutils.version import LooseVersion
 
 from easybuild.framework.easyblock import EasyBlock
 from easybuild.tools.build_log import EasyBuildError
-from easybuild.tools.filetools import adjust_permissions, mkdir
+from easybuild.tools.filetools import adjust_permissions, copy_dir, copy_file, mkdir
 from easybuild.tools.modules import get_software_root, get_software_version
 from easybuild.tools.run import run_cmd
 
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/s/snphylo.py
```diff
@@ -76,11 +77,16 @@ def install_step(self):
...
+        if LooseVersion(self.version) >= LooseVersion('20160204'):
+            predir = 'SNPhylo'
+
         try:
             mkdir(bindir, parents=True)
             for binfile in binfiles:
-                shutil.copy2(os.path.join(self.builddir, binfile), bindir)
-            shutil.copytree(os.path.join(self.builddir, 'scripts'), os.path.join(self.installdir, 'scripts'))
+                shutil.copy2(os.path.join(self.builddir, predir, binfile), bindir)
+            shutil.copytree(os.path.join(self.builddir, predir, 'scripts'), os.path.join(self.installdir, 'scripts'))
```
boegel: @vanzod Let's also use `copy_file` and `copy_dir` while we're touching this easyblock (and remove the `try`/`except`):

```python
mkdir(bindir, parents=True)
for binfile in binfiles:
    copy_file(os.path.join(self.builddir, binfile), bindir)
copy_dir(os.path.join(self.builddir, predir, 'scripts'), os.path.join(self.installdir, 'scripts'))

boegel:
lgtm, thanks @vanzod!

PR #1633, opened 2019-02-13
fix finding of Python 3 include subdir + libpython*.so in OpenBabel easyblock

boegel:


akesandgren:
Going in, thanks @boegel!

PR #1634, opened 2019-02-16
Add better support for TensorRT to TensorFlow.

akesandgren:
Required for https://github.com/easybuilders/easybuild-easyconfigs/pull/7246

### DIFF
--- a/easybuild/easyblocks/t/tensorflow.py
+++ b/easybuild/easyblocks/t/tensorflow.py
@@ -208,6 +208,7 @@ class EB_TensorFlow(PythonPackage):
             'TF_NEED_OPENCL': ('0', '1')[bool(opencl_root)],
             'TF_NEED_OPENCL_SYCL': '0',
             'TF_NEED_S3': '0',  # Amazon S3 File System
+            'TF_NEED_TENSORRT': '0',
             'TF_NEED_VERBS': '0',
             'TF_NEED_AWS': '0',  # Amazon AWS Platform
             'TF_NEED_KAFKA': '0',  # Amazon Kafka Platform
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/t/tensorflow.py
```diff
@@ -209,7 +209,6 @@ def configure_step(self):
             'TF_NEED_OPENCL_SYCL': '0',
             'TF_NEED_S3': '0',  # Amazon S3 File System
             'TF_NEED_VERBS': '0',
-            'TF_NEED_TENSORRT': ('0', '1')[bool(tensorrt_root)],
```
boegel: @akesandgren Better set to `0` by default, to avoid surprises?

akesandgren: Done

easybuild/easyblocks/t/tensorflow.py
```diff
@@ -208,8 +208,8 @@ def configure_step(self):
             'TF_NEED_OPENCL': ('0', '1')[bool(opencl_root)],
             'TF_NEED_OPENCL_SYCL': '0',
             'TF_NEED_S3': '0',  # Amazon S3 File System
-            'TF_NEED_VERBS': '0',
             'TF_NEED_TENSORRT': ('0', '1')[bool(tensorrt_root)],
```
boegel: @akesandgren I really mean `0` here, since it's revisited correctly later:

```suggestion
             'TF_NEED_TENSORRT': '0',

akesandgren: It isn't changed below anymore.

akesandgren: But you're right. It shouldn't be set to 1 unless cuda is set. I'll change it back again.

boegel:
Thanks @akesandgren!

PR #1635, opened 2019-02-19
Add support for newer versions of wxPython to wxpython.py.

akesandgren:
Required for https://github.com/easybuilders/easybuild-easyconfigs/pull/7679

### DIFF
--- a/easybuild/easyblocks/w/wxpython.py
+++ b/easybuild/easyblocks/w/wxpython.py
@@ -47,7 +47,10 @@ class EB_wxPython(PythonPackage):
         if LooseVersion(self.version) >= LooseVersion("4"):
             prebuild_opts = self.cfg['prebuildopts']
             script = 'build.py'
-            wxflag = ''
+            self.wxflag = ''
+            if get_software_root('wxWidgets'):
+                self.wxflag = '--use_syswx'
+
             BUILD_CMD = "%(prebuild_opts)s %(pycmd)s %(script)s --prefix=%(prefix)s -v"
             base_cmd = BUILD_CMD % {
                 'prebuild_opts': prebuild_opts,
@@ -55,13 +58,13 @@ class EB_wxPython(PythonPackage):
                 'script': script,
                 'prefix': self.installdir,
             }
-            if get_software_root('wxWidgets'):
-                wxflag = '--use_syswx'
-            else:
+
+            # Do we need to build wxWidgets inetrnally?
+            if self.wxflag == '':
                 cmd = base_cmd + " build_wx"
                 run_cmd(cmd, log_all=True, simple=True)
 
-            cmd = base_cmd + " %s build_py" % wxflag
+            cmd = base_cmd + " %s build_py" % self.wxflag
             run_cmd(cmd, log_all=True, simple=True)
 
     def install_step(self):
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/w/wxpython.py
```diff
@@ -41,41 +43,106 @@ class EB_wxPython(PythonPackage):
...
+        """Custom install procedure for wxPython."""
         # wxPython configure, build, and install with one script
-        script = os.path.join('wxPython', 'build-wxpython.py')
-        cmd = "{0} {1} --prefix={2} --wxpy_installdir={2} --install".format(self.python_cmd, script, self.installdir)
+        preinst_opts = self.cfg['preinstallopts']
+        INSTALL_CMD = "%(preinst_opts)s %(pycmd)s %(script)s --prefix=%(prefix)s"
+        if LooseVersion(self.version) >= LooseVersion("4"):
+            wxflag = ''
+            if get_software_root('wxWidgets'):
+                wxflag = '--use_syswx'
```
wpoely86: if you do this check multiple times, do it once and put it in a class variable?

easybuild/easyblocks/w/wxpython.py
```diff
@@ -41,41 +43,106 @@ class EB_wxPython(PythonPackage):
...
         majver = '.'.join(self.version.split('.')[:2])
         shlib_ext = get_shared_lib_ext()
-        py_bins = ['alacarte', 'alamode', 'crust', 'shell', 'wrap', 'wxrc']
+        py_bins = ['crust', 'shell', 'wxrc']
+        files = []
+        dirs = []
+        if LooseVersion(self.version) < LooseVersion("4"):
+            files.extend([os.path.join('bin', 'wxrc')])
+            dirs.extend(['include', 'share'])
+            py_bins.extend(['alacarte', 'alamode', 'wrap'])
```
wpoely86: why not add `wxrc` here too?

easybuild/easyblocks/w/wxpython.py
```diff
@@ -41,41 +43,106 @@ class EB_wxPython(PythonPackage):
...
         }
 
         # test using 'import wx' (i.e. don't use 'import wxPython')
         self.options['modulename'] = 'wx'
 
-        # also test importing wxversion
-        custom_commands = [(self.python_cmd, '-c "import wxversion"')]
+        if LooseVersion(self.version) < LooseVersion("4"):
+            # also test importing wxversion
+            custom_commands = [(self.python_cmd, '-c "import wxversion"')]
```
wpoely86: this doesn't work anymore in version > 4?

easybuild/easyblocks/w/wxpython.py
akesandgren: Ask the person who wrote the ver<4 part.... but probably because there is a wxrc binary and not just a pywxrc? And it is already in py_bins a couple of lines higher up.

easybuild/easyblocks/w/wxpython.py
akesandgren: Nope, wxversion doesn't exist in 4.x

easybuild/easyblocks/w/wxpython.py
```diff
@@ -41,41 +43,106 @@ class EB_wxPython(PythonPackage):
...
+
+            BUILD_CMD = "%(prebuild_opts)s %(pycmd)s %(script)s --prefix=%(prefix)s -v"
+            base_cmd = BUILD_CMD % {
+                'prebuild_opts': prebuild_opts,
+                'pycmd': self.python_cmd,
+                'script': script,
+                'prefix': self.installdir,
+            }
+
+            # Do we need to build wxWidgets inetrnally?
```
boegel: @akesandgren please fix typo: `inetrnally` -> `internally`

boegel:
Tested with all existing `wxPython` easyconfigs, so good to go (easybuilders/easybuild-easyconfigs#7679 is still blocked due to dependencies for now).

PR #1637, opened 2019-02-21
add no_qa pattern for interactive installation command for ABAQUS

boegel:
This avoids that the installation is cancelled because the output "hangs" at what is assumed to be an unknown question, while it's actually still going...


akesandgren:
Going in, thanks @boegel!

PR #1638, opened 2019-02-21
update OCaml easyblock to support OPAM 2.x

boegel:


easybuild/easyblocks/o/ocaml.py
```diff
@@ -28,17 +28,50 @@
...
+def det_opam_version():
+    """Determine OPAM version (using 'opam --version')."""
+
+    opam_version_cmd = 'opam --version'
+    (out, ec) = run_cmd(opam_version_cmd)
+    if ec == 0:
+        res = re.search('^[0-9.]+$', out.strip())
+        if res:
+            opam_ver = res.group(0)
+        else:
```
wpoely86: this else needs an indent less no?

PR #1639, opened 2019-02-24
replace 'except OSError, err:' pattern with 'except OSError as err:' (for compatibility with Python 3.x)

boegel:


easybuild/easyblocks/i/imkl.py
```diff
@@ -253,11 +253,8 @@ def post_install_step(self):
             fftw3libs = ['fftw3xc', 'fftw3xf']
 
             interfacedir = os.path.join(self.installdir, intsubdir)
-            try:
-                os.chdir(interfacedir)
-                self.log.info("Changed to interfaces directory %s" % interfacedir)
-            except OSError, err:
-                raise EasyBuildError("Can't change to interfaces directory %s", interfacedir)
+            change_dir(interfacedir)
+            self.log.info("Changed to interfaces directory %s" % interfacedir)
```
wpoely86: `self.log.info("Changed to interfaces directory %s", interfacedir)`

easybuild/easyblocks/m/metis.py
```diff
@@ -109,7 +109,7 @@ def install_step(self):
                     dst = os.path.join(includedir, f)
                     shutil.copy2(src, dst)
                     os.chmod(dst, 0755)
```
wpoely86: correct this one too while you're at it?

PR #1640, opened 2019-02-28
make easyblock tests pass on top of Python 3.x

boegel:


test/easyblocks/general.py
```diff
@@ -98,17 +99,15 @@ def test_custom_easyblocks_repo(self):
...
         # see also http://blog.olgabotvinnik.com/blog/2014/03/03/2014-03-03-pythonpath-is-a-liar-site-py-and-easy-install-pth-tell/
         import easybuild.framework
         framework_path = up(easybuild.framework.__file__, 3)
-        import vsc.utils.generaloption
-        vsc_path = up(vsc.utils.generaloption.__file__, 3)
 
         # prepend path to easybuild-easyblocks repo to $PYTHONPATH, so we're in full(?) control
         pythonpaths = os.environ.get('PYTHONPATH', '').split(os.pathsep)
-        os.environ['PYTHONPATH'] = os.pathsep.join([easyblocks_path, framework_path, vsc_path])
+        os.environ['PYTHONPATH'] = os.pathsep.join([easyblocks_path, framework_path] + pythonpaths)
```
wpoely86: Wait, how did this work before? `pythonpaths` was not used?

wpoely86: I'm also not sure I understand why the split/join on pathsep is needed?

easybuild/easyblocks/e/easybuildmeta.py
```diff
@@ -159,8 +159,7 @@ def sanity_check_step(self):
         for tool in eb_dirs.keys():
             self.log.debug("Trying %s.." % tool)
             try:
-                exec "from %s import setup" % tool
-                del setup
+                exec("from %s import setup" % tool)
```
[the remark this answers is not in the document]
boegel: It never was really needed, it's cleanup, but it fails with Python 3, and it's tricky to get right (because you have to try hard to make `exec` actually define things in the current scope)

easybuild/easyblocks/f/fftw.py
```diff
@@ -28,15 +28,16 @@
 @author: Kenneth Hoste (HPC-UGent)
 """
 from distutils.version import LooseVersion
-from vsc.utils.missing import nub
 
 from easybuild.easyblocks.generic.configuremake import ConfigureMake
 from easybuild.framework.easyconfig import CUSTOM
 from easybuild.toolchains.compiler.gcc import TC_CONSTANT_GCC
+from easybuild.tools.build_log import EasyBuildError
```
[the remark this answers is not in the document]
boegel: It didn't work, the import was always missing, it's just that we didn't notice.

Perhaps this should be a separate PR, since it's a silly bug that's also relevant for current `master` & `develop`. I'll do exactly that.

test/easyblocks/general.py
boegel: It worked fine before despite the bug of basically ignoring the current `$PYTHONPATH` because if you had easyblocks, framework and `vsc-base` in `$PYTHONPATH`, that was fine.

I guess there's little point in doing the `split`, but it doesn't hurt either?

easybuild/easyblocks/f/fftw.py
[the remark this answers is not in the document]
boegel: fixed in #1641

test/easyblocks/general.py
wpoely86: Doesn't it make things needlessly complex?

PR #1643, opened 2019-03-01
add new custom easyblock for CUTEst

boegel:


easybuild/easyblocks/c/cutest.py
```diff
@@ -0,0 +1,227 @@
...
+        """Custom prepare step for CUTEst."""
+        super(EB_CUTEst, self).prepare_step()
+
+        # determine subdirectories for each component
+        for comp in COMPONENTS:
+            paths = glob.glob(os.path.join(self.installdir, comp + '-*'))
+            if len(paths) == 1:
+                self.subdirs[comp] = os.path.basename(paths[0])
+            else:
+                raise EasyBuildError("Failed to isolate %s subdirectory: %s", comp, paths)
```
wpoely86: you still need to add the case for dry run here

boegel: fixed in 849f9d9

akesandgren:
@boegel ping

boegel:
@akesandgren Consider this on hold for now, the `CUTEst` installation that we have now have some problems, it's not clear yet whether some tweaking will be needed here, so I'll mark this as `WIP`

PR #1645, opened 2019-03-03
Use the newly added pretestopts in test_step where applicable.

akesandgren:


akesandgren:
This makes https://github.com/easybuilders/easybuild-easyconfigs/pull/7762 work if one adds
```
pretestopts = 'export OMP_NUM_THREADS=1 && '
```

easybuild/easyblocks/generic/configuremake.py
```diff
@@ -278,7 +278,7 @@ def test_step(self):
         """
 
         if self.cfg['runtest']:
-            cmd = "make %s" % (self.cfg['runtest'])
+            cmd = "%s make %s" % (self.cfg['pretestopts'], self.cfg['runtest'])
```
boegel: @akesandgren Also pick up on `testopts` while you're at it?

```python
cmd = ' '.join([self.cfg['pretestopts'], "make %s" % self.cfg['runtest'], self.cfg['testopts']])

akesandgren: I left that out because you can basically set runtest to include any othe options you need to have.
So I'm not sure that there is a reason for testopts.

boegel: In some contexts, `runtest` can be a boolean (I think for Python packages), so we should allow both `runtest` and `testopts` to compose the test command, I see no reason not to.

akesandgren: Ok, will fix for the ones i changed already. Leaving the more complicated ones for later.

boegel:
lgtm, thanks @akesandgren!

PR #1646, opened 2019-03-03
Make serial, smpar and dm+sm builds of WRF work and pass tests.

akesandgren:
run_test only checks rsl.error.0000 which isn't created for a serial run.

### DIFF
--- a/easybuild/easyblocks/w/wrf.py
+++ b/easybuild/easyblocks/w/wrf.py
@@ -158,7 +158,7 @@ class EB_WRF(EasyBlock):
 
         # fetch selected build type (and make sure it makes sense)
         known_build_types = ['serial', 'smpar', 'dmpar', 'dm+sm']
-        self.parallel_build_types = ["dmpar", "smpar", "dm+sm"]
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/w/wrf.py
```diff
@@ -158,7 +158,7 @@ def configure_step(self):
 
         # fetch selected build type (and make sure it makes sense)
         known_build_types = ['serial', 'smpar', 'dmpar', 'dm+sm']
-        self.parallel_build_types = ["dmpar", "smpar", "dm+sm"]
+        self.parallel_build_types = ["dmpar", "dm+sm"]
```
boegel: @akesandgren This doesn't look OK to me... `smpar` is a parallel build type, isn't it?

akesandgren: No, the parallel build type only applies to MPI builds, basically it's used to decide which tests to run and if mpirun is to be used. It's not used for building stuff. SO a more correct name would have been mpi_build_types.

PR #1649, opened 2019-03-09
clean up Java easyblock to use filetools functionality

boegel:


akesandgren:
Going in, thanks @boegel!

PR #1650, opened 2019-03-11
 configure CP2K with correct DATA_DIR value rather than relying on $CP2K_DATA_DIR environment variable

boegel:
Relying on `$CP2K_DATA_DIR` is problematic in some contexts; for example, it requires that the MPI processes also run in an environment where `$CP2K_DATA_DIR` is (still) defines.

Change made on recommendation of @dev-zero (one of the CP2K core developers).

(see also https://github.com/cp2k/cp2k/issues/234)

### DIFF
--- a/easybuild/easyblocks/c/cp2k.py
+++ b/easybuild/easyblocks/c/cp2k.py
@@ -240,6 +240,9 @@ class EB_CP2K(EasyBlock):
 
         options['LIBS'] = "-Wl,--start-group %s -Wl,--end-group" % options['LIBS']
 
+        # specify correct location for 'data' directory in final installation
+        options['DATA_DIR'] = os.path.join(self.installdir, 'data')
+
         # create arch file using options set
         archfile = os.path.join(self.cfg['start_dir'], 'arch', '%s.%s' % (self.typearch, self.cfg['type']))
         txt = self._generate_makefile(options)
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/c/cp2k.py
```diff
@@ -870,12 +867,3 @@ def sanity_check_step(self):
...
         super(EB_CP2K, self).sanity_check_step(custom_paths=custom_paths)
-
-    def make_module_extra(self):
-        """Set up a CP2K_DATA_DIR environment variable to find CP2K provided basis sets"""
-
-        txt = super(EB_CP2K, self).make_module_extra()
-        datadir = os.path.join(self.installdir, 'data')
-        if os.path.exists(datadir):
-            txt += self.module_generator.set_environment('CP2K_DATA_DIR', datadir)
-        return txt
```
akesandgren: Removing this might be bad for users that's gotten used to being able to copy files using that env var.
Not sure if this is really a problem but it doesn't really hurt to keep it, does it?

wpoely86: :+1:

boegel: I was told CP2K first looks in the current directory, so unsetting this environment variable shouldn't matter, it shouldn't break existing setups.

The search order is: i) current directory, ii) configured location in installation, iii) `$CP2K_DATA_DIR` (not sure about the order of ii) vs iii)).

Note that no longer setting `$CP2K_DATA_DIR` also doesn't prevent users from using `data` files that are parked something else, since in that case they were probably redefining `$CP2K_DATA_DIR`, which still works if `$CP2K_DATA_DIR` is not defined.

I guess the only potential impact could be if users were used to use `$CP2K_DATA_DIR` to find the `data` directory, that will no longer work (they'd have to use `$EBROOTCP2K/data` instead, which wouldn't be mentioned in the CP2K documentation), so that would be a good reason to keep it anyway.

@dev-zero: Any downsides to both configuring the location of the `data` directory correctly via `DATA_DIR` in the `Makefile` and setting `$CP2K_DATA_DIR`?

boegel: Defining `$CP2K_DATA_DIR` environment variable restored in 210f50626

PR #1657, opened 2019-03-15
install sitecustomize.py with Python to support Python package installations for multiple Python versions in a single directory

bartoldeman:
This is put into $EBROOTPYTHON/easybuild/python, which is itself
added to PYTHONPATH, and listens to $EBPYTHONPREFIXES as a list
of prefixes to which lib/pythonx.y/site-packages is added for
every entry, where x.y depends on the running Python version.

$EBROOTPYTHON/easybuild/python needs to be put in PYTHONPATH
to make sure this directory is used even inside virtual environments.


boegel:
Good to go, necessary changes for `PythonPackage` and `PythonBundle` to set `$EBPYTHONPREFIXES` when appropriate be tackled in a follow-up PR...

Thanks @bartoldeman!

PR #1660, opened 2019-03-19
Make sure Perl scripts are installed in the right location

### LINKED ISSUE #1659: Perl-5.28.0-GCCcore-7.3.0.eb  not building due to perldoc sanity checks failing
```
$ eb --version
This is EasyBuild 3.8.1 (framework: 3.8.1, easyblocks: 3.8.1)
```

From [my build log](https://www.dropbox.com/s/0rmwuqpju9kfqiy/install_foss_2018b_toolchain_job.sge.o3588049.gz?dl=0) for Perl-5.28.0-GCCcore-7.3.0.eb the following looks like a major problem:
```
 /usr/local/community/rse/EasyBuild/software/Perl/5.28.0-GCCcore-7.3.0/lib/perl5/5.28.0/xCouldn't copy cpan/podlators/blib/script/pod2man to /usr/local/scripts/pod2man: No such file or directory
```

I.e., it looks like it is trying to install stuff in the wrong place.

@boegel thinks the problem is that the Perl install process finds a `/usr/local/scripts` directory in my environment and incorrectly assumes that's where I'd like it to install scripts.  

[More background info](https://openpkg-dev.openpkg.narkive.com/bGejYSaD/bugdb-perl-possible-build-problem-copying-into-usr-local-scripts-pr-133) (from 16 years ago!)

Suggested fix: add the following to the Perl easyconfig (not tested yet):
```python
configopts = "-Dscriptdirexp=%(installdir)s/bin"
```

NB I've not yet had chance to test this!

ocaisa:
Should fix #1659

### DIFF
--- a/easybuild/easyblocks/p/perl.py
+++ b/easybuild/easyblocks/p/perl.py
@@ -58,6 +58,8 @@ class EB_Perl(ConfigureMake):
             '-Dcc="{0}"'.format(os.getenv('CC')),
             '-Dccflags="{0}"'.format(os.getenv('CFLAGS')),
             '-Dinc_version_list=none',
+            # Guarantee that scripts are installed in the installation directory (and not in a guessed path)
+            '-Dscriptdirexp="{0}"/bin'.format(self.installdir),
         ]
         if self.cfg['use_perl_threads']:
             configopts.append('-Dusethreads')
--- a/easybuild/easyblocks/p/perl.py
+++ b/easybuild/easyblocks/p/perl.py
@@ -45,6 +45,8 @@ class EB_Perl(ConfigureMake):
     def extra_options():
         """Add extra config options specific to Perl."""
         extra_vars = {
+            'force_scripts_installdir': [False, "Force all scripts to be installed in the installation bin directory "
+                                                "via setting -Dscriptdirexp, -Dscriptdir, -Dinstallscript", CUSTOM],
             'use_perl_threads': [True, "Enable use of internal Perl threads via -Dusethreads configure option", CUSTOM],
         }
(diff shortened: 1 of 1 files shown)


ocaisa:
@willfurnass Can you give these changes your seal of approval?

willfurnass:
@ocaisa Yep, although not sure if setting `installdir` and `scriptdirexp` will cause problems for those using [AFS](https://en.wikipedia.org/wiki/Andrew_File_System) - see https://perldoc.perl.org/Config.html

easybuild/easyblocks/p/perl.py
```diff
@@ -45,6 +45,8 @@ class EB_Perl(ConfigureMake):
     def extra_options():
         """Add extra config options specific to Perl."""
         extra_vars = {
+            'force_scripts_installdir': [False, "Force all scripts to be installed in the installation bin directory "
```
ocaisa: `False` is backwards compatible, `True` would automatically solve this problem and probably does no harm...but I can't be sure.

boegel: @ocaisa I see no reason to not always do this... scripts to into `<prefix>/bin` by default anyway, right?

damianam: I'd consider that any of these scripts not ending up in `$EBROOTPERL/bin` is a bug and therefore I wouldn't worry about backwards compatibility. I wouldn't make this change optional either. @boegel ?

boegel: @ocaisa https://github.com/ocaisa/easybuild-easyblocks/pull/16

PR #1661, opened 2019-03-19
enhance SAMtools easyblock to also install libhts.a and include/htslib (when applicable)

boegel:
`libhts.a` is required for `MetaBAT`, see also https://github.com/easybuilders/easybuild-easyconfigs/pull/7931


boegel:
Tested this with *all* existing `SAMtools` easyconfigs (57 of them), works like a charm.

PR #1664, opened 2019-03-24
tweak PythonPackage & PythonBundle to update $EBPYTHONPREFIXES rather than $PYTHONPATH for multi-Python installations + make sanity_check_step aware of multi_deps

boegel:
follow-up for ~~#1657~~, requires ~~https://github.com/easybuilders/easybuild-framework/pull/2813~~ to be merged first

cc @bartoldeman

### DIFF
--- a/easybuild/easyblocks/generic/pythonpackage.py
+++ b/easybuild/easyblocks/generic/pythonpackage.py
@@ -41,7 +41,7 @@ from vsc.utils import fancylogger
 from vsc.utils.missing import nub
 
 import easybuild.tools.environment as env
-from easybuild.easyblocks.python import EXTS_FILTER_PYTHON_PACKAGES
+from easybuild.easyblocks.python import EBPYTHONPREFIXES, EXTS_FILTER_PYTHON_PACKAGES
 from easybuild.framework.easyconfig import CUSTOM
 from easybuild.framework.extensioneasyblock import ExtensionEasyBlock
 from easybuild.tools.build_log import EasyBuildError
@@ -233,6 +233,9 @@ class PythonPackage(ExtensionEasyBlock):
             self.cfg['download_dep_fail'] = self.master.cfg['exts_download_dep_fail']
             self.log.info("Inherited setting for detection of downloaded dependencies from parent: %s",
                           self.cfg['download_dep_fail'])
+
+        # figure out whether this Python package is being installed for multiple Python versions
+        self.multi_python = 'Python' in self.cfg['multi_deps']
 
         # determine install command
         self.use_setup_py = False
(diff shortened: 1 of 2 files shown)


easybuild/easyblocks/generic/pythonpackage.py
```diff
@@ -620,11 +647,17 @@ def make_module_req_guess(self):
...
-        for path in self.all_pylibdirs:
-            fullpath = os.path.join(self.installdir, path)
-            # only extend $PYTHONPATH with existing, non-empty directories
-            if os.path.exists(fullpath) and os.listdir(fullpath):
-                txt += self.module_generator.prepend_paths('PYTHONPATH', path)
+
+        # update $EBPYTHONPREFIXES rather than $PYTHONPATH
+        # if this Python package was installed for multiple Python versions
+        if self.multi_python:
+            txt += self.module_generator.prepend_paths(EBPYTHONPREFIXES, '')
```
damianam: Doesn't it make sense to have the same checks than for the regular installation?

easybuild/easyblocks/generic/pythonpackage.py
```diff
@@ -588,6 +592,29 @@ def sanity_check_step(self, *args, **kwargs):
...
             self.log.debug("Detection of downloaded dependencies not enabled")
 
+        # inject directory path that uses %(pyshortver)s template into default value for sanity_check_paths,
+        # but only for stand-alone installations, not for extensions;
+        # this is relevant for installations of Python packages for multiple Python versions (via multi_deps)
+        # (we can not pass this via custom_paths, since then the %(pyshortver)s template value will not be resolved)
+        if not self.is_extension and not self.cfg['sanity_check_paths']:
+            self.cfg['sanity_check_paths'] = {
+                'files': [],
+                'dirs': [os.path.join('lib', 'python%(pyshortver)s', 'site-packages')],
```
damianam: Doesn't this mean that the sanity check is just performed in a single directory (eg: `python2.7`), instead of all the directories needed when using `multi_deps`?

easybuild/easyblocks/generic/pythonpackage.py
boegel: Not really, because we're just adding the top-level installation directory to `$EBPYTHONPREFIXES`, not specific subdirectories like `lib/python*/site-packages`.

The `site-customize.py` script in framework determines which subdirectories are considered for the paths listed in `$EBPYTHONPREFIXES`

easybuild/easyblocks/generic/pythonpackage.py
boegel: No, because in https://github.com/easybuilders/easybuild-framework/pull/2813 the `sanity_check_step` is done iteratively, once for every entry in `multi_deps`.

When `Python` 2.7.x is listed as a build dep, `%(pyshortver)s` will resolve to `2.7`; when `Python` 3.6.x is used, it will resolve to `3.6`.

easybuild/easyblocks/generic/pythonpackage.py
damianam: Yes, but the top-level directory is useless if the directories inside do not exist. That's my point, if checking for the existence of those directories make sense.

boegel: That's done via the `sanity_check_paths` default we inject?

Also, if those directories don't exist, the `import` check will fail anyway (unless those modules are available somewhere else for the right Python version, I guess)

PR #1666, opened 2019-03-26
only install mkl-dnn by default with TensorFlow on x86-64 systems

edmondac:
This is just a part of the TensorFlow on POWER9 story. See https://github.com/easybuilders/easybuild-framework/issues/2655 and https://github.com/easybuilders/easybuild-framework/issues/2654 for example. Other PRs will follow in time.


easybuild/easyblocks/t/tensorflow.py
```diff
@@ -81,8 +81,13 @@ def extra_options():
             'cuda_compute_capabilities': [[], "List of CUDA compute capabilities to build with", CUSTOM],
             'path_filter': [[], "List of patterns to be filtered out in paths in $CPATH and $LIBRARY_PATH", CUSTOM],
             'with_jemalloc': [None, "Make TensorFlow use jemalloc (usually enabled by default)", CUSTOM],
-            'with_mkl_dnn': [True, "Make TensorFlow use Intel MKL-DNN", CUSTOM],
         }
+
```
damianam: This makes sense to me, but I think we can simultaneously tackle a small issue. Down in the easyblock we enable `mkl-dnn` https://github.com/easybuilders/easybuild-easyblocks/blob/98984e2e6393c09642b4111b9f97a490007f857e/easybuild/easyblocks/t/tensorflow.py#L382 However, that line doesn't depend on CUDA. So it is actually an error, since you can't have simultaneously a tensorflow build with `mkl-dnn` and `cuDNN`. Or can you? @akesandgren ? Does it make sense to make that block an `else:` of `if cuda_root:`?

edmondac: I've gone for "if cudnn is present then don't use mkl-dnn"... does that look ok?

akesandgren:
On x86_64 system we do want mkl-dnn even if cudnn is available. So please do not drop it for all archs.

edmondac:
Sorry - I was confused by the earlier comment. I've now put that whole block in an `if get_cpu_architecture() == X86_64` statement... Is that now right?

easybuild/easyblocks/t/tensorflow.py
```diff
@@ -81,8 +81,13 @@ def extra_options():
...
+        if get_cpu_architecture() == X86_64:
+            extra_vars['with_mkl_dnn'] = [True, "Make TensorFlow use Intel MKL-DNN", CUSTOM]
+        else:
+            extra_vars['with_mkl_dnn'] = [False, "Make TensorFlow use Intel MKL-DNN", CUSTOM]
```
boegel: We should avoid copy-pasting things here, as follows:

```python
# only install mkl-dnn by default on x86_64 systems
with_mkl_dnn_default = get_cpu_architecture() == X86_64
extra_vars['with_mkl_dnn'] = [with_mkl_dnn_default, "Make TensorFlow use Intel MKL-DNN", CUSTOM]

File: easybuild/easyblocks/t/tensorflow.py [hunk 3]
```diff
@@ -374,16 +379,17 @@ def build_step(self):
...
-        mkl_root = get_software_root('mkl-dnn')
-        if mkl_root:
-            cmd.extend(['--config=mkl'])
-            cmd.insert(0, "export TF_MKL_DOWNLOAD=0 &&")
-            cmd.insert(0, "export TF_MKL_ROOT=%s &&" % mkl_root)
-        elif self.cfg['with_mkl_dnn']:
-            # this makes TensorFlow use mkl-dnn (cfr. https://github.com/01org/mkl-dnn)
-            cmd.extend(['--config=mkl'])
-            cmd.insert(0, "export TF_MKL_DOWNLOAD=1 && ")
+        if get_cpu_architecture() == X86_64:
```
boegel: I'm not sure this is actually the right way to go here...

Do we *always* want to skip on `mkl-dnn` entirely on non-x86_64 systems (also in the future)? Note that this is broader than "not on POWER"...

I think this change should be reverted to what it was, and only change the default value for `with_mkl_dnn`.

On POWER systems, you can then avoid the use of `mkl-dnn` by configuring EasyBuild with `--filter-deps=mkl-dnn`?

Unless you're confident it will never make sense to even consider `mkl-dnn` on non-x86_64 systems, then this is ok :)

easybuild/easyblocks/t/tensorflow.py
```diff
@@ -45,7 +45,7 @@
 from easybuild.tools.filetools import is_readable, which, write_file
 from easybuild.tools.modules import get_software_root, get_software_version
 from easybuild.tools.run import run_cmd
-from easybuild.tools.systemtools import get_os_name, get_os_version
+from easybuild.tools.systemtools import get_os_name, get_os_version, get_cpu_architecture, X86_64
```
boegel: @edmondac Please sort alphabetically here, and list constants first, so:

```python
from easybuild.tools.systemtools import X86_64, get_cpu_architecture, get_os_name, get_os_version

File: easybuild/easyblocks/t/tensorflow.py (hunk 3 above)
Comment (edmondac) [reply easybuild/easyblocks/t/tensorflow.py]: That's how I had it to start with :-) I'll put it back.

**Comment by edmondac** [comment]
Is this how you'd like it @boegel ?

File: easybuild/easyblocks/t/tensorflow.py [hunk 5]
```diff
@@ -83,10 +83,9 @@ def extra_options():
             'with_jemalloc': [None, "Make TensorFlow use jemalloc (usually enabled by default)", CUSTOM],
         }
 
-        if get_cpu_architecture() == X86_64:
-            extra_vars['with_mkl_dnn'] = [True, "Make TensorFlow use Intel MKL-DNN", CUSTOM]
-        else:
-            extra_vars['with_mkl_dnn'] = [False, "Make TensorFlow use Intel MKL-DNN", CUSTOM]
+        # only install mkl-dnn by default on x86_64 systems
+        with_mkl_dnn_default = get_cpu_architecture() == X86_64
```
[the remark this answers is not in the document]
edmondac: @boegel what's your opinion? If you guys come up with a consensus then I'm happy to do it.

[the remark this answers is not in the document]
edmondac: I've done what @akesandgren asked, since it makes sense to me! See 5f9f739.

boegel:
@edmondac Thank you very much for your persistence on this, I think we've found the right way to tackle this now, so merging...

PR #1668, opened 2019-04-01
replace -openmp with -fopenmp when building TINKER with Intel compilers (+ modernize easyblock)

boegel:


akesandgren:
Since I'm not MotW for easyblocks this week you probably won't find out :-)

PR #1669, opened 2019-04-01
remove CMakeMake easyblock

boegelbot:
CMake should die in a fire, therefore we should stop reporting it.

Humans with admin rights in this repository: please merge ASAP.


akesandgren:
I know it's April 1st, but this isn't funny...

PR #1671, opened 2019-04-02
always set self.pylibdir in PythonPackage.set_pylibdirs(), which is important when installing Python packages for multiple Python versions via multi_deps

boegel:
Ran into a hard to diagnose error like this when installing `h5py` for both Python 2.7.15 and 3.7.2 via `multi_deps` without this change:

```
ImportError: dynamic module does not define module export function (PyInit__multiarray_umath)
```

I think it's fine to simply always set `self.pylibdir`, regardless of whether it was set already or not. It's basically a case of premature optimization as far as I can tell, and `set_pylibdirs` should only get called once in a normal (non-multi-deps) installation.


akesandgren:
Going in, thanks @boegel!

PR #1673, opened 2019-04-03
Make sure meson has -Dlibdir=lib so it doesn't install libraries in lib/x86_64-linux-gnu on Debian/Ubuntu multiarch systems

akesandgren:
Make sure meson has -Dlibdir=lib so it doesn't install libraries in lib/x86_64-linux-gnu on Debian/Ubuntu multiarch systems.

### DIFF
--- a/easybuild/easyblocks/generic/mesonninja.py
+++ b/easybuild/easyblocks/generic/mesonninja.py
@@ -70,7 +70,9 @@ class MesonNinja(EasyBlock):
 
         # Make sure libdir doesn't get set to lib/x86_64-linux-gnu or something
         # on Debian/Ubuntu multiarch systems and others.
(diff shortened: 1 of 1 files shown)


boegel:
@akesandgren This seems to break the existing `at-spi2-core-2.26.3-fosscuda-2018b.eb`.

The fact that this changes makes the following appear in the configure command may explain it: `--libdir lib  -Dlibdir=lib`.

boegel:
Same for `at-spi2-core-2.26.3-fosscuda-2018b.eb`

boegel:
One more: `libepoxy-1.5.3-fosscuda-2018b.eb`

boegel:
Works like a charm now for all easyconfigs using `MesonNinja`, thanks @akesandgren!

Issue #1670, opened 2019-04-01
Qt5 builds do not respect EASYBUILD_OPTARCH - or any other CFLAGS, CXXFLAGS, LDFLAGS etc


casparvl:
**Issue**

So, the issue is, Qt doesn't listen to the 'normal' CFLAGS, CXXFLAGS, and LDFLAGS.

**Description**

To be able to play around, I've dumped an env script with eb's `--dump-env-script`:

```
source Qt5-5.10.1-intel-2018b.env
```

Then, my CFLAGS are

```
[casparl@tcn900 qt-everywhere-src-5.10.1]$ echo $CFLAGS
-O2 -O2 -xAVX -axCORE-AVX2,CORE-AVX512 -ftz -fp-speculation=safe -fp-model source
```

I run the configure line that EasyBuild generates for the intel-2018b toolchain:

```
./configure -prefix /home/casparl/.local/EB_regression/RedHatEnterpriseServer7/software/Qt5/5.10.1-intel-2018b -skip qtgamepad -release  -platform linux-icc-64
```

Then, when I start building, you can see the optimization flags that are actually passed to the compiler. Again, I use the 'make' line generated by EasyBuild (for clarity, I do a serial build):

```
LD_LIBRARY_PATH=/tmp/casparl/Qt5/5.10.1/intel-2018b/qt-everywhere-src-5.10.1/qtbase/lib:/tmp/casparl/Qt5/5.10.1/intel-2018b/qt-everywhere-src-5.10.1/qtdeclarative/lib:$LD_LIBRARY_PATH  make -j 1
[...]

PR #1677, opened 2019-04-08
pass down compiler names and options to Qt5 5.8 & more recent via corresponding QMAKE_* configuration options

boegel:
fix for problem reported in #1670 by @casparvl

These `QMAKE_*` options are only supported for `Qt5` 5.8.x and newer.


migueldiascosta:
Going in, thanks @boegel!

PR #1679, opened 2019-04-10
Update CGAL easyblock  to support CGAL 4.12 & newer

damianam:


easybuild/easyblocks/c/cgal.py
```diff
@@ -60,8 +63,12 @@ def sanity_check_step(self):
         shlib_ext = get_shared_lib_ext()
         libdirs = ('lib', 'lib64')
         libs = [tuple(os.path.join(d, 'libCGAL%s.%s' % (l, shlib_ext)) for d in libdirs) for l in ['', '_Core']]
+        if LooseVersion(self.version) > LooseVersion('4.12'):
```
boegel: @damianam Should be `>=`, since this also applies to the CGAL 4.12 release.

easybuild/easyblocks/c/cgal.py
```diff
@@ -60,8 +63,12 @@ def sanity_check_step(self):
...
+            dirs = ['include/CGAL', ('lib/cmake/CGAL', 'lib64/cmake/CGAL')]
+        else:
+            dirs = ['include/CGAL', ('lib/CGAL', 'lib64/CGAL')]
```
boegel: @damianam The EasyBuild framework will automatically consider `lib64/*` when `lib/*` is specified, and vice versa, so maybe just specify one of them here?

My vote goes to `lib64`, which is the one I see getting installed...

easybuild/easyblocks/c/cgal.py
```diff
@@ -42,26 +45,31 @@ class EB_CGAL(CMakeMake):
...
+
         shlib_ext = get_shared_lib_ext()
-        libdirs = ('lib', 'lib64')
-        libs = [tuple(os.path.join(d, 'libCGAL%s.%s' % (l, shlib_ext)) for d in libdirs) for l in ['', '_Core']]
+        libs = [os.path.join('lib64', 'libCGAL%s.%s' % (l, shlib_ext)) for l in ['', '_Core']]
+        dirs = [os.path.join('include', 'CGAL')]
+        if LooseVersion(self.version) >= LooseVersion('4.12'):
+            dirs += os.path.join('lib64', 'cmake', 'CGAL')
+        else:
+            dirs += os.path.join('lib64', 'CGAL')
```
boegel: @damianam please use `dirs.append(...)` here to add an item to the list, same above

boegel:
Going in, thanks @damianam! Nice try on trying to inject a bug with a silly easy-to-overlook typo... ;)

Issue #1680, opened 2019-04-11
Ensure ldconfig can be found in CUDA easyblock


willfurnass:
At present the `cuda.py` easyblock (v3.8.1) depends on being able to find and run `ldconfig`:

```python
    def post_install_step(self):
       # ...
       # Run ldconfig to create missing symlinks in the stubs directory (libcuda.so.1, etc)
       run_cmd("ldconfig -N " + os.path.join(self.installdir, 'lib64', 'stubs'))
```

However, on Centos 7x `ldconfig` is in `/sbin/` and `/usr/sbin/` but neither dir may be on the `PATH` e.g. in certain batch job environments.  

Should this case be handled in the easyblock?

akesandgren:
The first question here is if we should just add /sbin:/usr/sbin to PATH before executing the ldconfig, or check if it can be found and bail out if not?
@boegel any comments on that?
I think adding /sbin:/usr/sbin and then checking it might be the best way here...

akesandgren:
Proposed fix in https://github.com/easybuilders/easybuild-easyblocks/pull/1684

boegel:
closing since #1684 just got merged, fix will be included in upcoming EasyBuild v3.9.1

PR #1684, opened 2019-04-12
Make sure ldconfig can be found before using it.

akesandgren:
This fixes https://github.com/easybuilders/easybuild-easyblocks/issues/1680

### DIFF
--- a/easybuild/easyblocks/c/cuda.py
+++ b/easybuild/easyblocks/c/cuda.py
@@ -137,15 +137,24 @@ class EB_CUDA(Binary):
         for comp in (self.cfg['host_compilers'] or []):
             create_wrapper('nvcc_%s' % comp, comp)
 
-        # ldconfig is usually in /sbin or /usr/sbin
-        curr_path = os.environ.get('PATH', '')
-        setvar('PATH', os.pathsep.join(['/sbin', '/usr/sbin', curr_path]))
+        saved_path = os.environ.get('PATH', '')
+        ldconfig = which('ldconfig')
+        if ldconfig:
+            self.log.info("ldconfig found at %s", ldconfig)
+        else:
+            # ldconfig is usually in /sbin or /usr/sbin
+            for cand_path in ['/sbin', '/usr/sbin']:
+                if os.path.exists(os.path.join(cand_path, 'ldconfig')):
+                    self.log.info("ldconfig found at %s, so appending that location to $PATH" % cand_path)
+                    setvar('PATH', os.pathsep.join([saved_path, cand_path]))
+
+        # fail if we couldn't find ldconfig, because it's really needed
         if not which('ldconfig'):
-            raise EasyBuildError("Unable to find the 'ldconfig' in PATH")
+            raise EasyBuildError("Unable to find 'ldconfig' in $PATH = %s" os.environ.get('PATH', ''))
 
         # Run ldconfig to create missing symlinks in the stubs directory (libcuda.so.1, etc)
         run_cmd("ldconfig -N " + os.path.join(self.installdir, 'lib64', 'stubs'))
-        setvar('PATH', curr_path)
+        setvar('PATH', saved_path)
 
         super(EB_CUDA, self).post_install_step()
 
--- a/easybuild/easyblocks/c/cuda.py
+++ b/easybuild/easyblocks/c/cuda.py
@@ -143,8 +143,8 @@ class EB_CUDA(Binary):
             # ldconfig is usually in /sbin or /usr/sbin
             for cand_path in ['/sbin', '/usr/sbin']:
                 if os.path.exists(os.path.join(cand_path, 'ldconfig')):
-                    self.log.info("ldconfig found at %s, so using that." % cand_path)
                     ldconfig = os.path.join(cand_path, 'ldconfig')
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/c/cuda.py
```diff
@@ -136,8 +137,15 @@ def create_wrapper(wrapper_name, wrapper_comp):
         for comp in (self.cfg['host_compilers'] or []):
             create_wrapper('nvcc_%s' % comp, comp)
 
+        # ldconfig is usually in /sbin or /usr/sbin
+        curr_path = os.environ.get('PATH', '')
+        setvar('PATH', os.pathsep.join(['/sbin', '/usr/sbin', curr_path]))
```
boegel: @akesandgren Maybe we should only do this if `ldconfig` was not available yet?

Also, we can actually make very sure whether `ldconfig` is in `/sbin` or `/usr/sbin` before prepending the location to `$PATH`, using `os.path.exists`.

easybuild/easyblocks/c/cuda.py
```diff
@@ -136,8 +137,15 @@ def create_wrapper(wrapper_name, wrapper_comp):
...
+        if not which('ldconfig'):
+            raise EasyBuildError("Unable to find the 'ldconfig' in PATH")
```
boegel: Please mention `$PATH` in the error message, just `PATH` is a bit ambiguous

easybuild/easyblocks/c/cuda.py
akesandgren: Well, that's what which does basically, so adding them and running which is shorter.

easybuild/easyblocks/c/cuda.py
akesandgren: You mean:
```
raise EasyBuildError("Unable to find the 'ldconfig' in $PATH")
 
```

akesandgren: @boegel ping

boegel: Yes, like that is better (although I guess bonus points can be earned if you mention the `$PATH` value as well in the error message.

easybuild/easyblocks/c/cuda.py
boegel: I would prefer not pushing stuff into `$PATH` that we don't need, and I'd also prefer *appending* to `$PATH`, so something like this:

```python
ldconfig = which('ldconfig')
if ldconfig:
    self.log.info("ldconfig found at %s", ldconfig)
else:
    # ldconfig is usually in /sbin or /usr/sbin, so try and find it there
    for cand_path in ['/sbin', '/usr/sbin']:
        if os.path.exists(os.path.join(cand_path, 'ldconfig')):
            self.log.info("ldconfig found at %s, so appending that location to $PATH")
            setvar('PATH', os.pathsep.join(os.environ.get('PATH', ''), cand_path))
            break

    # fail if we couldn't find ldconfig, because it's really needed
    if not which('ldconfig'):
        raise EasyBuildError("Unable to find the 'ldconfig' in $PATH")
```

Maybe we should make a dedicated function for this in framework, it may come in useful in other places too (doesn't need to be a blocker for this PR, we can start using it here when it's in place)?

easybuild/easyblocks/c/cuda.py
```diff
@@ -136,8 +137,24 @@ def create_wrapper(wrapper_name, wrapper_comp):
...
+        saved_path = os.environ.get('PATH', '')
+        ldconfig = which('ldconfig')
+        if ldconfig:
+            self.log.info("ldconfig found at %s", ldconfig)
+        else:
+            # ldconfig is usually in /sbin or /usr/sbin
+            for cand_path in ['/sbin', '/usr/sbin']:
+                if os.path.exists(os.path.join(cand_path, 'ldconfig')):
+                    self.log.info("ldconfig found at %s, so appending that location to $PATH" % cand_path)
+                    setvar('PATH', os.pathsep.join([saved_path, cand_path]))
```
boegel: @akesandgren I think it makes sense to avoid updating `$PATH` all together... We can just assign `ldconfig` to the full path to `ldconfig`, and then use the `ldconfig` variable in `run_cmd`, so we (always) call `ldconfig` full path?

That's safer than pushing `/sbin` into `$PATH` (and less code since we don't need to set/restore `$PATH`)

easybuild/easyblocks/c/cuda.py
```diff
@@ -136,8 +136,25 @@ def create_wrapper(wrapper_name, wrapper_comp):
...
+                    break
+
+        # fail if we couldn't find ldconfig, because it's really needed
+        if not ldconfig:
+            raise EasyBuildError("Unable to find 'ldconfig' in $PATH = %s " +
+                                 "nor in /sbin or /usr/sbin" % os.environ.get('PATH', ''))
+
         # Run ldconfig to create missing symlinks in the stubs directory (libcuda.so.1, etc)
-        run_cmd("ldconfig -N " + os.path.join(self.installdir, 'lib64', 'stubs'))
+        cmd = " ".join(ldconfig, '-N', os.path.join(self.installdir, 'lib64', 'stubs'))
```
boegel: @akesandgren This `join` is incorrect, fixed in https://github.com/akesandgren/easybuild-easyblocks/pull/9

Issue #1511, opened 2018-09-13
PythonPackage pip install should not write in $HOME/.cache


akesandgren:
We should try to avoid writing in $HOME as much as possible. One step would be to stop pip install from doing it.

boegel:
@akesandgren Looks like this is quite easy to do: just pass `--no-cache-dir` to `pip`. Or set `$XDG_CACHE_HOME`, see https://pip.pypa.io/en/stable/reference/pip_install/#caching.

The latter is probably better/safer, unless we check when `--no-cache-dir` was added and we check the `pip` version.

akesandgren:
According to that url --no-cache-dir was added with v6.0 when the cache was first invented.

akesandgren:
Not sure where in pythonpackage it is best to set XDG_CACHE_HOME.
It need to be set even before test_step since the result of compose_install_command is used there too.

boegel:
fixed with #1687

surak:
Is there an option, like sourcepath, but for pip? We are running into the problem where the compute nodes don't have network access, and the ones which do, have no gpu.

PR #1687, opened 2019-04-15
use temporary directory for pip cache (rather than $HOME/.cache/pip)

boegel:
fix for #1511


akesandgren:
Going in, thanks @boegel!

PR #1688, opened 2019-04-15
allow oversubscription when testing FFTW on top of OpenMPI >= 3.0

boegel:


akesandgren:
Going in, thanks @boegel!

PR #1689, opened 2019-04-22
drop requirement for setuptools as runtime dependency 

boegel:


akesandgren:
Going in, thanks @boegel!

PR #1690, opened 2019-04-23
fix import statement for DUMMY_TOOLCHAIN_NAME

boegel:
This is required after the merge of https://github.com/easybuilders/easybuild-framework/pull/2836 where some constants had to be moved a bit...


akesandgren:
Going in, thanks @boegel!

PR #1691, opened 2019-04-23
make sure ~/.cache/pip isn't used while installing Tkinter

boegel:
I deliberately broke the use of `$HOME/.cache/pip` in my test account to catch occurrences where `pip` tries to use it (after we told it not to in #1687.

`Tkinter` installed failed because of that, with a `Permission denied` on `$HOME/.cache/pip/wheels`:

```
Could not install packages due to an EnvironmentError: [Errno 13] Permission denied: '/home/example/.cache/pip/wheels/f7/0b/c4/81c147d50fe89c71703e455053466f33e558af45fbab1812f4'
```

which can be circumvented by pointing `$XDG_CACHE_HOME` to a temporary location

### DIFF
--- a/easybuild/easyblocks/t/tkinter.py
+++ b/easybuild/easyblocks/t/tkinter.py
@@ -28,17 +28,18 @@ module to use Tcl/Tk.
 
 @author: Adam Huffman (The Francis Crick Institute)
 @author: Ward Poelmans (Free University of Brussels)
+@author: Kenneth Hoste (HPC-UGent)
 """
-import os
-import shutil
 import glob
+import os
 import tempfile
 from distutils.version import LooseVersion
 
+import easybuild.tools.environment as env
 from easybuild.easyblocks.generic.pythonpackage import det_pylibdir
 from easybuild.easyblocks.python import EB_Python
 from easybuild.tools.build_log import EasyBuildError
-from easybuild.tools.filetools import copy, mkdir, rmtree2
+from easybuild.tools.filetools import copy, move_file, remove_dir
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/t/tkinter.py
```diff
@@ -77,14 +83,11 @@ def install_step(self):
...
-        rmtree2(self.installdir)
+        remove_dir(self.installdir)
 
-        mkdir(pylibdir, parents=True)
-        try:
-            shutil.move(os.path.join(tmpdir, tkparts[0]), pylibdir)
-            shutil.move(os.path.join(tmpdir, os.path.basename(tkparts[1])), pylibdir)
-        except (IOError, OSError) as err:
-            raise EasyBuildError("Failed to move Tkinter back to the install directory: %s", err)
+        move_file(os.path.join(tmpdir, tkparts[0]), os.path.join(pylibdir, tkparts[0]))
```
akesandgren: Can't move_file take a dir as the second arg? Because if it can you should only need pylibdir there and not the path.join with tkparts[0].

And the same for the one below.

boegel: Right now, it can't.

The problem is that you can't distinguish between a file or directory if the target doesn't exist yet in case you're moving a file: should the target directory be created, and the file moved into it, or should the parent of the target be created as a directory, and then the file moved to the specified target?

akesandgren: Aha, forgot the non-exist dir situation...

akesandgren:
Going in, thanks @boegel!

PR #1692, opened 2019-04-24
update QuantumESPRESSO easyblock for v6.4.1

Micket:
1. Generalize binary copying code
2. Add full include paths for ELPA (2017+ we have skipped the symlinks, might as well just always use the right include path)
3. Remove sanity check for binaries dropped from > v6.4

### DIFF
--- a/easybuild/easyblocks/q/quantumespresso.py
+++ b/easybuild/easyblocks/q/quantumespresso.py
@@ -142,16 +142,14 @@ class EB_QuantumESPRESSO(ConfigureMake):
                                      "version %s or newer" % (self.version, elpa_min_ver))
 
             if self.toolchain.options.get('openmp', False):
-                elpa_include = os.path.join(elpa, 'include', 'elpa_openmp-%s' % elpa_v)
-            else:
-                elpa_include = os.path.join(elpa, 'include', 'elpa-%s' % elpa_v)
-            repls.append(('IFLAGS', '-I%s' % os.path.join(elpa_include, 'modules'), True))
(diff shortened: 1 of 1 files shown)


migueldiascosta:
@Micket did you try this with the 6.4 "release pack", besides 6.4.1? I had to add a special case for `install_subdir`,

```
@@ -62,7 +62,10 @@
         super(EB_QuantumESPRESSO, self).__init__(*args, **kwargs)
 
         if LooseVersion(self.version) >= LooseVersion("6"):
-            self.install_subdir = "qe-%s" % self.version
+            if self.version == '6.4':
+                self.install_subdir = "qe_release_%s" % self.version
+            else:
+                self.install_subdir = "qe-%s" % self.version
         else:
             self.install_subdir = "espresso-%s" % self.version
 

```

Micket:
@migueldiascosta I didn't test 6.4
What do you say about having an option to specify the installdir as part of the config file instead so that we don't have to hack in more version specifics into the block?

boegel:
@Micket As discussed during conf call, having a fallback option to manually specify the install subdir makes sense to me, but we should keep giving preference to handling it in the easyblock...

Micket:
As generalizing the install subdir path isn't necessary for 6.4.1, and can be done independently of these fixes, I'm suggesting someone reviews this PR as it is. I will make a new one for the subdir option, to make them easier to review.

boegel:
I think this is good to go, I've tested it with a couple of existing `QuantumESPRESSO` easyconfigs (old & new), and test report for are on their way. If no issues pop up, I'll merge it tomorrow...

migueldiascosta:
Going in, thanks @Micket!

PR #1693, opened 2019-04-25
OpenBLAS with POWER support

edmondac:
For POWER8/9 we need to say 'TARGET=POWER8'

### DIFF
--- a/easybuild/easyblocks/o/openblas.py
+++ b/easybuild/easyblocks/o/openblas.py
@@ -1,0 +1,20 @@
+"""
+EasyBuild support for building and installing OpenBLAS, implemented as an easyblock
+
+@author: Andrew Edmondson (University of Birmingham)
+"""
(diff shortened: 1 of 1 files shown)


edmondac:
Thanks!

boegel:
It would be nice to also have a custom `sanity_check_step` implementation in place in this custom OpenBLAS easyblock, so we can consider switching over the current `OpenBLAS` easyconfig files to use this easyblock (rather than the generic `ConfigureMake` easyblock).

Also auto-skipping `configure_step` would be nice (with a simple `pass`), so the `skipsteps = ['configure']` from the easyconfig files can be omitted, and automatically adding certain build options like `CC="$CC" FC="$F77"'` and install options like `PREFIX=%(installdir)s` (and maybe also the `BINARY`, `USE_OPENMP` and `USE_THREAD` ones, at least by default).

@edmondac Are you up for looking into that?

edmondac:
Yep - I'll do that @boegel

PR #1694, opened 2019-04-25
vtune: no license needed since 2019u3

jose-d:
as written in [release notes](https://software.intel.com/en-us/articles/intel-advisor-release-notes) no license is required since this $update.

### DIFF
--- a/easybuild/easyblocks/v/vtune.py
+++ b/easybuild/easyblocks/v/vtune.py
@@ -50,6 +50,14 @@ class EB_VTune(IntelBase):
         elif loosever >= LooseVersion('2018'):
             self.subdir = 'vtune_amplifier'
 
+    def prepare_step(self, *args, **kwargs):
+        """Since 2019u3 there is no license check."""
+        if LooseVersion(self.version) >= LooseVersion('2019_update3'):
+            kwargs['requires_runtime_license'] = False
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/v/vtune.py
```diff
@@ -50,6 +50,14 @@ def __init__(self, *args, **kwargs):
         elif loosever >= LooseVersion('2018'):
             self.subdir = 'vtune_amplifier'
 
+    def prepare_step(self, *args, **kwargs):
+        """Since 2019u3 there is no license check."""
+        if LooseVersion(self.version) >= LooseVersion('2019_update3'):
```
bartoldeman: Just a minor thing, you could just use an if here without the else, as in:

```
if LooseVersion(self.version) >= LooseVersion('2019_update3'):
    kwargs['requires_runtime_license'] = False
super(EB_VTune, self).prepare_step(*args, **kwargs)
```

jose-d: oki, I used same snippet in advisor so I'll change it in both PRs.

jose-d: fixed and build results with updated easyblock uploaded into [related easyconfig](https://github.com/easybuilders/easybuild-easyconfigs/pull/8139#issuecomment-486669786)

PR #1695, opened 2019-04-25
advisor: since 2019u3 license not required

jose-d:
[release notes](https://software.intel.com/en-us/articles/intel-advisor-release-notes#2019u3):

> Simplified installation and licensing (serial numbers and license files are no longer required for this product).

### DIFF
--- a/easybuild/easyblocks/a/advisor.py
+++ b/easybuild/easyblocks/a/advisor.py
@@ -48,12 +48,10 @@ class EB_Advisor(IntelBase):
             self.subdir = 'advisor'
 
     def prepare_step(self, *args, **kwargs):
-        """Since 2019u3 there is no license check."""
+        """Since 2019u3 there is no license required."""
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/a/advisor.py
```diff
@@ -46,6 +47,14 @@ def __init__(self, *args, **kwargs):
         else:
             self.subdir = 'advisor'
 
+    def prepare_step(self, *args, **kwargs):
+        """Since 2019u3 there is no license check."""
+        if LooseVersion(self.version) >= LooseVersion('2019_update3'):
```
bartoldeman: Same comment as for vtune, no need for the `else:`

jose-d: fixed, updated easyblock re-tested in [related easyconfig PR](https://github.com/easybuilders/easybuild-easyconfigs/pull/8140#issuecomment-486672811).

PR #1697, opened 2019-04-25
make OpenBLAS easyblock set default build options + use custom sanity check

edmondac:
Follow on PR from ~~#1693~~

### DIFF
--- a/easybuild/easyblocks/o/openblas.py
+++ b/easybuild/easyblocks/o/openblas.py
@@ -1,0 +1,20 @@
+"""
+EasyBuild support for building and installing OpenBLAS, implemented as an easyblock
+
+@author: Andrew Edmondson (University of Birmingham)
+"""
+from easybuild.easyblocks.generic.configuremake import ConfigureMake
+from easybuild.tools.systemtools import POWER, get_cpu_architecture
+
+
+class EB_OpenBLAS(ConfigureMake):
+    """Support for building/installing OpenBLAS."""
+
+    def build_step(self):
+        """Custom build procedure for OpenBLAS."""
+
+        if get_cpu_architecture() == POWER:
+            # There doesn't seem to be a POWER9 option yet, but POWER8 should work.
+            self.cfg['buildopts'] += ' TARGET=POWER8'
+
+        super(EB_OpenBLAS, self).build_step()
--- a/easybuild/easyblocks/o/openblas.py
+++ b/easybuild/easyblocks/o/openblas.py
@@ -11,26 +11,25 @@ from easybuild.tools.systemtools import POWER, get_cpu_architecture, get_shared_
 class EB_OpenBLAS(ConfigureMake):
     """Support for building/installing OpenBLAS."""
 
-    def build_step(self):
-        """Custom build procedure for OpenBLAS."""
+    def configure_step(self):
+        """ set up some options - but no configure command to run"""
 
-        self.cfg['buildopts'] += 'BINARY=64 USE_THREAD=1 USE_OPENMP=1 CC="%s" FC="%s"' % (
-            os.environ['CC'], os.environ['FC'])
+        if 'BINARY=' not in self.cfg['buildopts']:
+            self.cfg.update('buildopts', 'BINARY=64')
+        if 'USE_THREAD=' not in self.cfg['buildopts']:
+            self.cfg.update('buildopts', 'USE_THREAD=1')
+        if 'USE_OPENMP=' not in self.cfg['buildopts']:
+            self.cfg.update('buildopts', 'USE_OPENMP=1')
+        if 'CC=' not in self.cfg['buildopts']:
+            self.cfg.update('buildopts', 'CC=%s' % os.environ['CC'])
+        if 'FC=' not in self.cfg['buildopts']:
+            self.cfg.update('buildopts', 'FC=%s' % os.environ['FC'])
 
         if get_cpu_architecture() == POWER:
             # There doesn't seem to be a POWER9 option yet, but POWER8 should work.
-            self.cfg['buildopts'] += ' TARGET=POWER8'
+            self.cfg.update('buildopts', ' TARGET=POWER8')
 
-        super(EB_OpenBLAS, self).build_step()
-
-    def install_step(self):
-        """Custom install procedure for OpenBLAS."""
-        self.cfg['installopts'] += 'USE_THREAD=1 USE_OPENMP=1 PREFIX=%s' % self.installdir
-        super(EB_OpenBLAS, self).install_step()
-
-    def configure_step(self):
-        """ nothing to do for OpenBLAS """
-        pass
+        self.cfg.update('installopts', 'PREFIX=%s' % self.installdir)
 
     def sanity_check_step(self):
         """ Custom sanity check for OpenBLAS """
--- a/easybuild/easyblocks/o/openblas.py
+++ b/easybuild/easyblocks/o/openblas.py
@@ -4,7 +4,7 @@ EasyBuild support for building and installing OpenBLAS, implemented as an easybl
 @author: Andrew Edmondson (University of Birmingham)
 """
 from easybuild.easyblocks.generic.configuremake import ConfigureMake
-from easybuild.tools.systemtools import POWER, get_cpu_architecture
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/o/openblas.py
```diff
@@ -18,3 +18,18 @@ def build_step(self):
...
+        pass
+
+    def sanity_check_step(self):
+        """ Custom sanity check for OpenBLAS """
+        custom_paths = {
+            'files': ['include/cblas.h', 'include/f77blas.h', 'include/lapacke_config.h', 'include/lapacke.h',
+                      'include/lapacke_mangling.h', 'include/lapacke_utils.h', 'include/openblas_config.h',
+                      'lib/libopenblas.a', 'lib/libopenblas.%s' % get_shared_lib_ext()],
+
+            'dirs': [''],
```
boegel: @edmondac Please change this to `[]` (check for a single directory with an empty name is silly)

edmondac:
I think I've done the requested changes to this, but GitHub seems to disagree. What's your opinion @boegel?

easybuild/easyblocks/o/openblas.py
```diff
@@ -13,8 +14,31 @@ class EB_OpenBLAS(ConfigureMake):
     def build_step(self):
         """Custom build procedure for OpenBLAS."""
 
+        self.cfg['buildopts'] += 'BINARY=64 USE_THREAD=1 USE_OPENMP=1 CC="%s" FC="%s"' % (
+            os.environ['CC'], os.environ['F77'])
```
akesandgren: This should rather be FC and not F77. Use of F77 should be avoided when possible.
Especially since it's used to set "FC"

edmondac: Yes, sorry! Copy-paste mistake I think.

akesandgren: Well, all the existing easyconfigs for OpenBLAS also uses F77 so not your fault :-)

edmondac:
Looks like travis blew up... but not to do with my changes.

akesandgren:
We should make this more controillable from the easyconfig (like turning on/off threads/openmp etc)
And we should also make it produce multiple variants of the library, serial and SMP, so we can use the serial build for $LIBBLAS and the SMP one for $LIBBLAS_MT.
But this can be done later.

I'm ok with this one. @boegel ?

easybuild/easyblocks/o/openblas.py
```diff
@@ -13,8 +14,31 @@ class EB_OpenBLAS(ConfigureMake):
     def build_step(self):
         """Custom build procedure for OpenBLAS."""
 
+        self.cfg['buildopts'] += 'BINARY=64 USE_THREAD=1 USE_OPENMP=1 CC="%s" FC="%s"' % (
+            os.environ['CC'], os.environ['FC'])
```
boegel: @edmondac Ideally, this is done a bit more carefully. If the easyconfig already specifies `BINARY=...`, then we shouldn't hard inject `BINARY=64` here. Likewise for the other stuff.

Also, you must use `self.cfg.update('buildopts', "...")` here, `+=` doesn't do what you think it does (the stuff you're injecting will not be picked up because of underlying implementation details related to templating).

easybuild/easyblocks/o/openblas.py
```diff
@@ -13,8 +14,31 @@ class EB_OpenBLAS(ConfigureMake):
...
+
         if get_cpu_architecture() == POWER:
             # There doesn't seem to be a POWER9 option yet, but POWER8 should work.
             self.cfg['buildopts'] += ' TARGET=POWER8'
 
         super(EB_OpenBLAS, self).build_step()
+
+    def install_step(self):
+        """Custom install procedure for OpenBLAS."""
+        self.cfg['installopts'] += 'USE_THREAD=1 USE_OPENMP=1 PREFIX=%s' % self.installdir
```
boegel: Also here, you'll need to use `self.cfg.update`.

And I think it makes sense to pick up `buildopts` here, and just add `PREFIX=...` to it in `installopts`, to ensure consistency?

easybuild/easyblocks/o/openblas.py
```diff
@@ -13,8 +14,31 @@ class EB_OpenBLAS(ConfigureMake):
...
+        super(EB_OpenBLAS, self).install_step()
+
+    def configure_step(self):
```
boegel: Please move this up, above `build_step` (try to keep steps in logical order)

easybuild/easyblocks/o/openblas.py
```diff
@@ -13,8 +14,31 @@ class EB_OpenBLAS(ConfigureMake):
...
+        """ nothing to do for OpenBLAS """
+        pass
```
boegel: This would actually be a good place to define/tweak `buildopts` and `installopts`, which is basically configuring the installation?

easybuild/easyblocks/o/openblas.py
```diff
@@ -13,8 +14,31 @@ class EB_OpenBLAS(ConfigureMake):
...
+
+    def sanity_check_step(self):
+        """ Custom sanity check for OpenBLAS """
+        custom_paths = {
+            'files': ['include/cblas.h', 'include/f77blas.h', 'include/lapacke_config.h', 'include/lapacke.h',
+                      'include/lapacke_mangling.h', 'include/lapacke_utils.h', 'include/openblas_config.h',
+                      'lib/libopenblas.a', 'lib/libopenblas.%s' % get_shared_lib_ext()],
+
```
boegel: style nitpicking: drop this empty line please

edmondac:
@boegel do these seem ok to you?

easybuild/easyblocks/o/openblas.py
```diff
@@ -3,18 +3,42 @@
...
+        if 'USE_THREAD=' not in self.cfg['buildopts']:
+            self.cfg.update('buildopts', 'USE_THREAD=1')
+            self.cfg.update('installopts', 'USE_THREAD=1')
+        if 'USE_OPENMP=' not in self.cfg['buildopts']:
+            self.cfg.update('buildopts', 'USE_OPENMP=1')
+            self.cfg.update('installopts', 'USE_OPENMP=1')
+        if 'CC=' not in self.cfg['buildopts']:
+            self.cfg.update('buildopts', 'CC=%s' % os.environ['CC'])
+        if 'FC=' not in self.cfg['buildopts']:
+            self.cfg.update('buildopts', 'FC=%s' % os.environ['FC'])
```
boegel: This is very repetitive, and can be done a lot cleaner using a `dict`:

```python
default_opts = {
    'BINARY': '64',
    'CC': os.getenv('CC'),
    'FC': os.getenv('FC'),
    'USE_OPENMP': '1',
    'USE_THREAD': '1',
}
for key in sorted(default_opts.keys()):
    for opts_key in ['buildopts', 'installopts']:
        if '%s=' % key not in self.cfg[opts_key]:
            self.cfg.update(opts_key, '%s=%s' % (key, default_opts[key]))
```

This will also add `CC=...` to `installopts`, but that shouldn't cause any problems.

easybuild/easyblocks/o/openblas.py
```diff
@@ -3,18 +3,42 @@
...
 
         if get_cpu_architecture() == POWER:
             # There doesn't seem to be a POWER9 option yet, but POWER8 should work.
-            self.cfg['buildopts'] += ' TARGET=POWER8'
+            self.cfg.update('buildopts', 'TARGET=POWER8')
```
boegel: Not sure if this is relevant for `installopts`, but maybe it's better to join this with the above, as follows (before the `for` loop):

```python
if get_cpu_architecture() == POWER:
    default_opts['TARGET'] = 'POWER8'

edmondac: Done

easybuild/easyblocks/o/openblas.py
edmondac: Done

boegel:
No issues uncovered during testing, default configuration is still the same for the easyconfigs touched in https://github.com/easybuilders/easybuild-easyconfigs/pull/8345 compared to before, so good to go, thanks @edmondac!

PR #1698, opened 2019-04-26
add support to PythonPackage to fix 'python' shebang lines in specified files to '#!/usr/bin/env python'

boegel:
edit: now requires ~~https://github.com/easybuilders/easybuild-framework/pull/2844~~

### DIFF
--- a/easybuild/easyblocks/generic/pythonpackage.py
+++ b/easybuild/easyblocks/generic/pythonpackage.py
@@ -46,7 +46,7 @@ from easybuild.easyblocks.python import EBPYTHONPREFIXES, EXTS_FILTER_PYTHON_PAC
 from easybuild.framework.easyconfig import CUSTOM
 from easybuild.framework.extensioneasyblock import ExtensionEasyBlock
 from easybuild.tools.build_log import EasyBuildError
-from easybuild.tools.filetools import apply_regex_substitutions, mkdir, remove_dir, remove_file, which
+from easybuild.tools.filetools import apply_regex_substitutions, mkdir, remove_dir, which
 from easybuild.tools.modules import get_software_root
 from easybuild.tools.run import run_cmd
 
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/generic/pythonpackage.py
```diff
@@ -558,6 +560,15 @@ def install_step(self):
         if pythonpath is not None:
             env.setvar('PYTHONPATH', pythonpath, verbose=False)
 
+        for glob_pattern in self.cfg.get('fix_python_shebang_for', []):
+            paths = glob.glob(os.path.join(self.installdir, glob_pattern))
+            self.log.info("Fixing shebang to '%s' for files that match '%s': %s", PYTHON_SHEBANG, glob_pattern, paths)
+            for path in paths:
+                apply_regex_substitutions(path, [(r'^#!.*python$', PYTHON_SHEBANG)])
```
akesandgren: Need to be python[0-9.]*$ to cover all cases.

easybuild/easyblocks/generic/pythonpackage.py
```diff
@@ -558,6 +560,15 @@ def install_step(self):
...
+
+                # remove backup that is made by apply_regex_substitutions
+                remove_file(path + '.orig.eb')
```
akesandgren: Doesn't apply_regex have a flag to not create backup file?

easybuild/easyblocks/generic/pythonpackage.py
akesandgren: For python 3.7.2 it puts
#!/hpc2n/eb/software/Compiler/GCCcore/8.2.0/Python/3.7.2/bin/python3.7
in the file it created.

easybuild/easyblocks/generic/pythonpackage.py
boegel: No, it doesn't. We could add one though...

easybuild/easyblocks/generic/pythonpackage.py
boegel: Good point, thanks.

boegel: Fixed

easybuild/easyblocks/generic/pythonpackage.py
boegel: @akesandgren support for skipping backup added in https://github.com/easybuilders/easybuild-framework/pull/2844, and started using it in cc1464a31

akesandgren:
Going in, thanks @boegel!

PR #1702, opened 2019-04-27
Handle new multi-deps python setup with $EBPYTHONPREFIXES for tensorflow

akesandgren:


easybuild/easyblocks/t/tensorflow.py
```diff
@@ -362,9 +362,11 @@ def build_step(self):
         # this is required to make sure that Python packages included as extensions are found at build time;
         # see also https://github.com/tensorflow/tensorflow/issues/22395
         pythonpath = os.getenv('PYTHONPATH', '')
-        env.setvar('PYTHONPATH', '%s:%s' % (os.path.join(self.installdir, self.pylibdir), pythonpath))
+        env.setvar('PYTHONPATH', os.pathsep.join([os.path.join(self.installdir, self.pylibdir), pythonpath]))
 
         cmd.append('--action_env=PYTHONPATH')
+        # Also export EBPYTHONPREFIXES to handle the new multi-deps python setup
```
boegel: Drop the `new` (won't be "new" for long), and maybe include https://github.com/easybuilders/easybuild-easyblocks/pull/1664 as a reference?

Maybe we should also make this conditional, i.e. only pass this if `$EBPYTHONPREFIXES` is actually defined?
Otherwise it'll look weird in situations where it's not used...

boegel:
Thanks @akesandgren!

PR #1703, opened 2019-04-29
enhance CP2K easyblock to support building with CUDA support

Micket:
Expose nv arch option.
Add PW_CUDA flag when building CUDA support.

Add flags to nvcc command (else, compile errors).

### DIFF
--- a/easybuild/easyblocks/c/cp2k.py
+++ b/easybuild/easyblocks/c/cp2k.py
@@ -103,6 +103,7 @@ class EB_CP2K(EasyBlock):
             'plumed': [None, "Enable PLUMED support", CUSTOM],
             'type': ['popt', "Type of build ('popt' or 'psmp')", CUSTOM],
             'typeopt': [True, "Enable optimization", CUSTOM],
+            'nvarch': ['sm_72', "NVidia architecture", CUSTOM],
         }
         return EasyBlock.extra_options(extra_vars)
 
@@ -231,9 +232,11 @@ class EB_CP2K(EasyBlock):
         # CUDA
         cuda = get_software_root('CUDA')
         if cuda:
-            options['DFLAGS'] += ' -D__ACC -D__DBCSR_ACC'
+            options['DFLAGS'] += ' -D__ACC -D__DBCSR_ACC -D__PW_CUDA'
             options['LIBS'] += ' -lcudart -lcublas -lcufft -lrt'
             options['NVCC'] = ' nvcc'
+            self.log.info("Using Nvidia arch: %s" % self.cfg['nvarch'])
+            options['NVFLAGS'] = options['DFLAGS'] + ' -O3 -arch ' + self.cfg['nvarch']
 
         # avoid group nesting
         options['LIBS'] = options['LIBS'].replace('-Wl,--start-group', '').replace('-Wl,--end-group', '')


easybuild/easyblocks/c/cp2k.py
```diff
@@ -103,6 +103,7 @@ def extra_options():
             'plumed': [None, "Enable PLUMED support", CUSTOM],
             'type': ['popt', "Type of build ('popt' or 'psmp')", CUSTOM],
             'typeopt': [True, "Enable optimization", CUSTOM],
+            'nvarch': ['sm_72', "NVidia architecture", CUSTOM],
```
boegel: @Micket The default value should be "neutral", so I'd go with `None` (and check below whether any other value is specified before trying to use it)?

easybuild/easyblocks/c/cp2k.py
```diff
@@ -231,9 +232,11 @@ def configure_step(self):
         # CUDA
         cuda = get_software_root('CUDA')
         if cuda:
-            options['DFLAGS'] += ' -D__ACC -D__DBCSR_ACC'
+            options['DFLAGS'] += ' -D__ACC -D__DBCSR_ACC -D__PW_CUDA'
             options['LIBS'] += ' -lcudart -lcublas -lcufft -lrt'
             options['NVCC'] = ' nvcc'
+            self.log.info("Using Nvidia arch: %s" % self.cfg['nvarch'])
+            options['NVFLAGS'] = options['DFLAGS'] + ' -O3 -arch ' + self.cfg['nvarch']
```
boegel: @Micket This should be under an `if self.cfg['nvarch']`, and we should avoid hardcoding the `-O3`?

boegel:
@wpoely86 Thoughts on this?

easybuild/easyblocks/c/cp2k.py
Micket: Well, OK, but I would consider it basically an error to not set it (and CUDA is used). If not set, nvcc defaults to sm_30. Everyone building do want to set this, so that they aren't unwittingly getting a poor build.

easybuild/easyblocks/c/cp2k.py
Micket: I put in a -O3 because this was what other CP2K example build scripts did. Appearantly, this is just the build flags for the CPU code (the gpu code has the equivalent O3 level by default) **correct me if I'm wrong**. But, since this is the CPU code, perhaps we should put a value here based on the toolchainopts?

akesandgren:
I have a user comment about c2pk built by EB being very slow compared to his own build.
I'm going to investigate that as soon as I can so we're likely going to change flags on this one... But controlled from the easyconfig.

easybuild/easyblocks/c/cp2k.py
boegel: OK, then add a check to make sure it's set, which is better than using a default that'll get outdated in no time (and which we can't update easily without breaking backward compatibiltiy).

easybuild/easyblocks/c/cp2k.py
boegel: Indeed, this should ideally be based on what the toolchain sets.

Not sure how trivial it is to only get the `-O2` (or whatever) from `self.toolchain` though...

Micket: @boegel I'd rather expose extra_nvflags and let the config decide?

wpoely86: I've done it slightly different:

```diff
@@ -103,6 +103,7 @@ class EB_CP2K(EasyBlock):
             'plumed': [None, "Enable PLUMED support", CUSTOM],
             'type': ['popt', "Type of build ('popt' or 'psmp')", CUSTOM],
             'typeopt': [True, "Enable optimization", CUSTOM],
+            'cuda_cc': [[], "List of CUDA compute capability to compile for", CUSTOM],
         }
         return EasyBlock.extra_options(extra_vars)
 
@@ -231,9 +232,14 @@ class EB_CP2K(EasyBlock):
         # CUDA
         cuda = get_software_root('CUDA')
         if cuda:
-            options['DFLAGS'] += ' -D__ACC -D__DBCSR_ACC'
+            options['DFLAGS'] += ' -D__ACC -D__DBCSR_ACC -D__PW_CUDA'
             options['LIBS'] += ' -lcudart -lcublas -lcufft -lrt'
             options['NVCC'] = ' nvcc'
+            self.log.info("Using CUDA Compute capabilities: %s" % self.cfg['cuda_cc'])
+            capabilities = [''.join(x.split('.')) for x in self.cfg['cuda_cc']]
+            options['NVFLAGS'] = options['DFLAGS'] + ' '
+            options['NVFLAGS'] += ' '.join(['-gencode arch=compute_%s,code=\"sm_%s,compute_%s\"' % (x, x, x)
+                                            for x in capabilities])
 
```

To we can set in the easyconfig:
```python
cuda_cc = ['3.5', '6.0']
```

Micket: @wpoely86 I was about to add a similar change, but I was holding off to deal with the rest of the compile flags for nvcc. I'm not sure if they matter greatly, but CP2K's example build files suggested the use of -O3 at least.

wpoely86: These makefiles are just examples. If you want to use `-O3`, that's fine but make it an option.

akesandgren:
@Micket Any progress on this? I haven't had time to look at performance yet...

Micket:
@akesandgren No, not at all. In fact, we have users who complain that the CP2K we compile (at C3SE) is actually broken, and my coworker had been trying to investigate this (but I don't think we have any conclusions there either)

akesandgren:
@Micket should we use cuda_compute_capabilities instead?

PR #1704, opened 2019-04-29
Hopefully the final fix to make Bazel NOT use $HOME/.cache/bazel in TensorFlow

akesandgren:


easybuild/easyblocks/t/tensorflow.py
```diff
@@ -257,6 +257,8 @@ def configure_step(self):
         apply_regex_substitutions('configure.py', regex_subs)
 
         cmd = self.cfg['preconfigopts'] + './configure ' + self.cfg['configopts']
+        # Tell Bazel to not use $HOME/.cache/bazel at all
+        env.setvar('TEST_TMPDIR', os.path.join(tmpdir, 'output_root'))
```
boegel: @akesandgren Can you move this above the `cmd =`?

Is there a reference to the Bazel documentation that mentions the use of `$TEST_TMPDIR`?

akesandgren: Yes, there is, site/docs/output_directories.md

boegel: No URL to include? :-/

easybuild/easyblocks/t/tensorflow.py
```diff
@@ -256,6 +256,9 @@ def configure_step(self):
                        r"\1, '--output_base=%s', '--install_base=%s'" % (tmpdir, os.path.join(tmpdir, 'inst_base')))]
         apply_regex_substitutions('configure.py', regex_subs)
 
+        # Tell Bazel to not use $HOME/.cache/bazel at all
+        # See site/docs/output_directories.md
```
boegel: Please point to https://docs.bazel.build/versions/master/output_directories.html

boegel:
Going in, thanks @akesandgren!

PR #1705, opened 2019-04-29
also set $XDG_CACHE_HOME during installation of Python, to ensure that 'pip' that comes along with it doesn't touch $HOME/.cache/pip

boegel:
Without this, `make install` fails for Python 3.6.6 if `$HOME/.cache/pip` is read-only at the time it is executed...


akesandgren:
Going in, thanks @boegel!

PR #1707, opened 2019-05-01
CHARMM - Add prebuildopts and runtest options

vanzod:
The current implementation of `charmm.py` does not support the use of `prebuildopts`. This is important to build `charmm/v43b2` since the `mpiifort` compiler must be forced by passing the `MPIIFORT=YES` variable.
In addition, I have also added the option to skip the tests, which helps a lot when just testing the build.

### DIFF
--- a/easybuild/easyblocks/c/charmm.py
+++ b/easybuild/easyblocks/c/charmm.py
@@ -110,7 +110,8 @@ class EB_CHARMM(EasyBlock):
         if self.toolchain.comp_family() == toolchain.INTELCOMP:
             build_options += " IFORT"
 
-        cmd = "%s ./install.com %s %s %s" % (self.cfg['prebuildopts'], self.arch, self.cfg['system_size'], build_options)
+        cmd = "%s ./install.com %s %s %s" % \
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/c/charmm.py
```diff
@@ -109,18 +110,21 @@ def build_step(self, verbose=False):
         if self.toolchain.comp_family() == toolchain.INTELCOMP:
             build_options += " IFORT"
 
-        cmd = "./install.com %s %s %s" % (self.arch, self.cfg['system_size'], build_options)
+        cmd = "%s ./install.com %s %s %s" % \
+              (self.cfg['prebuildopts'], self.arch, self.cfg['system_size'], build_options)
```
boegel: Would be cleaner as:

```python
cmd = ' '.join([self.cfg['prebuildopts'], './install.com', self.arch, self.cfg['system_size'], build_options])

boegel:
Re-tested with `CHARMM-37b2-intel-2016a.eb`, still works like a ... charm, so good to go, thanks @vanzod!

PR #1708, opened 2019-05-02
Make sure perl install dir follows a standard format

### LINKED ISSUE #1700: Perl sanity check fails if prefix contains the string "perl"
When installing `Perl-5.28.1-GCCcore-8.2.0.eb` with EB 3.9.0 (also happens with EB 3.8.1) the sanity check fails with following error: 
```
== FAILED: Installation ended unsuccessfully (build directory: /software/build-tmp/Perl/5.28.1/GCCcore-8.2.0): build failed (first 300 chars): Sanity check failed: no (non-empty) directory found at 'lib/perl5/5.28.1' in /software/2019/software/perl/5.28.1-gcccore-8.2.0
== Results of the build can be found in the log file(s) /tmp/eb-2_vi5K/easybuild-Perl-5.28.1-20190426.135243.FZqzX.log
ERROR: Build of /software/system/easyconfigs/easybuild/easyconfigs/p/Perl/Perl-5.28.1-GCCcore-8.2.0.eb failed (err: "build failed (first 300 chars): Sanity check failed: no (non-empty) directory found at 'lib/perl5/5.28.1' in /software/2019/software/perl/5.28.1-gcccore-8.2.0")
```
The install folder misses the `perl5` folder:
```
[easybuild@clip-login-1 ~]$ ls -la /software/2019/software/perl/5.28.1-gcccore-8.2.0
total 28
drwxrwxr-x. 5 easybuild easybuild  4096 Apr 26 14:25 .
drwxrwxr-x. 3 easybuild easybuild  4096 Apr 26 14:25 ..
drwxr-xr-x. 2 easybuild easybuild 12288 Apr 26 14:35 bin
drwxr-xr-x. 4 easybuild easybuild  4096 Apr 26 14:25 lib
drwxr-xr-x. 4 easybuild easybuild  4096 Apr 26 14:25 man
```
The build log indicatges that botih `site_perl` as well as the `5.28.1` folder are directly installed into the `lib` folder and not into `perl5`:

```
Installation prefix to use for add-on modules and utilities? (~name ok) [/software/2019/software/perl/5.28.1-gcccore-8.2.0]  

The installation process will create a directory for
site-specific extensions and modules.  Most users find it convenient
to place all site-specific files in this directory rather than in the
main distribution directory.

Pathname for the site-specific library files? (~name ok) [/software/2019/software/perl/5.28.1-gcccore-8.2.0/lib/site_perl/5.28.1]  
```
According to https://metacpan.org/pod/distribution/perl/INSTALL#Common-Configure-options this seems to be expected behavior:

> If your prefix contains the string "perl", then the suggested directory structure is simplified. For example, if you use prefix=/opt/perl, then Configure will suggest /opt/perl/lib instead of /opt/perl/lib/perl5/

ocaisa:
Fixes #1700

### DIFF
--- a/easybuild/easyblocks/p/perl.py
+++ b/easybuild/easyblocks/p/perl.py
@@ -53,6 +53,7 @@ class EB_Perl(ConfigureMake):
         """
         Configure Perl build: run ./Configure instead of ./configure with some different options
         """
+        majver = self.version.split('.')[0]
         configopts = [
             self.cfg['configopts'],
             '-Dcc="{0}"'.format(os.getenv('CC')),
@@ -64,6 +65,9 @@ class EB_Perl(ConfigureMake):
             '-Dinstallscript=%(installdir)s/bin',
             '-Dscriptdir=%(installdir)s/bin',
             '-Dscriptdirexp=%(installdir)s/bin',
+            # guarantee that the install directory has the form lib/perlX/
+            # see https://github.com/easybuilders/easybuild-easyblocks/issues/1700
+            "-Dinstallstyle = 'lib/perl%s'" % majver,
         ]
         if self.cfg['use_perl_threads']:
             configopts.append('-Dusethreads')
--- a/easybuild/easyblocks/p/perl.py
+++ b/easybuild/easyblocks/p/perl.py
@@ -67,7 +67,7 @@ class EB_Perl(ConfigureMake):
             '-Dscriptdirexp=%(installdir)s/bin',
             # guarantee that the install directory has the form lib/perlX/
             # see https://github.com/easybuilders/easybuild-easyblocks/issues/1700
-            "-Dinstallstyle = 'lib/perl%s'" % majver,
+            "-Dinstallstyle='lib/perl%s'" % majver,
         ]
         if self.cfg['use_perl_threads']:
             configopts.append('-Dusethreads')


timeu:
Doesn't this also require the `exts_default_options = {'configopts': ''}` statement ?

ocaisa:
No, this is specifically for the perl installation, the perl extensions are handled by another easyblock (which is unaffected by this change).
With this change you wouldn't need either of the options you had to include in your easyconfig solution.

timeu:
@ocaisa : Ah I understand, makes sense Thanks for clarifying !

boegel:
@timeu Can you re-test with an unmodified `Perl` easyconfig (so without the changes used in #1700 to dance around the problem) with this patched easyblock (using `eb --include-easyblocks perl.py ...`), to confirm the fix?

easybuild/easyblocks/p/perl.py
```diff
@@ -64,6 +65,9 @@ def configure_step(self):
             '-Dinstallscript=%(installdir)s/bin',
             '-Dscriptdir=%(installdir)s/bin',
             '-Dscriptdirexp=%(installdir)s/bin',
+            # guarantee that the install directory has the form lib/perlX/
+            # see https://github.com/easybuilders/easybuild-easyblocks/issues/1700
+            "-Dinstallstyle = 'lib/perl%s'" % majver,
```
boegel: @ocaisa let's drop the spaces around `=` here, to be in sync with the other configure options

boegel: This may cause this problem I'm seeing when testing rebuild of `Perl` using this updated easyblock...

```
Sanity check failed: no (non-empty) directory found at 'lib/perl5/5.28.1'
```

timeu:
@boegel Do you still need me to test it ?

ocaisa:
Yes, can you retest please just so we know it actually does fix the problem

timeu:
Ok tried it with the unmodified easyconfig and the patched easyblock and it seems to work: 
```
[easybuild@stg-login-0 Perl]$ eb Perl-5.28.1-GCCcore-8.2.0.eb --include-easyblocks /software/system/easyblocks/perl.py --force
== temporary log file in case of crash /tmp/eb-lvfRPB/easybuild-4F4MKc.log
== processing EasyBuild easyconfig /software/system/easyconfigs/easybuild/easyconfigs/p/Perl/Perl-5.28.1-GCCcore-8.2.0.eb
== building and installing perl/5.28.1-gcccore-8.2.0...
== fetching files...
== creating build dir, resetting environment...
== unpacking...
...
== installing extension App::cpanminus 1.7044 (289/289)...
== restore after iterating...
== postprocessing...
== sanity checking...
== cleaning up...
== creating module...
== permissions...
== packaging...
== COMPLETED: Installation ended successfully
== Results of the build can be found in the log file(s) /software/2019/software/perl/5.28.1-gcccore-8.2.0/easybuild/easybuild-Perl-5.28.1-20190503.121419.log
== Build succeeded for 1 out of 1
== Temporary log file(s) /tmp/eb-lvfRPB/easybuild-4F4MKc.log* have been removed.
== Temporary directory /tmp/eb-lvfRPB has been removed.
```

boegel:
Thanks a lot for confirming the fix @timeu!

PR #1710, opened 2019-05-03
Handle components similar to extensions in framework

ocaisa:


ocaisa:
Requires https://github.com/easybuilders/easybuild-framework/pull/2855

ocaisa:
Closing this since there is a more straightforward solution to the particular problem in https://github.com/easybuilders/easybuild-easyblocks/pull/1711

PR #1711, opened 2019-05-06
Attach source_urls directly to sources for components in generic Bundle easyblock

### LINKED ISSUE #1674: Bundle easyblock and downloading of files.
I ran into an issue with the Bundle EasyBlock that could either be a feature with undesirable side-effects or a bug depending on how you look at it. See the EasyConfig which I attached to this message.

It turns out that Bundle during the fetch phase throws the "source_urls" for all components into one big pool, then starts fetching the requested files not trying only the source_urls given for that bundle component, but trying all given source_urls in order. In the attached EasyConfig this goes wrong after the second bundle component (libcerf). The server serving that source_urls has the undesirable feature to not produce an error message when requesting a different .tar.gz or .tar.bz2 file than the one served (libcerf) and instead create a file with the requested name and the latest version of libcerf into it. To see this (apart from the text in the log files), go to the directory where sources are stored after the fetch step and do a tar -tf of gmp-6.1.2.tar.bz2, mpc-1.1.0.tar.gz and mpfr-4.0.2.tar.xz. The file for mpfr has the expected contents, the once for gmp and mpc do not but contains a version of libcerf.

It is not an absolute showstopper. In fact, in comments in the EasyConfig I show the workaround (by specifying the source location of libcerf in a different way), but it is very confusing while developing the bundle EasyConfig... Of course one could blame the server, but I've run in other servers already with similar undesirable behaviour when installing packages that have multiple source files (I don't remember which one), where one needed to be very careful with the order of the URLs (which again I solved by using the same trick as I did in this EasyConfig to make things more robust).

Attachement: a sample EasyConfig showing the problem.
[baselibs-devel-2019a-GCCcore-8.2.0-downloadMPC.eb.txt](https://github.com/easybuilders/easybuild-easyblocks/files/3043435/baselibs-devel-2019a-GCCcore-8.2.0-downloadMPC.eb.txt)

ocaisa:
Fixes #1674

### DIFF
--- a/easybuild/easyblocks/generic/bundle.py
+++ b/easybuild/easyblocks/generic/bundle.py
@@ -85,70 +85,71 @@ class Bundle(EasyBlock):
             if len(comp) == 3:
                 comp_specs = comp[2]
 
-            cfg = self.cfg.copy()
+            comp_cfg = self.cfg.copy()
 
             easyblock = comp_specs.get('easyblock') or self.cfg['default_easyblock']
             if easyblock is None:
-                raise EasyBuildError("No easyblock specified for component %s v%s", cfg['name'], cfg['version'])
+                raise EasyBuildError("No easyblock specified for component %s v%s", comp_cfg['name'], comp_cfg['version'])
             elif easyblock == 'Bundle':
                 raise EasyBuildError("The Bundle easyblock can not be used to install components in a bundle")
 
-            cfg.easyblock = get_easyblock_class(easyblock, name=cfg['name'])
+            comp_cfg.easyblock = get_easyblock_class(easyblock, name=comp_cfg['name'])
 
             # make sure that extra easyconfig parameters are known, so they can be set
-            extra_opts = cfg.easyblock.extra_options()
-            cfg.extend_params(copy.deepcopy(extra_opts))
+            extra_opts = comp_cfg.easyblock.extra_options()
+            comp_cfg.extend_params(copy.deepcopy(extra_opts))
 
-            cfg['name'] = comp_name
-            cfg['version'] = comp_version
-            cfg.generate_template_values()
+            comp_cfg['name'] = comp_name
+            comp_cfg['version'] = comp_version
+            comp_cfg.generate_template_values()
 
             # do not inherit easyblock to use from parent (since that would result in an infinite loop in install_step)
-            cfg['easyblock'] = None
+            comp_cfg['easyblock'] = None
 
             # reset list of sources/source_urls/checksums
-            cfg['sources'] = cfg['source_urls'] = cfg['checksums'] = []
+            comp_cfg['sources'] = comp_cfg['source_urls'] = comp_cfg['checksums'] = []
 
             for key in self.cfg['default_component_specs']:
-                cfg[key] = self.cfg['default_component_specs'][key]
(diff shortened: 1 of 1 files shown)


akesandgren:
Looks sane, should there be a test for this too?

ocaisa:
There aren't specific tests for functionality within individual easyblocks are there? There are some tests for `--module-only` but other than that I can't find any. I don't know if I'm familiar enough with the testing to set that all up from scratch.

easybuild/easyblocks/generic/bundle.py
```diff
@@ -120,15 +120,25 @@ def __init__(self, *args, **kwargs):
...
+                # If per-component source URLs are provided, attach them directly to the relevant sources
+                if cfg['source_urls']:
+                    for source in cfg['sources']:
+                        if isinstance(source, basestring):
+                            self.cfg.update('sources', [{'filename': source, 'source_urls': cfg['source_urls']}])
+                        elif isinstance(source, dict):
+                            # Update source_urls in the 'source' dict to use the one for the components
+                            # (if it doesn't already exist)
+                            comp_source_urls = source.get('source_urls', cfg['source_urls'])
+                            self.cfg.update('sources', [source.update('source_urls', comp_source_urls)])
```
boegel: This doesn't look correct...

The `.update` for `self.cfg` is something we've implemented ourselves (see `EasyConfig` class).

The `.update` of regular `dict` value (like `source` here) only accepts a `dict` value to update, and it doesn't return anything.

I think this is what you're intending:

```python
source['source_urls'] = comp_source_urls
self.cfg.update('sources', source)

**Comment by boegel** [comment]
There's no easy way to come up with a test for this in the `easyblocks` test suite, the only real way to test is to use this updated `Bundle` easyblock for easyconfigs that use it (+ use `components`), which is `X11`, `FALCON` and recent `PyQt(5)` easyconfigs.

That, and checking whether the problem @klust reported is actually fixed.

**Comment by ocaisa** [comment]
I checked this against X11 and against the easyconfig that prompted the PR (as well as the easyconfig solution for that where the source was a dict), all worked with the latest version.

File: easybuild/easyblocks/generic/bundle.py [hunk 2]
```diff
@@ -120,15 +120,25 @@ def __init__(self, *args, **kwargs):
...
-                self.cfg.update('sources', cfg['sources'])
+                # If per-component source URLs are provided, attach them directly to the relevant sources
+                if cfg['source_urls']:
+                    for source in cfg['sources']:
+                        if isinstance(source, basestring):
+                            self.cfg.update('sources', [{'filename': source, 'source_urls': cfg['source_urls']}])
+                        elif isinstance(source, dict):
+                            # Update source_urls in the 'source' dict to use the one for the components
+                            # (if it doesn't already exist)
+                            source['source_urls'] = source.get('source_urls', cfg['source_urls'])
```
boegel: I think this makes more sense/is clearer:

```python
if 'source_urls' not in source:
    source['source_urls'] = cfg['source_urls']

ocaisa:
@boegel I did the renaming

boegel:
re-tested with `X11`, `PyQt`, `PyQt5` and `FALCON` bundles that use `components`, all's good...

Thanks a lot for looking into this @ocaisa!

PR #1712, opened 2019-05-06
MATLAB: allow specifying the license file directly

smoors:


easybuild/easyblocks/m/matlab.py
```diff
@@ -74,15 +74,16 @@ def configure_step(self):
         licport = self.cfg['license_server_port']
         if licport is None:
             licport = os.getenv('EB_MATLAB_LICENSE_SERVER_PORT', '00000')
```
migueldiascosta: for consistency, everything regarding `licserv` and `licport` should also be inside `if licfile is None:` then, since it's only used there?

Micket:
While, not a huge problem, we are a bit inconsistent when it comes to specifying licenses.
Rather than hardcoding anything into any configs, shouldn't we just keep environment variables for all of them?
Among software I've installed, I found;
```
export LMCOMSOL_LICENSE_FILE=...
export INTEL_LICENSE_FILE=...
export EB_ABAQUS_LICENSE_FILE=...
```
I think we should, at least, allow for the use of `export EB_MATLAB_LICENSE_FILE=...`

damianam:
@Micket I don't disagree with you, but I think this might be better suited for a broader license related PR. This one does just one thing, and seems to do it well. I see no reason to don't merge it.

I'll wait for your reply to merge it.

bartoldeman:
@migueldiascosta ping?

migueldiascosta:
Going in, thanks @smoors!

PR #1713, opened 2019-05-07
avoid that --no-build-isolation option is specified multiple times

boegel:
fixes minor issue where `--no-build-isolation` gets passed multiple times for Python packages being installed via `multi_deps` (which doesn't cause any problems, but it's just silly)


migueldiascosta:
Going in, thanks @boegel!

PR #1716, opened 2019-05-08
Enable GPU offloading in Clang if CUDA is included as a dep

ocaisa:
and support building lld

### DIFF
--- a/easybuild/easyblocks/c/clang.py
+++ b/easybuild/easyblocks/c/clang.py
@@ -222,7 +222,7 @@ class EB_Clang(CMakeMake):
             self.cfg.update('configopts', "-DLLVM_ENABLE_ASSERTIONS=OFF")
 
         if self.cfg["build_lld"]:
-            self.cfg.update('configopts', "-DLLVM_ENABLE_PROJECTS=lld ")
+            self.cfg.update('configopts', "-DLLVM_ENABLE_PROJECTS=lld")
 
         build_targets = self.cfg['build_targets']
         if build_targets is None:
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/c/clang.py
```diff
@@ -237,6 +242,15 @@ def configure_step(self):
         else:
             self.cfg['configopts'] += "-DLLVM_ENABLE_ASSERTIONS=OFF "
 
+        if self.cfg["build_lld"]:
+            self.cfg['configopts'] += "-DLLVM_ENABLE_PROJECTS=lld "
```
boegel: @ocaisa I don't think this works, does it?

Should be

```python
self.cfg.update('configopts', "-DLLVM_ENABLE_PROJECTS=lld")

File: easybuild/easyblocks/c/clang.py [hunk 2]
```diff
@@ -237,6 +242,15 @@ def configure_step(self):
...
+
+        # If CUDA is included as a dep, add it as a target
+        if get_software_root("CUDA") and "NVPTX" not in self.cfg['build_targets']:
+            self.cfg['build_targets'] += ["NVPTX"]
```
boegel: Use `self.cfg.update`, this most likely doesn't have the intended effect of adding `NVPTX` to the `build_targets` list...

The problem is that `self.cfg['build_targets']` returns a *copy* of the value (with templates resolved), which is why you have to use `self.cfg.update`

easybuild/easyblocks/c/clang.py
```diff
@@ -237,6 +242,15 @@ def configure_step(self):
...
+            if self.cfg["usepolly"]:
+                self.cfg['configopts'] += "-DPOLLY_ENABLE_GPGPU_CODEGEN=ON  "
```
boegel: also must be `self.cfg.update` here for this to work

easybuild/easyblocks/c/clang.py
ocaisa: It's the same as existing code, see 3 lines up. Should I change it everywhere?

easybuild/easyblocks/c/clang.py
ocaisa: It does indeed work but I don't mind changing it

boegel: Hmm, ok, I'm probably confusing with a similar thing that doesn't work as expected.

I can confirm that the existing `+=` statements for e.g. `-DCMAKE_BUILD_TYPE` work as intended, but using `self.cfg.update` makes more sense.

ocaisa:
I moved defining the build targets until the configure step so that we can automatically include architectures that require a dependency (`CUDA`) while still respecting if someone sets the architectures explicitly.

boegel:
lgtm, tested with Clang 8.0.0 easyconfigs from #8254 and #8276, so going in, thanks @ocaisa!

PR #1717, opened 2019-05-09
add support to Clang easyblock for also building libc++ and RTTI

ocaisa:


ocaisa:
Need to add an additional sanity check once I know what the libs are called

ocaisa:
Mostly following the lead of https://github.com/spack/spack/blob/e4ce4e5c2c3ae83d24965ed95b3d6cf6aa94b9cf/var/spack/repos/builtin/packages/llvm/package.py here

easybuild/easyblocks/c/clang.py
```diff
@@ -397,6 +410,10 @@ def sanity_check_step(self):
         if self.cfg["build_lld"]:
             custom_paths['files'].extend(["bin/lld"])
 
+        if self.cfg["libcxx"]:
+            custom_paths['files'].extend(["lib/libc++.so"])
+            custom_paths['files'].extend(["lib/libc++abi.so"])
```
Micket: This should use `shlib_ext` right?

boegel: Ideally, yes.

easybuild/easyblocks/c/clang.py
```diff
@@ -215,21 +222,27 @@ def configure_step(self):
         self.cfg.update('configopts', "-DGCC_INSTALL_PREFIX='%s'" % gcc_prefix)
         self.log.debug("Using %s as GCC_INSTALL_PREFIX", gcc_prefix)
 
+        # Configure some default options
         self.cfg.update('configopts', "-DCMAKE_BUILD_TYPE=Release")
+        self.cfg.update('configopts', '-DLLVM_REQUIRES_RTTI=ON')
+        self.cfg.update('configopts', '-DLLVM_ENABLE_RTTI=ON')
+        self.cfg.update('configopts', '-DLLVM_ENABLE_EH=ON')
+        self.cfg.update('configopts', '-DCLANG_DEFAULT_OPENMP_RUNTIME=libomp')
```
boegel: @ocaisa What if someone wants to build with different options?

Also, do all of these make sense for the oldest `Clang` versions we support?

easybuild/easyblocks/c/clang.py
```diff
@@ -215,21 +222,27 @@ def configure_step(self):
...
+
         if self.cfg['assertions']:
             self.cfg.update('configopts', "-DLLVM_ENABLE_ASSERTIONS=ON")
         else:
             self.cfg.update('configopts', "-DLLVM_ENABLE_ASSERTIONS=OFF")
 
-        if self.cfg["build_lld"]:
-            self.cfg.update('configopts', "-DLLVM_ENABLE_PROJECTS=lld")
```
boegel: Why is this being dropped? Is it not required when `build_lld` is enabled after all?

Micket: I verified that I got both libc++ and lld when I used this for my Clang PR.

easybuild/easyblocks/c/clang.py
ocaisa: Will change this

easybuild/easyblocks/c/clang.py
ocaisa: Answer to that is I have no idea, I don't know what they do, I just know Slack have them set. I can just wrap these options in a version check (for Clang 8)?

easybuild/easyblocks/c/clang.py
ocaisa: It's because I placed the lld sources in a different directory, when they are placed there there is no longer a need for this option.

boegel: ok, thanks for clarifying

boegel:
@ocaisa Should having `enable_rtti` or `default_openmp_runtime` set also trigger more check in the sanity check? Not sure whether it's possible/easy...

ocaisa:
@boegel I don't know if these options actually trigger additional files being built (but I suspect not). Maybe there a way to query these from the clang compiler? If google finds me a solution I'll add it

ocaisa:
Ok, what I found was that RTTI is disabled by default, what the `enable_rtti` related settings do is enable it by default (see https://llvm.org/docs/Packaging.html#c-features and https://github.com/idaholab/package_builder/issues/114 ). I can't find a way to have clang dump what it's settings are though so I can't verify these in the sanity check.

For the `libomp` related flag, [this discussion](http://clang-developers.42468.n3.nabble.com/switching-CLANG-DEFAULT-OPENMP-RUNTIME-to-libomp-td4045825.html) implies this is already the default since 2015 (and you can override it already with `-fopenmp=<something else>`) so there is no need for it in easyconfigs (see ) but it also does no harm to include as an option.

boegel:
lgtm now, running some more tests but I expect this to go in really soon...

boegel:
Thanks @ocaisa!

PR #1718, opened 2019-05-10
Allow Boost to build both Python2 and Python3 libraries

ocaisa:
via use of `multi_deps`

### DIFF
--- a/easybuild/easyblocks/b/boost.py
+++ b/easybuild/easyblocks/b/boost.py
@@ -227,7 +227,8 @@ class EB_Boost(EasyBlock):
         """Install Boost by copying files to install dir."""
 
         self.log.info("Copying %s to installation dir %s" % (self.objdir, self.installdir))
-        if self.cfg['multi_deps'] and self.iter_idx > 0:
+        # self.iter_idx will have already been bumped to prepare for the next iteration by the time we check it here
+        if self.cfg['multi_deps'] and self.iter_idx > 1:
             self.log.info("Main installation should already exist, only copying over missing Python libraries.")
             copy(glob.glob(os.path.join(self.objdir, 'lib', 'libboost_python*')), os.path.join(self.installdir, 'lib'))
         else:
--- a/easybuild/easyblocks/b/boost.py
+++ b/easybuild/easyblocks/b/boost.py
@@ -228,7 +228,7 @@ class EB_Boost(EasyBlock):
 
         self.log.info("Copying %s to installation dir %s" % (self.objdir, self.installdir))
         # self.iter_idx will have already been bumped to prepare for the next iteration by the time we check it here
-        if self.cfg['multi_deps'] and self.iter_idx > 1:
+        if self.cfg['multi_deps'] and self.current_iteration > 0:
             self.log.info("Main installation should already exist, only copying over missing Python libraries.")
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/b/boost.py
```diff
@@ -227,7 +227,15 @@ def install_step(self):
         """Install Boost by copying file to install dir."""
 
         self.log.info("Copying %s to installation dir %s" % (self.objdir, self.installdir))
-        copy(glob.glob(os.path.join(self.objdir, '*')), self.installdir)
+        try:
+            copy(glob.glob(os.path.join(self.objdir, '*')), self.installdir)
+        except EasyBuildError as return_error:
```
boegel: @ocaisa Rather than trying and letting it fail, we should just check which iteration is running when `multi_deps` is used, and only copy the Python libraries from the 2nd iteration onwards.

```python
if self.cfg['multi_deps'] and self.iter_idx > 0:
    # only copy Python libs
else:
    copy(glob.glob(os.path.join(self.objdir, '*')), self.installdir)
```

Also, the sanity check should probably be enhanced so it checks for both `libboost_python*` libraries?

ocaisa: You mean the sanity check is not run independently for each iteration? I thought it was, other wise how can you check a python import?

easybuild/easyblocks/b/boost.py
```diff
@@ -224,10 +224,14 @@ def build_step(self):
         run_cmd(cmd, log_all=True, simple=True)
 
     def install_step(self):
-        """Install Boost by copying file to install dir."""
+        """Install Boost by copying files to install dir."""
 
         self.log.info("Copying %s to installation dir %s" % (self.objdir, self.installdir))
-        copy(glob.glob(os.path.join(self.objdir, '*')), self.installdir)
+        if self.cfg['multi_deps'] and self.iter_idx > 0:
```
ocaisa: @boegel `self.iter_idx` is never zero here, for the first iteration is it already 1. Is that expected?

ocaisa:
Now requires https://github.com/easybuilders/easybuild-framework/pull/2886

ocaisa:
So injecting the Python specific files is hard, currently I don't have a way of accessing the specific environment of one of iterations in the easyblock. Ideas @boegel? I need different different `custom_paths` for each iteration but currently I generate this once and use it twice.

Ideas @boegel ?

ocaisa:
The only workaround I can figure out is to leverage `custom_commands`, these are the only things that are (currently) executed in the correct environment during the sanity check (file checks are done before fake modules are loaded).

easybuild/easyblocks/b/boost.py
boegel: No, fixed in https://github.com/easybuilders/easybuild-framework/pull/2886

easybuild/easyblocks/b/boost.py
boegel: The sanity check is run for both iterations, my bad.

boegel:
@ocaisa The changes in https://github.com/ocaisa/easybuild-easyblocks/pull/17 should make this good to go (I tested it)

boegel:
tested with https://github.com/easybuilders/easybuild-easyconfigs/pull/8415, good to go, thanks @ocaisa!

PR #1719, opened 2019-05-10
Allow option to build a thread MPI version of GROMACS

branfosj:
This allows for building GROMACS with thread MPI turned on, but leaves the default as it is.


akesandgren:
Gromacs easyblock already builds the threadMPI version and the MPI version at the same time.
So there is no need for this.

akesandgren:
To clarify, the initial configure_step configures a threadMPI build, which is then installed first thing in the install_step. (First call to super(EB_GROMACS, self).install_step() in install_step)
the Install step then (line 377) reconfigures for MPI if usempi is set and runs the configure/build/test/install steps again.

branfosj:
Doh!
Thanks for the explanation. I'd misunderstood what the earlier part was doing and not realised that that gives us the threadMPI version.

PR #1721, opened 2019-05-10
file() was removed in Python 3, use write_file() instead in Rosette easyblock

cclauss:


easybuild/easyblocks/r/rosetta.py
```diff
@@ -163,7 +163,7 @@ def configure_step(self):
         us_fp = os.path.join(self.srcdir, "tools/build/user.settings")
         try:
             self.log.debug("Creating '%s' with: %s" % (us_fp, txt))
-            f = file(us_fp, 'w')
+            f = open(us_fp, 'w')
```
boegel: Nice catch, we clearly overlooked this one when porting the easyblocks to also support Python 3 (which is currently in need for more testing, otherwise we would've noticed already).

Since we have support in the EasyBuild framework for a dedicated `write_file` function (cfr. https://easybuild.readthedocs.io/en/latest/api/easybuild.tools.filetools.html#easybuild.tools.filetools.write_file), we should use that instead here (added benefit is that we can drop the `try`/`catch` as well, since clean error reporting is handled by `write_file` itself).

So this whole block should become something like:

```python
us_fp = os.path.join(self.srcdir, "tools/build/user.settings")
self.log.debug("Creating '%s' with: %s", us_fp, txt)
write_file(us_fp, txt)

Comment (cclauss) [reply easybuild/easyblocks/r/rosetta.py]: No f"strings"?  ;-)...  __self.log.debug(f"Creating '{us_fp}' with: {txt}")__

File: easybuild/easyblocks/r/rosetta.py [hunk 2]
```diff
@@ -161,13 +161,8 @@ def configure_step(self):
...
         us_fp = os.path.join(self.srcdir, "tools/build/user.settings")
-        try:
-            self.log.debug("Creating '%s' with: %s" % (us_fp, txt))
-            f = file(us_fp, 'w')
-            f.write(txt)
-            f.close()
-        except IOError as err:
-            raise EasyBuildError("Failed to write settings file %s: %s", us_fp, err)
+        self.log.debug("Creating '%s' with: %s", us_fp, txt)
+        write_file(us_fp, txt)
```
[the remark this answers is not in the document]
boegel: @cclauss You’ll need to add the missing import statement as well (from easybuild.tools.filetools), and note that we try to keep import lines sorted alphabetically

PR #1722, opened 2019-05-10
fixed PGI easyblock sanity check for version 19

mboisson:
PGI 19 and more recent drops pgf77. This PR fixes the sanity checks.

### DIFF
--- a/easybuild/easyblocks/p/pgi.py
+++ b/easybuild/easyblocks/p/pgi.py
@@ -172,8 +172,13 @@ class EB_PGI(PackedBinary):
     def sanity_check_step(self):
         """Custom sanity check for PGI"""
         prefix = self.pgi_install_subdir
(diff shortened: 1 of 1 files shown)


mboisson:
Ah, I see somebody else had removed pgf77 already. I suggest modifying it for pgf90 instead, since this is present in PGI 19.

boegel:
@mboisson Indeed, see https://github.com/easybuilders/easybuild-easyblocks/pull/1653 .

I don't mind checking for `pgf90` instead, but you'll need to fix the conflict first here (or open a new PR starting from the `develop` branch)...

boegel:
@mboisson `pgf90` has been there since PGI 15.1, which is the oldest version we have easyconfigs for, so no need to the version check imho...

mboisson:
Alright, we can close this PR then, and just drop pgf77.

PR #1724, opened 2019-05-12
fix Python classifiers in setup.py, should be (only) Python 2.6 & 2.7

boegel:


setup.py
```diff
@@ -90,7 +90,8 @@ def read(fname):
         "Intended Audience :: System Administrators",
```
akesandgren: Same thing here, change doesn't match easybuild 507.

boegel: @akesandgren Those additional classifiers were only added in the `4.x` branch (for the upcoming EasyBuild 4.0).

Here, I'm just fixing the Python version (like we also did in `develop` in the `easybuild` repo, see https://github.com/easybuilders/easybuild/pull/506/files)

akesandgren: okidoki, too many PRs too keep track of :-)

PR #1726, opened 2019-05-13
clean up support for 'fix_python_shebang_for' in PythonPackage

boegel:
since it's now implemented (slightly better) as a generic easyconfig parameter via https://github.com/easybuilders/easybuild-framework/pull/2878


akesandgren:
Going in, thanks @boegel!

PR #1728, opened 2019-05-15
use is_system_toolchain() rather than checking toolchain name against DUMMY_TOOLCHAIN_NAME constant

boegel:
requires ~~https://github.com/easybuilders/easybuild-framework/pull/2877~~


akesandgren:
Going in, thanks @boegel!

PR #1729, opened 2019-05-15
WIP: Flang

ocaisa:


easybuild/easyblocks/f/flang.py
```diff
@@ -0,0 +1,185 @@
...
+        change_dir(self.flang_build_dir)
+        super(EB_Clang, self).install_step()
+        # Copy over the flang executable from LLVM
+        copy_file(
+            os.path.join(self.llvm_obj_dir, 'bin', 'flang'),
+            os.path.join(self.installdir, 'bin', 'flang'),
+        )
+        # Also copy over the libomp libraries
+        copy_file(
+            os.path.join(self.llvm_obj_dir, 'lib', 'libomp.%s' % shlib_ext),
```
ocaisa: Bit dubious about whether this is enough there are a number of libomp libraries in that directory but this is the only one that appears in the *Flang Binary Release File List* at https://github.com/flang-compiler/flang/releases

ocaisa: I'm also concerned that an identical library (same source code) is created by Clang, are these compatible?

ocaisa:
Regarding `libomp` here, this is a bit of a problem because since `libomp` appears for both clang and flang we have a name clash if we couple these. This is an issue because Flang looks like it is going to be always be a generation behind Clang, so if we make a toolchain we need to know that when we use Flang we get the libomp from Flang, and when we use Clang we get Clang's libomp. 

We can mostly achieve this by aliasing flang with
```
flang -Wl,-rpath %(installdir)s/lib
```
And being careful that whenever we define a full LLVM toolchain that flang appears before clang (so the clang libomp is picked up by default).

The other option is to stick to a single LLVM version for a clang/flang toolchain and use the libomp from clang for flang as well. This asssumes that since they are the same code they must be compatible (be warned though, they are built with an LLVM fork for flang in the normal case, I have no idea what that means in practice).

ocaisa:
@geimer Maybe you might know the implications of the libomp issue?

geimer:
@ocaisa I haven't built `flang` myself yet.  But from the description/instructions, it reads as if the build includes the full suite of compilers, and the `flang` folks only needed to tweak the compiler driver to support Fortran.  That is, I would expect to have only one OpenMP runtime library used for all compilers.  But I may be mistaken...

easybuild/easyblocks/f/flang.py
```diff
@@ -0,0 +1,186 @@
...
+        LLVM_CONFIG = os.path.join(self.llvm_obj_dir, 'bin', 'llvm-config')
+        shlib_ext = get_shared_lib_ext()
+        LIBOMP = os.path.join(self.llvm_obj_dir, 'lib', 'libomp.%s' % shlib_ext)
+
+        options = "-DCMAKE_INSTALL_PREFIX=%s " % self.installdir
+        options += "-DCMAKE_C_COMPILER='%s' " % CC
+        options += "-DCMAKE_CXX_COMPILER='%s' " % CXX
+        options += "-DCMAKE_Fortran_COMPILER='%s' " % FC
+        options += "-DLLVM_CONFIG='%s' " % LLVM_CONFIG
+        options += "-DFLANG_LIBOMP='%s' " % LIBOMP
```
ocaisa: If we added Clang as a dependency I think it would be possible to just point to that `libomp` at this point (and not build Flang `libomp` at all).

easybuild/easyblocks/f/flang.py
```diff
@@ -0,0 +1,186 @@
...
+        # we've built llvm
+        self.flang_source_dir = os.path.join(self.llvm_src_dir, '..', 'flang')
+        find_source_dir('flang-flang_*', self.flang_source_dir)
+
+        for src in self.src:
+            for (dirname, new_path) in src_dirs.items():
+                if src['name'].startswith(dirname):
+                    old_path = os.path.join(src['finalpath'], dirname)
+                    try:
+                        shutil.move(old_path, new_path)
```
boegel: use `move_file` function from `filetools` (doesn't need to be in `try`/`except`, already does proper error handling)

easybuild/easyblocks/f/flang.py
```diff
@@ -0,0 +1,186 @@
...
+                    except IOError as err:
+                        raise EasyBuildError("Failed to move %s to %s: %s", old_path,
+                                             new_path, err)
+                    src['finalpath'] = new_path
+                    break
```
boegel: Shouldn't we somehow checked that all entries in `self.src` got moved? There's no guarantee of that now...

easybuild/easyblocks/f/flang.py
```diff
@@ -0,0 +1,186 @@
...
+        # Configure.
+        CC = os.path.join(self.llvm_obj_dir, 'bin', 'clang')
+        CXX = os.path.join(self.llvm_obj_dir, 'bin', 'clang++')
+        FC = os.path.join(self.llvm_obj_dir, 'bin', 'flang')
+        LLVM_CONFIG = os.path.join(self.llvm_obj_dir, 'bin', 'llvm-config')
+        shlib_ext = get_shared_lib_ext()
+        LIBOMP = os.path.join(self.llvm_obj_dir, 'lib', 'libomp.%s' % shlib_ext)
+
+        options = "-DCMAKE_INSTALL_PREFIX=%s " % self.installdir
+        options += "-DCMAKE_C_COMPILER='%s' " % CC
```
boegel: This could be cleaned up quite a bit...

```python
opts_map = {
    'CMAKE_INSTALL_PREFIX': self.installdir,
    'CMAKE_C_COMPILER': cc,
}
options = ' '.join('-D%s=%s' % item for item in sorted(opts_map.items()))
options += ' ' + ' '.join(additional_options) + ' ' + self.cfg['configopts']

File: easybuild/easyblocks/f/flang.py [hunk 6]
```diff
@@ -0,0 +1,186 @@
...
+        options += self.cfg['configopts']
+
+        self.log.info("Configuring")
+        run_cmd("cmake %s %s" % (options, src_dir), log_all=True)
+
+        self.log.info("Building")
+        if parallel:
+            run_cmd("make %s" % self.make_parallel_opts, log_all=True)
+        else:
+            run_cmd("make", log_all=True)
```
boegel:
```python
cmd = 'make'
if parallel:
    cmd += ' ' + self.make_parallel_opts
run_cmd(cmd, log_all=True)

File: easybuild/easyblocks/f/flang.py [hunk 7]
```diff
@@ -0,0 +1,186 @@
...
+        shlib_ext = get_shared_lib_ext()
+        self.build_with_temporary_llvm(
+            self.flang_build_dir,
+            self.flang_source_dir,
+            parallel=False,  # Can misbehave in parallel
+            additional_options=[
+                '-DLIBPGMATH=%s' % os.path.join(self.pgmath_build_dir, 'lib', 'libpgmath.%s' % shlib_ext),
+                # Ensure libraries are rpath-ed to the install directory (most relevant for libomp.so link)
+                '-DCMAKE_BUILD_WITH_INSTALL_RPATH=1',
+                '-DCMAKE_INSTALL_RPATH=%s/lib' % self.installdir,
```
boegel:
```python
'-DCMAKE_INSTALL_RPATH=%s' % os.path.join(self.installdir, 'lib'),

File: easybuild/easyblocks/f/flang.py [hunk 8]
```diff
@@ -0,0 +1,186 @@
...
+    """Support for bootstrapping Flang."""
+
+    def __init__(self, *args, **kwargs):
+        """Initialize custom class variables for Flang."""
+
+        super(EB_Flang, self).__init__(*args, **kwargs)
+
+    def disable_sanitizer_tests(self):
+        # Not relevant for flang
+        pass
```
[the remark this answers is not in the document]
ocaisa: No, I need to override what is in the clang easyblock here which attempts to patch files in the runtime (these do not exist for Flang).

easybuild/easyblocks/f/flang.py
```diff
@@ -0,0 +1,186 @@
...
+
+    def extract_step(self):
+        """
+        Prepare a combined LLVM source tree.  The layout is:
```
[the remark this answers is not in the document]
ocaisa: Actually it is mostly correct...unfortunately the final folder for the `flang` runtime does need to be called `clang`

easybuild/easyblocks/f/flang.py
akesandgren: We would likely want Clang+Flang as a single entity, just like we just merged icc/ifort
It makes for a much improved user experience not to have to understand that both icc and ifort have to be loaded despite Lmods spider saying you only need one...

ocaisa: For Flang, I don't think that is actually a good idea. Flang is (unfortunately) working off a forked version of Clang.

This is likely to very soon be out of date anyway...though I admit to being confused by the whole thing: `f18` was supposed to replace `flang` but it seems it will actually take over the name `flang` (see [here](https://www.phoronix.com/scan.php?page=news_item&px=LLVM-f18-Will-Be-Flang)). Once that get's a release then it is very likely that incorporating flang will become a whole lot easier...it is probably best to wait it out.

ocaisa:
This is expected to be out of date in a few months, and a much simpler solution possible:
http://fortranwiki.org/fortran/show/Flang
> By the beginning of 2020, F18 is expected to replace Flang as a complete implementation of Fortran 2018.

ocaisa:
F18 has now become `flang` (see https://github.com/flang-compiler/f18) and can be built as an LLVM package in the normal way, this can be closed once it is included in an LLVM release.

PR #1730, opened 2019-05-15
rebuild internal libfabric for impi 2019.x & more recent

damianam:
…better fit the running system

### DIFF
--- a/easybuild/easyblocks/i/impi.py
+++ b/easybuild/easyblocks/i/impi.py
@@ -38,7 +38,7 @@ from distutils.version import LooseVersion
 from easybuild.easyblocks.generic.intelbase import IntelBase, ACTIVATION_NAME_2012, LICENSE_FILE_NAME_2012
 from easybuild.framework.easyconfig import CUSTOM
 from easybuild.tools.build_log import EasyBuildError
-from easybuild.tools.filetools import apply_regex_substitutions, mkdir, write_file
+from easybuild.tools.filetools import apply_regex_substitutions, extract_file, mkdir, write_file
 from easybuild.tools.run import run_cmd
 from easybuild.tools.systemtools import get_shared_lib_ext
 
@@ -50,6 +50,8 @@ class EB_impi(IntelBase):
     @staticmethod
     def extra_options():
         extra_vars = {
+            'rebuild_libfabric': [True, 'Rebuild the internal shipped libfabric instead of using the provided binary', CUSTOM],
+            'libfabric_configopts': ['', 'Configure options for the provided libfabric', CUSTOM],
             'ofi_internal': [True, 'Use internal shipped libfabric instead of external libfabric', CUSTOM],
             'set_mpi_wrappers_compiler': [False, 'Override default compiler used by MPI wrapper commands', CUSTOM],
             'set_mpi_wrapper_aliases_gcc': [False, 'Set compiler for mpigcc/mpigxx via aliases', CUSTOM],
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/i/impi.py
```diff
@@ -141,6 +143,22 @@ def post_install_step(self):
             for script in [os.path.join(script_path, 'mpivars.sh') for script_path in script_paths]:
                 apply_regex_substitutions(os.path.join(self.installdir, script), regex_subs)
 
+        # Recompile libfabric if necessary
+        if impiver >= LooseVersion('2019.0.0') and self.cfg['rebuild_libfabric']:
+            if self.cfg['ofi_internal']:
+                os.chdir(os.path.join(self.installdir, 'libfabric'))
+                extract_file('src.tgz', './')
+                libfabric_installpath = os.path.join(self.installdir, 'intel64', 'libfabric')
+                cmds = [
```
damianam: I don't particularly like recreating `ConfigureMake` here, but it might be better than importing another easyblock. That looks even more awkward.

easybuild/easyblocks/i/impi.py
```diff
@@ -50,6 +50,8 @@ class EB_impi(IntelBase):
     @staticmethod
     def extra_options():
         extra_vars = {
+            'rebuild_libfabric': [True, 'Rebuild the internal shipped libfabric instead of using the provided binary', CUSTOM],
```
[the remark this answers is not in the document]
boegel: @damianam please fix long line, for example as follows:

```python
def extra_options():
        extra_vars = {
            'rebuild_libfabric': [True, "Rebuild the internal shipped libfabric "
                                        "instead of using the provided binary", CUSTOM],
            ...
```

Also, please move `rebuild_libfabric` down (keep these sorted alphabetically).

Bonus points if you can come up with a naming that keeps both new parameters together (`libfabric_rebuild` maybe?)...

easybuild/easyblocks/i/impi.py
```diff
@@ -141,6 +143,22 @@ def post_install_step(self):
             for script in [os.path.join(script_path, 'mpivars.sh') for script_path in script_paths]:
                 apply_regex_substitutions(os.path.join(self.installdir, script), regex_subs)
 
+        # Recompile libfabric if necessary
+        if impiver >= LooseVersion('2019.0.0') and self.cfg['rebuild_libfabric']:
```
boegel: `LooseVersion('2019')` is sufficient

easybuild/easyblocks/i/impi.py
```diff
@@ -141,6 +143,22 @@ def post_install_step(self):
...
+            if self.cfg['ofi_internal']:
+                os.chdir(os.path.join(self.installdir, 'libfabric'))
```
boegel: please use `change_dir` from `easybuild.tools.filetools` (which spits out a proper error message in case the `chdir` fails)

easybuild/easyblocks/i/impi.py
```diff
@@ -141,6 +143,22 @@ def post_install_step(self):
...
+                libfabric_installpath = os.path.join(self.installdir, 'intel64', 'libfabric')
+                cmds = [
+                    './configure --prefix=%s %s' % (libfabric_installpath, self.cfg['libfabric_configopts']),
+                    'make',
+                    'make install'
+                ]
+                for cmd in cmds:
+                    run_cmd(cmd, log_all=True, simple=True)
+            else:
+                self.log.info("rebuild_libfabric is True, ofi_internal is false. Ignoring the rebuild...")
```
boegel: This should be an error imho...

easybuild/easyblocks/i/impi.py
boegel: Yeah, it may be difficult to leverage `ConfigureMake` here, since you'll need to convince it to *not* pick up on `configopts`, `buildopts`, etc. Also, lot's of the logic in there is totally irrelevant here.

This isn't too terrible (and chances are that the build procedure for `libfabric` will change anyway)...

easybuild/easyblocks/i/impi.py
```diff
@@ -141,6 +143,22 @@ def post_install_step(self):
...
 
+        # Recompile libfabric if necessary
+        if impiver >= LooseVersion('2019.0.0') and self.cfg['rebuild_libfabric']:
+            if self.cfg['ofi_internal']:
+                os.chdir(os.path.join(self.installdir, 'libfabric'))
+                extract_file('src.tgz', './')
+                libfabric_installpath = os.path.join(self.installdir, 'intel64', 'libfabric')
+                cmds = [
+                    './configure --prefix=%s %s' % (libfabric_installpath, self.cfg['libfabric_configopts']),
+                    'make',
```
boegel: Should we care about `make -j N` here, or is rebuilding `libfabric` quick enough that it doesn't matter too much?

boegel:
@damianam Maybe also add a comment above the `rebuild_libfabric` option to explain why it *enabled* by default?

boegel:
I'm a little bit hesitant to enable re-building of `libfabric` by default, but I didn't run into any issues with it on our systems when re-installing the various `impi` 2019.x easyconfigs we have.

@bartoldeman Can you take this for a spin with `impi` 2019.x, see if you run into any problems?

vanzod:
LGTM. I have tested it with `impi-2019.3.199-iccifort-2019.3.199-GCC-8.3.0-2.32.eb` and it builds fine.

bartoldeman:
Builds fine for me too, even on CentOS-6 with `--try-toolchain=GCC,8.2.0-2.31.1`

PR #1731, opened 2019-05-15
update Q&A patterns in SAS easyblock for version 9.4 + add support for custom license file path

vanzod:
On our installation of the latest release of `SAS/9.4` a few additional questions were asked that were not explicitly contemplated by the easyblock. This PR adds definitions for such questions.

In addition, it now supports the ability to specify a custom path for the SAS Installation Data File. If it is left undefined, SAS will pick the default path within the extracted sources directory.

### DIFF
--- a/easybuild/easyblocks/s/sas.py
+++ b/easybuild/easyblocks/s/sas.py
@@ -61,7 +61,7 @@ class EB_SAS(EasyBlock):
             "SAS Home:": self.installdir,
             "Install SAS Software (default: Yes):": '',
             "Configure SAS Software (default: Yes):": '',
-            "SAS Installation Data File:": '',
+            "SAS Installation Data File:": self.license_file,
             "Press Enter to continue:": '',
             "Configure as a Unicode server (default: No):": 'N',
             "SAS/ACCESS Interface to MySQL (default: Yes):": 'N',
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/s/sas.py
```diff
@@ -34,6 +34,19 @@
...
+    def __init__(self, *args, **kwargs):
+        """Custom constructor for SAS easyblock, initialize custom class variables."""
+
+        super(EB_SAS, self).__init__(*args, **kwargs)
+
+        # Use default SAS Installation Data File path
+        self.license_file = ''
+
+        # Set custom SAS Installation Data File path if defined and existing
+        if self.cfg['license_file'] is not None and os.path.isfile(self.cfg['license_file']):
```
boegel: You can simplify this (the `is not None` is a bit silly, and otherwise you'll run into problems if `license_file` is set to `''`):

```python
if self.cfg['license_file'] and os.path.isfile(self.cfg['license_file']):

File: easybuild/easyblocks/s/sas.py [hunk 2]
```diff
@@ -34,6 +34,19 @@
...
+            self.license_file = self.cfg['license_file']
+            self.log.info("Custom SAS Installation Data File found: %s" % self.license_file)
```
boegel: `s/ % /, ` ->

```python
self.log.info("Custom SAS Installation Data File found: %s", self.license_file)

easybuild/easyblocks/s/sas.py
vanzod: That is something we should correct on other easyblocks since it is a recurring pattern.

boegel: open an issue on it if you don't have time to dive in?

boegel:
re-tested with `SAS-9.4.eb`, still works great, so going in, thanks @vanzod!

Issue #1350, opened 2018-01-27
binutils: zlib not correctly embedded


geimer:
The way in which `zlib` is embedded in `libbfd` is half-way broken.  While the `zlib` symbols correctly show up in `libbfd.so`
```
$ nm $EBROOTBINUTILS/lib/libbfd.so | grep inflate
00000000000f2d80 T inflate
00000000000f50a0 T inflateCodesUsed
00000000000f4e00 T inflateCopy
00000000000f49b0 T inflateEnd
[...]
```
they do not in `libbfd.a`:
```
$ nm $EBROOTBINUTILS/lib/libbfd.a | grep inflate
nm: libz.a: File format not recognized
                 U inflate
                 U inflateEnd
                 U inflateInit_
                 U inflateReset
```
As can be seen from the first line of the output, the reason is that the whole `libz.a` is included in `libbfd.a` rather than the individual object files of `libz.a`.  This leads to problems when linking a program statically.

Example:
```
$ cat foo.c
#define PACKAGE_NAME
#define PACKAGE_VERSION

#include <stdio.h>

#include <bfd.h>

int main(int argc, char** argv)
{
   bfd_init();

   bfd * bfdFile = bfd_openr( "/bin/ls", "elf64-x86-64" );
   if ( bfdFile == NULL )
   {
       printf( "Error [%x]: %s\n", bfd_get_error(), bfd_errmsg(bfd_get_error()) );
       return 1;
   }
   if ( !bfd_check_format( bfdFile, bfd_object ))
   {
       printf( "Error [%x]: %s\n", bfd_get_error(), bfd_errmsg(bfd_get_error()) );
       return 1;
   }

   bfd_close( bfdFile );
   return 0;
}

$ gcc -o foo foo.c -static -lbfd -liberty -ldl
/.../binutils/2.29.1/lib/libbfd.a(plugin.o):plugin.c:function try_load_plugin: warning: Using 'dlopen' in statically linked applications requires at runtime the shared libraries from the glibc version used for linking
/.../binutils/2.29.1/lib/libbfd.a(compress.o):compress.c:function decompress_contents: error: undefined reference to 'inflateInit_'
/.../binutils/2.29.1/lib/libbfd.a(compress.o):compress.c:function decompress_contents: error: undefined reference to 'inflateReset'
/.../binutils/2.29.1/lib/libbfd.a(compress.o):compress.c:function decompress_contents: error: undefined reference to 'inflate'
[...]

PR #1732, opened 2019-05-17
binutils: Only embed `zlib` if listed as a build dependency

geimer:
Addresses https://github.com/easybuilders/easybuild-easyblocks/issues/1350 in conjunction with https://github.com/easybuilders/easybuild-easyconfigs/pull/8340

This does *not* address the issue that `libz.a` is included in `libbfd.a` when specified as a build dependency. At the moment, I don't believe that this is fixable at all without rewriting parts of the Makefile(s).

### DIFF
--- a/easybuild/easyblocks/b/binutils.py
+++ b/easybuild/easyblocks/b/binutils.py
@@ -83,24 +83,24 @@ class EB_binutils(ConfigureMake):
         if zlibroot:
             self.cfg.update('configopts', '--with-system-zlib')
 
-        # statically link to zlib only if it is a build dependency
-        # see https://github.com/easybuilders/easybuild-easyblocks/issues/1350
-        build_deps = self.cfg.dependencies(build_only=True)
-        if any(dep['name'] == 'zlib' for dep in build_deps):
-            libz_path = os.path.join(zlibroot, get_software_libdir('zlib'), 'libz.a')
+            # statically link to zlib only if it is a build dependency
+            # see https://github.com/easybuilders/easybuild-easyblocks/issues/1350
+            build_deps = self.cfg.dependencies(build_only=True)
+            if any(dep['name'] == 'zlib' for dep in build_deps):
+                libz_path = os.path.join(zlibroot, get_software_libdir('zlib'), 'libz.a')
 
-            # for recent binutils versions, we need to override ZLIB in Makefile.in of components
-            if LooseVersion(self.version) >= LooseVersion('2.26'):
-                regex_subs = [
-                    (r"^(ZLIB\s*=\s*).*$", r"\1%s" % libz_path),
-                    (r"^(ZLIBINC\s*=\s*).*$", r"\1-I%s" % os.path.join(zlibroot, 'include')),
-                ]
-                for makefile in glob.glob(os.path.join(self.cfg['start_dir'], '*', 'Makefile.in')):
-                    apply_regex_substitutions(makefile, regex_subs)
+                # for recent binutils versions, we need to override ZLIB in Makefile.in of components
+                if LooseVersion(self.version) >= LooseVersion('2.26'):
+                    regex_subs = [
+                        (r"^(ZLIB\s*=\s*).*$", r"\1%s" % libz_path),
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/b/binutils.py
```diff
@@ -175,8 +180,9 @@ def sanity_check_step(self):
                 os.path.join('include', 'libiberty.h'),
             ])
 
-        # if zlib is listed as a dependency, it should get linked in statically
-        if get_software_root('zlib'):
+        # if zlib is listed as a build dependency, it should get linked in statically
```
boegel: `should get linked` -> `should have been linked`

easybuild/easyblocks/b/binutils.py
```diff
@@ -78,10 +78,15 @@ def configure_step(self):
...
+        # configure using `--with-system-zlib` if zlib is a (build) dependency
         zlibroot = get_software_root('zlib')
         if zlibroot:
             self.cfg.update('configopts', '--with-system-zlib')
+
+        # statically link to zlib only if it is a build dependency
+        # see https://github.com/easybuilders/easybuild-easyblocks/issues/1350
+        build_deps = self.cfg.dependencies(build_only=True)
+        if any(dep['name'] == 'zlib' for dep in build_deps):
             libz_path = os.path.join(zlibroot, get_software_libdir('zlib'), 'libz.a')
```
boegel: @geimer This should be intended one level more, there's no point in checking whether `zlib` was a build-only dep if `zlibroot` is not defined (since `get_software_libdir` will crash hard then)

boegel:
@geimer Since we're currently only including `zlib` as a build dep, this doesn't actually change much for existing easyconfigs (except when people have their own `binutils` easyconfigs where `zlib` is a runtime dep already), right?

I'm aware that we're changing that in https://github.com/easybuilders/easybuild-easyconfigs/pull/8340, just trying to make sure I have a good view on this w.r.t. backwards compatibility, since the idea is to include this with EasyBuild v3.9.1...

How much of a problem is it that `libz.a` is still embedded in `libbfd.a`?
In other words, if we merge this together with https://github.com/easybuilders/easybuild-easyconfigs/pull/8340, is #1350 actually fixed, or not?

geimer:
> Since we're currently only including zlib as a build dep, this doesn't actually change much for existing easyconfigs (except when people have their own binutils easyconfigs where zlib is a runtime dep already), right?

It should be a drop-in replacement with existing easyconfigs, correct.

> How much of a problem is it that libz.a is still embedded in libbfd.a?

With the modified easyconfigs for newer `binutils`, this should only happen for `binutils` built with `dummy`, which is only used for bootstrapping `GCCcore` and `binutils` built with that `GCCcore` (and stuff explicitly using the `dummy`-built `binutils`, but these should be rare, no?).  Thus it's unlikely to be a huge problem.  If you want that also fixed, I currently only see two options: (a) require `zlib` as an OS dependency for the `dummy`-built `binutils` and not embed anything (thus dancing around the conflicts and `module switch` problems) or (b) mess around in the Makefiles even more. While (b) *may* be doable, it requires much more investigation, i.e., postponing the fix to some release in the future.

> In other words, if we merge this together with easybuilders/easybuild-easyconfigs#8340, is #1350 actually fixed, or not?

With both PRs merged, #1350 *should* be fixed for everything built on top of `GCCcore` using the `binutils` built with that `GCCcore`, as the latter no longer embeds `zlib`.

easybuild/easyblocks/b/binutils.py
geimer: It won't crash, as there isn't a way to list `zlib` as a build dependency without `zlibroot` being defined. But I agree, it could be indented.

boegel:
Tested with a range of existing `binutils` easyconfigs old & new (which still have `zlib` as a build-only dep), no problems found.
Working test reports for https://github.com/easybuilders/easybuild-easyconfigs/pull/8340 are under way...

boegel:
No surprises during testing, so going in, thanks @geimer!

PR #1735, opened 2019-05-20
Don't check for python module when using dummy/system toolchain

wpoely86:


easybuild/easyblocks/generic/pythonbundle.py
```diff
@@ -83,7 +84,7 @@ def prepare_step(self, *args, **kwargs):
         """Prepare for installing bundle of Python packages."""
         super(Bundle, self).prepare_step(*args, **kwargs)
 
-        if get_software_root('Python') is None:
+        if get_software_root('Python') is None and self.toolchain.name != DUMMY_TOOLCHAIN_NAME:
```
boegel: @wpoely86 That's not going to work out well, since `PythonPackage` may depend on `get_software_root` not returning `None`.

When installing Python tools with the system Python, use this in the easyconfig file:

```python
allow_system_deps = [('Python', SYS_PYTHON_VERSION)]
```

cfr. https://github.com/easybuilders/easybuild-easyconfigs/blob/master/easybuild/easyconfigs/e/EasyBuild/EasyBuild-3.9.0.eb#L43

PR #1740, opened 2019-05-31
fix GROMACS easyblock for list-type configopts

ostueker:
GROMACS easyblock currently fails in `sanity_check_step` when used
with a list-type `configopts` config like:

```
configopts = [
    "-DGMX_THREAD_MPI=OFF -DGMX_DOUBLE=OFF",
    "-DGMX_THREAD_MPI=OFF -DGMX_DOUBLE=ON",
]
```

This patch fixes this issue by correctly iterating over multiple
configopts strings to determine which binary and lib names to
check for.

### DIFF
--- a/easybuild/easyblocks/g/gromacs.py
+++ b/easybuild/easyblocks/g/gromacs.py
@@ -490,7 +490,8 @@ class EB_GROMACS(CMakeMake):
             dirs.append(os.path.join(self.lib_subdir, 'pkgconfig'))
 
         custom_paths = {
-            'files': [os.path.join('bin', b) for b in bin_files] + [os.path.join(self.lib_subdir, l) for l in lib_files],
+            'files': [os.path.join('bin', b) for b in bin_files] + 
+                     [os.path.join(self.lib_subdir, l) for l in lib_files],
             'dirs': dirs,
         }
         super(EB_GROMACS, self).sanity_check_step(custom_paths=custom_paths)
--- a/easybuild/easyblocks/g/gromacs.py
+++ b/easybuild/easyblocks/g/gromacs.py
@@ -468,18 +468,29 @@ class EB_GROMACS(CMakeMake):
             libnames.extend([libname + mpisuff for libname in libnames])
 
         suff = ''
-        # add the _d suffix to the suffix, in case of the double precission
-        if re.search('DGMX_DOUBLE=(ON|YES|TRUE|Y|[1-9])', self.cfg['configopts'], re.I):
-            suff = '_d'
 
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/g/gromacs.py
```diff
@@ -468,18 +468,29 @@ def sanity_check_step(self):
...
 
         suff = ''
-        # add the _d suffix to the suffix, in case of the double precission
-        if re.search('DGMX_DOUBLE=(ON|YES|TRUE|Y|[1-9])', self.cfg['configopts'], re.I):
-            suff = '_d'
 
-        libs = ['lib%s%s.%s' % (libname, suff, self.libext) for libname in libnames]
+        # make sure that configopts is a list:
+        configopts = self.cfg['configopts']
+        if type(configopts) is str:
```
boegel: Please use `isinstance` here. 
Also, I would use a more explicit name for the list of `configopts` values:

```python
configopts_list = self.cfg['configopts']
if isinstance(configopts_list, basestring):
    configopts_list = [configopts_list]

File: easybuild/easyblocks/g/gromacs.py [hunk 2]
```diff
@@ -468,18 +468,29 @@ def sanity_check_step(self):
...
+            configopts = [configopts]
+
+        libs = []
+        binaries = []
+
+        for cfgopts in configopts:
```
boegel: I would avoid `cfgopts` vs `configopts` here, since that'll lead to confusion

```python
for configopts in configopts_list:
```

easybuild/easyblocks/g/gromacs.py
```diff
@@ -468,18 +468,29 @@ def sanity_check_step(self):
...
+            # add the _d suffix to the suffix, in case of the double precission
+            if re.search('DGMX_DOUBLE=(ON|YES|TRUE|Y|[1-9])', cfgopts, re.I):
```
boegel: `s/cfgopts/configopts/g`

easybuild/easyblocks/g/gromacs.py
```diff
@@ -468,18 +468,29 @@ def sanity_check_step(self):
...
+                suff = '_d'
+
+            libs += ['lib%s%s.%s' % (libname, suff, self.libext) for libname in libnames]
+            binaries += [b + suff for b in bins]
```
boegel: please use `.extend` rather than `+=` (since that avoid creating a new list on every iteration):

```python
libs.extend([...])
binaries.extend([...])
```

Maybe also rename `binaries` to `bin_files` (and `libs` to `lib_files` for consistency)?

ostueker:
Thank you, @boegel for all the shiny EasyBuild releases!

PR #1741, opened 2019-06-05
don't try to patch CROSSTOOL script for recent Bazel versions (since it's no longer there)

boegel:
The `CROSSTOOL` script is no longer there in Bazel 0.24.0 and newer.

It seems like patching something else is not required, since building Bazel 0.26.0 with this updated easyblocks works just fine (see https://github.com/easybuilders/easybuild-easyconfigs/pull/8439)


akesandgren:
Going in, thanks @boegel!

Issue #605, opened 2015-04-30
also run numpy.test rather than only a simply numpy.dot performance test in EB_numpy.test_step


boegel:
Note: this will result in a failed test with numpy built with Intel compilers, which is actually a broken test, see https://github.com/numpy/numpy/issues/3844#issuecomment-97910358

problem reported by @olavks

boegel:
@pescobar: up for tackling this at some point?

edmondac:
With this easyconfig, at least some numpy tests are run, but the failures aren't detected:

```
name = 'numpy'
version = '1.15.0'
versionsuffix = '-Python-%(pyver)s'

homepage = 'http://www.numpy.org'
description = """NumPy is the fundamental package for scientific computing with Python. It contains among other things:
 a powerful N-dimensional array object, sophisticated (broadcasting) functions, tools for integrating C/C++ and Fortran
 code, useful linear algebra, Fourier transform, and random number capabilities. Besides its obvious scientific uses,
 NumPy can also be used as an efficient multi-dimensional container of generic data. Arbitrary data-types can be
 defined. This allows NumPy to seamlessly and speedily integrate with a wide variety of databases."""

toolchain = {'name': 'foss', 'version': '2018b'}

source_urls = [PYPI_SOURCE]
sources = [SOURCE_ZIP]
checksums = [
    'f28e73cf18d37a413f7d5de35d024e6b98f14566a10d82100f9dc491a7d449f9',  # numpy-1.15.0.zip
]

dependencies = [
    ('Python', '3.6.6'),
    ('pytest', '4.4.0', versionsuffix),
]

moduleclass = 'math'
```

Snippet of output:
```
== 2019-06-06 14:44:13,059 run.py:506 INFO cmd "export PYTHONPATH=/scratch/edmondac-admin/eb-7fJRj3/tmp744GE1/lib/python3.6/site-packages:$PYTHONPATH &&cd .. && /rds/bear-apps/devel/2018b/edmondac-eb-6/EL7/EL7-power9/software/Python/3.6.
6-foss-2018b/bin/python -c 'import numpy; numpy.test(verbose=2)'" exited with exit code 0 and output:
NumPy version 1.15.0
NumPy relaxed strides checking option: True
================================================= test session starts =================================================
platform linux -- Python 3.6.6, pytest-4.4.0, py-1.8.0, pluggy-0.9.0
[...]

PR #1743, opened 2019-06-07
Changing numpy test command to fix the f2py errors

edmondac:
Ref https://github.com/easybuilders/easybuild-easyblocks/issues/605

This fixes the four fortran compile errors when we run the numpy test suite

### DIFF
--- a/easybuild/easyblocks/n/numpy.py
+++ b/easybuild/easyblocks/n/numpy.py
@@ -66,7 +66,7 @@ class EB_numpy(FortranPythonPackage):
         self.sitecfg = None
         self.sitecfgfn = 'site.cfg'
         self.testinstall = True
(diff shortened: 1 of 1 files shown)


edmondac:
Thanks to @ocaisa for the help :-)

Now we just need the easyblock to notice if the tests fail/error

edmondac:
https://github.com/easybuilders/easybuild-easyconfigs/pull/8458 is probably the better solution. If we merge that, then this PR isn't needed anymore...

edmondac:
See also https://github.com/easybuilders/easybuild-easyblocks/pull/1744

ocaisa:
In #1744 I've unset `LDFLAGS` during testing which also seems to work, this is similar to what has been elsewhere in the scipy/numpy easyblocks because of problems caused by having `LDFLAGS` set (and is easier than trying to figure out what should be included in the flag).

edmondac:
Closing in favour of #1744

PR #1744, opened 2019-06-07
Catch errors or failures in numpy/scipy testsuites

### LINKED ISSUE #605: also run numpy.test rather than only a simply numpy.dot performance test in EB_numpy.test_step

ocaisa:
Fixes #605


easybuild/easyblocks/n/numpy.py
```diff
@@ -48,6 +48,21 @@
 from distutils.version import LooseVersion
 
 
+def parse_numpy_test_suite_output(full_testsuite_output):
+    # Parse the last (non-empty) line of the testcmd_output which contains the summary from the test suite
+    testsuite_summary = [i for i in full_testsuite_output.split('\n') if i != ''][-1]
```
edmondac: I think you could use `testsuite_summary = full_testsuite_output.strip().splitlines()[-1]`

ocaisa: That works, thanks.

ocaisa:
This works but is now reporting the errors in the scipy testsuite that we have been ignoring until now.
@boegel From what I could find out the failing tests are something to do with MKL:
```
fftpack/tests/test_real_transforms.py::TestIDCTIVFloat::test_definition FAILED                                                                                        [  3%]
special/tests/test_data.py::test_boost[<Data for legendre_p_via_assoc_: legendre_p_large_ipp-legendre_p_large>] FAILED                                                [ 87%]
special/tests/test_data.py::test_boost[<Data for legendre_p_via_lpmn: legendre_p_large_ipp-legendre_p_large>] FAILED                                                  [ 87%]
``` 
What do we do in this case? The first of these is reported at https://github.com/scipy/scipy/issues/9799 and the others have popped up as well https://github.com/scipy/scipy/issues/6338 . Do we patch these tests out of the test suite?

ocaisa:
From https://software.intel.com/en-us/forums/intel-fortran-compiler-for-linux-and-mac-os-x/topic/593518 this looks like a compiler optimisation issue/bug

edmondac:
Seems to work - and catch the scipy failures:

```
== installing extension scipy 1.2.1 (2/5)...
  >> defining build environment for foss/2019a toolchain
  >> running command:
        [started at: 2019-06-11 16:21:33]
        [output logged in /scratch/edmondac-admin/eb-xRXKB7/easybuild-run_cmd-NiWpen.log]
        tar xzf /rds/bear-sysadmin/configmgmt/easybuild/sources/s/SciPy-bundle/extensions/scipy-1.2.1.tar.gz
  >> running command:
        [started at: 2019-06-11 16:21:35]
        [output logged in /scratch/edmondac-admin/eb-xRXKB7/easybuild-run_cmd-PCL_nk.log]
        unset LDFLAGS &&   /rds/bear-apps/devel/github/edmondac-eb-contrib/EL7/EL7-haswell/software/Python/3.7.2-GCCcore-8.2.0/bin/python setup.py build  --fcompiler=gnu95
  >> command completed: exit 0, ran in 00h05m55s
  >> running command:
        [started at: 2019-06-11 16:27:31]
        [output logged in /scratch/edmondac-admin/eb-xRXKB7/easybuild-run_cmd-gkN4pm.log]
        export PYTHONPATH=/scratch/edmondac-admin/eb-xRXKB7/tmp8ZAzcL/lib/python3.7/site-packages:$PYTHONPATH &&  unset LDFLAGS &&   pip install --prefix=/scratch/edmondac-admin/eb-xRXKB7/tmp8ZAzcL  --no-deps  --ignore-installed  --no-build-isolation  .
  >> command completed: exit 0, ran in 00h00m17s
  >> running command:
        [started at: 2019-06-11 16:27:48]
        [output logged in /scratch/edmondac-admin/eb-xRXKB7/easybuild-run_cmd-d7_8Ie.log]
        export PYTHONPATH=/scratch/edmondac-admin/eb-xRXKB7/tmp8ZAzcL/lib/python3.7/site-packages:$PYTHONPATH &&  cd .. && unset LDFLAGS && /rds/bear-apps/devel/github/edmondac-eb-contrib/EL7/EL7-haswell/software/Python/3.7.2-GCCcore-8.2.0/bin/python -c 'import numpy; import scipy; scipy.test(verbose=2)'
  >> command completed: exit 0, ran in 00h06m38s
== FAILED: Installation ended unsuccessfully (build directory: /dev/shm/build-edmondac-admin/edmondac-admin-contrib/SciPybundle/2019.03/foss-2019a): build failed (first 300 chars): Found errors or failures in scipy testsuite output:
 {'skipped': 1291, 'warnings': 316969, 'xfailed': 78, 'xpassed': 7, 'failed': 4, 'deselected': 1223, 'passed': 14097}
```

ocaisa:
@edmondac Can you grep the log for that build for `FAILED`  and post those failures here to see if they have any overlap with mine (I have 3 not 4)

ocaisa:
I should mention that I'm building a Python2 bundle which includes numpy and scipy using GCC and MKL

edmondac:
```$ grep -C 2 FAILED easybuild-SciPy-bundle-2019.03-20190611.161804.LnhpN.log
fftpack/tests/test_real_transforms.py::TestIDCTIIIInt::test_definition PASSED                                                                                                                             [  3%]
fftpack/tests/test_real_transforms.py::TestIDCTIVDouble::test_definition PASSED                                                                                                                           [  3%]
fftpack/tests/test_real_transforms.py::TestIDCTIVFloat::test_definition FAILED                                                                                                                            [  3%]
fftpack/tests/test_real_transforms.py::TestIDCTIVInt::test_definition PASSED                                                                                                                              [  3%]
fftpack/tests/test_real_transforms.py::TestDSTIDouble::test_definition PASSED                                                                                                                             [  3%]
--
special/tests/test_data.py::test_boost[<Data for assoc_legendre_p_boost_: assoc_legendre_p_ipp-assoc_legendre_p>] PASSED                                                                                  [ 87%]
special/tests/test_data.py::test_boost[<Data for legendre_p_via_assoc_: legendre_p_ipp-legendre_p>] PASSED                                                                                                [ 87%]
special/tests/test_data.py::test_boost[<Data for legendre_p_via_assoc_: legendre_p_large_ipp-legendre_p_large>] FAILED                                                                                    [ 87%]
special/tests/test_data.py::test_boost[<Data for legendre_p_via_lpmn: legendre_p_ipp-legendre_p>] PASSED                                                                                                  [ 87%]
special/tests/test_data.py::test_boost[<Data for legendre_p_via_lpmn: legendre_p_large_ipp-legendre_p_large>] FAILED                                                                                      [ 87%]
special/tests/test_data.py::test_boost[<Data for lpn_: legendre_p_ipp-legendre_p>] PASSED                                                                                                                 [ 87%]
special/tests/test_data.py::test_boost[<Data for lpn_: legendre_p_large_ipp-legendre_p_large>] PASSED                                                                                                     [ 87%]
--
special/tests/test_orthogonal.py::test_roots_sh_legendre PASSED                                                                                                                                           [ 90%]
special/tests/test_orthogonal.py::test_roots_laguerre PASSED                                                                                                                                              [ 90%]
special/tests/test_orthogonal.py::test_roots_genlaguerre FAILED                                                                                                                                           [ 90%]
special/tests/test_orthogonal.py::test_gh_6721 PASSED                                                                                                                                                     [ 90%]
special/tests/test_orthogonal_eval.py::test_eval_chebyt PASSED                                                                                                                                            [ 90%]
--
== 2019-06-11 16:34:28,388 run.py:559 INFO parse_log_for_error msg: Command used: export PYTHONPATH=/scratch/edmondac-admin/eb-xRXKB7/tmp8ZAzcL/lib/python3.7/site-packages:$PYTHONPATH &&  cd .. && unset LDFLAGS && /rds/bear-apps/devel/github/edmondac-eb-contrib/EL7/EL7-haswell/software/Python/3.7.2-GCCcore-8.2.0/bin/python -c 'import numpy; import scipy; scipy.test(verbose=2)' 
== 2019-06-11 16:34:28,389 run.py:561 INFO parse_log_for_error (some may be harmless) regExp (?<![(,-]|\w)(?:error|segmentation fault|failed)(?![(,-]|\.?\w) found:
fftpack/tests/test_real_transforms.py::TestIDCTIVFloat::test_definition FAILED                                                                                                                            [  3%]
special/tests/test_data.py::test_boost[<Data for legendre_p_via_assoc_: legendre_p_large_ipp-legendre_p_large>] FAILED                                                                                    [ 87%]
special/tests/test_data.py::test_boost[<Data for legendre_p_via_lpmn: legendre_p_large_ipp-legendre_p_large>] FAILED                                                                                      [ 87%]
special/tests/test_orthogonal.py::test_roots_genlaguerre FAILED                                                                                                                                           [ 90%]
            # difference is due to fftw using a better algorithm w.r.t error
>                   err_msg="Size %d failed" % i)
E           Size 1024 failed
== 2019-06-11 16:34:28,389 run.py:518 WARNING Found 7 errors in command output (output: fftpack/tests/test_real_transforms.py::TestIDCTIVFloat::test_definition FAILED                                                                                                                            [  3%], special/tests/test_data.py::test_boost[<Data for legendre_p_via_assoc_: legendre_p_large_ipp-legendre_p_large>] FAILED                                                                                    [ 87%], special/tests/test_data.py::test_boost[<Data for legendre_p_via_lpmn: legendre_p_large_ipp-legendre_p_large>] FAILED                                                                                      [ 87%], special/tests/test_orthogonal.py::test_roots_genlaguerre FAILED                                                                                                                                           [ 90%],             # difference is due to fftw using a better algorithm w.r.t error, >                   err_msg="Size %d failed" % i), E           Size 1024 failed)
```

edmondac:
@ocaisa that was for `SciPy-bundle-2019.03-foss-2019a.eb` on a haswell node

ocaisa:
@edmondac Ok, thanks, 3 of the failures are the same and then you have an additional one:
```
special/tests/test_orthogonal.py::test_roots_genlaguerre FAILED
```
It's weird that the errors are there for both MKL and OpenBLAS.

easybuild/easyblocks/generic/pythonpackage.py
```diff
@@ -526,11 +526,17 @@ def test_step(self):
             if self.testcmd:
                 testcmd = self.testcmd % {'python': self.python_cmd}
                 cmd = ' '.join([extrapath, self.cfg['pretestopts'], testcmd, self.cfg['testopts']])
-                run_cmd(cmd, log_all=True, simple=True)
+                if return_testcmd_output:
+                    testcmd_output, testcmd_exit_code = run_cmd(cmd, log_all=True, simple=False)
+                else:
+                    run_cmd(cmd, log_all=True, simple=True)
```
boegel: You can simplify this safely to just:

```python
testcmd_output, testcmd_exit_code = run_cmd(cmd, log_all=True, simple=False)
```

That way you'll always catch the test output/exit code, yet only actually return it if `return_testcmd_output` is `True`.

I think it would even be safe to simply *always* return.

Since nothing is returned now (which implies returning `None`), you can't really break something by making that change (unless weird things are being done).

easybuild/easyblocks/n/numpy.py
```diff
@@ -48,6 +48,28 @@
 from distutils.version import LooseVersion
 
 
+def parse_numpy_test_suite_output(full_testsuite_output):
+    # Grab the line of the testcmd_output which contains the summary from the test suite.
+    # Summary looks like:
+    # ========== 4860 passed, 14 skipped, 88 deselected, 7 xfailed, 4 error in 63.72 seconds ==========
+    testsuite_summary = [i for i in full_testsuite_output.split('\n') if
+                         i.startswith('=') and 'passed' in i and 'seconds' in i]
```
boegel: Using a regex would be better here...

Something like:

```python
summary_regex = re.compile('^=.*passed.*skipped.*seconds', re.M)
res = summary_regex.search(full_testsuite_output)
if res:
    testsuite_summary = res.group(0)
else:
    raise EasyBuildError("Failed to extract summary from test suite output using pattern '%s': %s",
                         summary_regex.pattern, full_testsuite_output)

File: easybuild/easyblocks/n/numpy.py [hunk 4]
```diff
@@ -48,6 +48,28 @@
...
+
+    # Verify we got a summary like expected (should be a list with a single element)
+    if not testsuite_summary:
+        raise EasyBuildError("Could not extract summary from test suite output, is it written over multiple lines?")
+
+    # Use regex to help parse it
+    regexp = r"([\w]+)\ ([\w]+)(?:\ in|,)"
```
boegel: Are you grabbing the `failed` and `error` values here? If so, this should be improved significantly to actually use `error` and `failed`...

We're only expecting one summary line, correct?

easybuild/easyblocks/n/numpy.py
```diff
@@ -66,7 +88,9 @@ def __init__(self, *args, **kwargs):
         self.sitecfg = None
         self.sitecfgfn = 'site.cfg'
         self.testinstall = True
-        self.testcmd = "cd .. && %(python)s -c 'import numpy; numpy.test(verbose=2)'"
+        # unset flags that can influence the testsuite
```
boegel: Hmm, can we add some more details here, for example *how* are tests failing if `LDFLAGS` is not unset? @edmondac?

easybuild/easyblocks/n/numpy.py
```diff
@@ -214,9 +238,20 @@ def get_libs_for_mkl(varname):
...
-        super(EB_numpy, self).test_step()
 
-        # temporarily install numpy, it doesn't alow to be used straight from the source dir
+        if self.cfg['runtest']:
+            # Let's handle the output from the numpy test suite ourselves
+            testcmd_output, testcmd_exit_code = super(EB_numpy, self).test_step(return_testcmd_output=True)
+
+            numpy_testsuite_summary = parse_numpy_test_suite_output(testcmd_output)
+
+            if 'failed' in numpy_testsuite_summary or 'error' in numpy_testsuite_summary:
```
boegel: Do we need to be careful here with `failed` vs `xfailed`?

easybuild/easyblocks/s/scipy.py
```diff
@@ -58,6 +62,19 @@ def configure_step(self):
...
                 self.cfg.update('preinstallopts', "unset LDFLAGS && ")
 
+    def test_step(self):
+        """Run available scipy unit tests"""
+        if self.cfg['runtest']:
+            # Let's handle the output from the scipy test suite ourselves
+            testcmd_output, testcmd_exit_code = super(EB_scipy, self).test_step(return_testcmd_output=True)
+            scipy_testsuite_summary = parse_numpy_test_suite_output(testcmd_output)
+
+            if 'failed' in scipy_testsuite_summary or 'error' in scipy_testsuite_summary:
```
boegel: Same here, `failed` vs `xfailed`?

easybuild/easyblocks/n/numpy.py
ocaisa: I'm grabbing everything in the summary line, the resulting dictionary looks like:
```
{'skipped': 1291, 'warnings': 316969, 'xfailed': 78, 'xpassed': 7, 'failed': 4, 'deselected': 1223, 'passed': 14097}
```

easybuild/easyblocks/n/numpy.py
ocaisa: See https://github.com/numpy/numpy/issues/2411

easybuild/easyblocks/n/numpy.py
ocaisa: `numpy_testsuite_summary` is a dict at this point, I'm searching for a specific key.

easybuild/easyblocks/s/scipy.py
ocaisa: `numpy_testsuite_summary` is a dict, not a string

easybuild/easyblocks/n/numpy.py
ocaisa: I knew it was logged somewhere: https://github.com/easybuilders/easybuild-easyblocks/blob/develop/easybuild/easyblocks/generic/fortranpythonpackage.py#L57

easybuild/easyblocks/n/numpy.py
ocaisa: > We're only expecting one summary line, correct?

I wasn't sure what happens when the line gets really long, I thought it might split it over two but no, it doesn't. In other words, yes it is just a single summary line.

boegel: OK, the regex was confusing me a bit, since you're using `\w` twice, both for the count and the 'label'.
Please use `\d` instead to grab the count (this makes the regexp a little bit stricter, but avoids issues with the `int(...)` used below).
Let's clarify the regular expression a bit too, since it's not immediately obvious:

```python
# regex to match all "<count> <label>" instances in test summary
regexp = r"([\d]+)\ ([\w]+)(?:\ in|,)"
```

easybuild/easyblocks/n/numpy.py
boegel: I guess my question is actually, what does `xfailed` mean? Expected failures?

easybuild/easyblocks/n/numpy.py
```diff
@@ -214,9 +241,23 @@ def get_libs_for_mkl(varname):
...
+                failure_text = "Found errors or failures in numpy testsuite output:\n %s" % numpy_testsuite_summary
+                if self.cfg['ignore_test_failures']:
+                    self.log.warning("Ignoring: ", failure_text)
```
boegel: @ocaisa This will just be a warning in the log file, which is perhaps too easy to miss?

How about making this an actual verbose warning, via `print_warning`?

easybuild/easyblocks/n/numpy.py
ocaisa: Hard to find but the answer is yes (from looking at https://github.com/scipy/scipy/commit/2c27ec48c249e51bdeeb9ba7a3d96db04cfbb6de)

boegel:
I spent some time in fixing the problem I was hitting where building a stand-alone old `numpy` version, which was failing with `Failed to extract summary from test suite output...` because the only summary line in the output for old `numpy` versions looks like this:

```
FAILED (KNOWNFAIL=6, SKIP=11, errors=1, failures=4)
```

I've made changes to be more forgiving and just log a warning when the test suite summary was not found, see https://github.com/ocaisa/easybuild-easyblocks/pull/18 .

There's more to be done however... Basically, this PR is trying to tackle a bunch of things at once:

i) catching failures in the numpy/scipy test suites (and optionally failing the installation if there are any)
ii) also running `numpy.test` when `numpy` is used as an extension (in `Python` or `SciPy-bundle` for example), i.e. fix #605 

Currently, it does i) only for recent numpy versions, and fails hard for old numpy versions because it can't figure out the failures, so that's a blocker.
That problem is resolved with https://github.com/ocaisa/easybuild-easyblocks/pull/18, but it flat out ignores not being able to determine the failures; we can do better?

Also, ii) isn't fixed at all as far as I can tell: `numpy.test` still isn't being run when `numpy` is used as an extension, only when it's being installed stand-alone through a dedicated `numpy` easyconfig.

I can't spare the time right now to figure out these issues before the EasyBuild v3.9.3 release, so we'll need to follow up on them later i'm afraid...

ocaisa:
>Also, ii) isn't fixed at all as far as I can tell: numpy.test still isn't being run when numpy is used as an extension, only when it's being installed stand-alone through a dedicated numpy easyconfig.

Not sure what you mean here since none of the previous behaviour has changed, all this PR does is actually catch the failures when the tests are run.  Basically this is just picking up on the default behaviour that is being set in `PythonPackage` (and that is the same as it was before, nothing has changed). The issue with bundles is that the default behaviour of `PythonPackage` is not used *unless* you are using `PythonBundle`, which is true for all recent python bundles. In the past the tests were simply never run.

If you want the tests to automatically run with legacy easyconfigs when numpy/scipy is included then this would require more work since you will need to fiddle with the default value `runtest` for such cases.

I don't see that as necessary for this PR though, all that is addressed here is how to handle the test suite results when the tests are run.

akesandgren:
Agree, the important thing is to catch errors moving forward.

easybuild/easyblocks/s/scipy.py
```diff
@@ -35,18 +35,31 @@
...
+from easybuild.tools.build_log import EasyBuildError, print_warning
 import easybuild.tools.toolchain as toolchain
 
 
 class EB_scipy(FortranPythonPackage):
     """Support for installing the scipy Python package as part of a Python installation."""
 
+    @staticmethod
+    def extra_options():
+        """Easyconfig parameters specific to numpy."""
```
akesandgren: Nah, this is scipy, not numpy.

easybuild/easyblocks/n/numpy.py
```diff
@@ -214,9 +253,22 @@ def get_libs_for_mkl(varname):
...
-        super(EB_numpy, self).test_step()
 
-        # temporarily install numpy, it doesn't alow to be used straight from the source dir
+        if self.cfg['runtest']:
+            # Let's handle the output from the numpy test suite ourselves
+            testcmd_output, testcmd_exit_code = super(EB_numpy, self).test_step(return_testcmd_output=True)
+
+            testsuite_summary = parse_numpy_test_suite_output(testcmd_output)
+
+            if testsuite_summary and ('failed' in testsuite_summary or 'error' in testsuite_summary):
```
akesandgren: Does this handle all types of return values from parse_numpy_test_suite_output?
res = summary_regex.search
res = None
res = dict

ocaisa: Yes, `summary_regex.search` only get's returned if it's 'truthiness' is false, `None` is also false, and the dict is explicitly handled.
I can make `parse_numpy_test_suite_output` explicitly return False for the first two cases though, just to very clear on the return possibilities.

ocaisa: Returning only a dict or `None` made more sense so I've implemented that

akesandgren (dismissed):

akesandgren:
@boegel are you satisfied with this? Your change request looks fixed.

edmondac:
I know this is for a past problem, but it'd be nice to have to catch future test failures on POWER9, for example. Any chance of it in the next release @boegel ?

ocaisa:
@boegel I think we are waiting on the nod from you to get this merged

akesandgren:
@ocaisa conflict resolution
@boegel Happy with this yet?

ocaisa:
@akesandgren conflict resolved

akesandgren:
So, leaving it to @boegel then since he had comments

Flamefire:
The numpy.test function returns a True/False based on success. Was there a reason why that wasn't used?

ocaisa:
@Flamefire I guess we probably didn't look at what it returned, only at what the exit values were (always 0).

Flamefire:
So what would be a good way to proceed here? Shall I make a new PR which does check the return value?

easybuild/easyblocks/n/numpy.py
```diff
@@ -66,7 +105,9 @@ def __init__(self, *args, **kwargs):
         self.sitecfg = None
         self.sitecfgfn = 'site.cfg'
         self.testinstall = True
-        self.testcmd = "cd .. && %(python)s -c 'import numpy; numpy.test(verbose=2)'"
+        # LDFLAGS should not be set when testing numpy/scipy, because it overwrites whatever numpy/scipy sets
+        # see http://projects.scipy.org/numpy/ticket/182
+        self.testcmd = "unset LDFLAGS && cd .. && %(python)s -c 'import numpy; numpy.test(verbose=2)'"
```
ocaisa, line 110: @Flamefire To convert the Boolean return value to an exit code, would the suggestion here be something like:
```
self.testcmd = "unset LDFLAGS && cd .. && %(python)s -c 'import numpy; import sys; sys.exit(!int(numpy.test(verbose=2))'"
```

Flamefire: Yes, that would likely work. To keep the changes and the work done here I'd suggest to extract that change (likely the only one required) to a new PR and test it with a couple numpys

ocaisa: Agree, please keep the `unset LDFLAGS` and associated comment (which would have solved the issue you ran into)

Flamefire: Done: #2238

migueldiascosta:
@ocaisa can we close this as superseded by #2238 and #2241?

boegel:
@migueldiascosta Agreed, let's close this and follow up in the other PRs/issues.

PR #1749, opened 2019-06-11
Workaround for XCrySDen for Tcl 8.6

Micket:
XCrySDen is old, and patching all it's broken Tcl code doesn't sound fun. Workaround for now.

I also made mesa optional, so that I can use this block myself (since we filter-deps Mesa)

### DIFF
--- a/easybuild/easyblocks/x/xcrysden.py
+++ b/easybuild/easyblocks/x/xcrysden.py
@@ -52,8 +52,7 @@ class EB_XCrySDen(ConfigureMake):
         and set make target and installation prefix.
         """
 
-        # check dependencies
-        deps = ["Mesa", "Tcl", "Tk"]
+        deps = ['Tcl', 'Tk']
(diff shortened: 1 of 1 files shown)


boegel:
@Micket Can you clarify what effect using `-DUSE_INTERP_RESULT` has exactly?

Micket:
The Tcl API have optional return strings (which XCrySDen uses all over the place, for error messages if i recall correctly), but this part of the API was deprecated in 8.6. `-DUSE_INTERP_RESULT` reactivates this feature, as it was prior to 8.6. Do you want me to comment the code?

boegel:
@Micket Yes, please clarify this using a comment above the line adding `-DUSE_INTERP_RESULT` to `$CFLAGS`.

smoors:
tested with https://github.com/easybuilders/easybuild-easyconfigs/pull/8477

PR #1750, opened 2019-06-12
don't assume cmake arguments when configure_cmd is set in CMakeMake easyblock

migueldiascosta:
even if `configure_cmd` is simply a script running `cmake`, it may not accept the flags in `options_string` and the `src_dir` as arguments, and as it is they cannot be removed - the proposed solution gives back control to the easyconfig, imho

### DIFF
--- a/easybuild/easyblocks/generic/cmakemake.py
+++ b/easybuild/easyblocks/generic/cmakemake.py
@@ -134,12 +134,18 @@ class CMakeMake(ConfigureMake):
 
         options_string = ' '.join(options)
 
-        command = ' '.join([
-            self.cfg['preconfigopts'],
-            self.cfg.get('configure_cmd') or DEFAULT_CONFIGURE_CMD,
-            options_string,
-            self.cfg['configopts'],
-            srcdir])
+        if self.cfg.get('configure_cmd'):
+            command = ' '.join([
+                    self.cfg['preconfigopts'],
+                    self.cfg.get('configure_cmd'),
+                    self.cfg['configopts']])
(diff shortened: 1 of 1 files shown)


migueldiascosta:
closing and reopening to trigger travis

easybuild/easyblocks/generic/cmakemake.py
```diff
@@ -134,12 +134,18 @@ def configure_step(self, srcdir=None, builddir=None):
 
         options_string = ' '.join(options)
 
-        command = ' '.join([
-            self.cfg['preconfigopts'],
-            self.cfg.get('configure_cmd') or DEFAULT_CONFIGURE_CMD,
-            options_string,
-            self.cfg['configopts'],
-            srcdir])
+        if self.cfg.get('configure_cmd') != DEFAULT_CONFIGURE_CMD:
```
boegel: @migueldiascosta Do you mind changing this to positive logic, i.e.:

```python
if self.cfg.get('configure_cmd') == DEFAULT_CONFIGURE_CMD:
    ...
else:
    ...
```

migueldiascosta:
> @migueldiascosta What use case did you hit that triggered you to make this change?

https://github.com/easybuilders/easybuild-easyconfigs/pull/8478

easybuild/easyblocks/generic/cmakemake.py
migueldiascosta: will do

boegel:
Thanks for the fix @migueldiascosta!

PR #1754, opened 2019-06-14
also add --without-go configure option for SWIG as we ''disable everything by default"

jose-d:
..as we here

> disable everything by default

we should disable go too.

tested on SWIG `SWIG-3.0.12-foss-2018b-Python-2.7.15.eb` and `SWIG-4.0.0-foss-2019a-Python-2.7.15`.


boegel:
Tested with several `SWIG` easyconfigs, incl. an ancient one for SWIG 2.0.4, works like a charm, thanks @jose-d!

PR #1758, opened 2019-06-19
update WIEN2k easyblock to support version 19.1 (REVIEW)

migueldiascosta:
easyconfigs PR coming soon

### DIFF
--- a/easybuild/easyblocks/w/wien2k.py
+++ b/easybuild/easyblocks/w/wien2k.py
@@ -46,7 +46,7 @@ import easybuild.tools.toolchain as toolchain
 from easybuild.framework.easyblock import EasyBlock
 from easybuild.framework.easyconfig import CUSTOM
 from easybuild.tools.build_log import EasyBuildError
-from easybuild.tools.filetools import extract_file, mkdir, read_file, rmtree2, write_file
+from easybuild.tools.filetools import apply_regex_substitutions, extract_file, mkdir, read_file, rmtree2, write_file
 from easybuild.tools.modules import get_software_root, get_software_version
 from easybuild.tools.run import run_cmd, run_cmd_qa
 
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/w/wien2k.py
```diff
@@ -261,11 +277,29 @@ def configure_step(self):
...
+                        'variant': 'elpa_openmp' if self.toolchain.get_flag('openmp') else 'elpa'}
+
+                    # find correct line number
+                    elpa_line = '0'
+                    elpa_dir = "%(root)s/include/%(variant)s-%(version)s" % elpa_dict
+                    elpa_find_dirs_cmd = 'find %s -mindepth 1 -type d -name "elpa*"' % elparoot
+                    (elpa_dirs, _) = run_cmd(elpa_find_dirs_cmd, log_all=True, simple=False)
+                    for num, line in enumerate(elpa_dirs.split('\n'), 1):
+                        if line == elpa_dir:
+                            elpa_line = str(num)
```
migueldiascosta: here I reproduce how the `siteconfig` script looks for elpa versions and variants in order to select the requested line number, but maybe there is a way to look at the script output instead of reproducing it?

boegel: @migueldiascosta Can you give an example of the output?

`run_cmd_qa` supports using a part of the question (e.g. a number specifying which option to use) as the answer, which is what we do need to do for WRF for example, that may come in useful here?
See https://github.com/easybuilders/easybuild-easyblocks/blob/master/easybuild/easyblocks/w/wrf.py#L199

migueldiascosta: the output (below) is not numbered...

so we need to look at the whole list in any case, but if we could do that on the actual output instead of reproducing it, it would be better

```
Checking for ELPA version ...
 More than one set of include files in your ELPA-ROOT directory. Pick one (enter line number) or enter 0 to manually specify a version!:
.../software/ELPA/2018.11.001-intel-2019a/share/doc/elpa
.../software/ELPA/2018.11.001-intel-2019a/include/elpa-2018.11.001
.../software/ELPA/2018.11.001-intel-2019a/include/elpa-2018.11.001/elpa
.../software/ELPA/2018.11.001-intel-2019a/include/elpa_openmp-2018.11.001
.../software/ELPA/2018.11.001-intel-2019a/include/elpa_openmp-2018.11.001/elpa
```

Of course, initially I tried replying `0` and providing the version manually, the problem is that it then asks "Is this correct? (Y,n):" *before* allowing to change and the answer to this question is already yes in the `qanda` list, and I wasn't able to differentiate this particular question (maybe `std_qa` can also help there, if it allows matching multiple lines?)

boegel: I think you can match with multiple lines using `qa` too.

The main difference with `std_qa` is that it supports using regular expression patterns for matching.

Why is it a problem if the `Is this correct` question is already answered with `yes` though?

migueldiascosta: w.r.t. to matching multiple lines, then I must have been doing something wrong, will try again

w.r.t. to `Is this correct`, it asks *before* we choose, so we need to answer `no` in order to choose the correct one...

but I'll have a better look and come back to this

Micket:
A complete oversight from me, I forgot about the block itself when approving, https://github.com/easybuilders/easybuild-easyconfigs/pull/8503

Sorry for this.

ocaisa:
@Micket Merging this on the basis of the tests you did for the easyconfig

migueldiascosta:
@Micket github should really support PR dependencies (meanwhile, if you were using `--merge-pr`, this might have helped: https://github.com/easybuilders/easybuild-framework/pull/2607)

PR #1759, opened 2019-06-24
SNPhylo - Fix scripts installation path

vanzod:
The current easyblock installs the scripts provided by `SNPhylo/20160204` in `$EBROOTSNPHYLO/scripts` while the `snphylo.sh` main executable in `SNPhylo/20160204` looks for them in `$EBROOTSNPHYLO/bin/scripts` and fails with the following error:
```
Start to remove low quality data.
/opt/easybuild/software/MPI/GCC/6.4.0-2.28/OpenMPI/2.1.1/Python/3.6.3/bin/python: can't open file '/opt/easybuild/software/MPI/GCC/6.4.0-2.28/OpenMPI/2.1.1/SNPhylo/20160204-Python-3.6.3-R-3.4.3/bin/scripts/remove_no_genotype_data.py': [Errno 2] No such file or directory
```

### DIFF
--- a/easybuild/easyblocks/s/snphylo.py
+++ b/easybuild/easyblocks/s/snphylo.py
@@ -44,6 +44,12 @@ from easybuild.tools.run import run_cmd
 class EB_SNPhylo(EasyBlock):
     """Support for building and installing SNPhylo."""
 
+    def __init__(self, *args, **kwargs):
+        """Set custom class variables."""
+        super(EB_SNPhylo, self).__init__(*args, **kwargs)
+
+        self.scripts_path = os.path.join('bin', 'scripts')
+
     def configure_step(self):
         """No configure step for SNPhylo."""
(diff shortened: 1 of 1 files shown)


boegel:
@vanzod Can we enhance the sanity check to catch this?

vanzod:
@boegel Ok, so I tested with the older `SNPhylo/20140701` and that one also requires the scripts in `$EBROOTSNPHYLO/bin/scripts`. This simplifies the issue since both versions should install the scripts in the same path. The last code refactoring now works for both versions and the sanity checks have also been updated.

boegel:
@vanzod Is there any way to come up with a sanity check command using `snphylo.sh` that could help avoid re-introducing this issue?

vanzod:
@boegel No, in order to do that you need a dataset and start an actual run.

PR #1760, opened 2019-06-25
QuantumEspresso: Parsing of additional compilation parameters

Danzelot:
Adds the ability to specify compilation parameters in the easyconfig file which so far had to be changed laboriously in the source files.
These compilation parameters are specified in the QuantumEspresso [user manual in section 2.5](https://www.quantum-espresso.org/Doc/user_guide/node12.html)

To test it you have to specify one or multiple of the parameters in the easyconfig. Is there a way to write a test case?

PS: This is my first trail in modifying a easyblock but I think it might be useful for other users.

### DIFF
--- a/easybuild/easyblocks/q/quantumespresso.py
+++ b/easybuild/easyblocks/q/quantumespresso.py
@@ -54,6 +54,7 @@ class EB_QuantumESPRESSO(ConfigureMake):
             'hybrid': [False, "Enable hybrid build (with OpenMP)", CUSTOM],
             'with_scalapack': [True, "Enable ScaLAPACK support", CUSTOM],
             'with_ace': [False, "Enable Adaptively Compressed Exchange support", CUSTOM],
+            'ntypx': [10, "Max number of different types of atom", CUSTOM],
         }
         return ConfigureMake.extra_options(extra_vars)
 
@@ -232,6 +233,11 @@ class EB_QuantumESPRESSO(ConfigureMake):
             make_ext = '.inc'
         else:
             make_ext = '.sys'
+
+        if self.cfg['ntypx']:
+          regex_subs = [(r"^ntypx\s*=\s*\d+", r"ntypx = %s" % ntypx)]
+           apply_regex_substitutions(os.path.join('Modules', 'parameters.f90'), regex_subs)
+           print('REGEX: {}'.format(regex_subs))
 
         # patch make.sys file
         fn = os.path.join(self.cfg['start_dir'], 'make' + make_ext)
--- a/easybuild/easyblocks/q/quantumespresso.py
+++ b/easybuild/easyblocks/q/quantumespresso.py
@@ -238,21 +238,19 @@ class EB_QuantumESPRESSO(ConfigureMake):
         else:
             make_ext = '.sys'
 
+        # patch Modules/parameter.f90 file to set compilation parameters
+        regex_para = []
         if self.cfg['ntypx']:
-            regex_subs = [(r"ntypx\s*=\s*\d+", r"ntypx = %s" % self.cfg['ntypx'])]
-            apply_regex_substitutions(os.path.join('Modules', 'parameters.f90'), regex_subs)
+            regex_para.append((r"ntypx\s*=\s*\S+", r"ntypx = %s" % self.cfg['ntypx']))
         if self.cfg['nsx']:
-            regex_subs = [(r"nsx\s*=\s*\w+", r"nsx = %s" % self.cfg['nsx'])]
-            apply_regex_substitutions(os.path.join('Modules', 'parameters.f90'), regex_subs)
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/q/quantumespresso.py
```diff
@@ -233,6 +238,22 @@ def configure_step(self):
         else:
             make_ext = '.sys'
 
+        if self.cfg['ntypx']:
+          regex_subs = [(r"ntypx\s*=\s*\d+", r"ntypx = %s" % self.cfg['ntypx'])]
```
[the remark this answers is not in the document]
boegel: @Danzelot Do you mind fixing these style issues reported by the bot? Indentation should be done in multiples of 4 spaces.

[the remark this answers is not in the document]
Danzelot: I fixed it.
Don't know where this originates from, probably copy-paste.

Danzelot:
Thank you for the suggestions.
I fixed both issues.

Danzelot:
I noticed that I get a compilation error when I try to change *nsx* to 30. [q-e_building.log](https://github.com/easybuilders/easybuild-easyblocks/files/3330560/q-e_building.log)
I will have a look if I can fix that.

easybuild/easyblocks/q/quantumespresso.py
```diff
@@ -233,6 +238,20 @@ def configure_step(self):
...
+        if self.cfg['ntypx']:
+            regex_para.append((r"ntypx\s*=\s*\S+", r"ntypx = %s," % self.cfg['ntypx']))
+        if self.cfg['nsx']:
+            regex_para.append((r"nsx\s*=\s*\S+", r"nsx = %s," % self.cfg['nsx']))
+        if self.cfg['npk']:
+            regex_para.append((r"npk\s*=\s*\S+", r"npk = %s," % self.cfg['npk']))
+        if self.cfg['lmaxx']:
+            regex_para.append((r"lmaxx\s*=\s*\S+", r"lmaxx = %s," % self.cfg['lmaxx']))
+        if self.cfg['lqmax']:
+            regex_para.append((r"lqmax\s*=\s*\S+", r"lqmax = %s" % self.cfg['lqmax']))
```
boegel: @Danzelot Looks like this can be cleaned up even more, it's basically a for loop. And you can simplify the regex too using `.*`:

```python
for key in ['lmaxx', 'lqmax', 'npk', 'nsx', 'ntypx']:
    if self.cfg[key]:
        regex_para.append((key + r'\s*=.*', key + ' = ' + self.cfg[key]))

File: easybuild/easyblocks/q/quantumespresso.py [hunk 3]
```diff
@@ -54,6 +54,11 @@ def extra_options():
             'hybrid': [False, "Enable hybrid build (with OpenMP)", CUSTOM],
             'with_scalapack': [True, "Enable ScaLAPACK support", CUSTOM],
             'with_ace': [False, "Enable Adaptively Compressed Exchange support", CUSTOM],
+            'ntypx': [False, "Max number of different types of atom", CUSTOM],
+            'nsx': [False, "max number of atomic species (CP)", CUSTOM],
+            'npk': [False, "max number of k-points", CUSTOM],
+            'lmaxx': [False, "max non local angular momentum (l=0 to lmaxx)", CUSTOM],
+            'lqmax': [False, "max number of angular momenta of Q", CUSTOM],
```
boegel: @Danzelot Please keep these lines sorted alphabetically (mixed with the ones that are already there)

Danzelot: Done. Good thing to watch for.

easybuild/easyblocks/q/quantumespresso.py
Danzelot: Nice suggestion.
I had to change the regex a little to not mess up the fortran90 parameter file.
Should all work now. ( At least on my system)

easybuild/easyblocks/q/quantumespresso.py
```diff
@@ -233,6 +238,13 @@ def configure_step(self):
         else:
             make_ext = '.sys'
 
+        # patch Modules/parameter.f90 file to set compilation parameters
+        regex_para = []
+        for key in ['lmaxx', 'lqmax', 'npk', 'nsx', 'ntypx']:
```
akesandgren: Since we have to have ntypx >= nsx it might be worth the effort to check that it is indeed the case.
I.e., start by picking up both from parameters.f90 then verify the requirement using cfg[key].
And you need to pick up the values from parameters in case someone sets only one of them in the easyconfig file.

akesandgren:
@Danzelot Any progress on this?

ocaisa:
Since we're moving to a whole new CMake-based easyblock for QE in #3257, this PR is being closed

PR #1763, opened 2019-07-02
Healpix.py: modextrapaths implemented on easyblock level, sanity_check_step, gcc_target added

jose-d:
as discussed in [#8491](https://github.com/easybuilders/easybuild-easyconfigs/pull/8491) and #1752 :

* `LIBRARY_PATH`, `PATH`, `CPATH` extra prepends moved from easyconfig to this easyblock
* `sanity_check_step` returned back, just `libpsht_healpix_f.a` removed

and additionally:
* support for `gcc_target` option implemented

### DIFF
--- a/easybuild/easyblocks/h/healpix.py
+++ b/easybuild/easyblocks/h/healpix.py
@@ -37,17 +37,27 @@ from easybuild.tools.build_log import EasyBuildError
 from easybuild.tools.modules import get_software_root
 from easybuild.tools.run import run_cmd, run_cmd_qa
 from easybuild.tools.systemtools import get_shared_lib_ext
-from distutils.version import LooseVersion
+from easybuild.framework.easyconfig import CUSTOM
 
 
 class EB_HEALPix(ConfigureMake):
     """Support for building/installing HEALPix."""
+
+    @staticmethod
+    def extra_options():
+        """There 3 variants of GCC build"""
+        extra_vars = {
+            'gcc_target': ['generic_gcc', "Use generic_gcc target", CUSTOM],
+        }
+
+        return ConfigureMake.extra_options(extra_vars)
 
     def __init__(self, *args, **kwargs):
         """Initialisation of custom class variables for HEALPix."""
         super(EB_HEALPix, self).__init__(*args, **kwargs)
 
         self.build_in_installdir = True
+        self.target_string = None
 
     def extract_step(self):
         """Extract sources."""
@@ -62,11 +72,32 @@ class EB_HEALPix(ConfigureMake):
         if not cfitsio:
             raise EasyBuildError("Failed to determine root for CFITSIO, module not loaded?")
 
+        # target:
+        #   1: basic_gcc
+        #   2: generic_gcc
+        #   3: linux_icc
+        #   4: optimized_gcc
+        #   5: osx
+        #   6: osx_icc
+
         self.comp_fam = self.toolchain.comp_family()
         if self.comp_fam == toolchain.INTELCOMP:  # @UndefinedVariable
-            cxx_config = '4'  # linux_icc
+            cxx_config = '3'  # linux_icc
+            self.target_string = 'linux_icc'
         elif self.comp_fam == toolchain.GCC:  # @UndefinedVariable
-            cxx_config = '2'  # generic_gcc
+            if self.cfg['gcc_target'] == 'basic_gcc':
+                cxx_config = '1'
+                self.target_string = 'basic_gcc'
+            elif self.cfg['gcc_target'] == 'generic_gcc':
+                cxx_config = '2'
+                self.target_string = 'generic_gcc'
+            elif self.cfg['gcc_target'] == 'optimized_gcc':
+                cxx_config = '4'
+                self.target_string = 'optimized_gcc'
+            else:
+                # by default let's go with generic_gcc:
+                cxx_config = '2'
+                self.target_string = 'generic_gcc'
         else:
             raise EasyBuildError("Don't know how which C++ configuration for the used toolchain.")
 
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/h/healpix.py
```diff
@@ -112,20 +143,27 @@ def install_step(self):
...
-                'dirs': [os.path.join('include')],
-            }
-            super(EB_HEALPix, self).sanity_check_step(custom_paths=custom_paths)
+        """sanity checks"""
+        custom_paths = {
+            'files': [os.path.join('bin', x) for x in ['alteralm', 'anafast', 'hotspot', 'map2gif', 'median_filter',
+                                                       'plmgen', 'sky_ng_sim', 'sky_ng_sim_bin', 'smoothing',
+                                                       'synfast', 'ud_grade']] +
+                     [os.path.join('lib', 'lib%s.a' % x) for x in ['chealpix', 'gif', 'healpix', 'hpxgif']] +
+                     [os.path.join('lib', 'libchealpix.%s' % get_shared_lib_ext())],
```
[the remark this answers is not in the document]
jose-d: that should be fine. Sane format as in previous version, checked with Pycharm IDE autoformat tool.

easybuild/easyblocks/h/healpix.py
```diff
@@ -112,20 +143,27 @@ def install_step(self):
...
-                         [os.path.join('lib', 'libchealpix.%s' % get_shared_lib_ext())],
-                'dirs': [os.path.join('include')],
-            }
-            super(EB_HEALPix, self).sanity_check_step(custom_paths=custom_paths)
+        """sanity checks"""
+        custom_paths = {
+            'files': [os.path.join('bin', x) for x in ['alteralm', 'anafast', 'hotspot', 'map2gif', 'median_filter',
+                                                       'plmgen', 'sky_ng_sim', 'sky_ng_sim_bin', 'smoothing',
+                                                       'synfast', 'ud_grade']] +
+                     [os.path.join('lib', 'lib%s.a' % x) for x in ['chealpix', 'gif', 'healpix', 'hpxgif']] +
```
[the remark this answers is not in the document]
jose-d: that should be fine. Sane format as in previous version, checked with Pycharm IDE auto-format tool.

easybuild/easyblocks/h/healpix.py
```diff
@@ -37,17 +37,27 @@
...
 
 
 class EB_HEALPix(ConfigureMake):
     """Support for building/installing HEALPix."""
 
+    @staticmethod
+    def extra_options():
+        """There 3 variants of GCC build"""
+        extra_vars = {
+            'gcc_target': ['generic_gcc', "Use generic_gcc target", CUSTOM],
```
boegel: The description here could be improved a bit to clarify what it represents

easybuild/easyblocks/h/healpix.py
```diff
@@ -37,17 +37,27 @@
 from easybuild.tools.modules import get_software_root
 from easybuild.tools.run import run_cmd, run_cmd_qa
 from easybuild.tools.systemtools import get_shared_lib_ext
-from distutils.version import LooseVersion
+from easybuild.framework.easyconfig import CUSTOM
```
boegel: Please keep this sorted alphabetically.

easybuild/easyblocks/h/healpix.py
```diff
@@ -62,11 +72,32 @@ def configure_step(self):
...
+        #   2: generic_gcc
+        #   3: linux_icc
+        #   4: optimized_gcc
+        #   5: osx
+        #   6: osx_icc
+
         self.comp_fam = self.toolchain.comp_family()
         if self.comp_fam == toolchain.INTELCOMP:  # @UndefinedVariable
-            cxx_config = '4'  # linux_icc
+            cxx_config = '3'  # linux_icc
```
boegel: the changing number here is a sign of trouble (changing number for the same target across versions)

I've change things in https://github.com/jose-d/easybuild-easyblocks/pull/2 such that you don't need to hardcode the numbers anymore, you can extract it from the question (like we do in the WRF easyblock).

Works for both `HEALPix-2.20a-ictce-5.5.0.eb` and `HEALPix-3.50-foss-2018b.eb`

easybuild/easyblocks/h/healpix.py
```diff
@@ -62,11 +72,32 @@ def configure_step(self):
...
+                cxx_config = '1'
+                self.target_string = 'basic_gcc'
+            elif self.cfg['gcc_target'] == 'generic_gcc':
+                cxx_config = '2'
+                self.target_string = 'generic_gcc'
+            elif self.cfg['gcc_target'] == 'optimized_gcc':
+                cxx_config = '4'
+                self.target_string = 'optimized_gcc'
+            else:
+                # by default let's go with generic_gcc:
```
boegel: This should be an error imho, to protect against undesired behavior in case a silly typo is made...

easybuild/easyblocks/h/healpix.py
```diff
@@ -112,20 +143,27 @@ def install_step(self):
...
+            'files': [os.path.join('bin', x) for x in ['alteralm', 'anafast', 'hotspot', 'map2gif', 'median_filter',
+                                                       'plmgen', 'sky_ng_sim', 'sky_ng_sim_bin', 'smoothing',
+                                                       'synfast', 'ud_grade']] +
+                     [os.path.join('lib', 'lib%s.a' % x) for x in ['chealpix', 'gif', 'healpix', 'hpxgif']] +
+                     [os.path.join('lib', 'libchealpix.%s' % get_shared_lib_ext())],
+            'dirs': [
+                os.path.join(self.installdir, 'include'),
+                os.path.join(self.installdir, 'src/cxx', self.target_string, 'bin'),
+                os.path.join(self.installdir, 'src/cxx', self.target_string, 'lib'),
+                os.path.join(self.installdir, 'src/cxx', self.target_string, 'include'),
```
boegel: the `self.installdir` here is not actually needed (paths are checked relative to the installation directory)

easybuild/easyblocks/h/healpix.py
jose-d: TIL: There in Pycharm, it is not enough to click on 

> code → reformat

but one has to click on 

> Code → Optimize Imports

to be fully PEP8 compliant.

easybuild/easyblocks/h/healpix.py
jose-d: good job.

jose-d:
> @jose-d Requested changes are already made in [jose-d#2](https://github.com/jose-d/easybuild-easyblocks/pull/2)
I'm glad that you found the time to review my non-perfect contribution. Next time it will be better I hope.

easybuild/easyblocks/h/healpix.py
```diff
@@ -112,20 +135,42 @@ def install_step(self):
...
+        custom_paths = {
+            'files': binaries + libraries + [os.path.join('lib', 'libchealpix.%s' % get_shared_lib_ext())],
+            'dirs': ['include'] + [os.path.join('src', 'cxx', self.target_string, x) for x in target_subdirs],
+        }
+
+        custom_commands = []
+        if LooseVersion(self.version) >= LooseVersion('3'):
+            custom_commands.extend([
+                "cd %s && make test &> build/testresults.txt" % self.installdir,
+                '[ "$(cat %s/build/testresults.txt | grep -c \'test completed\')" -eq "4" ]' % self.installdir,
```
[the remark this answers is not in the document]
boegel: @jose-d Using the simpler notation makes sense.

We can deal with the numbers when they change?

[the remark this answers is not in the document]
jose-d: roger that. 
I'll change the notation and we can improve the test logic when new version arrives.

easybuild/easyblocks/h/healpix.py
```diff
@@ -112,20 +135,42 @@ def install_step(self):
...
+            'files': binaries + libraries + [os.path.join('lib', 'libchealpix.%s' % get_shared_lib_ext())],
+            'dirs': ['include'] + [os.path.join('src', 'cxx', self.target_string, x) for x in target_subdirs],
+        }
+
+        custom_commands = []
+        if LooseVersion(self.version) >= LooseVersion('3'):
+            custom_commands.extend([
+                "cd %s && make test &> build/testresults.txt" % self.installdir,
+                'test $(grep -c "test completed" %s/build/testresults.txt) -eq 4' % self.installdir,
+                'test $(grep -c "normal completion" %s/build/testresults.txt) -eq 10' % self.installdir,
```
boegel: @jose-d How about loosening this up a little bit by using `-gt` rather than `-eq`?

If more tests pass or complete normally, there's no problem?

jose-d: my opinion: when there is change in test suite "test completed" and "normal completion" counts, we can expect major change of testsuite. So far I have seen non-conservative changes in whole Healpix project across versions, testsuite is returning '0' RC even when failing, and minor issues are not changing the final summary. 

I'd propose to keep this as it is, and possibly in next version:
a) patch the testsuite to behave like in mainstream projects = nonzero return code when fail
b) move magic constants "4" and "10" into easyconfig if we'll find them heavy-version specific
c) something better

easybuild/easyblocks/h/healpix.py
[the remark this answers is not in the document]
jose-d: bashisms improved in https://github.com/easybuilders/easybuild-easyblocks/pull/1763/commits/1fb62fa28cad58fcc97c945f8acc2ef6d5542fcb

easybuild/easyblocks/h/healpix.py
boegel: Thanks for clarifying @jose-d!

boegel:
Going in, thanks a lot for your efforts on this @jose-d!

PR #1764, opened 2019-07-03
Using dict-type checksums for CUDA, with custom source filename

edmondac:
CUDA needs a different source file per architecture


edmondac:
See also https://github.com/easybuilders/easybuild-framework/pull/2940 and https://github.com/easybuilders/easybuild-easyconfigs/pull/8136

boegel:
@edmondac So we can't use `%(arch)s` after all?

boegel:
@edmondac Due to the mixup in #1765, this appears merged, but the changes were actually reverted in #1765 too.

So, you'll need to open a new PR to follow-up on this...

edmondac:
I've replaced this with #1766

PR #1765, opened 2019-07-04
update OpenBLAS easyblock to be aware of POWER9 support in OpenBLAS 0.3.6

edmondac:
OpenBLAS 0.3.6 has explicit support for POWER9 so we don't need to tell it to use POWER8 any more.

Also, 0.3.5 has quite a few bugs on POWER9, so including a warning message.


boegel:
Thanks for the update @edmondac!

PR #1770, opened 2019-07-08
Add easyblock for MSM

vanzod:
Add easyblock for `MSM`

### DIFF
--- a/easybuild/easyblocks/m/msm.py
+++ b/easybuild/easyblocks/m/msm.py
@@ -38,7 +38,7 @@ from easybuild.tools.modules import get_software_root
 
 class EB_MSM(MakeCp):
     """Support for building and installing MSM."""
-    
+
     def __init__(self, *args, **kwargs):
         """Initialize MSM specific variables and purge previous installations."""
         super(EB_MSM, self).__init__(*args, **kwargs)
@@ -76,11 +76,11 @@ class EB_MSM(MakeCp):
         (out, _) = run_cmd(cmd, log_all=True, simple=False)
         fslmachtype = out.strip()
         self.log.debug("FSL machine type: %s" % fslmachtype)
-        
+
         env.setvar('FSLDEVDIR', self.installdir)
         env.setvar('FSLCONFDIR', os.path.join(fsl_root, 'fsl', 'config'))
         env.setvar('FSLMACHTYPE', fslmachtype)
-        
+
     def build_step(self):
         """Build MSM one component at a time."""
 
@@ -91,7 +91,7 @@ class EB_MSM(MakeCp):
             paracmd = "-j %s" % self.cfg['parallel']
 
         cmd = ' '.join(['make', paracmd, 'install'])
-        
+
         for comp in components:
             target_dir = os.path.join(self.sources_root, 'src', comp)
             self.log.debug("Building %s in directory %s" % (comp, target_dir))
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/m/msm.py
```diff
@@ -0,0 +1,108 @@
...
+
+    def __init__(self, *args, **kwargs):
+        """Initialize MSM specific variables and purge previous installations."""
+        super(EB_MSM, self).__init__(*args, **kwargs)
+        self.sources_root = os.path.join(self.builddir, 'MSM_HOCR-%s' % self.version)
+
+        # Ensure that nothing has been left over from previous installation attempts.
+        # This is necessary here since directories must be created before building
+        # and not removed before the installation step.
+        remove_dir(self.installdir)
```
boegel: @vanzod This is troublesome under `--module-only`, should be moved to `configure_step`?

easybuild/easyblocks/m/msm.py
```diff
@@ -0,0 +1,108 @@
...
+        """Create directories, copy required files and set env vars."""
+
+        # Create directories recursively
+        dirpath = self.installdir
+        for d in ['extras', 'include']:
+            try:
+                dirpath = os.path.join(dirpath, d)
+                mkdir(dirpath)
+                self.log.debug("Created directory: %s" % dirpath)
+            except OSError as err:
```
boegel: No need for `try`/`except` when using `mkdir` from `filetools`, you'll never hit an `OSError` when using that (`mkdir` will raise an `EasyBuildError` already when creating the directory failed)

easybuild/easyblocks/m/msm.py
```diff
@@ -0,0 +1,108 @@
...
+        # This is necessary here since directories must be created before building
+        # and not removed before the installation step.
+        remove_dir(self.installdir)
+
+    def configure_step(self):
+        """Create directories, copy required files and set env vars."""
+
+        # Create directories recursively
+        dirpath = self.installdir
+        for d in ['extras', 'include']:
```
boegel: Please don't use single-letter variables outside of list comprehension, use `dirname`?

easybuild/easyblocks/m/msm.py
```diff
@@ -0,0 +1,108 @@
...
+
+        cmd = ' '.join(['make', paracmd, 'install'])
+
+        for comp in components:
+            target_dir = os.path.join(self.sources_root, 'src', comp)
+            self.log.debug("Building %s in directory %s" % (comp, target_dir))
+            try:
+                change_dir(target_dir)
+                run_cmd(cmd, log_all=True, simple=True, log_output=True)
+            except OSError as err:
```
boegel: Same here, you'll never hit an `OSError` here, so might as well lose the `try`/`except`

easybuild/easyblocks/m/msm.py
```diff
@@ -0,0 +1,108 @@
...
+
+        paracmd = ''
+        if self.cfg['parallel']:
+            paracmd = "-j %s" % self.cfg['parallel']
+
+        cmd = ' '.join(['make', paracmd, 'install'])
+
+        for comp in components:
+            target_dir = os.path.join(self.sources_root, 'src', comp)
+            self.log.debug("Building %s in directory %s" % (comp, target_dir))
```
boegel: Cleaner:

```python
self.log.debug("Building %s in directory %s", comp, target_dir)
```

(no need for `%` in `self.log.debug` & co)

PR #1771, opened 2019-07-11
only check for qtwebengine if glibc is sufficiently recent

migueldiascosta:
allows installing recent Qt5 easyconfigs on CentOS 6 (e.g. https://github.com/easybuilders/easybuild-easyconfigs/pull/7249#issuecomment-446068152)


ocaisa:
Thanks @migueldiascosta

PR #1772, opened 2019-07-16
openfoam v1906 reorganized their wmake rules and now rely on General/…

mboisson:
…*/{c,c++} for compiler parameters

### DIFF
--- a/easybuild/easyblocks/o/openfoam.py
+++ b/easybuild/easyblocks/o/openfoam.py
@@ -69,7 +69,7 @@ class EB_OpenFOAM(EasyBlock):
 
         # version may start with 'v' for some variants of OpenFOAM
         # we need to strip this off to avoid problems when comparing LooseVersion instances in Python 3
-        self.looseversion = LooseVersion(self.version.lstrip('v'))
+        self.looseversion = LooseVersion(self.version.strip('v+'))
 
         if 'extend' in self.name.lower():
             if self.looseversion >= LooseVersion('3.0'):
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/o/openfoam.py
```diff
@@ -181,6 +181,8 @@ def configure_step(self):
 
         # inject compiler variables into wmake/rules files
         ldirs = glob.glob(os.path.join(self.builddir, self.openfoamdir, 'wmake', 'rules', 'linux*'))
+        if LooseVersion(self.version) >= LooseVersion('v1906'):
```
boegel: This comparison is only meaningful if the version is also of the form `v[0-9]`, so make sure that's the case?

```python
if self.version.startswith('v') and LooseVersion(self.version) >= LooseVersion('v1906'):
```

easybuild/easyblocks/o/openfoam.py
```diff
@@ -210,6 +212,9 @@ def configure_step(self):
             'c++OPT': os.environ['CXXFLAGS'],
         }
         for wmake_rules_file in wmake_rules_files:
+            # the cOpt and c++Opt files don't exist in the General directories
```
boegel: should be reworded?

```python
# the cOpt and c++Opt files don't exist in the General directories (which are included for recent versions)

File: easybuild/easyblocks/o/openfoam.py (hunk 1 above)
Comment (mboisson) [reply easybuild/easyblocks/o/openfoam.py]: done

File: easybuild/easyblocks/o/openfoam.py (hunk 2 above)
Comment (mboisson) [reply easybuild/easyblocks/o/openfoam.py]: done

File: easybuild/easyblocks/o/openfoam.py [hunk 3]
```diff
@@ -181,6 +181,8 @@ def configure_step(self):
 
         # inject compiler variables into wmake/rules files
         ldirs = glob.glob(os.path.join(self.builddir, self.openfoamdir, 'wmake', 'rules', 'linux*'))
+        if self.version.startswith('v') and LooseVersion(self.version) >= LooseVersion('v1906'):
```
boegel: @mboisson please sync with current `develop` and use `self.looseversion > LooseVersion('1906'):`, see changes in #1794

mboisson: Done

akesandgren:
@boegel are you satisified with the fix to your request? Then please handle it.

boegel:
@mboisson Blocked by conflicts with current `develop`...

mboisson:
@boegel it should be merged now.

akesandgren:
Going in, thanks @mboisson!

PR #1773, opened 2019-07-16
removed checks for Boussinesq and sonic solvers for OpenFOAM 7, since those have been deprecated

mboisson:
In the OpenFOAM 7 release notes : 

Solvers: The Boussinesq equation of state can now be applied to any buoyancy solver, deprecating the specialized buoyantBoussinesq[SP]impleFoam solvers [ commit 9bf346 ].

Solvers:  The sonicFoam, sonicDyMFoam and sonicLiquidFoam are deprecated since their functionality has been merged into the rhoPimpleFoam solver, running the transonic option [ commit 3341f9 ].  Added LTS capability to pimpleFoam, providing the option to accelerate solutions to steady-state [ commit e1e3e2 ]

### DIFF
--- a/easybuild/easyblocks/o/openfoam.py
+++ b/easybuild/easyblocks/o/openfoam.py
@@ -361,6 +361,11 @@ class EB_OpenFOAM(EasyBlock):
                [os.path.join(toolsdir, "surface%s" % x) for x in ["Add", "Find", "Smooth"]] + \
                [os.path.join(toolsdir, x) for x in ['blockMesh', 'checkMesh', 'deformedGeom', 'engineSwirl',
                                                     'modifyMesh', 'refineMesh']]
+
+        # remove Boussinesq and sonic since for OpenFOAM >= 7, since those solvers are deprecated
+        if LooseVersion(self.version) >= LooseVersion('7'):
+            bins = [x for x in bins if "Boussinesq" not in x and "sonic" not in x]
+
         # check for the Pstream and scotchDecomp libraries, there must be a dummy one and an mpi one
         if 'extend' in self.name.lower():
             libs = [os.path.join(libsdir, "libscotchDecomp.%s" % shlib_ext),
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/o/openfoam.py
```diff
@@ -361,6 +361,11 @@ def sanity_check_step(self):
                [os.path.join(toolsdir, "surface%s" % x) for x in ["Add", "Find", "Smooth"]] + \
                [os.path.join(toolsdir, x) for x in ['blockMesh', 'checkMesh', 'deformedGeom', 'engineSwirl',
                                                     'modifyMesh', 'refineMesh']]
+
+        # remove Boussinesq and sonic since for OpenFOAM >= 7, since those solvers are deprecated
+        if LooseVersion(self.version) >= LooseVersion('7'):
+            bins = [x for x in bins if "Boussinesq" not in x and "sonic" not in x]
```
boegel: @mboisson Rather than adding them first and then removing then again, we should only add them for versions < `7`, something like:

```python
bins = [...
       [os.path.join(toolsdir, "buoyantSimpleFoam")] + \
       [os.path.join(toolsdir, "%sFoam" % x) for x in ["boundary", "engine"]] + \
       ...
]

# only include Boussinesq and sonic since for OpenFOAM < 7, since those solvers have been deprecated
if LooseVersion(self.version) < LooseVersion('7'):
    bins.extend([
        os.path.join(toolsdir, "buoyantBoussinesqSimpleFoam"),
        os.path.join(toolsdir, "sonicFoam")
    ])
```

mboisson: done.

boegel:
closing and re-opening PR to re-trigger enhanced Travis tests (but should be a formality for this PR...)

boegel:
Tested and approved, going in, thanks @mboisson!

PR #1774, opened 2019-07-18
[parmetis] Make compatible with custom configopts

lexming:
Currently the easyblock for ParMETIS fails the sanity check with `configopts = '-DSHARED=1'`, which builds its shared libraries but disables the static ones. This patch adds a conditional check on configopts to the sanity check function.

This patch also enforces that symlinks are removed before creation to avoid `file already exists` errors  with builds with multiple iterations.

### DIFF
--- a/easybuild/easyblocks/p/parmetis.py
+++ b/easybuild/easyblocks/p/parmetis.py
@@ -41,7 +41,6 @@ from easybuild.tools.filetools import mkdir, symlink, remove_file
 from easybuild.tools.run import run_cmd
 from easybuild.tools.systemtools import get_shared_lib_ext
 
-SHARED = 'SHARED'
 
 class EB_ParMETIS(EasyBlock):
     """Support for building and installing ParMETIS."""
@@ -187,7 +186,7 @@ class EB_ParMETIS(EasyBlock):
         config_list = self.cfg['configopts'] if isinstance(self.cfg['configopts'], list) else [self.cfg['configopts']]
         for configopt in config_list:
             for opt in configopt.split():
-                if SHARED in opt and any(trueval in opt for trueval in config_true):
+                if 'SHARED' in opt and any(trueval in opt for trueval in config_true):
                     config_shared = True
 
         if config_shared:
--- a/easybuild/easyblocks/p/parmetis.py
+++ b/easybuild/easyblocks/p/parmetis.py
@@ -193,7 +193,7 @@ class EB_ParMETIS(EasyBlock):
 
         custom_paths = {
                         'files': ['include/%smetis.h' % x for x in ["", "par"]] +
-                                 parmetis_libs,
+                                  parmetis_libs,
                         'dirs': ['Lib']
                        }
 
--- a/easybuild/easyblocks/p/parmetis.py
+++ b/easybuild/easyblocks/p/parmetis.py
@@ -30,6 +30,7 @@ EasyBuild support for ParMETIS, implemented as an easyblock
 @author: Kenneth Hoste (Ghent University)
 @author: Pieter De Baets (Ghent University)
 @author: Jens Timmerman (Ghent University)
+@author: Alex Domingo (Vrije Universiteit Brussel)
 """
 import os
 import shutil
--- a/easybuild/easyblocks/p/parmetis.py
+++ b/easybuild/easyblocks/p/parmetis.py
@@ -54,6 +54,19 @@ class EB_ParMETIS(EasyBlock):
         Run 'cmake' in the build dir to get rid of a 'user friendly'
         help message that is displayed without this step.
         """
+
+        # Detect if this iteration is building static or shared libs to do proper sanity check
+        static_build = True
+        config_true = ['1', 'ON', 'YES', 'TRUE', 'Y']  # True values in CMake
+        for configopt in self.cfg['configopts'].split():
+            if 'SHARED' in configopt and any(trueval in configopt for trueval in config_true):
+                static_build = False
+
+        if static_build:
+            self.config_static = True
+        else:
+            self.config_shared = True
+
         if LooseVersion(self.version) >= LooseVersion("4"):
             # tested with 4.0.2, now actually requires cmake to be run first
             # for both parmetis and metis
@@ -181,20 +194,13 @@ class EB_ParMETIS(EasyBlock):
     def sanity_check_step(self):
         """Custom sanity check for ParMETIS."""
 
-        # Paths to libraries defined depending on SHARED configuration option
-        config_shared = False
-        config_true = ['1', 'ON', 'YES', 'TRUE', 'Y']  # True values in CMake
-        config_list = self.cfg['configopts'] if isinstance(self.cfg['configopts'], list) else [self.cfg['configopts']]
-        for configopt in config_list:
-            for opt in configopt.split():
-                if 'SHARED' in opt and any(trueval in opt for trueval in config_true):
-                    config_shared = True
-
-        if config_shared:
+        parmetis_libs = [os.path.join('lib', 'libmetis.a')]
+        # Add static and shared libs depending on configopts
+        if hasattr(self, "config_shared") and self.config_shared is True:
             shlib_ext = get_shared_lib_ext()
-            parmetis_libs = ['lib/libparmetis.%s' % shlib_ext, 'lib/libmetis.a']
-        else:
-            parmetis_libs = ['lib/lib%smetis.a' % x for x in ["", "par"]]
+            parmetis_libs.append(os.path.join('lib', 'libparmetis.%s' % shlib_ext))
+        if hasattr(self, "config_static") and self.config_static is True:
+            parmetis_libs.append(os.path.join('lib', 'libparmetis.a'))
 
         custom_paths = {
             'files': ['include/%smetis.h' % x for x in ["", "par"]] + parmetis_libs,
--- a/easybuild/easyblocks/p/parmetis.py
+++ b/easybuild/easyblocks/p/parmetis.py
@@ -192,13 +192,13 @@ class EB_ParMETIS(EasyBlock):
 
         if config_shared:
             shlib_ext = get_shared_lib_ext()
-            parmetis_libs = ['lib/libparmetis.%s' % shlib_ext] + ['lib/libmetis.a']
+            parmetis_libs = ['lib/libparmetis.%s' % shlib_ext, 'lib/libmetis.a']
         else:
             parmetis_libs = ['lib/lib%smetis.a' % x for x in ["", "par"]]
 
         custom_paths = {
-                        'files': ['include/%smetis.h' % x for x in ["", "par"]] + parmetis_libs,
-                        'dirs': ['Lib']
-                       }
+            'files': ['include/%smetis.h' % x for x in ["", "par"]] + parmetis_libs,
+            'dirs': ['Lib']
+        }
 
         super(EB_ParMETIS, self).sanity_check_step(custom_paths=custom_paths)
--- a/easybuild/easyblocks/p/parmetis.py
+++ b/easybuild/easyblocks/p/parmetis.py
@@ -168,13 +168,11 @@ class EB_ParMETIS(EasyBlock):
         # and header files in the Lib directory (capital L). The following symlink are hence created.
         try:
             caplibdir = os.path.join(self.installdir, 'Lib')
-            if os.path.lexists(caplibdir):
-                remove_file(caplibdir)
+            remove_file(caplibdir)
             symlink(libdir, caplibdir)
             for header_file in ['metis.h', 'parmetis.h']:
                 header_path = os.path.join(libdir, header_file)
-                if os.path.lexists(header_path):
-                    remove_file(header_path)
+                remove_file(header_path)
                 symlink(os.path.join(includedir, header_file), header_path)
         except OSError as err:
             raise EasyBuildError("Something went wrong during symlink creation: %s", err)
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/p/parmetis.py
```diff
@@ -167,19 +168,31 @@ def install_step(self):
...
+            shlib_ext = get_shared_lib_ext()
+            custom_paths = {
+                        'files': ['include/%smetis.h' % x for x in ["", "par"]] +
+                                 ['lib/libparmetis.%s' % shlib_ext],
+                        'dirs': ['Lib']
+                       }
+        else:
+            custom_paths = {
                         'files': ['include/%smetis.h' % x for x in ["", "par"]] +
                                  ['lib/lib%smetis.a' % x for x in ["", "par"]],
```
boegel: @lexming Since only the library extension is different, we can do better...

In addition, you're only checking for `libparmetis.so`, not `libmetis.so`?

Something like this:

```python
if '-DSHARED=1' in self.cfg['configopts']:
    lib_ext = get_shared_lib_ext()
else:
    lib_ext = 'a'

custom_paths = {
    'files': ['include/%smetis.h' % x for x in ['', 'par']] +
             ['lib/lib%smetis.%s' % (x, lib_ext) for x in ['', 'par']],
}
```

This is not perfect though, since there are probably other ways to enable building shared libraries (`-DSHARED=ON` for example)...

easybuild/easyblocks/p/parmetis.py
```diff
@@ -167,19 +168,31 @@ def install_step(self):
         # and header files in the Lib directory (capital L). The following symlink are hence created.
         try:
             llibdir = os.path.join(self.installdir, 'Lib')
+            if os.path.lexists(llibdir):
+                os.remove(llibdir)
```
boegel: please use the `remove_file` and `symlink` functions from `easybuild.tools.filetools`

easybuild/easyblocks/p/parmetis.py
```diff
@@ -167,19 +168,31 @@ def install_step(self):
...
             os.symlink(libdir, llibdir)
             for f in ['metis.h', 'parmetis.h']:
+                if os.path.lexists(os.path.join(libdir, f)):
+                    os.remove(os.path.join(libdir, f))
```
boegel: please use the `remove_file` and `symlink` functions from `easybuild.tools.filetools`, and change the code to not use a single-letter variable here:

```python
for header_file in ['metis.h', 'parmetis.h']:
    if os.path.lexists(os.path.join(libdir, header_file)):
        remove_file(os.path.join(libdir, header_file))
        symlink(os.path.join(includedir, header_file), os.path.join(libdir, header_file))

**Comment by akesandgren** [comment]
@lexming ping!

**Comment by lexming** [comment]
@akesandgren forgot about this one, will come back to it asap.

File: easybuild/easyblocks/p/parmetis.py (hunk 1 above)
Comment (lexming) [reply easybuild/easyblocks/p/parmetis.py]: Unfortunately ParMETIS builds `libparmetis.so` and `libmetis.a` with `SHARED=1`. Nonetheless, I simplified the code setting the library paths in `custom_paths` and also added a check on `SHARED` that considers all possible `true` strings in CMake.

File: easybuild/easyblocks/p/parmetis.py (hunk 2 above)
Comment (lexming) [reply easybuild/easyblocks/p/parmetis.py]: Fixed :+1:

File: easybuild/easyblocks/p/parmetis.py (hunk 3 above)
Comment (lexming) [reply easybuild/easyblocks/p/parmetis.py]: Fixed :+1:

File: easybuild/easyblocks/p/parmetis.py [hunk 4]
```diff
@@ -166,20 +167,33 @@ def install_step(self):
...
+            shlib_ext = get_shared_lib_ext()
+            parmetis_libs = ['lib/libparmetis.%s' % shlib_ext] + ['lib/libmetis.a']
+        else:
+            parmetis_libs = ['lib/lib%smetis.a' % x for x in ["", "par"]]
+
         custom_paths = {
                         'files': ['include/%smetis.h' % x for x in ["", "par"]] +
-                                 ['lib/lib%smetis.a' % x for x in ["", "par"]],
-                        'dirs':['Lib']
+                                 parmetis_libs,
```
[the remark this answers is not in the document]
lexming: can anybody clarify why is this line misaligned?

lexming:
@boegel Thanks for the feedback and sorry for the delay. Updated with your suggestions.

easybuild/easyblocks/p/parmetis.py
```diff
@@ -182,9 +182,9 @@ def install_step(self):
     def sanity_check_step(self):
         """Custom sanity check for ParMETIS."""
 
-        # Paths to libraries depend on SHARED configuration option
-        config_trues = ['1', 'ON', 'YES', 'TRUE', 'Y']  # strings that evaluate as True in CMake
-        if set(['-DSHARED=%s' % opt for opt in config_trues]) & set(self.cfg['configopts']):
+        # Paths to libraries defined depending on SHARED configuration option
+        config_shared = ['-DSHARED=%s' % val for val in ['1', 'ON', 'YES', 'TRUE', 'Y']]  # True values in CMake
```
akesandgren: Lower case values are also allowed if i remember correctly? And "True", "Yes", "On", so a lower/upper case compare is likely better.

akesandgren:
@boegel happy with the changes for your reviews?

easybuild/easyblocks/p/parmetis.py
lexming: You're right. I made the comparison case insensitive :+1:

akesandgren:
@boegel ping

lexming:
@boegel all checks are ok now and I addressed all comments. Let me know if there is anything else that needs to be done for this PR.

easybuild/easyblocks/p/parmetis.py
```diff
@@ -166,20 +167,33 @@ def install_step(self):
         # other applications depending on ParMETIS (SuiteSparse for one) look for both ParMETIS libraries
         # and header files in the Lib directory (capital L). The following symlink are hence created.
         try:
-            llibdir = os.path.join(self.installdir, 'Lib')
-            os.symlink(libdir, llibdir)
-            for f in ['metis.h', 'parmetis.h']:
-                os.symlink(os.path.join(includedir, f), os.path.join(libdir, f))
+            caplibdir = os.path.join(self.installdir, 'Lib')
+            if os.path.lexists(caplibdir):
```
smoors: @lexming remove_file() already does this check, so it's not needed.

easybuild/easyblocks/p/parmetis.py
```diff
@@ -166,20 +167,33 @@ def install_step(self):
...
+                remove_file(caplibdir)
+            symlink(libdir, caplibdir)
+            for header_file in ['metis.h', 'parmetis.h']:
+                header_path = os.path.join(libdir, header_file)
+                if os.path.lexists(header_path):
```
smoors: same here.

easybuild/easyblocks/p/parmetis.py
```diff
@@ -166,20 +167,33 @@ def install_step(self):
...
+                    remove_file(header_path)
+                symlink(os.path.join(includedir, header_file), header_path)
         except OSError as err:
             raise EasyBuildError("Something went wrong during symlink creation: %s", err)
 
     def sanity_check_step(self):
         """Custom sanity check for ParMETIS."""
 
+        # Paths to libraries defined depending on SHARED configuration option
+        config_shared = ['-DSHARED=%s' % val for val in ['1', 'ON', 'YES', 'TRUE', 'Y']]  # True values in CMake
```
smoors: @lexming there are many more valid ways to specify this cmake option, eg. `-DSHARED='1'`, `-D SHARED=1`, `-DSHARED:BOOL=ON`, and so on.

to make this more robust, it's probably better to create a `sharedlibs` parameter, like is done for example in the ATLAS easyblock.

easybuild/easyblocks/p/parmetis.py
```diff
@@ -166,20 +167,33 @@ def install_step(self):
...
+        config_optlist = [opt.upper() for configopt in self.cfg['configopts'] for opt in configopt.split()]
+        if set(config_shared) & set(config_optlist):
+            shlib_ext = get_shared_lib_ext()
+            parmetis_libs = ['lib/libparmetis.%s' % shlib_ext] + ['lib/libmetis.a']
```
smoors: nit: merge this into one list:
```
parmetis_libs = ['lib/libparmetis.%s' % shlib_ext, 'lib/libmetis.a']
```

easybuild/easyblocks/p/parmetis.py
```diff
@@ -37,9 +37,11 @@
 
 from easybuild.framework.easyblock import EasyBlock
 from easybuild.tools.build_log import EasyBuildError
-from easybuild.tools.filetools import mkdir
+from easybuild.tools.filetools import mkdir, symlink, remove_file
 from easybuild.tools.run import run_cmd
+from easybuild.tools.systemtools import get_shared_lib_ext
 
+SHARED = 'SHARED'
```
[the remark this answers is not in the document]
lexming: fixed

easybuild/easyblocks/p/parmetis.py
lexming: Thanks for pointing this out, fixed :+1:

easybuild/easyblocks/p/parmetis.py
lexming: thx!

easybuild/easyblocks/p/parmetis.py
lexming: absolutely, merged :+1:

easybuild/easyblocks/p/parmetis.py
lexming: I thought about using `sharedibs` as an extra option, but it is not trivial for `parmetis` as it is necessary to run multiple iterations to have an installation with both static and shared libs. So, to still allow for such builds, I improved the detection of the `SHARED` option in `configopts` to be more flexible and work for both lists and strings.

lexming:
Tested in `develop` with
- no `configopts`
- configopts = '-DSHARED=1' 
- configopts = ['', '-DSHARED=0', '-DSHARED:BOOL=ON', "-DSHARED='1'"]

smoors:
test report for this easyblock: https://github.com/easybuilders/easybuild-easyconfigs/pull/9886

easybuild/easyblocks/p/parmetis.py
```diff
@@ -166,20 +168,37 @@ def install_step(self):
...
+        # Paths to libraries defined depending on SHARED configuration option
+        config_shared = False
+        config_true = ['1', 'ON', 'YES', 'TRUE', 'Y']  # True values in CMake
+        config_list = self.cfg['configopts'] if isinstance(self.cfg['configopts'], list) else [self.cfg['configopts']]
+        for configopt in config_list:
+            for opt in configopt.split():
+                if 'SHARED' in opt and any(trueval in opt for trueval in config_true):
+                    config_shared = True
+
+        if config_shared:
```
boegel: @lexming There's one small issue here...

Even with an iterated installation (when using a list of `configopts`), the sanity check is only run once.

So, depending on whether `SHARED=1` is used first or last, you'll only check for `libparmetis.a` or `libparmetis.so`...

It would be better if you move the code that checks whether `SHARED=1` (or equivalent) was used to the `configure_step`, and set `self.config_shared` there, also introduce a `self.config_static` (and set it to `True` if `SHARED=1` was not found), both initialize them to `False` in the easyblock constructor, and then change the code here to:

```python
parmetis_libs = [os.path.join('lib', 'libmetis.a')]
if self.config_shared:
    parmetis_libs.append(os.path.join('lib', 'libparmetis.%s' % shlib_ext))
if self.config_static:
    parmetis_libs.append(os.path.join('lib', 'libparmetis.a'))

File: easybuild/easyblocks/p/parmetis.py [hunk 12]
```diff
@@ -48,10 +50,23 @@ def configure_step(self):
...
-        
-        Run 'cmake' in the build dir to get rid of a 'user friendly' 
+
+        Run 'cmake' in the build dir to get rid of a 'user friendly'
         help message that is displayed without this step.
         """
+
+        # Detect if this iteration is building static or shared libs to do proper sanity check
+        static_build = True
+        config_true = ['1', 'ON', 'YES', 'TRUE', 'Y']  # True values in CMake
```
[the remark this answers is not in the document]
lexming: What changes? I checked with `black -l 120 -S --diff parmetis.py` and it would not change anything in that section.

easybuild/easyblocks/p/parmetis.py
lexming: That's very true. Moving the detection of `SHARED` to the config step also simplifies quite a bit the code, as the `configopts` will always be a string inside `config_step()`.
Fixed :+1:

easybuild/easyblocks/p/parmetis.py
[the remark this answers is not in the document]
boegel: OK to ignore this, fixed with #1950

easybuild/easyblocks/p/parmetis.py
```diff
@@ -166,20 +181,30 @@ def install_step(self):
...
+                symlink(os.path.join(includedir, header_file), header_path)
         except OSError as err:
             raise EasyBuildError("Something went wrong during symlink creation: %s", err)
 
     def sanity_check_step(self):
         """Custom sanity check for ParMETIS."""
 
+        parmetis_libs = [os.path.join('lib', 'libmetis.a')]
+        # Add static and shared libs depending on configopts
+        if hasattr(self, "config_shared") and self.config_shared is True:
```
boegel: @lexming no need for `hasattr` as long as you initialize `self.config_shared` in the constructor, done in https://github.com/lexming/easybuild-easyblocks/pull/1

PR #1776, opened 2019-08-09
MSM - Moved file copy and sanity check to easyblock

vanzod:
This allows a cleaner easyconfig file by moving the definition of the files to copy and the sanity check to the easyblock.

### DIFF
--- a/easybuild/easyblocks/m/msm.py
+++ b/easybuild/easyblocks/m/msm.py
@@ -120,12 +120,12 @@ class EB_MSM(MakeCp):
 
         self.cfg['files_to_copy'] = [(['src/MSM/%s' % f for f in self.binfiles], 'bin')]
         super(EB_MSM, self).install_step()
-    
+
     def sanity_check_step(self):
         """Custom sanity check for MSM."""
 
         custom_paths = {
-            'files':['bin/%s' % f for f in self.binfiles],
+            'files': ['bin/%s' % f for f in self.binfiles],
             'dirs': ['doc', 'include', 'lib']
         }
         super(EB_MSM, self).sanity_check_step(custom_paths=custom_paths)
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/m/msm.py
```diff
@@ -98,6 +110,22 @@ def build_step(self):
...
         """Override installdir creation"""
+
         self.log.warning("Not removing installation directory %s" % self.installdir)
         self.cfg['keeppreviousinstall'] = True
         super(EB_MSM, self).make_installdir()
+
+    def install_step(self):
+        """Define files to be copied at installation."""
+
+        self.cfg['files_to_copy'] = [(['src/MSM/%s' % f for f in self.binfiles], 'bin')]
```
boegel: please use `os.path.join`:

```python
self.cfg['files_to_copy'] = [([os.path.join('src', 'MSM', f) for f in self.binfiles], 'bin')]
```

Also, aren't `doc` & co missing? How will those get installed?

easybuild/easyblocks/m/msm.py
```diff
@@ -98,6 +110,22 @@ def build_step(self):
...
+        super(EB_MSM, self).install_step()
+
+    def sanity_check_step(self):
+        """Custom sanity check for MSM."""
+
+        custom_paths = {
+            'files': ['bin/%s' % f for f in self.binfiles],
```
boegel: here too:

```python
[os.path.join('bin', f) for f in self.binfiles]

easybuild/easyblocks/m/msm.py
vanzod: @boegel `doc` are installed automatically by `make install`. As a matter of fact the `doc` directory is already checked in `sanity_check_step`.

boegel: Wait, `make install` installs the docs, but not the binaries?! 🤦‍♂

vanzod: Yup

PR #1781, opened 2019-08-22
enable 'check_ldshared' in generic PythonPackage easyblock by default

boegel:
Enabling this by default makes sense, if problems arise (haven't seen any so far), we can still disable this if the need would arise...


Micket:
So, do we not want to make this dependent on toolchain version?
For reference:
https://github.com/easybuilders/easybuild-easyblocks/pull/1455#issuecomment-457529348

Alternatively, we could go through old easyconfigs where it's currently not set and set `check_ldconfig=False`?

boegel:
replaced with #1788

PR #1787, opened 2019-08-26
also define $ZLIB_HOME when installing Hadoop with native bindings for zlib

boegel:
potential fix for issue raised by @vanzod in https://github.com/easybuilders/easybuild-easyconfigs/pull/7485

### DIFF
--- a/easybuild/easyblocks/h/hadoop.py
+++ b/easybuild/easyblocks/h/hadoop.py
@@ -32,6 +32,7 @@ import os
 import re
 import shutil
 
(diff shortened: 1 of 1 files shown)


vanzod:
@boegel No, this does not work. I suspect that such variable should be defined in CMake as `-DZLIB_HOME=...`. However since the `cmake` command is generated by Maven, I have no clue how this can be done if Maven does not generate it automatically upon passing `-Dzlib.prefix=`

PR #1788, opened 2019-08-27
enable 'check_ldshared' in generic PythonPackage easyblock by default

boegel:
replacement for #1781, now to correct branch (`develop`)

### DIFF
--- a/easybuild/easyblocks/generic/pythonpackage.py
+++ b/easybuild/easyblocks/generic/pythonpackage.py
@@ -181,7 +181,7 @@ class PythonPackage(ExtensionEasyBlock):
             extra_vars = {}
         extra_vars.update({
             'buildcmd': ['build', "Command to pass to setup.py to build the extension", CUSTOM],
-            'check_ldshared': [False, 'Check Python value of $LDSHARED, correct if needed to "$CC -shared"', CUSTOM],
+            'check_ldshared': [True, 'Check Python value of $LDSHARED, correct if needed to "$CC -shared"', CUSTOM],
             'download_dep_fail': [None, "Fail if downloaded dependencies are detected", CUSTOM],
             'install_target': ['install', "Option to pass to setup.py", CUSTOM],
             'pip_ignore_installed': [True, "Let pip ignore installed Python packages (i.e. don't remove them)", CUSTOM],
--- a/easybuild/easyblocks/generic/pythonpackage.py
+++ b/easybuild/easyblocks/generic/pythonpackage.py
@@ -61,6 +61,13 @@ SETUP_PY_DEVELOP_CMD = "%(python)s setup.py develop --prefix=%(prefix)s %(instal
 UNKNOWN = 'UNKNOWN'
 
 
+def det_python_version(python_cmd):
+    """Determine version of specified 'python' command."""
+    pycode = 'import sys; print("%s.%s.%s" % sys.version_info[:3])'
+    out, _ = run_cmd("%s -c '%s'" % (python_cmd, pycode), simple=False, trace=False)
+    return out.strip()
+
+
 def pick_python_cmd(req_maj_ver=None, req_min_ver=None):
     """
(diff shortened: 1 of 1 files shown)


Micket:
I'll just copy my previous comment:
So, do we not want to make this dependent on toolchain version?
For reference:
#1455 (comment)

Alternatively, we could go through old easyconfigs where it's currently not set and set check_ldconfig=False?
We could just find all PythonPackage/PythonBundle configs for intel/iccifort/iimpi/iimpic/intelcuda that currently doesn't have `check_ldconfig` set at all, and apply `check_ldconfig = False`

boegel:
@Micket The implementation is pretty safe, since it checks first whether the `$LDSHARED` used to build Python packages already uses the toolchain compiler.

If not (when `intel` toolchain is used together with `Python` built with `GCCcore`), only then is a different `$LDSHARED` value set.

So, I don't think we need to make this specific for certain toolchains, or go back to easyconfigs and set `check_ldshared = False` specifically.

We can do that in case having this enabled by default is problematic, but it shouldn't be (and if it is, we can still try to fix that).

We're changing this default for the upcoming EasyBuild v4.0 release, which gives us an excuse to make "breaking" changes (although nothing should actually break due to this change).

Micket:
I know, but, but the reason why we did it "opt-in" to start with hasn't actually changed; it broke some existing old configs. These old configs relies on broken behavior of distutils.

To be specific, merging this PR would break the build for: `mpi4py-2.0.0-intel-2016b-Python-2.7.12.eb` (and several others equally old configs).

I'm not sure if we really care about these old versions (I don't use them), but, it also wouldn't be that terribly hard to fix them. Or drop them.

boegel:
@Micket OK, just re-discovered my own fairly thorough testing of this (cfr. https://github.com/easybuilders/easybuild-easyblocks/pull/1455#issuecomment-457529348), you're right...

I'm not a big fan of hunting down easyconfigs where we should set `check_ldshared = False`, so we should come up with an additional check before we fiddle with `$LDSHARED`...

Maybe something as simple as simply implementing a bypass for any Python 2.x versions older than `2.7.15` and Python 3.x versions older than `3.7.2` is enough?

Micket:
But we also can't put a runtime check inside `extra_options` so shall we instead do:
```python
'check_ldshared': [None, 'Check Python value of $LDSHARED, correct if needed to "$CC -shared"', CUSTOM],
```
and then, then used:
```python
if (check_ldshared is None and old_broken_versions) or check_ldshared == False:
    # Not touching LDSHARED
else:
   # Do touch LDSHARED
```
(ugh.. so ugly, perhaps a 3 value enum would be better)

boegel:
@Micket Updated to `check_ldshared` is only set to `True` for sufficiently recent Python versions, should be good to go now?

easybuild/easyblocks/generic/pythonpackage.py
```diff
@@ -453,6 +458,17 @@ def configure_step(self):
             except IOError:
                 raise EasyBuildError("Creating %s failed", self.sitecfgfn)
 
+        # conservatively auto-enable checking of $LDSHARED if it is not explicitely enabled or disabled
+        # only do this for sufficiently recent Python versions (>= 3.7 or Python 2.x >= 2.7.15)
+        if self.cfg.get('check_ldshared') is None:
+            pyver = det_python_version(self.python_cmd)
+            recent_py2 = pyver.startswith('2') and LooseVersion(pyver) >= LooseVersion('2.7.15')
+            if recent_py2 or LooseVersion(pyver) >= LooseVersion('3.7'):
```
Micket: Do we need to go this recent?

boegel: It's a conservative choice.

We're not actually building `Python` with `GCCcore` anyway (in the central repos) with older versions.

Micket: Are you possibly using rpath?

PR #1780, opened 2019-08-17
add custom easyblock for OpenMPI

boegel:
to be included with EasyBuild 4.0 (not 3.x)

PR #1789, opened 2019-08-27
add custom easyblock for OpenMPI

boegel:
replacement for #1780, now to correct target branch (`develop`, not `4.x` which is gone now)


boegel:
@akesandgren Please check this again. I've updated this based on your suggestion in https://github.com/easybuilders/easybuild-easyblocks/pull/1780/files/ac48256fa462423a887f98d5e2cffd3a0ff5b6ed#diff-9fe1f1d737f3e5809029a835098489d8 to also consider `PMIx` and `libevent`.
I'm not sure whether we should hardwire the requirement for `libevent` if `PMIx` is used though, we should probably leave that up to the `OpenMPI` `configure` script?

Also, what about `--with-slurm`, should we try and auto-detect that if it's not specified? And if so, how? Would checking for a `slurm` OS dependency make sense?

akesandgren:
If building OpenMPI with an external PMIx one MUST use the same libevent that PMIx was built with.
Can't quite remember if one can use an external libevent to build OpenMPIs internal PMIx or not.
Never tried but I guess it's possible.

I would leave the use of Slurm up to the site, i.e. a hook. It can come from too many directions. And even if it is there, the site may not want to use it, i.e. we don't. We only use PMIx, not Slurm, for OpenMPI's that can. One reason for leaving Slurm out of OpenMPI is the constant changing of API's, and thus changing the libslurm.so.x.y numbering, forcing a rebuild of OpenMPI against an upgraded Slurm.

boegel:
@akesandgren So it's even stricter than just "if PMIx is used, libevent must be used too".

I seem to recall that the OpenMPI configure script produces a proper error message if you try to use PMIx without libevent, so we don't need to check for it ourselves in this easyblock?

boegel:
@akesandgren Indeed, if `$EBROOTLIBEVENT` is set, then it will be passed to `--with-libevent` with the current approach.

With that settled, this is good to go in?

easybuild/easyblocks/o/openmpi.py
```diff
@@ -0,0 +1,120 @@
...
+            'mpicxx': os.getenv('CXX', 'g++'),
+            'mpifort': os.getenv('FC', 'gfortran'),
+            'mpif90': os.getenv('F90', 'gfortran'),
+        }
+        # actual pattern for gfortran is "GNU Fortran"
+        for key in ['mpifort', 'mpif90']:
+            if expected[key] == 'gfortran':
+                expected[key] = "GNU Fortran"
+
+        custom_commands = ["%s --version | grep '%s'" % (key, expected[key]) for key in sorted(expected.keys())]
```
akesandgren: Is this verified to work with Intel and PGI too?

boegel: It's exactly for those compilers this was added, see https://github.com/easybuilders/easybuild-easyconfigs/blob/master/easybuild/easyconfigs/o/OpenMPI/OpenMPI-1.10.2-PGI-16.4-GCC-5.3.0-2.26.eb#L33 and https://github.com/easybuilders/easybuild-easyconfigs/blob/master/easybuild/easyconfigs/o/OpenMPI/OpenMPI-1.10.3-iccifort-2016.3.210-GCC-5.4.0-2.26.eb#L32

akesandgren:
Going in, thanks @boegel!

PR #1793, opened 2019-08-29
stop trying to use setuptools.setup in setup.py, always use distutils.core.setup instead

boegel:
See also https://github.com/easybuilders/easybuild-framework/pull/2988 and cfr. https://github.com/easybuilders/easybuild-framework/issues/2941 .


migueldiascosta:
Going in, thanks @boegel!

PR #1797, opened 2019-09-03
Setting a cudaarch template var so that it can be used in sources

edmondac:
Moving the cudaarch setting from fetch_sources to easyconfig template_values means it still works when building CUDA, and also works in the EasyConfig unit tests now.

### DIFF
--- a/easybuild/easyblocks/c/cuda.py
+++ b/easybuild/easyblocks/c/cuda.py
@@ -49,6 +49,7 @@ class EB_CUDA(Binary):
     """
     Support for installing CUDA.
     """
+
     @staticmethod
     def extra_options():
         """Create a set of wrappers based on a list determined by the easyconfig file"""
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/c/cuda.py
```diff
@@ -73,11 +59,18 @@ def fetch_sources(self, sources=None, checksums=None):
...
-        for source in sources:
-            modified_sources.append(source % {'cudaarch': cudaarch})
+        super(EB_CUDA, self).__init__(*args, **kwargs)
+
+        self.cfg.template_values['cudaarch'] = cudaarch
+        self.cfg.generate_template_values()
 
-        return Binary.fetch_sources(self, modified_sources, checksums)
+    @staticmethod
+    def extra_options():
```
boegel: @edmondac Please move this back up, above the `__init__` method (that's just a style issue, not a technical requirement)

edmondac: Done

edmondac:
Thanks @wpoely86 :-)

PR #1799, opened 2019-09-06
Add custom easyblock to install iccifort as self-contained software.

bartoldeman:
It inherits everything from ifort, except for the sanity check for
which it calls both icc's and then ifort's sanity check.

### DIFF
--- a/easybuild/easyblocks/i/iccifort.py
+++ b/easybuild/easyblocks/i/iccifort.py
@@ -1,0 +1,46 @@
+##
+# Copyright 2019-2019 Bart Oldeman, McGill University, Compute Canada
+#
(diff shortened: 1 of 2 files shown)


bartoldeman:
To be used in https://github.com/easybuilders/easybuild-easyconfigs/pull/8879

boegel:
@bartoldeman Some enhancements/cleanup for this in https://github.com/ComputeCanada/easybuild-easyblocks/pull/4, some of it related to the framework changes in https://github.com/easybuilders/easybuild-framework/pull/2995

boegel:
Tested with `iccifort` easyconfig in https://github.com/easybuilders/easybuild-easyconfigs/pull/8879, good to go, thanks @bartoldeman!

PR #1800, opened 2019-09-09
filter out empty entries in $CPATH or $C_INCLUDE_PATH when building Perl, since that breaks the build

boegel:
fixes #8859 and #8874


akesandgren:
Going in, thanks @boegel!

PR #1801, opened 2019-09-10
fix sanity check command pattern for output of 'mpif90 --version' with PGI compilers

boegel:


migueldiascosta:
Going in, thanks @boegel!

PR #1805, opened 2019-09-16
don't load modules for dependencies in CrayToolchain.prepare_step

boegel:
requires https://github.com/easybuilders/easybuild-framework/pull/3008

This should have been there all along, rather than relying on using `dummy` as a version for the `dummy` toolchain (which is deprecated in EasyBuild v4.0).

See also https://github.com/eth-cscs/production/pull/1174#pullrequestreview-255440220

cc @victorusu, @gppezzi, @lucamar

### DIFF
--- a/easybuild/easyblocks/generic/craytoolchain.py
+++ b/easybuild/easyblocks/generic/craytoolchain.py
@@ -41,6 +41,13 @@ class CrayToolchain(Bundle):
     """
     Compiler toolchain: generate module file only, nothing to build/install
     """
+    def prepare_step(self, *args, **kwargs):
+        """Prepare build environment (skip loaded of dependencies)."""
+
+        kwargs['load_tc_deps_modules'] = False
+
+        super(CrayToolchain, self).prepare_step(*args, **kwargs)
+
     def make_module_dep(self):
         """
         Generate load/swap statements for dependencies in the module file


lucamar:
Thanks @boegel, this fix should make redundant the temporary workaround that I have used in `craytoolchain.py`, as reported by @victorusu in https://github.com/eth-cscs/production/pull/1174 (namely, defining `AVAIL_PRGENVS` instead of `KNOWN_PRGENVS`). How can I test the fix? Do I need to bootstrap EasyBuild files from `easybuilders:develop`? With release 3.9.1 I get
`TypeError: prepare_step() got an unexpected keyword argument 'load_tc_deps_modules'`

boegel:
@lucamar To test, you would need to apply the changes made to `easybuild/framework/easyblock.py` from https://github.com/easybuilders/easybuild-framework/pull/3008/files#diff-a1bcbbc1aab6b59688b219f298fddb56 (which are fairly limited and backwards-compatible), and then download the updated `craytoolchain.py` and make EasyBuild use it via `eb --include-easyblocks craytoolchain.py`.

I tested it myself on Piz Daint, with this change (and the other fixes from https://github.com/easybuilders/easybuild-framework/pull/3008/files) I was able to install both `CrayGNU/19.06` and `HPL/2.3-CrayGNU-19.06` on top of it.

These fixes will be included in the upcoming EasyBuild v4.0; with EasyBuild v3.x you use the workaround of using `dummy` as toolchain version (but you'll still hit the problem with the faulty `source` being included in the output of `modulecmd python load`).

PR #1814, opened 2019-09-24
fix Boost easyblock for versions >=1.71.0

smoors:


easybuild/easyblocks/b/boost.py
```diff
@@ -167,20 +167,26 @@ def configure_step(self):
...
-        run_cmd("./bjam --clean-all", log_all=True, simple=True)
+        run_cmd("./%s --clean-all" % bjamcmd, log_all=True, simple=True)
 
     def build_step(self):
         """Build Boost with bjam tool."""
 
+        if LooseVersion(self.version) >= LooseVersion("1.71.0"):
+            bjamcmd = 'b2'
+        else:
+            bjamcmd = 'bjam'
```
boegel: Why not move this in the constructor and use `self.bjamcmd`, so you don't have to pass it to `self.build_boost_variant`?

smoors: good idea.

PR #1815, opened 2019-09-25
update RepeatMasker easyblock for version 4.0.9

smoors:


smoors:
not yet entirely working as it should

PR #1817, opened 2019-09-27
avoid UnicodeDecodeError when reading 'configure' script and checking for 'Generated by GNU Autoconf' in ConfigureMake

boegel:
fix for following error (hit this on CentOS 8 when installing `Bison-3.3.2.eb` with `$LC_ALL` set to `C`)

```
  File "/tmp/kehoste/eb400/lib/python3.6/site-packages/easybuild/main.py", line 112, in build_and_install_software
  File "/tmp/kehoste/eb400/lib/python3.6/site-packages/easybuild/framework/easyblock.py", line 3051, in build_and_install_one
  File "/tmp/kehoste/eb400/lib/python3.6/site-packages/easybuild/framework/easyblock.py", line 2959, in run_all_steps
  File "/tmp/kehoste/eb400/lib/python3.6/site-packages/easybuild/framework/easyblock.py", line 2829, in run_step
    step_method(self)()
  File "/tmp/kehoste/eb400/lib/python3.6/site-packages/easybuild/easyblocks/generic/configuremake.py", line 219, in configure_step
    if os.path.exists(configure_command) and AUTOCONF_GENERATED_MSG in read_file(configure_command):
  File "/tmp/kehoste/eb400/lib/python3.6/site-packages/easybuild/tools/filetools.py", line 188, in read_file
    txt = handle.read()
  File "/usr/lib64/python3.6/encodings/ascii.py", line 26, in decode
    return codecs.ascii_decode(input, self.errors)[0]
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 741356: ordinal not in range(128)
```


damianam:
If that makes you happy we can merge that, but I would add some comments explaining why. Someone will read it at some point and think "why is this a bytes literal"?

Besides that, that's IMO an environment problem. I have had that before in completely different scenarios that had nothing to do with EB. If you environment is not sanitized to handle UTF-8 that is bound to happen sooner or later, doesn't matter if you try to hide it in this particular case.

boegel:
@damianam I agree this is an edge case and perhaps a shoot-in-your-own-foot situation, but since this surprised me a bit when I was hitting it, it's in a very commonly used generic easyblock, and it's pretty much trivial to fix, I would like to get this merged.

It's not like we can shoot ourselves in the foot here, can we? ;)

I tested this with both Python 2 & 3, on different systems, should be OK to merge without introducing problems...

damianam:
@boegel That's fine for me. I am not sure I agree this is our duty to fix this, since whoever is using this will hit the same error further down the road, but I guess it a reasonable safeguard for this easyblock. Going in!

PR #1819, opened 2019-09-27
Add easyblock for installing MotionCor2.

akesandgren:
Also installs wrapper to load correct version of CUDA.


easybuild/easyblocks/m/motioncor2.py
```diff
@@ -0,0 +1,114 @@
...
+     - creates wrapper that loads the correct version of CUDA before
+     - running the actual binary
+    """
+
+    def __init__(self, *args, **kwargs):
+        super(EB_MotionCor2, self).__init__(*args, **kwargs)
+
+    def extract_step(self):
+        """Extract the files"""
+        super(EB_MotionCor2, self).extract_step()
```
boegel: You can drop both `__init__` and `extract_step`, since you're not customizing anything here.

easybuild/easyblocks/m/motioncor2.py
```diff
@@ -0,0 +1,114 @@
...
+        adjust_permissions(wrapper, stat.S_IRWXU, add=True)
+
+    def sanity_check_step(self):
+        """
+        Custom sanity check for MotionCor2
+        """
+
+        custom_paths = {
+                        'files': [os.path.join('bin', x) for x in ['motioncor2', self.motioncor2_bin]],
+                        'dirs': []
```
boegel: Please fix indent (4 spaces)

easybuild/easyblocks/m/motioncor2.py
```diff
@@ -0,0 +1,114 @@
...
+    def build_step(self):
+        """No compilation, this is binary software"""
+        pass
+
+    def install_step(self):
+        """
+        Install binary and a wrapper that loads correct CUDA version.
+        """
+
+        cuda_mod_name = ""
```
boegel: Use `None` as default value, makes more sense... And also initialize `cuda_name`:

```
cuda_mod_name, cuda_name = None, None
```

easybuild/easyblocks/m/motioncor2.py
```diff
@@ -0,0 +1,114 @@
...
+        for dep in self.toolchain.dependencies:
+            if dep['name'] == 'CUDA':
+                cuda_mod_name = dep['short_mod_name']
+                cuda_name = cuda_mod_name.split('/')[0]
+                break
+
+        if cuda_mod_name == "":
```
boegel: change from `== ""` to `is None`

easybuild/easyblocks/m/motioncor2.py
```diff
@@ -0,0 +1,114 @@
...
+            raise EasyBuildError("CUDA must be a direct dependency of MotionCor2")
+        cuda_ver = get_software_version('CUDA')
```
boegel: Why not grab the version above in the for loop (`dep['version']`?

We should also move this to the constructor (`__init__`), to ensure `self.motioncor2_bin` is defined when `--module-only` is used... That also means we can't use `get_software_version` (since modules for deps are not loaded yet until `prepare_step` is run, but if you'll use `dep['version']`, that's fine.

easybuild/easyblocks/m/motioncor2.py
```diff
@@ -0,0 +1,114 @@
...
+        self.motioncor2_bin = 'MotionCor2_%s-Cuda%s' % (self.version, "".join(cuda_ver.split('.')[:2]))
```
boegel: define a `cuda_shortver` to clean this up a bit?

easybuild/easyblocks/m/motioncor2.py
```diff
@@ -0,0 +1,114 @@
...
+        src_mc2_bin = os.path.join(self.builddir, self.motioncor2_bin)
+        if not os.path.exists(src_mc2_bin):
+            raise EasyBuildError("Specified CUDA version has no corresponding MotionCor2 binary")
+
+        bindir = os.path.join(self.installdir, 'bin')
+        mkdir(bindir)
+
+        dst_mc2_bin = os.path.join(bindir, self.motioncor2_bin)
+        copy_file(src_mc2_bin, dst_mc2_bin)
+        adjust_permissions(dst_mc2_bin, stat.S_IRWXU, add=True)
```
boegel: Do we really need write permissions on `dst_mc2_bin`?

easybuild/easyblocks/m/motioncor2.py
```diff
@@ -0,0 +1,114 @@
...
+    def install_step(self):
+        """
+        Install binary and a wrapper that loads correct CUDA version.
+        """
+
+        cuda_mod_name = ""
+        for dep in self.toolchain.dependencies:
+            if dep['name'] == 'CUDA':
+                cuda_mod_name = dep['short_mod_name']
+                cuda_name = cuda_mod_name.split('/')[0]
```
boegel: I think you actually want to just drop the last part of the module name, since in theory you chould have `foo/bar/CUDA/10.1.234` as module name, and then you want to retain the `foo/bar/CUDA`.
Maybe (ab)use `os.path.dirname(cuda_mod_name)`?

easybuild/easyblocks/m/motioncor2.py
```diff
@@ -0,0 +1,112 @@
...
+    """
+
+    def __init__(self, *args, **kwargs):
+        super(EB_MotionCor2, self).__init__(*args, **kwargs)
+
+        self.cuda_mod_name, self.cuda_name = None, None
+        for dep in self.cfg.dependencies():
+            if dep['name'] == 'CUDA':
+                self.cuda_mod_name = dep['short_mod_name']
+                self.cuda_name = self.cuda_mod_name.rsplit('/', 1)[0]
```
boegel: @akesandgren Any reason not to use `os.path.dirname`? Easier to understand imho, and the module name is basically a file path, so it makes sense imho...

easybuild/easyblocks/m/motioncor2.py
```diff
@@ -0,0 +1,112 @@
...
+        if not os.path.exists(src_mc2_bin):
+            raise EasyBuildError("Specified CUDA version has no corresponding MotionCor2 binary")
+
+        bindir = os.path.join(self.installdir, 'bin')
+        mkdir(bindir)
+
+        dst_mc2_bin = os.path.join(bindir, self.motioncor2_bin)
+        copy_file(src_mc2_bin, dst_mc2_bin)
+
+        exe_perms = stat.S_IXUSR | stat.S_IXGRP | stat.S_IXOTH
```
boegel: Are we sure that read permissions are enable for all?

boegel: Are we sure that read permissions are enable for all?

easybuild/easyblocks/m/motioncor2.py
akesandgren: Didn't think of it. For me, treating it as a path is odd in this context, but if you prefer it that way I can change it.

easybuild/easyblocks/m/motioncor2.py
akesandgren: I assume you intended to make the comment in two places...

Anyway, for the wrapper it should be unless user has set an odd umask that write_file picks up on.
I'll add reads, just forgot it when i changed it...

boegel: Doesn't hurt to make sure here :)

easybuild/easyblocks/m/motioncor2.py
boegel: I think it would be a lot easier to understand what's going on here, compared to the `rsplit` you're using now.

In fact, it's even wrong:

```
>>> 'foo/CUDA/1.0'.rsplit('/')[0]
'foo'
>>> os.path.dirname('foo/CUDA/1.0')
'foo/CUDA'
```

akesandgren: No it isn't wrong, you're just not doing what I am :-)

rsplit('/', 1)

boegel: Well, that only proves my point that `os.path.dirname` makes more sense. ;)

boegel:
@akesandgren I've fixed a similar issues in the tests in https://github.com/easybuilders/easybuild-easyblocks/pull/1825 as you need here, should be fairly trivial to make the tests pass now here too if you take a look at that...

akesandgren:
You mean add stuff to  test/easyblocks/module.py ...
so for the module-only part i should just have to add motioncor2.py and CUDA to that list of req_deps?

And something similar in test/easyblocks/init_easyblocks.py ??

boegel:
@akesandgren If you move what's now being done in `__init__` to `prepare_step` (which is not skipped when `--module-only` is used) instead, then only making the obvious changes (which you mentioned) in `test/easyblocks/module.py` should be sufficient...

easybuild/easyblocks/m/motioncor2.py
```diff
@@ -0,0 +1,123 @@
...
+                self.cuda_name = os.path.dirname(self.cuda_mod_name)
+                cuda_ver = dep['version']
+                cuda_short_ver = "".join(cuda_ver.split('.')[:2])
+                self.motioncor2_bin = 'MotionCor2_%s-Cuda%s' % (self.version, cuda_short_ver)
+                break
+
+    def prepare_step(self, *args, **kwargs):
+        super(EB_MotionCor2, self).prepare_step(*args, **kwargs)
+
+        # This is technically already done in __init__
```
boegel: @akesandgren Why not do everything in `prepare_step`?

You ran into trouble earlier because the workaround in the tests only "hacks" `get_software_root`, not `.dependencies()`.

This looks a bit messy now, we shouldn't be dancing around limitations in the tests by tweaking an easyblock like this...

akesandgren: So self.cfg.dependencies() is still available in prepare_step, should have thought of that..

I'l fix it...

easybuild/easyblocks/m/motioncor2.py
```diff
@@ -0,0 +1,117 @@
...
+    """
+
+    def prepare_step(self, *args, **kwargs):
+        super(EB_MotionCor2, self).prepare_step(*args, **kwargs)
+
+        if not get_software_root('CUDA'):
+            raise EasyBuildError("CUDA must be a direct (build)dependency of MotionCor2")
+
+        self.cuda_mod_name, self.cuda_name = None, None
+        self.motioncor2_bin = None
```
boegel: @akesandgren Sorry for the nitpicking, but the initialization of class variables belongs in the constructor (`__init__`). :)

Can you move these two lines to `__init__`?

```python
    """Constructor of MotionCor2 easyblock."""
    super(EB_MotionCor2, self).__init__(*args, **kwargs)

    self.cuda_mod_name, self.cuda_name = None, None
    self.motioncor2_bin = None

File: easybuild/easyblocks/m/motioncor2.py [hunk 13]
```diff
@@ -0,0 +1,117 @@
...
+
+
+class EB_MotionCor2(EasyBlock):
+    """
+    Support for installing MotionCor2
+     - creates wrapper that loads the correct version of CUDA before
+     - running the actual binary
+    """
+
+    def prepare_step(self, *args, **kwargs):
```
boegel: add a docstring please:

```python
def prepare_step(self, *args, **kwargs):
    """
    Determine name of MotionCor2 binary to install based on CUDA version.
    """

boegel:
Looks good, tested with https://github.com/easybuilders/easybuild-easyconfigs/pull/8942, so good to go, thanks @akesandgren!

PR #1822, opened 2019-10-02
add --verbose flag to 'pip install' when running EB in debug mode

smoors:
this is often useful when pip is compiling from source: it will show the actual compilation commands being used.

### DIFF
--- a/easybuild/easyblocks/generic/pythonpackage.py
+++ b/easybuild/easyblocks/generic/pythonpackage.py
@@ -44,6 +44,7 @@ from easybuild.easyblocks.python import EBPYTHONPREFIXES, EXTS_FILTER_PYTHON_PAC
 from easybuild.framework.easyconfig import CUSTOM
 from easybuild.framework.extensioneasyblock import ExtensionEasyBlock
 from easybuild.tools.build_log import EasyBuildError
+from easybuild.tools.config import build_option
 from easybuild.tools.filetools import mkdir, remove_dir, which
 from easybuild.tools.modules import get_software_root
 from easybuild.tools.py2vs3 import string_type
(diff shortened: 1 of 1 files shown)


verdurin:
Looks like a good idea to me.

easybuild/easyblocks/generic/pythonpackage.py
```diff
@@ -56,6 +56,8 @@
 EASY_INSTALL_TARGET = "easy_install"
 EASY_INSTALL_INSTALL_CMD = "%(python)s setup.py " + EASY_INSTALL_TARGET + " --prefix=%(prefix)s %(installopts)s %(loc)s"
 PIP_INSTALL_CMD = "pip install --prefix=%(prefix)s %(installopts)s %(loc)s"
+if fancylogger.getLogger().level == fancylogger.logging.DEBUG:
```
boegel: @smoors This makes sense, but you should use `if build_option('debug')` here rather than talking to `fancylogger`.

Also, you shouldn't do it here, but in `PythonPackage.__init__` where `self.install_cmd = PIP_INSTALL_CMD` is used.

easybuild/easyblocks/generic/pythonpackage.py
```diff
@@ -222,6 +223,10 @@ def __init__(self, *args, **kwargs):
 
         self.install_cmd_output = ''
 
+        if build_option('debug'):
+            global PIP_INSTALL_CMD
+            PIP_INSTALL_CMD += ' --verbose'
```
boegel: @smoors I would move this below, where `self.install_cmd = PIP_INSTALL_CMD` is used, as:

```python
if build_option('debug'):
    self.install_cmd += ' --verbose'

smoors: @boegel care to explain why you think that's better?

boegel:
Tested & approved, thanks @smoors!

PR #1825, opened 2019-10-04
add support for installing QScintilla on top of PyQt5

Darkless012:


easybuild/easyblocks/q/qscintilla.py
```diff
@@ -86,32 +86,43 @@ def install_step(self):
...
                 '--qsci-sipdir %s' % qsci_sipdir,
                 '--qsci-incdir %s' % os.path.join(self.installdir, 'include'),
                 '--qsci-libdir %s' % os.path.join(self.installdir, 'lib'),
-                '--pyqt-sipdir %s' % os.path.join(pyqt, 'share', 'sip', 'PyQt4'),
+                '--pyqt-sipdir %s' % os.path.join(self.pyqt_root, 'share', 'sip', self.pyqt_version),
                 '--apidir %s' % os.path.join(self.installdir, 'qsci', 'api', 'python'),
                 '--no-stubs',
+                '--no-dist-info',
             ]
+            if pyqt5:
```
boegel: You'll need to change this if you move things to `prepare_step`:

```python
if self.pyqt_pkg_name == 'PyQt5':
```

easybuild/easyblocks/q/qscintilla.py
```diff
@@ -136,11 +150,13 @@ def sanity_check_step(self):
...
         # also install Python bindings if Python is included as a dependency
         python = get_software_root('Python')
+
+        run_cmd("python -c 'import sys; print(sys.path)'")
         custom_commands = []
         if python:
-            custom_paths['dirs'] += [os.path.join(det_pylibdir(), 'PyQt4'), os.path.join('qsci', 'api', 'python'),
-                                     os.path.join('share', 'sip', 'PyQt4')]
-            custom_commands = ["python -c 'import PyQt4.Qsci'"]
+            custom_paths['dirs'] += [os.path.join(det_pylibdir(), self.pyqt_version), os.path.join('qsci', 'api', 'python'),
```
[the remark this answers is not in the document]
boegel: Use one entry per line, and use `.extend` rather than `+=` while you're at it:

```python
custom_paths['dirs'].extend([
    os.path.join(det_pylibdir(), self.pyqt_version),
    os.path.join('qsci', 'api', 'python'),
    os.path.join('share', 'sip', self.pyqt_version),
])

File: easybuild/easyblocks/q/qscintilla.py [hunk 3]
```diff
@@ -136,11 +150,13 @@ def sanity_check_step(self):
...
+                                     os.path.join('share', 'sip', self.pyqt_version)]
+            custom_commands = ["python -c 'import "+self.pyqt_version+".Qsci'"]
```
boegel: Hmm, I would expect Hound to bitch about lack of spaces around `+` here (please fix)

Or maybe use:

```python
custom_commands = ["python -c 'import %s.Qsci'" % self.pyqt_pkg_name]

File: easybuild/easyblocks/q/qscintilla.py [hunk 4]
```diff
@@ -86,32 +86,43 @@ def install_step(self):
...
                 '--qsci-incdir %s' % os.path.join(self.installdir, 'include'),
                 '--qsci-libdir %s' % os.path.join(self.installdir, 'lib'),
-                '--pyqt-sipdir %s' % os.path.join(pyqt, 'share', 'sip', 'PyQt4'),
+                '--pyqt-sipdir %s' % os.path.join(self.pyqt_root, 'share', 'sip', self.pyqt_version),
                 '--apidir %s' % os.path.join(self.installdir, 'qsci', 'api', 'python'),
                 '--no-stubs',
+                '--no-dist-info',
             ]
+            if pyqt5:
+                 cfgopts.append("--pyqt=PyQt5")
```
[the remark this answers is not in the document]
Darkless012: good remark, according to the docs it would, but this flag was only added in version 2.11.*.
Previous version didn't have it AFAIK. This will need some additional logic there.

easybuild/easyblocks/q/qscintilla.py
```diff
@@ -86,32 +86,43 @@ def install_step(self):
...
+            pyqt5 = get_software_root('PyQt5')
+            pyqt = get_software_root('PyQt')
+            if pyqt5:
+                self.pyqt_root = pyqt5
+                self.pyqt_version = "PyQt5"
+            elif pyqt:
+                self.pyqt_root = pyqt
+                self.pyqt_version = "PyQt4"
+            else:
+                raise EasyBuildError("Failed to determine PyQt installation prefix, PyQt/PyQt5 not included as dependency?")
```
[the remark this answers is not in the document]
Darkless012: Well, I'd rather keep the "dependency" word since for me as beginner is direct hint what is wrong. But rephrased to be shorter.

easybuild/easyblocks/q/qscintilla.py
```diff
@@ -136,11 +150,13 @@ def sanity_check_step(self):
         }
         # also install Python bindings if Python is included as a dependency
         python = get_software_root('Python')
+
+        run_cmd("python -c 'import sys; print(sys.path)'")
```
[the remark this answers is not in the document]
Darkless012: oops

easybuild/easyblocks/q/qscintilla.py
```diff
@@ -87,31 +107,33 @@ def install_step(self):
...
             cfgopts = [
                 '--destdir %s' % os.path.join(self.installdir, pylibdir),
                 '--qsci-sipdir %s' % qsci_sipdir,
                 '--qsci-incdir %s' % os.path.join(self.installdir, 'include'),
                 '--qsci-libdir %s' % os.path.join(self.installdir, 'lib'),
-                '--pyqt-sipdir %s' % os.path.join(pyqt, 'share', 'sip', 'PyQt4'),
+                '--pyqt-sipdir %s' % os.path.join(self.pyqt_root, 'share', 'sip', self.pyqt_pkg_name),
                 '--apidir %s' % os.path.join(self.installdir, 'qsci', 'api', 'python'),
                 '--no-stubs',
+                '--no-dist-info',
```
boegel: @Darkless012 It looks like this `--no-dist-info` flag is only supported in recent version of `QScintilla`, certainly not for version 2.10:

```
configure.py: error: no such option: --no-dist-info
```

boegel: It was added in 2.10.7, see https://www.riverbankcomputing.com/news/qscintilla-2107, so this needs a version check...

boegel: fixed in 5de67dc

PR #1826, opened 2019-10-05
make sure "generic=True" actually turns on generic build in GCC easyblock

akesandgren:


boegel:
Thanks for the fix @akesandgren!

PR #1827, opened 2019-10-08
Make an easyblock for Gctf in the same style as MotionCor2.

akesandgren:
Gctf also needs a wrapper around it to handle CUDA

### DIFF
--- a/easybuild/easyblocks/g/gctf.py
+++ b/easybuild/easyblocks/g/gctf.py
@@ -1,0 +1,131 @@
+##
+# Copyright 2019-2019 Ghent University
+#
+# This file is part of EasyBuild,
+# originally created by the HPC team of Ghent University (https://ugent.be/hpc/en),
+# with support of Ghent University (https://ugent.be/hpc),
+# the Flemish Supercomputer Centre (VSC) (https://www.vscentrum.be),
+# Flemish Research Foundation (FWO) (https://www.fwo.be/en)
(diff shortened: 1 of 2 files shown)


easybuild/easyblocks/g/gctf.py
```diff
@@ -0,0 +1,131 @@
...
+
+        src_gctf_bin = os.path.join(self.builddir, 'bin', self.gctf_bin)
+        if not os.path.exists(src_gctf_bin):
+            raise EasyBuildError("Specified CUDA version has no corresponding Gctf binary")
+
+        bindir = os.path.join(self.installdir, 'bin')
+        mkdir(bindir)
+
+        dst_gctf_bin = os.path.join(bindir, self.gctf_bin)
+        copy_file(src_gctf_bin, dst_gctf_bin)
```
smoors: @akesandgren minor suggestion: maybe copy `gctf_bin` to a location that is not in the user's $PATH, to somewhat discourage them from using it directly without the wrapper?

akesandgren: could have done that, but i was lazy... and if they really want to shoot themselves in the foot noone can really stop them.

The main usage of gctf (and motioncor2) is probably RELION and Scipion and from there we control the usage...

boegel: The name of the binary is weird enough that most people will use `gctf` anyway.

I understand @smoors' suggestion, but it doesn't make much of a difference in practice I think.

PR #1828, opened 2019-10-14
update VEP easyblock to make installation compatible with Bio::EnsEMBL::XS

lexming:
Update for `vep` easyblock to add sanity checks and set specific environment variables for the extension `Bio::EnsEMBL::XS` if it is present in the list of extension list.

### DIFF
--- a/easybuild/easyblocks/v/vep.py
+++ b/easybuild/easyblocks/v/vep.py
@@ -101,10 +101,8 @@ class EB_VEP(EasyBlock):
         }
 
         if 'Bio::EnsEMBL::XS' in [ext[0] for ext in self.cfg['exts_list']]:
-            perl_majver = get_major_perl_version()
-            perl_ver = get_software_version('Perl')
             custom_paths['files'] += ['lib/perl%s/site_perl/%s/x86_64-linux-thread-multi/Bio/EnsEMBL/XS.pm'\
-                                     % (perl_majver, perl_ver)]
+                                     % (get_major_perl_version(), get_software_version('Perl'))]
 
         custom_commands = ['vep --help']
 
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/v/vep.py
```diff
@@ -93,10 +94,16 @@ def install_step(self):
...
     def sanity_check_step(self):
         """Custom sanity check for VEP."""
+
         custom_paths = {
             'files': ['vep'],
             'dirs': ['modules/Bio/EnsEMBL/VEP'],
         }
+
+        if 'Bio::EnsEMBL::XS' in [ext[0] for ext in self.cfg['exts_list']]:
+            custom_paths['files'] += ['lib/perl%s/site_perl/%s/x86_64-linux-thread-multi/Bio/EnsEMBL/XS.pm'\
```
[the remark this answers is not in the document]
boegel: @lexming Please use `extend`, use `os.path.join` and avoid breaking the long line by using a local variable:

```python
if 'Bio::EnsEMBL::XS' in [ext[0] for ext in self.cfg['exts_list']]:
    perl_maj_ver = get_major_perl_version()
    perl_ver = get_software_version('Perl')
    perl_lib_subdir = os.path.join('lib', 'perl' + perl_maj_ver, 'site_perl', perl_ver)
    bio_ensembl_xs_perl_mod = os.path.join(perl_lib_subdir, 'x86_64-linux-thread-multi', Bio', 'EnsEMBL', 'XS.pm')
    custom_paths['files'].extend(bio_ensembl_xs_perl_mod)

File: easybuild/easyblocks/v/vep.py [hunk 2]
```diff
@@ -105,9 +112,13 @@ def make_module_req_guess(self):
         """Custom guesses for environment variables (PATH, ...) for VEP."""
         perl_majver = get_major_perl_version()
 
+        perl_libpath = [self.api_mods_subdir]
+        if 'Bio::EnsEMBL::XS' in [ext[0] for ext in self.cfg['exts_list']]:
+            perl_libpath += ['lib/perl%s/site_perl/%s' % (perl_majver, get_software_version('Perl'))]
```
boegel: same here, please use `.extend` and `os.path.join`

boegel:
```python
perl_maj_ver = get_major_perl_version()
perl_ver = get_software_version('Perl')
perl_libpath.extend(os.path.join('lib', 'perl' + perl_maj_ver, 'site_perl', perl_ver)

easybuild/easyblocks/v/vep.py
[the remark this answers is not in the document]
lexming: Fixed with your suggestions. Thanks for the feedback.

easybuild/easyblocks/v/vep.py
lexming: Fixed as well :+1:

boegel:
tested with existing `VEP` easyconfigs and https://github.com/easybuilders/easybuild-easyconfigs/pull/8733, good to go

Thanks @lexming!

PR #1831, opened 2019-10-15
fix compatibility of Trilinos easyblock with Python 3

boegel:
This slipped through during the previous regression test (which was only partially done using Python 3)...


akesandgren:
Going in, thanks @boegel!

PR #1836, opened 2019-10-22
fix for conda packages that rely on particular versions of Python

jerowe:
Hi, @boegel!

I made a rather bone headed error when I was figuring out the logic for conda installs. There doesn't need to be two separate commands for first creating the conda env and then installing the requirement. It should be one, or else you run into errors.

An example error would be if you have a build dep of `Miniconda3/4.7.10` along with the following easyconfig - 

```
##
# This is an easyconfig file for EasyBuild, see https://github.com/easybuilders/easybuild
##

easyblock = 'Conda'

name = "hisat2"
version = "2.1.0"

homepage = 'https://bioconda.github.io/recipes/hisat2/README.html'
description = """graph-based alignment of next generation sequencing reads to a population of genomes"""

toolchain = SYSTEM

requirements = "%(name)s=%(version)s"
channels = ['bioconda', 'conda-forge']

builddependencies = [('Miniconda3', '4.7.10')]

sanity_check_paths = {
    'files': ['bin/python'],
    'dirs': ['bin']
}

moduleclass = 'tools'
```

### DIFF
--- a/easybuild/easyblocks/generic/conda.py
+++ b/easybuild/easyblocks/generic/conda.py
@@ -84,25 +84,24 @@ class Conda(Binary):
             run_cmd(cmd, log_all=True, simple=True)
 
         else:
-            cmd = "%s conda create --force -y -p %s" % (self.cfg['preinstallopts'], self.installdir)
-            run_cmd(cmd, log_all=True, simple=True)
 
+	    self.set_conda_env()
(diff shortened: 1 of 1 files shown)


boegel:
Let's see if I can wake up those tests with a close/re-open...

PR #1837, opened 2019-10-23
New xmipp easyblock handling Xmipp based on Scons.

akesandgren:
Builds both with and without CUDA support.

Since this is a complete rewrite don't look at the diff against the old one, look at the complete file.

### DIFF
--- a/easybuild/easyblocks/x/xmipp.py
+++ b/easybuild/easyblocks/x/xmipp.py
@@ -1,5 +1,5 @@
 ##
-# Copyright 2019 Ghent University
+# Copyright 2015-2019 Ghent University
 #
 # This file is part of EasyBuild,
 # originally created by the HPC team of Ghent University (http://ugent.be/hpc/en),
@@ -25,6 +25,9 @@
 """
 EasyBuild support for building and installing Xmipp, implemented as an easyblock
 
+@author: Jens Timmerman (Ghent University)
+@author: Pablo Escobar (sciCORE, SIB, University of Basel)
+@author: Kenneth Hoste (Ghent University)
 @author: Ake Sandgren (HPC2N, Umea University)
 """
 import os
@@ -50,12 +53,13 @@ class EB_Xmipp(SCons):
         self.srcdir = os.path.join(self.builddir, 'src')
         self.cfg['start_dir'] = self.builddir
         self.use_cuda = False
+        self.cfgfile = os.path.join(self.builddir, 'xmipp.conf')
 
     def extract_step(self):
         """Extract Xmipp sources."""
         # Xmipp assumes that everything is unpacked in a "src" dir
         mkdir(self.srcdir)
-        self.cfg.update('unpack_options', '--directory src')
+        self.cfg.update('unpack_options', '--directory %s' % os.path.basename(self.srcdir))
         super(EB_Xmipp, self).extract_step()
         for module in self.xmipp_modules:
             symlink('%s-%s' % (module, self.version), os.path.join(self.srcdir, module), use_abspath_source=False)
@@ -79,10 +83,15 @@ class EB_Xmipp(SCons):
         """Custom configuration procedure for Xmipp."""
 
         # Initialize the config file and then patch it with the correct values
-        self.cfgfile = os.path.join(self.builddir, 'xmipp.conf')
-        cmd = '%s src/xmipp/xmipp config' % self.cfg['preconfigopts']
+        cmd = ' '.join([
+            self.cfg['preconfigopts'],
+            os.path.join(self.srcdir, 'xmipp', 'xmipp'),
+            'config',
+            self.cfg['configopts'],
+        ])
         run_cmd(cmd, log_all=True, simple=True)
 
+        # Parameters to be set in the config file
         params = {
             'CC': os.environ['CC'],
             'CXX': os.environ['CXX'],
@@ -95,7 +104,7 @@ class EB_Xmipp(SCons):
         }
 
         deps = [
-            # dep name, is required dep?
+            # Name of dependency, name of env var to set, required or not
             ('HDF5', None, True),
             ('SQLite', None, True),
             ('LibTIFF', None, True),
@@ -103,8 +112,6 @@ class EB_Xmipp(SCons):
             ('Java', 'JAVA_HOME', True),
             ('MATLAB', 'MATLAB_DIR', False),
         ]
-
-        regex_subs = []
 
         cuda_root = get_software_root('CUDA')
         if cuda_root:
@@ -145,6 +152,9 @@ class EB_Xmipp(SCons):
         jnipath = os.pathsep.join([jnipath, os.path.join(jnipath, 'linux')])
         params.update({'JNI_CPPPATH': jnipath})
 
+        regex_subs = []
+
+        # Set the variables in the config file to match the environment from EasyBuild
         for (key, val) in params.items():
             regex_subs.extend([(r'^%s\s*=.*$' % key, r'%s = %s' % (key, val))])
 
@@ -165,19 +175,26 @@ class EB_Xmipp(SCons):
     def build_step(self):
         """Custom build step for Xmipp."""
 
-        # First build cuFFTAdvisor
+        # First build cuFFTAdvisor, XmippCore depends on this when CUDA is enabled.
         if self.use_cuda:
             cwd = change_dir(os.path.join(self.srcdir, 'cuFFTAdvisor'))
-            cmd = '%s make all' % self.cfg['preinstallopts']
+            cmd = ' '.join([
+                self.cfg['prebuildopts'],
+                'make',
+                'all',
+                self.cfg['buildopts'],
+            ])
             run_cmd(cmd, log_all=True, simple=True)
             change_dir(cwd)
             xmipp_lib = os.path.join(self.srcdir, 'xmipp', 'lib')
             mkdir(xmipp_lib)
-            copy_file(os.path.join(self.srcdir, 'cuFFTAdvisor', 'build', 'libcuFFTAdvisor.so'), xmipp_lib)
+            shlib_ext = get_shared_lib_ext()
+            libname = 'libcuFFTAdvisor.%s' % shlib_ext
+            copy_file(os.path.join(self.srcdir, 'cuFFTAdvisor', 'build', libname), xmipp_lib)
 
         self.cfg.update('buildopts', '--verbose')
         for module in self.xmipp_modules:
-            moddir = os.path.join('src', module)
+            moddir = os.path.join(os.path.basename(self.srcdir), module)
             symlink(self.cfgfile, os.path.join(self.srcdir, module, 'install', 'xmipp.conf'))
             cwd = change_dir(moddir)
             super(EB_Xmipp, self).build_step()
@@ -188,7 +205,13 @@ class EB_Xmipp(SCons):
 
         # Use the xmipp script to do the install, there's too much
         # to replicate here.
-        cmd = '%s src/xmipp/xmipp install %s' % (self.cfg['preinstallopts'], self.installdir)
+        cmd = ' '.join([
+            self.cfg['preinstallopts'],
+            os.path.join(self.srcdir, 'xmipp', 'xmipp'),
+            'install',
+            self.installdir,
+            self.cfg['installopts'],
+        ])
         run_cmd(cmd, log_all=True, simple=True)
 
         # Remove the bash and fish files. Everything should be in the module
@@ -204,8 +227,8 @@ class EB_Xmipp(SCons):
         libs = ['XmippCore', 'XmippJNI', 'XmippParallel', 'Xmipp']
         custom_paths = {
             'files': [os.path.join('bin', x) for x in bins] +
-                     [os.path.join('bindings', 'python', 'xmippViz.py')] +
-                     [os.path.join('lib', 'lib%s.%s') % (x, shlib_ext) for x in libs],
+                 [os.path.join('bindings', 'python', 'xmippViz.py')] +
+                 [os.path.join('lib', 'lib%s.%s') % (x, shlib_ext) for x in libs],
             'dirs': ['resources'],
         }
         return super(EB_Xmipp, self).sanity_check_step(custom_paths=custom_paths)
@@ -218,7 +241,7 @@ class EB_Xmipp(SCons):
         return txt
 
     def make_module_req_guess(self):
-        """Custom guesses for environment variables (PATH, ...) for Xmipp."""
+        """Custom guesses for environment variables for Xmipp."""
         guesses = super(EB_Xmipp, self).make_module_req_guess()
         guesses.update({
             'LD_LIBRARY_PATH': ['lib', os.path.join('bindings', 'python')],
--- a/easybuild/easyblocks/x/xmipp.py
+++ b/easybuild/easyblocks/x/xmipp.py
@@ -82,6 +82,8 @@ class EB_Xmipp(SCons):
     def configure_step(self):
         """Custom configuration procedure for Xmipp."""
 
+        # Tell xmipp config that there is no Scipion.
+        env.setvar('XMIPP_NOSCIPION', 'True')
         # Initialize the config file and then patch it with the correct values
         cmd = ' '.join([
             self.cfg['preconfigopts'],
--- a/easybuild/easyblocks/x/xmipp.py
+++ b/easybuild/easyblocks/x/xmipp.py
@@ -1,5 +1,5 @@
 ##
-# Copyright 2015-2019 Ghent University
+# Copyright 2019 Ghent University
 #
 # This file is part of EasyBuild,
 # originally created by the HPC team of Ghent University (http://ugent.be/hpc/en),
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/x/xmipp.py
```diff
@@ -25,174 +25,196 @@
...
+        bins = ['xmipp_%s' % x for x in ['angular_rotate', 'classify_kerdensom', 'mpi_ml_refine3d']]
+        libs = ['XmippCore', 'XmippJNI', 'XmippParallel', 'Xmipp']
         custom_paths = {
-            # incomplete list, random picks, cfr. http://xmipp.cnb.csic.es/twiki/bin/view/Xmipp/ListOfProgramsv3
-            'files': ['bin/xmipp_%s' % x for x in ['compile', 'imagej', 'mpi_run', 'phantom_create',
-                                                   'transform_filter', 'tomo_project', 'volume_align']],
-            'dirs': ['lib'],
+            'files': [os.path.join('bin', x) for x in bins] +
+                     [os.path.join('bindings', 'python', 'xmippViz.py')] +
+                     [os.path.join('lib', 'lib%s.%s') % (x, shlib_ext) for x in libs],
```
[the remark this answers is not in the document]
akesandgren: I can never remember how these are supposed to look like...  please help...

akesandgren:
Need to fix NVCC_CXXFLAGS so more changes coming.

easybuild/easyblocks/x/xmipp.py
```diff
@@ -1,5 +1,5 @@
 ##
-# Copyright 2015-2019 Ghent University
+# Copyright 2019 Ghent University
```
boegel: keep the `2015-2019` please (even though it's a rewrite)

easybuild/easyblocks/x/xmipp.py
```diff
@@ -25,174 +25,203 @@
 """
 EasyBuild support for building and installing Xmipp, implemented as an easyblock
 
-@author: Jens Timmerman (Ghent University)
-@author: Pablo Escobar (sciCORE, SIB, University of Basel)
-@author: Kenneth Hoste (Ghent University)
```
boegel: retain these original authors?

easybuild/easyblocks/x/xmipp.py
```diff
@@ -25,174 +25,203 @@
...
+        self.cfg['start_dir'] = self.builddir
+        self.use_cuda = False
 
     def extract_step(self):
         """Extract Xmipp sources."""
-        # strip off 'xmipp' part to avoid having everything in a 'xmipp' subdirectory
-        self.cfg.update('unpack_options', '--strip-components=1')
+        # Xmipp assumes that everything is unpacked in a "src" dir
+        mkdir(self.srcdir)
+        self.cfg.update('unpack_options', '--directory src')
```
boegel: use `os.path.basename(self.srcdir)` rather than hardcoding `src`

```python
self.cfg.update('unpack_options', '--directory %s' % os.path.basename(self.srcdir))
```

easybuild/easyblocks/x/xmipp.py
```diff
@@ -25,174 +25,203 @@
...
+        """Patch files from self.srcdir dir."""
+        super(EB_Xmipp, self).patch_step(beginpath=self.srcdir)
+
+    def setup_xmipp_env(self):
+        """Setup environment before running SCons."""
+        env.setvar('LINKERFORPROGRAMS', os.environ['CXX'])
+        env.setvar('MPI_CC', os.getenv('MPICC'))
+        env.setvar('MPI_CXX', os.getenv('MPICXX'))
+        env.setvar('MPI_LINKERFORPROGRAMS', os.getenv('MPICXX'))
+        incpath = os.path.join(get_software_root('Java'), 'include')
```
boegel: This will crash hard if `Java` is not included as a dep, should be handled with:

```python
java_root = get_software_root('Java')
if java_root:
    java_incpath = os.path.join(java_root, 'include')
else:
    raise EasyBuildError("Java not included as dependency")
```

easybuild/easyblocks/x/xmipp.py
```diff
@@ -25,174 +25,203 @@
...
-        ])
-
-        # define list of configure options, which will be passed to Xmipp's install.sh script via --configure-args
-        self.cfg['configopts'] = configure_args
-        self.log.info("Configure arguments for Xmipp install.sh script: %s", self.cfg['configopts'])
+        """Custom configuration procedure for Xmipp."""
+
+        # Initialize the config file and then patch it with the correct values
+        self.cfgfile = os.path.join(self.builddir, 'xmipp.conf')
+        cmd = '%s src/xmipp/xmipp config' % self.cfg['preconfigopts']
```
boegel: also pick up on `configopts`?

```python
cmd = ' '.join([
    self.cfg['preconfigopts'],
    os.path.join(self.srcdir, 'xmipp', 'xmipp'),
    'config',
    self.cfg['configopts'],
])

File: easybuild/easyblocks/x/xmipp.py [hunk 7]
```diff
@@ -25,174 +25,203 @@
...
-            'LIBPATH=%s' % os.getenv('LD_LIBRARY_PATH'),
-        ])
-
-        # define list of configure options, which will be passed to Xmipp's install.sh script via --configure-args
-        self.cfg['configopts'] = configure_args
-        self.log.info("Configure arguments for Xmipp install.sh script: %s", self.cfg['configopts'])
+        """Custom configuration procedure for Xmipp."""
+
+        # Initialize the config file and then patch it with the correct values
+        self.cfgfile = os.path.join(self.builddir, 'xmipp.conf')
```
boegel: this line should be moved into `__init__`

easybuild/easyblocks/x/xmipp.py
```diff
@@ -25,174 +25,203 @@
...
+            'LINKERFORPROGRAMS': os.environ['CXX'],
+            'LIBDIRFLAGS': '',
+            # I don't think Xmipp can actually build without MPI.
+            'MPI_CC': os.environ.get('MPICC', 'UNKNOWN'),
+            'MPI_CXX': os.environ.get('MPICXX', 'UNKNOWN'),
+            'MPI_LINKERFORPROGRAMS': os.environ.get('MPICXX', 'UNKNOWN'),
+        }
+
+        deps = [
+            # dep name, is required dep?
```
boegel: clarify comment, 2nd element is environment variable to set?

easybuild/easyblocks/x/xmipp.py
```diff
@@ -25,174 +25,203 @@
...
+            ('HDF5', None, True),
+            ('SQLite', None, True),
+            ('LibTIFF', None, True),
+            ('libjpeg-turbo', None, True),
+            ('Java', 'JAVA_HOME', True),
+            ('MATLAB', 'MATLAB_DIR', False),
+        ]
+
+        regex_subs = []
```
boegel: move down to where this is first used

easybuild/easyblocks/x/xmipp.py
```diff
@@ -25,174 +25,203 @@
...
+
+        cuda_root = get_software_root('CUDA')
+        if cuda_root:
+            params.update({'CUDA_BIN': os.path.join(cuda_root, 'bin')})
+            params.update({'CUDA_LIB': os.path.join(cuda_root, 'lib64')})
+            params.update({'NVCC': os.environ['CUDA_CXX']})
+            # Their default for NVCC is to use g++-5, fix that
+            nvcc_flags = '--x cu -D_FORCE_INLINES -Xcompiler -fPIC '
+            nvcc_flags = nvcc_flags + '-Wno-deprecated-gpu-targets -std=c++11'
```
boegel: Aren't we going to regret hardcoding the `std=c++11` later?

easybuild/easyblocks/x/xmipp.py
```diff
@@ -25,174 +25,203 @@
...
+        if missing_deps:
+            raise EasyBuildError("One or more required dependencies not available: %s", ', '.join(missing_deps))
+
+        # Here we know that Java exists since it is a required dependency
+        jnipath = os.path.join(get_software_root('Java'), 'include')
+        jnipath = os.pathsep.join([jnipath, os.path.join(jnipath, 'linux')])
+        params.update({'JNI_CPPPATH': jnipath})
+
+        for (key, val) in params.items():
+            regex_subs.extend([(r'^%s\s*=.*$' % key, r'%s = %s' % (key, val))])
```
boegel: This is a bit cryptic, please add a comment to explain what's going on here?

easybuild/easyblocks/x/xmipp.py
```diff
@@ -25,174 +25,203 @@
...
+        env.setvar('CPATH', os.pathsep.join([os.environ['CPATH'], pyincpath]))
+
+        super(EB_Xmipp, self).configure_step()
 
     def build_step(self):
-        """No custom build step (see install step)."""
-        pass
+        """Custom build step for Xmipp."""
+
+        # First build cuFFTAdvisor
```
boegel: Is there a specific reason to build this first?

easybuild/easyblocks/x/xmipp.py
```diff
@@ -25,174 +25,203 @@
...
+        if self.use_cuda:
+            cwd = change_dir(os.path.join(self.srcdir, 'cuFFTAdvisor'))
+            cmd = '%s make all' % self.cfg['preinstallopts']
```
boegel: Should be `prebuildopts`, no?

```python
cmd = ' '.join([
    self.cfg['prebuildopts'],
    'make',
    'all',
    self.cfg['buildopts'],
])
```

No parallel build here?

easybuild/easyblocks/x/xmipp.py
```diff
@@ -25,174 +25,203 @@
...
+            run_cmd(cmd, log_all=True, simple=True)
+            change_dir(cwd)
+            xmipp_lib = os.path.join(self.srcdir, 'xmipp', 'lib')
+            mkdir(xmipp_lib)
+            copy_file(os.path.join(self.srcdir, 'cuFFTAdvisor', 'build', 'libcuFFTAdvisor.so'), xmipp_lib)
```
boegel: please avoid hardcoding `.so`, use `get_shared_lib_ext()`

```python
libname = 'libcuFFTAdvisor.%s' % shlib_ext
copy_file(os.path.join(self.srcdir, 'cuFFTAdvisor', 'build', libname), xmipp_lib)

File: easybuild/easyblocks/x/xmipp.py [hunk 15]
```diff
@@ -25,174 +25,203 @@
...
-            sys.stdout.write(line)
-
-        cmd = './install.sh -j %s --unattended=true %s' % (self.cfg['parallel'], ' '.join(cmd_opts))
-        out, _ = run_cmd(cmd, log_all=True, simple=False)
-
-        if not re.search("Xmipp has been successfully compiled", out):
-            raise EasyBuildError("Xmipp installation did not complete successfully?")
+        # Use the xmipp script to do the install, there's too much
+        # to replicate here.
+        cmd = '%s src/xmipp/xmipp install %s' % (self.cfg['preinstallopts'], self.installdir)
```
boegel:
```python
cmd = ' '.join([
    self.cfg['preinstallopts'],
    os.path.join(self.srcdir, 'xmipp', 'xmipp'),
    'install',
    self.cfg['installopts'],
])

File: easybuild/easyblocks/x/xmipp.py (hunk 1 above)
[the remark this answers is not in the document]
Comment (boegel) [reply easybuild/easyblocks/x/xmipp.py]: I think it wants you to indent by 4 spaces relative to `'files':`

File: easybuild/easyblocks/x/xmipp.py [hunk 16]
```diff
@@ -25,174 +25,203 @@
...
-        """Define Xmipp specific variables in generated module file, i.e. XMIPP_HOME."""
+        """Set the install directory as XMIPP_HOME"""
         txt = super(EB_Xmipp, self).make_module_extra()
         txt += self.module_generator.set_environment('XMIPP_HOME', self.installdir)
-        txt += self.module_generator.prepend_paths('PYTHONPATH', self.xmipp_pythonpaths)
+        self.log.debug("make_module_extra added this: %s", txt)
         return txt
+
+    def make_module_req_guess(self):
+        """Custom guesses for environment variables (PATH, ...) for Xmipp."""
```
boegel: drop the `(PATH, ...)`, since we're not customizing `$PATH`

easybuild/easyblocks/x/xmipp.py
akesandgren: What do you mean? Xmipp assumes that it lives in a directory called "src", it's used in the scripts.

self.srcdir is not named src is it?

easybuild/easyblocks/x/xmipp.py
akesandgren: It won't since that function is only used after veryfying that the required deps (of which Java is one) are already there.

See the deps struct and the missing_deps loop.

easybuild/easyblocks/x/xmipp.py
akesandgren: Ahh, I had actually set it...

easybuild/easyblocks/x/xmipp.py
akesandgren: Maybe, I just didn't feel like changing that from what they use at the moment.

easybuild/easyblocks/x/xmipp.py
akesandgren: It's required to be built first. Fails if not. Enhanced the comment a bit.

easybuild/easyblocks/x/xmipp.py
akesandgren: No, doesn't take that long and i don't trust their makefile.

easybuild/easyblocks/x/xmipp.py
```diff
@@ -28,171 +28,223 @@
...
+
+        bins = ['xmipp_%s' % x for x in ['angular_rotate', 'classify_kerdensom', 'mpi_ml_refine3d']]
+        libs = ['XmippCore', 'XmippJNI', 'XmippParallel', 'Xmipp']
         custom_paths = {
-            # incomplete list, random picks, cfr. http://xmipp.cnb.csic.es/twiki/bin/view/Xmipp/ListOfProgramsv3
-            'files': ['bin/xmipp_%s' % x for x in ['compile', 'imagej', 'mpi_run', 'phantom_create',
-                                                   'transform_filter', 'tomo_project', 'volume_align']],
-            'dirs': ['lib'],
+            'files': [os.path.join('bin', x) for x in bins] +
+                 [os.path.join('bindings', 'python', 'xmippViz.py')] +
```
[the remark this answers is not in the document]
akesandgren: Apparently not that...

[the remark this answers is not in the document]
akesandgren: Gah, made a typo...

akesandgren:
@boegel anything else?

easybuild/easyblocks/x/xmipp.py
boegel: OK, I failed to take into account the order of methods being called

easybuild/easyblocks/x/xmipp.py
boegel: OK, fine, we'll see

easybuild/easyblocks/x/xmipp.py
```diff
@@ -28,171 +28,223 @@
...
+        jnipath = os.pathsep.join([jnipath, os.path.join(jnipath, 'linux')])
+        params.update({'JNI_CPPPATH': jnipath})
+
+        regex_subs = []
+
+        # Set the variables in the config file to match the environment from EasyBuild
+        for (key, val) in params.items():
+            regex_subs.extend([(r'^%s\s*=.*$' % key, r'%s = %s' % (key, val))])
+
+        apply_regex_substitutions(self.cfgfile, regex_subs)
```
boegel: Testing this updated easyblock with #9257 lead to:

```
== FAILED: Installation ended unsuccessfully (build directory: /tmp/easybuild_build/Xmipp/3.19.04-Apollo/foss-2019a-Python-2.7.15): build failed (first 300 chars): Failed to patch /tmp/easybuild_build/Xmipp/3.19.04-Apollo/foss-2019a-Python-2.7.15/xmipp.conf: [Errno 2] No such file or directory: '/tmp/easybuild_build/Xmipp/3.19.04-Apollo/foss-2019a-Python-2.7.15/xmipp.conf' -> '/tmp/easybuild_build/Xmipp/3.19.04-Apollo/foss-2019a-Pyth (took 14 sec)
```

Is there supposed to be an existing `xmipp.conf` which you're patching? I coulnd't find a file like that in the build dir.

If you want to create a new file named `xmipp.conf`, you can't use `apply_regex_substitutions`...

easybuild/easyblocks/x/xmipp.py
```diff
@@ -28,171 +28,225 @@
...
+
+        bins = ['xmipp_%s' % x for x in ['angular_rotate', 'classify_kerdensom', 'mpi_ml_refine3d']]
+        libs = ['XmippCore', 'XmippJNI', 'XmippParallel', 'Xmipp']
         custom_paths = {
-            # incomplete list, random picks, cfr. http://xmipp.cnb.csic.es/twiki/bin/view/Xmipp/ListOfProgramsv3
-            'files': ['bin/xmipp_%s' % x for x in ['compile', 'imagej', 'mpi_run', 'phantom_create',
-                                                   'transform_filter', 'tomo_project', 'volume_align']],
-            'dirs': ['lib'],
+            'files': [os.path.join('bin', x) for x in bins] +
+                [os.path.join('bindings', 'python', 'xmippViz.py')] +
```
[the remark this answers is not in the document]
akesandgren: what now??? I didn't touch that part...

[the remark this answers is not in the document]
boegel: doesn't sense to me either, ignoring it...

PR #1839, opened 2019-10-24
added easyblock to install a module only

mboisson:
This is to be used for moving cuda at the core level (but could have other uses).

### DIFF
--- a/easybuild/easyblocks/generic/moduleonly.py
+++ b/easybuild/easyblocks/generic/moduleonly.py
@@ -1,0 +1,76 @@
+##
+# Copyright 2018-2019 Ghent University
+#
+# This file is part of EasyBuild,
+# originally created by the HPC team of Ghent University (http://ugent.be/hpc/en),
+# with support of Ghent University (http://ugent.be/hpc),
+# the Flemish Supercomputer Centre (VSC) (https://www.vscentrum.be),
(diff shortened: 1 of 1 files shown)


migueldiascosta:
@mboisson if I understand correctly you ended up not using this at CC, but still want to contribute it in case there are other uses?

boegel:
Also, can you motivate the need for this next to the existing `Bundle`?

mboisson:
Yes, Bart ended up using the Bundle module instead of this, mostly to not add new code, I think. 

As for motivation, I guess it is just that I found the Bundle easyblock way over-complicated for something that is not supposed to install anything. I realize now that it states in the description that it does not build or install, but the code is much more complicated than doing nothing. It was not at all obvious to me that this easyblock ends up doing nothing but create a module. The name is also not obvious, ModuleOnly is much more explicit.

mboisson:
How do you reconcile 
```
    """
    Bundle of modules: only generate module files, nothing to build/install
    """
```

with
```
            'default_component_specs': [{}, "Default specs to use for every component", CUSTOM],
            'components': [(), "List of components to install: tuples w/ name, version and easyblock to use", CUSTOM],
```

mboisson:
I don't think is needed anymore. Closing. Let me know if you think otherwise. @bartoldeman

PR #1840, opened 2019-10-25
Minor fix to name of Gctf binary.

akesandgren:
Other easyconfigs/blocks (RELION/Scipion) already depends on Gctf as name.
Make sure both Gctf and gctf exists so nothing breaks.


migueldiascosta:
Going in, thanks @akesandgren!

PR #1843, opened 2019-11-05
fixes for TensorFlow easyblocks w.r.t. Bazel build options & __init__ in top-level google-protobuf package dir

Flamefire:
Fix the warning 

> WARNING: The following configs were expanded more than once: [v2, cuda, using_cuda]. For repeatable flags, repeats are counted twice and may lead to unexpected behavior. 

in TF 2 (see https://github.com/easybuilders/easybuild-easyblocks/commit/7f62330e84710626ed87ef404d2c9a170cfbd729#diff-925e7942309e22e4c64cb2d99b50c206) and remove `google/__init__.py` for Python 3.3 (see https://github.com/easybuilders/easybuild-easyconfigs/issues/9207)

### DIFF
--- a/easybuild/easyblocks/t/tensorflow.py
+++ b/easybuild/easyblocks/t/tensorflow.py
@@ -38,12 +38,12 @@ from distutils.version import LooseVersion
 
 import easybuild.tools.environment as env
 import easybuild.tools.toolchain as toolchain
-from easybuild.easyblocks.generic.pythonpackage import PythonPackage
+from easybuild.easyblocks.generic.pythonpackage import PythonPackage, det_python_version
 from easybuild.easyblocks.python import EXTS_FILTER_PYTHON_PACKAGES
 from easybuild.framework.easyconfig import CUSTOM
 from easybuild.tools.build_log import EasyBuildError
 from easybuild.tools.filetools import adjust_permissions, apply_regex_substitutions, copy_file, mkdir, resolve_path
-from easybuild.tools.filetools import is_readable, read_file, which, write_file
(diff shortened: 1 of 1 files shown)


boegel:
Looks good on first sight, I'll test this with a couple of existing TensorFlow easyconfigs to make sure there are no regressions introduced...

Flamefire:
Highly appreciated!

Note that I "removed support" for Tensorflow 2 RCs which did require the `--config=v2`, but including that would add a lot of trash to the source (you need the option for 2 RC but not 2 and not 1 and not >=2.0.1) because (in difference to SemVer) `2.0.0-rc1>=2.0.0`. With the release of the 2.0.0 I'd rather not include that code.

boegel:
Agreed, I don't think having no support for 2.0.0-rc* is an issue. That also explains why the `--config=v2` was there (I updated the easyblock using 2.0.0-rc2)

boegel:
Tested with a variety of existing `TensorFlow` easyconfigs, including:

* `TensorFlow-2.0.0-foss-2019a-Python-3.7.2.eb`
* `TensorFlow-1.14.0-foss-2019a-Python-3.7.2.eb`
* `TensorFlow-1.13.1-foss-2018b-Python-3.6.6.eb`
* `TensorFlow-1.8.0-intel-2018a-Python-3.6.4.eb`
* `TensorFlow-1.3.0-intel-2017a-Python-2.7.13.eb`

Works as expected, so no reason to hold this back. Thanks a lot for the fixes @Flamefire!

PR #1844, opened 2019-11-06
add GitHub Actions workflow to run easybuild-easyblocks test suite

boegel:
Equivalent to the `easybuild-framework` repo (cfr. https://github.com/easybuilders/easybuild-framework/pull/3039), this PR add the configuration file to run the easyblocks test suite in the native GitHub CI, next to Travis

The long term goal is to stop relying on Travis (which we'll keep using as is for now, and soon only for running the tests on top of Python 2.6).

General availability for GitHub Actions is planned for mid November, it's probably worth waiting until shortly after before fully relying on it...

### DIFF
--- a/.github/workflows/unit_tests.yml
+++ b/.github/workflows/unit_tests.yml
@@ -1,0 +1,129 @@
+name: easyblocks unit tests
+on: [push, pull_request]
+jobs:
+  build:
+    runs-on: ubuntu-18.04
+    strategy:
+      matrix:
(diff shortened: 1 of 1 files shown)


boegel:
Works as expected, see https://github.com/boegel/easybuild-easyblocks/commit/99f3e3776a6d65f58c778a498bc8a3c5228b549e/checks?check_suite_id=298313945

akesandgren:
Going in, thanks @boegel!

PR #1845, opened 2019-11-08
fix support for sequential version in MUMPS easyblock

migueldiascosta:


akesandgren:
Going in, thanks @migueldiascosta!

PR #1848, opened 2019-11-13
change default value of files_to_copy to None in MakeCp generic easyblock + code cleanup & use change_dir, copy_dir, copy_file and mkdir function from filetools

boegel:
Workaround for https://github.com/easybuilders/easybuild-framework/issues/3049, and the code cleanup is good as a side effect.

Re-tested with different easyconfigs relying on `MakeCp`, including `HISAT2-2.1.0-foss-2018b.eb`, `prodigal-2.6.3-GCCcore-8.2.0.eb` and `BamTools-2.5.1-foss-2018b.eb`.

### DIFF
--- a/easybuild/easyblocks/generic/makecp.py
+++ b/easybuild/easyblocks/generic/makecp.py
@@ -55,10 +55,6 @@ class MakeCp(ConfigureMake):
         extra.update(extra_vars)
         return ConfigureMake.extra_options(extra_vars=extra)
 
-    def __init__(self, *args, **kwargs):
-        """Constructor for MakeCp easyblock."""
-        super(MakeCp, self).__init__(*args, **kwargs)
-
     def configure_step(self, cmd_prefix=''):
         """
         Configure build if required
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/generic/makecp.py
```diff
@@ -47,14 +47,18 @@ def extra_options(extra_vars=None):
...
-            'files_to_copy': [[], "List of files or dirs to copy", MANDATORY],
+            'files_to_copy': [None, "List of files or dirs to copy", MANDATORY],
             'with_configure': [False, "Run configure script before building", BUILD],
         }
         if extra_vars is None:
             extra_vars = {}
         extra.update(extra_vars)
         return ConfigureMake.extra_options(extra_vars=extra)
 
+    def __init__(self, *args, **kwargs):
```
wpoely86: is this needed?

boegel: It's not, I was planning something else first, and then overlook not actually doing anything in the constructor

Fixed in 0f6992e8e

PR #1849, opened 2019-11-14
add support to Toolchain generic easyblock for defining $EB* environment variables for toolchain components that use an external module

boegel:
This is required to construct a toolchain using external modules (provided by OpenHPC packages, for example).

The metadata for external modules is only picked up for dependencies, not for the toolchain (partially because that would require access to the easyconfig file rather than just an existing module for the toolchain itself).

In addition, the installation prefix & version for the toolchain components must be available via the corresponding `$EBROOT*` and `$EBVERSION*` environment variables when setting up the build environment.
For EasyBuild-generated modules those environment variables are set by the module files, but for external modules we must leverage the metadata files to take care of that ourselves.

cc @koomie @adrianreber

edit: requires ~~https://github.com/easybuilders/easybuild-framework/pull/3088~~

### DIFF
--- a/easybuild/easyblocks/generic/toolchain.py
+++ b/easybuild/easyblocks/generic/toolchain.py
@@ -65,7 +65,11 @@ class Toolchain(Bundle):
 
                 mod_name = dep['full_mod_name']
                 metadata = dep['external_module_metadata']
-                names, versions = metadata.get('name'), metadata.get('version')
+                names, versions = metadata.get('name', []), metadata.get('version')
+
+                # if no versions are available, use None as version (for every software name)
+                if versions is None:
+                    versions = [None] * len(names)
 
                 if names:
                     self.log.info("Adding environment variables for %s provided by external module %s", names, mod_name)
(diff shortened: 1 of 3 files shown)


easybuild/easyblocks/generic/toolchain.py
```diff
@@ -31,12 +31,75 @@
...
+        # in the generated module file for this toolchain;
+        # this is required to let the EasyBuild framework set up the build environment based on the toolchain
+        if self.cfg.get('set_env_external_modules', False):
+            for dep in [d for d in self.cfg['dependencies'] if d['external_module']]:
+
+                mod_name = dep['full_mod_name']
+                metadata = dep['external_module_metadata']
+                names, versions, prefix = metadata.get('name'), metadata.get('version'), metadata.get('prefix')
+
+                if names:
```
wpoely86: can this not be definied? Why this if?

boegel: Yes, it's possible the metadata doesn't specify the software name, and then we can't define anything at all for that particular external module.

boegel:
@wpoely86 Test added, but needs https://github.com/easybuilders/easybuild-framework/pull/3088 to be merged first

PR #1852, opened 2019-11-15
Improved configuration for netcdf and netcdf4_python

lexming:
Currently `netcdf` sets wrong CMake variables for configuring the build with HDF5. Even though we can find the following comment in `CMakeLists.txt` 

> Accommodate developers who have hdf5 libraries and
> headers on their system, but do not have a the hdf
> .cmake files.  If this is the case, they should
> specify HDF5_HL_LIB, HDF5_LIB, HDF5_INCLUDE_DIR manually.

The variables are actually called `HDF5_C_LIBRARY` and `HDF5_HL_LIBRARY`. This PR fixes those variables and also adds support for building `netCDF` with `PnetCDF`.

These fixes for `netcdf` have been tested with
* netCDF-4.4.1-foss-2016b.eb
* netCDF-4.5.0-foss-2017b.eb
* netCDF-4.6.0-foss-2018a.eb
* netCDF-4.6.1-foss-2018b.eb
* netCDF-4.6.2-gompi-2019a.eb

Additionally, `netcdf4_python` is updated to properly handle external `cURL` and `JPEG` libraries.

### DIFF
--- a/easybuild/easyblocks/n/netcdf.py
+++ b/easybuild/easyblocks/n/netcdf.py
@@ -68,7 +68,10 @@ class EB_netCDF(CMakeMake):
             ConfigureMake.configure_step(self)
 
         else:
-            self.cfg.update('configopts', '-DCMAKE_BUILD_TYPE=RELEASE -DCMAKE_C_FLAGS_RELEASE="-DNDEBUG " ')
+            if self.toolchain.options['debug']:
+                self.cfg.update('configopts', '-DCMAKE_BUILD_TYPE=DEBUG ')
+            else:
+                self.cfg.update('configopts', '-DCMAKE_BUILD_TYPE=RELEASE -DCMAKE_C_FLAGS_RELEASE="-DNDEBUG " ')
 
             for (dep, libname) in [('cURL', 'curl'), ('HDF5', 'hdf5'), ('Szip', 'sz'), ('zlib', 'z'),
                                    ('PnetCDF', 'pnetcdf')]:
--- a/easybuild/easyblocks/n/netcdf.py
+++ b/easybuild/easyblocks/n/netcdf.py
@@ -69,7 +69,8 @@ class EB_netCDF(CMakeMake):
 
         else:
             self.cfg.update('configopts', '-DCMAKE_BUILD_TYPE=RELEASE -DCMAKE_C_FLAGS_RELEASE="-DNDEBUG " ')
-            for (dep, libname) in [('cURL', 'curl'), ('HDF5', 'hdf5'), ('Szip', 'sz'), ('zlib', 'z')]:
+            for (dep, libname) in [('cURL', 'curl'), ('HDF5', 'hdf5'), ('Szip', 'sz'), ('zlib', 'z'),
+                                   ('PnetCDF', 'pnetcdf')]:
                 dep_root = get_software_root(dep)
                 dep_libdir = get_software_libdir(dep)
                 if dep_root:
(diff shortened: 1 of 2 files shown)


easybuild/easyblocks/n/netcdf.py
```diff
@@ -69,18 +69,22 @@ def configure_step(self):
 
         else:
             self.cfg.update('configopts', '-DCMAKE_BUILD_TYPE=RELEASE -DCMAKE_C_FLAGS_RELEASE="-DNDEBUG " ')
```
akesandgren: While you're working on it, this line is wrong.
It should not set this if toolchainopt debug= True but rather BUILD_TYPE=Debug and so on...

easybuild/easyblocks/n/netcdf.py
```diff
@@ -69,18 +69,22 @@ def configure_step(self):
...
                 dep_libdir = get_software_libdir(dep)
                 if dep_root:
                     incdir = os.path.join(dep_root, 'include')
                     self.cfg.update('configopts', '-D%s_INCLUDE_DIR=%s ' % (dep.upper(), incdir))
                     if dep == 'HDF5':
                         env.setvar('HDF5_ROOT', dep_root)
+                        self.cfg.update('configopts', '-DUSE_HDF5=ON')
                         libhdf5 = os.path.join(dep_root, dep_libdir, 'libhdf5.%s' % shlib_ext)
-                        self.cfg.update('configopts', '-DHDF5_LIB=%s ' % libhdf5)
+                        self.cfg.update('configopts', '-DHDF5_C_LIBRARY=%s ' % libhdf5)
```
akesandgren: This is netCDF version dependent.
v4.3.0 uses HDF5_LIB, HDF5_HL_LIB, HDF5_INCLUDE_DIR
v4.4.0 is the first version to use HDF5_C_LIBRARY, HDF5_HL_LIBRARY, HDF5_INCLUDE_DIR

Please take this into account.

lexming: Thanks for the warning, after reaching version 4.4 I was convinced that the names `HDF5_LIB` and  `HDF5_HL_LIB` were just an error in that comment. I updated the easyblock adding a dict with variable names per version to ease any additional changes that might be introduced in the future.

easybuild/easyblocks/n/netcdf.py
lexming: Added a conditional on `toolchain.options['debug']`. CMakeLists already adds flags for debug builds, so I removed any further settings.

easybuild/easyblocks/n/netcdf.py
```diff
@@ -69,22 +69,36 @@ def configure_step(self):
...
                     if dep == 'HDF5':
                         env.setvar('HDF5_ROOT', dep_root)
                         self.cfg.update('configopts', '-DUSE_HDF5=ON')
-                        libhdf5 = os.path.join(dep_root, dep_libdir, 'libhdf5.%s' % shlib_ext)
-                        self.cfg.update('configopts', '-DHDF5_C_LIBRARY=%s ' % libhdf5)
-                        libhdf5_hl = os.path.join(dep_root, dep_libdir, 'libhdf5_hl.%s' % shlib_ext)
-                        self.cfg.update('configopts', '-DHDF5_HL_LIBRARY=%s ' % libhdf5_hl)
+
+                        hdf5cmvars = {
+                            'hdf5': ('LIB', 'C_LIBRARY'),
```
akesandgren: Please add a clarifying comment here, something like
```
# library name, (old cmake option suffix, new cmake option suffix)
```

lexming: Fixed :+1:

akesandgren:
Works well with old netCDF 4.3.3.1 and 4.4.1.1

akesandgren:
Going in, thanks @lexming!

PR #1853, opened 2019-11-15
add support for running 'pip check' during sanity check in generic PythonPackage easyblock

boegel:
cc @Flamefire

### DIFF
--- a/easybuild/easyblocks/generic/pythonpackage.py
+++ b/easybuild/easyblocks/generic/pythonpackage.py
@@ -179,7 +179,9 @@ def det_pylibdir(plat_specific=False, python_cmd=None):
 
 def det_pip_version():
     """Determine version of currently active 'pip' command."""
+
     pip_version = None
+    log = fancylogger.getLogger('det_pylibdir', fname=False)
 
     out, _ = run_cmd("pip --version", verbose=False, simple=False, trace=False)
 
--- a/easybuild/easyblocks/generic/pythonpackage.py
+++ b/easybuild/easyblocks/generic/pythonpackage.py
@@ -663,7 +663,7 @@ class PythonPackage(ExtensionEasyBlock):
         if self.cfg.get('sanity_pip_check', False):
             pip_version = det_pip_version()
             if pip_version:
-                if LooseVersion(pip_version) >= LooseVersion(pip_version):
+                if LooseVersion(pip_version) >= LooseVersion('9.0.0'):
                     run_cmd("pip check")
                 else:
                     raise EasyBuildError("pip >= 9.0.0 is required for running 'pip check', found %s", pip_version)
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/generic/pythonpackage.py
```diff
@@ -645,6 +660,16 @@ def sanity_check_step(self, *args, **kwargs):
                 exts_filter = (orig_exts_filter[0].replace('python', self.python_cmd), orig_exts_filter[1])
                 kwargs.update({'exts_filter': exts_filter})
 
+        if self.cfg.get('sanity_pip_check', False):
+            pip_version = det_pip_version()
+            if pip_version:
+                if LooseVersion(pip_version) >= LooseVersion(pip_version):
```
wpoely86: say what?

boegel: Ugh. Nice catch.

boegel: fixed in ea82401

easybuild/easyblocks/generic/pythonpackage.py
```diff
@@ -177,6 +177,24 @@ def det_pylibdir(plat_specific=False, python_cmd=None):
     return pylibdir
 
 
+def det_pip_version():
```
Flamefire: I'd refactor this to only do the required work: 
```suggestion
def det_pip_version():
 """Determine version of currently active 'pip' command."""

    out, _ = run_cmd("pip --version", verbose=False, simple=False, trace=False)

    pip_version_regex = re.compile('^pip ([0-9.]+)')
    res = pip_version_regex.search(out)
    if res:
        return res.group(1)
    else:
        log = fancylogger.getLogger('det_pylibdir', fname=False)
        log.warning("Failed to determine pip version from '%s' using pattern '%s'", out, pip_version_regex.pattern)
        return None
```

easybuild/easyblocks/generic/pythonpackage.py
```diff
@@ -645,6 +660,16 @@ def sanity_check_step(self, *args, **kwargs):
                 exts_filter = (orig_exts_filter[0].replace('python', self.python_cmd), orig_exts_filter[1])
                 kwargs.update({'exts_filter': exts_filter})
 
+        if self.cfg.get('sanity_pip_check', False):
+            pip_version = det_pip_version()
+            if pip_version:
+                if LooseVersion(pip_version) >= LooseVersion('9.0.0'):
+                    run_cmd("pip check")
```
wpoely86: this is non - fatal if it fails?

boegel: It's fatal, `run_cmd` will raise an error for commands that exit with a non-zero exit code by default.

easybuild/easyblocks/generic/pythonpackage.py
boegel: I'll enhance the function a bit to use more logging instead...

PR #1854, opened 2019-11-18
fix cuda header paths

smoors:
needed for building custom TF ops


akesandgren:
Going in, thanks @smoors!

smoors:
successful test report can be found here: https://github.com/easybuilders/easybuild-easyconfigs/pull/9322

PR #1857, opened 2019-11-24
add custom easyblock for SCOOP

smoors:
this easyblock patches the remote launchers for broker and workers over ssh to use SCOOP with multiple nodes.

### DIFF
--- a/easybuild/easyblocks/s/scoop.py
+++ b/easybuild/easyblocks/s/scoop.py
@@ -1,0 +1,70 @@
+##
+# Copyright 2009-2019 Ghent University
+#
+# This file is part of EasyBuild,
+# originally created by the HPC team of Ghent University (http://ugent.be/hpc/en),
+# with support of Ghent University (http://ugent.be/hpc),
+# the Flemish Supercomputer Centre (VSC) (https://www.vscentrum.be),
+# Flemish Research Foundation (FWO) (http://www.fwo.be/en)
+# and the Department of Economy, Science and Innovation (EWI) (http://www.ewi-vlaanderen.be/en).
+#
+# https://github.com/easybuilders/easybuild
+#
+# EasyBuild is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation v2.
+#
+# EasyBuild is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with EasyBuild.  If not, see <http://www.gnu.org/licenses/>.
+##
+"""
+EasyBuild support for SCOOP, implemented as an easyblock
+
+@author: Samuel Moors, Vrije Universiteit Brussel (VUB)
+"""
+import os
+import re
+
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/s/scoop.py
```diff
@@ -0,0 +1,70 @@
...
+    def install_step(self):
+        """
+        Patch for launching the broker and workers over ssh
+        Emulate login shell and load SCOOP module
+        """
+
+        super(EB_SCOOP, self).install_step()
+
+        login_shell = 'source /etc/profile && '
+        login_shell += '(source ~/.bash_profile || source ~/.bash_login || source ~/.profile || echo)'
```
boegel: What's this about?

This feels dirty, to say the least...

easybuild/easyblocks/s/scoop.py
```diff
@@ -0,0 +1,70 @@
...
+        mod_name = ActiveMNS().det_full_module_name(self.cfg)
+        set_env = "'%s && module load %s && '" % (login_shell, mod_name)
+        pyshortver = '.'.join(get_software_version('Python').split('.')[:2])
+        base_path = os.path.join(self.installdir, 'lib', 'python%s' % pyshortver, 'site-packages', 'scoop', 'launch')
```
boegel: use `det_pylibdir` function from `pythonpackage.py`?

easybuild/easyblocks/s/scoop.py
```diff
@@ -0,0 +1,70 @@
...
+
+        # patch workerLaunch.py
+        sub = """subprocess.Popen(sshCmd + [
+                    self.hostname,
+                    %s,
+                    self.getCommand(),
+                ],""" % set_env
```
boegel: This is probably more readable:

```python
sub = '\n'.join([
    "subprocess.Popen(sshCmd + [",
    "    self.hostname,",
    "    " + self.env,
    "    self.getCommand()",
    "],",
])

File: easybuild/easyblocks/s/scoop.py (hunk 1 above)
Comment (smoors) [reply easybuild/easyblocks/s/scoop.py]: it does exactly what a login shell does in bash (see the bash man pages).
we need to initialize the environment on the remote machine before we can load the module.
another way to do it is this:
```
ssh <server> "bash --login -c '<commands>'"
```
but that is not trivial given the way the remote launcher is implemented.
feel free to suggest a better way..

smoors: @boegel after thinking about this a bit, I think I will create a regular patch file instead, adding a custom env var for loading the module that will be defined in the easyblock. this will also get rid of the login_shell stuff you complained about.
does that sound good to you?

boegel: It sounds better, but let's see the patch first. ;)

Also, if you just need an extra env var, you can use `modextravars` rather than using an easyblock, or am I missing something?

smoors: the problem is that the full module name depends on the site: HMNS or not, environment modules or lmod, that's why I'd define it in an easyblock.

boegel: We could add a template for the module name in framework?

smoors: yes that would be even better. there is already another software that could make use of it too.
I'll see if I can figure out how to implement it.

smoors:
closing this PR as the patch was already merged, see https://github.com/easybuilders/easybuild-easyconfigs/pull/9366

Issue #1858, opened 2019-11-25
NCL intel and netcdf libdirs are set incorrectly by the EasyBlock


casparvl:
**Issue**
We are using EasyBuild's RPATH support. Thus, EasyBuild does an RPATH sanity check where LD_LIBRARY_PATH is emptied, and EasyBuild checks if all binaries can still find their respective libraries. This failed for the NCL-6.6.2-intel-2018b.eb for various libraries:

```
libnetcdf.so.13 => not found
libiomp5.so => not found
libifcore.so.5 => not found
libifport.so.5 => not found
libpng16.so.16 => not found
libjpeg.so.8 => not found
```

The latter two are because of missing dependencies, which has to be fixed on the EasyConfig side (see https://github.com/easybuilders/easybuild-easyconfigs/issues/9332). The first four are because of the way that the NCL EasyBlock passes the library path (i.e. the paths passed with ```-L``` to the linker). 

**Issue with the intel libs**

From the current EasyBlock:

```
ifort = get_software_root('ifort')
if LooseVersion(get_software_version('ifort')) < LooseVersion('2011.4'):
                ctof_libs = '-lm -L%s/lib/intel64 -lifcore -lifport' % ifort
            else:
                ctof_libs = '-lm -L%s/compiler/lib/intel64 -lifcore -lifport' % ifort
```

However, the ```ifort``` directory structure is:

```
$EBROOTIFORT/compilers_and_libraries/linux/lib/intel64
```

**Proposed fix for intel libs**
Replace the ```ctof_libs``` in the else statement with

```
else:
                ctof_libs = '-lm -L%s/compilers_and_libraries/linux/lib/intel64/ -lifcore -lifport' % ifort
```

**Issue with the netCDF libs**
Libdirs are automatically for

```
deps = ["HDF5", "JasPer", "netCDF", "HDF", "g2lib", "g2clib", "Szip", "UDUNITS"]
```
by this part:
```
for dep in deps:
            root = get_software_root(dep)
            if not root:
                raise EasyBuildError("%s not available", dep)
            libs += ' -L%s/lib ' % root
            includes += ' -I%s/include ' % root
```
Something similar is done for optional dependencies, ```opt_deps = ["netCDF-Fortran", "GDAL"]```.

The problem is that for netCDF, I only have a ```libs64``` directory:

```
[casparl@int2 ~]$ ls -al $EBROOTNETCDF
total 28
[...]

PR #1862, opened 2019-11-27
fix intel and netcdf libdirs in NCL easyblock

casparvl:
This PR fixes the

```
libnetcdf.so.13 => not found
libiomp5.so => not found
libifcore.so.5 => not found
libifport.so.5 => not found
```

errors for the RPATH sanity check of NCL-6.6.2 (i.e. this issue: https://github.com/easybuilders/easybuild-easyblocks/issues/1858)

### DIFF
--- a/easybuild/easyblocks/n/ncl.py
+++ b/easybuild/easyblocks/n/ncl.py
@@ -78,10 +78,10 @@ class EB_NCL(EasyBlock):
         ctof_libs = ''
         ifort = get_software_root('ifort')
         if ifort:
-            if LooseVersion(get_software_version('ifort')) < LooseVersion('2011.4'):
-                ctof_libs = '-lm -L%s/lib/intel64 -lifcore -lifport' % ifort
+            if os.path.exists('%s/lib/intel64/' % ifort) and os.access('%s/lib/intel64/' % ifort, os.R_OK):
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/n/ncl.py
```diff
@@ -78,15 +78,19 @@ def configure_step(self):
...
         if ifort:
-            if LooseVersion(get_software_version('ifort')) < LooseVersion('2011.4'):
-                ctof_libs = '-lm -L%s/lib/intel64 -lifcore -lifport' % ifort
+            if os.path.exists('%s/lib/intel64/' % ifort) and os.access('%s/lib/intel64/' % ifort, os.R_OK):
+                ctof_libs += '-lm -L%s/lib/intel64/ -lifcore -lifport' % ifort
             else:
-                ctof_libs = '-lm -L%s/compiler/lib/intel64 -lifcore -lifport' % ifort
+                self.log.warning(
+                    "Can't find a libdir for ifortran libraries -lifcore -lifport: "
+                    "%s/lib/intel64 doesn't exist or is not accessible." % (ifort, ifort)
```
akesandgren: One too many ifort in the args list there

casparvl: Well spotted, I probably had a different phrase earlier on... corrected it now :)

PR #1871, opened 2019-12-04
Fix CUDA EB for CUDA 10.1

Flamefire:
On Power9 installation fails due to some missing files/folders which is fixed in CUDA 10.2. Using the workaround from https://github.com/easybuilders/easybuild-easyconfigs/issues/9437#issuecomment-561667670

Additionally using non-default locale (LANG env var) crashes some installations with an out-of-bounds exception. Fixed by setting LANG=C

Fixes https://github.com/easybuilders/easybuild-easyconfigs/issues/9437

### DIFF
--- a/easybuild/easyblocks/c/cuda.py
+++ b/easybuild/easyblocks/c/cuda.py
@@ -101,6 +101,22 @@ class EB_CUDA(Binary):
             self.cfg.update('installopts', "--silent --toolkit --toolkitpath=%s --defaultroot=%s" % (
                             self.installdir, self.installdir))
 
+        if LooseVersion("10.0") < LooseVersion(self.version) < LooseVersion("10.2") and get_cpu_architecture() == POWER:
+            # Workaround for
+            # https://devtalk.nvidia.com/default/topic/1063995/cuda-setup-and-installation/cuda-10-1-243-10-1-update-2-ppc64le-run-file-installation-issue/
+            install_script = " && ".join([
+                "mkdir -p %(installdir)s/targets/ppc64le-linux/include",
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/c/cuda.py
```diff
@@ -101,6 +101,24 @@ def install_step(self):
             self.cfg.update('installopts', "--silent --toolkit --toolkitpath=%s --defaultroot=%s" % (
                             self.installdir, self.installdir))
 
+        if LooseVersion("10.0") < LooseVersion(self.version) < LooseVersion("10.2") and get_cpu_architecture() == POWER:
+            # Workaround for
+            # https://devtalk.nvidia.com/default/topic/1063995/cuda-setup-and-installation/cuda-10-1-243-10-1-update-2-ppc64le-run-file-installation-issue/
+            install_script = " && ".join([
+                "echo 'Configuring lol'",
```
bartoldeman: do we need this, or are these accidentally still there?

easybuild/easyblocks/c/cuda.py
```diff
@@ -101,6 +101,24 @@ def install_step(self):
...
+                "mkdir -p %(installdir)s/targets/ppc64le-linux/include",
+                "([ -e %(installdir)s/include ] || ln -s targets/ppc64le-linux/include %(installdir)s/include)",
+                "cp -r %(builddir)s/builds/cublas/src %(installdir)s/.",
+                "echo 'Configuring bar'",
```
bartoldeman: same as above

easybuild/easyblocks/c/cuda.py
Flamefire: done

PR #1874, opened 2019-12-06
Add EasyBlock for cryptography to fix missing -pthread for all versions

Flamefire:
This fixes https://github.com/easybuilders/easybuild-easyconfigs/issues/9446 for all versions of cryptography as all suffer from the same issue. See https://github.com/pyca/cryptography/issues/5084

The approach was taken from https://github.com/bear-rsg/easybuild-easyconfigs/blob/2019b/easybuild/easyconfigs/p/Python/Python-3.7.4-GCCcore-8.3.0.eb#L119

The alternative would be to go and fix every single EC installing the package to add this to the preinstallopts.

Note also that I'm adding `-pthread` as opposed to `-lpthread` because the former a) is independent of the position in the command line and b) may add other compile definitions as required. Cursory check seems to verify that all used compilers support this flag: clang, gcc, icc, pgcc (at least via patch)

### DIFF
--- a/easybuild/easyblocks/c/cryptography.py
+++ b/easybuild/easyblocks/c/cryptography.py
@@ -1,0 +1,43 @@
+##
+# Copyright 2017-2019 Ghent University
+#
+# This file is part of EasyBuild,
+# originally created by the HPC team of Ghent University (http://ugent.be/hpc/en),
+# with support of Ghent University (http://ugent.be/hpc),
+# the Flemish Supercomputer Centre (VSC) (https://www.vscentrum.be),
+# Flemish Research Foundation (FWO) (http://www.fwo.be/en)
+# and the Department of Economy, Science and Innovation (EWI) (http://www.ewi-vlaanderen.be/en).
+#
+# https://github.com/easybuilders/easybuild
+#
+# EasyBuild is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation v2.
+#
+# EasyBuild is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with EasyBuild.  If not, see <http://www.gnu.org/licenses/>.
+##
+"""
+EasyBuild support for building and installing cryptography, implemented as an easyblock
+
+@author: Alexander Grund
+"""
+from easybuild.easyblocks.generic.pythonpackage import PythonPackage
+
+
+class EB_cryptography(PythonPackage):
+    """Support for building/installing cryptography."""
+
+    def __init__(self, *args, **kwargs):
+        """Initialize cryptography easyblock."""
+        super(EB_cryptography, self).__init__(*args, **kwargs)
+
+        # cryptography compiles a library using pthreads but does not link against it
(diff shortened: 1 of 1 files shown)


Flamefire:
@boegel I created an upstream PR which is likely to be accepted and fix the issue: https://github.com/pyca/cryptography/pull/5086

As this is a blocker for TensorFlow on Power I'd like a decision here:
1. Use this easyblock with version guard due to the huge amount of versions affected
2. Extract a patch for Python 3.7.4 (and any EC using `Paramiko >= 2.5.0`) to allow the sanity check to pass and backporting to older ECs when someone actually uses that part of cryptography on an affected system
3. Add the patch to all ECs using cryptography

boegel:
I don't mind adding a custom easyblock for `cryptography` to resolve this problem in one single place at all, that's exactly what easyblocks are for :+1:

If your patch is indeed fixed upstream we can make this fix conditional so it's only done for older versions.

Why hasn't this popped up for others yet though, that surprises me a bit... The problem doesn't look specific to POWER at all?

easybuild/easyblocks/c/cryptography.py
```diff
@@ -0,0 +1,43 @@
...
+
+class EB_cryptography(PythonPackage):
+    """Support for building/installing cryptography."""
+
+    def __init__(self, *args, **kwargs):
+        """Initialize cryptography easyblock."""
+        super(EB_cryptography, self).__init__(*args, **kwargs)
+
+        # cryptography compiles a library using pthreads but does not link against it
+        # which causes 'undefined symbol: pthread_atfork', see issue #9446
```
boegel: Please include the full URL to https://github.com/easybuilders/easybuild-easyconfigs/issues/9446, and maybe mention https://github.com/pyca/cryptography/issues/5084 too?

Flamefire:
> Why hasn't this popped up for others yet though, that surprises me a bit... The problem doesn't look specific to POWER at all?

See my comment https://github.com/easybuilders/easybuild-easyconfigs/issues/9446#issuecomment-562126078. The TL&DR is: On (e.g.) Power `libpthread` is a linker script which links in a static part of pthread in addition.  It needs to come **after** the library using it.

Additionally this specific extension doesn't seem to be used that often and an import check did not find it until a change in another module exposed it.

> If your patch is indeed fixed upstream we can make this fix conditional so it's only done for older versions.

I'm waiting for it to be merged. Note that my patch is slightly different there: Here I use `-pthread`, there I append `-lpthread`. The former is supposedly independent of the position but maybe not universally supported. But without a patch I don't see how to achieve appending the option in an EB.

easybuild/easyblocks/c/cryptography.py
Flamefire: Done

boegel:
@Flamefire Should we put this on hold until the upstream fix is merged so we can add the version condition?

Flamefire:
I wouldn't wait (to long). This does not conflict with a potentially fixed upstream so a later added version check does not hurt. Up to you.

boegel:
Is this sufficient to make the vanilla `Python-3.7.4-GCCcore-8.3.0.eb` pass on POWER?

If so, can you upload a test report and share it here?

Flamefire:
Yes. Will do on monday

Flamefire:
https://gist.github.com/fc346506bb8444bb55a8a01d4c8bd1d3

Flamefire:
https://github.com/pyca/cryptography/pull/5086 has been merged, so the next release will hopefully contain this fix

boegel:
Tested this with several `Python` old & recent easyconfigs (on x86-64), no problems encountered, so going in...

PR #1875, opened 2019-12-06
change Bazel easyblock to prefer using Java dependency rather than includued JDK (fix for POWER9)

Flamefire:
Bazel fails to install on Power9 because it tried to download and install an (internal?) JDK but does not provide one for PPC.

This fixes the installation by using the "local jdk", supposedly the one in the current environment which is probably even the right thing to do for EB (or Java could be made a build dependency)

Related issues: https://github.com/easybuilders/easybuild-easyconfigs/issues/9450 and https://github.com/bazelbuild/bazel/issues/10377

### DIFF
--- a/easybuild/easyblocks/b/bazel.py
+++ b/easybuild/easyblocks/b/bazel.py
@@ -96,7 +96,15 @@ class EB_Bazel(EasyBlock):
             self.log.info("Not patching Bazel build scripts, installation prefix for binutils/GCC not found")
 
         # enable building in parallel
-        env.setvar('EXTRA_BAZEL_ARGS', '--jobs=%d' % self.cfg['parallel'])
+        bazel_args = '--jobs=%d' % self.cfg['parallel']
+
+        # Bazel provides a JDK by itself for some architectures
+        # We want to enforce it using the JDK we provided via modules
+        # This is required for Power where Bazel does not have a JDK, but requires it for building itself
+        # See https://github.com/bazelbuild/bazel/issues/10377
+        bazel_args += ' --host_javabase=@local_jdk//:jdk'
+
+        env.setvar('EXTRA_BAZEL_ARGS', bazel_args)
 
     def build_step(self):
         """Custom build procedure for Bazel."""


boegel:
This seems like something we should be doing regardless of the processor architecture...

Flamefire:
I'm not fully aware of the implications of that, i.e. I haven't tried building a Java application with a such configured Bazel and don't even know what the option specifically does. Does it bundle the host java with the Bazel binary? Does it reference it? Does one need it to invoke the built Bazel?

Until this question is answered I'd limit the impact and leave it as-is. See also the related Bazel issue.

boegel:
OK, fair enough to err on the side of caution here, so lgtm.

I'll hold off merging this until @branfosj can confirm the fix with a test report in https://github.com/easybuilders/easybuild-easyconfigs/pull/9455

Flamefire:
In the linked Bazel issue someone wrote:

> Bazel built this way will use the system JDK.

He doesn't seem to be a Bazel dev, but rather a TF dev. Anyway I'd use this option for all archs based on this.

boegel:
@Flamefire What do they mean mean with "system JDK" though?
Hardcoded `/usr/bin/java`, or `java` found via `$PATH` (we want the latter, obviously).

Does this only affects the build of Bazel itself, or also the use of Bazel (just to know if we should re-test `TensorFlow` as well to ensure we're not breaking anything if we change this to all archs)?

Flamefire:
From my experiments the `$PATH`-java and yes both use and build (that was my question to him). `TensorFlow` built successfully on our system with that change enabled globally.

boegel:
@Flamefire OK, let's give it a shot, since it's relatively easy to test this (just rebuild all Bazel + TensorFlow modules we have, which is relatively easy to do for us).

Flamefire:
Updated to use this flag for all archs

Flamefire:
@boegel Give it a try on your systems. On ours it does work :)

Flamefire:
> easybuild/easyblocks/b/bazel.py

@Sa1Beer777 excuse me?

boegel:
> > easybuild/easyblocks/b/bazel.py
> 
> @Sa1Beer777 excuse me?

@Flamefire What was that about? :)

boegel:
re-tested installing all `Bazel` easyconfigs (on x86-86), no problems encountered

I'm now also re-testing the installation of a bunch of `TensorFlow` easyconfigs on top, but looking good so far...

boegel:
No problems encountered with reinstalling a bunch of `TensorFlow` easyconfigs on top of a `Bazel` that was installed with this easyblock:

* `TensorFlow-1.8.0-foss-2018a-Python-3.6.4.eb`
* `TensorFlow-1.8.0-intel-2018a-Python-3.6.4.eb`
* `TensorFlow-1.12.0-foss-2018b-Python-3.6.6.eb`
* `TensorFlow-1.13.1-foss-2019a-Python-3.7.2.eb`
* `TensorFlow-1.14.0-foss-2019a-Python-3.7.2.eb`
* `TensorFlow-1.15.0-foss-2019b-Python-3.7.4.eb`
* `TensorFlow-1.15.0-fosscuda-2019b-Python-3.7.4.eb`
* `TensorFlow-2.0.0-foss-2019a-Python-3.7.2.eb`
* `TensorFlow-2.0.0-fosscuda-2019b-Python-3.7.4.eb`

So good to go in, thanks @Flamefire!

PR #1876, opened 2019-12-10
[Python] Remove obsolete configure args, build with optimizations + build with LTO for GCC >= 8.0

Flamefire:
Current Python versions (3.0+ for unicode, 3.7+ for threads) include the asked features and removed the configure options resulting in `configure: WARNING: unrecognized options: --enable-unicode`

This PR removes those for the relevant python versions (checked on source checkouts)

Additionally Python 3.5.3 introduced optimized builds employing PGO and LTO for faster runtimes. Users report 5-30% improvement (see https://github.com/deadsnakes/python3.6/pull/1) As EB is HPC centric using an optimized Python-runtime makes sense to me.

On the downside this adds 1h to the build time on our power machine due to the double build and test runs.

I added options to both enable the optimized build as well as `--with-lto` which does not use profiling to speed up runtime but cross-object-optimizations. That will be easier on build time (expected slight increase) but with less performance benefit. Both can be combined

As a follow-up I'd suggest making output like "WARNING: unrecognized options: " a hard failure to detect changed configure arguments. Imagine this was an important option or e.g. a rename (as happened from `--with-optimizations` to `--enable-optimizations`)

### DIFF
--- a/easybuild/easyblocks/p/python.py
+++ b/easybuild/easyblocks/p/python.py
@@ -109,6 +109,7 @@ class EB_Python(ConfigureMake):
         extra_vars = {
             'ulimit_unlimited': [False, "Ensure stack size limit is set to '%s' during build" % UNLIMITED, CUSTOM],
             'ebpythonprefixes': [True, "Create sitecustomize.py and allow use of $EBPYTHONPREFIXES", CUSTOM],
+            'optimized': [True, "Enable expensive, stable optimizations (PGO, etc)", CUSTOM],
         }
         return ConfigureMake.extra_options(extra_vars)
 
@@ -156,6 +157,10 @@ class EB_Python(ConfigureMake):
                 self.cfg.update('configopts', "--enable-unicode=ucs2")
             else:
                 raise EasyBuildError("Unknown maxunicode value for your python: %d" % sys.maxunicode)
+
+        # Enable further optimizations at the cost of a longer build
+        if self.cfg['optimized'] and LooseVersion(self.version) >= LooseVersion('3.5.3'):
+            self.cfg.update('configopts', "--enable-optimizations")
 
         modules_setup_dist = os.path.join(self.cfg['start_dir'], 'Modules', 'Setup.dist')
 
--- a/easybuild/easyblocks/p/python.py
+++ b/easybuild/easyblocks/p/python.py
@@ -141,15 +141,21 @@ class EB_Python(ConfigureMake):
 
     def configure_step(self):
         """Set extra configure options."""
-        self.cfg.update('configopts', "--with-threads --enable-shared")
+        self.cfg.update('configopts', "--enable-shared")
 
+        # Explicitely enable thread support on < 3.7 (always on 3.7+)
+        if LooseVersion(self.version) < LooseVersion('3.7'):
+            self.cfg.update('configopts', "--with-threads")
+
+        # Explicitely enable unicode on Python 2, always on for Python 3
         # Need to be careful to match the unicode settings to the underlying python
-        if sys.maxunicode == 1114111:
-            self.cfg.update('configopts', "--enable-unicode=ucs4")
-        elif sys.maxunicode == 65535:
-            self.cfg.update('configopts', "--enable-unicode=ucs2")
-        else:
-            raise EasyBuildError("Unknown maxunicode value for your python: %d" % sys.maxunicode)
+        if LooseVersion(self.version) < LooseVersion('3.0'):
+            if sys.maxunicode == 1114111:
+                self.cfg.update('configopts', "--enable-unicode=ucs4")
+            elif sys.maxunicode == 65535:
+                self.cfg.update('configopts', "--enable-unicode=ucs2")
+            else:
+                raise EasyBuildError("Unknown maxunicode value for your python: %d" % sys.maxunicode)
 
         modules_setup_dist = os.path.join(self.cfg['start_dir'], 'Modules', 'Setup.dist')
 
--- a/easybuild/easyblocks/p/python.py
+++ b/easybuild/easyblocks/p/python.py
@@ -49,6 +49,7 @@ from easybuild.tools.modules import get_software_libdir, get_software_root, get_
 from easybuild.tools.filetools import symlink, write_file
 from easybuild.tools.run import run_cmd
 from easybuild.tools.systemtools import get_shared_lib_ext
+import easybuild.tools.toolchain as toolchain
 
 
 EXTS_FILTER_PYTHON_PACKAGES = ('python -c "import %(ext_name)s"', "")
@@ -109,7 +110,11 @@ class EB_Python(ConfigureMake):
         extra_vars = {
             'ulimit_unlimited': [False, "Ensure stack size limit is set to '%s' during build" % UNLIMITED, CUSTOM],
             'ebpythonprefixes': [True, "Create sitecustomize.py and allow use of $EBPYTHONPREFIXES", CUSTOM],
-            'use_lto': [False, "Build with Link Time Optimization. Potentially unstable on some toolchains", CUSTOM],
+            'use_lto': [
+                None,
+                "Build with Link Time Optimization. Potentially unstable on some toolchains. Default: Auto-detect",
+                CUSTOM
+                ],
             'optimized': [True, "Enable expensive, stable optimizations (PGO, etc)", CUSTOM],
         }
         return ConfigureMake.extra_options(extra_vars)
@@ -141,6 +146,17 @@ class EB_Python(ConfigureMake):
                 self.log.debug(msg, param, self.cfg[param])
             self.cfg[param] = ''
 
+    def auto_detect_lto_support(self):
+        """Return True, if LTO should be enabled for current toolchain"""
+        result = False
+        # GCC >= 8 should be stable enough for LTO
+        if self.toolchain.comp_family() == toolchain.GCC:
+            gcc_ver = get_software_version('GCCcore') or get_software_version('GCC')
+            if gcc_ver and LooseVersion(gcc_ver) >= LooseVersion('8.0'):
+                self.log.info("Auto-enabling LTO since GCC >= v8.0 is used as toolchain compiler")
+                result = True
+        return result
+
     def configure_step(self):
         """Set extra configure options."""
         self.cfg.update('configopts', "--enable-shared")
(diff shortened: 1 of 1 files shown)


bartoldeman:
Good point about `WARNING: unrecognized options`, this comes up from time to time (seen it with OpenMPI as well)

Flamefire:
Seems it already exists since v1: https://github.com/easybuilders/easybuild-easyblocks/issues/157

easybuild/easyblocks/p/python.py
```diff
@@ -109,6 +109,8 @@ def extra_options():
         extra_vars = {
             'ulimit_unlimited': [False, "Ensure stack size limit is set to '%s' during build" % UNLIMITED, CUSTOM],
             'ebpythonprefixes': [True, "Create sitecustomize.py and allow use of $EBPYTHONPREFIXES", CUSTOM],
+            'use_lto': [False, "Build with Link Time Optimization. Potentially unstable on some toolchains", CUSTOM],
```
boegel: @Flamefire Using `use_lto = True` doesn't do anything at all right now?

Also, can you clarify the "potentially unstable on some toolchains"?

Is it worth using `None` as a default here, and auto-enabling LTO for blessed toolchains (e.g. when a sufficiently recent version of GCC or Intel compiler is used as toolchain compiler)?

easybuild/easyblocks/p/python.py
```diff
@@ -141,15 +143,25 @@ def prepare_for_extensions(self):
...
-            raise EasyBuildError("Unknown maxunicode value for your python: %d" % sys.maxunicode)
+        if LooseVersion(self.version) < LooseVersion('3.0'):
+            if sys.maxunicode == 1114111:
+                self.cfg.update('configopts', "--enable-unicode=ucs4")
+            elif sys.maxunicode == 65535:
+                self.cfg.update('configopts', "--enable-unicode=ucs2")
+            else:
+                raise EasyBuildError("Unknown maxunicode value for your python: %d" % sys.maxunicode)
+
+        # Enable further optimizations at the cost of a longer build
```
boegel: Maybe clarify the comment to explain the version check >= 3.5.3?

Bonus points for a reference to the relevant section in the Python documentation...

boegel:
FWIW: version checks look OK to me based on this:

```
$ ls -d /software/Python/*
/software/Python/2.7.12-foss-2016b
/software/Python/2.7.13-foss-2017a
/software/Python/2.7.14-foss-2018a
/software/Python/2.7.14-GCCcore-6.4.0-bare
/software/Python/2.7.14-intel-2018a
/software/Python/2.7.15-foss-2018b
/software/Python/2.7.15-foss-2019b
/software/Python/2.7.15-fosscuda-2018b
/software/Python/2.7.15-GCCcore-7.3.0-bare
/software/Python/2.7.15-GCCcore-8.2.0
/software/Python/2.7.15-GCCcore-8.3.0-bare
/software/Python/2.7.16-GCCcore-8.3.0
/software/Python/3.5.2-foss-2016b
/software/Python/3.6.4-foss-2018a
/software/Python/3.6.6-foss-2018b
/software/Python/3.6.6-foss-2019b
/software/Python/3.6.6-fosscuda-2018b
/software/Python/3.7.0-foss-2018b
/software/Python/3.7.0-foss-2019b
/software/Python/3.7.2-GCCcore-8.2.0
/software/Python/3.7.4-GCCcore-8.3.0
```

```
$ bzgrep "configure: WARNING: unrecognized options" /software/Python/*/easybuild/*log*
/software/Python/3.5.2-foss-2016b/easybuild/easybuild-Python-3.5.2-20191119.123533.log.bz2:configure: WARNING: unrecognized options: --enable-unicode
/software/Python/3.5.2-foss-2016b/easybuild/easybuild-Python-3.5.2-20191119.123533.log.bz2:configure: WARNING: unrecognized options: --enable-unicode
/software/Python/3.6.4-foss-2018a/easybuild/easybuild-Python-3.6.4-20191104.200920.log.bz2:configure: WARNING: unrecognized options: --enable-unicode
/software/Python/3.6.4-foss-2018a/easybuild/easybuild-Python-3.6.4-20191104.200920.log.bz2:configure: WARNING: unrecognized options: --enable-unicode
/software/Python/3.6.6-foss-2018b/easybuild/easybuild-Python-3.6.6-20190828.231037.log.bz2:configure: WARNING: unrecognized options: --enable-unicode
/software/Python/3.6.6-foss-2018b/easybuild/easybuild-Python-3.6.6-20190828.231037.log.bz2:configure: WARNING: unrecognized options: --enable-unicode
/software/Python/3.6.6-foss-2019b/easybuild/easybuild-Python-3.6.6-20190912.165329.log.bz2:configure: WARNING: unrecognized options: --enable-unicode
/software/Python/3.6.6-foss-2019b/easybuild/easybuild-Python-3.6.6-20190912.165329.log.bz2:configure: WARNING: unrecognized options: --enable-unicode
/software/Python/3.7.0-foss-2018b/easybuild/easybuild-Python-3.7.0-20190828.195437.log.bz2:configure: WARNING: unrecognized options: --with-threads, --enable-unicode
/software/Python/3.7.0-foss-2018b/easybuild/easybuild-Python-3.7.0-20190828.195437.log.bz2:configure: WARNING: unrecognized options: --with-threads, --enable-unicode
/software/Python/3.7.0-foss-2019b/easybuild/easybuild-Python-3.7.0-20190914.112033.log.bz2:configure: WARNING: unrecognized options: --with-threads, --enable-unicode
/software/Python/3.7.0-foss-2019b/easybuild/easybuild-Python-3.7.0-20190914.112033.log.bz2:configure: WARNING: unrecognized options: --with-threads, --enable-unicode
/software/Python/3.7.2-GCCcore-8.2.0/easybuild/easybuild-Python-3.7.2-20191121.112113.log.bz2:configure: WARNING: unrecognized options: --with-threads, --enable-unicode
/software/Python/3.7.2-GCCcore-8.2.0/easybuild/easybuild-Python-3.7.2-20191121.112113.log.bz2:configure: WARNING: unrecognized options: --with-threads, --enable-unicode
/software/Python/3.7.4-GCCcore-8.3.0/easybuild/easybuild-Python-3.7.4-20191121.113430.log.bz2:configure: WARNING: unrecognized options: --with-threads, --enable-unicode
/software/Python/3.7.4-GCCcore-8.3.0/easybuild/easybuild-Python-3.7.4-20191121.113430.log.bz2:configure: WARNING: unrecognized options: --with-threads, --enable-unicode
```

easybuild/easyblocks/p/python.py
Flamefire: Ups, rebase conflict gone wrong. Thanks

From https://docs.python.org/3.7/whatsnew/changelog.html#python-3-7-5-final

> A –with-lto configure option has been added that will enable link time optimizations at build time during a make profile-opt. Some compilers and toolchains are known to not produce stable code when using LTO, be sure to test things thoroughly before relying on it. It can provide a few % speed up over profile-opt alone.

The suggestion with `None`==auto-detect is good. https://bugs.python.org/issue25702 seems to state that GCC>=4.8 is fine and apparently "Ubuntu" might have same kind of whitelist or blacklist when to enable this for Python. Not sure where to look for that though.

easybuild/easyblocks/p/python.py
Flamefire: https://docs.python.org/3.7/whatsnew/changelog.html#python-3-5-3-release-candidate-1

> bpo-26359: Rename –with-optimiations to –enable-optimizations.

This seems to be where it was added initially, see the linked issue in the changelog. I also downloaded a couple python releases (source) and grepped for this option and it seems to consistently appear after 3.5.3

easybuild/easyblocks/p/python.py
Flamefire: https://bugs.python.org/issue28032 indicates GCC 5.4+

boegel: I guess the question is how to determine if the resulting build is stable, and how to "test things thoroughly".

If it's only a few %, I'm not sure it's worth taking the risk to auto-enable it?

easybuild/easyblocks/p/python.py
boegel: OK, just clarify that a bit in the comment then?

easybuild/easyblocks/p/python.py
Flamefire: The Python devs seem to be quite confident and if Ubuntu does it... But maybe we can at least do it in the ECS where we know the compilers, so no surprises

boegel: How about this:

* use `None` as default, so people can opt-in (`True`) or opt-out (`False`) in the `Python` easyconfig easily
* auto-enable if `None` is used, but only if a sufficiently recent `GCC` is used as compiler (probably not worth the auto-enable when Intel compilers are used in toolchain for installing `Python`, so we pulled down `Python` to using `GCCcore` as toolchain a while ago)

Flamefire: Sure. What would you deem "sufficiently recent GCC"? GCC 6+? 8+?
Do you have a snipped how to check for the used compiler and version?

boegel: I'd prefer being conservative here, so why not go with GCC 8+.

Should be something like this:

```python
if self.cfg['uselto'] is None and self.toolchain.comp_family() == toolchain.GCC:
    gcc_ver = get_software_version('GCCcore') or get_software_version('GCC')
    if LooseVersion(gcc_ver) >= LooseVersion('8.0'):
        self.log.info("Auto-enabling --with-lto since GCC >= v8.0 is used as toolchain compiler")
        self.cfg['uselto']  = True

Comment (Flamefire) [reply easybuild/easyblocks/p/python.py]: Ok. Slightly adapted and removed `get_software_version('GCCcore')`. If I got it right the if-check already ensures we have `GCC`

File: easybuild/easyblocks/p/python.py (hunk 2 above)
Comment (Flamefire) [reply easybuild/easyblocks/p/python.py]: ok

File: easybuild/easyblocks/p/python.py (hunk 1 above)
Comment (Flamefire) [reply easybuild/easyblocks/p/python.py]: Oh, it seems GCCcore is needed because it does not load the GCC module. This is an annoying pitfall and I'm not the only one:

`easybuild-easyblocks/easybuild/easyblocks/l/libsmm.py:            gccVersion = LooseVersion(get_software_version('GCC'))`

There are probably more... Maybe make `get_software_version('GCC`) also check `GCCcore'?

File: easybuild/easyblocks/p/python.py [hunk 3]
```diff
@@ -109,6 +110,12 @@ def extra_options():
         extra_vars = {
             'ulimit_unlimited': [False, "Ensure stack size limit is set to '%s' during build" % UNLIMITED, CUSTOM],
             'ebpythonprefixes': [True, "Create sitecustomize.py and allow use of $EBPYTHONPREFIXES", CUSTOM],
+            'use_lto': [
+                None,
+                "Build with Link Time Optimization. Potentially unstable on some toolchains. Default: Auto-detect",
+                CUSTOM
+                ],
+            'optimized': [True, "Enable expensive, stable optimizations (PGO, etc)", CUSTOM],
```
boegel: @Flamefire We should keep these ordered alphabetically, and I think they can be slightly clarified too, see https://github.com/Flamefire/easybuild-easyblocks/pull/1

boegel:
Tested by rebuilding various `Python` 3.6 & 3.7 + 2.7.16 versions with this updated easyblock, and various Python packages on top (`SciPy-bundle`, `TensorFlow`).

No problems encounered, so good to go... Thanks @Flamefire!

PR #1879, opened 2019-12-11
Remove optarch warning in GORMACS @ CRAY

victorusu:
Users were getting confused about the warning
"No target architecture specified based on OPTARCH configuration option 'HASWELL'"

So, I am proposing to remove it and let GROMACS auto-detect for the
Cray.
We would also make the mapping for all Cray supported architectures, but this is
not a trivial quest...


ocaisa:
In CSCS I trust...I can't test this but the code change looks fine for me.

PR #1882, opened 2019-12-13
disable running of 'sudo apt-get update' in GitHub CI config, since it's failing (and we don't really need it)

boegel:
(see also https://github.com/easybuilders/easybuild-easyconfigs/pull/9492)


akesandgren:
Going in, thanks @boegel!

PR #1883, opened 2019-12-13
add easyblock for ELSI

migueldiascosta:
there are more libraries besides ELPA that can be specified as external instead of built, and other optional libraries, see "Solver Libraries and External Dependencies" in https://wordpress.elsi-interchange.org/index.php/download/, not sure if there is demand for those

Motivation was its use in the new "MaX release version" of SIESTA (PR coming soon)

### DIFF
--- a/easybuild/easyblocks/e/elsi.py
+++ b/easybuild/easyblocks/e/elsi.py
@@ -48,7 +48,7 @@ class EB_ELSI(CMakeMake):
     def extra_options():
         """Define custom easyconfig parameters for ELSI."""
         extra_vars = {
-            'build_internal_pexsi': [False, "Build internal PEXSI solver", CUSTOM],
+            'build_internal_pexsi': [True, "Build internal PEXSI solver", CUSTOM],
         }
         return CMakeMake.extra_options(extra_vars)
 
--- a/easybuild/easyblocks/e/elsi.py
+++ b/easybuild/easyblocks/e/elsi.py
@@ -49,7 +49,7 @@ class EB_ELSI(CMakeMake):
     def extra_options():
         """Define custom easyconfig parameters for ELSI."""
         extra_vars = {
-            'build_internal_pexsi': [True, "Build internal PEXSI solver", CUSTOM],
+            'build_internal_pexsi': [None, "Build internal PEXSI solver", CUSTOM],
             'build_shared_libs': [True, "Build shared libraries instead of static", CUSTOM],
         }
         return CMakeMake.extra_options(extra_vars)
@@ -85,17 +85,23 @@ class EB_ELSI(CMakeMake):
             self.log.info("No external ELPA specified as dependency, building internal ELPA.")
 
         pexsi = get_software_root('PEXSI')
-        if pexsi and self.cfg['build_internal_pexsi']:
-            raise EasyBuildError("Both build_internal_pexsi and external PEXSI dependency found, only one can be set.")
+        if pexsi:
+            if self.cfg['build_internal_pexsi']:
+                raise EasyBuildError("Both build_internal_pexsi and external PEXSI dependency found, "
+                                     "only one can be set.")
+            self.log.info("Using external PEXSI.")
+            self.cfg.update('configopts', "-DUSE_EXTERNAL_PEXSI=ON")
+            external_libs.append('pexsi')
+        elif self.cfg['build_internal_pexsi'] is not False:
+            self.log.info("No external PEXSI specified as dependency and internal PEXSI not explicitly disabled, "
+                          "building internal PEXSI.")
+            self.cfg['build_internal_pexsi'] = True
+        else:
+            self.log.info("No external PEXSI specified as dependency and internal PEXSI was explicitly disabled, "
+                          "building ELSI without PEXSI.")
+
         if pexsi or self.cfg['build_internal_pexsi']:
-            self.log.info("Enabling PEXSI solver.")
             self.cfg.update('configopts', "-DENABLE_PEXSI=ON")
-            if pexsi:
-                self.log.info("Using external PEXSI.")
-                self.cfg.update('configopts', "-DUSE_EXTERNAL_PEXSI=ON")
-                external_libs.append('pexsi')
-            else:
-                self.log.info("No external PEXSI specified as dependency, building internal PEXSI.")
 
         slepc = get_software_root('SLEPc')
         if slepc:
--- a/easybuild/easyblocks/e/elsi.py
+++ b/easybuild/easyblocks/e/elsi.py
@@ -49,6 +49,7 @@ class EB_ELSI(CMakeMake):
         """Define custom easyconfig parameters for ELSI."""
         extra_vars = {
             'build_internal_pexsi': [True, "Build internal PEXSI solver", CUSTOM],
+            'build_shared_libs': [True, "Build shared libraries instead of static", CUSTOM],
         }
         return CMakeMake.extra_options(extra_vars)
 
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/e/elsi.py
```diff
@@ -0,0 +1,94 @@
...
+
+class EB_ELSI(CMakeMake):
+    """Support for building ELSI."""
+
+    @staticmethod
+    def extra_options():
+        """Define custom easyconfig parameters for ELSI."""
+
+        extra_vars = {
+            'enable_pexsi': [False, "Enable PEXSI solver", CUSTOM],
```
boegel: Any reason not to enable it by default?

easybuild/easyblocks/e/elsi.py
```diff
@@ -0,0 +1,94 @@
...
+
+        if self.cfg['enable_pexsi']:
+            self.cfg.update('configopts', "-DENABLE_PEXSI=1")
+
+        if self.cfg['runtest']:
+            self.cfg.update('configopts', "-DENABLE_TESTS=1")
+            self.cfg.update('configopts', "-DENABLE_C_TESTS=1")
+            self.cfg['runtest'] = 'test'
+
+        libs = [lib[3:-2] for lib in os.environ['SCALAPACK_STATIC_LIBS'].split(',')]
```
boegel: What's the `3:-2` about? At least clarify with a comment? Does this work as expected for multiple toolchains?

easybuild/easyblocks/e/elsi.py
```diff
@@ -0,0 +1,94 @@
...
+
+        elpa = get_software_root('ELPA')
+        if elpa:
+            self.log.info("Using external ELPA.")
+            elpa_ver = get_software_version('ELPA')
+            self.cfg.update('configopts', "-DUSE_EXTERNAL_ELPA=1")
+            self.cfg.update('configopts', "-DINC_PATHS='%s/include/elpa-%s/modules'" % (elpa, elpa_ver))
+            self.cfg.update('configopts', "-DLIB_PATHS='%s/lib'" % elpa)
+            libs = ['elpa'] + libs
```
boegel:
```python
libs.insert(0, 'elpa')

File: easybuild/easyblocks/e/elsi.py (hunk 1 above)
Comment (migueldiascosta) [reply easybuild/easyblocks/e/elsi.py]: not that I know of (other than false being the default in ELSI)

File: easybuild/easyblocks/e/elsi.py (hunk 2 above)
Comment (migueldiascosta) [reply easybuild/easyblocks/e/elsi.py]: I think the `...STATIC_LIBS...` environment variables are always a comma separated list of static library filenames, this is just a way of converting it to the format expected by ELSI (semi-colon separated and without `lib` and '.a')

I suppose something like `re.findall('lib(.*?)\.a', os.environ['SCALAPACK_STATIC_LIBS']` would be clearer?

File: easybuild/easyblocks/e/elsi.py [hunk 4]
```diff
@@ -0,0 +1,151 @@
...
+        """Initialize ELSI-specific variables."""
+        super(EB_ELSI, self).__init__(*args, **kwargs)
+        self.enable_sips = False
+        self.env_suff = '_MT' if self.toolchain.options.get('openmp', None) else ''
+
+    @staticmethod
+    def extra_options():
+        """Define custom easyconfig parameters for ELSI."""
+        extra_vars = {
+            'build_internal_pexsi': [True, "Build internal PEXSI solver", CUSTOM],
```
akesandgren: I wonder if it would be more appropriate to have this as False or None.
Our default modus operandi is to rely on explicit dependencies, not internal ones.

easybuild/easyblocks/e/elsi.py
```diff
@@ -0,0 +1,151 @@
...
+            external_libs.extend(['slepc', 'petsc', 'HYPRE', 'umfpack', 'klu', 'cholmod', 'btf', 'ccolamd', 'colamd',
+                                  'camd', 'amd', 'suitesparseconfig', 'metis', 'ptesmumps',
+                                  'ptscotchparmetis', 'ptscotch', 'ptscotcherr', 'esmumps', 'scotch', 'scotcherr',
+                                  'stdc++', 'dl'])
+            if get_software_root('imkl') or get_software_root('FFTW'):
+                external_libs.extend(re.findall(r'lib(.*?)\.a', os.environ['FFTW_STATIC_LIBS%s' % self.env_suff]))
+            else:
+                raise EasyBuildError("Could not find FFTW library or interface.")
+
+        if get_software_root('imkl') or get_software_root('SCALAPACK'):
```
akesandgren: It's ScaLAPACK

easybuild/easyblocks/e/elsi.py
```diff
@@ -0,0 +1,151 @@
...
+            external_libs.extend(re.findall(r'lib(.*?)\.a', os.environ['SCALAPACK%s_STATIC_LIBS' % self.env_suff]))
+        else:
+            raise EasyBuildError("Could not find SCALAPACK library or interface.")
```
akesandgren: Same here, ScaLAPACK

easybuild/easyblocks/e/elsi.py
```diff
@@ -0,0 +1,151 @@
...
+        if self.cfg['build_internal_pexsi']:
+            modules.append('elsi_pexsi')
+            libs.extend(['pexsi', 'ptscotch', 'ptscotcherr', 'ptscotchparmetis',
+                         'scotch', 'scotcherr', 'scotchmetis', 'superlu_dist'])
+
+        if self.enable_sips:
+            modules.append('elsi_sips')
+            libs.append('sips')
+
+        libext = '.so' if self.cfg['build_shared_libs'] else '.a'
```
akesandgren: Use get_shared_lib_ext, from easybuild.tools.systemtools instead of hardcoding .so

easybuild/easyblocks/e/elsi.py
```diff
@@ -0,0 +1,151 @@
...
+
+        custom_paths = {
+            'files': ['include/%s.mod' % mod for mod in modules] + ['lib/lib%s%s' % (lib, libext) for lib in libs],
```
akesandgren: Use os.path.join

easybuild/easyblocks/e/elsi.py
```diff
@@ -0,0 +1,151 @@
...
+        external_libs = []
+        inc_paths = os.environ['CMAKE_INCLUDE_PATH'].split(':')
+        lib_paths = os.environ['CMAKE_LIBRARY_PATH'].split(':')
+
+        elpa_root = get_software_root('ELPA')
+        if elpa_root:
+            self.log.info("Using external ELPA.")
+            self.cfg.update('configopts', "-DUSE_EXTERNAL_ELPA=ON")
+            elpa_lib = 'elpa_openmp' if self.toolchain.options.get('openmp', None) else 'elpa'
+            inc_paths.append('%s/include/%s-%s/modules' % (elpa_root, elpa_lib, get_software_version('ELPA')))
```
akesandgren: Use os.path.join instead of "/"

easybuild/easyblocks/e/elsi.py
migueldiascosta: you're probably right... (but see above, https://github.com/easybuilders/easybuild-easyblocks/pull/1883#discussion_r357669460)

migueldiascosta: I'm fine either way

akesandgren: Yes, but that comment was about enabling pexsi, not about using the internal pexsi.

I'd say building_internal_pexsi should not be the default.
I.e. set it to None, and if there is no external pexsi, then use internal unless explicitly set to false in easyconfig file.
And complain if explicitly set to true with external pexsi dep defined.

migueldiascosta: Ah, you're right, will do

akesandgren:
Going in, thanks @migueldiascosta!

PR #1884, opened 2019-12-13
limit MPI ranks used for running WRF test cases to max. 4 + include contents of rsl.error.0000 output file in case test failed

boegel:
I ran into problems with running the WRF 4.0.1 test cases (using the easyconfigs in https://github.com/easybuilders/easybuild-easyconfigs/pull/7085) when too many cores were available.
Some tests were crashing because too many MPI ranks were used, and the domain size was too small...

It took me a while to realize why the tests were failing, because they were just exiting with exit code `1`, not producing any output, so I've also enhanced the test step to read the `rsl.error.0000` output file (from MPI rank 0) and include the output in the error message that is produced when the test failed to run.

### DIFF
--- a/easybuild/easyblocks/w/wrf.py
+++ b/easybuild/easyblocks/w/wrf.py
@@ -295,34 +295,46 @@ class EB_WRF(EasyBlock):
                     if test in self.testcases:
                         self.testcases.remove(test)
 
-            # determine parallel setting (1/2 of available processors + 1)
-            n = self.cfg['parallel'] / 2 + 1
+            # determine number of MPI ranks to use in tests (1/2 of available processors + 1);
+            # we need to limit max number of MPI ranks (8 is too high for some tests, 4 is OK),
+            # since otherwise run may fail because domain size is too small
+            n_mpi_ranks = min(self.cfg['parallel'] / 2 + 1, 4)
 
             # prepare run command
 
             # stack limit needs to be set to unlimited for WRF to work well
             if self.cfg['buildtype'] in self.parallel_build_types:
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/w/wrf.py
```diff
@@ -295,34 +295,46 @@ def test_step(self):
...
             # stack limit needs to be set to unlimited for WRF to work well
             if self.cfg['buildtype'] in self.parallel_build_types:
                 test_cmd = "ulimit -s unlimited && %s && %s" % (self.toolchain.mpi_cmd_for("./ideal.exe", 1),
-                                                                self.toolchain.mpi_cmd_for("./wrf.exe", n))
+                                                                self.toolchain.mpi_cmd_for("./wrf.exe", n_mpi_ranks))
             else:
                 test_cmd = "ulimit -s unlimited && ./ideal.exe && ./wrf.exe >rsl.error.0000 2>&1"
 
+            # regex to check for successful test run
+            re_success = re.compile("SUCCESS COMPLETE WRF")
```
akesandgren: Why move this outside of run_test ??

boegel: Because there's no point in recompiling the same regex over and over again :)

akesandgren: Thought that would be it, but doesn't it have to be in self. then??
Still don't quite understand Python...

akesandgren: Ah, should have looked at the whole file...
run_test is inside test_step, that explains it.

boegel: `run_test` is an inner method, sort of like a local variable in the `test_step` method (but it's a function, not a variable), so it has access to all local variables in `test_step`

akesandgren:
Going in, thanks @boegel!

PR #1885, opened 2019-12-13
fix linking of Hypre to external BLAS/LAPACK + support building (only) static/shared library

Micket:
Old code was incorrect *and* also ignored by Hypre (only supports external blas/lapack when building shared library).

Using the flags were limited, and still, looked like hypre still built it's own lapack/blas stuff, so this needed patching either way, and it was a lot simpler to make the patch just pick up EB's environment variables directly instead of going through the broken configure scripts.

### DIFF
--- a/easybuild/easyblocks/h/hypre.py
+++ b/easybuild/easyblocks/h/hypre.py
@@ -82,7 +82,7 @@ class EB_Hypre(ConfigureMake):
 
         custom_paths = {
             'files': hypre_libs,
-            'dirs':['include']
+            'dirs': ['include']
         }
 
         super(EB_Hypre, self).sanity_check_step(custom_paths=custom_paths)
--- a/easybuild/easyblocks/h/hypre.py
+++ b/easybuild/easyblocks/h/hypre.py
@@ -46,7 +46,7 @@ class EB_Hypre(ConfigureMake):
         self.cfg.update('configopts', '--enable-shared')
 
         # While there are a --with-{blas|lapack}-libs flag, it's not useable, because of how Hypre treats it.
-        # We need to patch the code anyway to prevent it from building it's own BLAS packages.
+        # We need to patch the code anyway to prevent it from building its own BLAS packages.
 
         super(EB_Hypre, self).configure_step()
 
--- a/easybuild/easyblocks/h/hypre.py
+++ b/easybuild/easyblocks/h/hypre.py
@@ -26,10 +26,12 @@
 EasyBuild support for Hypre, implemented as an easyblock
 
 @author: Kenneth Hoste (Ghent University)
+@author: Mikael OEhman (Chalmers University of Technology)
 """
 import os
 
 from easybuild.easyblocks.generic.configuremake import ConfigureMake
+from easybuild.tools.systemtools import get_shared_lib_ext
 
 
 class EB_Hypre(ConfigureMake):
@@ -40,11 +42,11 @@ class EB_Hypre(ConfigureMake):
 
         self.cfg.update('configopts', '--with-MPI-include=%s' % os.getenv('MPI_INC_DIR'))
 
-        for dep in ["BLAS", "LAPACK"]:
-            libs = ' '.join(os.getenv('%s_STATIC_LIBS' % dep).split(','))
-            self.cfg.update('configopts', '--with-%s-libs="%s"' % (dep.lower(), libs))
-            self.cfg.update('configopts', '--with-%s-lib-dirs="%s"' % (dep.lower(),
-                                                                      os.getenv('%s_LIB_DIR' % dep)))
+        # Only supports external libraries when building a shared library.
+        self.cfg.update('configopts', '--enable-shared')
+
+        # While there are a --with-{blas|lapack}-libs flag, it's not useable, because of how Hypre treats it.
+        # We need to patch the code anyway to prevent it from building it's own blas packages. 
 
         super(EB_Hypre, self).configure_step()
 
@@ -52,7 +54,7 @@ class EB_Hypre(ConfigureMake):
         """Custom sanity check for Hypre."""
 
         custom_paths = {
-                        'files':['lib/libHYPRE.a'],
+                        'files':['lib/libHYPRE.' + get_shared_lib_ext()],
                         'dirs':['include']
                        }
 
--- a/easybuild/easyblocks/h/hypre.py
+++ b/easybuild/easyblocks/h/hypre.py
@@ -46,7 +46,7 @@ class EB_Hypre(ConfigureMake):
         self.cfg.update('configopts', '--enable-shared')
 
         # While there are a --with-{blas|lapack}-libs flag, it's not useable, because of how Hypre treats it.
-        # We need to patch the code anyway to prevent it from building it's own blas packages. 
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/h/hypre.py
```diff
@@ -40,20 +42,20 @@ def configure_step(self):
 
         self.cfg.update('configopts', '--with-MPI-include=%s' % os.getenv('MPI_INC_DIR'))
 
-        for dep in ["BLAS", "LAPACK"]:
-            libs = ' '.join(os.getenv('%s_STATIC_LIBS' % dep).split(','))
-            self.cfg.update('configopts', '--with-%s-libs="%s"' % (dep.lower(), libs))
-            self.cfg.update('configopts', '--with-%s-lib-dirs="%s"' % (dep.lower(),
-                                                                      os.getenv('%s_LIB_DIR' % dep)))
+        # Only supports external libraries when building a shared library.
+        self.cfg.update('configopts', '--enable-shared')
```
smoors: @Micket does this mean that the static lib not is no longer built? if so, can't we make sure both are built (even though the static lib does not support external libs) ?

easybuild/easyblocks/h/hypre.py
```diff
@@ -40,20 +42,20 @@ def configure_step(self):
...
+
+        # While there are a --with-{blas|lapack}-libs flag, it's not useable, because of how Hypre treats it.
+        # We need to patch the code anyway to prevent it from building it's own blas packages.
```
smoors: typo: building its own BLAS packages

easybuild/easyblocks/h/hypre.py
Micket: I actually intentionally excluded this to make sure we didn't get a gimped version.

boegel: The only easyconfigs we have that depend on `Hypre` are `PETSc`, so easy to test if the static `Hypre` lib is actually required...

easybuild/easyblocks/h/hypre.py
boegel: @Micket ping on fixing the typo?

akesandgren:
Was there some other thread on PETSc/Hypre somewhere? Or is this related to that thread?

Anyway, what's the status on this?
@Micket @boegel @smoors

Micket:
I don't recall any thread.
I think this is good to go. The previous way is definitely not working correctly (the buildsystem they wrote just can't link to external libraries when creating a static library (without a fair bit of patching)).

akesandgren:
@boegel Ok with the typo changes?

easybuild/easyblocks/h/hypre.py
boegel: @Micket Typo is still there: `it's` should be `its` ;)

lexming:
The static libs of Hypre do use symbols from external BLAS/LAPACK if configured to do so. As an example, this is the table in `libHYPRE.a` for `FGmres.o` where LAPACK symbols such as `daxpy` differ
* left: compiled removing `--with-blas-libs` and friends
* right: default compilation in EB using `Hypre-2.18.2-foss-2020a.eb` in current `develop`
```
FGmres.o:                                                                       FGmres.o:
                                                                              >                  U daxpy_
                                                                              >                  U dcopy_
                                                                              >                  U ddot_
                                                                              >                  U dscal_
0000000000000070 T FGMRES_ParaSails                                             0000000000000070 T FGMRES_ParaSails
                 U free                                                                          U free
                 U _GLOBAL_OFFSET_TABLE_                                                         U _GLOBAL_OFFSET_TABLE_
                 U hypre_daxpy                                                <
                 U hypre_dcopy                                                <
                 U hypre_ddot                                                 <
                 U hypre_dscal                                                <
                 U hypre_MAlloc                                                                  U hypre_MAlloc
                 U hypre_MPI_Allreduce                                                           U hypre_MPI_Allreduce
                 U hypre_MPI_Comm_rank                                                           U hypre_MPI_Comm_rank
                 U hypre_printf                                                                  U hypre_printf
0000000000000000 t InnerProd                                                    0000000000000000 t InnerProd
0000000000000000 r .LC0                                                         0000000000000000 r .LC0
0000000000000008 r .LC2                                                         0000000000000008 r .LC2
0000000000000000 r .LC3                                                         0000000000000000 r .LC3
0000000000000010 r .LC4                                                         0000000000000010 r .LC4
0000000000000000 r .LC5                                                         0000000000000000 r .LC5
                 U MatrixMatvec                                                                  U MatrixMatvec
                 U ompi_mpi_double                                                               U ompi_mpi_double
                 U ompi_mpi_op_sum                                                               U ompi_mpi_op_sum
                 U ParaSailsApply                                                                U ParaSailsApply
```

lexming:
According to the explanation given by the devs in https://github.com/hypre-space/hypre/issues/41, Hypre builds in all cases the internal BLAS/LAPACK and it is only in the final link step that it re-links the libraries to external ones if configured to do so. This fits with my experience in https://github.com/easybuilders/easybuild-easyconfigs/pull/10132

@Micket  I updated this easyblock to support builds with static and shared libs and use external BLAS/LAPACK in both cases. I made a PR to your branch in https://github.com/Micket/easybuild-easyblocks/pull/3 . Please test it and let me know if it works.

Micket:
I did some more digging; and, yeah, Hypre's build system sure is misleading.
If you build statically, you can actually put any gibberish you want in the `--with-blas-libs=xxx ` flags, since they aren't actually used beyond setting a define flag that adds their hypre_ prefix

```
Building libHYPRE ...
ar -rcu libHYPRE.a /home/c3-micke/hypre-2.15.1/src/FEI_mv/fei-hypre/*.o  /home/c3-micke/hypre-2.15.1/src/IJ_mv/*.o
...
ar -rcu libHYPRE.a /home/c3-micke/hypre-2.15.1/src/utilities/*.o /home/c3-micke/hypre-2.15.1/src/blas/*.o /home/c3-micke/hypre-2.15.1/src/lapack/*.o
ranlib libHYPRE.a
```
it just packs in their `hypre_dgemm` versions, and leaves `dgemm_` undefined.

But, then you read the code and see uses of their "hypre_dgemm" being called everwhere.. but, it was just a trick! Because then they have these 4 macros, conditionally set and used:
```
#define dgemm_   hypre_dgemm
#define hypre_dgemm   hypre_F90_NAME_BLAS(dgemm ,DGEMM )
#define hypre_F90_NAME(name,NAME) hypre_F90_NAME_2(name,NAME)
#define hypre_F90_NAME_2(name,NAME) name##_
```
Sheeeesh

boegel:
@lexming @Micket So, what the best way forward with this?

Should https://github.com/Micket/easybuild-easyblocks/pull/3 get merged first?

And when that's done, should this then be good to go?

lexming:
@boegel in my opinion we should just let Hypre perform its convoluted mess (as described by @Micket). Its library might be bloated with a non-used BLAS/LAPACK but as long as it uses the external libs is not an issue. PR https://github.com/Micket/easybuild-easyblocks/pull/3 is ready to be merged as far as I can tell.

boegel:
Tested with all existing Hypre easyconfigs + PETSc on top, good to go, thanks @Micket and @lexming!

PR #1886, opened 2019-12-14
update WPS easyblock for recent versions: set $WRF_DIR to point to location of WRF installation

boegel:


akesandgren:
Going in, thanks @boegel!

PR #1887, opened 2019-12-14
make sure $LIBLAPACK_MT is set before using it in ESMF easyblock

boegel:
When looking into a new easyconfig for the latest version of ESMF, I tried installing it with the `iimpi` toolchain which led to a confusing compilation problem involving `None`.

The problem was that `$LIBLAPACK_MT` was not defined in the environment, but the ESMF easyblock was using the value obtained via `os.getenv('LIBLAPACK_MT')` (which yields `None` if `$LIBLAPACK_MT` is not defined) without checking it...

### DIFF
--- a/easybuild/easyblocks/e/esmf.py
+++ b/easybuild/easyblocks/e/esmf.py
@@ -73,11 +73,11 @@ class EB_ESMF(ConfigureMake):
         # specify decent LAPACK lib
         env.setvar('ESMF_LAPACK', 'user')
         ldflags = os.getenv('LDFLAGS')
-        liblapack_mt = os.getenv('LIBLAPACK_MT')
-        if liblapack_mt is None:
-            raise EasyBuildError("$LIBLAPACK_MT not defined, no BLAS/LAPACK in %s toolchain?", self.toolchain.name)
+        liblapack = os.getenv('LIBLAPACK_MT') or os.getenv('LIBLAPACK')
+        if liblapack is None:
+            raise EasyBuildError("$LIBLAPACK(_MT) not defined, no BLAS/LAPACK in %s toolchain?", self.toolchain.name)
         else:
-            env.setvar('ESMF_LAPACK_LIBS', ldflags + ' ' + liblapack_mt)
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/e/esmf.py
```diff
@@ -73,7 +72,12 @@ def configure_step(self):
 
         # specify decent LAPACK lib
         env.setvar('ESMF_LAPACK', 'user')
-        env.setvar('ESMF_LAPACK_LIBS', '%s %s' % (os.getenv('LDFLAGS'), os.getenv('LIBLAPACK_MT')))
+        ldflags = os.getenv('LDFLAGS')
+        liblapack_mt = os.getenv('LIBLAPACK_MT')
+        if liblapack_mt is None:
+            raise EasyBuildError("$LIBLAPACK_MT not defined, no BLAS/LAPACK in %s toolchain?", self.toolchain.name)
```
akesandgren: This is technically wrong.
If LIBLAPACK_MT is not defined it should just use LIBLAPACK
I.e., if the easyconfig didn't set openmp: True use LIBLAPACK, if it did use openmp: True use LIB LAPACK_MT

boegel: The ESMF easyconfig doesn't set `openmp` at all, see https://github.com/easybuilders/easybuild-easyconfigs/pull/9513/files .

Related discussion: https://github.com/easybuilders/easybuild-framework/issues/2350

akesandgren: Ah, I'd forgotten about that thread.

boegel: @akesandgren How about if I use this instead:

```python
liblapack = os.getenv('LIBLAPACK_MT') or os.getenv('LIBLAPACK')

akesandgren: Yeah, that makes sense. Although I still think we should get something in the way of 2350 in place.

migueldiascosta:
Going in, thanks @boegel!

PR #1891, opened 2019-12-17
make sure $PYTHONNOUSERSITE it set when performing sanity check for (bundles of) Python package(s)

### LINKED ISSUE #1877: new sanity_pip_check should ignore modules in user home
Hi, I just tried to install `Python-3.7.4-GCCcore-8.3.0.eb` with EasyBuild 4.1.0, but it failed during the sanity check because I had some TensorFlow installation lying around in `~/.local` which depended on an old Python module (that was not loaded ofc). So apparently, the "pip check" also tried to check this tensorflow installation and failed because it could not find numpy, protobuf etc.

```
== sanity checking...
== FAILED: Installation ended unsuccessfully (build directory: /dev/shm/easybuild-build/Python/3.7.4/GCCcore-8.3.0): build failed (first 300 chars): cmd "pip check" exited with exit code 1 and output:
tensorflow-metadata 0.15.0 requires protobuf, which is not installed.
tensorflow-datasets 1.3.0 requires absl-py, which is not installed.
tensorflow-datasets 1.3.0 requires numpy, which is not installed.
tensorflow-datasets 1.3.0 requires protobuf, (took 14 min 33 sec)
```

Deleting `~/.local` solved this problem, but, wouldn't it be good if EasyBuild somehow ignored modules from the user home for the purpose of sanity checking? Not sure if this is even possible..

boegel:
We are already defining `$PYTHONNOUSERSITE` when installing Python packages, but since we reset the environment at the start of the sanity check, we need to make sure it's defined again.

fixes #1877 reported by @mstud

cc @Flamefire, @smoors

### DIFF
--- a/easybuild/easyblocks/generic/pythonpackage.py
+++ b/easybuild/easyblocks/generic/pythonpackage.py
@@ -615,12 +615,32 @@ class PythonPackage(ExtensionEasyBlock):
         self.test_step()
         self.install_step()
 
+    def load_module(self, *args, **kwargs):
+        """
+        Make sure that $PYTHONNOUSERSITE is defined after loading module file for this software."""
+
+        super(PythonPackage, self).load_module(*args, **kwargs)
+
+        # don't add user site directory to sys.path (equivalent to python -s),
+        # to avoid that any Python packages installed in $HOME/.local/lib affect the sanity check;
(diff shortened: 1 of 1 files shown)


Flamefire:
LGTM!

PR #1892, opened 2019-12-18
Change options based on arch for Mesa

edmondac:
arch-specific config opts for Mesa

### DIFF
--- a/easybuild/easyblocks/m/mesa.py
+++ b/easybuild/easyblocks/m/mesa.py
@@ -45,17 +45,18 @@ class EB_Mesa(MesonNinja):
 
         features = set(get_cpu_features())
         arches = []
-        if 'avx' in features or 'avx1.0' in features:
-            arches.append('avx')
-        if 'avx2' in features:
-            arches.append('avx2')
-        if 'avx512f' in features:
-            # AVX-512 Foundation - introduced in Skylake
-            arches.append('skx')
-        if 'avx512er' in features:
-            # AVX-512 Exponential and Reciprocal Instructions implemented in Knights Landing
-            arches.append('knl')
-        if arches:
-            self.cfg.update('configopts', "-Dswr-arches=%s" % ','.join(arches))
+        if arch == X86_64:
+            if 'avx' in features or 'avx1.0' in features:
+                arches.append('avx')
+            if 'avx2' in features:
+                arches.append('avx2')
+            if 'avx512f' in features:
+                # AVX-512 Foundation - introduced in Skylake
+                arches.append('skx')
+            if 'avx512er' in features:
+                # AVX-512 Exponential and Reciprocal Instructions implemented in Knights Landing
+                arches.append('knl')
+            if arches:
+                self.cfg.update('configopts', "-Dswr-arches=%s" % ','.join(arches))
 
         return super(EB_Mesa, self).configure_step(cmd_prefix=cmd_prefix)
--- a/easybuild/easyblocks/m/mesa.py
+++ b/easybuild/easyblocks/m/mesa.py
@@ -45,20 +45,16 @@ class EB_Mesa(MesonNinja):
             elif arch == POWER:
                 self.cfg.update('configopts', "-Dgallium-drivers='swrast'")
 
-        features = set(get_cpu_features())
-        arches = []
-        if arch == X86_64:
-            if 'avx' in features or 'avx1.0' in features:
-                arches.append('avx')
-            if 'avx2' in features:
-                arches.append('avx2')
-            if 'avx512f' in features:
-                # AVX-512 Foundation - introduced in Skylake
-                arches.append('skx')
-            if 'avx512er' in features:
-                # AVX-512 Exponential and Reciprocal Instructions implemented in Knights Landing
-                arches.append('knl')
-            if arches:
-                self.cfg.update('configopts', "-Dswr-arches=%s" % ','.join(arches))
+        if 'swr-arches' not in self.cfg['configopts']:
+            # Set cpu features of SWR for current architecture
+            cpu_features = set(get_cpu_features())
+            swr_arches = []
+            if arch == X86_64:
+                # avx512f: AVX-512 Foundation - introduced in Skylake
+                # avx512er: AVX-512 Exponential and Reciprocal Instructions implemented in Knights Landing
+                x86_features = {'avx': 'avx', 'avx1.0': 'avx', 'avx2': 'avx2', 'avx512f': 'skx', 'avx512er': 'knl'}
+                swr_arches = [farch for fname, farch in x86_features.items() if fname in cpu_features]
+            if swr_arches:
+                self.cfg.update('configopts', "-Dswr-arches=%s" % ','.join(swr_arches))
 
         return super(EB_Mesa, self).configure_step(cmd_prefix=cmd_prefix)
--- a/easybuild/easyblocks/m/mesa.py
+++ b/easybuild/easyblocks/m/mesa.py
@@ -26,6 +26,7 @@
 EasyBuild support for installing Mesa, implemented as an easyblock
 
 @author: Andrew Edmondson (University of Birmingham)
+"""
 
 from easybuild.easyblocks.generic.mesonninja import MesonNinja
 from easybuild.tools.systemtools import POWER, X86_64, get_cpu_architecture
--- a/easybuild/easyblocks/m/mesa.py
+++ b/easybuild/easyblocks/m/mesa.py
@@ -1,4 +1,31 @@
-# This easyblock was created by the BEAR Software team at the University of Birmingham
+##
+# Copyright 2009-2020 Ghent University
+#
+# This file is part of EasyBuild,
+# originally created by the HPC team of Ghent University (http://ugent.be/hpc/en),
+# with support of Ghent University (http://ugent.be/hpc),
+# the Flemish Supercomputer Centre (VSC) (https://www.vscentrum.be),
+# Flemish Research Foundation (FWO) (http://www.fwo.be/en)
+# and the Department of Economy, Science and Innovation (EWI) (http://www.ewi-vlaanderen.be/en).
+#
+# https://github.com/easybuilders/easybuild
+#
+# EasyBuild is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation v2.
+#
+# EasyBuild is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with EasyBuild.  If not, see <http://www.gnu.org/licenses/>.
+##
+"""
+EasyBuild support for installing Mesa, implemented as an easyblock
+
+@author: Andrew Edmondson (University of Birmingham)
 
 from easybuild.easyblocks.generic.mesonninja import MesonNinja
 from easybuild.tools.systemtools import POWER, X86_64, get_cpu_architecture
--- a/easybuild/easyblocks/m/mesa.py
+++ b/easybuild/easyblocks/m/mesa.py
@@ -29,7 +29,7 @@ EasyBuild support for installing Mesa, implemented as an easyblock
 """
 
 from easybuild.easyblocks.generic.mesonninja import MesonNinja
-from easybuild.tools.systemtools import POWER, X86_64, get_cpu_architecture
+from easybuild.tools.systemtools import POWER, X86_64, get_cpu_architecture, get_cpu_features
 
 
 class EB_Mesa(MesonNinja):
@@ -42,4 +42,20 @@ class EB_Mesa(MesonNinja):
             self.cfg.update('configopts', "-Dgallium-drivers='swrast,swr'")
         elif arch == POWER:
             self.cfg.update('configopts', "-Dgallium-drivers='swrast'")
+
+        features = set(get_cpu_features())
+        arches = []
+        if 'avx' in features:
+            arches.append('avx')
+        if 'avx2' in features:
+            arches.append('avx2')
+        if 'avx512f' in features:
+            # AVX-512 Foundation - introduced in Skylake
+            arches.append('skx')
+        if 'avx512er' in features:
+            # AVX-512 Exponential and Reciprocal Instructions implemented in Knights Landing
+            arches.append('knl')
+        if arches:
+            self.cfg.update('configopts', "-Dswr-arches=%s" % ','.join(arches))
+
         return super(EB_Mesa, self).configure_step(cmd_prefix=cmd_prefix)
--- a/easybuild/easyblocks/m/mesa.py
+++ b/easybuild/easyblocks/m/mesa.py
@@ -38,10 +38,12 @@ class EB_Mesa(MesonNinja):
         Customise the configopts based on the platform
         """
         arch = get_cpu_architecture()
-        if arch == X86_64:
-            self.cfg.update('configopts', "-Dgallium-drivers='swrast,swr'")
-        elif arch == POWER:
-            self.cfg.update('configopts', "-Dgallium-drivers='swrast'")
+        if 'gallium-drivers' not in self.cfg['configopts']:
+            # Install appropriate Gallium drivers for current architecture
+            if arch == X86_64:
+                self.cfg.update('configopts', "-Dgallium-drivers='swrast,swr'")
+            elif arch == POWER:
+                self.cfg.update('configopts', "-Dgallium-drivers='swrast'")
 
         features = set(get_cpu_features())
         arches = []
--- a/easybuild/easyblocks/m/mesa.py
+++ b/easybuild/easyblocks/m/mesa.py
@@ -45,7 +45,7 @@ class EB_Mesa(MesonNinja):
 
         features = set(get_cpu_features())
         arches = []
-        if 'avx' in features:
+        if 'avx' in features or 'avx1.0' in features:
             arches.append('avx')
         if 'avx2' in features:
             arches.append('avx2')
--- a/easybuild/easyblocks/m/mesa.py
+++ b/easybuild/easyblocks/m/mesa.py
@@ -1,0 +1,17 @@
+# This easyblock was created by the BEAR Software team at the University of Birmingham
+
+from easybuild.easyblocks.generic.mesonninja import MesonNinja
+from easybuild.tools.systemtools import POWER, X86_64, get_cpu_architecture
+
+
+class EB_Mesa(MesonNinja):
+    def configure_step(self, cmd_prefix=''):
+        """
+        Customise the configopts based on the platform
+        """
+        arch = get_cpu_architecture()
+        if arch == X86_64:
+            self.cfg.update('configopts', "-Dgallium-drivers='swrast,swr' -Dswr-arches=avx,avx2,skx,knl")
+        elif arch == POWER:
+            self.cfg.update('configopts', "-Dgallium-drivers='swrast'")
+        return super(EB_Mesa, self).configure_step(cmd_prefix=cmd_prefix)
--- a/easybuild/easyblocks/m/mesa.py
+++ b/easybuild/easyblocks/m/mesa.py
@@ -11,7 +11,7 @@ class EB_Mesa(MesonNinja):
         """
         arch = get_cpu_architecture()
         if arch == X86_64:
-            self.cfg.update('configopts', "-Dgallium-drivers='swrast,swr')
+            self.cfg.update('configopts', "-Dgallium-drivers='swrast,swr'")
         elif arch == POWER:
             self.cfg.update('configopts', "-Dgallium-drivers='swrast'")
         return super(EB_Mesa, self).configure_step(cmd_prefix=cmd_prefix)
--- a/easybuild/easyblocks/m/mesa.py
+++ b/easybuild/easyblocks/m/mesa.py
@@ -11,7 +11,7 @@ class EB_Mesa(MesonNinja):
         """
         arch = get_cpu_architecture()
         if arch == X86_64:
-            self.cfg.update('configopts', "-Dgallium-drivers='swrast,swr' -Dswr-arches=avx,avx2,skx,knl")
+            self.cfg.update('configopts', "-Dgallium-drivers='swrast,swr')
         elif arch == POWER:
             self.cfg.update('configopts', "-Dgallium-drivers='swrast'")
         return super(EB_Mesa, self).configure_step(cmd_prefix=cmd_prefix)


easybuild/easyblocks/m/mesa.py
```diff
@@ -0,0 +1,17 @@
...
+
+
+class EB_Mesa(MesonNinja):
+    def configure_step(self, cmd_prefix=''):
+        """
+        Customise the configopts based on the platform
+        """
+        arch = get_cpu_architecture()
+        if arch == X86_64:
+            self.cfg.update('configopts', "-Dgallium-drivers='swrast,swr' -Dswr-arches=avx,avx2,skx,knl")
```
boegel: @edmondac Can you clarify what this does exactly? Does this yield a "fat" build of Mesa, which works on different processor architectures?
Any links to Mesa documentation for this?

edmondac: This is what was in https://github.com/easybuilders/easybuild-easyconfigs/blob/develop/easybuild/easyconfigs/m/Mesa/Mesa-19.2.1-GCCcore-8.3.0.eb

I can revert to the https://github.com/easybuilders/easybuild-easyconfigs/blob/develop/easybuild/easyconfigs/m/Mesa/Mesa-19.1.7-GCCcore-8.3.0.eb options if you like, which doesn't have the `-Dswr-arches` flag at all...

edmondac: And I don't know what the difference is... I just didn't want to lose a potential improvement. The swr-arches bit in the later easyconfig is all about x86 so wouldn't work on power9, for example.

boegel: Maybe @wpoely86 is familiar with this?

boegel: Or @bartoldeman, since the `Mesa` easyconfig that includes this was contributed by him in https://github.com/easybuilders/easybuild-easyconfigs/pull/9143

wpoely86: As far as I remember, swr by default builds a version for avx, for avx2, etc and picks the correct one at runtime.

edmondac: I now think that it shouldn't have this bit at all, since it won't build on anything older than Skylake...
```
-Dswr-arches=avx,avx2,skx,knl"
```
Patch coming.

A specific easyconfig could still add that in if desired, like the one in easybuilders/easybuild-easyconfigs#9143

easybuild/easyblocks/m/mesa.py
```diff
@@ -0,0 +1,17 @@
...
+
+
+class EB_Mesa(MesonNinja):
+    def configure_step(self, cmd_prefix=''):
+        """
+        Customise the configopts based on the platform
+        """
+        arch = get_cpu_architecture()
+        if arch == X86_64:
+            self.cfg.update('configopts', "-Dgallium-drivers='swrast,swr')
```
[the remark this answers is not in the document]
edmondac: Fixed

easybuild/easyblocks/m/mesa.py
```diff
@@ -0,0 +1,17 @@
+# This easyblock was created by the BEAR Software team at the University of Birmingham
+
+from easybuild.easyblocks.generic.mesonninja import MesonNinja
+from easybuild.tools.systemtools import POWER, X86_64, get_cpu_architecture
+
+
+class EB_Mesa(MesonNinja):
+    def configure_step(self, cmd_prefix=''):
```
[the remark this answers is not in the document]
edmondac: I don't understand that... @boegel ?

easybuild/easyblocks/m/mesa.py
```diff
@@ -0,0 +1,61 @@
...
+        """
+        arch = get_cpu_architecture()
+        if arch == X86_64:
+            self.cfg.update('configopts', "-Dgallium-drivers='swrast,swr'")
+        elif arch == POWER:
+            self.cfg.update('configopts', "-Dgallium-drivers='swrast'")
+
+        features = set(get_cpu_features())
+        arches = []
+        if 'avx' in features:
```
smoors:
```suggestion
        if 'avx' in features or 'avx1.0' in features:
```

bartoldeman:
Just a few comments:

the reason why I put in the all the swr arches unconditionally is that they are selected at runtime, and you may be "cross-compiling". The problem with the recipes was that in the 19.1.* they run into a GCC bug (fixed in GCC 9 IIRC) where `-march=native` is *not* overridden with `-march=skylake-avx512` or similar, so avx512 code when compiled with `-march=native -march=skylake-avx512` errors out when the build machine does not support avx512 . That is why I included a patch in the easyconfig for 19.2..1

edmondac:
> Just a few comments:
> 
> the reason why I put in the all the swr arches unconditionally is that they are selected at runtime, and you may be "cross-compiling". The problem with the recipes was that in the 19.1.* they run into a GCC bug (fixed in GCC 9 IIRC) where `-march=native` is _not_ overridden with `-march=skylake-avx512` or similar, so avx512 code when compiled with `-march=native -march=skylake-avx512` errors out when the build machine does not support avx512 . That is why I included a patch in the easyconfig for 19.2..1

Hi @bartoldeman - the problem with this approach is that it doesn't build on POWER9 CPUs:
```
src/gallium/drivers/swr/meson.build:200:2: ERROR: Problem encountered: Cannot find AVX support for swr. (these are required for SWR an all architectures.)
```

The new approach does, see https://github.com/easybuilders/easybuild-easyconfigs/pull/9764#issuecomment-596584673

Is cross-compilation "supported" by EasyBuild? Not using it, I wouldn't know. So many things auto-detect their CPU features that I'd have thought it would mostly not build with the right support - but I've not tried it.

easybuild/easyblocks/m/mesa.py
```diff
@@ -0,0 +1,61 @@
...
+        if 'avx2' in features:
+            arches.append('avx2')
+        if 'avx512f' in features:
+            # AVX-512 Foundation - introduced in Skylake
+            arches.append('skx')
+        if 'avx512er' in features:
+            # AVX-512 Exponential and Reciprocal Instructions implemented in Knights Landing
+            arches.append('knl')
+        if arches:
+            self.cfg.update('configopts', "-Dswr-arches=%s" % ','.join(arches))
```
akesandgren: Shjouldn't this whole block be inside the "if arch == X86_64"

Flamefire: Judging from the name I'd expect this only to be used when `swr` is built, so I guess yes.

> Is cross-compilation "supported" by EasyBuild? Not using it, I wouldn't know. So many things auto-detect their CPU features that I'd have thought it would mostly not build with the right support - but I've not tried it.

Kinda. For example we do have 3 installation trees: Power, sandy bridge, haswell. The haswell is also used for (a few) newer ones. So while this is no real cross compilation it is compiling for support on the maximal common arch. So I'd include the arches unconditionally based on 

>  they are selected at runtime

edmondac: lol - this is what it was like earlier before https://github.com/easybuilders/easybuild-easyblocks/pull/1892#pullrequestreview-371226403

What's the conclusion @Flamefire and @akesandgren ?

Flamefire: I'd add them always if swr is built or do that in the EC. I have the feeling a compiler might choke if e.g. knl is added but the compiler is to old. Not sure how smart the configure is. If it is smart enough: Add always in EasyBlock, else in EasyConfig

akesandgren: Not sure what you mean by that. That whole block of checking for avx,skx,knl are only valid if it is a X86_64 architecture in the first place, so why do it even if the base arch is POWER?

smoors:
```suggestion
        if arch == X86_64:
            if 'avx' in features or 'avx1.0' in features:
                arches.append('avx')
            if 'avx2' in features:
                arches.append('avx2')
            if 'avx512f' in features:
                # AVX-512 Foundation - introduced in Skylake
                arches.append('skx')
            if 'avx512er' in features:
                # AVX-512 Exponential and Reciprocal Instructions implemented in Knights Landing
                arches.append('knl')
            if arches:
                self.cfg.update('configopts', "-Dswr-arches=%s" % ','.join(arches))
```

Flamefire: Sorry for the confusion:
- `swr-arches` (most likely) applies only when building swr, hence yes this block need to be moved inside the X86_64 check
- I would not do the arch check at all and simply add all available archs to `swr-arches` inside X86_64 to allow selection at runtime
- If the configure of Mesa would add flags that are incompatible with older compilers when e.g. `swr-arches=knl` is set (happens of Mesa doesn't check those flags before adding, but I don't know if it does check) then adding `swr-arches` should be done in an EasyConfig instead (where compiler support for such flags is known)

Flamefire: Ok I took the time to verify the last point: Mesa *does* add the flags unconditionally. Hence if `swr-arches=knl` is given and the compiler doesn't support this flag the configure will fail.

-> Add those archs via EasyConfig

akesandgren: well compiler support is sort of knowable in the easyblock too by checking compiler version

easybuild/easyblocks/m/mesa.py
```diff
@@ -0,0 +1,62 @@
...
+class EB_Mesa(MesonNinja):
+    def configure_step(self, cmd_prefix=''):
+        """
+        Customise the configopts based on the platform
+        """
+        arch = get_cpu_architecture()
+        if arch == X86_64:
+            self.cfg.update('configopts', "-Dgallium-drivers='swrast,swr'")
+        elif arch == POWER:
+            self.cfg.update('configopts', "-Dgallium-drivers='swrast'")
```
lexming:
```suggestion
        if 'gallium-drivers' not in self.cfg['configopts']:
            # Install appropriate Gallium drivers for current architecture
            if arch == X86_64:
                self.cfg.update('configopts', "-Dgallium-drivers='swrast,swr'")
            elif arch == POWER:
                self.cfg.update('configopts', "-Dgallium-drivers='swrast'")
```

easybuild/easyblocks/m/mesa.py
```diff
@@ -0,0 +1,62 @@
...
+            if 'avx2' in features:
+                arches.append('avx2')
+            if 'avx512f' in features:
+                # AVX-512 Foundation - introduced in Skylake
+                arches.append('skx')
+            if 'avx512er' in features:
+                # AVX-512 Exponential and Reciprocal Instructions implemented in Knights Landing
+                arches.append('knl')
+            if arches:
+                self.cfg.update('configopts', "-Dswr-arches=%s" % ','.join(arches))
```
lexming:
```suggestion
        if 'swr-arches' not in self.cfg['configopts']:
            # Set cpu features of SWR for current architecture
            cpu_features = set(get_cpu_features())
            swr_arches = []
            if arch == X86_64:
                # avx512f: AVX-512 Foundation - introduced in Skylake
                # avx512er: AVX-512 Exponential and Reciprocal Instructions implemented in Knights Landing
                x86_features = {'avx': 'avx', 'avx1.0': 'avx', 'avx2': 'avx2', 'avx512f': 'skx', 'avx512er': 'knl'}
                swr_arches = [farch for fname, farch in x86_features.items() if fname in cpu_features]
            if swr_arches:
                self.cfg.update('configopts', "-Dswr-arches=%s" % ','.join(swr_arches))
```

easybuild/easyblocks/m/mesa.py
lexming: compiler support for `swr-arches` is not an issue right now because this easyblock for `mesa` will only be used with recent toolchains (i.e. old easyconfigs for Mesa will stay with `configuremake`). In the future, if there are compiler dependent `configopts`, we can just add specific conditionals to the `mesa` easyblock.

easybuild/easyblocks/m/mesa.py
smoors: @lexming this is not entirely correct: if `avx1.0` is present, `avx` feature should be added.

lexming: That's exactly what it does with `{'avx1.0': 'avx'}` in `x86_features`, isn't it?. If the key 'avx1.0' is in `cpu_features` it adds its value 'avx' to `swr_arches`

smoors: ah you're right, sorry I misread..

easybuild/easyblocks/m/mesa.py
edmondac: OK - implemented in the suggested changes below from @smoors :+1:

easybuild/easyblocks/m/mesa.py
```diff
@@ -0,0 +1,60 @@
...
+
+        if 'swr-arches' not in self.cfg['configopts']:
+            # Set cpu features of SWR for current architecture
+            cpu_features = set(get_cpu_features())
+            swr_arches = []
+            if arch == X86_64:
+                # avx512f: AVX-512 Foundation - introduced in Skylake
+                # avx512er: AVX-512 Exponential and Reciprocal Instructions implemented in Knights Landing
+                x86_features = {'avx': 'avx', 'avx1.0': 'avx', 'avx2': 'avx2', 'avx512f': 'skx', 'avx512er': 'knl'}
+                swr_arches = [farch for fname, farch in x86_features.items() if fname in cpu_features]
```
Flamefire: Shouldn't this be a set? Otherwise avx might be added twice

Flamefire:
This might have been forgotten but what was the reasoning for NOT adding all the archs independent of the current CPU? See https://github.com/easybuilders/easybuild-easyblocks/pull/1892#issuecomment-596628357 and https://github.com/easybuilders/easybuild-easyblocks/pull/1892#discussion_r395472877

As they are selected on runtime it is better they are always included to allow the (officially suggested) use case of compiling on the maximum arch common to all platforms that will be using this.   
Additionally always including all allows to find typos and such in the names. Currently a typo would only be found in the rare case it is compiled on that specific arch.

easybuild/easyblocks/m/mesa.py
lexming: it should not be an issue if it is added twice

lexming:
@Flamefire check https://github.com/easybuilders/easybuild-easyblocks/pull/1892#issuecomment-597007388

Flamefire:
@lexming I've seen that and answered in https://github.com/easybuilders/easybuild-easyblocks/pull/1892#discussion_r395472877. There was some (kind of) discussion afterwards but nothing against always setting all arches. Or did I miss it?

lexming:
@Flamefire according to the reply by @edmondac, indiscriminately adding AVX archs on all cases breaks the build on POWER9

Flamefire:
@lexming: See: https://github.com/easybuilders/easybuild-easyblocks/pull/1892#discussion_r395549476
> I would not do the arch check at all and simply add all available archs to swr-arches inside X86_64 to allow selection at runtime

swr is not build on Power with this PR so no problem

lexming:
@Flamefire that's the point, those two options go hand by hand. If SWR is not present in non-x86 archs, there is no point in setting `swr-archs`. It won't be possible to do anything with those, as SWR won't be present. If I'm missing something please let me know.

Flamefire:
You are missing the point: I proposed to add all options to swr-archs on x86:

        if 'swr-arches' not in self.cfg['configopts']:
            if arch == X86_64:
                self.cfg.update('configopts', "-Dswr-arches=avx,avx2,skx,knl")

lexming:
@Flamefire I see, ok I can try if that works on something old like an IvyBridge

lexming:
@Flamefire your proposal does not work, have you tested it?

Building Mesa on a machine with an IvyBridge CPU using `-Dgallium-drivers='swrast,swr'  -Dswr-arches=knl,skx,avx2,avx` crashes with many errors related to AVX512
> ../mesa-19.1.7/src/gallium/drivers/swr/rasterizer/common/simdlib_512_avx512_core.inl:162:42: error: cannot convert 'SIMDImpp
     l::SIMD512Impl::Integer' to '__m512i' {aka '__vector(8) long long int'} 

Building on the same IvyBridge with `-Dgallium-drivers='swrast,swr'  -Dswr-arches=avx` works well.

Flamefire:
It seems to have worked: https://github.com/easybuilders/easybuild-easyconfigs/pull/9764/files#diff-a56c37857393d269498e8fd48f40500dL64

Maybe the problem mentioned in https://github.com/easybuilders/easybuild-easyblocks/pull/1892#issuecomment-596628357

akesandgren:
As long as the compiler is new enough to understand the various swr-arches there is no problem to build it.
I just built this on our AMD Bulldozer system using https://github.com/easybuilders/easybuild-easyconfigs/pull/10175

And this easyublock will only be used in Mesa Ec's using a new enough compiler toolchain.
The older Mesa ECs will not be touched.

So I judge this good to go. Any one opposed? If not I'll merge this in < 2h

Flamefire:
I feel being ignored.

I do see value in having runtime selection of more optimized implementations than available on the build machine

akesandgren:
@Flamefire Ah, you wanted the swr-arches to be complete, regardless of where it is built....

Let me recheck that part... I forgot to double check what the current code does...

akesandgren:
And yes, building it with
```
if 'swr-arches' not in self.cfg['configopts']:
            if arch == X86_64:
                self.cfg.update('configopts', "-Dswr-arches=avx,avx2,skx,knl")
```
works very well on our AMD Bulldozer, so I'm for doing it that way and let it find the best swr-arch at runtime.

@lexming what where you really testing when it failed??

lexming:
@akesandgren I tried exactly that in an IvyBridge and it fails with errors related to AVX512

lexming:
@Flamefire you are far from being ignored, I did listen to your request and went as far as testing it. Unfortunately, your idea does not work out-of-the-box in all micro-architectures, which is a must for this easyblock. Therefore, IMO the current state for this easyblock is the safe path forward.

akesandgren:
Using self.cfg.update('configopts', "-Dswr-arches=avx,avx2,skx,knl") works on our broadwell nodes at least.

Flamefire:
My point was that there is a pending investigation but this was about to be merged.
The existing easyconfig with all archs has worked so far, so the proposed approach is reasonable. 
The question is: why did it fail for you? Can you post the easyconfig and easyblock used together with the log? I suggested that the problem might be in Mesa and already known and solved. Is that the case?

lexming:
Using `-Dgallium-drivers='swrast,swr' -Dswr-arches=knl,skx,avx2,avx` (your proposal) hits bug #105029 (https://bugs.freedesktop.org/show_bug.cgi?id=105029) on IvyBridge for easyconfigs of Mesa < v19.3.7 using GCC-8.

The current easyblock for `mesa` in this PR does not hit that bug because it compiles Mesa with `-Dgallium-drivers='swrast,swr' -Dswr-arches=avx`.

Since there is already a PR open (https://github.com/easybuilders/easybuild-easyconfigs/pull/9764) to use the `mesa` easyblock on Mesa easyconfigs below v19.3.7, this is a serious concern.

As I said, the current state of this PR seems to be a safer path forward. If any user wants to do cross-compiling, this easyblock already allows the user to do that by manually overriding `swr-arches` in the easyconfig (ie adding `-Dswr-arches=knl,skx,avx2,avx` to `configopts`)

This is just my advice, I leave the final decision to other maintainers.

Flamefire:
Do I misunderstood the bug report because to me this looks like an issue with a compiler not with the architecture.
That's why I suggested to just add the archs in easyconfigs where the compiler is known. Currently I expect this to fail with Mesa < 19.3.7 on archs supporting avx 512 but with older compilers (?) 

So this makes this an arch dependent failure which is even worse

And I mentioned that this is probably what was mentioned in https://github.com/easybuilders/easybuild-easyblocks/pull/1892#issuecomment-596628357

easybuild/easyblocks/m/mesa.py
akesandgren: So after finding that Mesa < 19.3.7 bug, suggested code would be:
```
if arch == X86_64:
  x86_features = {'avx': 'avx', 'avx1.0': 'avx', 'avx2': 'avx2', 'avx512f': 'skx', 'avx512er': 'knl'}
  if LooseVersion(self.version) < LooaseVersion('19.3.7'):
    swr_arches = [farch for fname, farch in x86_features.items() if fname in cpu_features]
  else:
    swr_arches = [farch for fname, farch in x86_features.items() ]
```
Or something similar.

akesandgren:
@Flamefire @lexming good enough for both of you?

And if so, @edmondac can you make that effect if the4y both agree?

Flamefire:
Again the question: is this a bug related to the current arch or the current compiler? I understood it to be the latter. Then this is not acceptable as it will fail when one tries to build on knl or skylake(?)

lexming:
Based on my tests, the current `mesa` easyblock (commit 3c02f75) works well in the following cases:

* Mesa-19.0.1-GCCcore-8.2.0.eb: IvyBridge and Skylake
* Mesa-19.1.7-GCCcore-8.3.0.eb: IvyBridge and Skylake
* Mesa-20.0.2-GCCcore-9.3.0.eb: IvyBridge and Skylake

So the bug occurs in a combination of architecture and compiler when using `-Dswr-arches=knl,skx,avx2,avx` on non-AVX512 archs (?) That is unclear, but in any case, my tests are all positive so far with this easyblock as it is.

@akesandgren I would really prefer avoiding the burden and complication of maintaining version checks of the compiler or Mesa unless we find a case where this easyblock fails This is not currently the case. Moreover (I repeat myself) users wanting to do cross-compiling can set `swr-arches` in configopts and this easyblock will respect that.

Having said all of this, if a majority of you prefer to add version checks on the compiler and Mesa, I won't oppose this change obviously.

Flamefire:
And you are still missing the point. To summarize:

- All evidence suggest that there is a known issue with Mesa < 19.3.7 (?) which runs into a GCC 8 bug that makes it fail to compile if knl or skx is specified for swr-arches. This is from https://bugs.freedesktop.org/show_bug.cgi?id=105029 and https://github.com/easybuilders/easybuild-easyblocks/pull/1892#issuecomment-596628357
- These flags will be used with the current EasyBlock if skylake or knl architectures are used to build mesa
- Hence there is a case where this easyblock fails. You did not test compiling on skylake, did you? Hence this isn't noticed yet but it will fail

> users wanting to do cross-compiling can set swr-arches in configopts

See above: The bug is the compiler with older Mesa versions. There was not any evidence that this bug only applies when running on non-AVX-512 archs. The linked issue shows a compile failure, not a runtime failure.

>  I would really prefer avoiding the burden and complication of maintaining version checks of the compiler or Mesa

The burden of having a bug that is only reproducible on 1 architecture due to the EasyBlock changing behavior based on the architecture is IMO higher.

Furthermore there exists a patch for that which has been mentioned in https://github.com/easybuilders/easybuild-easyblocks/pull/1892#issuecomment-596628357: https://github.com/easybuilders/easybuild-easyconfigs/pull/9764/files#diff-a56c37857393d269498e8fd48f40500dR29

lexming:
>Hence there is a case where this easyblock fails. You did not test compiling on skylake, did you? Hence this isn't noticed yet but it will fail

/facepalm

I already tested this easyblock in a Skylake (Intel(R) Xeon(R) Gold 6126) and it works. Precisely with `Mesa-19.0.1-GCCcore-8.2.0.eb`, `Mesa-19.1.7-GCCcore-8.3.0.eb` and `Mesa-20.0.2-GCCcore-9.3.0.eb`. I just said that in my previous comment (https://github.com/easybuilders/easybuild-easyblocks/pull/1892#issuecomment-602046224). Are you reading them?

I am tired of repeating myself. All those tests listed in https://github.com/easybuilders/easybuild-easyblocks/pull/1892#issuecomment-602046224 with this easyblock as it is in commit cac0c0c are **successful**.

I won't participate in this discussion any more unless you provide some evidence to your claims. So far you have talked a lot, but have not submitted the results of any test. I already provided tests of several Mesa easyconfigs in different architectures, I tested both the current easyblock and your proposed changes, and I am the one who found that we could be affected by bug 105029 by using indiscriminate `swr-archs`. So I really doubt that I'm missing any point here.

Flamefire:
Apologies, it seems it is KNL affected by the bug you mentioned, not skylake. Got fooled by the mentioning of avx512. The upstream fix is this: https://gitlab.freedesktop.org/mesa/mesa/-/commit/f1fbeb1a53 and it seems to be fixed in 18.1.0. So that doesn't look like the bug you encountered. But since no test reports of the failure were submitted despite me asking I couldn't know better.

> So far you have talked a lot, but have not submitted the results of any test.

I did link https://github.com/easybuilders/easybuild-easyconfigs/pull/9764/files#diff-a56c37857393d269498e8fd48f40500dR29 and https://github.com/easybuilders/easybuild-easyblocks/pull/1892#issuecomment-596628357 multiple times where all archs are passed for an existing EasyConfig and all tests (submitted when this was merged and current usage) show that using all archs is possible.

> and I am the one who found that we could be affected by bug 105029 by using indiscriminate swr-archs. So I really doubt that I'm missing any point here.

And I'm the one who pointed out that we might be affected by the issue reported in https://github.com/easybuilders/easybuild-easyblocks/pull/1892#issuecomment-596628357 and asked to verify this. It seems you must have missed https://github.com/easybuilders/easybuild-easyblocks/pull/1892#issuecomment-601896496.

I did just do this: I used this EC: https://gist.github.com/Flamefire/6730d3b40d75dabf47d9df3a3ca515f

It is the current Mesa-19.1.7-GCCcore-8.3.0.eb with `-Dswr-arches=avx,avx2,skx,knl` and the mentioned patch added. I successfully built this on a sandy-bridge. I also verified that removing the patch makes it fail with errors like:

>  ../mesa-19.1.7/src/gallium/drivers/swr/rasterizer/core/format_types.h:1023:45: error: cannot convert simd16scalar {aka SIMDImpl::SIMD512Impl::Float} to __m512 {aka __vector(16) float}
result = _mm512_mask_blend_ps(mask, result2, result);



So conclusion: We run into https://gcc.gnu.org/bugzilla/show_bug.cgi?id=69471 as mentioned by @bartoldeman in https://github.com/easybuilders/easybuild-easyblocks/pull/1892#issuecomment-596628357 and there is a patch available so we can continue to build for all architectures.

However building only with the swr for the current arch happens to work as `-march=native` and the swr-arch-specific `-march` don't conflict in that situation. This is why the current state works but has the downside of reduced performance when building for the highest common CPU architecture used.

easybuild/easyblocks/m/mesa.py
boegel: I agree with @lexming here that including this version check is not a good idea, especially since there's a patch available to fix the problem you're dancing around here.

boegel:
Seems like it's time I pitch in here... :)

EasyBuild usually builds for the *host* processor architecture (usually by just using `-march=native`), so that's also what this easyblock should do by default...

Also listing other values in the `swr-arches` configuration option which correspond to older processor architectures (for example also listing `avx` when building on Intel Skylake) is fine though. In some sense that's cross-compiling, but only for older stuff that is definitely supported for the build host, so I don't see an issue there.

I'm not in favor of always listing all supported architectures to try and obtain a Mesa installation that will perform "optimally" on hosts that have newer processors than the build host. That's pretty atypical for EasyBuild, so that shouldn't be the default.

In the current state (commit cac0c0c) there's an easy way to override the default of building for the host architecture: as soon as you include `swr-arches=...` via `configopts` you're in full control, and you can, for example, opt for also listing `skx` when building on a Haswell host to obtain a Mesa installation that will switch at runtime to something optimized for Skylake.

@Flamefire I understand your point that obtaining a Mesa installation that will perform well on other (newer) architectures can be useful, but the problem that @bartoldeman fixed with a patch shows that this is also a bit risky.
Although it's fixed now, I can envision a future where a similar issue is introduced again.

So I would prefer going forward with the current implementation, which fits better with what EasyBuild usually does by default, and is less likely to cause build problems going forward.

I hope that settles the discussion.

I would like to see some minor stylistic changes to the easyblock (like stretching out the `x86_features` dict to have one entry per line), as well as adding a custom `sanity_check_step` method, before merging this. I'll look into this myself right now...

Flamefire:
> EasyBuild usually builds for the host processor architecture (usually by just using -march=native), so that's also what this easyblock should do by default...

> I'm not in favor of always listing all supported architectures to try and obtain a Mesa installation that will perform "optimally" on hosts that have newer processors than the build host. That's pretty atypical for EasyBuild, so that shouldn't be the default.

Ok, to me it looked like it should do that as it was removed from the EC: https://github.com/easybuilders/easybuild-easyconfigs/pull/9764/files#diff-a56c37857393d269498e8fd48f40500dL64
So I guess it should be added there again to not change existing behavior.

BTW: What is done for other Software that does runtime selection? Doesn't OpenBLAS or FFTW does something similar? This is why I didn't consider it atypical.

boegel:
> > EasyBuild usually builds for the host processor architecture (usually by just using -march=native), so that's also what this easyblock should do by default...
> 
> > I'm not in favor of always listing all supported architectures to try and obtain a Mesa installation that will perform "optimally" on hosts that have newer processors than the build host. That's pretty atypical for EasyBuild, so that shouldn't be the default.
> 
> Ok, to me it looked like it should do that as it was removed from the EC: https://github.com/easybuilders/easybuild-easyconfigs/pull/9764/files#diff-a56c37857393d269498e8fd48f40500dL64
> So I guess it should be added there again to not change existing behavior.

I consider this approach better than hardcoding `avx,avx2,skx,knl`, so I consider the changed behavior as a bug fix, not a regression (so it's OK to change).

> BTW: What is done for other Software that does runtime selection? Doesn't OpenBLAS or FFTW does something similar? This is why I didn't consider it atypical.

I don't know of any other software which does runtime selection to an arch that is newer than what the software was built for.
For FFTW, we only enable building for CPU features supported by the build host, and then leave it to FFTW itself to do runtime selection, so very similar to what we do here for Mesa.
For OpenBLAS, the easyblock is only aware of building for a specific architecture as far as making sure a `TARGET=...` that is included in `buildopts` is also set in `testopts` and `installopts`, but it doesn't specify any target arch at all and leaves it up to OpenBLAS itself (which detects the host CPU and targets that by default).

boegel:
@edmondac Please take a look at https://github.com/bear-rsg/easybuild-easyblocks/pull/17, I took this a step further, mostly w.r.t. the sanity check...

easybuild/easyblocks/m/mesa.py
```diff
@@ -0,0 +1,122 @@
...
+            gles_inc_files = [('GLES', 'gl.h'), ('GLES2', 'gl2.h'), ('GLES3', 'gl3.h')]
+            header_files = [os.path.join('include', 'GL', x) for x in gl_inc_files]
+            header_files.extend([os.path.join('include', x, y) for (x, y) in gles_inc_files])
+
+        custom_paths = {
+            'files': [os.path.join('lib', 'libOSMesa.%s' % shlib_ext)] + header_files,
+            'dirs': [os.path.join('include', 'GL', 'internal')],
+        }
+
+        if self.swr_arches:
```
Flamefire: Whoops, this is indeed wrong: It expects swr to be build without checking the configopts

boegel: @Flamefire Did this get fixed?

boegel: I guess it was in #2006...

PR #1896, opened 2019-12-21
impi: don't rebuild libfabric if the source code is not present.

bartoldeman:
libfabric source is no longer included with Intel MPI 2019.6

### DIFF
--- a/easybuild/easyblocks/i/impi.py
+++ b/easybuild/easyblocks/i/impi.py
@@ -51,7 +51,7 @@ class EB_impi(IntelBase):
     def extra_options():
         extra_vars = {
             'libfabric_configopts': ['', 'Configure options for the provided libfabric', CUSTOM],
-            'libfabric_rebuild': [True, 'Rebuild the internal libfabric instead of using the provided binary', CUSTOM],
+            'libfabric_rebuild': [True, 'Try to rebuild the internal libfabric instead of using the provided binary', CUSTOM],
             'ofi_internal': [True, 'Use internal shipped libfabric instead of external libfabric', CUSTOM],
             'set_mpi_wrappers_compiler': [False, 'Override default compiler used by MPI wrapper commands', CUSTOM],
             'set_mpi_wrapper_aliases_gcc': [False, 'Set compiler for mpigcc/mpigxx via aliases', CUSTOM],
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/i/impi.py
```diff
@@ -126,7 +126,8 @@ def install_step(self):
             run_cmd(cmd, log_all=True, simple=True)
 
         # recompile libfabric (if requested)
-        if impiver >= LooseVersion('2019') and self.cfg['libfabric_rebuild']:
+        if (impiver >= LooseVersion('2019') and self.cfg['libfabric_rebuild'] and
+                os.path.exists(os.path.join(self.installdir, 'libfabric'))):
             if self.cfg['ofi_internal']:
                 change_dir(os.path.join(self.installdir, 'libfabric'))
```
boegel: @bartoldeman Let's introduce an additional variable for the `libfabric` path?

```python
if impiver >= LooseVersion('2019') and self.cfg['libfabric_rebuild']:
    libfabric_path = os.path.join(self.installdir, 'libfabric')
    # some Intel MPI versions (like 2019 update 5) no longer ship libfabric sources
    if self.cfg['ofi_internal'] and os.path.exists(libfabric_path):
        change_dir(libfabric_path)

Comment (bartoldeman) [reply easybuild/easyblocks/i/impi.py]: I can't move the condition to the second `if` statement (the `else` won't work properly then), but I could introduce the variable so I did that. Also the comment applies to update 6 only (I checked update 5!)

File: easybuild/easyblocks/i/impi.py [hunk 2]
```diff
@@ -126,9 +126,12 @@ def install_step(self):
             run_cmd(cmd, log_all=True, simple=True)
 
         # recompile libfabric (if requested)
-        if impiver >= LooseVersion('2019') and self.cfg['libfabric_rebuild']:
+        # some Intel MPI versions (like 2019 update 6) no longer ship libfabric sources
+        libfabric_path = os.path.join(self.installdir, 'libfabric')
+        if (impiver >= LooseVersion('2019') and self.cfg['libfabric_rebuild'] and
+                os.path.exists(os.path.join(libfabric_path, 'src.tgz'))):
```
boegel: A couple of things I don't like here:

* wrapping the if condition (especially while there's another one below)
* hardcoding the `src.tgz` in two places
* no logging why we're skipping `libfabric` rebuild even though it's requested

fixed in https://github.com/ComputeCanada/easybuild-easyblocks/pull/5

PR #1897, opened 2019-12-24
also copy component patches to self.cfg in Bundle generic easyblock

### LINKED ISSUE #1893: Bundle Easyblock does not put patches into root config
At https://github.com/easybuilders/easybuild-easyblocks/blob/b99cc5a4dfb98cafbbd4a8827ea9bfb444724e27/easybuild/easyblocks/generic/bundle.py#L156 the patches checksums are added to the root, but the patches are not which makes e.g. the unit tests fail, see https://github.com/easybuilders/easybuild-easyconfigs/pull/9546

Should the patches be added? From the logic in the check_checksums it seems: yes

boegel:
fixes #1893

This doesn't affect the way in which a bundle of components is installed, but it's important for the checksum check we do in the easyconfigs test suite (see #1893)...

### DIFF
--- a/easybuild/easyblocks/generic/bundle.py
+++ b/easybuild/easyblocks/generic/bundle.py
@@ -74,6 +74,8 @@ class Bundle(EasyBlock):
         # list of sources for bundle itself *must* be empty
         if self.cfg['sources']:
             raise EasyBuildError("List of sources for bundle itself must be empty, found %s", self.cfg['sources'])
+        if self.cfg['patches']:
+            raise EasyBuildError("List of patches for bundle itself must be empty, found %s", self.cfg['patches'])
 
         # disable templating to avoid premature resolving of template values
         self.cfg.enable_templating = False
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/generic/bundle.py
```diff
@@ -109,7 +111,7 @@ def __init__(self, *args, **kwargs):
             comp_cfg['easyblock'] = None
 
             # reset list of sources/source_urls/checksums
-            comp_cfg['sources'] = comp_cfg['source_urls'] = comp_cfg['checksums'] = []
+            comp_cfg['sources'] = comp_cfg['source_urls'] = comp_cfg['checksums'] = comp_cfg['patches'] = []
```
Flamefire: This is a serious bug. It sets all of them to the same list! So if you do `comp_cfg['sources'].append('foo')` it will also append it to the others unless some copy trickery is done

PR #1901, opened 2019-12-29
add new easyblock for Desmond

smoors:


easybuild/easyblocks/d/desmond.py
```diff
@@ -0,0 +1,80 @@
+##
+# Copyright 2009-2019 Ghent University
```
boegel: @smoors please bump to `2020`

easybuild/easyblocks/d/desmond.py
```diff
@@ -0,0 +1,80 @@
...
+from easybuild.tools.filetools import rmtree2
+from easybuild.tools.run import run_cmd_qa
+
+
+class EB_Desmond(Tarball):
+    """Support for building/installing Desmond."""
+
+    def install_step(self):
+        """Custom install procedure for Desmond."""
+        rmtree2(self.installdir)
```
boegel: please use `remove_dir` instead (and log that you're removing the install directory)

boegel:
@smoors I think you were still getting the Hound CI message for Black because your `desmond` branch didn't include the updated `setup.cfg` from #1769 yet?

cc @Flamefire

easybuild/easyblocks/d/desmond.py
```diff
@@ -0,0 +1,85 @@
...
+        guesses = super(EB_Desmond, self).make_module_req_guess()
+
+        guesses["PATH"] = [""]
+
+        return guesses
+
+    def make_module_extra(self):
+        """Set up SCHRODINGER environment variable"""
+        txt = super(EB_Desmond, self).make_module_extra()
+        txt += self.module_generator.set_environment("SCHRODINGER", self.installdir)
```
akesandgren: This conflicts with what the Schrodinger EC does with the SCHRODINGER env var.
Which is:
modextrapaths = {'SCHRODINGER': ''}

Flamefire: Are they related? If not how come there is SCHRODINGER in Desmont?

Flamefire:
> I think you were still getting the Hound CI message for Black because your desmond branch didn't include the updated setup.cfg from #1769 yet?

I agree. I suggest a rebase on develop

akesandgren:
And I think this can just as easily be done in a simple Easyblock in the same way the Schrodinger EC works. It is the same INSTALL script and thus takes the same parameters.

So a simple Binary based EC with
```
install_cmd = 'tar xf Schrodinger_Suites_%(version)s%(versionsuffix)s.tar && '
install_cmd += 'cd Schrodinger_Suites_%(version)s%(versionsuffix)s && '
install_cmd += './INSTALL -d . -s %(installdir)s -k %(builddir)s -t %(installdir)s/thirdparty -b *'
```

Changed to desmond instead should work.

See https://github.com/easybuilders/easybuild-easyconfigs/pull/762 which never got merged.
We're using it.

But since we now have two things that basically need to be handled the same way we might have use for an easyblock for this. But it should not be called Desmond, Schrodinger is the more appropriate name for it if it is to handle both.

easybuild/easyblocks/d/desmond.py
akesandgren: The README for Desmond says:
```
In addition to the files provided by Schrodinger, this distribution
contains the following.
```
and also
```
Set SCHRODINGER environment variable
```

Parts of the stuff in Desmond comes from the Schrodinger distribution, (Maestro).

Basically Schrodinger and Desmond must be conflicting with each other in the module file due to this.

There is an academic free Desmond also a version from Schrodinger (for commercial licensing)

smoors:
ugh, so there is an option for non-interactive install. jeuj for undocumented features :/
ok for me to make an easyblock that handles both (not sure if/when I'll have time to do that).

PR #1904, opened 2020-01-04
iccifort: remove LIBRARY_PATH entries already known to icc et al

bartoldeman:
This makes it consistent with Intel's own script that don't set
LIBRARY_PATH either (except for aux components such as TBB)

But only do this for iccifort (single installation) so that in a split
installation icc can still find ifort's libraries and vice versa.

### DIFF
--- a/easybuild/easyblocks/i/iccifort.py
+++ b/easybuild/easyblocks/i/iccifort.py
@@ -61,4 +61,12 @@ class EB_iccifort(EB_ifort, EB_icc):
         # including it causes problems, e.g. with complex.h and std::complex
         # cfr. https://software.intel.com/en-us/forums/intel-c-compiler/topic/338378
         # whereas EB_ifort adds 'include' but that's only needed if icc and ifort are separate
-        return EB_icc.make_module_req_guess(self)
+        guesses = EB_icc.make_module_req_guess(self)
+
+        # remove entries from LIBRARY_PATH that icc and co already know about at compile time
+        # only do this for iccifort merged installations so that icc can still find ifort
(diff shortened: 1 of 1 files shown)


boegel:
Tested with standalone `iccifort` that is part of `intel/2019b` and `intel/2020.00` + a couple of installation using that `iccifort` as toolchain, no problems encountered, so good to go in...

Thanks @bartoldeman!

boegel:
@bartoldeman I think this broke something after all...

The installation of `BLAST+-2.9.0-iimpi-2019b.eb` is failing for me when the `iccifort` module is not including the `compiler/lib/intel64` subdirectory in `$LIBRARY_PATH` with:

```
ld.gold: error: cannot find -liomp5
```

The problem doesn't occur when the `compiler/lib/intel64` path is still being added to `$LIBRARY_PATH` by the `iccifort` module...

boegel:
After taking a closer look, I noticed the failing link command is using `gcc` rather than the expected `icc` or `mpiicc`, so I guess that's the real problem with building `BLAST+` with `iimpi/2019b`, and this change just exposes that problem better...

boegel:
To avoid a "wisdom of the ancients" situation: `gcc` was only used to build a small helper utility during the build of `BLAST+`, and `$LIBS` (containing `-liomp5`) was being picked up, which led to the reported `cannot find `-liomp5` error.

Fixed by unsetting `$LIBS`, see https://github.com/easybuilders/easybuild-easyconfigs/pull/9795/commits/4c400333c9ceff512a5d1d85c478eae721e388c5

PR #1908, opened 2020-01-06
define $EB_PYTHON in module for EasyBuild installation, to make sure correct Python version is used at runtime

boegel:
If EasyBuild is being installed using `eb` that is running on top of a Python version that is anything else than the system `python` command, we need to make sure `$EB_PYTHON` is set so the `eb` command will use the correct `python` binary (basically the one it was installed with).

It should be OK to simply always define `$EB_PYTHON` in the generated `EasyBuild` module...


ocaisa:
What happens if the python installation at `EB_PYTHON` is removed? Does it fall back to whatever is in the path?

ocaisa:
Never mind, I looked it up and that case is covered already.

boegel:
@ocaisa Indeed, then `command -v /path/to/python` will fail, and `eb` will move on to the next `python*` command it can find.

PR #1915, opened 2020-01-13
add custom easyconfig parameter 'buildtype' to generic CMakeMake easyblock

Flamefire:
Heavily based on code from Ward Poelmans from #911

IMO setting the buildtype is valuable to not have to work with the quite verbose configopts.

Note that configopts is passed AFTER this option so it can still override CMAKE_BUILD_TYPE and that there is NO change to the default behavior: Nothing is added unless buildtype is set


Flamefire:
@wpoely86 Hope you don't mind me taking this, I added your name to the commit message

I also made it more minimal and simpler

wpoely86:
No problem!

Flamefire:
Follow-up: Replace and validate all usages of CMAKE_BUILD_TYPE in EBs, I've seen "release" and "RELEASE" which is likely wrong

PR #1916, opened 2020-01-13
Do not sanity check on MATLAB compiler

crubb:
I would suggest removing the sanity checks on the MATLAB compiler, since it is a separate license (which for example is not included at my institution).


boegel:
Hmm, this should've been part of #1514...

Makes sense, so going in, thanks @crubb!

PR #1917, opened 2020-01-14
Fix MPI-CXX dependency of PETSc

Flamefire:
Due to SuperLU (automatically downloaded and build with PETSc) using mpicxx
the resulting libpetsc.so will have a dependency on C++ MPI symbols causing
failures in C applications linking against it even though those symbols are never used.

This removes the include to the mpi cxx extension inside openmpis mpi.h.
Downside is that all consumers of PETSc will have this define too as
PETSc saves its used command line flags.

IMO this is OK as to my knowledge nobody uses those extensions anyway, especially in the PETSc consumer list (e.g. SLEPc)

edit (by @boegel): see also https://github.com/easybuilders/easybuild-easyconfigs/issues/9610

### DIFF
--- a/easybuild/easyblocks/p/petsc.py
+++ b/easybuild/easyblocks/p/petsc.py
@@ -112,13 +112,16 @@ class EB_PETSc(ConfigureMake):
             self.cfg.update('configopts', '--with-fc="%s"' % os.getenv('F90'))
 
             # compiler flags
+            # Don't build with MPI c++ bindings as this leads to a hard dependency
+            # on libmpi and libmpi_cxx even for C code and non-MPI code
+            cxxflags = os.getenv('CXXFLAGS') + ' -DOMPI_SKIP_MPICXX -DMPICH_SKIP_MPICXX'
             if LooseVersion(self.version) >= LooseVersion("3.5"):
                 self.cfg.update('configopts', '--CFLAGS="%s"' % os.getenv('CFLAGS'))
-                self.cfg.update('configopts', '--CXXFLAGS="%s"' % os.getenv('CXXFLAGS'))
+                self.cfg.update('configopts', '--CXXFLAGS="%s"' % cxxflags)
                 self.cfg.update('configopts', '--FFLAGS="%s"' % os.getenv('F90FLAGS'))
             else:
                 self.cfg.update('configopts', '--with-cflags="%s"' % os.getenv('CFLAGS'))
-                self.cfg.update('configopts', '--with-cxxflags="%s"' % os.getenv('CXXFLAGS'))
+                self.cfg.update('configopts', '--with-cxxflags="%s"' % cxxflags)
                 self.cfg.update('configopts', '--with-fcflags="%s"' % os.getenv('F90FLAGS'))
 
             if not self.toolchain.comp_family() == toolchain.GCC:  #@UndefinedVariable
--- a/easybuild/easyblocks/p/petsc.py
+++ b/easybuild/easyblocks/p/petsc.py
@@ -36,10 +36,12 @@ import easybuild.tools.toolchain as toolchain
 from easybuild.easyblocks.generic.configuremake import ConfigureMake
 from easybuild.framework.easyconfig import BUILD, CUSTOM
 from easybuild.tools.build_log import EasyBuildError
-from easybuild.tools.filetools import symlink
+from easybuild.tools.filetools import symlink, apply_regex_substitutions
 from easybuild.tools.modules import get_software_root
 from easybuild.tools.run import run_cmd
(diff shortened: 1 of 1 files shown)


boegel:
@Flamefire Can you clarify what you mean by "PETSc consumers"?

Is that anything depending on PETSc?

If so, this change will break other stuff, only to support building PETSc on top of OpenMPI configured with `--enable-mpi-cxx`, which is a site customization (since the `OpenMPI` easyconfigs in the central repo don't use that configure option).

So, we should be more careful here and make sure we apply the necessary updates to anything that uses `PETSc`...

Flamefire:
Yes. PETSc has a config file with flags that is used by e.g. SLEPc to build SLEPc.

The problem is not that easy: The goal is to have a PETSc which does not depend on mpi-cxx. Building PETSc is fine without this. But (especially C) consumers fail to try-link against PETSc due to the mpi-cxx dependency which is unnecessarily pulled in by PETSc through SuperLU through OpenMPIs mpi.h

So yes this will break consumers of PETSc using the PETSc config params and (trying) to use mpi-cxx. However it will fix all consumers that do NOT use mpicxx (e.g. C applications). There is at least SLEPc matching that last category and none I know of matching the first. Hence me voting for this.

Better solution: Use that flag only for building PETSc (to be exact: the included/downloaded SuperLU) but not propagating it through PETSc to consumers. That would be correct. But I don't know of any possibility to do it besides building SuperLU via EC (cause PETSc-downloaded-deps to be a lie)

It is **not** a solution to fix consumers, because PETSc is build wrongly (has the mpicxx dependency which it should not) and may not even be possible.

Maybe patch the generated files and remove this flag after installation? How?

boegel:
Patching the generated files that pass on `-DOMPI_SKIP_MPICXX` sounds reasonable.

Do you know if the SuperLU they pull in is a standard version, or a patched one?

If it's a standard version, I'd prefer taking the approach of providing a SuperLU built with EasyBuild...

Flamefire:
No idea, I guess standard. Problem is: It seems PETSc-downloaded-deps was specifically created NOT to use EC modules judging from the naming.

boegel:
Indeed, but I think the main reason for that was two-fold: effort required, and the fact that some dependencies are patched.
We now have a good reason to flesh out `SuperLU` specifically (unless PETSc uses a patched version).

Flamefire:
I investigated again: It does not use patches as far as I can tell. However it is not SuperLU but SuperLU_DIST which causes the failure (as I've written above but confused myself due to not knowing there are multiple flavors). The Download and configuration process used by PETSc is easy though, so I guess I can come up with a suitable EB and EC (or maybe only EC for now and see if that works already). This could end up being quite some work though, so not sure if it is reasonable after all.

easybuild/easyblocks/p/petsc.py
```diff
@@ -112,13 +112,16 @@ def configure_step(self):
             self.cfg.update('configopts', '--with-fc="%s"' % os.getenv('F90'))
 
             # compiler flags
+            # Don't build with MPI c++ bindings as this leads to a hard dependency
+            # on libmpi and libmpi_cxx even for C code and non-MPI code
+            cxxflags = os.getenv('CXXFLAGS') + ' -DOMPI_SKIP_MPICXX'
```
akesandgren: The correct flag depends on which MPI is used.
OMPI_SKIP_MPICXX for OpenMPI and MPICH_SKIP_MPICXX for IMPI (and any other MPICH based MPI)
So better set both and be done with it.

Flamefire:
```suggestion
            cxxflags = os.getenv('CXXFLAGS') + ' -DOMPI_SKIP_MPICXX -DMPICH_SKIP_MPICXX'
```

akesandgren:
Was this everything needed to fix this part of the problem?
Have not been able to keep track of all the discussions going on the past weeks... only seen lots of stuff on petsc and other things and i'm probably mixing them up...

Flamefire:
@boegel suggested to patch out these added flags so consumers can decide if they want MPICXX. I'll check how hard that would be

Flamefire:
@akesandgren I rebased to develop and added a commit which removes the flags from petscvariables (which AFAIK is the file used for packages using PETSc) so consumers can use MPICXX if the system / module is configured for it

--> Ready to go

akesandgren:
Trying to kick Hound

akesandgren:
Kicked

akesandgren:
@Flamefire Please repush to kick Hound again.

Flamefire:
@akesandgren Done and passed

akesandgren:
Going in, thanks @Flamefire!

PR #1918, opened 2020-01-14
Build HDF5 without MPI C++ extension

Flamefire:
This avoids the direct dependency on MPI/MPI cxx for consumers even for C libraries

I verified that this flag is not passed to consumer by means of `h5cc -show` and `h5c++ -show` or in fact anything else (grep used)

### DIFF
--- a/easybuild/easyblocks/h/hdf5.py
+++ b/easybuild/easyblocks/h/hdf5.py
@@ -84,6 +84,9 @@ class EB_HDF5(ConfigureMake):
             mpich_mpi_families = [toolchain.INTELMPI, toolchain.MPICH, toolchain.MPICH2, toolchain.MVAPICH2]
             if self.toolchain.mpi_family() in mpich_mpi_families:
                 self.cfg.update('buildopts', 'CXXFLAGS="$CXXFLAGS -DMPICH_IGNORE_CXX_SEEK"')
+            # Skip MPI cxx extensions to avoid hard dependency
+            if self.toolchain.mpi_family() in toolchain.OPENMPI:
+                self.cfg.update('buildopts', 'CXXFLAGS="$CXXFLAGS -DOMPI_SKIP_MPICXX"')
         else:
             self.cfg.update('configopts', "--disable-parallel")
 


easybuild/easyblocks/h/hdf5.py
```diff
@@ -84,6 +84,9 @@ def configure_step(self):
             mpich_mpi_families = [toolchain.INTELMPI, toolchain.MPICH, toolchain.MPICH2, toolchain.MVAPICH2]
             if self.toolchain.mpi_family() in mpich_mpi_families:
                 self.cfg.update('buildopts', 'CXXFLAGS="$CXXFLAGS -DMPICH_IGNORE_CXX_SEEK"')
+            # Skip MPI cxx extensions to avoid hard dependency
+            if self.toolchain.mpi_family() in toolchain.OPENMPI:
```
boegel: Hmm, should be `==` rather than `in`?

boegel:
@Flamefire What kind of implications does this have exactly? How sure are we that nothing we build on top of `HDF5` actually requires this?

easybuild/easyblocks/h/hdf5.py
Flamefire: Oh... How did this work then? Double-checking

Flamefire:
No implications. This is only used to build HDF5 itself. By pure accident (or  better: bad design) some symbols are defined which need to be resolved by linking mpi_cxx which now are not. Consumers can still link against HDF5 and/or mpi_cxx and use what they want

easybuild/easyblocks/h/hdf5.py
Flamefire: It did work though: `== 2020-01-14 11:16:31,794 run.py:534 INFO cmd " make -j 12  CXXFLAGS="$CXXFLAGS -DOMPI_SKIP_MPICXX"  FC="mpifort" " exited with exit code 0 and output:
`

Flamefire: Fixed in https://github.com/easybuilders/easybuild-easyblocks/pull/1919

boegel: It worked because `toolchain.OPENMPI` is a string value (but you shouldn't rely on that, in theory it could change).

boegel:
Thanks for clarifying @Flamefire!

PR #1919, opened 2020-01-14
[HDF5] Fix 'in' vs '==' confusion

Flamefire:
Followup to https://github.com/easybuilders/easybuild-easyblocks/pull/1918

But not critical as both versions seem to work fine


boegel:
The `in` only works by accident (because `toolchain.OPENMPI` is a string value), but we should really use `==`.

Issue #1920, opened 2020-01-14
Dolfin build in Debug mode


Flamefire:
@JensTimmerman You introduced this change: https://github.com/easybuilders/easybuild-easyblocks/commit/de16713b1a744354e2022000c84a9f08ba3665aa#diff-ae44a1c5fa077d44e8ada75a8cabd117R47

What was the reasoning in building in debug mode (aka -O0 etc)? The comment "run cmake in debug mode" makes me think you want debug output from CMake not a debug build

JensTimmerman:
Oh, euh, this was 7 years ago. 
I remember dolfin being a particularly hard project to build and indeed I was not able to finish it and @boegel took over from me at some point. 

It is entirely possible that I enabled this flag to get some more insights into the process or avoid some weird condition and later forgot to turn it of or could not find a way to make this work without it.

boegel:
The comment tells me that @JensTimmerman was under the assumption that using `-DCMAKE_BUILD_TYPE=Debug` indeed put `cmake` in debug mode.

The only (non-archive) easyconfig we still have around for `DOLFIN` is DOLFIN-2018.1.0.post1-foss-2018a-Python-3.6.4.eb, as long as dropping the `-DCMAKE_BUILD_TYPE=Debug` doesn't break installing that easyconfig, we can change this imho.

PR #1921, opened 2020-01-14
change various easyblock deriving from CMakeMake to use 'build_type' easyconfig parameter

Flamefire:
This replaces raw usages of `-DCMAKE_BUILD_TYPE=...` in other ECs

I also added a commit which renames the option to build_type because the base class ConfigureMake has this and having both might be confusing to users. This lets the last one win, in this case CMake

### DIFF
--- a/easybuild/easyblocks/c/clang.py
+++ b/easybuild/easyblocks/c/clang.py
@@ -70,7 +70,8 @@ class EB_Clang(CMakeMake):
 
     @staticmethod
     def extra_options():
-        extra_vars = {
+        extra_vars = CMakeMake.extra_options()
+        extra_vars.update({
             'assertions': [True, "Enable assertions.  Helps to catch bugs in Clang.", CUSTOM],
             'build_targets': [None, "Build targets for LLVM (host architecture if None). Possible values: " +
                                     ', '.join(CLANG_TARGETS), CUSTOM],
@@ -84,9 +85,9 @@ class EB_Clang(CMakeMake):
             'skip_all_tests': [False, "Skip running of tests", CUSTOM],
             # The sanitizer tests often fail on HPC systems due to the 'weird' environment.
             'skip_sanitizer_tests': [True, "Do not run the sanitizer tests", CUSTOM],
-        }
-
-        return CMakeMake.extra_options(extra_vars)
+        })
+        extra_vars['build_type'][0] = 'Release'
+        return extra_vars
 
     def __init__(self, *args, **kwargs):
         """Initialize custom class variables for Clang."""
@@ -226,7 +227,6 @@ class EB_Clang(CMakeMake):
         self.log.debug("Using %s as GCC_INSTALL_PREFIX", gcc_prefix)
 
         # Configure some default options
-        self.cfg.update('configopts', "-DCMAKE_BUILD_TYPE=Release")
         if self.cfg["enable_rtti"]:
             self.cfg.update('configopts', '-DLLVM_REQUIRES_RTTI=ON')
             self.cfg.update('configopts', '-DLLVM_ENABLE_RTTI=ON')
--- a/easybuild/easyblocks/d/dirac.py
+++ b/easybuild/easyblocks/d/dirac.py
@@ -31,9 +31,7 @@ import shutil
 import tempfile
 
 import easybuild.tools.environment as env
-import easybuild.tools.toolchain as toolchain
 from easybuild.easyblocks.generic.cmakemake import CMakeMake
-from easybuild.framework.easyconfig import CUSTOM, MANDATORY
 from easybuild.tools.build_log import EasyBuildError
 from easybuild.tools.config import build_option
 from easybuild.tools.run import run_cmd
@@ -41,6 +39,12 @@ from easybuild.tools.run import run_cmd
 
 class EB_DIRAC(CMakeMake):
     """Support for building/installing DIRAC."""
+
+    @staticmethod
+    def extra_options():
+        extra_vars = CMakeMake.extra_options()
+        extra_vars['build_type'][0] = 'release'  # TODO: Verify if this is really lowercase
+        return extra_vars
 
     def configure_step(self):
         """Custom configuration procedure for DIRAC."""
@@ -54,7 +58,7 @@ class EB_DIRAC(CMakeMake):
                 raise EasyBuildError("Failed to remove existing install directory %s: %s", self.installdir, err)
 
         self.cfg['separate_build_dir'] = True
-        self.cfg.update('configopts', "-DENABLE_MPI=ON -DCMAKE_BUILD_TYPE=release")
+        self.cfg.update('configopts', "-DENABLE_MPI=ON")
 
         # complete configuration with configure_method of parent
         super(EB_DIRAC, self).configure_step()
--- a/easybuild/easyblocks/d/dolfin.py
+++ b/easybuild/easyblocks/d/dolfin.py
@@ -47,6 +47,12 @@ from easybuild.tools.systemtools import get_shared_lib_ext
 class EB_DOLFIN(CMakePythonPackage):
     """Support for building and installing DOLFIN."""
 
+    @staticmethod
+    def extra_options():
+        extra_vars = CMakePythonPackage.extra_options()
+        extra_vars['build_type'][0] = 'Release'
+        return extra_vars
+
     def __init__(self, *args, **kwargs):
         """Initialize class variables."""
         super(EB_DOLFIN, self).__init__(*args, **kwargs)
@@ -80,9 +86,6 @@ class EB_DOLFIN(CMakePythonPackage):
         self.cfg.update('configopts', '-DCMAKE_CXX_FLAGS="%s"' % cxxflags)
         self.cfg.update('configopts', '-DCMAKE_Fortran_FLAGS="%s"' % fflags)
 
-        # run cmake in debug mode
-        self.cfg.update('configopts', '-DCMAKE_BUILD_TYPE=Debug')
-
         # set correct compilers to be used at runtime
         self.cfg.update('configopts', '-DMPI_C_COMPILER="$MPICC"')
         self.cfg.update('configopts', '-DMPI_CXX_COMPILER="$MPICXX"')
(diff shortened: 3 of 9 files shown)


easybuild/easyblocks/o/opencv.py
```diff
@@ -90,9 +91,6 @@ def prepare_step(self, *args, **kwargs):
     def configure_step(self):
         """Custom configuration procedure for OpenCV."""
 
-        if 'CMAKE_BUILD_TYPE' not in self.cfg['configopts']:
```
boegel: Shouldn't we add this check whether `CMAKE_BUILD_TYPE` is already specified` in `configopts` in `CMakeMake`?

easybuild/easyblocks/p/psi.py
```diff
@@ -56,6 +56,7 @@ def __init__(self, *args, **kwargs):
         self.psi_srcdir = None
         self.install_psi_objdir = None
         self.install_psi_srcdir = None
+        self.cfg['build_type'] = 'Release'
```
boegel: Same issue w.r.t. default in `eb -a`, change via `extra_options`?

easybuild/easyblocks/p/psi.py
```diff
@@ -143,9 +144,7 @@ def configure_step(self):
             if self.name == 'PSI4' and LooseVersion(self.version) >= LooseVersion("1.2"):
                 self.log.info("Remove the CMAKE_BUILD_TYPE test in PSI4 source and the downloaded dependencies!")
                 self.log.info("Use PATCH_COMMAND in the corresponding CMakeLists.txt")
-                self.cfg.update('configopts', ' -DCMAKE_BUILD_TYPE=EasyBuildRelease')
-            else:
-                self.cfg.update('configopts', ' -DCMAKE_BUILD_TYPE=Release')
+                self.cfg['build_type'] = 'EasyBuildRelease'
```
boegel: This should be in the `if`?

easybuild/easyblocks/d/dirac.py
```diff
@@ -54,7 +54,7 @@ def configure_step(self):
                 raise EasyBuildError("Failed to remove existing install directory %s: %s", self.installdir, err)
 
         self.cfg['separate_build_dir'] = True
-        self.cfg.update('configopts', "-DENABLE_MPI=ON -DCMAKE_BUILD_TYPE=release")
+        self.cfg['build_type'] = 'release'  # TODO: Verify if this is really lowercase
```
boegel: As far as I know the values passed to CMake options are case-insensitive (but you should confirm that)

Flamefire: Not always. If it is used directly casing does matter (in a strequal e.g.)
I verified the other uses but can't for this one as access to source requires a license

easybuild/easyblocks/d/dolfin.py
```diff
@@ -54,6 +54,8 @@ def __init__(self, *args, **kwargs):
         self.boost_dir = None
         self.saved_configopts = None
         self.cfg['separate_build_dir'] = True
+        # run cmake in debug mode
+        self.cfg['build_type'] = 'Debug'
```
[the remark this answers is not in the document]
Flamefire: Yes. I opened an issue requesting clarification why this was added

easybuild/easyblocks/n/netcdf.py
```diff
@@ -68,10 +68,7 @@ def configure_step(self):
             ConfigureMake.configure_step(self)
 
         else:
-            if self.toolchain.options['debug']:
-                self.cfg.update('configopts', '-DCMAKE_BUILD_TYPE=DEBUG ')
-            else:
-                self.cfg.update('configopts', '-DCMAKE_BUILD_TYPE=RELEASE -DCMAKE_C_FLAGS_RELEASE="-DNDEBUG " ')
+            self.cfg['build_type'] = 'Debug' if self.toolchain.options.get('debug', None) else 'Release'
```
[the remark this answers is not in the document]
Flamefire: Cmake does add it by default. Can double check

easybuild/easyblocks/o/opencv.py
Flamefire: Not required. This is added first so configopts can overwrite it

easybuild/easyblocks/p/psi.py
Flamefire: It is, the else was removed

easybuild/easyblocks/t/trilinos.py
```diff
@@ -98,11 +98,8 @@ def configure_step(self):
         else:
             self.cfg.update('configopts', "-DBUILD_SHARED_LIBS:BOOL=OFF")
 
-        # release or debug get_version
-        if self.toolchain.options['debug']:
-            self.cfg.update('configopts', "-DCMAKE_BUILD_TYPE:STRING=DEBUG")
-        else:
-            self.cfg.update('configopts', "-DCMAKE_BUILD_TYPE:STRING=RELEASE")
+        # release or debug gversion
+        self.cfg['build_type'] = 'Debug' if self.toolchain.options.get('debug', None) else 'Release'
```
Flamefire: @boegel I think it makes sense to do this by default. Is the toolchainopt debug used for anything else?

Flamefire: @boegel I think it makes sense to do this by default. Is the toolchainopt debug used for anything else?

Flamefire: @boegel I think it makes sense to do this by default. Is the toolchainopt debug used for anything else?

easybuild/easyblocks/c/clang.py
```diff
@@ -96,6 +96,7 @@ def __init__(self, *args, **kwargs):
         self.llvm_obj_dir_stage2 = None
         self.llvm_obj_dir_stage3 = None
         self.make_parallel_opts = ""
+        self.cfg['build_type'] = 'Release'
```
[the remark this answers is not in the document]
Flamefire: Ok. Does it still need to by python 2 compatible?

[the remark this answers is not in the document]
boegel: Yes, EasyBuild is still compatible with Python 2.6, 2.7 and 3.5+ (see https://easybuild.readthedocs.io/en/latest/Python-2-3-compatibility.html)

easybuild/easyblocks/o/opencv.py
boegel: OK

easybuild/easyblocks/t/trilinos.py
boegel: We may use it in some places to inject `-g` into `$CXXFLAGS` & co for example.

Always switching to `-DDCMAKE_BUILD_TYPE=Debug` when the `debug` toolchain option is enabled makes sense I guess...

easybuild/easyblocks/d/dolfin.py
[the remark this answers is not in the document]
boegel: see #1920

easybuild/easyblocks/n/netcdf.py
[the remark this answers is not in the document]
boegel: You mean it adds it when `-DCMAKE_BUILD_TYPE=Release` is used, not always, right?

easybuild/easyblocks/d/dolfin.py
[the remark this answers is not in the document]
Flamefire: From the discussion I conclude that this was indeed a mistake. I'm currently building it with Release instead and will keep that if that works

easybuild/easyblocks/n/netcdf.py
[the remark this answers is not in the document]
Flamefire: Short answer: Correct

To be exact I meant that it adds it to CMAKE_C/CXX_FLAGS_RELEASE by default which is used for the release build type. Yes there is a difference but doesn't matter for "make" as the generator.

easybuild/easyblocks/t/trilinos.py
Flamefire: The question would be: How to handle 'debug' vs 'build_type'? My proposal: if build_type is not None, it overwrites 'debug'. Else it is used as it is here: Debug or Release
But this borders into the discussion in https://github.com/easybuilders/easybuild-easyblocks/pull/911#issuecomment-564466693: What for projects failing to work in "Release" mode? TBH I think this is a non-issue as optflags are usually used so only difference is (almost always) the added `-DNDEBUG`

To be conservative one could treat debug=None as not doing anything which makes it:
- if build_type is not None, use it
- else if debug is not None, use Release or Debug
- don't use anything

But then: Which ECs use `debug: False` explicitly? So I'd treat debug=None as false and if anything breaks, set it in the EC (again because I see very small chances for anything in ECs break)

Flamefire: On the other hand current behavior for e.g. gromacs and netcdf is that toolchainopts has preference over build_type. Maybe use that?

Flamefire:
Rebased and updated please reverify. I'd also like to get the behavior with toolchainopts.debug defined here. See https://github.com/easybuilders/easybuild-easyblocks/pull/1921#discussion_r366776798 My current stance would be that if toolchainopts.debug==True then build_type is set to Debug. That would match similar uses in existing ECs except that it still can be overwritten by passing it in configopts

easybuild/easyblocks/d/dolfin.py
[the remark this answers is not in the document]
Flamefire: DOLFIN-2018.1.0.post1-foss-2018a-Python-3.6.4.eb build successfully with build_type=Release

easybuild/easyblocks/n/netcdf.py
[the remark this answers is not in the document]
Flamefire: Verified by checking output of a test build: `-- CMAKE_C_FLAGS_RELEASE:   -O3 -DNDEBUG`

Flamefire:
I ingested this into #1929 as changes requested there create conflicts with these changes and the scope became hard to tell apart. please continue discussion there

PR #1923, opened 2020-01-14
ignore "Black would make changes" produced by flake8-black

boegel:
see also https://github.com/easybuilders/easybuild-framework/pull/3162


akesandgren:
Going in, thanks @boegel!

PR #1925, opened 2020-01-14
fixes for GAMESS-US easyblock (WIP)

boegel:
(not finished, but it's getting on my nerves having this in my working copy)

cc @guacke

### DIFF
--- a/easybuild/easyblocks/g/gamess_us.py
+++ b/easybuild/easyblocks/g/gamess_us.py
@@ -37,18 +37,19 @@ EasyBuild support for building and installing GAMESS-US, implemented as an easyb
 import fileinput
 import glob
 import os
+import random
 import re
 import shutil
 import sys
 import tempfile
 
-import easybuild.tools.toolchain as toolchain
 from easybuild.framework.easyblock import EasyBlock
 from easybuild.framework.easyconfig import CUSTOM, MANDATORY
 from easybuild.tools.build_log import EasyBuildError
 from easybuild.tools.config import build_option
-from easybuild.tools.filetools import mkdir, read_file, write_file
+from easybuild.tools.filetools import change_dir, read_file, write_file
 from easybuild.tools.modules import get_software_root, get_software_version
+from easybuild.tools.py2vs3 import ascii_letters
 from easybuild.tools.run import run_cmd, run_cmd_qa
 from easybuild.tools.systemtools import get_platform_name
 from easybuild.tools import toolchain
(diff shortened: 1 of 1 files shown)


branfosj:
I've been looking at GAMESS-US for one of our users. I've not successfully completed it yet, but noticed that `maxcpus` and `maxnodes` are defined as variables in the EasyBlock and defined in the EasyConfigs. However, they do not seem to ever be used. I've added the following to the WIP I have for the EasyBlock:
```
        compddi = os.path.join(self.builddir, 'ddi/compddi')
        try:
            for line in fileinput.input(compddi, inplace=1, backup='.orig'):
                line = re.sub(r"^\s*set\s*MAXCPUS=.*", r"set MAXCPUS=%s" % self.cfg['maxcpus'], line)
                line = re.sub(r"^\s*set\s*MAXNODES=.*", r"set MAXNODES=%s" % self.cfg['maxnodes'], line)
                sys.stdout.write(line)
        except IOError as err:
            raise EasyBuildError("Failed to patch %s: %s", compddi, err)
```

easybuild/easyblocks/g/gamess_us.py
```diff
@@ -78,7 +79,10 @@ def __init__(self, *args, **kwargs):
             self.testdir = tempfile.mkdtemp()
             # make sure test dir doesn't contain [ or ], rungms csh script doesn't handle that well ("set: No match")
             if re.search(r'[\[\]]', self.testdir):
-                raise EasyBuildError("Temporary dir for tests '%s' will cause problems with rungms csh script", self.testdir)
+                raise EasyBuildError("Temporary dir for tests '%s' will cause problems with rungms csh script",
```
akesandgren, line 82: Should probably tell the user why it will cause problems.

easybuild/easyblocks/g/gamess_us.py
```diff
@@ -188,22 +195,48 @@ def configure_step(self):
...
+            salt = ''.join(random.choice(ascii_letters) for i in range(10))
+            self.scratch_dir = os.path.join(self.scratch_dir, self.name + '-' + salt)
+
+            # make sure that specified directory is accessible via 'ssh', since rungms relies on that;
+            # touch a file, and then use "ssh localhost ls" to see if it's accessible;
+            # we must use shell commands for this, since scratch_dir may contain names of environment variables
+            test_file = os.path.join(self.scratch_dir, 'test.txt')
+            run_cmd("mkdir -p %s" % self.scratch_dir)
+            run_cmd("touch %s" % test_file)
+            test_cmd = 'ssh localhost "ls %s"' % test_file
```
akesandgren, line 213: This is not a good thing to do. Think building in a container...
If rungms tries to do any ssh-based stuff it is broken and should be fixed.
I don't think I had that problem back in the days I built this by hand.

akesandgren (changes requested):

lexming:
Closing as superseded by https://github.com/easybuilders/easybuild-easyblocks/pull/3047

PR #616, opened 2015-05-28
Default out of tree build for CMakeMake

wpoely86:
Needs testing

PR #1933, opened 2020-01-16
enable out of tree build by default in CMakeMake easyblock

Flamefire:
This supersedes #616 and includes #1929 which in turn includes an improved version of #616 concerning defaults for out-of-tree builds while this contains the setting only.

Better: It is intended that 1929 is merged first, and this rebased on the merged result (currently it is rebased on 1929) to see the small changes made by this PR. 1929 contains explicit setting of out-of-tree builds where required which is need for this PR to work.

I already did extensive tests with existing EasyConfigs using True as the default. The ECs using CMake are 
[cmakeECs.txt](https://github.com/easybuilders/easybuild-easyblocks/files/4071432/cmakeECs.txt) as found by grepping the ECs and EBs. I chose a subset of those to test excluding ones using old Intel toolchains as I don't have old intel compilers. I previously excluded all intel ECs hence the names below

So:
- Tested ECs: [cmakeECs-non-intel.txt](https://github.com/easybuilders/easybuild-easyblocks/files/4090324/cmakeECs-non-intel.txt)
- Successful ECs: [cmakeECs-OK.txt](https://github.com/easybuilders/easybuild-easyblocks/files/4090325/cmakeECs-OK.txt)
- Failed ECs with reason: [cmakeECs-Error.txt](https://github.com/easybuilders/easybuild-easyblocks/files/4071461/cmakeECs-Error.txt)

From this extensive test the failure introduced was always a build failure so can be easily detected. To the best of my knowledge (according to the tests) there are no failures introduced for existing ECs. Most non-EB ECs should be caught by the changed EasyBlocks that set the out-of-tree build to False where required.

From those results I see no reason NOT to enable it by default. I'm against using the deprecating behavior as this would force all ECs/EBs to include setting the new default which will be rendered superflous, I Included it nonetheless to spark some discussion but would like to remove it.

### DIFF
--- a/easybuild/easyblocks/generic/cmakemake.py
+++ b/easybuild/easyblocks/generic/cmakemake.py
@@ -81,7 +81,7 @@ class CMakeMake(ConfigureMake):
                                  "Defaults to 'Release' or 'Debug' depending on toolchainopts[debug]", CUSTOM],
             'configure_cmd': [DEFAULT_CONFIGURE_CMD, "Configure command to use", CUSTOM],
             'srcdir': [None, "Source directory location to provide to cmake command", CUSTOM],
-            'separate_build_dir': [False, "Perform build in a separate directory", CUSTOM],
+            'separate_build_dir': [True, "Perform build in a separate directory", CUSTOM],
         })
         return extra_vars
 
@@ -118,7 +118,7 @@ class CMakeMake(ConfigureMake):
 
         setup_cmake_env(self.toolchain)
 
-        if builddir is None and self.cfg.get('separate_build_dir', False):
+        if builddir is None and self.cfg.get('separate_build_dir', True):
             builddir = os.path.join(self.builddir, 'easybuild_obj')
 
         if builddir:
--- a/easybuild/easyblocks/generic/cmakemake.py
+++ b/easybuild/easyblocks/generic/cmakemake.py
@@ -94,7 +94,7 @@ class CMakeMake(ConfigureMake):
     def lib_ext(self):
         """Return the extension for libraries build based on `build_shared_libs` or None if that is unset"""
         if self._lib_ext is None:
-            build_shared_libs = self.cfg['build_shared_libs']
+            build_shared_libs = self.cfg.get('build_shared_libs')
             if build_shared_libs:
                 self._lib_ext = get_shared_lib_ext()
             elif build_shared_libs is not None:
@@ -108,7 +108,7 @@ class CMakeMake(ConfigureMake):
     @property
     def build_type(self):
         """Build type set in the EasyConfig with default determined by toolchainopts"""
-        build_type = self.cfg['build_type']
+        build_type = self.cfg.get('build_type')
         if build_type is None:
             build_type = 'Debug' if self.toolchain.options.get('debug', None) else 'Release'
         return build_type
@@ -138,7 +138,7 @@ class CMakeMake(ConfigureMake):
         options = ['-DCMAKE_INSTALL_PREFIX=%s' % self.installdir]
 
         if '-DCMAKE_BUILD_TYPE=' in self.cfg['configopts']:
-            if self.cfg['build_type'] is not None:
+            if self.cfg.get('build_type') is not None:
                 self.log.warning('CMAKE_BUILD_TYPE is set in configopts. Ignoring build_type')
         else:
             options.append('-DCMAKE_BUILD_TYPE=%s' % self.build_type)
@@ -149,7 +149,7 @@ class CMakeMake(ConfigureMake):
 
         # Set flag for shared libs if requested
         # Not adding one allows the project to choose a default
-        build_shared_libs = self.cfg['build_shared_libs']
+        build_shared_libs = self.cfg.get('build_shared_libs')
         if build_shared_libs is not None:
             # Contrary to other options build_shared_libs takes precedence over configopts which may be unexpected.
             # This is to allow self.lib_ext to be determined correctly.
--- a/easybuild/easyblocks/generic/cmakemake.py
+++ b/easybuild/easyblocks/generic/cmakemake.py
@@ -130,7 +130,8 @@ class CMakeMake(ConfigureMake):
 
         if srcdir is None:
             if self.cfg.get('srcdir', None) is not None:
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/generic/cmakemake.py
```diff
@@ -86,8 +90,15 @@ def configure_step(self, srcdir=None, builddir=None):
 
         setup_cmake_env(self.toolchain)
 
-        if builddir is None and self.cfg.get('separate_build_dir', False):
-            builddir = os.path.join(self.builddir, 'easybuild_obj')
+        if builddir is None:
+            separate_build_dir = self.cfg['separate_build_dir']
+            if separate_build_dir is None:
+                msg = "Easyconfig parameter 'separate_build_dir' (custom to CMakeMake easyblock) has not set "
```
akesandgren: "has not been set " (or "is not set ")

Flamefire: I'd rather remove this commit entirely, see reasoning in description
It would also introduce needless depreciation warnings to almost all ecs

Flamefire: I removed this commit. It leads to more trouble than it is worth: It would print a deprecation warning for pretty much all EasyConfigs. As mentioned all tested EasyConfigs work with the new default and if a third-party EC fails it is very easy to fix. Having said that: It shouldn't fail in the first place as this would be a sign for something non-standard happening there.

boegel:
@Flamefire Please sync this with `develop` now #1929 is merged, to get a good view on the remaining changes.

Should be good to go, already tested extensively...

Flamefire:
Done. Please don't forget to merge https://github.com/easybuilders/easybuild-easyconfigs/pull/9738

easybuild/easyblocks/generic/cmakemake.py
```diff
@@ -118,7 +118,7 @@ def configure_step(self, srcdir=None, builddir=None):
 
         setup_cmake_env(self.toolchain)
 
-        if builddir is None and self.cfg.get('separate_build_dir', False):
+        if builddir is None and self.cfg['separate_build_dir']:
```
boegel, line 121: Please keep this as it was, just change the `False` to `True` (or, if you want to ensure this stays consistent, add a `DEFAULT_SEPARATE_BUILD_DIR` class constant and use that in both places.

We're careful with this because they may be easyblocks that derive from `CMakeMake` but blindly overwrite custom easyconfig parameters in `extra_options`, causing `'separate_build_dir'` to be undefined in `self.cfg`.

That may not be the case of any of the easyblocks in the central repository, but there may be other easyblocks out there that are less carefully constructed.

```python
f builddir is None and self.cfg.get('separate_build_dir', True):

File: easybuild/easyblocks/generic/cmakemake.py [hunk 3]
```diff
@@ -130,7 +130,8 @@ def configure_step(self, srcdir=None, builddir=None):
 
         if srcdir is None:
             if self.cfg.get('srcdir', None) is not None:
-                srcdir = self.cfg['srcdir']
+                # Note that the join returns srcdir if it is absolute
+                srcdir = os.path.join(default_srcdir, self.cfg['srcdir'])
```
boegel, line 134: @Flamefire Can you clarify why this is needed?

boegel:
This change also required https://github.com/easybuilders/easybuild-easyconfigs/pull/10300 (which is fine, since that easyconfig is only there in `develop`)

easybuild/easyblocks/generic/cmakemake.py
Flamefire: > but blindly overwrite custom easyconfig parameters in extra_options, causing separate_build_dir' to be undefined in self.cfg.

Wouldn't that be a bug (to do that)? IMO it is good that this would fail in that case.

There is also a usage of e.g. `self.cfg.get('configure_cmd')` which in that case (blindly overwriting) would return `None` and fail. Same for `build_type` and the property.

So I'd rather keep this change to make people fix their broken EasyBlocks (if there are any) unless you really want that

easybuild/easyblocks/generic/cmakemake.py
Flamefire: Programs can set `srcdir` and usually do so using a relative path: https://github.com/easybuilders/easybuild-easyconfigs/pull/10300/files#diff-6573a83383dfd6ecdc9fd30fe232478cL76

This path is relative `self.cfg['start_dir']` but if we change the CWD then the below configure command can't find the start_dir.

easybuild/easyblocks/generic/cmakemake.py
boegel: fixed in 6695267

easybuild/easyblocks/generic/cmakemake.py
boegel: OK, thanks for clarifying!

boegel:
Extensively tested and approved, so going in, thanks @Flamefire!

klust:
This change, combined with the fact that the out-of-source build tree has a fixed name, wreaks havoc with Bundle-based eb-files that contain more than one package built with CMake as the second build fails because EasyBuild tries to build it in the directory of the first one.  It is of course easily circumvented by setting separate_build_dir back to its old default False in those EB-files. But if out-of-tree builds are considered important, it would be better to also encode the name/version of the package in the build directory, and even better to also encode the iteration number for builds with multiple iterations (e.g., an array of configopts) in the directory name for the build.

Flamefire:
Good point. Possible mitigations:
- Build inside a subfolder of the source. Usual workflow I see outside EB is: `mkdir build && cd build; cmake .. <other opts>; make && make install`
- Remove build folder after install step / before build step at the point where it is to be created. IMO this is a good approach as the point of an out-of-source dir is a clean build so that folder should not exist or be empty

boegel:
Ideally, an issue should be opened for the (valid) comments raised by @klust, keeping track of things to change in merged PRs is very difficult w.r.t. staying on top of things.

@klust Do you mind opening an issue for this via https://github.com/easybuilders/easybuild-easyblocks/issues/new?

PR #1934, opened 2020-01-16
[TensorFlow] Limit amount of parallelism

Flamefire:
Otherwise Bazel might hang when run with to many jobs. See https://github.com/easybuilders/easybuild-easyconfigs/pull/9646#issuecomment-574699216

Setting maxparallel does not work, because TF is built as an extension, not as a direct EC, hence using the setting manually.


akesandgren:
Going in, thanks @Flamefire!

PR #1936, opened 2020-01-21
Add EasyBlock for CMake

Flamefire:
Handles all custom parameters that are currently passed to configure so CMake ECs can be simpler

Handles non-loaded modules gracefully instead of passing -DZLIB_LIBRARY_RELEASE=$EBROOTZLIB/lib/libz.so which results in
-DZLIB_LIBRARY_RELEASE=/lib/libz.so if e.g. filter-deps=zlib is used

FollowUp: Use this in all CMake ECs

### DIFF
--- a/easybuild/easyblocks/c/cmake.py
+++ b/easybuild/easyblocks/c/cmake.py
@@ -23,7 +23,7 @@
 # along with EasyBuild.  If not, see <http://www.gnu.org/licenses/>.
 ##
 """
-@author: Alexander Grund
+@author: Alexander Grund (TU Dresden)
 """
 import os
 
@@ -110,9 +110,9 @@ class EB_CMake(ConfigureMake):
         Custom sanity check for CMake.
         """
         paths = {
-            'files': ["bin/%s" % x for x in ['ccmake', 'cmake', 'cpack', 'ctest']],
+            'files': [os.path.join('bin', x) for x in ('ccmake', 'cmake', 'cpack', 'ctest')],
             'dirs': [],
         }
-        commands = [('cmake', '--help'), ('ccmake', '--help')]
+        commands = ['cmake --help', 'ccmake --help']
 
         super(EB_CMake, self).sanity_check_step(custom_paths=paths, custom_commands=commands)
--- a/easybuild/easyblocks/c/cmake.py
+++ b/easybuild/easyblocks/c/cmake.py
@@ -109,9 +109,10 @@ class EB_CMake(ConfigureMake):
         """
         Custom sanity check for CMake.
         """
-        custom_paths = {
+        paths = {
             'files': ["bin/%s" % x for x in ['ccmake', 'cmake', 'cpack', 'ctest']],
             'dirs': [],
         }
+        commands = [('cmake', '--help'), ('ccmake', '--help')]
 
-        super(EB_CMake, self).sanity_check_step(custom_paths=custom_paths)
+        super(EB_CMake, self).sanity_check_step(custom_paths=paths, custom_commands=commands)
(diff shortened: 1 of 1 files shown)


Flamefire:
Ccing @boegel @pescobar, see Slack

easybuild/easyblocks/c/cmake.py
```diff
@@ -0,0 +1,118 @@
...
+# EasyBuild is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with EasyBuild.  If not, see <http://www.gnu.org/licenses/>.
+##
+"""
+@author: Alexander Grund
```
akesandgren: Please add affiliation

easybuild/easyblocks/c/cmake.py
```diff
@@ -0,0 +1,118 @@
...
+
+    def configure_step(self):
+        """
+        Run qmake on the GUI, if necessary
+        """
+        configopts = self.cfg['configopts'].split(' -- ', 1)
+        configure_opts = configopts[0]
+        cmake_opts = configopts[1] if len(configopts) == 2 else ''
+
+        add_configure_opts = {'parallel': '%(parallel)s'}
```
akesandgren: This should probably be something like
add_configure_opts = {}
if self.cfg['parallel']:
  add_configure_opts['parallel'] = self.cfg['parallel']

easybuild/easyblocks/c/cmake.py
```diff
@@ -0,0 +1,118 @@
...
+
+    def sanity_check_step(self):
+        """
+        Custom sanity check for CMake.
+        """
+        paths = {
+            'files': ["bin/%s" % x for x in ['ccmake', 'cmake', 'cpack', 'ctest']],
+            'dirs': [],
+        }
+        commands = [('cmake', '--help'), ('ccmake', '--help')]
```
akesandgren: Simple strings work well here, i.e.
commands = ['cmake --help', 'ccmake --help']

easybuild/easyblocks/c/cmake.py
```diff
@@ -0,0 +1,118 @@
...
+        self.cfg['configopts'] = configure_opts + ' -- ' + cmake_opts
+
+        super(EB_CMake, self).configure_step()
+
+    def sanity_check_step(self):
+        """
+        Custom sanity check for CMake.
+        """
+        paths = {
+            'files': ["bin/%s" % x for x in ['ccmake', 'cmake', 'cpack', 'ctest']],
```
akesandgren: nitpicking, should be os.path.join('bin', x) for x...

boegel:
@Flamefire Please also open a PR for a CMake easyconfig that uses this, so we have something to test with (either update latest CMake easyconfig, or create a new one for latest CMake on top of `GCCcore/9.2.0`)

easybuild/easyblocks/c/cmake.py
Flamefire: Really? In the CMake ECs the unconditional '%(parallel)s' works well and IMO `parallel` is always set by EB

Flamefire:
Done: https://github.com/easybuilders/easybuild-easyconfigs/pull/9871

easybuild/easyblocks/c/cmake.py
akesandgren: The default for parallel in ./easybuild/framework/easyconfig/default.py is None but should probably always be set but...

Flamefire: It will be set as "None" means "autodetect": https://github.com/easybuilders/easybuild-framework/blob/090f8c1d0cf0837eda084abd6f7ba53aaa476d7e/easybuild/framework/easyblock.py#L1618-L1633

akesandgren:
Going in, thanks @Flamefire!

PR #1939, opened 2020-01-27
Some changes made by @mand35

valtandor:
An erstwhile colleague of mine made some trivial changes to the Boost easyblock, including error handling on making a certain directory. Most of his changes were indeed implemented in parallel by other EasyBuild developers and have therefore been dropped from this PR.

### DIFF
--- a/easybuild/easyblocks/b/boost.py
+++ b/easybuild/easyblocks/b/boost.py
@@ -36,6 +36,7 @@ EasyBuild support for Boost, implemented as an easyblock
 @author: Guilherme Peretti-Pezzi (CSCS)
 @author: Joachim Hein (Lund University)
 @author: Michele Dolfi (ETH Zurich)
+@author: Mandes Schoenherr (NIWA)
 """
 from distutils.version import LooseVersion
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/b/boost.py
```diff
@@ -120,8 +121,12 @@ def configure_step(self):
             raise EasyBuildError("When enabling building boost_mpi, also enable the 'usempi' toolchain option.")
 
         # create build directory (Boost doesn't like being built in source dir)
-        self.objdir = os.path.join(self.builddir, 'obj')
-        mkdir(self.objdir)
+        try:
+            self.objdir = os.path.join(self.builddir, 'obj')
+            mkdir(self.objdir)
```
akesandgren: The mkdir from easybuild filetools already handle this.

smoors: @valtandor what @akesandgren is saying is that the error handling you added is already present the `mkdir` function.
in view of that, I think we can close this PR, do you agree?

PR #1941, opened 2020-01-27
Gurobi licence server modifications

orbsmiv:
Modifies block to check for the pre-existence of the `GRB_LICENSE_FILE` env var and uses its value if present, else uses `license_file` value from the easyconfig.

### DIFF
--- a/easybuild/easyblocks/g/gurobi.py
+++ b/easybuild/easyblocks/g/gurobi.py
@@ -27,6 +27,7 @@
 EasyBuild support for installing Gurobi, implemented as an easyblock
 
 @author: Bob Dröge (University of Groningen)
+modified by James Carpenter (University of Birmingham)
 """
 import os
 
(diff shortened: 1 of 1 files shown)


smoors:
@orbsmiv for reproducibility it is better to use the easyconfig parameter `license_file` instead of relying on an enviroment variable.
do you have a good reason why you cannot use `license_file`?

orbsmiv:
@smoors thanks for looking into this. I based the idea off the blocks for Ansys & Matlab where they can use either the `license_file` variable or an environment variable. In our case (and for a few pieces of software) we need to provide different licence files with differing licence server hostnames due to the varying DNS resolution across the various bits of our cluster. Hard-coding a licence file into an application's installation directory therefore doesn't work for us.

smoors:
@orbsmiv ok that makes sense.

but in any case the easyconfig parameter should take precedence, as is the case in Ansys and Matlab.

also, it would be nice if you could distinguish between errors `file not found` and `no license file specified`.

orbsmiv:
@smoors I've made the modifications that you suggested – can you take another look?

easybuild/easyblocks/g/gurobi.py
```diff
@@ -71,7 +83,11 @@ def make_module_extra(self):
         """Custom extra module file entries for Gurobi."""
         txt = super(EB_Gurobi, self).make_module_extra()
         txt += self.module_generator.set_environment('GUROBI_HOME', self.installdir)
-        txt += self.module_generator.set_environment('GRB_LICENSE_FILE', os.path.join(self.installdir, 'gurobi.lic'))
+
+        # Only define GRB_LICENSE_FILE env var if it doesn't already exist
+        if not os.getenv("GRB_LICENSE_FILE"):
```
smoors: @orbsmiv also here, the `license_file` variable must take precendence:
```
if self.cfg['license_file']:
```

easybuild/easyblocks/g/gurobi.py
```diff
@@ -43,13 +44,24 @@ class EB_Gurobi(Tarball):
...
         """Install Gurobi and license file."""
-        # make sure license file is available
-        if self.cfg['license_file'] is None or not os.path.exists(self.cfg['license_file']):
-            raise EasyBuildError("No existing license file specified: %s", self.cfg['license_file'])
+
+        # Check if the client license file is already defined in the appropriate easyconfig variable
+        licfile = self.cfg['license_file']
+        if licfile is None:
+            # Fallback to standard Gurobi license file environment variable
+            licfile = os.getenv('GRB_LICENSE_FILE')
```
boegel: Have you considered using the `find_flexlm_license` function provided by `filetools`? See https://github.com/easybuilders/easybuild-framework/blob/develop/easybuild/tools/filetools.py#L1670

See the COMSOL & PGI easyblocks, as well as the `IntelBase` generic easyblock used for the Intel tools.

easybuild/easyblocks/g/gurobi.py
```diff
@@ -71,7 +78,11 @@ def make_module_extra(self):
         """Custom extra module file entries for Gurobi."""
         txt = super(EB_Gurobi, self).make_module_extra()
         txt += self.module_generator.set_environment('GUROBI_HOME', self.installdir)
-        txt += self.module_generator.set_environment('GRB_LICENSE_FILE', os.path.join(self.installdir, 'gurobi.lic'))
+
+        # Only define GRB_LICENSE_FILE env var if it doesn't already exist
+        if not os.getenv("GRB_LICENSE_FILE"):
+            txt += self.module_generator.set_environment('GRB_LICENSE_FILE',
+                                                         os.path.join(self.installdir, 'gurobi.lic'))
```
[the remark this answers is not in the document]
akesandgren: (This was a fairly old comment that I had forgotten to press submit on)

boegel:
This PR has gone stale, so I'm closing it.

If the proposed changes are still relevant, please open a new PR on top of the latest `gurobi.py` from `develop` branch

PR #1942, opened 2020-01-28
Add XALT EasyBlock

samcmill:
Specialized variant of the ConfigureMake EasyBlock for building and installing [XALT](https://github.com/xalt/xalt).

cc @rtmclay

### DIFF
--- a/easybuild/easyblocks/x/xalt.py
+++ b/easybuild/easyblocks/x/xalt.py
@@ -142,7 +142,8 @@ class EB_XALT(ConfigureMake):
         # directory in the Singularity container and preload the XALT library
         # https://xalt.readthedocs.io/en/latest/050_install_and_test.html#xalt-modulefile
         txt += self.module_generator.prepend_paths('SINGULARITY_BINDPATH', '')
-        txt += self.module_generator.prepend_paths('SINGULARITYENV_LD_PRELOAD', 'lib64/libxalt_init.%s' % get_shared_lib_ext())
+        txt += self.module_generator.prepend_paths('SINGULARITYENV_LD_PRELOAD',
+                                                   'lib64/libxalt_init.%s' % get_shared_lib_ext())
 
         return txt
 
--- a/easybuild/easyblocks/x/xalt.py
+++ b/easybuild/easyblocks/x/xalt.py
@@ -50,6 +50,7 @@ class EB_XALT(ConfigureMake):
             'logging_url': [None, "Logging URL for transmission", CUSTOM],
             'mysql': [False, "Build with MySQL support", CUSTOM],
             'scalar_sampling': [True, "Enable scalar sampling", CUSTOM],
+            'static_cxx': [False, "Statically link libstdc++ and libgcc_s", CUSTOM],
             'syshost': [None, "System name", MANDATORY],
             'transmission': [None, "Data tranmission method", MANDATORY],
         }
@@ -120,6 +121,12 @@ class EB_XALT(ConfigureMake):
         else:
             self.cfg.update('configopts', '--with-MySQL=no')
 
+        # If XALT is built with a more recent compiler than the system
+        # compiler, then XALT likely will depend on symbol versions not
+        # available in the system libraries. Link statically as a workaround.
+        if self.cfg['static_cxx'] is True:
+            self.cfg.update('configopts', 'LDFLAGS="${LDFLAGS} -static-libstdc++ -static-libgcc"')
+
         # Configure
         super(EB_XALT, self).configure_step()
 
@@ -159,5 +166,7 @@ class EB_XALT(ConfigureMake):
                       'lib64/libxalt_init.%s' % get_shared_lib_ext()],
             'dirs': ['bin', 'libexec', 'sbin'],
         }
+        custom_commands = ['xalt_configuration_report']
 
-        super(EB_XALT, self).sanity_check_step(custom_paths=custom_paths)
+        super(EB_XALT, self).sanity_check_step(custom_commands=custom_commands,
+                                               custom_paths=custom_paths)
--- a/easybuild/easyblocks/x/xalt.py
+++ b/easybuild/easyblocks/x/xalt.py
@@ -105,13 +105,11 @@ class EB_XALT(ConfigureMake):
             # Default value, enable GPU tracking if nvml.h is present
             # and the CUDA module is loaded
             cuda_root = get_software_root('CUDA')
-            if not cuda_root:
-                # Try using the system CUDA
-                cuda_root = '/usr/local/cuda'
-            nvml_h = os.path.join(cuda_root, "include", "nvml.h")
-            if os.path.isfile(nvml_h):
-                self.cfg.update('configopts', '--with-trackGPU=yes')
-                self.cfg['gpu_tracking'] = True
+            if cuda_root:
+                nvml_h = os.path.join(cuda_root, "include", "nvml.h")
+                if os.path.isfile(nvml_h):
+                    self.cfg.update('configopts', '--with-trackGPU=yes')
+                    self.cfg['gpu_tracking'] = True
         else:
             # User disabled
             self.cfg.update('configopts', '--with-trackGPU=no')
@@ -128,9 +126,7 @@ class EB_XALT(ConfigureMake):
     def make_module_extra(self, *args, **kwargs):
         txt = super(EB_XALT, self).make_module_extra(*args, **kwargs)
 
-        txt += self.module_generator.prepend_paths('LD_PRELOAD', 'lib64/libxalt_init.so')
-        txt += self.module_generator.prepend_paths('SINGULARITY_BINDPATH', '')
-        txt += self.module_generator.prepend_paths('SINGULARITYENV_LD_PRELOAD', 'lib64/libxalt_init.so')
+        txt += self.module_generator.prepend_paths('LD_PRELOAD', 'lib64/libxalt_init.%s' % get_shared_lib_ext())
         txt += self.module_generator.set_environment('XALT_DIR', self.installdir)
         txt += self.module_generator.set_environment('XALT_ETC_DIR', '%s' % os.path.join(self.installdir, 'etc'))
         txt += self.module_generator.set_environment('XALT_EXECUTABLE_TRACKING',
@@ -142,19 +138,25 @@ class EB_XALT(ConfigureMake):
         txt += self.module_generator.set_environment('XALT_SCALAR_SAMPLING',
                                                      ('no', 'yes')[bool(self.cfg['scalar_sampling'])])
 
+        # In order to track containerized executables, bind mount the XALT
+        # directory in the Singularity container and preload the XALT library
+        # https://xalt.readthedocs.io/en/latest/050_install_and_test.html#xalt-modulefile
+        txt += self.module_generator.prepend_paths('SINGULARITY_BINDPATH', '')
+        txt += self.module_generator.prepend_paths('SINGULARITYENV_LD_PRELOAD', 'lib64/libxalt_init.%s' % get_shared_lib_ext())
+
         return txt
 
     def make_module_req_guess(self):
-        """ Limit default guess paths
-        Do not set CPATH, LD_LIBRARY_PATH, LIBRARY_PATH, include sbin in PATH """
+        """Custom guesses for environment variables"""
         return {'COMPILER_PATH': 'bin',
                 'PATH': 'bin'}
 
     def sanity_check_step(self):
         """Custom sanity check"""
         custom_paths = {
-            'files': ['lib64/libxalt_init.%s' % get_shared_lib_ext()],
-            'dirs': ['lib64'],
+            'files': ['bin/ld', 'bin/ld.gold', 'bin/xalt_extract_record',
+                      'lib64/libxalt_init.%s' % get_shared_lib_ext()],
+            'dirs': ['bin', 'libexec', 'sbin'],
         }
 
         super(EB_XALT, self).sanity_check_step(custom_paths=custom_paths)
--- a/easybuild/easyblocks/x/xalt.py
+++ b/easybuild/easyblocks/x/xalt.py
@@ -1,0 +1,94 @@
+##
+# Copyright 2020 NVIDIA
+#
+# This file is triple-licensed under GPLv2 (see below), MIT, and
+# BSD three-clause licenses.
+#
+# This file is part of EasyBuild,
+# originally created by the HPC team of Ghent University (http://ugent.be/hpc/en),
+# with support of Ghent University (http://ugent.be/hpc),
+# the Flemish Supercomputer Centre (VSC) (https://www.vscentrum.be),
+# Flemish Research Foundation (FWO) (http://www.fwo.be/en)
+# and the Department of Economy, Science and Innovation (EWI) (http://www.ewi-vlaanderen.be/en).
+#
+# https://github.com/easybuilders/easybuild
+#
+# EasyBuild is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation v2.
+#
+# EasyBuild is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with EasyBuild.  If not, see <http://www.gnu.org/licenses/>.
+##
+"""
+EasyBuild support for XALT, implemented as an easyblock
+@author: Scott McMillan (NVIDIA)
+"""
+import os
+import re
+
+from easybuild.easyblocks.generic.configuremake import ConfigureMake
+from easybuild.framework.easyconfig import CUSTOM, MANDATORY
+from easybuild.tools.build_log import EasyBuildError
+from easybuild.tools.modules import get_software_root
+from easybuild.tools.systemtools import check_os_dependency, get_shared_lib_ext
+
+
+class EB_XALT(ConfigureMake):
+    """Support for building and installing XALT."""
+
+    @staticmethod
+    def extra_options():
+        extra_vars = {
+            'config_py': [None, "XALT site filter file", MANDATORY],
+            'syshost': ['hardcode:cluster', "System name", CUSTOM],
+            'transmission': ['syslog', "", CUSTOM],
+        }
+        return ConfigureMake.extra_options(extra_vars)
+
+    def configure_step(self):
+        """Custom configuration step for XALT."""
+
+        # Remove the subdir from the install prefix to resolve the
+        # following configure error.  XALT automatically creates a
+        # versioned directory hierarchy.
+        #
+        #    Do not install XALT with 2.7.27 as part of the
+        #    prefix.  It will lead to many many problems with future
+        #    installs of XALT
+        #
+        #    Executables built with the current version of XALT
+        #    will not work with future installs of XALT!!!
+        #
+        #    If you feel that you know better than the developer
+        #    of XALT then you can configure XALT with the configure
+        #    --with-IrefuseToInstallXALTCorrectly=yes and set the prefix
+        #    to include the version
+        self.installdir = os.path.normpath(re.sub(self.install_subdir, '', self.installdir))
+
+        # XALT site filter config file is mandatory
+        if not self.cfg['config_py']:
+            raise EasyBuildError('XALT site filter config must be specified. '
+                                 'Use "--try-amend=config_py=<file>"')
+        self.cfg.update('configopts', '--with-config=%s' % self.cfg['config_py'])
+
+        if self.cfg['syshost']:
+            self.cfg.update('configopts', '--with-syshostConfig=%s' % self.cfg['syshost'])
+
+        if self.cfg['transmission']:
+            self.cfg.update('configopts', '--with-transmission=%s' % self.cfg['transmission'])
+
+        super(EB_XALT, self).configure_step()
+
+        # Reassemble the default install path
+        self.installdir = os.path.join(self.installdir, self.install_subdir)
+
+    def make_module_req_guess(self):
+        """ Limit default guess paths """
+        return {'COMPILER_PATH': 'bin',
+                'PATH': 'bin'}
--- a/easybuild/easyblocks/x/xalt.py
+++ b/easybuild/easyblocks/x/xalt.py
@@ -59,23 +59,13 @@ class EB_XALT(ConfigureMake):
     def configure_step(self):
         """Custom configuration step for XALT."""
 
-        # Remove the subdir from the install prefix to resolve the
-        # following configure error.  XALT automatically creates a
-        # versioned directory hierarchy.
-        #
-        #    Do not install XALT with 2.7.28 as part of the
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/x/xalt.py
```diff
@@ -0,0 +1,92 @@
...
+            self.cfg.update('configopts', '--with-syshostConfig=%s' % self.cfg['syshost'])
+
+        if self.cfg['transmission']:
+            self.cfg.update('configopts', '--with-transmission=%s' % self.cfg['transmission'])
+
+        super(EB_XALT, self).configure_step()
+
+        # Reassemble the default install path
+        self.installdir = os.path.join(self.installdir, self.install_subdir)
+
```
victorusu: I think it needs a function to defined environment variables
```
def make_module_extra(self, *args, **kwargs):
    txt = super(EB_XALT, self).make_module_extra(*args, **kwargs)
    txt += self.module_generator.prepend_paths('LD_PRELOAD', 'lib64/libxalt_init.so')
    txt += self.module_generator.prepend_paths('ALT_LINKER', 'bin/%s' % self.cfg['linker'])
    txt += self.module_generator.prepend_paths('COMPILER_PATH', 'bin')
'ALT_LINKER': '%(installdir)s/bin/ld'
    txt += self.module_generator.set_environment('XALT_EXECUTABLE_TRACKING',  '%s' % self.cfg['executable_tracking'])
    txt += self.module_generator.set_environment('XALT_SCALAR_SAMPLING',  '%s' % self.cfg['scalar_sampling'])
    txt += self.module_generator.set_environment('XALT_DIR', '%s' % self.installdir)
    txt += self.module_generator.set_environment('XALT_ETC_DIR', '%s/etc' % self.installdir)
    if get_software_root('cuda'):
        txt += self.module_generator.set_environment('XALT_GPU_TRACKING', 'yes')
    else:
        txt += self.module_generator.set_environment('XALT_GPU_TRACKING', 'no')
    if self.cfg['transmission'] == "curl":
        txt += self.module_generator.set_environment('XALT_LOGGING_URL', '%s' % self.cfg['transmission_url'])
    return txt
```

easybuild/easyblocks/x/xalt.py
```diff
@@ -0,0 +1,92 @@
...
+class EB_XALT(ConfigureMake):
+    """Support for building and installing XALT."""
+
+    @staticmethod
+    def extra_options():
+        extra_vars = {
+            'config_py': [None, "XALT site filter file", MANDATORY],
+            'syshost': ['hardcode:cluster', "System name", CUSTOM],
+            'transmission': ['syslog', "", CUSTOM],
+        }
```
victorusu: I think it should have these additional extra options
```
'executable_tracking': ['yes', '', CUSTOM],
'scalar_sampling': ['yes', '', CUSTOM],
'linker': ['ld', '', CUSTOM],
'logging_url': [None, 'logging url when using curl transmission method', CUSTOM],
```

easybuild/easyblocks/x/xalt.py
```diff
@@ -0,0 +1,92 @@
...
+            raise EasyBuildError('XALT site filter config must be specified. '
+                                 'Use "--try-amend=config_py=<file>"')
+        self.cfg.update('configopts', '--with-config=%s' % self.cfg['config_py'])
+
+        if self.cfg['syshost']:
+            self.cfg.update('configopts', '--with-syshostConfig=%s' % self.cfg['syshost'])
+
+        if self.cfg['transmission']:
+            self.cfg.update('configopts', '--with-transmission=%s' % self.cfg['transmission'])
+
```
victorusu: I think it should include GPU tracking support, with something like
```
if get_software_root('cuda'):
        txt += self.module_generator.set_environment('XALT_GPU_TRACKING', 'yes')
    else:
        txt += self.module_generator.set_environment('XALT_GPU_TRACKING', 'no')
```

victorusu: I think it should add a test for the link type
```
allowed_linkers =['ld', 'ld.gold']
if not self.cfg['linker'] in allowed_linkers:
    raise EasyBuildError("Allowed linker values are %s" % allowed_linkers)
```

victorusu: I think it should add a test for the config type
```
yes_or_no =['YES', 'NO']
if not self.cfg['executable_tracking'].upper() in yes_or_no:
    raise EasyBuildError("Allowed executable_tracking values are yes or no")

if not self.cfg['scalar_sampling'].upper() in yes_or_no:
    raise EasyBuildError("Allowed scalar_sampling values are yes or no")
```

easybuild/easyblocks/x/xalt.py
```diff
@@ -0,0 +1,92 @@
...
+        # XALT site filter config file is mandatory
+        if not self.cfg['config_py']:
+            raise EasyBuildError('XALT site filter config must be specified. '
+                                 'Use "--try-amend=config_py=<file>"')
+        self.cfg.update('configopts', '--with-config=%s' % self.cfg['config_py'])
+
+        if self.cfg['syshost']:
+            self.cfg.update('configopts', '--with-syshostConfig=%s' % self.cfg['syshost'])
+
+        if self.cfg['transmission']:
```
victorusu: I think it should add a check for "curl" transmission style
```
if self.cfg['transmission'] == "curl":
    if not self.cfg['logging_url'] in allowed_linkers:
        raise EasyBuildError("You need to setup a logging url if you want to use the curl transmission style")
```

easybuild/easyblocks/x/xalt.py
```diff
@@ -0,0 +1,92 @@
...
+            self.cfg.update('configopts', '--with-transmission=%s' % self.cfg['transmission'])
+
+        super(EB_XALT, self).configure_step()
+
+        # Reassemble the default install path
+        self.installdir = os.path.join(self.installdir, self.install_subdir)
+
+    def make_module_req_guess(self):
```
victorusu: If you use the suggested `def make_module_extra`, you do not need this function

samcmill:
Thanks for the feedback @victorusu.  

The approach I took was to keep the EasyBlock as lean as possible and keep the specifics in the [XALT EasyConfig](https://github.com/easybuilders/easybuild-easyconfigs/pull/9792).  Additional XALT EasyConfigs could be created for other permutations, such as the curl transmission method.  (If not for the XALT prefix behavior, I likely would have used the standard ConfigureMake EasyBlock.)

Alternatively, the EasyBlock could be a much more customized version of the ConfigureMake EasyBlock that hard codes more XALT options, as you are suggesting.  

From what I can tell, the first approach seems to be more closely aligned with the EasyBuild "spirit".

@boegel, do you have a preference?

boegel:
I'll try to take a look at this soon, perhaps also polling @rtmclay (the XALT developer) for feedback.

Thanks a lot for the contribution @samcmill, hopefully we can get this merged soon in one form or another...

samcmill:
@boegel would still appreciate your feedback on this.  Thanks!

boegel:
@samcmill Thanks for the bump!
I'll reach out to @rtmclay to get some feedback from him on this...

easybuild/easyblocks/x/xalt.py
boegel: I agree that having easyconfig parameters for `executable_tracking` and `scalar_sampling` makes sense, and having them enabled by default makes sense too. That way the corresponding environment variables and/or configuration options (probably only the latter?) can be put in place by the easyblock...

Same goes for GPU tracking I guess, since having a `gpu_tracking` option with default `None` could maybe even allow EasyBuild to figure out whether or not GPU tracking *can* be enabled (based on whether the required dependencies are available, like `nvidia.h` and CUDA).

easybuild/easyblocks/x/xalt.py
```diff
@@ -0,0 +1,92 @@
...
+        #    installs of XALT
+        #
+        #    Executables built with the current version of XALT
+        #    will not work with future installs of XALT!!!
+        #
+        #    If you feel that you know better than the developer
+        #    of XALT then you can configure XALT with the configure
+        #    --with-IrefuseToInstallXALTCorrectly=yes and set the prefix
+        #    to include the version
+        self.installdir = os.path.normpath(re.sub(self.install_subdir, '', self.installdir))
```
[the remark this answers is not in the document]
boegel: After discussing this with @rtmclay, I think it makes sense to wait until there's an XALT release that has support for picking up on `$XALT_DIR` (see @samcmill's PR to add that at https://github.com/xalt/xalt/pull/26).

If XALT picks up on `$XALT_DIR`, none of this trickery with the installation prefix is needed anymore, and we can have several XALT installations side by side. As long as `$XALT_DIR` points to an existing XALT installation when static binaries that have been hijacked by XALT are used, XALT tracking will still work.

I suggested to @rtmclay to let XALT log a message to syslog when it can't find its tool anymore (which is what `$XALT_DIR` would be for, and which is why XALT is currently so pedantic about where it's being installed), which is a decent way of signalling that XALT tracking got broken because XALT's tool were (re)moved.

easybuild/easyblocks/x/xalt.py
```diff
@@ -0,0 +1,92 @@
...
+
+        # XALT site filter config file is mandatory
+        if not self.cfg['config_py']:
+            raise EasyBuildError('XALT site filter config must be specified. '
+                                 'Use "--try-amend=config_py=<file>"')
+        self.cfg.update('configopts', '--with-config=%s' % self.cfg['config_py'])
```
boegel: Throwing an error here makes sense, since there's no way we can come up with a generic configuration file.

I would change a couple of things here though:

* use positive logic
* check whether specified config.py actually exists
* improve the error message
   ```python
   config_py = self.cfg['config_py']
   if config_py:
        if os.path.exists(config_py):
            self.cfg.update('configopts', '--with-config=%s' % config_py)
        else:
            raise EasyBuildError("Specified XALT configuration file %s does not exist!", config_py)
    else:
        error_msg = "Location of XALT configuration file must be specified via 'config_py' easyconfig parameter. "
        error_msg += "You can edit the easyconfig file, or use 'eb --try-amend=config_py=<path>'. "
        error_msg += "See https://xalt.readthedocs.io/en/latest/030_site_filtering.html for more information."
        raise EasyBuildError(error_msg)

easybuild/easyblocks/x/xalt.py
```diff
@@ -0,0 +1,92 @@
...
+
+
+class EB_XALT(ConfigureMake):
+    """Support for building and installing XALT."""
+
+    @staticmethod
+    def extra_options():
+        extra_vars = {
+            'config_py': [None, "XALT site filter file", MANDATORY],
+            'syshost': ['hardcode:cluster', "System name", CUSTOM],
```
boegel: We should leave `syshost` unspecified, and also require that a value is specified (similar to `config_py`), since there's no way we can guess a good default value (and hardcoding `cluster` is a bad idea imho).

The error message produced should point to https://xalt.readthedocs.io/en/latest/020_site_configuration.html ?

easybuild/easyblocks/x/xalt.py
```diff
@@ -0,0 +1,92 @@
...
+            'transmission': ['syslog', "", CUSTOM],
```
boegel: Likewise here, I'm not sure we can pick a decent default for this, we should enforce that this is specified at installation time (and point to https://xalt.readthedocs.io/en/latest/020_site_configuration.html#xalt-data-transmission for more information).

easybuild/easyblocks/x/xalt.py
boegel: Yeah, I think that makes sense...

Where's the documentation on the `curl` transmission mode though, and is the `in allowed_linkers` part correct @victorusu?

easybuild/easyblocks/x/xalt.py
boegel: That would require that `XALT` gets installed with `CUDA` listed as a dependency, which I don't think is wise...

@rtmclay told me what XALT actually needs is `nvidia.h` to be available the system, so if we can figure out a way to check for that, we can auto-enable GPU tracking based on that (unless `gpu_tracking = False` is used to override the auto-detection).

easybuild/easyblocks/x/xalt.py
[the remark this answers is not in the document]
samcmill: Makes sense to me.

When XALT no longer requires special installation prefix handling, then I think a dedicated XALT EasyBlock will not be necessary?

easybuild/easyblocks/x/xalt.py
boegel: Where would these linkers be used exactly, can you clarify @victorusu?

boegel: `executable_tracking` is a boolean setting, so should only accept `True` and `False` as values (not strings); same goes for `scalar_sampling`

easybuild/easyblocks/x/xalt.py
boegel: Agreed, we can handle setting of environment variables in the easyblock, and keep the easyconfigs clean...

easybuild/easyblocks/x/xalt.py
boegel: +1

boegel:
@samcmill I'm happy to clarify anything if needed, either via this PR or more directly ([EasyBuild Slack channel](https://easybuild.readthedocs.io/en/latest/#getting-help), or a short conf call), please don't hesitate to reach out!

easybuild/easyblocks/x/xalt.py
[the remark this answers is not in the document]
boegel: It would still be useful I think, w.r.t. the required values for `config_py` & co...

easybuild/easyblocks/x/xalt.py
samcmill: `make_module_req_guess` is necessary to prevent setting unneeded (and potentially harmful in this case) environment variables like `LD_LIBRARY_PATH`.

easybuild/easyblocks/x/xalt.py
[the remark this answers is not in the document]
samcmill: Addressed the original comment, so making this as resolved.

easybuild/easyblocks/x/xalt.py
samcmill: Now setting the environment variables in the EasyBlock.  Please open a new comment if you have feedback on the latest commit.

easybuild/easyblocks/x/xalt.py
samcmill: The latest commit tries to find CUDA using `get_software_root('CUDA')` and falls back to `/usr/local/cuda` is the former is not defined.  It's not perfect (the CUDA module must be loaded prior to running `eb`).  Please let me know if there are better approaches.

samcmill:
@boegel @victorusu The latest commit should address most of your feedback.  Thanks!

The step monkeying with the install path is still there.  I'll revisit that once https://github.com/xalt/xalt/pull/26 is merged.

I'm unsure about the alternative linker stuff, so no changes for that yet.  @victorusu - can you please clarify?

samcmill:
The latest changes incorporate the `--with-siteControlledPrefix` option added to XALT 2.7.29 by @rtmclay, so the easyblock code that monkeys with the prefix has been removed.

This is ready for another review.

samcmill:
@boegel @victorusu can you please review this again?

samcmill:
@boegel @victorusu can you please review this again?

boegel:
@samcmill I'm sorry for not getting back to this yet, it's on my TODO list for EasyBuild v4.2.1 (see https://github.com/orgs/easybuilders/projects/5)

easybuild/easyblocks/x/xalt.py
```diff
@@ -0,0 +1,160 @@
...
+        if self.cfg['gpu_tracking'] is True:
+            # User enabled
+            self.cfg.update('configopts', '--with-trackGPU=yes')
+        elif self.cfg['gpu_tracking'] is None:
+            # Default value, enable GPU tracking if nvml.h is present
+            # and the CUDA module is loaded
+            cuda_root = get_software_root('CUDA')
+            if not cuda_root:
+                # Try using the system CUDA
+                cuda_root = '/usr/local/cuda'
```
boegel, line 110: I would not use this hardcoded fallback path, EasyBuild should not make decision based on what is installed in the OS.

Generally, CUDA is supplied via EasyBuild, and if it's not, people can still opt-in to building XALT with GPU tracking support by setting `gpu_tracking` to `True`.

easybuild/easyblocks/x/xalt.py
```diff
@@ -0,0 +1,160 @@
...
+        else:
+            self.cfg.update('configopts', '--with-MySQL=no')
+
+        # Configure
+        super(EB_XALT, self).configure_step()
+
+    def make_module_extra(self, *args, **kwargs):
+        txt = super(EB_XALT, self).make_module_extra(*args, **kwargs)
+
+        txt += self.module_generator.prepend_paths('LD_PRELOAD', 'lib64/libxalt_init.so')
```
boegel, line 131: Please also use `get_shared_lib_ext()` here rather than hardcoding `.so`.

easybuild/easyblocks/x/xalt.py
```diff
@@ -0,0 +1,160 @@
...
+        txt += self.module_generator.prepend_paths('SINGULARITY_BINDPATH', '')
+        txt += self.module_generator.prepend_paths('SINGULARITYENV_LD_PRELOAD', 'lib64/libxalt_init.so')
```
boegel, line 133: Do these really belong in the `XALT` easyblock?

At the very least they should be done in a separate code block below, with an accompanying comment to clarify why it makes sense to let the `XALT` module file define update these `$SINGULARITY*` environment variables.

easybuild/easyblocks/x/xalt.py
```diff
@@ -0,0 +1,160 @@
...
+        txt += self.module_generator.set_environment('XALT_SCALAR_SAMPLING',
+                                                     ('no', 'yes')[bool(self.cfg['scalar_sampling'])])
+
+        return txt
+
+    def make_module_req_guess(self):
+        """ Limit default guess paths
+        Do not set CPATH, LD_LIBRARY_PATH, LIBRARY_PATH, include sbin in PATH """
+        return {'COMPILER_PATH': 'bin',
+                'PATH': 'bin'}
```
boegel, line 151: docstring doesn't match what is actually being done (and it could be formatted differently):

```python
return {
    'COMPILER_PATH': 'bin',
    'PATH': 'bin',
}
```

easybuild/easyblocks/x/xalt.py
```diff
@@ -0,0 +1,160 @@
...
+
+    def sanity_check_step(self):
+        """Custom sanity check"""
+        custom_paths = {
+            'files': ['lib64/libxalt_init.%s' % get_shared_lib_ext()],
+            'dirs': ['lib64'],
```
boegel, line 157: Checking for a non-empty `lib64` directory isn't very useful if we're also checking for a specific file in this directory.

There's a couple of other important files we *should* be checking for here though: `bin/ld`, `bin/ld.gold`, and also the `libexec` and `sbin` directories.

easybuild/easyblocks/x/xalt.py
boegel: @victorusu ping?

boegel: Checking whether CUDA is loaded via `get_software_root('CUDA')` is fine, but I don't think the fallback to `/usr/local/cuda` makes sense...

Users can easily opt-in to GPU tracking if they really want it when not providing CUDA via EasyBuild?

victorusu: [We do install XALT](https://github.com/eth-cscs/production/blob/da8e1f42dcfe66511778750101d8770b08513962/easybuild/easyconfigs/x/xalt/xalt-2.8.eb#L73) with CUDA as a build dependency and it works. CUDA does not need to be present at run time.

victorusu: XALT2 has [wrapper scripts](https://github.com/xalt/xalt/blob/c21e24b02a0d1e8d698b271775bec30806ab8257/sh_src/) around `ld` and `ld.gold`. It uses them to [do some magic](https://github.com/xalt/xalt/blob/c21e24b02a0d1e8d698b271775bec30806ab8257/sh_src/ld.in#L268).
Currently, the linkers are the same `ld.gold` points to `ld` (see [here](https://github.com/xalt/xalt/blob/master/sh_src/ld.gold.in)). But the time that I suggested this change, I was not aware that they were actually the same thing.
Even though in the future, `ld` and `ld.gold` may be different things, I do not think that you need to do this change now.

easybuild/easyblocks/x/xalt.py
victorusu: @boegel, one can use the `get_shared_lib_ext()`, but XALT only works on Linux...

easybuild/easyblocks/x/xalt.py
victorusu: The suggestion to add these variables comes from the [XALT documentation](https://xalt.readthedocs.io/en/latest/050_install_and_test.html#xalt-modulefile)

easybuild/easyblocks/x/xalt.py
```diff
@@ -0,0 +1,160 @@
...
+    def configure_step(self):
+        """Custom configuration step for XALT."""
+
+        # By default, XALT automatically appends 'xalt/<version>' to the
+        # prefix, i.e., --prefix=/opt will actually install in
+        # /opt/xalt/<version>.  To precisely control the install prefix and
+        # not append anything to the prefix, use the configure option
+        # '--with-siteControlledPrefix=yes'.
+        # See https://xalt.readthedocs.io/en/latest/050_install_and_test.html
+        self.cfg.update('configopts', '--with-siteControlledPrefix=yes')
```
victorusu, line 67: You can avoid the use of this configuration option, you do something similar to what [we do](https://github.com/eth-cscs/production/blob/da8e1f42dcfe66511778750101d8770b08513962/easybuild/easyblocks/xalt.py#L46)[

easybuild/easyblocks/x/xalt.py
boegel: Sure, it's more of a code style thing, we try not to hardcode `.so` anywhere. Perhaps XALT can be fixed on macOS in the future, who knows...

easybuild/easyblocks/x/xalt.py
victorusu: @boegel, there is no documentation...

easybuild/easyblocks/x/xalt.py
```diff
@@ -0,0 +1,163 @@
...
+
+    def sanity_check_step(self):
+        """Custom sanity check"""
+        custom_paths = {
+            'files': ['bin/ld', 'bin/ld.gold', 'bin/xalt_extract_record',
+                      'lib64/libxalt_init.%s' % get_shared_lib_ext()],
+            'dirs': ['bin', 'libexec', 'sbin'],
+        }
+
+        super(EB_XALT, self).sanity_check_step(custom_paths=custom_paths)
```
boegel, line 163: We should also run `xalt_configuration_report` as a sanity check command, see also https://github.com/easybuilders/easybuild-easyconfigs/pull/9792#issuecomment-631248656

boegel:
Tested with https://github.com/easybuilders/easybuild-easyconfigs/pull/9792, good to go, thanks a lot for your efforts on this @samcmill!

PR #1943, opened 2020-02-03
get_tbb_gccprefix: support GCCcore and gcc4.8 (if that dir exists)

bartoldeman:
This fixes the issue that icc/iccifort/recent tbb installations may no longer set LD_LIBRARY_PATH and LIBRARY_PATH, since the library files may now be in a `gcc4.8` subdirectory.

This fix is as minimal as possible, though I doubt that gcc<4.4 is relevant any more.


akesandgren:

PR #1944, opened 2020-02-04
[TensorFlow] Disable ROCM and Android workspace explicitely

Flamefire:
Minor change which keeps existing behavior but leads to cleaner logs and more explicit configuration with recent TF.


akesandgren:
Going in, thanks @Flamefire!

PR #1946, opened 2020-02-06
A path to making OpenBLAS optarch aware.

terjekv:
- A first and very raw implementation, mapping optarch's march
    directly to targets known to OpenBLAS.
  - A better mapping is required. Skylake-avx512 vs SKYLAKEX as
    an obvious example.

### DIFF
--- a/easybuild/easyblocks/o/openblas.py
+++ b/easybuild/easyblocks/o/openblas.py
@@ -3,20 +3,42 @@ EasyBuild support for building and installing OpenBLAS, implemented as an easybl
 
 @author: Andrew Edmondson (University of Birmingham)
 @author: Alex Domingo (Vrije Universiteit Brussel)
+@author: Terje Kvernes (University of Oslo)
 """
 import os
+import re
 from distutils.version import LooseVersion
+from easybuild.base import fancylogger
 from easybuild.easyblocks.generic.configuremake import ConfigureMake
+from easybuild.framework.easyconfig import CUSTOM
+from easybuild.tools.build_log import EasyBuildError, print_warning
+from easybuild.tools.config import ERROR, build_option
+from easybuild.tools.filetools import read_file
 from easybuild.tools.systemtools import POWER, get_cpu_architecture, get_shared_lib_ext
-from easybuild.tools.build_log import print_warning
-from easybuild.tools.config import ERROR
 from easybuild.tools.run import run_cmd, check_log_for_errors
+
+_log = fancylogger.getLogger('systemtools', fname=False)
+
+try:
+    from archspec import cpu as archspec_cpu
+    HAVE_ARCHSPEC = True
+except ImportError as err:
+    _log.debug("Failed to import 'archspec' Python module: %s", err)
+    HAVE_ARCHSPEC = False
 
 TARGET = 'TARGET'
 
 
 class EB_OpenBLAS(ConfigureMake):
     """Support for building/installing OpenBLAS."""
+
+    @staticmethod
+    def extra_options():
+        """Custom easyconfig parameters for OpenBLAS"""
+        extra_vars = {
+            'targetfile': ['TargetList.txt', "File containing OpenBLAS target list", CUSTOM],
+        }
+        return ConfigureMake.extra_options(extra_vars)
 
     def configure_step(self):
         """ set up some options - but no configure command to run"""
@@ -29,6 +51,15 @@ class EB_OpenBLAS(ConfigureMake):
             'USE_THREAD': '1',
         }
 
+        # Handle the possibility of setting the target architecture as dynamic,
+        # where OpenBLAS will optimize the kernel at runtime.
+        self._dynamic_target = False
+
+        compiler_optarch = self._optarch_for_compiler(build_option('optarch'))
+
+        # Retain the (m)arch part of the optarch settings across the entire object.
+        self._optarch_architecture = compiler_optarch
+
         if '%s=' % TARGET in self.cfg['buildopts']:
             # Add any TARGET in buildopts to default_opts, so it is passed to testopts and installopts
             for buildopt in self.cfg['buildopts'].split():
@@ -39,6 +70,38 @@ class EB_OpenBLAS(ConfigureMake):
             # There doesn't seem to be a POWER9 option yet, but POWER8 should work.
             print_warning("OpenBLAS 0.3.5 and lower have known issues on POWER systems")
             default_opts[TARGET] = 'POWER8'
+        elif compiler_optarch:
+            compiler_family = self.toolchain.comp_family()
+            self.log.info("EasyBuild full optarch requested for %s: %s" % (compiler_family, compiler_optarch))
+            optarch_as_target = self._parse_optarch(compiler_optarch)
+            mapped_target = None
+
+            if optarch_as_target:
+                # Note that _parse_optarch returns lowercased results, so GENERIC has become 'generic'.
+                if optarch_as_target == 'generic':
+                    self._set_dynamic_architecture(default_opts)
+                    mapped_target = 'generic'
+                else:
+                    self.log.info("EasyBuild march: %s" % optarch_as_target)
+                    openblas_targets = self._get_openblas_targets(self.cfg['targetfile'])
+                    mapped_target = self._get_mapped_target(optarch_as_target, openblas_targets)
+            else:
+                self.log.info("Optarch specified for %s, but no march detected", compiler_family)
+
+            if mapped_target is None:
+                print_warning("optarch for %s given as '%s'\n"
+                              "EasyBuild was unable to map this to an equivalent OpenBLAS target!\n"
+                              "OpenBLAS will be built to optimize its kernel at runtime!\n"
+                              % (self.toolchain.comp_family(), compiler_optarch))
+                self.log.warning("Unable to map %s to an OpenBLAS target, falling back to runtime optimization."
+                                 % compiler_optarch)
+                self._set_dynamic_architecture(default_opts)
+            elif mapped_target == 'generic':
+                self.log.info("Optarch 'GENERIC' requested, will enable runtime optimization.")
+            else:
+                mapped_target = mapped_target.upper()
+                self.log.info("Optarch mapped between EasyBuild and OpenBLAS to: " + mapped_target)
+                default_opts['TARGET'] = mapped_target
 
         for key in sorted(default_opts.keys()):
             for opts_key in ['buildopts', 'testopts', 'installopts']:
@@ -57,6 +120,17 @@ class EB_OpenBLAS(ConfigureMake):
                 build_parts += ['re_lapack']
         build_parts += ['shared']
 
+        # If we're doing either a dynamic build or utilizing optarch,
+        # strip march from all environment variables except the EBVAR-prefixed ones.
+        # For dynamic builds we should ignore optarch completely and for optarch-set builds
+        # we need to adhere to TARGET and not march.
+        if self._dynamic_target is True or self._optarch_architecture is True:
+            self.log.info('Dynamic build requested, stripping march settings from environment variables')
+            for k in os.environ.keys():
+                optarch_to_strip = '-' + self._optarch_architecture
+                if 'EBVAR' not in k and self._optarch_architecture in os.environ[k]:
+                    os.environ[k] = os.environ[k].replace(optarch_to_strip, '')
+
         # Pass CFLAGS through command line to avoid redefinitions (issue xianyi/OpenBLAS#818)
         cflags = 'CFLAGS'
         if os.environ[cflags]:
@@ -72,7 +146,7 @@ class EB_OpenBLAS(ConfigureMake):
         run_cmd(cmd, log_all=True, simple=True)
 
     def test_step(self):
-        """ Mandatory test step plus optional runtest"""
+        """ Mandatory test step plus optional runtest """
 
         run_tests = ['tests']
         if self.cfg['runtest']:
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/o/openblas.py
```diff
@@ -44,3 +61,42 @@ def sanity_check_step(self):
             'dirs': [],
         }
         super(EB_OpenBLAS, self).sanity_check_step(custom_paths=custom_paths)
+    
+    def _parse_optarch( self, cfg_optarch):
+        """ Plucks the march out of the optarch settings. See
+        https://easybuild.readthedocs.io/en/latest/Controlling_compiler_optimization_flags.html """
+
+        pieces = cfg_optarch.split('=')
```
boegel: @terjekv We already have logic to parse the `optarch` value, I don't think it's a good idea to re-implement that here...

See `_postprocess_optarch` in https://github.com/easybuilders/easybuild-framework/blob/develop/easybuild/tools/options.py#L851 .

You currently can't easily leverage that code from an easyblock though, and you can't access the parsed value either currently, I think. That can be changed in framework though (in a backwards-compatible way)

easybuild/easyblocks/o/openblas.py
```diff
@@ -44,3 +61,42 @@ def sanity_check_step(self):
...
+                return pieces[3]
+        else:
+            return cfg_optarch
+
+    def _get_openblas_targets(self, targetfile):
+        """ Parse the openblas target file and generate a list of targets """
+        targets = []
+
+        with open( targetfile ) as f:
+            lines = f.readlines()
```
boegel: Use `lines = read_file(targetfile).splitlines()`, and have a backup plan in case the specified file doesn't exist?

easybuild/easyblocks/o/openblas.py
```diff
@@ -44,3 +61,42 @@ def sanity_check_step(self):
...
+        
+        return targets
+
+    def _get_mapped_target( self, cfg_optarch, openblas_targets ):
+        """ Attempts to find the desired optarch in the list of openblas targets.
+        Returns None if the target was not found. """
+
+        for target in openblas_targets:
+            if cfg_optarch.lower() == target.lower():
+                return target
```
boegel: We usually try to avoid inline `return`s (they make it harder to follow logic).

```python
result = None
for target in openblas_targets:
    if cfg_optarch.lower() == target.lower():
        result = target

return result

File: easybuild/easyblocks/o/openblas.py (hunk 1 above)
Comment (terjekv) [reply easybuild/easyblocks/o/openblas.py]: Right, this one will take a bit more time. Another work in progress.  :-)

File: easybuild/easyblocks/o/openblas.py (hunk 2 above)
Comment (terjekv) [reply easybuild/easyblocks/o/openblas.py]: Yup, fixed in https://github.com/easybuilders/easybuild-easyblocks/pull/1946/commits/d4da6d55bc8cdfe8534cfd4455adf07adcc5d972

File: easybuild/easyblocks/o/openblas.py (hunk 3 above)
Comment (terjekv) [reply easybuild/easyblocks/o/openblas.py]: Fixed in https://github.com/easybuilders/easybuild-easyblocks/pull/1946/commits/d4da6d55bc8cdfe8534cfd4455adf07adcc5d972

File: easybuild/easyblocks/o/openblas.py (hunk 1 above)
Comment (terjekv) [reply easybuild/easyblocks/o/openblas.py]: Okay, so I have done some work on the framework side, but I am unable to find any actual parsing of the contents of `optarch` in the framework. There are (limited) validation tests, but nothing that does something like "extract the march from this optarch". As such, the current iteration of the EasyBlock does not rely on changes in the framework, but a future PR will look at cleaning up some of the Framework code and moving generalised code from the EB to the Framework.

**Comment by terjekv** [comment]
@boegel and @tgamblin, I'm trying to return to resolving the lack of optarch support for OpenBLAS, but in a slightly more generic fashion. @boegel asked if leaning on https://github.com/archspec/archspec would be a good idea, and that most certainly seems like a very good idea indeed. However, there are a few thoughts on how to proceed:

1) My current thinking is that the goal should be to map an architecture specification, be it GCCs nomenclature, OpenBLAS' own, or something completely different, to a label from archspec. From there we can thus "translate" from one naming scheme to another at will.

2) The question the becomes of implementation. Should these distinct nomenclatures be integrated as "plugins" or something similar to archspec, or should these individual libraries be "consumers" of archspec, remaining completely distinct from the central project? Separation removes clutter, it means we will require more "bits" in the form of distinct libraries for the user to maintain.

3) Versioning of a given local nomenclature to archspec mappings. Archspec is already familiar with how to make certain supported compilers optimise for a given architecture, and what versions of the compilers are required, but we would need an even broader mapping for applications. The OpenBLAS [TargetList](https://github.com/xianyi/OpenBLAS/blob/develop/TargetList.txt) both adds and [renames](https://github.com/xianyi/OpenBLAS/commit/ea3a58c8442c9327b43e5cd2109865782759e6e8#diff-67620e0145e353d323940d99b2bc40d8) targets. This implies that a correct mapping will have to be release-dependent.

4) As a side note to this work... If we get this working, someone needs to decide the extent of EasyBuilds optarch support. Long term, support for cross-compiling will probably become a question that will require a definitive answer.

**Comment by tgamblin** [comment]
Did you look at the Spack `openblas` package?  The mapping from archspec to `openblas` is here:

* https://github.com/spack/spack/blob/develop/var/spack/repos/builtin/packages/openblas/package.py#L141

It's close enough that we just search for a target based on the archspec name, with a few special cases.

Given that OpenBLAS is one package, I'm not inclined to add its names to `archspec` -- I think the translation belongs in the package file.  If the goal is to avoid duplicating that work for lots of package managers, I'd say we should try to upstream archspec support to `openblas` -- it's already pretty close so they might go for something like this.

**Comment by terjekv** [comment]
Hi @tgamblin and thank you for your input! I poked at at the Spack implementation and it looks very smooth. I feel there needs to be a bit of version checking on the TargetList, but that's another problem all together.

As you say, cluttering the main archspec JSON is a bad idea, which is why I thought of secondary mappings or "plugins" to provide extended support. There are also cases beyond OpenBLAS that will eventually need addressing. However, in the short term, using archspec and the Spack implementation as a reference would solve the immediate issue, and then we can take it from there. Hm.

Thanks again for your input, very helpful!

File: easybuild/easyblocks/o/openblas.py [hunk 4]
```diff
@@ -94,3 +168,132 @@ def sanity_check_step(self):
...
+        :param compiler_optarch: An optarch for a given compiler.
+        https://easybuild.readthedocs.io/en/latest/Controlling_compiler_optimization_flags.html
+        """
+
+        target_arch = ''
+        pieces = compiler_optarch.split()
+
+        for piece in pieces:
+            spec = piece.split('=')
+            if spec[0] == 'march' or spec[0] == '-march':
```
Flamefire, line 209: What about `-mcpu`? I see that used on POWER

easybuild/easyblocks/o/openblas.py
```diff
@@ -39,6 +70,38 @@ def configure_step(self):
...
+                self.log.info("Optarch specified for %s, but no march detected", compiler_family)
+
+            if mapped_target is None:
+                print_warning("optarch for %s given as '%s'\n"
+                              "EasyBuild was unable to map this to an equivalent OpenBLAS target!\n"
+                              "OpenBLAS will be built to optimize its kernel at runtime!\n"
+                              % (self.toolchain.comp_family(), compiler_optarch))
+                self.log.warning("Unable to map %s to an OpenBLAS target, falling back to runtime optimization."
+                                 % compiler_optarch)
+                self._set_dynamic_architecture(default_opts)
```
Flamefire, line 98: I'm very unsure about this. If I build OpenBLAS currently, it seems to autodetect the current arch and optimizes for it. It seems to even overwrite our optarch settings etc which is fine when building for native. Additionally e.g. for POWER and OpenBLAS until 0.3.7 (IIRC) DYNAMIC_ARCH is broken to the point where even the simple tests fail (so in this case the EC wouldn't build)

easybuild/easyblocks/o/openblas.py
```diff
@@ -94,3 +168,132 @@ def sanity_check_step(self):
...
+        pieces = compiler_optarch.split()
+
+        for piece in pieces:
+            spec = piece.split('=')
+            if spec[0] == 'march' or spec[0] == '-march':
+                target_arch = spec[1]
+
+        return target_arch.lower()
+
+    def _get_openblas_targets(self, targetfile):
```
Flamefire, line 214: For all those functions: Could you make them free functions? In PyTorch/TensorFlow I found it a good idea so you can just import this from regular python and run/test it

easybuild/easyblocks/o/openblas.py
```diff
@@ -94,3 +168,132 @@ def sanity_check_step(self):
...
+        compiler = self.toolchain.comp_family()
+        compiler_specific_optarch_string = ''
+
+        if type(optarch) == str:
+            compiler_specific_optarch_string = optarch
+        elif type(optarch) == dict:
+            if compiler in optarch:
+                compiler_specific_optarch_string = optarch[compiler]
+        else:
+            raise EasyBuildError("optarch in an unexpected format: '%s'" % type(optarch)).__class__.__name__
```
Flamefire, line 192: Please remove the `.__class__.__name__`. Also maybe include the optarch value and use `,` instead of `%`

easybuild/easyblocks/o/openblas.py
```diff
@@ -94,3 +168,132 @@ def sanity_check_step(self):
...
+        :param optarch: A complete optarch statement.
+        https://easybuild.readthedocs.io/en/latest/Controlling_compiler_optimization_flags.html
+        """
+        if optarch is False:
+            return False
+
+        compiler = self.toolchain.comp_family()
+        compiler_specific_optarch_string = ''
+
+        if type(optarch) == str:
```
Flamefire, line 186: you can (and IIRC should) use `isinstance`

easybuild/easyblocks/o/openblas.py
```diff
@@ -57,6 +120,17 @@ def build_step(self):
                 build_parts += ['re_lapack']
         build_parts += ['shared']
 
+        # If we're doing either a dynamic build or utilizing optarch,
+        # strip march from all environment variables except the EBVAR-prefixed ones.
+        # For dynamic builds we should ignore optarch completely and for optarch-set builds
+        # we need to adhere to TARGET and not march.
+        if self._dynamic_target is True or self._optarch_architecture is True:
```
Flamefire, line 127: In which case can `self._optarch_architecture` be True? It is either False or a string isn't it?

easybuild/easyblocks/o/openblas.py
```diff
@@ -94,3 +168,132 @@ def sanity_check_step(self):
...
+
+    def _optarch_for_compiler(self, optarch):
+        """
+        Extracts the appropriate optarch for the compiler currently being used.
+        If it is not compiler-specific it is returned as-is.
+        If no optarch is found, False is returned.
+        :param optarch: A complete optarch statement.
+        https://easybuild.readthedocs.io/en/latest/Controlling_compiler_optimization_flags.html
+        """
+        if optarch is False:
```
Flamefire, line 180: For me (on POWER) optarch is `None`, so maybe `if not optarch:`? And I'd return `None` too as that seems to better fit the meaning

easybuild/easyblocks/o/openblas.py
```diff
@@ -94,3 +168,132 @@ def sanity_check_step(self):
...
+            except KeyError:
+                warning_string = "Archspec was asked to find '" + march + "' as an architecture, but failed!"
+                print_warning(warning_string)
+                self.log.warning(warning_string)
+                return None
+
+            if uarch.name in available_targets:
+                result = uarch.name
+            else:
+                self.log.info("No direct match for '" + march + "' between archspec and OpenBLAS, traversing ancestry.")
```
Flamefire, line 281: Don't we use formatted logs instead of string concatting?

Flamefire (changes requested):

easybuild/easyblocks/o/openblas.py
terjekv: We should certainly disable DYNAMIC_ARCH for Power. But the thing is, if you ask for `optarch=haswell` and OpenBLAS optimises for native, and you're on a skylake build host, your compute nodes will cease to function. For me (and I presume others), `optarch` is not something that may be ignored.

My strong preference is that if an EasyBuild user sets `optarch=haswell` EasyBuild offers the guarantee that the code produced will indeed run on haswell.

easybuild/easyblocks/o/openblas.py
terjekv: Hm, yeah. I think this was a suggested change from the editor of best practices for python. There was something odd about that. I'll check it out.

easybuild/easyblocks/o/openblas.py
terjekv: Yep, certainly.

easybuild/easyblocks/o/openblas.py
terjekv: Thanks, will do.

easybuild/easyblocks/o/openblas.py
terjekv: -mcpu is used, but it is a deprecated synonym for -mtune, which only takes two values on x86, generic and Intel, neither of which have an effect on OpenBLAS targeting, which only cares about CPU architecture. I have no idea what this option does on Power. :(

easybuild/easyblocks/o/openblas.py
terjekv: @lexming (I think) suggested generalising a number of these functions and shipping them off to Framework. The ones that are OpenBLAS specific should probably suffer a similar fate. This will require another rewrite though, not sure about the ETA on that.

easybuild/easyblocks/o/openblas.py
terjekv: Probably? I'll have to look at that.

easybuild/easyblocks/o/openblas.py
Flamefire: See https://stackoverflow.com/questions/42718572/gcc-mtune-vs-march-vs-mcpu

> -mcpu unfortunately has different semantics for different targets. It's deprecated for x86 (being a synonym for -mtune) but not for ARM, where it's a sum of -march and -mtune.

easybuild/easyblocks/o/openblas.py
Flamefire: Alright sure. I just wanted to highlight that the default (native) should still work as before (hence my restriction to "when building for native")

boegel:
@terjekv I think you'll agree with me that it's unlikely that this PR will get merged.

It was at least partially motivated by a need in EESSI, and we've started taking a different approach there in https://github.com/EESSI/software-layer/pull/260 by implementing a custom hook for OpenBLAS which injects `DYNAMIC_ARCH=1` when using `--optarch=GENERIC`.

That could/should probably also be implemented in the OpenBLAS easyblock, but in a different way than is being done here  - this PR does way more than just injecting `DYNAMIC_ARCH=1`.

terjekv:
I'd wager it could be merged if I rewrote it to use DYNAMIC_ARCH in a sane way when seeing optarch=GENERIC, but hooks work fine for now. So fine I've been using it for long while myself. 

Close and let a new, maybe working, PR arise if one ha the energy. :)

PR #1948, opened 2020-02-07
Add black format checking

ocaisa:
@boegel This is a proof of concept that only does `a` easyblocks. There are actually a lot of flake8 errors floating around in our easyblocks which I also had to resolve

### DIFF
--- a/easybuild/easyblocks/a/abaqus.py
+++ b/easybuild/easyblocks/a/abaqus.py
@@ -64,32 +64,29 @@ class EB_ABAQUS(Binary):
             setvar('NOLICENSECHECK', 'true')
         else:
             self.replayfile = os.path.join(self.builddir, "installer.properties")
-            txt = '\n'.join([
-                "INSTALLER_UI=SILENT",
-                "USER_INSTALL_DIR=%s" % self.installdir,
-                "MAKE_DEF_VER=true",
-                "DOC_ROOT=UNDEFINED",
-                "DOC_ROOT_TYPE=false",
-                "DOC_ROOT_ESCAPED=UNDEFINED",
-                "ABAQUSLM_LICENSE_FILE=@abaqusfea",
-                "LICENSE_SERVER_TYPE=FLEXNET",
-                "PRODUCT_NAME=Abaqus %s" % self.version,
-                "TMPDIR=%s" % self.builddir,
-                "INSTALL_MPI=1",
-            ])
+            txt = '\n'.join(
+                [
+                    "INSTALLER_UI=SILENT",
+                    "USER_INSTALL_DIR=%s" % self.installdir,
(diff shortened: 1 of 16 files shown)


easybuild/easyblocks/a/abaqus.py
```diff
@@ -64,32 +64,29 @@ def configure_step(self):
...
             change_dir(os.path.join(self.cfg['start_dir'], '1'))
-            qa = {
-                "Enter selection (default: Install):": '',
-            }
-            no_qa = [
-                '___',
-                '\(\d+ MB\)',
-            ]
+            qa = {"Enter selection (default: Install):": ''}
+            no_qa = ['___', r'\(\d+ MB\)']
```
boegel: This is going backwards, not forwards imho... ;)

Flamefire: That's always the problem with auto-formatters: You'll hit situations where manual formatting would be better. As an alternative to black I'd suggest yapf which allows for more fine grained control over formatting similar to clang-format. Anyhow formatting will always be opinionated so just decide on 1 auto-formatter and configuration and get over it. ;)

setup.cfg
```diff
@@ -15,4 +15,4 @@ builtins =
 
 # ignore "Black would make changes" produced by flake8-black
 # see also https://github.com/houndci/hound/issues/1769
-ignore = BLK100
+ignore = BLK100,E121,E123,E126,E226,E24,E704,W503,W504
```
Flamefire: That's a lot ignored. What are those and why are they ignored? I'd rather not do that.

pyproject.toml
```diff
@@ -0,0 +1,19 @@
+[tool.black]
+line-length = 120
+skip-string-normalization = true
+target-version = ['py27','py35','py36', 'py37', 'py38']
```
Flamefire: EB still supports py26, doesn't it?

setup.cfg
ocaisa: This is outdated now, I was manually adding the default ignores of flake8. It should be
```
extend-ignore = BLK100
```

pyproject.toml
ocaisa: Yes, but `black` doesn't. TBH I'm not sure that this is of any real use anyway.

ocaisa:
@Flamefire @shahzebsiddiqui I've added the pre-commit hook configuration here as well...but this still needs a wider discussion (and someone willing to spend the time to go over the entire repo and fix the flake8 problems)

shahzebsiddiqui:
I get in since easybuild code base is too big. Either you do incremental changes with yapf or black doesn't matter or abandon formatting completely. I don't mind whatever decision you guys take. For buildtest we atleast made the black --check a regular GitHub check which is useful as first step

PR #1949, opened 2020-02-07
Update config.guess for all R packages

Flamefire:
Factor out the check and obtain config.guess from ConfigureMake into free functions to be able to use them inside the rpackage EB.
Change the workflow for R packages to always extract which is more in line with the regular workflow for other EasyConfigs.
This then allows to search for config.guess files to replace by the downloaded version contained within EasyBuild.

This is an alternate, more complete approach to #1938 

I tested some random R based ECs including the underlying R successfully: rgdal-1.4-8-foss-2019b-R-3.6.2.eb ncdf4-1.16.1-foss-2019a-R-3.6.0.eb lavaan-0.6-4.1433-foss-2019a-R-3.6.0.eb

Requires:
- [x] https://github.com/easybuilders/easybuild-framework/pull/3195


easybuild/easyblocks/generic/configuremake.py
```diff
@@ -67,6 +67,101 @@
...
+    try:
+        config_guess_timestamp = datetime.fromtimestamp(os.stat(config_guess).st_mtime).isoformat()
+    except OSError as err:
+        if log:
+            log.warning("Failed to determine timestamp of %s: %s", config_guess, err)
+        config_guess_timestamp = None
+
+    # log version, timestamp & SHA256 checksum of config.guess
+    if log:
+        log.info("config.guess version: %s (last updated: %s, SHA256 checksum: %s)",
```
boegel: @Flamefire You should introduce a top-level `_log` like we do in other easyblocks, and use that instead?

```python
_log = fancylogger.getLogger('easyblocks.generic.configuremake')

File: easybuild/easyblocks/generic/configuremake.py [hunk 2]
```diff
@@ -67,6 +67,101 @@
...
+            # verify SHA256 checksum of download to avoid using a corrupted download
+            if verify_checksum(downloaded_path, CONFIG_GUESS_SHA256):
+                config_guess_path = downloaded_path
+                # add execute permissions
+                adjust_permissions(downloaded_path, stat.S_IXUSR | stat.S_IXGRP | stat.S_IXOTH, add=True)
+                if log:
+                    log.info("Downloaded recent %s to %s, using it if required", config_guess, config_guess_path)
+            else:
+                if log:
+                    log.warning("Checksum failed for downloaded file %s, not using it!", downloaded_path)
```
boegel: same here, use top-level `_log`

easybuild/easyblocks/generic/configuremake.py
```diff
@@ -67,6 +67,101 @@
...
+        log.info("config.guess version: %s (last updated: %s, SHA256 checksum: %s)",
+                 config_guess_version, config_guess_timestamp, config_guess_checksum)
+
+    result = True
+
+    if config_guess_version != CONFIG_GUESS_VERSION:
+        result = False
+        tup = (config_guess, config_guess_version, CONFIG_GUESS_VERSION)
+        if log:
+            log.warning("config.guess version at %s does not match expected version: %s vs %s" % tup)
```
boegel: stick to using `print_warning` here (which is a verbose warning, as opposed to just a warning in the log file)?

easybuild/easyblocks/generic/configuremake.py
```diff
@@ -67,6 +67,101 @@
 DEFAULT_INSTALL_CMD = 'make install'
 
 
+def check_config_guess(config_guess, log=None):
+    """Check timestamp & SHA256 checksum of config.guess script.
+
+    :param configu_gess: Path to config.guess script to check
```
boegel: typo: `config_guess`

easybuild/easyblocks/generic/rpackage.py
```diff
@@ -132,12 +133,8 @@ def make_cmdline_cmd(self, prefix=None):
         else:
             prefix = ''
 
-        if self.cfg['unpack_sources']:
-            loc = self.start_dir
-        elif self.patches:
-            loc = self.ext_dir
-        else:
-            loc = self.ext_src
+        loc = self.start_dir or self.ext_dir or self.ext_src
```
boegel: How confident are you that this code cleanup is OK to do? It's not exactly equivalent to what was there...

easybuild/easyblocks/generic/rpackage.py
```diff
@@ -152,23 +149,6 @@ def make_cmdline_cmd(self, prefix=None):
...
-            raise EasyBuildError("Don't know how to handle R packages with multiple sources.'")
-        else:
-            try:
-                shutil.copy2(self.src[0]['path'], self.builddir)
-            except OSError as err:
-                raise EasyBuildError("Failed to copy source to build dir: %s", err)
-            self.ext_src = self.src[0]['name']
-
-            # set final path since it can't be determined from unpacked sources (used for guessing start_dir)
-            self.src[0]['finalpath'] = self.builddir
```
boegel: How is it OK to drop this custom `extract_step` implementation?

easybuild/easyblocks/generic/rpackage.py
```diff
@@ -195,9 +175,19 @@ def install_R_package(self, cmd, inp=None):
         else:
             self.log.debug("R package %s installed succesfully" % self.name)
 
+    def update_config_guess(self, src_dir):
+        """Update any config.guess found in src_dir"""
```
boegel: `... in specified directory`

easybuild/easyblocks/generic/rpackage.py
```diff
@@ -195,9 +175,19 @@ def install_R_package(self, cmd, inp=None):
         else:
             self.log.debug("R package %s installed succesfully" % self.name)
 
+    def update_config_guess(self, src_dir):
```
boegel: `src_dir` -> `path`?

easybuild/easyblocks/generic/rpackage.py
```diff
@@ -195,9 +175,19 @@ def install_R_package(self, cmd, inp=None):
...
+        """Update any config.guess found in src_dir"""
+        for config_guess_dir in (root for root, _, files in os.walk(src_dir) if 'config.guess' in files):
+            config_guess = os.path.join(config_guess_dir, 'config.guess')
+            if not check_config_guess(config_guess):
+                updated_config_guess = obtain_config_guess(log=self.log)
+                copy_file(updated_config_guess, config_guess)
```
boegel: This deserves some logging, like `"Replacing outdated /path/to/config.guess with more recent copy available at /path/to/new/config.guess"`

easybuild/easyblocks/generic/rpackage.py
```diff
@@ -215,16 +205,15 @@ def run(self):
             lib_install_prefix = os.path.join(self.installdir, self.cfg['exts_subdir'])
             mkdir(lib_install_prefix, parents=True)
 
-        if self.patches:
-            super(RPackage, self).run(unpack_src=True)
-        else:
-            super(RPackage, self).run()
-
         if self.src:
+            super(RPackage, self).run(unpack_src=True)
```
boegel: This is also significantly changing the logic, why don't we need to check for patches anymore?

easybuild/easyblocks/generic/configuremake.py
Flamefire: The idea was to be able to NOT pass a log to have it be quiet. And I don't understand the global logger: Isn't it supposed to be the logger from the calling class? In this case the rpackage?

easybuild/easyblocks/generic/rpackage.py
Flamefire: Very. start_dir is set to the path to the extracted sources when used as a EasyBlock. ext_dir is set for use as an Extension. ext_src is just a fallback if neither is set.

The basic idea is to always extract, see https://github.com/easybuilders/easybuild-easyblocks/pull/1938#issuecomment-582362995

easybuild/easyblocks/generic/rpackage.py
Flamefire: Yes, In https://github.com/easybuilders/easybuild-easyblocks/pull/1938#issuecomment-582362995 I suggested to always extract, see also PR description. Hence there is no need to do the "maybe extract or copy"-dance

easybuild/easyblocks/generic/rpackage.py
Flamefire: Isn't https://github.com/easybuilders/easybuild-easyblocks/pull/1949/files#diff-aa6aaedfce9df12a4136ab4413f7971aR154 enough? But I can add it as a `debug` log entry

easybuild/easyblocks/generic/rpackage.py
Flamefire: Previous: Extract only if there are patches   
Now: Extract always because config.guess needs to be "patched" (so there are "patches")

Placed inside `self.src` check because the super-`run` was a no-op anyway if neither patches nor src were set

edmondac:
I've tried testing this, but there are too many things in the way. E.g. we need https://github.com/easybuilders/easybuild-easyconfigs/pull/9694 merging. Also older R version have checksum failures (as expected) and I'm not able to go through and fix them right now.

Saying that, it does seem to work for the bits that did build!

easybuild/easyblocks/generic/configuremake.py
Flamefire: Where is the warning function (from log) defined? Can it also print to stderr? Because print_warning (inconsistently) does not accept a log param to also print to log.

Flamefire: See https://github.com/easybuilders/easybuild-framework/pull/3195

easybuild/easyblocks/generic/configuremake.py
boegel: Why would you want it to be quiet, we're logging useful information here?

Passing a logger is something we should avoid, and since this function isn't part of an easyblock anymore, using a module-level logger is fine. Another option is to create a custom logger in this function, see `pythonpackage.py`

easybuild/easyblocks/generic/rpackage.py
boegel: OK, makes sense.

easybuild/easyblocks/generic/rpackage.py
boegel: Thanks for the clarification!

easybuild/easyblocks/generic/rpackage.py
```diff
@@ -195,9 +174,20 @@ def install_R_package(self, cmd, inp=None):
...
             self.log.debug("R package %s installed succesfully" % self.name)
 
+    def update_config_guess(self, path):
+        """Update any config.guess found in specified directory"""
+        for config_guess_dir in (root for root, _, files in os.walk(path) if 'config.guess' in files):
+            config_guess = os.path.join(config_guess_dir, 'config.guess')
+            if not check_config_guess(config_guess):
+                updated_config_guess = obtain_config_guess(log=self.log)
+                copy_file(updated_config_guess, config_guess)
+                self.log.debug("Replacing outdated %s with more recent copy at %s", config_guess, updated_config_guess)
```
boegel: Either make it `Replaced`, or move up this `log.debug` call (and maybe make it `log.info`)?

easybuild/easyblocks/generic/rpackage.py
boegel: OK, thanks for clarifying

easybuild/easyblocks/generic/configuremake.py
Flamefire: You are right. Using a logger where the name is shared between the 2 function which IMO is reasonable as they belong together. One now even calls the other

Flamefire:
@boegel Updated this with the following changes worth verifying:

- Use logger (same for check and obtain function)
- Verify existing file before using it (this avoids a cached config.guess to fail the build if it is an old one, see the removed comment about invalidating the cache which is kinda impossible from the EB side)
- Verify downloaded file through check_config_guess
- Honor force_download build option
- downgrade the print_warning to a log.warning again, but use print_warning in obtain_config_guess when download or verification fails. Otherwise the build output has a few warnings which are not really warnings because the file will be replaced later. I'd actually downgrade the warnings to info in check_config_guess if you agree

Pending approval: Remove arguments of `obtain_config_guess`. No EasyBlock uses them and Easyblock.obtain_file doesn't have those either. Also the final download location is pretty much opaque to the caller (appended with e.g. EB version) so it doesn't make sense to pass it

Flamefire:
@boegel Test results in https://github.com/easybuilders/easybuild-easyconfigs/pull/9830 show that this works and fixes the issue on Power.

What I like to do before merging this is to remove arguments of obtain_config_guess, see https://github.com/easybuilders/easybuild-easyblocks/pull/1949#issuecomment-593990282
This does not change behavior but makes things easier to maintain

Flamefire:
I did the change I proposed now for easier assesment in https://github.com/easybuilders/easybuild-easyblocks/pull/1949/commits/8af677b9d339eaa45deefa80ae0797d6d73c033c I even found that using the source-* argument didn't work at all which is another point for the change.

The current implementation is to raise an error that those arguments were removed. This caters for external users using that function. The alternative is to simply remove the arguments and let Python handle it by throwing a (IIRC) `TypeError` on unknown/unexpected arguments. But I guess the current implementation is more in line with EB handling such things. 

Not using deprecation on purpose as the implementation contained at least 1 bug. So removing it is safer IMO.

Flamefire:
@boegel *ping* again. This has been tested successfully in https://github.com/easybuilders/easybuild-easyconfigs/pull/10531 multiple times.

boegel:
I've tested this extensively by installing `R-4.0.0-foss-2020a.eb`, `R-3.6.3-foss-2020a.eb` and `R-3.4.3-foss-2017b-X11-20171023.eb` using this enhanced `RPackage` + a whole bunch of dependencies using`ConfigureMake`, no problems encountered, so this is good to go...

PR #1952, opened 2020-02-11
openblas: remove tests from build_step and raise error on failed tests

lexming:
Update for `openblas` easyblock to detect and raise an error on failed tests. Currently OpenBLAS is running its tests in the build step and the installation proceed regardless of their result.

* `build_step` will run a custom make command that is equivalent to `make all` but without the tests
* `test_step` will run the tests and raise an exception if any test fails.

Additional issues to be solved:
* (already present in current easyblock) many compilation variables are redefined. Example
> gcc -O2 -ftree-vectorize -march=native -fno-math-errno -O2 -DMAX_STACK_ALLOC=2048 -fopenmp -Wall -m64 -DF_INTERFACE_GFORT -fPIC -      DSMP_SERVER -DUSE_OPENMP -DNO_WARMUP -DMAX_CPU_NUMBER=24 -DMAX_PARALLEL_NUMBER=1 -DVERSION=\"0.3.7\" -march=skylake-avx512 **-DASMNAME= -DASMFNAME=_ -DNAME=_ -DCNAME= -DCHAR_NAME=\"_\" -DCHAR_CNAME=\"\"** -DNO_AFFINITY -I. *-O2 -DMAX_STACK_ALLOC=2048 -fopenmp -Wall -m64 -DF_INTERFACE_GFORT -fPIC -DSMP_SERVER -DUSE_OPENMP -DNO_WARMUP -DMAX_CPU_NUMBER=24 -DMAX_PARALLEL_NUMBER=1 -DVERSION=\"0.3.7\" -march=skylake-avx512 **-DASMNAME=cblas_isamax -DASMFNAME=cblas_isamax_ -DNAME=cblas_isamax_ -DCNAME=cblas_isamax -DCHAR_NAME=\"cblas_isamax_\" -DCHAR_CNAME=\"cblas_isamax\"** -DNO_AFFINITY -I.. -I. -UDOUBLE  -UCOMPLEX -DCBLAS -c -DUSE_ABS -UUSE_MIN imax.c -o cblas_isamax.o
* (introduced by this PR) Header file `lapack.h` is not installed

### DIFF
--- a/easybuild/easyblocks/o/openblas.py
+++ b/easybuild/easyblocks/o/openblas.py
@@ -47,6 +47,13 @@ class EB_OpenBLAS(ConfigureMake):
                 build_parts += ['re_lapack']
         build_parts += ['shared']
 
+        # Pass CFLAGS through command line to avoid redefinitions (issue xianyi/OpenBLAS#818)
+        cflags = 'CFLAGS'
+        if os.environ[cflags]:
+            self.cfg.update('buildopts', "%s='%s'" % (cflags, os.environ[cflags]))
+            del os.environ[cflags]
+            self.log.info("Environment variable %s unset and passed through command line" % cflags)
+
         cmd = "%s make %s %s" % (self.cfg['prebuildopts'], ' '.join(build_parts), self.cfg['buildopts'])
         run_cmd(cmd, log_all=True, simple=True)
 
--- a/easybuild/easyblocks/o/openblas.py
+++ b/easybuild/easyblocks/o/openblas.py
@@ -51,13 +51,18 @@ class EB_OpenBLAS(ConfigureMake):
         run_cmd(cmd, log_all=True, simple=True)
 
     def test_step(self):
-        """ Mandatory test step """
+        """ Mandatory test step plus optional runtest"""
 
         cmd = "%s make tests %s" % (self.cfg['pretestopts'], self.cfg['testopts'])
-        (out, ec) = run_cmd(cmd, log_all=True, simple=False, regexp=False)
+        (out, _) = run_cmd(cmd, log_all=True, simple=False, regexp=False)
 
         # Raise an error if any test failed
         check_log_for_errors(out, [('FATAL ERROR', ERROR)])
+
+        # Run any optional tests
+        if self.cfg['runtest']:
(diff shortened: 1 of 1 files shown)


akesandgren:
FYI, lapack.h has never been installed, lapacke.h is the one that gets installed.

martin-frbg:
The lapack.h thing is a recent change by Reference-LAPACK a.k.a netlib - in 3.9.0 the pure lapack definitions have been split out into their own header file for easier use with a future LAPACK++
The redefiniton of variables is (usually) an unwanted side effect of passing CFLAGS via the environment (OpenBLAS issue 818, I tried to fix it a few years ago but got nowhere with the current mess of makefiles that tend to include each other )

lexming:
@martin-frbg Thank you very much for the feedback, it's been really helpful to better understand this issues.

lexming:
@akesandgren @boegel all issues fixed, no longer a WIP

akesandgren:
Tested on broadwell and skylake with https://github.com/easybuilders/easybuild-easyconfigs/pull/9852
broadwell -> no problem
skyake -> caught the fatal errors

No duplication of variable settings.
lapack.h installed.

Perfect...

akesandgren:
Going in, thanks @lexming!

easybuild/easyblocks/o/openblas.py
```diff
@@ -29,12 +31,46 @@ def configure_step(self):
...
+        build_parts += ['shared']
+
+        # Pass CFLAGS through command line to avoid redefinitions (issue xianyi/OpenBLAS#818)
+        cflags = 'CFLAGS'
+        if os.environ[cflags]:
+            self.cfg.update('buildopts', "%s='%s'" % (cflags, os.environ[cflags]))
+            del os.environ[cflags]
+            self.log.info("Environment variable %s unset and passed through command line" % cflags)
+
+        cmd = "%s make %s %s" % (self.cfg['prebuildopts'], ' '.join(build_parts), self.cfg['buildopts'])
```
boegel: @lexming You accidentally disabled parallel building (`make -j`) here, which is done by default by `ConfigureMake.build_step`... fixed in #1962

lexming: I totally missed that, thanks for pointing it out

PR #1955, opened 2020-02-12
openblas: add lexming as author

lexming:
I forgot to add my attribution to the contribution made in PR #1952


akesandgren:
Going in, thanks @lexming!

PR #1956, opened 2020-02-12
update Ferret easyblock for versions >= 7.5.0

smoors:


easybuild/easyblocks/f/ferret.py
```diff
@@ -75,15 +79,32 @@ def configure_step(self):
...
                 (r"^FER_LOCAL_EXTFCNS = $(FER_DIR)", "FER_LOCAL_EXTFCNS = $(INSTALL_FER_DIR)/libs"),
             ])
 
+        if LooseVersion(self.version) >= LooseVersion("7.5"):
+            comp_vars = {
+                'CC': 'CC',
+                'FC': 'FC',
+            }
+
+            for x, y in comp_vars.items():
```
boegel: @smoors Please avoid using single-letter variable names outside of list comprehension

smoors: ok, done. there were a lot of other cases like this already present.

boegel:
tested with existing `Ferret-7.3-intel-2017b.eb` easyconfig + https://github.com/easybuilders/easybuild-easyconfigs/pull/9862, good to go, thanks @smoors!

PR #1958, opened 2020-02-12
openblas: add optional runtest to the catch for FATAL ERRORS

lexming:
Optional tests added with the `runtest` option were left out from the catch of `FATAL ERRORS`. However, tests such as `blas-test` can behave in a similar way as the default `make tests`, revealing build errors that are going unnoticed by EB.
This PR fixes this issue by parsing all test outputs and raising an error if `FATAL ERROR` is found in any of them.


akesandgren:
Going in, thanks @lexming!

PR #1960, opened 2020-02-13
openblas: pass any TARGET in buildopts to testopts and installopts

lexming:


easybuild/easyblocks/o/openblas.py
```diff
@@ -26,7 +26,14 @@ def configure_step(self):
             'USE_OPENMP': '1',
             'USE_THREAD': '1',
         }
-        if LooseVersion(self.version) < LooseVersion('0.3.6') and get_cpu_architecture() == POWER:
+
+        if 'TARGET' in self.cfg['buildopts']:
```
boegel: I would check for `TARGET=` to make it a little bit more strict (since you rely on `TARGET=` below basically)

easybuild/easyblocks/o/openblas.py
```diff
@@ -26,7 +26,14 @@ def configure_step(self):
...
+            # Add any TARGET in buildopts to default_opts, so it is passed to testopts and installopts
+            for buildopt in self.cfg['buildopts'].split():
+                optpair = buildopt.split('=')
+                if optpair[0] == 'TARGET':
```
boegel: Usually when we use a "magic string" in more than one place, we introduce a constant.

```python
TARGET = 'TARGET'
```

```python
if (TARGET + '=') in self.cfg['buildopts']:
    ...
    if optpair[0] == TARGET:
```

May look like overkill, but it helps avoid bugs.

easybuild/easyblocks/o/openblas.py
```diff
@@ -26,7 +26,14 @@ def configure_step(self):
...
+                    default_opts[optpair[0]] = optpair[1]
+        elif LooseVersion(self.version) < LooseVersion('0.3.6') and get_cpu_architecture() == POWER:
             # There doesn't seem to be a POWER9 option yet, but POWER8 should work.
             print_warning("OpenBLAS 0.3.5 and lower have known issues on POWER systems")
             default_opts['TARGET'] = 'POWER8'
```
boegel: Also use `TARGET` constant here if you introduce it

easybuild/easyblocks/o/openblas.py
lexming: Makes perfect sense, fixed :+1:

easybuild/easyblocks/o/openblas.py
lexming: Done :+1:

easybuild/easyblocks/o/openblas.py
lexming: Done :+1:

lexming:
Tested with `OpenBLAS-0.3.7-GCC-8.3.0.eb` and `OpenBLAS-0.3.8-GCC-9.2.0.eb` from `develop` by adding 
```
buildopts = ['TARGET=GENERIC']
```

easybuild/easyblocks/o/openblas.py
```diff
@@ -26,10 +26,18 @@ def configure_step(self):
             'USE_OPENMP': '1',
             'USE_THREAD': '1',
         }
-        if LooseVersion(self.version) < LooseVersion('0.3.6') and get_cpu_architecture() == POWER:
+
+        TARGET = 'TARGET'
```
boegel: @lexming Please move declaring the constant outside of the class (so at line 15, with 2 blank lines after it)

lexming: that's indeed much prettier, fixed :+1:

akesandgren:
@lexming it's just
buildopts = 'TARGET=GENERIC '

akesandgren:
@boegel how do you feel about the '%s=' % TARGET way of doing the test compared to your suggestion?

lexming:
@akesandgren just to clarify, I used the same syntax found in other parts of the easyblock. Check line 45 for instance.

akesandgren:
Yeah, just want to make sure there are no complaints later on :-)

akesandgren:
Going in, thanks @lexming!

boegel:
> @boegel how do you feel about the '%s=' % TARGET way of doing the test compared to your suggestion?

:+1:

PR #1961, opened 2020-02-13
ensure right 'python' command is used to determine Python lib dir when system Python is used with PythonBundle easyblock

boegel:
required for https://github.com/easybuilders/easybuild-easyconfigs/pull/9859 in case `python` is available in the OS, but `python3` is used to run EasyBuild; without this fix, `$PYTHONPATH` gets set incorrectly (using `lib/python2.7/site-packages` rather than `lib/python3.6/site-packages`) in the generated module file...

similar fix was done in `PythonPackage` in #1907, but that doesn't help with `PythonBundle`...


akesandgren:
Going in, thanks @boegel!

PR #1962, opened 2020-02-13
reinstate parallel building of OpenBLAS

boegel:
(was accidentally disabled in #1952)

### DIFF
--- a/easybuild/easyblocks/o/openblas.py
+++ b/easybuild/easyblocks/o/openblas.py
@@ -55,11 +55,11 @@ class EB_OpenBLAS(ConfigureMake):
             del os.environ[cflags]
             self.log.info("Environment variable %s unset and passed through command line" % cflags)
 
-        paracmd = ''
+        makecmd = 'make'
         if self.cfg['parallel']:
-            paracmd = "-j %s" % self.cfg['parallel']
+            makecmd += "-j %s" % self.cfg['parallel']
 
-        cmd = ' '.join([self.cfg['prebuildopts'], 'make', paracmd, ' '.join(build_parts), self.cfg['buildopts']])
+        cmd = ' '.join([self.cfg['prebuildopts'], makecmd, ' '.join(build_parts), self.cfg['buildopts']])
         run_cmd(cmd, log_all=True, simple=True)
 
     def test_step(self):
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/o/openblas.py
```diff
@@ -55,7 +55,11 @@ def build_step(self):
             del os.environ[cflags]
             self.log.info("Environment variable %s unset and passed through command line" % cflags)
 
-        cmd = "%s make %s %s" % (self.cfg['prebuildopts'], ' '.join(build_parts), self.cfg['buildopts'])
+        paracmd = ''
+        if self.cfg['parallel']:
+            paracmd = "-j %s" % self.cfg['parallel']
+
+        cmd = ' '.join([self.cfg['prebuildopts'], 'make', paracmd, ' '.join(build_parts), self.cfg['buildopts']])
```
lexming: Sorry for the nitpicking, but having a variable called `paracmd` that has no command in it might be confusing, I would suggest the following
```
        makecmd = 'make'
        if self.cfg['parallel']:
            makecmd = "%s -j %s" % (makecmd, self.cfg['parallel'])

        cmd = ' '.join([self.cfg['prebuildopts'], makecmd, ' '.join(build_parts), self.cfg['buildopts']])
```

easybuild/easyblocks/o/openblas.py
```diff
@@ -55,7 +55,11 @@ def build_step(self):
             del os.environ[cflags]
             self.log.info("Environment variable %s unset and passed through command line" % cflags)
 
-        cmd = "%s make %s %s" % (self.cfg['prebuildopts'], ' '.join(build_parts), self.cfg['buildopts'])
+        makecmd = 'make'
+        if self.cfg['parallel']:
+            makecmd += "-j %s" % self.cfg['parallel']
```
lexming: Ther is one space missing
```suggestion
            makecmd += " -j %s" % self.cfg['parallel']
```

boegel: I did notice that, but forgot to squeeze in the space correctly before pushing... 🤦‍♂ 

Thanks for checking this carefully!

lexming:
I reply my own question. The test step is not parallelized in `configuremake`, so we will do the same for `openblas`.

lexming:
Tested with `OpenBLAS-0.3.7-GCC-8.3.0.eb` and `OpenBLAS-0.3.8-GCC-9.2.0.eb` from `develop` and it is working as intended.

LGTM

akesandgren:
@lexming You need to approve the " " change so we can merge this.

PR #1963, opened 2020-02-14
NAMD: allow building with a separate charm++.

bartoldeman:
You just need to specify that in the easyconfig's "sources" parameter
and then it will be used instead of the included charm tarball.

We needed this to compile NAMD with UCX (using charm++ 6.10), as we're in the same situation as TACC:
https://www.ks.uiuc.edu/Research/namd/wiki/index.cgi?NamdAtTexas


bartoldeman:
I found an issue when there is no separate charm++. I'll fix it later.

akesandgren:
@bartoldeman still waiting for your "I'll fix it later", and there is a conflict to handle too.

PR #1964, opened 2020-02-17
add custom easyblock for LAMMPS

Darkless012:


easybuild/easyblocks/l/lammps.py
```diff
@@ -0,0 +1,362 @@
...
+
+            # if KOKKOS and CUDA
+            if cuda:
+                self.check_cuda_compute_capabilities(cfg_cuda_cc, ec_cuda_cc, cuda_cc)
+                nvcc_wrapper_path = os.path.join(self.start_dir, "lib", "kokkos", "bin", "nvcc_wrapper")
+                # nvcc_wrapper can't find OpenMP for some reason, need to specify OpenMP flags.
+                # https://software.intel.com/en-us/forums/intel-oneapi-base-toolkit/topic/842470
+                self.cfg.update('configopts', '-DOpenMP_CXX_FLAGS="-qopenmp"')
+                self.cfg.update('configopts', '-DOpenMP_CXX_LIB_NAMES="libiomp5"')
+                self.cfg.update('configopts', '-DOpenMP_libiomp5_LIBRARY=$EBROOTICC/lib/intel64_lin/libiomp5.so')
```
ocaisa: This is clearly toolchain specific behaviour so you will check the toolchain so that the behaviour is valid.

easybuild/easyblocks/l/lammps.py
```diff
@@ -0,0 +1,362 @@
...
+
+        if self.cfg['user_packages']:
+            for package in self.cfg['user_packages']:
+                self.cfg.update('configopts', '-DPKG_USER-%s=on' % package)
+
+        # Optimization settings
+        if '-DPKG_OPT=' not in self.cfg['configopts']:
+            self.cfg.update('configopts', '-DPKG_OPT=on')
+
+        if '-DPKG_USR-INTEL=' not in self.cfg['configopts']:
```
ocaisa: You would need to be using an Intel compiler for this to be valid so best to check the toolchain compiler family before setting this.

ocaisa:
@Darkless012 Thanks for looking into this, this has been on my todo list forever

easybuild/easyblocks/l/lammps.py
ocaisa: @boegel Do we have generic variables we can leverage for this case?

akesandgren: I think we do, LIB_MULTITHREAD

Darkless012: Good catch :) Just found out that this might not be needed - might be an issue with intel compilers not finding stuff. Trying workaround.

easybuild/easyblocks/l/lammps.py
Darkless012: I might get that wrong, but from what I've read I think that you don't need intel compilers for this: https://lammps.sandia.gov/doc/Speed_intel.html 
Those should be able to run being compiled with GCC as well.

easybuild/easyblocks/l/lammps.py
Darkless012: unsetting LIBS in easyconfig resolves the issue with compiler having trouble finding stuff here.

easybuild/easyblocks/l/lammps.py
```diff
@@ -0,0 +1,353 @@
...
+            warning_msg += "but 'cuda_compute_capabilities' option was specified."
+            print_warning(warning_msg)
+
+        return cuda_cc
+
+    def get_cpu_arch(self):
+        out, ec = run_cmd("python -c 'from archspec.cpu import host; print(host())'", simple=False)
+        if ec:
+            raise EasyBuildError("Failed to determine CPU architecture: %s", out)
+        # transform: 'skylake_avx512\n' => 'skylake'
```
boegel: I wouldn't mangle the output obtained from `archspec`.

`skylake` is actually "Skylake desktop", which doesn't have AVX512. So the Skylake expected by LAMMPS (to use `SKX` is actually `skylake_avx512` (sometimes called "Skylake X")

easybuild/easyblocks/l/lammps.py
```diff
@@ -0,0 +1,353 @@
...
+
+            cpu_arch = self.get_cpu_arch()
+
+            if cpu_arch == "sandybridge" or cpu_arch == "ivybridge":
+                processor_arch = 'SNB'
+            elif cpu_arch == "haswell":
+                processor_arch = 'HSW'
+            elif cpu_arch == "broadwell":
+                processor_arch = 'BDW'
+            elif cpu_arch == "skylake":
```
boegel: this should be `skylake_avx512` (a.k.a Skylake X)

easybuild/easyblocks/l/lammps.py
```diff
@@ -0,0 +1,353 @@
...
+# EasyBuild is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with EasyBuild.  If not, see <http://www.gnu.org/licenses/>.
+##
+"""
+@author: Pavel Grochal (INUITS)
```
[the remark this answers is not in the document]
Darkless012: I'm adding the original CMakeMake I've used as skeleton. 
The easyblock was done from scratch - I've started from semi-working CMakeMake and extracted bits out.

https://github.com/easybuilders/easybuild-easyconfigs/blob/0c7fa07b9b7a855df6d14b971bd9eb1a25f51dd8/easybuild/easyconfigs/l/LAMMPS/LAMMPS-24Oct2018-intel-2018b.eb

easybuild/easyblocks/l/lammps.py
```diff
@@ -0,0 +1,353 @@
...
+            if cpu_arch == "sandybridge" or cpu_arch == "ivybridge":
+                processor_arch = 'SNB'
+            elif cpu_arch == "haswell":
+                processor_arch = 'HSW'
+            elif cpu_arch == "broadwell":
+                processor_arch = 'BDW'
+            elif cpu_arch == "skylake":
+                processor_arch = 'SKX'
+            elif cpu_arch == "knights-landing":
+                processor_arch = 'KNL'
```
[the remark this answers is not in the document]
Darkless012: fair point :)

easybuild/easyblocks/l/lammps.py
```diff
@@ -0,0 +1,353 @@
...
+        cuda_cc = cfg_cuda_cc or ec_cuda_cc or []
+
+        # cmake has its own folder
+        self.cfg['srcdir'] = os.path.join(self.start_dir, 'cmake')
+
+        # verbose CMake
+        if '-DCMAKE_VERBOSE_MAKEFILE=' not in self.cfg['configopts']:
+            self.cfg.update('configopts', '-DCMAKE_VERBOSE_MAKEFILE=yes')
+
+        # Enable following packages, if not configured in easycofig
```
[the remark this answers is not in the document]
Darkless012: I've used the terminology LAMMPS has in documentation: https://lammps.sandia.gov/doc/Packages_standard.html

But you are absolutely right, the term is build options.

easybuild/easyblocks/l/lammps.py
```diff
@@ -0,0 +1,353 @@
...
+        # Is there a libjpeg-turbo
+        if '-DWITH_JPEG=' not in self.cfg['configopts']:
+            if get_software_root('libjpeg-turbo'):
+                self.cfg.update('configopts', '-DWITH_JPEG=yes')
+            else:
+                self.cfg.update('configopts', '-DWITH_JPEG=no')
+
+        # With Eigen dependency:
+        if '-DDOWNLOAD_EIGEN3=' not in self.cfg['configopts']:
+            self.cfg.update('configopts', '-DDOWNLOAD_EIGEN3=no')
```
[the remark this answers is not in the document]
Darkless012: I think this should stay. If I remember correctly the LAMMPS was failing to build with self-downloaded Eigen probably because of wrong or duplicated paths.

I don't think we should auto-download stuff, if something (cmake?) determines it is missing Eigen just because it will not find it? Especially if there is Easyblock for Eigen present, and you can override downloding settings from easyconfig - so if you know what you are doing, you can let lammps download eigen automatically.

easybuild/easyblocks/l/lammps.py
```diff
@@ -0,0 +1,353 @@
...
+        # MPI/OpenMP
+        if self.toolchain.options.get('usempi', None):
+            self.cfg.update('configopts', '-DBUILD_MPI=yes')
+        if self.toolchain.options.get('openmp', None):
+            self.cfg.update('configopts', '-DBUILD_OMP=yes')
+            self.cfg.update('configopts', '-DPKG_USER-OMP=on')
+
+        # FFT
+        if '-DFFT=' not in self.cfg['configopts']:
+            self.cfg.update('configopts', '-DFFT=FFTW3')
```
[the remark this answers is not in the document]
Darkless012: nice catch, thanks :)

easybuild/easyblocks/l/lammps.py
```diff
@@ -0,0 +1,353 @@
...
+            pyshortver = '.'.join(get_software_version('Python').split('.')[:2])
+            pythonpath = os.path.join('lib', 'python%s' % pyshortver, 'site-packages')
+            txt += self.module_generator.prepend_paths('PYTHONPATH', [pythonpath])
+
+        txt += self.module_generator.prepend_paths('PYTHONPATH', ["lib64"])
+        txt += self.module_generator.prepend_paths('LD_LIBRARY_PATH', ["lib64"])
+
+        return txt
+
+    def get_cuda_gpu_arch(self, cuda_cc):
```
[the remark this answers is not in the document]
Darkless012: Those are just helper functions and I think having them out of sight (at the bottom) is better, because I expect overridden steps (configure, build, ...) in easyblock the most. So for me it made sense to define helper functions at bottom, and when you encounter the function, you can always search for its definition.

easybuild/easyblocks/l/lammps.py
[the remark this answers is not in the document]
akesandgren: Or possibly LAMMPS computational package options, but that's a bit long.

easybuild/easyblocks/l/lammps.py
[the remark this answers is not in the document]
akesandgren: That part of the code should complaint loudly if Eigen is specified as a dependency in the easyconfig and the DOWNLOAD_EIGEN option is also specified.

easybuild/easyblocks/l/lammps.py
```diff
@@ -0,0 +1,353 @@
...
+            elif cpu_arch == "knights-landing":
+                processor_arch = 'KNL'
+            else:
+                error_msg = "Couldn't determine CPU architecture, you need to set 'kokkos_arch' manually."
+                raise EasyBuildError(error_msg)
+                exit(1)
+            print("Determined cpu arch: %s" % processor_arch)
+
+        if not cuda:
+            return processor_arch
```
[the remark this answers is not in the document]
Darkless012: My optimization habit I guess. Will change that to if/else.

easybuild/easyblocks/l/lammps.py
Darkless012: oh, didn't know that :) thanks

easybuild/easyblocks/l/lammps.py
```diff
@@ -0,0 +1,353 @@
...
+
+        elif cuda_cc:
+            warning_msg = "Missing CUDA package (in dependencies), "
+            warning_msg += "but 'cuda_compute_capabilities' option was specified."
+            print_warning(warning_msg)
+
+        return cuda_cc
+
+    def get_cpu_arch(self):
+        out, ec = run_cmd("python -c 'from archspec.cpu import host; print(host())'", simple=False)
```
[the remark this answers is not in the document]
Darkless012: Well archspec needs to be included as builddep. Maybe we should include archspec dependency in the init of easyblock or check for the root?
The comment would be "Toolchain independent mean to obtain CPU architecture - needs to be called as external script" ?

easybuild/easyblocks/l/lammps.py
```diff
@@ -0,0 +1,359 @@
...
+            # see https://developer.nvidia.com/cuda-gpus
+            'cuda_compute_capabilities': [[], "List of CUDA compute capabilities to build with", CUSTOM],
+            'general_packages': [None, "List of general packages without `PKG_` prefix.", MANDATORY],
+            'kokkos': [True, "Enable kokkos build.", CUSTOM],
+            'kokkos_arch': [None, "Set kokkos processor arch manually, if auto-detection doesn't work.", CUSTOM],
+            'user_packages': [None, "List user packages without `PKG_USER-` prefix.", MANDATORY],
+        }
+        return CMakeMake.extra_options(extra_vars)
+
+    def prepare_step(self):
```
boegel: @Darkless012 You'll need to change this to `def prepare_step(self, *args, **kwargs):` to pass the test suite...

easybuild/easyblocks/l/lammps.py
```diff
@@ -0,0 +1,359 @@
...
+        super(EB_LAMMPS, self).prepare_step()
```
boegel: pass `*args` and `**kwargs` down here too:

```python
super(EB_LAMMPS, self).prepare_step(*args, **kwargs)

File: easybuild/easyblocks/l/lammps.py [hunk 15]
```diff
@@ -0,0 +1,359 @@
...
+
+        # Unset LIBS when using both KOKKOS and CUDA - it will mix lib paths otherwise
+        if self.cfg['kokkos'] and get_software_root('CUDA'):
+            run_cmd("unset LIBS")
```
boegel: This won't work, you need to use `env.unset_env_vars['LIBS']`.

The commands specified in `run_cmd` are run in a temporary subshell, so you're not actually unsetting `$LIBS` at all here...

easybuild/easyblocks/l/lammps.py
```diff
@@ -0,0 +1,360 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with EasyBuild.  If not, see <http://www.gnu.org/licenses/>.
+##
+"""
+@author: Pavel Grochal (INUITS)
+@author: Kenneth Hoste (Ghent University)
+@author: Alan O'Cais
```
boegel: `(Jülich Supercomputing Centre)`

easybuild/easyblocks/l/lammps.py
```diff
@@ -0,0 +1,360 @@
...
+
+Based on work of Alan O'Cais (https://github.com/ocaisa)
+https://github.com/easybuilders/easybuild-easyconfigs/blob/0c7fa07b9b7a855df6d14b971bd9eb1a25f51dd8/easybuild/easyconfigs/l/LAMMPS/LAMMPS-24Oct2018-intel-2018b.eb
```
boegel: I would drop this part, Alan is already as an author?

easybuild/easyblocks/l/lammps.py
```diff
@@ -0,0 +1,360 @@
...
+    "3.5": "Kepler35",  # NVIDIA Kepler generation CC 3.5
+    "3.7": "Kepler37",  # NVIDIA Kepler generation CC 3.7
+    "5.0": "Maxwell50",  # NVIDIA Maxwell generation CC 5.0
+    "5.2": "Maxwell52",  # NVIDIA Maxwell generation CC 5.2
+    "5.3": "Maxwell53",  # NVIDIA Maxwell generation CC 5.3
+    "6.0": "Pascal60",  # NVIDIA Pascal generation CC 6.0
+    "6.1": "Pascal61",  # NVIDIA Pascal generation CC 6.1
+    "7.0": "Volta70",  # NVIDIA Volta generation CC 7.0
+    "7.2": "Volta72",  # NVIDIA Volta generation CC 7.2
+    "7.5": "Turing75",  # NVIDIA Turing generation CC 7.5
```
boegel: nitpicking, but I prefer single quotes for string values that don't have spaces (more readable imho)

easybuild/easyblocks/l/lammps.py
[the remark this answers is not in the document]
boegel: OK, I don't care strongly about it, it was just a suggestion.

We usually keep methods "in order".

Maybe you can move these outside of the `class` though, and make them regular functions where you just pass down the stuff they need (like `self.cfg['kokkos_arch']`)

easybuild/easyblocks/l/lammps.py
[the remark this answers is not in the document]
boegel: The comment should mention that direct import doesn't work, since `archspec` is expected to be included as a build dep (so it's only available after `prepare_step` has run).

boegel:
Tested with both `intel/2019b` (see https://github.com/easybuilders/easybuild-easyconfigs/pull/9883) and `foss/2019b` (see https://github.com/easybuilders/easybuild-easyconfigs/pull/9884), so good to go...

Thanks for all the hard work on this one @Darkless012!

PR #1966, opened 2020-02-18
make sure lib/python*/lib-dynload exists in Python installation

### LINKED ISSUE #1957: Python fails to install in OpenSuse 15.1
Python fails to install with this error

```
== 2020-02-12 15:59:50,754 build_log.py:169 ERROR EasyBuild crashed with an error (at easybuild/opensuse-leap/15.1/x86_64/software/EasyBuild/4.1.1/lib/python2.7/site-packages/easybuild/base/exceptions.py:124 in __init__): cmd " ulimit -s unlimited &&   /shared/earth/easybuild/opensuse-leap/15.1/x86_64/software/Python/2.7.16-GCCcore-8.3.0/bin/python setup.py build " exited with exit code 1 and output:
Could not find platform dependent libraries <exec_prefix>
Consider setting $PYTHONHOME to <prefix>[:<exec_prefix>]
  File "setup.py", line 9, in <module>
    import setuptools
  File "/tmp/eb-build/Python/2.7.16/GCCcore-8.3.0/setuptools/setuptools-41.2.0/setuptools/__init__.py", line 5, in <module>
    import functools
  File "/shared/earth/easybuild/opensuse-leap/15.1/x86_64/software/Python/2.7.16-GCCcore-8.3.0/lib/python2.7/functools.py", line 10, in <module>
    from _functools import partial, reduce
ImportError: No module named _functools
 (at easybuild/opensuse-leap/15.1/x86_64/software/EasyBuild/4.1.1/lib/python2.7/site-packages/easybuild/tools/run.py:529 in parse_cmd_output)
```
The error is related to this [bug](https://bugs.python.org/issue34058).

Workaround from @boegel and @omula 
```
installopts = " && cd %(installdir)s && ln -s ../../lib64/python2.7/lib-dynload lib/python2.7/lib-dynload"
```

boegel:
see #1957

@kserradell Are you up for testing this fix on OpenSUSE?

### DIFF
--- a/easybuild/easyblocks/p/python.py
+++ b/easybuild/easyblocks/p/python.py
@@ -279,6 +279,21 @@ class EB_Python(ConfigureMake):
         if self.cfg['ebpythonprefixes']:
             write_file(os.path.join(self.installdir, self.pythonpath, 'sitecustomize.py'), SITECUSTOMIZE)
 
+        # symlink lib/python*/lib-dynload to lib64/python*/lib-dynload if it doesn't exist;
+        # see https://github.com/easybuilders/easybuild-easyblocks/issues/1957
+        lib_dynload = 'lib-dynload'
+        lib_dynload_path = os.path.join(self.installdir, 'lib', 'python%s' % self.pyshortver, lib_dynload)
+        if not os.path.exists(lib_dynload_path):
+            lib64_dynload_path = os.path.join(self.installdir, 'lib64', 'python%s' % self.pyshortver, lib_dynload)
+            if os.path.exists(lib64_dynload_path):
+                lib_dynload_parent = os.path.dirname(lib_dynload_path)
+                mkdir(lib_dynload_parent, parents=True)
+                cwd = change_dir(lib_dynload_parent)
+                # use relative path as target, to avoid hardcoding path to install directory
+                target_lib_dynload = os.path.join('..', '..', 'lib64', 'python%s' % self.pyshortver, lib_dynload)
+                symlink(target_lib_dynload, lib_dynload)
+                change_dir(cwd)
+
     def sanity_check_step(self):
         """Custom sanity check for Python."""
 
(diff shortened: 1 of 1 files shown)


boegel:
@akesandgren This is good to go imho, the remaining problem in #1957 is something else entirely it seems...

easybuild/easyblocks/p/python.py
```diff
@@ -279,6 +279,21 @@ def install_step(self):
...
+        lib_dynload = 'lib-dynload'
+        lib_dynload_path = os.path.join(self.installdir, 'lib', 'python%s' % self.pyshortver, lib_dynload)
+        if not os.path.exists(lib_dynload_path):
+            lib64_dynload_path = os.path.join(self.installdir, 'lib64', 'python%s' % self.pyshortver, lib_dynload)
+            if os.path.exists(lib64_dynload_path):
+                lib_dynload_parent = os.path.dirname(lib_dynload_path)
+                mkdir(lib_dynload_parent, parents=True)
+                cwd = change_dir(lib_dynload_parent)
+                # use relative path as target, to avoid hardcoding path to install directory
+                target_lib_dynload = os.path.join('..', '..', 'lib64', 'python%s' % self.pyshortver, lib_dynload)
```
akesandgren: In line with your usual comments for code that repeats itself...
put os.path.join('lib64', 'python%s' % self.pyshortver, lib_dynload))
into some variable and reuse in both places
or even the ('python%s' % self.pyshortver, lib_dynload) part...

easybuild/easyblocks/p/python.py
```diff
@@ -279,6 +279,21 @@ def install_step(self):
...
+                symlink(target_lib_dynload, lib_dynload)
```
akesandgren: Doesn't this need to be symlink(target_lib_dynload, lib_dynload, use_abspath_source=False) to make sure it doesn't get fiddled with...

boegel: No, we want to avoid hardcoding the path to the installation directory in the symlink (we had a good reason to use `../..` rather than using the full path even)

easybuild/easyblocks/p/python.py
boegel: fixed in b73b6ed47

akesandgren:
Going in, thanks @boegel!

PR #1969, opened 2020-02-20
add custom easyblock for FreeFEM

boegel:
This is an updated version of #1488.


smoors:
@boegel is there a difference in performance between setting 
OPENBLAS_NUM_THREAD or OMP_NUM_THREADS to 1?

easybuild/easyblocks/f/freefem.py
```diff
@@ -0,0 +1,165 @@
...
+            # avoid using too many OpenMP threads, which may slow tests down too much,
+            # since this could lead to failing tests (60 sec. time limit is imposed by the test run script)
+            parallel = self.cfg['parallel']
+            if parallel:
+                try:
+                    par = int(parallel)
+                except ValueError as err:
+                    raise EasyBuildError("Failed to parse 'parallel' value '%s' as integer: %s", parallel, err)
+
+                # use number of threads equal to 1/4th of available cores (and ensure at least one)
```
smoors: @boegel this is quite conservative?

easybuild/easyblocks/f/freefem.py
```diff
@@ -0,0 +1,165 @@
...
+            # reconfigure for FreeFEM build
+            super(EB_FreeFEM, self).configure_step()
+
+    def test_step(self):
+        """Run tests."""
+
+        # run tests unless they're disabled explicitly
+        if self.cfg['runtest'] is None or self.cfg['runtest']:
+            # avoid oversubscribing, by using both OpenMP threads and having OpenBLAS use threads as well
+            env.setvar('OPENBLAS_NUM_THREAD', '1')
```
smoors: @boegel I think this should be OPENBLAS_NUM_THREADS ? (added 'S')

smoors:
@boegel also, according to the OpenBLAS documentation https://github.com/xianyi/OpenBLAS
'OpenBLAS ignores OPENBLAS_NUM_THREADS and GOTO_NUM_THREADS when compiled with USE_OPENMP=1.'
so, IIUC, this means that setting OPENBLAS_NUM_THREADS does not do anything?

akesandgren:
Precisely, one should use OMP_NUM_THREAD and nothing else.

easybuild/easyblocks/f/freefem.py
```diff
@@ -0,0 +1,165 @@
...
+
+        configopts = [
+            '--disable-optim',  # disable custom optimizations (not needed, $CFLAGS set by EasyBuild is picked up)
+            '--enable-download',  # enable downloading of dependencies
+            '--disable-openblas',  # do not download OpenBLAS
+        ]
+
+        blas_family = self.toolchain.blas_family()
+        if blas_family == toolchain.OPENBLAS:
+            configopts.append("--with-blas=openblas")
```
lexming: It seems that `--with-blas` expects a path to the libs
```suggestion
            openblas_root = get_software_root('OpenBLAS')                           
            configopts.append('--with-blas=%s' % os.path.join(openblas_root, 'lib'))
```

easybuild/easyblocks/f/freefem.py
```diff
@@ -0,0 +1,165 @@
...
+        elif blas_family == toolchain.INTELMKL:
+            mkl_root = get_software_root('imkl')
+            configopts.append("--with-mkl=%s" % os.path.join(mkl_root, 'mkl', 'lib', 'intel64'))
+
```
lexming: The `configure` script has a specific option for `OpenMPI`. We could also add an option to disable linking to MPI on toolchains without MPI if `usempi` is not set to `true`
```
        if self.toolchain.mpi_family() in toolchain.OPENMPI:   
            self.cfg.update('configopts', '--with-mpi=openmpi')
        elif self.toolchain.mpi_family() in toolchain.INTELMPI:
            self.cfg.update('configopts', '--with-mpi=mpic++') 
        elif self.toolchain.options.get('usempi', None):       
            self.cfg.update('configopts', '--with-mpi=yes')    
        else:                                                  
            self.cfg.update('configopts', '--with-mpi=no')     
```

easybuild/easyblocks/f/freefem.py
```diff
@@ -0,0 +1,165 @@
...
+                    raise EasyBuildError("Failed to parse 'parallel' value '%s' as integer: %s", parallel, err)
+
+                # use number of threads equal to 1/4th of available cores (and ensure at least one)
+                n_thr = max(par // 4, 1)
+            else:
+                n_thr = 1
+
+            env.setvar('OMP_NUM_THREADS', '%d' % n_thr)
+
+            run_cmd("make check", log_all=True, simple=False)
```
lexming: I would make the installation fail if any of the tests fail or error out
```suggestion
            (out, _) = run_cmd("make check", log_all=True, simple=False, regexp=False)
                                                                                      
            # Raise an error if any test failed                                       
            check_log_for_errors(out, [('# (ERROR|FAIL): +[1-9]', ERROR)])            
```

easybuild/easyblocks/f/freefem.py
```diff
@@ -0,0 +1,165 @@
...
+
+    def sanity_check_step(self):
+        # define FreeFem++ sanity_check_paths here
+
+        custom_paths = {
+            'files': ['bin/%s' % x for x in ['bamg', 'cvmsh2', 'ffglut', 'ffmedit']] +
```
lexming: There is an additional binary installed
```suggestion
            'files': ['bin/%s' % x for x in ['bamg', 'cvmsh2', 'ffglut', 'ffmaster', 'ffmedit']] +
```

easybuild/easyblocks/f/freefem.py
akesandgren: Yes, needs the 'S'

akesandgren: And it should be OMP_NUM_THREADS since we build OpenBLAS the way we do.

boegel:
@lexming Suggestions (finally!) tackled, can you take another look?

lexming:
Positive test in https://gist.github.com/eaee20c537c572ba0cdfa1a542bfd8fb

lexming:
Going in, thanks for the update @boegel !

PR #1973, opened 2020-02-25
add NCDF_PARALLEL and METIS support to Siesta easyblock

migueldiascosta:
and skip check for broken utils in 4.1-MaX-1.0 release

(created using `eb --new-pr`)


boegel:
@migueldiascosta Can you clarify what `NCDF_PARALLEL` does?

migueldiascosta:
@boegel it enables parallel IO support (which is still off by default, needs to be explicitly enabled in the input file, so there doesn't seem to be a need to make it an easyblock option)

I regtested all existing Siesta easyconfigs and manually run one of the tests explicitly setting parallel IO in the input, it seems to work fine

Also tested adding METIS as a dependency to all Siesta easyconfigs and manually tested one of the Order-N tests, where METIS ordering would be used, again no problems

But we can wait for feedback from others (e.g. @mstud, who asked for this in the mailing list?)

PR #1977, opened 2020-03-02
don't insist that pylibdir always exists in OpenCV easyblock

pm67nz:
Python-less builds of OpenCV fail their sanity check if the easyblock requires that self.pylibdir exists, and the "python -c 'import cv2'" command is a sufficient Python test.


boegel:
makes sense, thanks for the contribution @pm67nz!

PR #1978, opened 2020-03-02
add cascadelake to KOKKOS_CPU_MAPPING in LAMMPS easyblock

boegel:
fix for `Couldn't determine CPU architecture, you need to set 'kokkos_arch' manually.` when building LAMMPS on an Intel Cascade Lake system (e.g. Intel Xeon Gold 6240), avoid having to add `kokkos_arch = 'SKX'` in the easyconfig file...


ocaisa:
Hound is asleep...

PR #1979, opened 2020-03-02
also set TRINITY_HOME environment variable in Trinity easyblock

lexming:
Many scripts in `Trinity` rely on having the environmental variable `TRINITY_HOME` pointing to the root of the installation directory.


lexming:
Re-triggering tests

lexming:
Re-triggering tests

PR #1980, opened 2020-03-02
add support for 'use_pip_extras' custom easyconfig parameter in PythonPackage easyblock

boegel:
required for https://github.com/nanoporetech/tombo, which should be installed with `pip install .[full]` in order to be fully featured?


akesandgren:
Going in, thanks @boegel!

PR #1981, opened 2020-03-03
fix sanity check for OpenFOAM-Extend >= 4.1

boegel:
binaries and libraries for `OpenFOAM-Extend` 4.1 are now also installed into `applications/bin/linux64GccDPOpt` rather than `applications/bin/linux64GccDPInt32Opt`, and `lib/linux64GccDPInt32Opt` rather than `lib/linux64GccDPOpt` (like was already the case for `OpenFOAM` >= 3.0)


akesandgren:
Going in, thanks @boegel!

PR #1983, opened 2020-03-04
libdrm easyblock for custom sanity check on different architectures

edmondac:
This is for https://github.com/easybuilders/easybuild-easyconfigs/pull/9694


edmondac:
@boegel please see successful test reports on https://github.com/easybuilders/easybuild-easyconfigs/pull/9694 for this easyblock :-)

PR #1987, opened 2020-03-09
gromacs.py: Add support for building gmxapi.

akesandgren:
gmxapi requires Python and shared libs.

### DIFF
--- a/easybuild/easyblocks/g/gromacs.py
+++ b/easybuild/easyblocks/g/gromacs.py
@@ -72,6 +72,7 @@ class EB_GROMACS(CMakeMake):
         super(EB_GROMACS, self).__init__(*args, **kwargs)
         self.lib_subdir = ''
         self.pre_env = ''
+        self.dynamic = self.toolchain.options.get('dynamic', False)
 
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/g/gromacs.py
```diff
@@ -427,6 +441,13 @@ def install_step(self):
 
                 self.log.info("A full regression test suite is available from the GROMACS web site")
 
+        # Reset installopts for the benefit of the gmxapi extension
+        self.cfg['installopts'] = self.save_installopts
+
+        # Set runtest to None so that the gmxapi extension doesn't try to
+        # run "check" as a command
+        self.cfg['runtest'] = None
```
bartoldeman: Does this apply now even when gmxapi is not build?

akesandgren: Not sure what you really mean here, but resetting it to None doesn't hur.
And this code will be moved around a bit in the new gromacs easyblock I've almost finsihed.

akesandgren:
Superseded by #1991

PR #1988, opened 2020-03-12
Added SEPP easyblock

Darkless012:


easybuild/easyblocks/s/sepp.py
```diff
@@ -0,0 +1,69 @@
...
+
+        # Configure sepp
+        run_cmd("python setup.py config -c")
+
+    def install_step(self, *args, **kwargs):
+        super(EB_SEPP, self).install_step(*args, **kwargs)
+
+        # Python version
+        pyver = ".".join(get_software_version('Python').split(".")[:-1])
+        python_site_packages_dir = os.path.join(self.installdir, "lib", "python%s" % pyver, "site-packages")
```
boegel: You can just use `self.pylibdir` (which is set by `PythonPackage.configure_step`)

easybuild/easyblocks/s/sepp.py
```diff
@@ -0,0 +1,69 @@
...
+
+        # Old (wrong) SEPP paths
+        sepp_wrong_config_dir = os.path.join(self.builddir, self.name.lower()+"-"+self.version, '.sepp')
```
boegel: spaces around `+` please (I'm surprised Hound doesn't complain about that)

easybuild/easyblocks/s/sepp.py
```diff
@@ -0,0 +1,69 @@
...
+
+        # New (correct) SEPP paths
+        sepp_correct_home_path_file = os.path.join(python_site_packages_dir, "home.path")
```
boegel: please try to use `'` consistently for string values that don't have spaces (why `"` here and `'` below)

easybuild/easyblocks/s/sepp.py
```diff
@@ -0,0 +1,69 @@
...
+        sepp_correct_config_dir = os.path.join(python_site_packages_dir, '.sepp')
+        sepp_correct_config_file = os.path.join(sepp_correct_config_dir, 'main.config')
+
+        # Create correct home.path file
+        self.log.debug("Creating home.path file for SEPP at %s", sepp_correct_home_path_file)
```
boegel: `log.info` makes more sense here I think

Darkless012:
Wrong develop, closing. Will reopen with correct one.

easybuild/easyblocks/s/sepp.py
```diff
@@ -0,0 +1,69 @@
...
+        write_file(sepp_correct_home_path_file, sepp_correct_config_dir)
+
+        # Copy .sepp folder with configurations and bundled stuff from builddir to installdir
+        copy_dir(sepp_wrong_config_dir, sepp_correct_config_dir)
+
+        # Replace wrong paths in SEPP configfile with correct ones
+        run_cmd("sed -i 's#%s#%s#g' %s" % (sepp_wrong_config_dir, sepp_correct_config_dir, sepp_correct_config_file))
```
boegel: we have the `apply_regex_substitutions` function for runtime patching, please don't rely on `sed` for this

boegel:
replaced with #1998

PR #1989, opened 2020-03-12
tarball: add support for preinstallopts and install in subdirectory

lexming:
Update for tarball easyblock to make it (1) more flexible and (2) usable as a component in a bundle:
* Currently, `tarball` deletes and re-creates `installdir` on each installation, removing any installation from previous components. This PR adds an extra variable `final_dir` that is similar to `start_dir` and can be used to target sub-directories of `installdir`, preserving installations from previous components. The default behavior of `tarball` will be unaffected.
* Adds support for using `preinstallopts` with `tarball`. Any commands given through `preinstallopts` will be executed before the copy of files to install directory.

### DIFF
--- a/easybuild/easyblocks/generic/tarball.py
+++ b/easybuild/easyblocks/generic/tarball.py
@@ -53,7 +53,7 @@ class Tarball(EasyBlock):
         """Extra easyconfig parameters specific to Tarball."""
         extra_vars = EasyBlock.extra_options(extra=extra_vars)
         extra_vars.update({
-            'final_dir': [None, "Override default installation path with provided path", CUSTOM],
+            'install_in_subdir': [False, "Extract tarball in a subdirectory with the name of the package", CUSTOM],
         })
         return extra_vars
 
@@ -86,8 +86,8 @@ class Tarball(EasyBlock):
         # Copy source directory
         source_path = src or self.cfg['start_dir']
 
-        if 'final_dir' in self.cfg and self.cfg['final_dir']:
-            install_path = os.path.join(self.installdir, self.cfg['final_dir'])
+        if self.cfg['install_in_subdir']:
+            install_path = os.path.join(self.installdir, self.name.lower())
         else:
             install_path = self.installdir
 
--- a/easybuild/easyblocks/generic/tarball.py
+++ b/easybuild/easyblocks/generic/tarball.py
@@ -54,6 +54,7 @@ class Tarball(EasyBlock):
         extra_vars = EasyBlock.extra_options(extra=extra_vars)
         extra_vars.update({
             'install_in_subdir': [False, "Extract tarball in a subdirectory with the name of the package", CUSTOM],
+            'preinstall_cmd': [None, "Command to execute before installation", CUSTOM],
         })
         return extra_vars
 
@@ -76,10 +77,14 @@ class Tarball(EasyBlock):
     def install_step(self, src=None):
         """Install by copying from specified source directory (or 'start_dir' if not specified)."""
 
-        # Run preinstallopts before copy of source directory
+        # Run preinstallopts and/or preinstall_cmd before copy of source directory
+        preinstall_cmd = None
         if self.cfg['preinstallopts']:
             preinstall_opts = self.cfg['preinstallopts'].split('&&')
             preinstall_cmd = '&&'.join([opt for opt in preinstall_opts if opt and not opt.isspace()])
+        if self.cfg['preinstall_cmd']:
+            preinstall_cmd = '&& '.join([cmd for cmd in [preinstall_cmd, self.cfg['preinstall_cmd']] if cmd])
+        if preinstall_cmd:
             self.log.info("Preparing installation of %s using command '%s'..." % (self.name, preinstall_cmd))
             run_cmd(preinstall_cmd, log_all=True, simple=True)
 
--- a/easybuild/easyblocks/generic/tarball.py
+++ b/easybuild/easyblocks/generic/tarball.py
@@ -34,6 +34,8 @@ implemented as an easyblock
 @author: Alex Domingo (Vrije Universiteit Brussel)
 """
 
+import os
+
 from easybuild.framework.easyblock import EasyBlock
 from easybuild.framework.easyconfig import CUSTOM
 from easybuild.tools.filetools import copy_dir, remove_dir
@@ -83,7 +85,12 @@ class Tarball(EasyBlock):
 
         # Copy source directory
         source_path = src or self.cfg['start_dir']
-        install_path = self.cfg['final_dir'] or self.installdir
+
+        if 'final_dir' in self.cfg and self.cfg['final_dir']:
+            install_path = os.path.join(self.installdir, self.cfg['final_dir'])
+        else:
+            install_path = self.installdir
+
         self.log.info("Copying tarball contents of %s to %s..." % (self.name, install_path))
         remove_dir(install_path)
         copy_dir(source_path, install_path, symlinks=self.cfg['keepsymlinks'])
--- a/easybuild/easyblocks/generic/tarball.py
+++ b/easybuild/easyblocks/generic/tarball.py
@@ -32,6 +32,7 @@ implemented as an easyblock
 @author: Pieter De Baets (Ghent University)
 @author: Jens Timmerman (Ghent University)
 @author: Alex Domingo (Vrije Universiteit Brussel)
+@author: Pavel Grochal (INUITS)
 """
 
 import os
--- a/easybuild/easyblocks/generic/tarball.py
+++ b/easybuild/easyblocks/generic/tarball.py
@@ -31,10 +31,13 @@ implemented as an easyblock
 @author: Kenneth Hoste (Ghent University)
 @author: Pieter De Baets (Ghent University)
 @author: Jens Timmerman (Ghent University)
+@author: Alex Domingo (Vrije Universiteit Brussel)
 """
 
 from easybuild.framework.easyblock import EasyBlock
+from easybuild.framework.easyconfig import CUSTOM
 from easybuild.tools.filetools import copy_dir, remove_dir
+from easybuild.tools.run import run_cmd
 
 
 class Tarball(EasyBlock):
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/generic/tarball.py
```diff
@@ -57,8 +75,25 @@ def build_step(self):
 
     def install_step(self, src=None):
         """Install by copying from specified source directory (or 'start_dir' if not specified)."""
-        remove_dir(self.installdir)
-        copy_dir(src or self.cfg['start_dir'], self.installdir, symlinks=self.cfg['keepsymlinks'])
+
+        # Run preinstallopts before copy of source directory
+        if self.cfg['preinstallopts']:
+            preinstall_opts = self.cfg['preinstallopts'].split('&&')
```
ocaisa, line 81: While this usage of `preinstallopts` is quite common, what happens here is reliant on the user knowing that.  As well as supporting this usage of `preinstallopts`, maybe we can primarily support an additional explicit parameter `preinstall_cmd` which we can explicitly document and attach this parsing of `preinstallopts` to it, that way you can support both cases. Something like
```
preinstall_cmd = None
if self.cfg['preinstallopts']:
    ...
    preinstall_cmd = '&&'.join([opt for opt in preinstall_opts if opt and not opt.isspace()])
if self.cfg['preinstall_cmd']:
    preinstall_cmd = ' '.join([preinstall_cmd, self.cfg['preinstall_cmd']])
if preinstall_cmd:
    self.log.info("Preparing installation of %s using command '%s'..." % (self.name, preinstall_cmd))
    run_cmd(preinstall_cmd, log_all=True, simple=True)
```

easybuild/easyblocks/generic/tarball.py
```diff
@@ -57,8 +75,25 @@ def build_step(self):
...
+            preinstall_cmd = '&&'.join([opt for opt in preinstall_opts if opt and not opt.isspace()])
+            self.log.info("Preparing installation of %s using command '%s'..." % (self.name, preinstall_cmd))
+            run_cmd(preinstall_cmd, log_all=True, simple=True)
+
+        # Copy source directory
+        source_path = src or self.cfg['start_dir']
+
+        if 'final_dir' in self.cfg and self.cfg['final_dir']:
```
ocaisa, line 89: I wonder if we need to be giving the user the power to name the installation subdirectory (and if we do we should check they're trying to pass an absolute path)? Instead, we could create an `install_in_subdir` boolean and use the name of the software as the subdir, this would lead to more predictable results (that we would be able to leverage somewhere else without parsing the associated easyconfig).

lexming: I agree :+1:  it is much safer to set the name of the subdir to the software name and it is a suitable solution for the easyconfig that triggered these changes.

easybuild/easyblocks/generic/tarball.py
lexming: I have to say that I always try first to use the different `opts` in the common steps of the build, but adding an explicit parameter is a good idea as well. I made it work in combination with `preinstallopts` as you suggested so that both options can be used at the same time.

lexming:
I tested the execution of the new `preinstall_cmd` and `preinstallopts` in the following cases and it successfully works in all of them:
* Single `preinstallopts`
```
'preinstallopts' = 'touch A && touch B && '
```
* Single `preinstall_cmd`
```
'preinstall_cmd' = 'touch C && touch D'
```
* Both `preinstallopts` and `preinstall_cmd`
```
'preinstallopts' = 'touch A && touch B && '
'preinstall_cmd' = 'touch C && touch D'
```

ocaisa:
@lexing Coincidentally there is a PR https://github.com/easybuilders/easybuild-framework/pull/3270 was opened  today in framework that intends to cover (some of) the same ground as this, can you check if that approach would suit your needs?

lexming:
@ocaisa that PR merges the folders instead of installing them in different subfolders. That will not work for bundles, at least not in a general sense, as the source code of different projects will probably have collisions.

lexming:
@ocaisa I think that both options can complement each other and  could be added to `tarball`. Maybe a variable called `install_type` that would accept `merge` or `subdir` as options. I'll wait to see what @Darkless012 has done.

Darkless012:
I might be repeating, but let me summarize what is needed by this change:

1. Don't remove install dir in tarball - on which we all agree, because this should be done by easyconfig and brings trouble to Bundles.
2. Allow changing (in some way) installdir so that conflicts will not arise.

Do you have any specific use-case other than using this in Bundle? 

I needed step 1 Because I had MakeCP and Tarball. MakeCP would create `installdir/bin/executable` and then Tarball should copy its content (into installdir which includes bin folder) as well. Which is my need in merging folders instead of remove/copy. So there should be extra param "merge folder" which should not be default as you mentioned.

I didn't thought about having multiple Tarballs in bundle, which is valid point. I like the idea of having install_in_subdir.
Just want point out that if you are already installing it into bundle, you might want to specify exact path within the installdir? Which is basically what I'm doing with MakeCp (files_to_copy).
Sure you can do `mv/ln -s` in postinstallcmds, where the install_in_subdir makes sence, because you will have always same name structure to copy from?

Does install_in_subdir has any other use besides Bundle? 
"merge_folder" probably doesn't have since installdir should be wiped in solo Tarball installation.

easybuild/easyblocks/generic/tarball.py
```diff
@@ -43,6 +48,20 @@ class Tarball(EasyBlock):
...
     """
 
+    @staticmethod
+    def extra_options(extra_vars=None):
+        """Extra easyconfig parameters specific to Tarball."""
+        extra_vars = EasyBlock.extra_options(extra=extra_vars)
+        extra_vars.update({
+            'install_in_subdir': [False, "Extract tarball in a subdirectory with the name of the package", CUSTOM],
+            'preinstall_cmd': [None, "Command to execute before installation", CUSTOM],
+        })
```
Darkless012, line 58: 'dirs_exist_ok': [False, "Merge into existing folder instead of throwing Exception if installdir exists.", CUSTOM],

easybuild/easyblocks/generic/tarball.py
```diff
@@ -57,8 +76,29 @@ def build_step(self):
...
+        source_path = src or self.cfg['start_dir']
+
+        if self.cfg['install_in_subdir']:
+            install_path = os.path.join(self.installdir, self.name.lower())
+        else:
+            install_path = self.installdir
+
+        self.log.info("Copying tarball contents of %s to %s..." % (self.name, install_path))
+        remove_dir(install_path)
+        copy_dir(source_path, install_path, symlinks=self.cfg['keepsymlinks'])
```
Darkless012, line 102: Should we remove dir? shouldn't easyblock wipe whole installdir in the begining?

copy_dir(source_path, install_path, symlinks=self.cfg['keepsymlinks'], dirs_exist_ok=self.cfg['dirs_exist_ok'])

ocaisa: Removing the directory is still required I think if this easyblock is used outside a bundle, but it would now need to be done conditionally, something like 
```
dirs_exist_ok = self.cfg['dirs_exist_ok']
if not dirs_exist_ok:
    remove_dir(install_path)
copy_dir(source_path, install_path, symlinks=self.cfg['keepsymlinks'], dirs_exist_ok=dirs_exist_ok)
```

lexming:
List of use cases and required actions needed for each of them:

1. Installation of a single tarball
    * Wipe existing installation directory (we do not want any unexpected behaviour caused by leftovers)
    * Unpack and copy tarball

2. Installation of a multiple tarballs, where one is the main package and the other are _addons_ that have to be decompressed in paths below the main one
    * Extract sources of main package using default settings
    * Extract sources of _addon_ tarballs in paths under main package with specific `extract_cmd` (see for instance https://github.com/easybuilders/easybuild-easyconfigs/pull/10351)
    * `tarball` easyblock will do a single copy of everything to installation directory.

3. Installation of multiple tarballs, where all of them have to be merged in a common installation directory. Useful for Bundles were `tarball` easyblock comes after other easyblocks that have already populated the installation directory.
    * Installation directory will be already wiped and populated by previous easyblock
    * Copy files to installation directory without wipe and enabling `dirs_exist_ok`

4. Installation of multiple tarballs, where each of them is independent from the others and live in their own subdirectory. Useful for Bundles with multiple `tarball` easyblocks.
    * Wipe installation sub-directory
    * Unpack and copy tarball to specific sub-directory

Cases 1 and 2 are already supported. The default  behaviour of `tarball` should not change to avoid breaking existing easyconfigs.
Support for case 3 and 4 will be added by this PR. Since these two options are exclusive, I propose to have a single option that can be set to one or the other.

lexming:
@Darkless012 @ocaisa I have added the option `install_type` to handle both installation merging the contents of the tarball or extracting in its own sub-directory. Please test it and let me know your thoughts.

I've tested it combining several `tarball` components in a single `bundle` with various combinations of options (None, merge, subdir) and all worked well on my end.

easybuild/easyblocks/generic/tarball.py
lexming: I added a single option called `install_type` covering both `install_in_subdir` and `dirs_exist_ok` as they should be exclusive.

easybuild/easyblocks/generic/tarball.py
lexming: I removed the `remove_dir()` call if `tarball` is used in merge mode.

ocaisa:
@lexming PR looks good, can you share that test case (here or in a gist) just so I repeat the tests for myself?

lexming:
@ocaisa I put the dummy easyconfig to test this PR in https://gist.github.com/lexming/3477ca1390b1965affb6338efc5f1dd7
It's just a stripped down version of my PR for Alpha (https://github.com/easybuilders/easybuild-easyconfigs/pull/9994) with the updated settings from this PR for `tarball`.
The version I'm sharing uses `subdir` and should fulfil the sanity checks. The python package has its own `lib` and `pyinclude` and `alpha` are in their own folder. You can change the settings for `install_type` to merge or None and check the different results.

PR #1991, opened 2020-03-14
rewrite GROMACS easyblock to build all four variations in the same build

akesandgren:
single precision without MPI, single precision with MPI,
double precision without MPI, double precision with MPI.

Change the default for double_precision from False to None
And a bunch of other changes.

This supersedes https://github.com/easybuilders/easybuild-easyblocks/pull/1987

### DIFF
--- a/easybuild/easyblocks/g/gromacs.py
+++ b/easybuild/easyblocks/g/gromacs.py
@@ -254,12 +254,13 @@ class EB_GROMACS(CMakeMake):
                     mpiexec = which(mpiexec)
                     if mpiexec:
                         self.cfg.update('configopts', "-DMPIEXEC=%s" % mpiexec)
-                        self.cfg.update('configopts', "-DMPIEXEC_NUMPROC_FLAG=%s" % self.cfg.get('mpiexec_numproc_flag'))
+                        self.cfg.update('configopts', "-DMPIEXEC_NUMPROC_FLAG=%s" %
+                                        self.cfg.get('mpiexec_numproc_flag'))
                         self.cfg.update('configopts', "-DNUMPROC=%s" % mpi_numprocs)
                     elif self.cfg['runtest']:
                         raise EasyBuildError("'%s' not found in $PATH", self.cfg.get('mpiexec'))
                 else:
-                        raise EasyBuildError("No value found for 'mpiexec'")
+                    raise EasyBuildError("No value found for 'mpiexec'")
                 self.log.info("Using %s as MPI executable when testing, with numprocs flag '%s' and %s tasks",
                               self.cfg.get('mpiexec'), self.cfg.get('mpiexec_numproc_flag'),
                               mpi_numprocs)
@@ -430,7 +431,8 @@ class EB_GROMACS(CMakeMake):
 
             super(EB_GROMACS, self).install_step()
 
-            # the GROMACS libraries get installed in different locations (deeper subdirectory), depending on the platform;
+            # the GROMACS libraries get installed in different locations (deeper subdirectory),
+            # depending on the platform;
             # this is determined by the GNUInstallDirs CMake module;
             # rather than trying to replicate the logic, we just figure out where the library was placed
 
--- a/easybuild/easyblocks/g/gromacs.py
+++ b/easybuild/easyblocks/g/gromacs.py
@@ -158,13 +158,15 @@ class EB_GROMACS(CMakeMake):
                 # otherwise warn about it and skip the double precision build
                 if self.cfg['double_precision']:
                     raise EasyBuildError("Double precision is not available for GPU build. " +
-                                         "Please explicitly set \"double_precision = False\" or remove it in the easyconfig file.")
+                                         "Please explicitly set \"double_precision = False\" " +
+                                         "or remove it in the easyconfig file.")
                 if re.search('DGMX_DOUBLE=(ON|YES|TRUE|Y|[1-9])', self.cfg.get('configopts'), re.I):
                     if self.cfg['double_precision'] is None:
                         # Only print warning once when trying double precision
                         # build the frst time
                         self.cfg['double_precision'] = False
-                        print_warning("Double precision is not available for GPU build. Skipping the double precision build.")
+                        print_warning("Double precision is not available for " +
+                                      "GPU build. Skipping the double precision build.")
 
                     print_msg("skipping configure step", silent=self.silent)
                     return
@@ -257,7 +259,8 @@ class EB_GROMACS(CMakeMake):
                     # check whether Python is loaded as a dependency
                     python_root = get_software_root('Python')
                     if python_root:
-                        self.cfg.update('configopts', "-DPYTHON_EXECUTABLE=%s" % os.path.join(python_root, 'bin', 'python'))
+                        self.cfg.update('configopts',
+                                        "-DPYTHON_EXECUTABLE=%s" % os.path.join(python_root, 'bin', 'python'))
                         self.cfg.update('configopts', "-DGMX_PYTHON_PACKAGE=ON")
 
             # Now patch GROMACS for PLUMED before cmake
@@ -383,7 +386,8 @@ class EB_GROMACS(CMakeMake):
 
         cuda = get_software_root('CUDA')
         if cuda:
-            if not self.cfg['double_precision'] and re.search('DGMX_DOUBLE=(ON|YES|TRUE|Y|[1-9])', self.cfg.get('configopts'), re.I):
+            if not self.cfg['double_precision'] and
+               re.search('DGMX_DOUBLE=(ON|YES|TRUE|Y|[1-9])', self.cfg.get('configopts'), re.I):
                 print_msg("skipping build step", silent=self.silent)
                 return
 
@@ -415,7 +419,8 @@ class EB_GROMACS(CMakeMake):
         # is for double precision
         cuda = get_software_root('CUDA')
         if cuda:
-            if not self.cfg['double_precision'] and re.search('DGMX_DOUBLE=(ON|YES|TRUE|Y|[1-9])', self.cfg.get('configopts'), re.I):
+            if not self.cfg['double_precision'] and
+               re.search('DGMX_DOUBLE=(ON|YES|TRUE|Y|[1-9])', self.cfg.get('configopts'), re.I):
                 print_msg("skipping install step", silent=self.silent)
                 return
 
@@ -632,7 +637,6 @@ class EB_GROMACS(CMakeMake):
             for mpitype in mpitypes:
                 self.variants_to_build += 1
                 versions_built.append('%s precision %s' % (prec, mpitype))
-                var_preconfopts = []
                 var_confopts = []
                 var_buildopts = []
                 var_installopts = []
--- a/easybuild/easyblocks/g/gromacs.py
+++ b/easybuild/easyblocks/g/gromacs.py
@@ -132,7 +132,7 @@ class EB_GROMACS(CMakeMake):
     def is_double_precision_cuda_build(self):
         """Check if the current build step involves double precision and CUDA"""
         cuda = get_software_root('CUDA')
-        return cuda and self.DP_pattern in self.cfg['configopts']
+        return cuda and self.double_prec_pattern in self.cfg['configopts']
 
     def prepare_step(self, *args, **kwargs):
         """Custom prepare step for GROMACS."""
@@ -166,7 +166,7 @@ class EB_GROMACS(CMakeMake):
                     raise EasyBuildError("Double precision is not available for GPU build. " +
                                          "Please explicitly set \"double_precision = False\" " +
                                          "or remove it in the easyconfig file.")
-                if self.DP_pattern in self.cfg['configopts']:
+                if self.double_prec_pattern in self.cfg['configopts']:
                     if self.cfg.get('double_precision') is None:
                         # Only print warning once when trying double precision
                         # build the first time
@@ -540,7 +540,7 @@ class EB_GROMACS(CMakeMake):
         if not get_software_root('CUDA'):
             for configopts in configopts_list:
                 # add the _d suffix to the suffix, in case of double precission
-                if self.DP_pattern in configopts:
+                if self.double_prec_pattern in configopts:
                     dsuff = '_d'
 
         if dsuff:
@@ -605,7 +605,7 @@ class EB_GROMACS(CMakeMake):
             }
 
         # Double precision pattern so search for in configopts
-        self.DP_pattern = prec_opts['double']
+        self.double_prec_pattern = prec_opts['double']
 
         if LooseVersion(self.version) < LooseVersion('5'):
             # For older versions we only build/install the mdrun part for
--- a/easybuild/easyblocks/g/gromacs.py
+++ b/easybuild/easyblocks/g/gromacs.py
@@ -570,16 +570,16 @@ class EB_GROMACS(CMakeMake):
         """
         # Save installopts so we can reset it later. The gmxapi pip install
         # can't handle the -j argument.
-        self.save_installopts = self.cfg.get('installopts')
+        self.save_installopts = self.cfg['installopts']
 
         # keep track of config/build/installopts specified in easyconfig
         # file, so we can include them in each iteration later
-        common_config_opts = self.cfg.get('configopts')
-        common_build_opts = self.cfg.get('buildopts')
-        common_install_opts = self.cfg.get('installopts')
+        common_config_opts = self.cfg['configopts']
+        common_build_opts = self.cfg['buildopts']
+        common_install_opts = self.cfg['installopts']
 
-        self.save_install_cmd = self.cfg.get('install_cmd')
-        self.save_build_cmd = self.cfg.get('build_cmd')
+        self.save_install_cmd = self.cfg['install_cmd']
+        self.save_build_cmd = self.cfg['build_cmd']
 
         self.cfg['configopts'] = []
         self.cfg['buildopts'] = []
--- a/easybuild/easyblocks/g/gromacs.py
+++ b/easybuild/easyblocks/g/gromacs.py
@@ -407,7 +407,7 @@ class EB_GROMACS(CMakeMake):
             if isinstance(self.cfg['runtest'], bool):
                 self.cfg['runtest'] = 'check'
 
-            # run 'make check' or whever the easyconfig specifies 
+            # run 'make check' or whever the easyconfig specifies
             # in parallel since it involves more compilation
             self.cfg.update('runtest', "-j %s" % self.cfg['parallel'])
             super(EB_GROMACS, self).test_step()
--- a/easybuild/easyblocks/g/gromacs.py
+++ b/easybuild/easyblocks/g/gromacs.py
@@ -548,11 +548,9 @@ class EB_GROMACS(CMakeMake):
             suffixes.extend([dsuff])
 
         lib_files.extend([
-            'lib%s%s.%s' % (libname, suff, self.libext) for libname in libnames + mpi_libnames for suff in suffixes
+            'lib%s%s.%s' % (x, suff, self.libext) for x in libnames + mpi_libnames for suff in suffixes
         ])
-        bin_files.extend([
-            b + suff for b in bins + mpi_bins for suff in suffixes
-        ])
+        bin_files.extend([b + suff for b in bins + mpi_bins for suff in suffixes])
 
         # pkgconfig dir not available for earlier versions, exact version to use here is unclear
         if LooseVersion(self.version) >= LooseVersion('4.6'):
--- a/easybuild/easyblocks/g/gromacs.py
+++ b/easybuild/easyblocks/g/gromacs.py
@@ -247,7 +247,7 @@ class EB_GROMACS(CMakeMake):
 
                 elif self.cfg.get('mpi_numprocs') > self.cfg['parallel']:
                     self.log.warning("Number of test MPI tasks (%s) is greater than value for 'parallel': %s",
-                                     self.cfg.get('mpi_numprocs'), self.cfg'parallel'])
+                                     self.cfg.get('mpi_numprocs'), self.cfg['parallel'])
 
                 mpiexec = which(self.cfg.get('mpiexec'))
                 if mpiexec:
--- a/easybuild/easyblocks/g/gromacs.py
+++ b/easybuild/easyblocks/g/gromacs.py
@@ -44,7 +44,7 @@ import easybuild.tools.toolchain as toolchain
 from easybuild.easyblocks.generic.configuremake import ConfigureMake
 from easybuild.easyblocks.generic.cmakemake import CMakeMake
 from easybuild.framework.easyconfig import CUSTOM
-from easybuild.tools.build_log import EasyBuildError, print_msg, print_warning
+from easybuild.tools.build_log import EasyBuildError, print_warning
 from easybuild.tools.config import build_option
 from easybuild.tools.filetools import download_file, extract_file, which
 from easybuild.tools.modules import get_software_libdir, get_software_root, get_software_version
--- a/easybuild/easyblocks/g/gromacs.py
+++ b/easybuild/easyblocks/g/gromacs.py
@@ -162,7 +162,7 @@ class EB_GROMACS(CMakeMake):
                     raise EasyBuildError("Double precision is not available for GPU build. " +
                                          "Please explicitly set \"double_precision = False\" " +
                                          "or remove it in the easyconfig file.")
-                if re.search('DGMX_DOUBLE=(ON|YES|TRUE|Y|[1-9])', self.cfg.get('configopts'), re.I):
+                if re.search('-DGMX_DOUBLE=(ON|YES|TRUE|Y|[1-9])', self.cfg.get('configopts'), re.I):
                     if self.cfg.get('double_precision') is None:
                         # Only print warning once when trying double precision
                         # build the first time
@@ -232,7 +232,7 @@ class EB_GROMACS(CMakeMake):
                 run_cmd(plumed_cmd, log_all=True, simple=True)
 
         else:
-            if re.search('DGMX_MPI=(ON|YES|TRUE|Y|[1-9])', self.cfg.get('configopts'), re.I):
+            if re.search('-DGMX_MPI=(ON|YES|TRUE|Y|[1-9])', self.cfg.get('configopts'), re.I):
                 if self.cfg.get('mpi_numprocs') == 0:
                     self.log.info("No number of test MPI tasks specified -- using default: %s" % self.cfg.get('parallel'))
                     self.cfg['mpi_numprocs'] = self.cfg.get('parallel')
@@ -383,7 +383,7 @@ class EB_GROMACS(CMakeMake):
         cuda = get_software_root('CUDA')
         if cuda:
             if (not self.cfg.get('double_precision') and
-                    re.search('DGMX_DOUBLE=(ON|YES|TRUE|Y|[1-9])', self.cfg.get('configopts'), re.I)):
+                    re.search('-DGMX_DOUBLE=(ON|YES|TRUE|Y|[1-9])', self.cfg.get('configopts'), re.I)):
                 print_msg("skipping build step", silent=self.silent)
                 return
 
@@ -416,7 +416,7 @@ class EB_GROMACS(CMakeMake):
         cuda = get_software_root('CUDA')
         if cuda:
             if (not self.cfg.get('double_precision') and
-                    re.search('DGMX_DOUBLE=(ON|YES|TRUE|Y|[1-9])', self.cfg.get('configopts'), re.I)):
+                    re.search('-DGMX_DOUBLE=(ON|YES|TRUE|Y|[1-9])', self.cfg.get('configopts'), re.I)):
                 print_msg("skipping install step", silent=self.silent)
                 return
 
@@ -532,9 +532,9 @@ class EB_GROMACS(CMakeMake):
         if self.cfg.get('double_precision') is None or self.cfg.get('double_precision'):
             for configopts in configopts_list:
                 # add the _d suffix to the suffix, in case of double precission
-                if re.search('enable-double', configopts, re.I):
+                if re.search('-enable-double', configopts, re.I):
                     suff = '_d'
-                if re.search('DGMX_DOUBLE=(ON|YES|TRUE|Y|[1-9])', configopts, re.I):
+                if re.search('-DGMX_DOUBLE=(ON|YES|TRUE|Y|[1-9])', configopts, re.I):
                     suff = '_d'
 
         lib_files.extend(['lib%s%s.%s' % (libname, suff, self.libext) for libname in libnames + mpi_libnames])
--- a/easybuild/easyblocks/g/gromacs.py
+++ b/easybuild/easyblocks/g/gromacs.py
@@ -387,7 +387,7 @@ class EB_GROMACS(CMakeMake):
         cuda = get_software_root('CUDA')
         if cuda:
             if not self.cfg['double_precision'] and
-               re.search('DGMX_DOUBLE=(ON|YES|TRUE|Y|[1-9])', self.cfg.get('configopts'), re.I):
+                re.search('DGMX_DOUBLE=(ON|YES|TRUE|Y|[1-9])', self.cfg.get('configopts'), re.I):
                 print_msg("skipping build step", silent=self.silent)
                 return
 
@@ -420,7 +420,7 @@ class EB_GROMACS(CMakeMake):
         cuda = get_software_root('CUDA')
         if cuda:
             if not self.cfg['double_precision'] and
-               re.search('DGMX_DOUBLE=(ON|YES|TRUE|Y|[1-9])', self.cfg.get('configopts'), re.I):
+                re.search('DGMX_DOUBLE=(ON|YES|TRUE|Y|[1-9])', self.cfg.get('configopts'), re.I):
                 print_msg("skipping install step", silent=self.silent)
                 return
 
--- a/easybuild/easyblocks/g/gromacs.py
+++ b/easybuild/easyblocks/g/gromacs.py
@@ -387,9 +387,9 @@ class EB_GROMACS(CMakeMake):
         cuda = get_software_root('CUDA')
         if cuda:
             if (not self.cfg['double_precision'] and
-                re.search('DGMX_DOUBLE=(ON|YES|TRUE|Y|[1-9])', self.cfg.get('configopts'), re.I)):
-                    print_msg("skipping build step", silent=self.silent)
-                    return
+                    re.search('DGMX_DOUBLE=(ON|YES|TRUE|Y|[1-9])', self.cfg.get('configopts'), re.I)):
+                print_msg("skipping build step", silent=self.silent)
+                return
 
         super(EB_GROMACS, self).build_step()
 
@@ -420,9 +420,9 @@ class EB_GROMACS(CMakeMake):
         cuda = get_software_root('CUDA')
         if cuda:
             if (not self.cfg['double_precision'] and
-                re.search('DGMX_DOUBLE=(ON|YES|TRUE|Y|[1-9])', self.cfg.get('configopts'), re.I)):
-                    print_msg("skipping install step", silent=self.silent)
-                    return
+                    re.search('DGMX_DOUBLE=(ON|YES|TRUE|Y|[1-9])', self.cfg.get('configopts'), re.I)):
+                print_msg("skipping install step", silent=self.silent)
+                return
 
         # run 'make install' in parallel since it involves more compilation
         self.cfg.update('installopts', "-j %s" % self.cfg['parallel'])
--- a/easybuild/easyblocks/g/gromacs.py
+++ b/easybuild/easyblocks/g/gromacs.py
@@ -456,7 +456,7 @@ class EB_GROMACS(CMakeMake):
                 raise EasyBuildError("Failed to determine lib subdirectory in %s", self.installdir)
 
             # Reset installopts etc for the benefit of the gmxapi extension
-            self.cfg['installopts'] = self.save_installopts
+            self.cfg['installopts'] = self.orig_installopts
 
     def extensions_step(self, fetch=False):
         """ Custom extensions step, only handle extensions after the last iteration round"""
@@ -569,7 +569,7 @@ class EB_GROMACS(CMakeMake):
         """
         # Save installopts so we can reset it later. The gmxapi pip install
         # can't handle the -j argument.
-        self.save_installopts = self.cfg['installopts']
+        self.orig_installopts = self.cfg['installopts']
 
         # keep track of config/build/installopts specified in easyconfig
         # file, so we can include them in each iteration later
@@ -577,8 +577,8 @@ class EB_GROMACS(CMakeMake):
         common_build_opts = self.cfg['buildopts']
         common_install_opts = self.cfg['installopts']
 
-        self.save_install_cmd = self.cfg['install_cmd']
-        self.save_build_cmd = self.cfg['build_cmd']
+        self.orig_install_cmd = self.cfg['install_cmd']
+        self.orig_build_cmd = self.cfg['build_cmd']
 
         self.cfg['configopts'] = []
         self.cfg['buildopts'] = []
@@ -672,7 +672,7 @@ class EB_GROMACS(CMakeMake):
         self.log.info("Building these variants of GROMACS: %s", ', '.join(versions_built))
         return super(EB_GROMACS, self).run_all_steps(*args, **kwargs)
 
-        self.cfg['install_cmd'] = self.save_install_cmd
-        self.cfg['build_cmd'] = self.save_build_cmd
+        self.cfg['install_cmd'] = self.orig_install_cmd
+        self.cfg['build_cmd'] = self.orig_build_cmd
 
         self.log.info("A full regression test suite is available from the GROMACS web site")
--- a/easybuild/easyblocks/g/gromacs.py
+++ b/easybuild/easyblocks/g/gromacs.py
@@ -669,7 +669,7 @@ class EB_GROMACS(CMakeMake):
         self.variants_to_build = len(self.cfg['configopts'])
 
         self.log.debug("List of configure options to iterate over: %s", self.cfg['configopts'])
-        self.log.info("Building these versions of GROMACS: %s", ', '.join(versions_built))
+        self.log.info("Building these variants of GROMACS: %s", ', '.join(versions_built))
         return super(EB_GROMACS, self).run_all_steps(*args, **kwargs)
 
         self.cfg['install_cmd'] = self.save_install_cmd
--- a/easybuild/easyblocks/g/gromacs.py
+++ b/easybuild/easyblocks/g/gromacs.py
@@ -129,6 +129,14 @@ class EB_GROMACS(CMakeMake):
 
         return res
 
+    def is_double_precision_cuda_build(self):
+        """Check if the current build step involves double precision and CUDA"""
+        cuda = get_software_root('CUDA')
+        if cuda:
+            if re.search(self.DP_pattern, self.cfg.get('configopts')):
+                return True
+        return False
+
     def prepare_step(self, *args, **kwargs):
         """Custom prepare step for GROMACS."""
 
@@ -161,7 +169,7 @@ class EB_GROMACS(CMakeMake):
                     raise EasyBuildError("Double precision is not available for GPU build. " +
                                          "Please explicitly set \"double_precision = False\" " +
                                          "or remove it in the easyconfig file.")
-                if re.search('-DGMX_DOUBLE=(ON|YES|TRUE|Y|[1-9])', self.cfg.get('configopts'), re.I):
+                if re.search(self.DP_pattern, self.cfg.get('configopts')):
                     if self.cfg.get('double_precision') is None:
                         # Only print warning once when trying double precision
                         # build the first time
@@ -381,16 +389,19 @@ class EB_GROMACS(CMakeMake):
         iteration is for double precision
         """
 
-        cuda = get_software_root('CUDA')
-        if cuda:
-            if re.search('-DGMX_DOUBLE=(ON|YES|TRUE|Y|[1-9])', self.cfg.get('configopts'), re.I):
-                self.log.info("skipping build step")
-                return
+        if is_double_precision_cuda_build():
+            self.log.info("skipping build step")
+            return
 
         super(EB_GROMACS, self).build_step()
 
     def test_step(self):
         """Run the basic tests (but not necessarily the full regression tests) using make check"""
+
+        if is_double_precision_cuda_build():
+            self.log.info("skipping test step")
+            return
+
         # allow to escape testing by setting runtest to False
         if not self.cfg.get('runtest') and not isinstance(self.cfg.get('runtest'), bool):
 
@@ -411,13 +422,10 @@ class EB_GROMACS(CMakeMake):
         """
         Custom install step for GROMACS; figure out where libraries were installed to.
         """
-        # Skipping the install step if CUDA is enabled and the current iteration
-        # is for double precision
-        cuda = get_software_root('CUDA')
-        if cuda:
-            if re.search('-DGMX_DOUBLE=(ON|YES|TRUE|Y|[1-9])', self.cfg.get('configopts'), re.I):
-                self.log.info("skipping install step")
-                return
+        # Skipping if CUDA is enabled and the current iteration is double precision
+        if is_double_precision_cuda_build():
+            self.log.info("skipping install step")
+            return
 
         # run 'make install' in parallel since it involves more compilation
         self.cfg.update('installopts', "-j %s" % self.cfg.get('parallel'))
@@ -532,10 +540,8 @@ class EB_GROMACS(CMakeMake):
         if not get_software_root('CUDA'):
             for configopts in configopts_list:
                 # add the _d suffix to the suffix, in case of double precission
-                if re.search('-enable-double', configopts, re.I):
-                    dsuff = '_d'
-                if re.search('-DGMX_DOUBLE=(ON|YES|TRUE|Y|[1-9])', configopts, re.I):
-                    dsuff = '_d'
+                if re.search(self.DP_pattern, configopts):
+                        dsuff = '_d'
 
         if dsuff:
             suffixes.extend([dsuff])
@@ -588,6 +594,8 @@ class EB_GROMACS(CMakeMake):
                 'nompi': '',
                 'mpi': '--enable-mpi'
             }
+            # Double precision pattern so search for in configopts
+            self.DP_pattern = '--(enable-double|disable-float)'
         else:
             prec_opts = {
                 'single': '',
@@ -597,6 +605,8 @@ class EB_GROMACS(CMakeMake):
                 'nompi': '-DGMX_MPI=OFF -DGMX_THREAD_MPI=ON',
                 'mpi': '-DGMX_MPI=ON -DGMX_THREAD_MPI=OFF'
             }
+            # Double precision pattern so search for in configopts
+            self.DP_pattern = '-DGMX_DOUBLE=(ON|YES|TRUE|Y|[1-9])'
 
         if LooseVersion(self.version) < LooseVersion('5'):
             # For older versions we only build/install the mdrun part for
--- a/easybuild/easyblocks/g/gromacs.py
+++ b/easybuild/easyblocks/g/gromacs.py
@@ -539,7 +539,7 @@ class EB_GROMACS(CMakeMake):
         if not get_software_root('CUDA'):
             for configopts in configopts_list:
                 # add the _d suffix to the suffix, in case of double precission
-                if re.search(self.DP_pattern, configopts):
+                if self.DP_pattern in configopts:
                     dsuff = '_d'
 
         if dsuff:
--- a/easybuild/easyblocks/g/gromacs.py
+++ b/easybuild/easyblocks/g/gromacs.py
@@ -664,9 +664,9 @@ class EB_GROMACS(CMakeMake):
                 var_buildopts.append(build_opts[mpitype])
                 var_installopts.append(install_opts[mpitype])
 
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/g/gromacs.py
```diff
@@ -444,21 +481,37 @@ def sanity_check_step(self):
...
+        mpi_bins = []
+        mpi_libnames = []
         if LooseVersion(self.version) < LooseVersion('5.1'):
-            bins.extend(['editconf', 'g_lie', 'genbox', 'genconf', 'mdrun'])
+            mpi_bins.extend(['mdrun'])
 
         if LooseVersion(self.version) >= LooseVersion('5.0'):
-            bins.append('gmx')
-            libnames.append('gromacs')
-            if LooseVersion(self.version) < LooseVersion('5.1') and self.toolchain.options.get('usempi', None):
```
Flamefire, line 455: Why was this removed?

easybuild/easyblocks/g/gromacs.py
```diff
@@ -480,13 +533,16 @@ def sanity_check_step(self):
         lib_files = []
         bin_files = []
 
-        for configopts in configopts_list:
-            # add the _d suffix to the suffix, in case of the double precission
-            if re.search('DGMX_DOUBLE=(ON|YES|TRUE|Y|[1-9])', configopts, re.I):
-                suff = '_d'
+        if self.cfg['double_precision'] is None or self.cfg['double_precision']:
```
Flamefire, line 536: You are updating configopts if this is set, don't you? So This check is not required. Removing it catches user-set -DGMX_DOUBLE

easybuild/easyblocks/g/gromacs.py
```diff
@@ -480,13 +533,16 @@ def sanity_check_step(self):
...
+            for configopts in configopts_list:
+                # add the _d suffix to the suffix, in case of double precission
+                if re.search('enable-double', configopts, re.I):
```
Flamefire, line 539: Combine the 2 ifs by OR (`||`) as that's what you do. Easier to understand.

easybuild/easyblocks/g/gromacs.py
```diff
@@ -498,3 +554,115 @@ def sanity_check_step(self):
...
+                'nompi': '',
+                'mpi': ''
+            }
+            install_opts = {
+                'nompi': 'install',
+                'mpi': 'install'
+            }
+
+        precs = ['single']
+        if self.cfg['double_precision'] is None or self.cfg['double_precision']:
```
Flamefire, line 625: So the default is actually True? Why not set that as the default value in the extra_options?

Or is the default (None, unset) to autodetect based on whether CUDA is used? If so, check for CUDA here.

easybuild/easyblocks/g/gromacs.py
```diff
@@ -498,3 +554,115 @@ def sanity_check_step(self):
...
+            build_opts = {
+                'nompi': '',
+                'mpi': ''
+            }
+            install_opts = {
+                'nompi': 'install',
+                'mpi': 'install'
+            }
+
+        precs = ['single']
```
Flamefire, line 624: nitpick: Use `precisions`. Had to scroll back to this to see what `for prec in precs:` means

easybuild/easyblocks/g/gromacs.py
```diff
@@ -498,3 +554,115 @@ def sanity_check_step(self):
...
+        if self.toolchain.options.get('usempi', None):
+            mpitypes.append('mpi')
+
+        # We need to count the numb er of variations to build.
+        self.variants_to_build = 0
+        versions_built = []
+        # Handle the different variants
+        for prec in precs:
+            for mpitype in mpitypes:
+                self.variants_to_build += 1
```
Flamefire, line 638: Just set it after the loop to `len(self.cfg['configopts'])`

easybuild/easyblocks/g/gromacs.py
```diff
@@ -498,3 +554,115 @@ def sanity_check_step(self):
...
+
+                var_buildopts.append(build_opts[mpitype])
+                var_installopts.append(install_opts[mpitype])
+
+                self.cfg.update('configopts', [' '.join(var_confopts) + ' ' + common_config_opts])
+                self.cfg.update('buildopts', [' '.join(var_buildopts) + ' ' + common_build_opts])
+                self.cfg.update('installopts', [' '.join(var_installopts) + ' ' + common_install_opts])
+
+        self.log.debug("List of configure options to iterate over: %s", self.cfg['configopts'])
+        print_msg("Building these versions of GROMACS: %s" % ', '.join(versions_built), silent=self.silent)
```
Flamefire, line 662: Isn't a verbose (stdout shown) message unusal? Why not `self.log.info`?

easybuild/easyblocks/g/gromacs.py
```diff
@@ -59,7 +59,7 @@ class EB_GROMACS(CMakeMake):
     @staticmethod
     def extra_options():
         extra_vars = {
-            'double_precision': [False, "Build with double precision enabled (-DGMX_DOUBLE=ON)", CUSTOM],
+            'double_precision': [None, "Build with double precision enabled (-DGMX_DOUBLE=ON)", CUSTOM],
```
Flamefire, line 62: A note on what `None` means would be good. I couldn't tell if what is done below is correct as I don't know what `None` is supposed to do

easybuild/easyblocks/g/gromacs.py
```diff
@@ -72,6 +72,9 @@ def __init__(self, *args, **kwargs):
         super(EB_GROMACS, self).__init__(*args, **kwargs)
         self.lib_subdir = ''
         self.pre_env = ''
+        self.dynamic = self.toolchain.options.get('dynamic', False)
+        if LooseVersion(self.version) >= LooseVersion('4.6'):
+            self.cfg['separate_build_dir'] = True
```
Flamefire, line 77: Does it fail to build when NOT using a separate build dir? If not, remove. It is the default now :)

akesandgren:
Attacking the comments from @Flamefire piece by piece...

easybuild/easyblocks/g/gromacs.py
```diff
@@ -147,6 +150,33 @@ def prepare_step(self, *args, **kwargs):
...
+                if self.cfg['double_precision']:
+                    raise EasyBuildError("Double precision is not available for GPU build. " +
+                                         "Please explicitly set \"double_precision = False\" " +
+                                         "or remove it in the easyconfig file.")
+                if re.search('DGMX_DOUBLE=(ON|YES|TRUE|Y|[1-9])', self.cfg.get('configopts'), re.I):
+                    if self.cfg['double_precision'] is None:
+                        # Only print warning once when trying double precision
+                        # build the frst time
+                        self.cfg['double_precision'] = False
+                        print_warning("Double precision is not available for " +
```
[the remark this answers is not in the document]
akesandgren: None means that the easyblock decides.

The warning "Double precision is not available for GPU build. Skipping the double precision build." is only printed if this is a CUDA build and -DGMX_DOUBLE=Y is found in the easyconfig and double_precision is unset in the easyconfig.

The "remove it in the easyconfig file." is from the warning before that.
So those warnings are doing the right thing. It's telling you that CUDA and double precision is not available, and that it is skipping it. Unless you have explicitly set double_precision = True, then it is bailing out and telling you it can't go on since double precision and CUDA is not possible.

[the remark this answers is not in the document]
Flamefire: > None means that the easyblock decides.

This should be documented together with how it decides.

> Unfortunately I can't check for CUDA in run_steps...

So that verbose warning will always be printed for default builds? That's not acceptable IMO. As I see the warning is not printed when `self.cfg['double_precision'] == False`, then why not set it to `False` in `run_all_steps`? Also what I don't understand: This now builds single and double precision in one go. So isn't `self.cfg['double_precision']` meaningless? Shouldn't `run_all_steps` take that into account? So `True=>only double`, `False=>Only single`, `None=>Both (with CUDA silently skipped)`? It should still be set to `False` by `run_all_steps` to avoid this warning.

easybuild/easyblocks/g/gromacs.py
akesandgren: To make the user aware that it is going to run x number of builds.

akesandgren:
Reg "Replace self.dynamic by self.cfg['build_shared_libs']" perhaps later. The dynamic flag comes from the cray toolchain and self.dynamic been hijacked a bit and I need to be able to enforce it to be on for later versions.

I can look at that later.

akesandgren:
Reg "Handle CUDA vs double in 1 place, e.g. build_all_steps", haven't look at what i can do in that.

Can look again later, leaving it alone right now.

easybuild/easyblocks/g/gromacs.py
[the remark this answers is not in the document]
akesandgren: In the end I'd prefer to drop the double_precision option. But that's my personal opinion.

Yes, could check if it is set in run_all_steps and only build accordingly, that's a good idea.

But since I can't check which dependencies are available in run_all_steps I need to check it after they become available.
And since it does print that it is going to build double precision (as long as it is set to None) , (the "self.log.debug("List of configure options to iterate over: %s", self.cfg.get('configopts'))" part), even for a CUDA build (since it doesn't know that yet) it's only proper to tell the user that it found out that it can't.

And setting it to False in run_all_steps will make it miss checking for double precision files in sanity_check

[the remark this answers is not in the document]
akesandgren: Oh, and I already do check for double_precision when setting up the build types.

easybuild/easyblocks/g/gromacs.py
Flamefire: Doesn't EB does this already? I remember "Running iteration n" messages

akesandgren: Yes it does but it doesn't tell you why

easybuild/easyblocks/g/gromacs.py
```diff
@@ -324,34 +375,62 @@ def configure_step(self):
...
 
+    def build_step(self):
+        """
+        Custom build step for GROMACS; Skip if CUDA is enabled and the current
+        iteration is for double precision
+        """
+
+        cuda = get_software_root('CUDA')
+        if cuda:
+            if (not self.cfg.get('double_precision') and
```
Flamefire, line 386: I don't understand this condition: If `double_precision` is None or False and it is enabled by configopts? Also what about the configure option? (`--enable-double` IIRC)

easybuild/easyblocks/g/gromacs.py
```diff
@@ -439,49 +478,68 @@ def sanity_check_step(self):
...
             configopts_list = [configopts_list]
 
         lib_files = []
         bin_files = []
 
-        for configopts in configopts_list:
-            # add the _d suffix to the suffix, in case of the double precission
-            if re.search('DGMX_DOUBLE=(ON|YES|TRUE|Y|[1-9])', configopts, re.I):
-                suff = '_d'
+        if self.cfg.get('double_precision') is None or self.cfg.get('double_precision'):
```
Flamefire, line 533: I think this change is wrong:

1. `double_precision` seems to only be used to set configopts and print messages, so this `if` can be removed which helps discussion in https://github.com/easybuilders/easybuild-easyblocks/pull/1991/files#r404754159
2. You moved the `lib_files.extend` out of the loop hence you only check for double precision libs.
3. The code before was wrong anyway and worked only if single precision came before double precision. `suff = ''` has to be in the loop

easybuild/easyblocks/g/gromacs.py
Flamefire: So does none of the other ECs/EBs. You have to look into the log for that.
I'm just mentioning this out of consistency. There has to be a guideline on what to print where IMO or the stdout will be cluttered to unreadability.

easybuild/easyblocks/g/gromacs.py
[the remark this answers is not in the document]
Flamefire: > Unfortunately I can't check for CUDA in run_steps...

An idea: Can't you do `isinstance(self.toolchain, Cuda)` with `from easybuild.toolchains.compiler.cuda import Cuda`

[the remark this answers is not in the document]
ocaisa: > > Unfortunately I can't check for CUDA in run_steps...
> 
> An idea: Can't you do `isinstance(self.toolchain, Cuda)` with `from easybuild.toolchains.compiler.cuda import Cuda`

This would introduce the requirement that CUDA is part of the toolchain, rather than potentially just a simple dependency (not everyone is on the `fosscuda` train yet)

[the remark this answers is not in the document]
Flamefire: But it would work for the offical ECs. And thirdparties have another reason to switch to toolchain-enabled CUDA: A fat warning printed by this EB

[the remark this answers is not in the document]
ocaisa: Of course, it is just a warning, sorry for the noise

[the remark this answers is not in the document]
Flamefire: No, don't misunderstand me. It is good you mentioning this. So we know that we could do that proposed check to avoid the `print_warning` being called for official ECs but also need the check for CUDA when non-official ECs are used.

easybuild/easyblocks/g/gromacs.py
akesandgren: CUDA builds are only available from GROMACS 4.6 and forward
--enable-double is only uesd for < 4.6 builds.

easybuild/easyblocks/g/gromacs.py
```diff
@@ -199,14 +230,47 @@ def configure_step(self):
...
+                if mpiexec:
+                    self.cfg.update('configopts', "-DMPIEXEC=%s" % mpiexec)
+                    self.cfg.update('configopts', "-DMPIEXEC_NUMPROC_FLAG=%s" % self.cfg['mpiexec_numproc_flag'])
+                    self.cfg.update('configopts', "-DNUMPROC=%s" % self.cfg['mpi_numprocs'])
+                elif self.cfg['runtest']:
+                    raise EasyBuildError("'%s' not found in $PATH", self.cfg['mpiexec'])
+                self.log.info("Using %s as MPI executable when testing, with numprocs flag '%s' and %s tasks",
+                              self.cfg['mpiexec'], self.cfg['mpiexec_numproc_flag'], self.cfg['mpi_numprocs'])
+
+            if LooseVersion(self.version) >= LooseVersion('2019'):
```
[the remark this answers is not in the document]
akesandgren: Yes, they changed to year based in 2018

easybuild/easyblocks/g/gromacs.py
akesandgren: Because it's handled by the mpi_bins.extend(['mdrun']) above

easybuild/easyblocks/g/gromacs.py
```diff
@@ -329,6 +378,21 @@ def configure_step(self):
                     if not regex.search(out):
                         raise EasyBuildError("Pattern '%s' not found in GROMACS configuration output.", pattern)
 
+    def build_step(self):
+        """
+        Custom build step for GROMACS; Skip if CUDA is enabled and the current
+        iteration is for double precision
+        """
+
+        cuda = get_software_root('CUDA')
```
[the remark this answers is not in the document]
akesandgren: To be able to skip building in the double_precision+CUDA case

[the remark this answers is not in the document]
Flamefire: Ah ok, skipping configure does not skip build. My bad.

easybuild/easyblocks/g/gromacs.py
```diff
@@ -324,34 +375,60 @@ def configure_step(self):
...
+            self.cfg['runtest'] = None
+
     def install_step(self):
         """
         Custom install step for GROMACS; figure out where libraries were installed to.
-        Also, install the MPI version of the executable in a separate step.
         """
+        # Skipping the install step if CUDA is enabled and the current iteration
+        # is for double precision
+        cuda = get_software_root('CUDA')
```
Flamefire, line 416: Ok after looking at this again: You have the same, non-trivial check at 3 places (configure, build, install)

I can't remember if you can simply omit the double-precision case for CUDA when generating the variant steps, that would be best. Second best: Factor the check into a function like `is_double_precision_cuda_build`   
That would also make the comment unnecessary.

BTW: I think the `re.I` is not required or even wrong. The part `-DGMX_DOUBLE=` has to be upper-case and to be fully correct for the remaining: Everything that is not false is true in CMake, so that check is incomplete anyway. But what you mainly want to check is for `-DGMX_DOUBLE=ON` which you add in https://github.com/easybuilders/easybuild-easyblocks/pull/1991/files#diff-a02b2451ce7221a206c1f64ba6758d33R594. So  `if '-DGMX_DOUBLE=ON' in  self.cfg.get('configopts', '')` should be enough. (or `self.cfg['configopts']` if we can assume it's always there)

If an EC author adds something different: Not your fault ;)

Flamefire:
One more thing I found: This change (and similar) is wrong:
```
        self.cfg.update('installopts', "-j %s" % self.cfg['parallel'])
        self.cfg.update('installopts', "-j %s" % self.cfg.get('parallel'))
```

`get()` would return `None` if not set which would fail the string substitution. In this case we can assume `parallel` to be set (because EB does it) so `get` is not required. The other calls must be checked whether they could be unset and whether the default (`None` if not given) is ok for the code.

easybuild/easyblocks/g/gromacs.py
akesandgren: Yeah, should probably filter out enable-double|DGMX_DOUBLE from configopts coming from the EC but...

Flamefire: Speaking of that: You should probably use `-DGMX_DOUBLE=OFF` for the single precision case or an EC passing `-DGMX_DOUBLE=...` will still make it compile as double. But besides that I'd say: Assume nothing is coming from the EC or it is at least well-behaved (using ON/OFF)

easybuild/easyblocks/g/gromacs.py
```diff
@@ -127,6 +129,14 @@ def get_gromacs_arch(self):
 
         return res
 
+    def is_double_precision_cuda_build(self):
+        """Check if the current build step involves double precision and CUDA"""
+        cuda = get_software_root('CUDA')
+        if cuda:
+            if self.DP_pattern in self.cfg['configopts']:
+                return True
+        return False
```
Flamefire, line 138:
```suggestion
        return cuda and self.DP_pattern in self.cfg['configopts']:
```

for simplicity :)

easybuild/easyblocks/g/gromacs.py
```diff
@@ -493,3 +562,119 @@ def sanity_check_step(self):
...
+            prec_opts = {
+                'single': '--disable-double',
+                'double': '--enable-double',
+            }
+            mpi_type_opts = {
+                'nompi': '--disable-mpi',
+                'mpi': '--enable-mpi'
+            }
+            # Double precision pattern so search for in configopts
+            self.DP_pattern = '--enable-double'
```
Flamefire, line 597: nitpick: Use that in `prec_opts`?

easybuild/easyblocks/g/gromacs.py
akesandgren: Done

easybuild/easyblocks/g/gromacs.py
akesandgren: Did it the other way around instead since that makes more sense.

Flamefire:
Ok so summary of what I think is missing:

- Check where to replace `self.cfg.get` by `self.cfg[]` (Everything from `extra_args` should use `get`, the rest not)
- Make sure the defaults for `self.cfg.get` are reasonable/don't crash the build
- Suggestion/Question in https://github.com/easybuilders/easybuild-easyblocks/pull/1991/files#r402769979

For the last it boils down to (copied from earlier comment):
- CMakeMake supports build_type and build_shared_libs
- Replace self.dynamic by `self.cfg['build_shared_libs']`
- Use `self.lib_ext` (from CMakeMake) IIRC

But check https://github.com/easybuilders/easybuild-easyblocks/pull/1991/files#r402769979 if I misunderstood anything

easybuild/easyblocks/g/gromacs.py
```diff
@@ -72,6 +72,9 @@ def __init__(self, *args, **kwargs):
         super(EB_GROMACS, self).__init__(*args, **kwargs)
         self.lib_subdir = ''
         self.pre_env = ''
+        self.dynamic = self.toolchain.options.get('dynamic', False)
```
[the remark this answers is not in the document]
akesandgren: Can't rely solely on build_shared_libs since it has to support pre CMake GROMACS versions.
Or ??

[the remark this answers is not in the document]
akesandgren: dynamic comes from the Cray toolchain, it was there before I started changing things. I just hijacked it.

But I'd like to take a look at that part build_shared_libs vs dynamic etc at a later point.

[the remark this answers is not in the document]
Flamefire: As far as I can tell you can simply:
- replace `self.dynamic` by `self.cfg['build_shared_libs']`
- get rid of some now unnecessary assignments to the latter
- Remove `-DBUILD_SHARED_LIBS` settings
- Likely replace https://github.com/easybuilders/easybuild-easyblocks/blob/52fc2230988cf1bdded69a7a7db03f68c56474fd/easybuild/easyblocks/g/gromacs.py#L435-L438 to what CMakeMake already sets (although you'd need to double-check it gets set even when CMakeMakes configure is not called)

So it's mostly a Search&Replace and removal of code.

One point where I'm not 100% sure: https://github.com/easybuilders/easybuild-easyblocks/blob/52fc2230988cf1bdded69a7a7db03f68c56474fd/easybuild/easyblocks/g/gromacs.py#L294-L295 why the `if` on `plumed_root` and what does `GMX_PREFER_STATIC_LIBS` do what `BUILD_SHARED_LIBS` does not?

If you don't feel like doing that please move this discussion into an issue.

akesandgren:
I think i got all self.cfg that can use [] done. As for making sure the cfg.get has proper default values I think only mpiexec/mpiexec_numproc_flag and possible one mpisuffix is at risk.
And only if someone uses the gromacs easytblock as super...

And they should really use the new mpirun stuff from framework, but i currently don't know how that works. Need to look into that later.

easybuild/easyblocks/g/gromacs.py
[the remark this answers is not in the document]
akesandgren: using lib_ext() when building a non-CMake version fails with:
    self.libext = self.lib_ext()
TypeError: 'NoneType' object is not callable

So leaving that as it is.

the "if plumed_root" was a left over from before the changes. Removed.

GMX_PREFER_STATIC_LIBS=ON makes it use static libs for external things regardless of BUILD_SHARED_LIBS
and vice versa. I'm not sure we need this but it doesn't hurt.

easybuild/easyblocks/g/gromacs.py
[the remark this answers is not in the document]
akesandgren: The warning is now only in the log. Still haven't found a way to detect CUDA in a convenient place that gets rid of the extra iterations for double precision.

Flamefire:
> I think i got all self.cfg that can use [] done.

What about `runtest` and the `*opts` and `*_cmd`? I think those can safely assumed to be present

easybuild/easyblocks/g/gromacs.py
```diff
@@ -324,34 +378,61 @@ def configure_step(self):
...
+            old_runtest = self.cfg['runtest']
             # make very sure OMP_NUM_THREADS is set to 1, to avoid hanging GROMACS regression test
             env.setvar('OMP_NUM_THREADS', '1')
 
-            self.cfg['runtest'] = 'check'
-            if self.cfg['parallel']:
-                # run 'make check' in parallel since it involves more compilation
-                self.cfg.update('runtest', "-j %s" % self.cfg['parallel'])
+            if isinstance(self.cfg['runtest'], bool):
+                self.cfg['runtest'] = 'check'
```
migueldiascosta, line 408: I think this should also be done if `runtest` is `Non`e, otherwise the `update` below will fail in that case

Flamefire: Actually: Why the condition at all? Why not just `self.cfg['runtest'] = 'check -j %s' % self.cfg['parallel']`?
I get that it may allow runtest to be a string with a custom target but that hasn't been the case before and I don't see why it is needed

easybuild/easyblocks/g/gromacs.py
```diff
@@ -324,34 +378,61 @@ def configure_step(self):
...
     def test_step(self):
         """Run the basic tests (but not necessarily the full regression tests) using make check"""
+
+        if self.is_double_precision_cuda_build():
+            self.log.info("skipping test step")
+            return
+
         # allow to escape testing by setting runtest to False
-        if not self.cfg['runtest'] and not isinstance(self.cfg['runtest'], bool):
+        if self.cfg['runtest'] is None or self.cfg['runtest']:
```
Flamefire, line 401: Good catch fixing this. Just a note that this changes behavior. Previously it would be executed only if `runtest=""` or `runtest=None` which is very weird.

easybuild/easyblocks/g/gromacs.py
akesandgren: I made it to make it easy to pick another make target if they decide to use one.

easybuild/easyblocks/g/gromacs.py
akesandgren: yes it does change behavior, but at least now it is correct.

migueldiascosta:
currently testing all existing GROMACS easyconfigs with this easyblock, so far so good

easybuild/easyblocks/g/gromacs.py
```diff
@@ -493,3 +563,118 @@ def sanity_check_step(self):
...
+                'single': '-DGMX_DOUBLE=OFF',
+                'double': '-DGMX_DOUBLE=ON',
+            }
+            mpi_type_opts = {
+                'nompi': '-DGMX_MPI=OFF -DGMX_THREAD_MPI=ON',
+                'mpi': '-DGMX_MPI=ON -DGMX_THREAD_MPI=OFF'
+            }
+
+        # Double precision pattern so search for in configopts
+        self.DP_pattern = prec_opts['double']
```
boegel, line 608: @akesandgren Please rename this to `self.double_prec_pattern` (mostly to avoid the capitals `DP`, which is confusing, capital letters are generally only used in class names)

easybuild/easyblocks/g/gromacs.py
```diff
@@ -201,14 +238,48 @@ def configure_step(self):
                 run_cmd(plumed_cmd, log_all=True, simple=True)
 
         else:
+            if '-DGMX_MPI=ON' in self.cfg['configopts']:
+                if self.cfg.get('mpi_numprocs', 0) == 0:
```
boegel, line 242: I would grab `mpi_numprocs` from `self.cfg` once, which cleans things up a bit below (and lowers the risk for introducing bugs by fat fingering on `'mpi_numprocs'`):

```python
mpi_numprocs = self.cfg.get('mpi_numprocs', 0)
if mpi_numprocs == 0:
    ...
    mpi_numprocs = self.cfg['parallel']
...

self.cfg.update('configopts', "-DNUMPROC=%s" % mpi_numprocs)

File: easybuild/easyblocks/g/gromacs.py [hunk 23]
```diff
@@ -201,14 +238,48 @@ def configure_step(self):
...
+                    self.log.info("No number of test MPI tasks specified -- using default: %s",
+                                  self.cfg['parallel'])
+                    self.cfg['mpi_numprocs'] = self.cfg['parallel']
+
+                elif self.cfg.get('mpi_numprocs') > self.cfg['parallel']:
+                    self.log.warning("Number of test MPI tasks (%s) is greater than value for 'parallel': %s",
+                                     self.cfg.get('mpi_numprocs'), self.cfg['parallel'])
+
+                mpiexec = which(self.cfg.get('mpiexec'))
+                if mpiexec:
```
boegel, line 252: You should be a bit more careful here, usign `which(None)` would end in tears:

```python
mpiexec = self.cfg.get('mpiexec')
if mpiexec:
    mpiexec = which(mpiexec)
    if mpi_exec:
        ...
    elif self.cfg['runtest']:
        raise EasyBuildError("'%s' not found in $PATH", mpiexec')
elif self.cfg['runtest']:
    raise EasyBuildError("No value found for 'mpiexec'")

File: easybuild/easyblocks/g/gromacs.py [hunk 24]
```diff
@@ -201,14 +238,48 @@ def configure_step(self):
...
+                              self.cfg.get('mpiexec'), self.cfg.get('mpiexec_numproc_flag'),
+                              self.cfg.get('mpi_numprocs'))
+
+            if LooseVersion(self.version) >= LooseVersion('2019'):
+                # Building the gmxapi interface requires shared libraries
+                self.cfg['build_shared_libs'] = True
+                self.cfg.update('configopts', "-DGMXAPI=ON")
+
+                if LooseVersion(self.version) >= LooseVersion('2020'):
+                    # check whether Python is loaded as a dependency
```
boegel, line 268:
```python
# build Python bindings if Python is loaded as a dependency
```

easybuild/easyblocks/g/gromacs.py
```diff
@@ -201,14 +238,48 @@ def configure_step(self):
...
+                    python_root = get_software_root('Python')
+                    if python_root:
+                        self.cfg.update('configopts',
+                                        "-DPYTHON_EXECUTABLE=%s" % os.path.join(python_root, 'bin', 'python'))
```
boegel, line 272:
```python
if python_root:
    bin_python = os.path.join(python_root, 'bin', 'python')
    self.cfg.update('configopts', "-DPYTHON_EXECUTABLE=%s" % bin_python)
    self.cfg.update('configopts', "-DGMX_PYTHON_PACKAGE=ON")

File: easybuild/easyblocks/g/gromacs.py [hunk 26]
```diff
@@ -324,34 +378,61 @@ def configure_step(self):
...
+        """
+        Custom build step for GROMACS; Skip if CUDA is enabled and the current
+        iteration is for double precision
+        """
+
+        if self.is_double_precision_cuda_build():
+            self.log.info("skipping build step")
+            return
+
+        super(EB_GROMACS, self).build_step()
```
boegel, line 391: Not a big fan of inline `return` statements, can be easily avoided here:

```python
if self.is_double_precision_cuda_build():
    self.log.info("skipping build step")
else:
    super(EB_GROMACS, self).build_step()

File: easybuild/easyblocks/g/gromacs.py [hunk 27]
```diff
@@ -324,34 +378,61 @@ def configure_step(self):
...
+
     def test_step(self):
         """Run the basic tests (but not necessarily the full regression tests) using make check"""
+
+        if self.is_double_precision_cuda_build():
+            self.log.info("skipping test step")
+            return
```
boegel, line 398:
```python
if self.is_double_precision_cuda_build():
    self.log.info("skipping test step")
else:
    ...

File: easybuild/easyblocks/g/gromacs.py [hunk 28]
```diff
@@ -324,34 +378,61 @@ def configure_step(self):
...
             env.setvar('OMP_NUM_THREADS', '1')
 
-            self.cfg['runtest'] = 'check'
-            if self.cfg['parallel']:
-                # run 'make check' in parallel since it involves more compilation
-                self.cfg.update('runtest', "-j %s" % self.cfg['parallel'])
+            if self.cfg['runtest'] is None or isinstance(self.cfg['runtest'], bool):
+                self.cfg['runtest'] = 'check'
+
+            # run 'make check' or whever the easyconfig specifies
```
boegel, line 410: `whever` -> `whatever`?

easybuild/easyblocks/g/gromacs.py
```diff
@@ -324,34 +378,61 @@ def configure_step(self):
...
+
     def install_step(self):
         """
         Custom install step for GROMACS; figure out where libraries were installed to.
-        Also, install the MPI version of the executable in a separate step.
         """
+        # Skipping if CUDA is enabled and the current iteration is double precision
+        if self.is_double_precision_cuda_build():
+            self.log.info("skipping install step")
+            return
```
boegel, line 424:
```python
if self.is_double_precision_cuda_build():
    self.log.info("skipping install step")
else:
    ...

File: easybuild/easyblocks/g/gromacs.py [hunk 30]
```diff
@@ -372,55 +453,19 @@ def install_step(self):
...
+        # Reset installopts etc for the benefit of the gmxapi extension
+        self.cfg['installopts'] = self.save_installopts
 
-                # clean up obj dir before reconfiguring
-                shutil.rmtree(os.path.join(self.builddir, 'easybuild_obj'))
+    def extensions_step(self, fetch=False):
+        """ Custom extensions step, only handle extensions after the last iteration round"""
+        if self.iter_idx < self.variants_to_build-1:
+            self.log.info("skipping extension step %s", self.iter_idx)
+            return
```
boegel, line 463: no inline `return` + spaces around `-`:

```python
if self.iter_idx < self.variants_to_build - 1:
    self.log.info("skipping extension step %s", self.iter_idx)
else:
    ...

File: easybuild/easyblocks/g/gromacs.py [hunk 31]
```diff
@@ -372,55 +453,19 @@ def install_step(self):
...
-                # rebuild/test/install with MPI options
-                super(EB_GROMACS, self).configure_step()
-                super(EB_GROMACS, self).build_step()
-                super(EB_GROMACS, self).test_step()
-                super(EB_GROMACS, self).install_step()
-
-                self.log.info("A full regression test suite is available from the GROMACS web site")
+        # Set runtest to None so that the gmxapi extension doesn't try to
+        # run "check" as a command
+        self.cfg['runtest'] = None
```
boegel, line 467: Shouldn't you restore the original `runtest` value afterwards, since a custom value may have been specified in the easyconfig file?

easybuild/easyblocks/g/gromacs.py
```diff
@@ -475,13 +536,22 @@ def sanity_check_step(self):
...
         bin_files = []
 
-        for configopts in configopts_list:
-            # add the _d suffix to the suffix, in case of the double precission
-            if re.search('DGMX_DOUBLE=(ON|YES|TRUE|Y|[1-9])', configopts, re.I):
-                suff = '_d'
+        dsuff = None
+        if not get_software_root('CUDA'):
+            for configopts in configopts_list:
+                # add the _d suffix to the suffix, in case of double precission
```
boegel, line 542: `precission` -> `precision`

easybuild/easyblocks/g/gromacs.py
```diff
@@ -475,13 +536,22 @@ def sanity_check_step(self):
...
+                    dsuff = '_d'
+
+        if dsuff:
+            suffixes.extend([dsuff])
 
-            lib_files.extend(['lib%s%s.%s' % (libname, suff, self.libext) for libname in libnames])
-            bin_files.extend([b + suff for b in bins])
+        lib_files.extend([
+            'lib%s%s.%s' % (libname, suff, self.libext) for libname in libnames + mpi_libnames for suff in suffixes
+        ])
```
boegel, line 551: I would replace `libname` with `x` here, which makes it shorter and more readable (due to the `libname` vs `libnames`):

```python
lib_files.extend(['lib%s%s.%s' % (x, suff, self.libext) for x in libnames + mpi_libnames for suff in suffixes])

File: easybuild/easyblocks/g/gromacs.py [hunk 34]
```diff
@@ -475,13 +536,22 @@ def sanity_check_step(self):
...
+        bin_files.extend([
+            b + suff for b in bins + mpi_bins for suff in suffixes
+        ])
```
boegel, line 554: Just collapse to a single line?

```python
        bin_files.extend([b + suff for b in bins + mpi_bins for suff in suffixes])

File: easybuild/easyblocks/g/gromacs.py [hunk 35]
```diff
@@ -493,3 +563,118 @@ def sanity_check_step(self):
...
+            build_opts = {
+                'nompi': '',
+                'mpi': ''
+            }
+            install_opts = {
+                'nompi': 'install',
+                'mpi': 'install'
+            }
+
+        precisions = ['single']
```
boegel, line 636: @akesandgren Does it make sense to opt-out of building single precision, via `single_precision = False`?

easybuild/easyblocks/g/gromacs.py
```diff
@@ -493,3 +563,118 @@ def sanity_check_step(self):
...
+        if self.cfg.get('double_precision') is None or self.cfg.get('double_precision'):
+            precisions.append('double')
+
+        mpitypes = ['nompi']
```
boegel, line 640: Make it possible to opt-out of non-MPI installation via `nompi_variant = False`?

easybuild/easyblocks/g/gromacs.py
```diff
@@ -493,3 +563,118 @@ def sanity_check_step(self):
...
+        if self.toolchain.options.get('usempi', None):
+            mpitypes.append('mpi')
+
+        # We need to count the numb er of variations to build.
```
boegel, line 644: `numb er` -> `number`

easybuild/easyblocks/g/gromacs.py
```diff
@@ -493,3 +563,118 @@ def sanity_check_step(self):
...
+        self.variants_to_build = 0
```
boegel, line 645: no need to initialize here, since we're counting with `len` below?

easybuild/easyblocks/g/gromacs.py
```diff
@@ -493,3 +563,118 @@ def sanity_check_step(self):
...
+        versions_built = []
+        # Handle the different variants
+        for precicion in precisions:
```
boegel, line 648: `precicion` -> `precision`

easybuild/easyblocks/g/gromacs.py
```diff
@@ -493,3 +563,118 @@ def sanity_check_step(self):
...
+                var_confopts = []
+                var_buildopts = []
+                var_installopts = []
+
+                var_confopts.append(mpi_type_opts[mpitype])
+                var_confopts.append(prec_opts[precicion])
+                if LooseVersion(self.version) < LooseVersion('4.6'):
+                    suffix = ''
+                    if mpitype == 'mpi':
+                        suffix = "--program-suffix={0}".format(self.cfg.get('mpisuffix'))
```
boegel, line 660: What if `mpisuffix` is `None`?

easybuild/easyblocks/g/gromacs.py
```diff
@@ -493,3 +563,118 @@ def sanity_check_step(self):
...
+                        if precicion == 'double':
+                            suffix += '_d'
+                    var_confopts.append(suffix)
+
+                var_buildopts.append(build_opts[mpitype])
+                var_installopts.append(install_opts[mpitype])
+
+                self.cfg.update('configopts', [' '.join(var_confopts) + ' ' + common_config_opts])
```
boegel, line 668: I think there's no need for the `[...]`? And you can avoid the `+ ' ' +` too:

```python
self.cfg.update('configopts', ' '.join(var_confopts + [common_config_opts]))

File: easybuild/easyblocks/g/gromacs.py [hunk 42]
```diff
@@ -493,3 +563,118 @@ def sanity_check_step(self):
...
+                self.cfg.update('buildopts', [' '.join(var_buildopts) + ' ' + common_build_opts])
+                self.cfg.update('installopts', [' '.join(var_installopts) + ' ' + common_install_opts])
+        self.variants_to_build = len(self.cfg['configopts'])
+
+        self.log.debug("List of configure options to iterate over: %s", self.cfg['configopts'])
+        self.log.info("Building these versions of GROMACS: %s", ', '.join(versions_built))
```
boegel, line 674:
```python
self.log.info("Building these %d versions of GROMACS: %s", len(versions_built), ', '.join(versions_built))
```

Maybe replace `versions` with `variants`?

easybuild/easyblocks/g/gromacs.py
```diff
@@ -493,3 +563,118 @@ def sanity_check_step(self):
...
+        self.save_installopts = self.cfg['installopts']
+
+        # keep track of config/build/installopts specified in easyconfig
+        # file, so we can include them in each iteration later
+        common_config_opts = self.cfg['configopts']
+        common_build_opts = self.cfg['buildopts']
+        common_install_opts = self.cfg['installopts']
+
+        self.save_install_cmd = self.cfg['install_cmd']
+        self.save_build_cmd = self.cfg['build_cmd']
```
boegel, line 582: rename to `self.orig_install_cmd`?

easybuild/easyblocks/g/gromacs.py
```diff
@@ -493,3 +563,118 @@ def sanity_check_step(self):
...
         }
         super(EB_GROMACS, self).sanity_check_step(custom_paths=custom_paths)
+
+    def run_all_steps(self, *args, **kwargs):
+        """
+        Put configure options in place for different variants, (no)mpi, single/double precision.
+        """
+        # Save installopts so we can reset it later. The gmxapi pip install
+        # can't handle the -j argument.
+        self.save_installopts = self.cfg['installopts']
```
boegel, line 573: rename to `self.orig_installopts`?

easybuild/easyblocks/g/gromacs.py
```diff
@@ -493,3 +563,118 @@ def sanity_check_step(self):
...
+        self.variants_to_build = len(self.cfg['configopts'])
+
+        self.log.debug("List of configure options to iterate over: %s", self.cfg['configopts'])
+        self.log.info("Building these versions of GROMACS: %s", ', '.join(versions_built))
+        return super(EB_GROMACS, self).run_all_steps(*args, **kwargs)
+
+        self.cfg['install_cmd'] = self.save_install_cmd
+        self.cfg['build_cmd'] = self.save_build_cmd
+
+        self.log.info("A full regression test suite is available from the GROMACS web site")
```
boegel, line 680: This log message is useful without a URL? :)

easybuild/easyblocks/g/gromacs.py
akesandgren: Done

easybuild/easyblocks/g/gromacs.py
akesandgren: Done

easybuild/easyblocks/g/gromacs.py
```diff
@@ -493,3 +563,118 @@ def sanity_check_step(self):
...
+
+            # Use the fact that for older versions we just need to
+            # build and install mdrun for the MPI part
+            build_opts = {
+                'nompi': '',
+                'mpi': 'mdrun'
+            }
+            install_opts = {
+                'nompi': 'install',
+                'mpi': 'install-mdrun'
```
boegel, line 625: You should change this to `make install-mdrun` and customize `self.cfg['install_cmd']` instead, since now we're running `make install install` for example, which is silly/wrong

easybuild/easyblocks/g/gromacs.py
akesandgren: Done

easybuild/easyblocks/g/gromacs.py
akesandgren: Done

easybuild/easyblocks/g/gromacs.py
akesandgren: Done

easybuild/easyblocks/g/gromacs.py
akesandgren: Done

easybuild/easyblocks/g/gromacs.py
akesandgren: Done

easybuild/easyblocks/g/gromacs.py
akesandgren: Done

easybuild/easyblocks/g/gromacs.py
akesandgren: Done

easybuild/easyblocks/g/gromacs.py
akesandgren: Done

easybuild/easyblocks/g/gromacs.py
akesandgren: Done

easybuild/easyblocks/g/gromacs.py
akesandgren: Do I need to? This is at the last iteration and extensions_step is that last thing done. So nothing else should happen as far as I know.

easybuild/easyblocks/g/gromacs.py
akesandgren: Done

easybuild/easyblocks/g/gromacs.py
akesandgren: Done

easybuild/easyblocks/g/gromacs.py
akesandgren: Not really, single precision is THE precision for Gromacs, double precision is only used in certain special cases.

easybuild/easyblocks/g/gromacs.py
akesandgren: You must have the nompi build or the basic tools doesn't exist.

easybuild/easyblocks/g/gromacs.py
akesandgren: Done

easybuild/easyblocks/g/gromacs.py
akesandgren: Done

easybuild/easyblocks/g/gromacs.py
akesandgren: Done

easybuild/easyblocks/g/gromacs.py
akesandgren: You mean if it's explicitly set to None in the easyconfig?
Then they get what they deserve, failure.... but fixed it.

easybuild/easyblocks/g/gromacs.py
akesandgren: Done

easybuild/easyblocks/g/gromacs.py
akesandgren: Done

easybuild/easyblocks/g/gromacs.py
akesandgren: Yeah, but perhaps one could pull homepage from the easyconfig easily?

easybuild/easyblocks/g/gromacs.py
akesandgren: Done

easybuild/easyblocks/g/gromacs.py
akesandgren: configopts is a list here, don't I need to give it a list then?

easybuild/easyblocks/g/gromacs.py
akesandgren: Ahh, forgot to change install_cmd for v >= 5
For v < 5 it's correctly done.

Fixed for all versions.

easybuild/easyblocks/g/gromacs.py
boegel: Why not mention a direct URL in the log message?

easybuild/easyblocks/g/gromacs.py
boegel: No, it will leverage it to a list automatically, see https://github.com/easybuilders/easybuild-framework/blob/develop/easybuild/framework/easyconfig/easyconfig.py#L575

easybuild/easyblocks/g/gromacs.py
boegel: OK, fine

easybuild/easyblocks/g/gromacs.py
boegel: OK then

easybuild/easyblocks/g/gromacs.py
boegel: It's good to avoid nasty crashes that just spit out a traceback

easybuild/easyblocks/g/gromacs.py
```diff
@@ -201,14 +238,54 @@ def configure_step(self):
...
+                mpiexec = self.cfg.get('mpiexec')
+                if mpiexec:
+                    mpiexec = which(mpiexec)
+                    if mpiexec:
+                        self.cfg.update('configopts', "-DMPIEXEC=%s" % mpiexec)
+                        self.cfg.update('configopts', "-DMPIEXEC_NUMPROC_FLAG=%s" %
+                                        self.cfg.get('mpiexec_numproc_flag'))
+                        self.cfg.update('configopts', "-DNUMPROC=%s" % mpi_numprocs)
+                    elif self.cfg['runtest']:
+                        raise EasyBuildError("'%s' not found in $PATH", self.cfg.get('mpiexec'))
```
boegel, line 261: also use `mpiexec` here:

```python
                         raise EasyBuildError("'%s' not found in $PATH", mpiexec')

File: easybuild/easyblocks/g/gromacs.py [hunk 48]
```diff
@@ -201,14 +238,54 @@ def configure_step(self):
...
+                                  self.cfg['parallel'])
+                    mpi_numprocs = self.cfg['parallel']
+
+                elif mpi_numprocs > self.cfg['parallel']:
+                    self.log.warning("Number of test MPI tasks (%s) is greater than value for 'parallel': %s",
+                                     mpi_numprocs, self.cfg['parallel'])
+
+                mpiexec = self.cfg.get('mpiexec')
+                if mpiexec:
+                    mpiexec = which(mpiexec)
```
boegel, line 254: better rename `mpiexec = ` to `mpiexec_path = ` here, so you can reuse `mpiexec` below?

easybuild/easyblocks/g/gromacs.py
```diff
@@ -201,14 +238,54 @@ def configure_step(self):
...
+                        self.cfg.update('configopts', "-DMPIEXEC=%s" % mpiexec)
+                        self.cfg.update('configopts', "-DMPIEXEC_NUMPROC_FLAG=%s" %
+                                        self.cfg.get('mpiexec_numproc_flag'))
+                        self.cfg.update('configopts', "-DNUMPROC=%s" % mpi_numprocs)
+                    elif self.cfg['runtest']:
+                        raise EasyBuildError("'%s' not found in $PATH", self.cfg.get('mpiexec'))
+                else:
+                    raise EasyBuildError("No value found for 'mpiexec'")
+                self.log.info("Using %s as MPI executable when testing, with numprocs flag '%s' and %s tasks",
+                              self.cfg.get('mpiexec'), self.cfg.get('mpiexec_numproc_flag'),
```
boegel, line 265: `self.cfg.get('mpiexec')` -> `mpiexec_path`

easybuild/easyblocks/g/gromacs.py
```diff
@@ -324,103 +384,91 @@ def configure_step(self):
...
+            self.cfg['installopts'] = self.orig_installopts
+
+    def extensions_step(self, fetch=False):
+        """ Custom extensions step, only handle extensions after the last iteration round"""
+        if self.iter_idx < self.variants_to_build - 1:
+            self.log.info("skipping extension step %s", self.iter_idx)
+        else:
+            # Set runtest to None so that the gmxapi extension doesn't try to
+            # run "check" as a command
+            self.cfg['runtest'] = None
```
boegel, line 470: still need to restore original `runtest` value here?

easybuild/easyblocks/g/gromacs.py
akesandgren: Done

easybuild/easyblocks/g/gromacs.py
akesandgren: Done

easybuild/easyblocks/g/gromacs.py
akesandgren: Done

easybuild/easyblocks/g/gromacs.py
akesandgren: Did you miss my question on the previous review of this line?

Repeating:
Do I need to? This is at the last iteration and extensions_step is that last thing done. So nothing else should happen as far as I know.

easybuild/easyblocks/g/gromacs.py
akesandgren: My question above was if I can pick the homepage value from the easyconfig easily and use that in the log message. More future proof

easybuild/easyblocks/g/gromacs.py
akesandgren: Done

easybuild/easyblocks/g/gromacs.py
akesandgren: Anyway, done

easybuild/easyblocks/g/gromacs.py
akesandgren: Anyway, done

easybuild/easyblocks/g/gromacs.py
boegel: Sure: `self.cfg['homepage']` (just like any other easyconfig parameter)

easybuild/easyblocks/g/gromacs.py
boegel: better to do, to avoid introducing bugs later when we make more changes

easybuild/easyblocks/g/gromacs.py
akesandgren: Done

boegel:
Retested extensively with a whole range of easyconfigs, old and new, works like a charm.

Thanks a lot @akesandgren!

PR #1992, opened 2020-03-15
update Open Babel easyblock for version 3.0.0

smoors:


easybuild/easyblocks/o/openbabel.py
```diff
@@ -97,7 +100,7 @@ def configure_step(self):
     def sanity_check_step(self):
         """Custom sanity check for OpenBabel."""
         custom_paths = {
-            'files': ['bin/babel', 'lib/libopenbabel.%s' % get_shared_lib_ext()],
+            'files': ['bin/obabel', 'lib/libopenbabel.%s' % get_shared_lib_ext()],
```
akesandgren, line 103: Is this always true for all versions?

boegel: Yes, it is.

boegel:
Thanks @smoors!

PR #1995, opened 2020-03-19
Allow differing names for TF wheel

Flamefire:
This provides better resistance against changes from the TF side but mainly allows to build custom version like "TensorFlow master" (which I did to have some local EC which allows me to install the current master)

Problem was: The wheel was still named `tensorflow-2.1.0.whl` or something while the version set in the EC could be 2nightly or so


akesandgren:
Going in, thanks @Flamefire!

PR #1997, opened 2020-03-21
Update lammps easyblock to be aware of MKL

ocaisa:


boegel:
close/re-open to wake up Travis (they had some maintenance going on this weekend)

PR #2000, opened 2020-03-24
QtWebEngine is not fully supported on POWER - so do not sanity check it

branfosj:
One of the components of QtWebEngine is not (currently) support on POWER, so we should skip the sanity check on it.


akesandgren:
Going in, thanks @branfosj!

PR #2001, opened 2020-03-24
use mpiexec to run import test in PythonPackage easyblock if usempi is enabled

lexming:
Python packages using MPI should run the sanity check command `python -c "import module_name"` with `mpirun` to avoid failed initializations of MPI.

Fixes: https://github.com/easybuilders/easybuild-easyconfigs/issues/10213

### DIFF
--- a/easybuild/easyblocks/generic/pythonpackage.py
+++ b/easybuild/easyblocks/generic/pythonpackage.py
@@ -691,7 +691,7 @@ class PythonPackage(ExtensionEasyBlock):
 
         if self.toolchain.options.get('usempi', None):
             # packages using MPI have to execute the sanity checks with mpirun
-            exts_filter = ("mpirun -n 1 " + exts_filter[0], exts_filter[1])
+            exts_filter = ("mpiexec -n 1 " + exts_filter[0], exts_filter[1])
 
         if not self.multi_python:
             # 'python' is replaced by full path to active 'python' command if a single Python version is used
--- a/easybuild/easyblocks/generic/pythonpackage.py
+++ b/easybuild/easyblocks/generic/pythonpackage.py
@@ -687,16 +687,18 @@ class PythonPackage(ExtensionEasyBlock):
             }
 
         # make sure 'exts_filter' is defined, which is used for sanity check
-        if self.multi_python:
-            # when installing for multiple Python versions, we must use 'python', not a full-path 'python' command!
-            if 'exts_filter' not in kwargs:
-                kwargs.update({'exts_filter': EXTS_FILTER_PYTHON_PACKAGES})
-        else:
-            # 'python' is replaced by full path to active 'python' command
+        exts_filter = EXTS_FILTER_PYTHON_PACKAGES
+
+        if self.toolchain.options.get('usempi', None):
+            # packages using MPI have to execute the sanity checks with mpirun
+            exts_filter = ("mpirun -n 1 " + exts_filter[0], exts_filter[1])
+
+        if not self.multi_python:
+            # 'python' is replaced by full path to active 'python' command if a single Python version is used
             # (which is required especially when installing with system Python)
-            if 'exts_filter' not in kwargs:
-                orig_exts_filter = EXTS_FILTER_PYTHON_PACKAGES
-                exts_filter = (orig_exts_filter[0].replace('python', self.python_cmd), orig_exts_filter[1])
+            exts_filter = (exts_filter[0].replace('python', self.python_cmd), exts_filter[1])
+
+        if 'exts_filter' not in kwargs:
                 kwargs.update({'exts_filter': exts_filter})
 
         if self.cfg.get('sanity_pip_check', False):


easybuild/easyblocks/generic/pythonpackage.py
```diff
@@ -687,16 +687,18 @@ def sanity_check_step(self, *args, **kwargs):
...
-            # when installing for multiple Python versions, we must use 'python', not a full-path 'python' command!
-            if 'exts_filter' not in kwargs:
-                kwargs.update({'exts_filter': EXTS_FILTER_PYTHON_PACKAGES})
-        else:
-            # 'python' is replaced by full path to active 'python' command
+        exts_filter = EXTS_FILTER_PYTHON_PACKAGES
+
+        if self.toolchain.options.get('usempi', None):
+            # packages using MPI have to execute the sanity checks with mpirun
+            exts_filter = ("mpirun -n 1 " + exts_filter[0], exts_filter[1])
```
ocaisa: This should either be configurable or should use `mpiexec` from the standard. `mpirun` would fail on the JSC systems where you are required to use `srun`.

lexming: I always forget that `mpirun` is actually not part of the standard

lexming:
Successfully tested with:
* h5py-2.10.0-foss-2020a-Python-3.8.2.eb (`usempi: True`)
* h5py-2.10.0-intel-2020a-Python-3.8.2.eb (`usempi: True`)
* networkx-2.4-foss-2020a-Python-3.8.2.eb (`usempi: False`)

easybuild/easyblocks/generic/pythonpackage.py
lexming: Changed it to `mpiexec` for simplicity

akesandgren:
Don't we already have a variable for this that should be used? @boegel ??

boegel:
@akesandgren There's `self.toolchain.mpi_cmd_for` which we use in a couple of easyblocks, but that uses `mpirun` rather than `mpiexec` too...

easybuild/easyblocks/generic/pythonpackage.py
```diff
@@ -687,16 +687,18 @@ def sanity_check_step(self, *args, **kwargs):
...
+        if not self.multi_python:
+            # 'python' is replaced by full path to active 'python' command if a single Python version is used
             # (which is required especially when installing with system Python)
-            if 'exts_filter' not in kwargs:
-                orig_exts_filter = EXTS_FILTER_PYTHON_PACKAGES
-                exts_filter = (orig_exts_filter[0].replace('python', self.python_cmd), orig_exts_filter[1])
+            exts_filter = (exts_filter[0].replace('python', self.python_cmd), exts_filter[1])
+
+        if 'exts_filter' not in kwargs:
                 kwargs.update({'exts_filter': exts_filter})
```
boegel: indent is off by one level here

easybuild/easyblocks/generic/pythonpackage.py
```diff
@@ -687,16 +687,18 @@ def sanity_check_step(self, *args, **kwargs):
             }
 
         # make sure 'exts_filter' is defined, which is used for sanity check
-        if self.multi_python:
-            # when installing for multiple Python versions, we must use 'python', not a full-path 'python' command!
```
boegel: please retain this comment below, so we're not left wondering why replacing `'python'` with the full path is not done when using multi-Python

easybuild/easyblocks/generic/pythonpackage.py
```diff
@@ -687,16 +687,18 @@ def sanity_check_step(self, *args, **kwargs):
...
-            if 'exts_filter' not in kwargs:
-                kwargs.update({'exts_filter': EXTS_FILTER_PYTHON_PACKAGES})
-        else:
-            # 'python' is replaced by full path to active 'python' command
+        exts_filter = EXTS_FILTER_PYTHON_PACKAGES
+
+        if self.toolchain.options.get('usempi', None):
+            # packages using MPI have to execute the sanity checks with mpirun
```
boegel: not mpirun, mpiexec

ocaisa:
At least that is configurable though with `mpi_cmd_template` (after investigating)

boegel:
> At least that is configurable though with `mpi_cmd_template` (after investigating)

Please open an issue in framework with some motivation, maybe we need to replace `mpirun` with `mpiexec` in `mpi_cmd_for`

akesandgren:
@boegel I though self.toolchain.mpi_cmd_for was configurable from the outside too?

ocaisa:
> Please open an issue in framework with some motivation, maybe we need to replace `mpirun` with `mpiexec` in `mpi_cmd_for`

Sure I can do that but what I meant was that if we used `mpi_cmd_for` then it is possible to handle funky site specifics with a configuration variable, see https://github.com/easybuilders/easybuild-framework/blob/develop/easybuild/tools/toolchain/mpi.py#L178

True that it is not a very obvious though, name is a a bit off.

lexming:
I got feedback from Intel on the original issue that triggered this change (https://github.com/easybuilders/easybuild-easyconfigs/issues/10213). They acknowledge the issue and it should be fixed at some point.

Therefore, this PR is **not** the way to go. It should not be necessary to use `mpiexec/mpirun` to perform simple executions of binaries linked to Intel MPI (or any other MPI implementation). So, the sanity checks in Python packages are correct as they are done now. In the case of the bug at hand with Intel MPI 2019.6, the best solution is to fix affected easyconfigs as in https://github.com/easybuilders/easybuild-easyconfigs/pull/10246 given that it won't be necessary in the future.

In conclusion, I'm closing this PR.

PR #2002, opened 2020-03-24
Fix TensorFlow in some environments

Flamefire:
In https://github.com/easybuilders/easybuild-easyconfigs/pull/9991 I noticed TensorFlow still uses the local user repository which we usually prevent with `PYTHONNOUSERSITE=1` but (of course) TensorFlow needs it slightly different.

Also 2 minor fixes:
- TF 1.12.1 introduced an experimental download clang option which is disabled by default. Explicitely disable it to avoid the prompt
- AWS support was enabled although it should be disabled. Previously it was `TF_NEED_AWS=0` but no it is an explicit switch.

### DIFF
--- a/easybuild/easyblocks/t/tensorflow.py
+++ b/easybuild/easyblocks/t/tensorflow.py
@@ -440,6 +440,10 @@ class EB_TensorFlow(PythonPackage):
         # https://docs.bazel.build/versions/master/user-manual.html#flag--verbose_failures
         cmd.extend(['--subcommands', '--verbose_failures'])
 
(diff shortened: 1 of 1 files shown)


Flamefire:
This has been tested with https://github.com/easybuilders/easybuild-easyconfigs/pull/9991 (4 TF versions) and 2.1.0 (fosscuda) so this does seemingly work ok

Flamefire:
@boegel Can we have this for 4.2.0 please? The failure is annoying and kinda hard to find as it happens when you e.g. install `tensorflow_datasets` into the user environment

akesandgren:
Going in, thanks @Flamefire!

PR #2005, opened 2020-03-29
make sure libxml2 is built with XZ provided as dependency

boegel:
Just to make sure we don't accidentally pick up the lzma provided by the OS...


akesandgren:
Going in, thanks @boegel!

PR #2006, opened 2020-03-29
take into account user defined gallium_drivers and swr_arches in Mesa easyblock

lexming:
Current mesa easyblock fails the sanity checks if user customizes `gallium-drivers` or `swr-arches` in `configopts`.

This PR fixes that issue by 
1. first retrieve the values for those options from `configopts`
2. `gallium-drivers` is only set by easyblock if missing from `configopts` (as before)
3. `swr-arches` is only set by easyblock if missing from `configopts` **and** `swr` is present in `gallium-drivers`
4. SWR libs are added to sanity checks depending on `swr-arches` and regardless of who defined `swr-arches`

Tested cases:
* default easyconfig of Mesa-20.0.2-GCCcore-9.3.0.eb
   * easyblock behaves as before this PR
* adding `configopts += "-Dgallium-drivers=''"` 
   * the build will fail as at least one driver is missing, but easyblock will respect this configuration if it's set by user
* adding `configopts += "-Dgallium-drivers='swrast'"`
   * no `swr-arches` added and sanity checks do not include any SWR libs
* adding `configopts += "-Dgallium-drivers='swrast,swr'"`
   * `swr-arches` added depending on CPU arch and respective SWR libs checked in sanity checks
* adding `configopts += "-Dgallium-drivers='swrast,swr' -Dswr-arches=avx2"`
   * `swr-arches` is not modified and `libswrAVX2.so` is added to sanity checks

### DIFF
--- a/easybuild/easyblocks/m/mesa.py
+++ b/easybuild/easyblocks/m/mesa.py
@@ -37,6 +37,7 @@ from easybuild.tools.filetools import copy_dir
 from easybuild.tools.systemtools import POWER, X86_64, get_cpu_architecture, get_cpu_features, get_shared_lib_ext
 from easybuild.tools.build_log import EasyBuildError
 
+
 class EB_Mesa(MesonNinja):
     """Custom easyblock for building and installing Mesa."""
 
--- a/easybuild/easyblocks/m/mesa.py
+++ b/easybuild/easyblocks/m/mesa.py
@@ -43,7 +43,7 @@ class EB_Mesa(MesonNinja):
     """Custom easyblock for building and installing Mesa."""
 
     def __init__(self, *args, **kwargs):
-        """Constructor for custom Mesa easyblock: figure out which vales to pass to swr-arches configuration option."""
+        """Constructor for custom Mesa easyblock: figure out which values to pass to swr-arches configuration option."""
 
         super(EB_Mesa, self).__init__(*args, **kwargs)
 
@@ -63,6 +63,8 @@ class EB_Mesa(MesonNinja):
                 gallium_drivers = arch_gallium_drivers[arch]
                 # Add configopt for additional Gallium drivers
                 self.gallium_configopts.append('-Dgallium-drivers=' + ','.join(gallium_drivers))
+
+        self.log.debug('Gallium driver(s) included in the installation: %s' % ', '.join(gallium_drivers))
 
         self.swr_arches = []
 
@@ -84,6 +86,8 @@ class EB_Mesa(MesonNinja):
                 self.swr_arches = sorted([swrarch for feat, swrarch in feat_to_swrarch.items() if feat in cpu_features])
                 # Add configopt for additional SWR arches
                 self.gallium_configopts.append('-Dswr-arches=' + ','.join(self.swr_arches))
+
+            self.log.debug('SWR Gallium driver will support: %s' % ', '.join(self.swr_arches))
 
     def get_configopt_value(self, configopt_name):
         """
--- a/easybuild/easyblocks/m/mesa.py
+++ b/easybuild/easyblocks/m/mesa.py
@@ -36,7 +36,6 @@ from distutils.version import LooseVersion
 from easybuild.easyblocks.generic.mesonninja import MesonNinja
 from easybuild.tools.filetools import copy_dir
 from easybuild.tools.systemtools import POWER, X86_64, get_cpu_architecture, get_cpu_features, get_shared_lib_ext
-from easybuild.tools.build_log import EasyBuildError
 
 
 class EB_Mesa(MesonNinja):
--- a/easybuild/easyblocks/m/mesa.py
+++ b/easybuild/easyblocks/m/mesa.py
@@ -92,17 +92,18 @@ class EB_Mesa(MesonNinja):
         """
         Return list of values for the given configuration option in configopts
         """
-        configopt_value = [opt.split('=')[1] for opt in self.cfg['configopts'].split() if configopt_name in opt]
+        configopt_args = [opt for opt in self.cfg['configopts'].split() if opt.startswith('-D%s=' % configopt_name)]
 
-        if configopt_value:
-            configopt_value = configopt_value[-1].replace("'", '').replace('"', '')
-            configopt_value = configopt_value.split(',')
-
-            if len(configopt_value) > 1:
-                configopt_pair = "%s = %s" % (configopt_name, ', '.join(configopt_value))
-                self.log.warning(
-                    "Found multiple instances of %s in configopts, using last one: %s", configopt_name, configopt_pair
-                )
+        if configopt_args:
+            if len(configopt_args) > 1:
+                self.log.warning("Found multiple instances of %s in configopts, using last one: %s",
+                                 configopt_name, configopt_args[-1])
+            # Get value of last option added
+            configopt_value = configopt_args[-1].split('=')[-1]
+            # Remove quotes and extract individual values
+            configopt_value = configopt_value.strip('"\'').split(',')
+        else:
+            configopt_value = None
 
         return configopt_value
 
--- a/easybuild/easyblocks/m/mesa.py
+++ b/easybuild/easyblocks/m/mesa.py
@@ -28,6 +28,7 @@ EasyBuild support for installing Mesa, implemented as an easyblock
 @author: Andrew Edmondson (University of Birmingham)
 @author: Kenneth Hoste (HPC-UGent)
 @author: Alex Domingo (Vrije Universiteit Brussel)
+@author: Alexander Grund (TU Dresden)
 """
 import os
 from distutils.version import LooseVersion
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/m/mesa.py
```diff
@@ -44,38 +45,61 @@ def __init__(self, *args, **kwargs):
...
+            self.gallium_configopts += ['-Dswr-arches=' + ','.join(self.swr_arches)]
+
+    def get_configopt_value(self, configopt_name):
+        """
+        Return list of values for the given configuration option in configopts
+        """
+        configopt_value = [opt.split('=')[1] for opt in self.cfg['configopts'].split() if configopt_name in opt]
+
+        if configopt_value:
+            configopt_value = configopt_value[0].replace("'", '').replace('"', '')
```
boegel: @lexming Maybe make sure `configopt_value` has exactly one element here, to avoid surprises?

```python
if len(configopt_value) > 1:
    raise EasyBuildError("Found multiple values for %s in configopts: %s", ', '.join(configopt_value))

File: easybuild/easyblocks/m/mesa.py [hunk 2]
```diff
@@ -44,38 +45,61 @@ def __init__(self, *args, **kwargs):
...
+            # Add appropriate Gallium drivers for current architecture
+            arch = get_cpu_architecture()
+            arch_gallium_drivers = {
+                'x86_64': ['swrast', 'swr'],
+                'POWER': ['swrast'],
+            }
+            if arch in arch_gallium_drivers:
+                gallium_drivers = arch_gallium_drivers[arch]
+                # Add configopt for additional Gallium drivers
+                self.gallium_configopts += ['-Dgallium-drivers=' + ','.join(gallium_drivers)]
```
boegel: please use `.append`:

```python
self.gallium_configopts.apend('-Dgallium-drivers=' + ','.join(gallium_drivers))

File: easybuild/easyblocks/m/mesa.py [hunk 3]
```diff
@@ -44,38 +45,61 @@ def __init__(self, *args, **kwargs):
...
+                'avx1.0': 'avx',  # on macOS, AVX is indicated with 'avx1.0' rather than 'avx'
+                'avx2': 'avx2',
+                'avx512f': 'skx',  # AVX-512 Foundation - introduced in Skylake
+                'avx512er': 'knl',  # AVX-512 Exponential and Reciprocal Instructions implemented in Knights Landing
+            }
+            # Determine list of values to pass to swr-arches configuration option
+            cpu_features = get_cpu_features()
+            self.swr_arches = sorted([swrarch for feat, swrarch in feat_to_swrarch.items() if feat in cpu_features])
+            # Add configopt for additional SWR arches
+            self.gallium_configopts += ['-Dswr-arches=' + ','.join(self.swr_arches)]
```
boegel: also use `.append` here please (which extends the existing list, rather than creating a new one)

easybuild/easyblocks/m/mesa.py
```diff
@@ -44,38 +45,61 @@ def __init__(self, *args, **kwargs):
...
-            # Install appropriate Gallium drivers for current architecture
-            if arch == X86_64:
-                self.cfg.update('configopts', "-Dgallium-drivers='swrast,swr'")
-            elif arch == POWER:
-                self.cfg.update('configopts', "-Dgallium-drivers='swrast'")
-
-        if self.swr_arches:
-            self.cfg.update('configopts', '-Dswr-arches=' + ','.join(self.swr_arches))
+        for configopt in self.gallium_configopts:
+            self.cfg.update('configopts', configopt)
```
boegel: `.update` supports list values (also for updating string values like `configopts`, so:

```python
self.cfg.update('configopts', self.gallium_configopts)

File: easybuild/easyblocks/m/mesa.py (hunk 1 above)
Comment (lexming) [reply easybuild/easyblocks/m/mesa.py]: good point, fixed :+1:

File: easybuild/easyblocks/m/mesa.py (hunk 2 above)
Comment (lexming) [reply easybuild/easyblocks/m/mesa.py]: fixed :+1:

File: easybuild/easyblocks/m/mesa.py (hunk 3 above)
Comment (lexming) [reply easybuild/easyblocks/m/mesa.py]: fixed :+1:

File: easybuild/easyblocks/m/mesa.py (hunk 4 above)
Comment (lexming) [reply easybuild/easyblocks/m/mesa.py]: `self.cfg.update()` is awesome! :smile:

**Comment by Flamefire** [comment]
I haven't seen this one when I opened #2007 but it is worth comparing them. What #2007 does better:
- Use the last option from `configopts` when multiple are present
- Handles the case when `swr-arches` was passed but swr is not build. I expect this PR to fail in this case.
- (Less changes)

I don't think `self.gallium_configopts` is actually required. (Same actually for #2007 and `self.swr_arches`) as IMO setting the config params should be done in the config step to make it easier to understand.

**Comment by lexming** [comment]
@Flamefire you are right, this easyblock would have failed with a `configopts` such as `"-Dgallium-drivers='swrast' -Dswr-arches=avx2"`. I've checked your PR and made the following changes:

* `swr-arches` configuration and sanity checks only considered if `swr` is in `gallium-drivers`
* function `get_configopt_value()` always takes last provided value and warns if multiple options were found
* added additional log messages

I testes these changes on Mesa-20.0.2-GCCcore-9.3.0.eb with the configration options listed in the description of https://github.com/easybuilders/easybuild-easyblocks/pull/2006#issue-395239860, plus the following ones:
* `configopts += "-Dgallium-drivers='swrast' -Dswr-arches=avx"`
    * SWR is not built and no sanity checks are added for SWR libs
* `configopts += "-Dswr-arches=avx"`
    * SWR will be built and  SWR libs checked depending on the architecture
* `configopts += "-Dgallium-drivers='swrast,swr' -Dswr-arches=avx -Dgallium-drivers='swrast,swr' -Dswr-arches=avx2"`
    * only last option is taken into account

Regarding the use of `self.gallium_configopts`, it is true that all code sitting in `__init__()` could be moved to `configure_step()` and `self.gallium_configopts` would not be necessary. I do not really mind which option is used, both have their advantages and disadvantages. I'll let this decision to @boegel .

File: easybuild/easyblocks/m/mesa.py [hunk 5]
```diff
@@ -40,42 +42,78 @@ class EB_Mesa(MesonNinja):
...
+            # Add appropriate Gallium drivers for current architecture
+            arch = get_cpu_architecture()
+            arch_gallium_drivers = {
+                'x86_64': ['swrast', 'swr'],
+                'POWER': ['swrast'],
+            }
+            if arch in arch_gallium_drivers:
+                gallium_drivers = arch_gallium_drivers[arch]
+                # Add configopt for additional Gallium drivers
+                self.gallium_configopts.append('-Dgallium-drivers=' + ','.join(gallium_drivers))
```
Flamefire: Why use `self.gallium_configopts` and not `self.cfg.update('configopts', '...')`? Seems cleaner to me and you won't need any class variables, as you can use get_configopt_value in sanity check path. Other classes use that too

easybuild/easyblocks/m/mesa.py
```diff
@@ -40,42 +42,78 @@ class EB_Mesa(MesonNinja):
...
-                # determine list of values to pass to swr-arches configuration option
+                # Determine list of values to pass to swr-arches configuration option
                 cpu_features = get_cpu_features()
                 self.swr_arches = sorted([swrarch for feat, swrarch in feat_to_swrarch.items() if feat in cpu_features])
+                # Add configopt for additional SWR arches
+                self.gallium_configopts.append('-Dswr-arches=' + ','.join(self.swr_arches))
+
+            self.log.debug('SWR Gallium driver will support: %s' % ', '.join(self.swr_arches))
+
+    def get_configopt_value(self, configopt_name):
```
Flamefire: Can you rename to `get_configopt_values`?  Or better: Just copy my function, there are more issues below.

easybuild/easyblocks/m/mesa.py
```diff
@@ -40,42 +42,78 @@ class EB_Mesa(MesonNinja):
...
+        """
+        Return list of values for the given configuration option in configopts
+        """
+        configopt_value = [opt.split('=')[1] for opt in self.cfg['configopts'].split() if configopt_name in opt]
```
Flamefire: Can have false-positives as it matches `configopt_name` anywhere

easybuild/easyblocks/m/mesa.py
```diff
@@ -40,42 +42,78 @@ class EB_Mesa(MesonNinja):
...
+
+        if configopt_value:
+            configopt_value = configopt_value[-1].replace("'", '').replace('"', '')
+            configopt_value = configopt_value.split(',')
+
+            if len(configopt_value) > 1:
```
Flamefire: This is at the wrong place. It shows a warning for `foo=bar,baz` if I'm not mistaken

easybuild/easyblocks/m/mesa.py
```diff
@@ -40,42 +42,78 @@ class EB_Mesa(MesonNinja):
...
-        if 'gallium-drivers' not in self.cfg['configopts']:
-            # Install appropriate Gallium drivers for current architecture
-            if arch == X86_64:
-                self.cfg.update('configopts', "-Dgallium-drivers='swrast,swr'")
-            elif arch == POWER:
-                self.cfg.update('configopts', "-Dgallium-drivers='swrast'")
-
-        if self.swr_arches:
-            self.cfg.update('configopts', '-Dswr-arches=' + ','.join(self.swr_arches))
+        if self.gallium_configopts:
```
Flamefire: This can be removed with suggestion in https://github.com/easybuilders/easybuild-easyblocks/pull/2006/files#diff-1cf1bc7f7b0deed686907a2e5a9e0a96R64

easybuild/easyblocks/m/mesa.py
lexming: I would still do the actual update of `configopts` in the `configure_step()`. I think that it is just more natural.

easybuild/easyblocks/m/mesa.py
lexming: I took most of your function

easybuild/easyblocks/m/mesa.py
lexming: Yep, this is totally misplaced. I fixed it and adapted it to your function.

easybuild/easyblocks/m/mesa.py
Flamefire: But with that change you can remove the whole custom configure step which is also something...
But I can live with that.

easybuild/easyblocks/m/mesa.py
lexming: I made it more strict following what you did.

easybuild/easyblocks/m/mesa.py
lexming: See my reply above

easybuild/easyblocks/m/mesa.py
Flamefire: Looks good, best from both :+1:

The name is still remaining. If it returns values (list) it should be named so, shouldn't it? Just in case we need/have something that doesn't split the value into a list

lexming: I would expect a function called `get_configopt_values` to return all values from `configopts`, not just the last one assigned.

easybuild/easyblocks/m/mesa.py
boegel: It sure is. I know the guy who implemented it, he's awesome too... ;)

boegel:
re-reviewed, tested and approved, thanks @lexming and @Flamefire!

PR #2007, opened 2020-03-30
[Mesa] Adapt sanitity check to actually used options

Flamefire:
When the easyconfig specifies `-Dgallium-drivers=swr-ast` then `swr` won't be built but might be checked for.

This checks configopts for the actually used values and uses those to generate the sanity check

Takes ideas from #2006 (seen after creating this):
- Add a class scoped method to get configure options
- Only add swr-arches if swr is build

Additions over #2006:
- Allow multiple `-DFOO=` in configopts and take last
- Differentiate between unset -DFOO and `-DFOO=''` (empty)
- Check swr-arches in sanity-check step (allows setting swr-arches but not building swr)
- No class scoped variables anymore (local reasoning is easier than tracking where and when variables are set)
- Debug log for what is detected

### DIFF
--- a/easybuild/easyblocks/m/mesa.py
+++ b/easybuild/easyblocks/m/mesa.py
@@ -95,6 +95,17 @@ class EB_Mesa(MesonNinja):
 
     def sanity_check_step(self):
         """Custom sanity check for Mesa."""
+        def get_multi_opt_value(opt_name):
+            """Get list of values given to a multi-value option like -DFOO=bar,baz"""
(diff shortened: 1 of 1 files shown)


Flamefire:
Tested with Mesa-19.2.1-GCCcore-8.3.0.eb. Output:
```
== 2020-03-30 10:08:23,704 mesa.py:130 DEBUG Gallium driver(s) built: ['swrast', 'swr']
== 2020-03-30 10:08:23,704 mesa.py:134 DEBUG SWR gallium driver built for ['avx', 'avx2']
```

Flamefire:
Closed in favor of #2006

PR #2008, opened 2020-04-01
make MATLAB runtime available from MATLAB

smoors:
with this update, the MCR is no longer needed if MATLAB is installed.


akesandgren:
I assume you have tested this?

And sigh.... why can't they learn how to properly package things.... libraries in bin... the stupidity

smoors:
@akesandgren yes, I tested it for 2017a, 2019a, and 2019b.
the test report in https://github.com/easybuilders/easybuild-easyconfigs/pull/10301 was done with this easyblock.

Issue #2009, opened 2020-04-02
root.py sets wrong PYTHON_INCLUDE_DIR and PYTHON_LIBRARY when using Python3


JackPerdue:
So while trying to built ROOT with Python3 (customer request), we see this during the configure_step:

-DPYTHON_INCLUDE_DIR=/sw/eb/sw/Python/3.6.6-foss-2018b/include/python3.6
-DPYTHON_LIBRARY=/sw/eb/sw/Python/3.6.6-foss-2018b/lib/libpython3.6.so

In both cases, the trailing python3.6 should be python3.6m.  This is not "easy" to work around because the easyblock setting overrides any configopts setttings.  So it requires a lot of work after the configure step to tweak the right files to change 3.6 to 3.6m.

I call this a bug.  I find it annoying.  Please fix it.

kelseymh:
As a followup (I'm the customer), it appears that the wrong setting is in [root.py line 84](https://github.com/easybuilders/easybuild-easyblocks/blob/af41c4774c7249511c50e8a53b3c04326d6d2dc6/easybuild/easyblocks/r/root.py#L84), where an example of a "right" solution is in [blender.py line 89](https://github.com/easybuilders/easybuild-easyblocks/blob/af41c4774c7249511c50e8a53b3c04326d6d2dc6/easybuild/easyblocks/b/blender.py#L89).

It seems as though (within th limitations of GitHub's search capability) that there are many instances of the python-setting code block throughout easyblocks.

kelseymh:
I have tested a fix for this issue, by copying the relevant individual lines from blender.py and adapting them into root.py.  I suspect that a proper fix, from the developers, would probably involve pusing the "find_glob_pattern" up to a base class like CMakeMake, or even higher, and maybe dealing with the set of python CMake options in a more encapsulated way, rather than in each individual easyblock.

Anyway, I've attached a diff for root.py showing what I've done.  If the developers would like me to do a pull request, I can, or I can leave it to them to do it right.
[root.py-diff.txt](https://github.com/easybuilders/easybuild-easyblocks/files/4424001/root.py-diff.txt)

JackPerdue:
My customer's Python kung fu is better than mine. :)  Anyway, submitted the easyconfig (also written by the customer) as a PR for reference.

kelseymh:
[...]

PR #2010, opened 2020-04-03
look for Python version directories with suffixes in ROOT easyblock

kelseymh:
Implementation adapted from blender.py.  Addresses issue #2009 .

### DIFF
--- a/easybuild/easyblocks/r/root.py
+++ b/easybuild/easyblocks/r/root.py
@@ -29,6 +29,7 @@ EasyBuild support for ROOT, implemented as an easyblock
 @author: Jens Timmerman (Ghent University)
 """
 from distutils.version import LooseVersion
+import glob
 import os
 
 from easybuild.framework.easyconfig import CUSTOM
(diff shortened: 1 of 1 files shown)


lexming:
Going in, thanks @kelseymh !

mboisson:
Instead of copying `find_glob_pattern`, this function should have been implemented in `filetools.py` and imported in both easyblocks.

kelseymh:
@mboisson You are quite right, of course.  Since this PR is already merged, I will open a new one with that improvement.

mboisson:
I created a PR for it here already 
https://github.com/easybuilders/easybuild-framework/pull/3297

kelseymh:
I'll do that.  Something I don't know how to deal with, though, is handling the cross-repository dependence.  Should I just wait until your [PR #3297](https://github.com/easybuilders/easybuild-framework/pull/3297) gets merged, and then submit mine?  Or is there a way to specify the dependence as part of the PR process?  Sorry for my GitHub ignorance :-/

mboisson:
I don't know that there is an official mechanism. I usually just comment and comment on the PR that this PR requires that PR.

boegel:
> Instead of copying `find_glob_pattern`, this function should have been implemented in `filetools.py` and imported in both easyblocks.

You're very right, and I thought of that, but I had to admit I was being lazy at the time (we were close to the EasyBuild v4.2.0 release, and I wanted this fix included in it).
I figured that if it came up again, or if somebody cared enough, we could still support it in framework later.

Thanks for bringing this up, and helping out!

PR #2012, opened 2020-04-06
re-enable 'apt-get update' + use '-o Acquire::Retries=3' to dance around Azure connectivity issues

boegel:


ocaisa:
Unfortunately there are still failures in here even with the change

boegel:
> Unfortunately there are still failures in here even with the change

Indeed, mentioned that in https://github.com/actions/virtual-environments/issues/675

boegel:
superseded by #2015

PR #2013, opened 2020-04-06
Enable TK, FLTK and OPENGL in VMD config (deps already present)

HugoStrand:
Dear EasyBuilders,

This is a fix for VMD that enables the already included dependencies FLTK, TK and OPENGL (through MESA). Without this fix the TK support is disabled and the extension menu in the GUI is empty.

Tested at C3SE.

Best, Hugo

### DIFF
--- a/easybuild/easyblocks/v/vmd.py
+++ b/easybuild/easyblocks/v/vmd.py
@@ -148,7 +148,8 @@ class EB_VMD(ConfigureMake):
         # PTHREADS: enable support for POSIX threads
         # COLVARS: enable support for collective variables (related to NAMD/LAMMPS)
         # NOSILENT: verbose build command
-        self.cfg.update('configopts', "LINUXAMD64 LP64 IMD PTHREADS COLVARS NOSILENT", allow_duplicate=False)
+        self.cfg.update(
+            'configopts', "LINUXAMD64 LP64 IMD PTHREADS COLVARS NOSILENT FLTK TK OPENGL", allow_duplicate=False)
 
         # add additional configopts based on available dependencies
         for key in deps:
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/v/vmd.py
```diff
@@ -148,7 +148,8 @@ def configure_step(self):
         # PTHREADS: enable support for POSIX threads
         # COLVARS: enable support for collective variables (related to NAMD/LAMMPS)
         # NOSILENT: verbose build command
```
Micket, line 150: Perhaps add descriptions to these additional parameters as well.

HugoStrand: Fixed below.

easybuild/easyblocks/v/vmd.py
```diff
@@ -148,7 +148,11 @@ def configure_step(self):
...
-        self.cfg.update('configopts', "LINUXAMD64 LP64 IMD PTHREADS COLVARS NOSILENT", allow_duplicate=False)
+        # FLTK: enable the standard FLTK GUI
+        # TK: enable TK to support extension GUI elements
+        # OPENGL: enable OpenGL
+        self.cfg.update(
+            'configopts', "LINUXAMD64 LP64 IMD PTHREADS COLVARS NOSILENT FLTK TK OPENGL", allow_duplicate=False)
```
boegel, line 155: Adding FTLK & co should be done conditionally, only if those dependencies are available?

There may be VMD easyconfigs out there that don't have those deps included.

You can use the `get_software_root` function to check whether a dep is provided, there are several examples of that in other easyblocks.

Let us know if you need help with that...

Micket: @boegel they are actually required (see earlier in the file, errors if not included)

Micket: @boegel they are actually required (see earlier in the file, errors if not included)

ocaisa:
@HugoStrand This PR is suffering from the issue that was fixed in #2015, can you please sync it with `develop` so you pull in those changes and the tests pass, then we can merge this.

HugoStrand:
Dear @ocaisa, I have rebased the pull request on the `develop` branch. I hope it can be merged now.

ocaisa:
LGTM
 
Going in, thanks @HugoStrand !

PR #2017, opened 2020-04-09
Added FusionCather Easyblock

Darkless012:


lexming:
Having an easyblock for FusionCatcher will be convenient. But as I said in the PR for the easyconfig, I do **not** like changing version checks in the software
* If the devs are sticking to a certain version of the dependencies and they go as far as putting explicit version checks in the code, we should assume that it must be for a good reason. Even if it is laziness, there might be consequences to upgrading a dependency that we cannot foresee.
* Moreover, users from Easybuild will not be able to guarantee that the results obtained with our FusionCatcher vX.Y are comparable to any other FusionCatcher vX.Y.

Darkless012:
Point taken

PR #2021, opened 2020-04-12
add custom 'subdir_version' easyconfig parameter for FLUENT

boegel:
to deal with divering software version (like '2019R3') vs version used in subdir (like 'v195')


smoors:
@boegel this is ok to me, but I wonder if you can't somehow extract the `subdir_version` from the sources directly?

akesandgren:
@smoors fluent is packedbinary so not much sources there...

akesandgren:
Going in, thanks @boegel!

PR #2027, opened 2020-04-14
use correct version in ELPA preprocessor flag used for QuantumESPRESSO v6.x

Darkless012:
Quantum Espresso EasyBlock always includes `dflags.append('-D__ELPA_2016')` for QE version >= 6, but it doesn't reflects current versions of ELPA, which made breaking changes in v 2017.11.

This results in inability to compile current version.

PR for Easyconfig: https://github.com/easybuilders/easybuild-easyconfigs/pull/10393

### DIFF
--- a/easybuild/easyblocks/q/quantumespresso.py
+++ b/easybuild/easyblocks/q/quantumespresso.py
@@ -40,7 +40,7 @@ import easybuild.tools.toolchain as toolchain
 from easybuild.easyblocks.generic.configuremake import ConfigureMake
 from easybuild.framework.easyconfig import CUSTOM
 from easybuild.tools.build_log import EasyBuildError
-from easybuild.tools.filetools import apply_regex_substitutions, copy_dir, copy_file
+from easybuild.tools.filetools import copy_dir, copy_file
 from easybuild.tools.modules import get_software_root, get_software_version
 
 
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/q/quantumespresso.py
```diff
@@ -134,15 +134,25 @@ def configure_step(self):
...
-                dflags.append('-D__ELPA_2016')
             else:
                 elpa_min_ver = "2015"
                 dflags.append('-D__ELPA_2015 -D__ELPA')
 
             if LooseVersion(elpa_v) < LooseVersion(elpa_min_ver):
-                raise EasyBuildError("QuantumESPRESSO %s needs ELPA to be " +
-                                     "version %s or newer" % (self.version, elpa_min_ver))
+                raise EasyBuildError(("QuantumESPRESSO %s needs ELPA to be " +
+                                     "version %s or newer") % (self.version, elpa_min_ver))
```
boegel, line 155: @Darkless012 You should be able to simplify this to:

```python
                raise EasyBuildError("QuantumESPRESSO %s needs ELPA to be " +
                                     "version %s or newer", self.version, elpa_min_ver)

boegel:
Tested with https://github.com/easybuilders/easybuild-easyconfigs/pull/10393 + existing `QuantumESPRESSO` easyconfigs, good to in, thanks @Darkless012!

PR #2028, opened 2020-04-14
sync develop with master after EasyBuild v4.2.0 release + bump version to 4.2.1dev

boegel:


easybuild/easyblocks/__init__.py
```diff
@@ -43,7 +43,7 @@
 # recent setuptools versions will *TRANSFORM* something like 'X.Y.Zdev' into 'X.Y.Z.dev0', with a warning like
 #   UserWarning: Normalizing '2.4.0dev' to '2.4.0.dev0'
 # This causes problems further up the dependency chain...
-VERSION = LooseVersion('4.1.2.dev0')
+VERSION = LooseVersion('4.2.0.dev0')
```
Micket, line 46: So, just checking if this is correct; should it be 4.2.0.dev0 here?

boegel: Ah, no, should be `4.2.1.dev0`, of course...

boegel: fixed with force push in e7e45bc44

PR #2029, opened 2020-04-16
generically find the arch folder and add top-level bin/lib symlinks in PDT & TAU easyblocks

Flamefire:
The arch folder is not $(uname -m), but determined by some custom if-chain
So short of parsing the configure output we simply symlink the bin/lib folder
of any folder containing those into the top-level folder.
This is what Spack does too so it is verified already

This fixes build on e.g. Power9

### DIFF
--- a/easybuild/easyblocks/p/pdt.py
+++ b/easybuild/easyblocks/p/pdt.py
@@ -14,6 +14,7 @@ EasyBuild support for building and installing PDT, implemented as an easyblock
 
 @author Bernd Mohr (Juelich Supercomputing Centre)
 @author Markus Geimer (Juelich Supercomputing Centre)
+@author Alexander Grund (TU Dresden)
 """
 
 import os
@@ -21,30 +22,36 @@ import os
 from easybuild.easyblocks.generic.configuremake import ConfigureMake
 import easybuild.tools.toolchain as toolchain
 from easybuild.tools.build_log import EasyBuildError
-from easybuild.tools.run import run_cmd
+
+
+def find_arch_dir(install_dir):
+    """Find architecture directory inside the install tree
+
+    For simplicity any top-level folder containing a bin and lib directory is collected
+    Raises an error if multiple or none matching folders are found
+    Returns the architecture specific folder
+    """
+    arch_dirs = []
+    for entry in os.listdir(install_dir):
+        full_path = os.path.join(install_dir, entry)
+        # Only check directories which are not symlinks
+        # E.g. on x86_64 there are symlinks craycnl, mic_linux, ... to x86_64
+        if not os.path.isdir(full_path) or os.path.islink(full_path):
+            continue
+        bin_dir = os.path.join(full_path, 'bin')
+        lib_dir = os.path.join(full_path, 'lib')
+        if all(os.path.isdir(dir) for dir in (bin_dir, lib_dir)):
+            arch_dirs.append(full_path)
+
+    if not arch_dirs:
+        raise EasyBuildError('Architecture specific directory not found in %s' % install_dir)
+    elif len(arch_dirs) != 1:
+        raise EasyBuildError('Found multiple architecture specific directories in: %s' % arch_dirs)
+    return arch_dirs[0]
 
 
 class EB_PDT(ConfigureMake):
     """Support for building/installing PDT."""
-
-    def __init__(self, *args, **kwargs):
-        """Initialisation of custom class variables for PDT."""
-        super(EB_PDT, self).__init__(*args, **kwargs)
-
-        out, _ = run_cmd("uname -m", simple=False)
-        self.machine = out.strip()
-        self.log.info("Using '%s' as machine label", self.machine)
-
-    def prepare_step(self, *args, **kwargs):
-        """Custom prepare step for PDT."""
-        super(EB_PDT, self).prepare_step(*args, **kwargs)
-
-        # create install directory and make sure it does not get cleaned up again in the install step;
-        # the first configure iteration already puts things in place in the install directory,
-        # so that shouldn't get cleaned up afterwards...
-        self.log.info("Creating install dir %s before starting configure-build-install iterations", self.installdir)
-        self.make_installdir()
-        self.cfg['keeppreviousinstall'] = True
 
     def configure_step(self):
         """Custom configuration procedure for PDT."""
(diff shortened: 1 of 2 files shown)


Flamefire:
Tested in https://github.com/easybuilders/easybuild-easyblocks/pull/2029 with success

easybuild/easyblocks/p/pdt.py
```diff
@@ -57,36 +64,53 @@ def configure_step(self):
...
-        else:
+        except KeyError:
             raise EasyBuildError("Compiler family not supported yet: %s" % comp_fam)
         self.cfg.update('configopts', compiler_opt)
 
+        if self.toolchain.options['pic']:
+            self.cfg.update('-useropt=-fPIC')
+
+        # Configure creates required subfolders in installdir, so create first
+        super(EB_PDT, self).make_installdir()
```
boegel, line 80: This call to `make_installdir` should be guarded, to ensure it's only done once in case of an iterated installation (for example when the `configopts` specified in the PDT easyconfig file is a list of string values, rather than a single string value).
The previous approach via `keeppreviousinstall` did that sort of too, but caused problems with a forced reinstallation, so this approach is better as long as the guard is there:

```python
# create installation directory (but only once, during first iteration)
if self.iter_idx == 0:
    super(EB_PDT, self).make_installdir()

File: easybuild/easyblocks/p/pdt.py [hunk 2]
```diff
@@ -14,37 +14,44 @@
...
 from easybuild.easyblocks.generic.configuremake import ConfigureMake
 import easybuild.tools.toolchain as toolchain
 from easybuild.tools.build_log import EasyBuildError
-from easybuild.tools.run import run_cmd
 
 
-class EB_PDT(ConfigureMake):
-    """Support for building/installing PDT."""
+def find_arch_dir(install_dir):
+    """Find architecture directory inside the install tree
```
boegel, line 28: Please change to (separate lines):

```python
"""
Find architecture directory inside the install tree.

File: easybuild/easyblocks/p/pdt.py [hunk 3]
```diff
@@ -14,37 +14,44 @@
...
+    for entry in os.listdir(install_dir):
+        full_path = os.path.join(install_dir, entry)
+        # Only check directories which are not symlinks
+        # E.g. on x86_64 there are symlinks craycnl, mic_linux, ... to x86_64
+        if not os.path.isdir(full_path) or os.path.islink(full_path):
+            continue
+        bin_dir = os.path.join(full_path, 'bin')
+        lib_dir = os.path.join(full_path, 'lib')
+        if all(os.path.isdir(dir) for dir in (bin_dir, lib_dir)):
+            arch_dirs.append(full_path)
```
boegel, line 44: bit shorter, since `bin_dir` or `lib_dir` are not used elsewhere:

```python
# only retain paths that both have a 'bin' and 'lib' subdirectory
if all(os.path.isdir(os.path.join(full_path, subdir)) for subdir in ['bin', 'lib']):

File: easybuild/easyblocks/p/pdt.py [hunk 4]
```diff
@@ -14,37 +14,44 @@
...
 
-        out, _ = run_cmd("uname -m", simple=False)
-        self.machine = out.strip()
-        self.log.info("Using '%s' as machine label", self.machine)
+    if not arch_dirs:
+        raise EasyBuildError('Architecture specific directory not found in %s' % install_dir)
```
boegel, line 47:
```python
raise EasyBuildError("No architecture specific directories not found in %s", install_dir)
```

(mostly about the ` %` -> `,`; for strings containing spaces, we usually use double quotes)

easybuild/easyblocks/p/pdt.py
```diff
@@ -14,37 +14,44 @@
...
+    elif len(arch_dirs) != 1:
+        raise EasyBuildError('Found multiple architecture specific directories in: %s' % arch_dirs)
```
boegel, line 49:
```python
        raise EasyBuildError("Found multiple architecture specific directories in: %s", ', '.join(arch_dirs))

File: easybuild/easyblocks/p/pdt.py [hunk 6]
```diff
@@ -57,36 +64,53 @@ def configure_step(self):
...
             toolchain.GCC: '-GNU',
             toolchain.INTELCOMP: '-icpc',
+            toolchain.PGI: '-pgCC',
         }
         comp_fam = self.toolchain.comp_family()
-        if comp_fam in known_compilers:
+        try:
             compiler_opt = known_compilers[comp_fam]
-        else:
+        except KeyError:
```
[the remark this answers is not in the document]
Flamefire: Just some DRY cleaning: No need to mention the key twice. Bonus: It's faster

easybuild/easyblocks/p/pdt.py
Flamefire: Thanks. Could swear I tried that before but got over the line length. Probably didn't had the list comprehension back then

easybuild/easyblocks/t/tau.py
```diff
@@ -237,32 +232,39 @@ def build_step(self):
...
+    def install_step(self):
+        """Create symlinks into arch-specific directories"""
+        super(EB_TAU, self).install_step()
+        # Link arch-specific directories into prefix
+        arch_dir = find_arch_dir(self.installdir)
+        self.log.debug('Found %s as architecture specific directory. Creating symlinks...', arch_dir)
+        for d in ('bin', 'lib'):
+            src = os.path.join(arch_dir, d)
+            dst = os.path.join(self.installdir, d)
+            if os.path.exists(dst):
```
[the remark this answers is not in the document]
Flamefire: I changed it to `lexists` which also checks for broken symlinks. The message was changed to info instead of debug.

If it is added and is correct all is good, sanity check covers the case where it points/contains something wrong.

easybuild/easyblocks/p/pdt.py
```diff
@@ -57,36 +64,53 @@ def configure_step(self):
...
+        if self.cfg['parallel']:
+            self.cfg.update('installopts', '-j %s' % self.cfg['parallel'])
+        super(EB_PDT, self).install_step()
+        # Link arch-specific directories into prefix
+        arch_dir = find_arch_dir(self.installdir)
+        self.log.debug('Found %s as architecture specific directory. Creating symlinks...', arch_dir)
+        for d in ('bin', 'lib'):
+            src = os.path.join(arch_dir, d)
+            dst = os.path.join(self.installdir, d)
+            if os.path.exists(dst):
```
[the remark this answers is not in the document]
Flamefire: see https://github.com/easybuilders/easybuild-easyblocks/pull/2029#discussion_r411159962

easybuild/easyblocks/p/pdt.py
Flamefire: Done here and in TAU

easybuild/easyblocks/p/pdt.py
Flamefire: >  for strings containing spaces, we usually use double quotes

Hm? What is the reasoning?   
For reference: https://stackoverflow.com/a/144163/1930508 and I agree: Using conventions when to use which is confusing ;)

Flamefire:
Resolved discussions with changes implemented as requested @boegel Please double-check the open ones.

Flamefire:
FWIW: Score-P and PDT build successfully with this. For TAU we only have an ancient EC that doesn't even work out of the box. But fixing the unrelated errors and omitting OTF (that version of TAU uses wrong names) it also builds.

So I assume that no regressions are introduced

easybuild/easyblocks/p/pdt.py
boegel: It's more of a personal habit, we don't have strict code stye rules on this one.

The only way to avoid discussions like this is to adopt [Black](https://github.com/psf/black), but it's a significant effort and it formats some things in questionable/weird ways, which make me reluctant to switch to it.

boegel:
Re-reviewed & tested with existing PDT easyconfigs, good to go, thanks @Flamefire!

PR #2032, opened 2020-04-16
let CMakeMake remove 'easybuild_obj' build directory if it already exists

Flamefire:
without also ensuring that the directories that are used for that build are also unique
there are build problems with bundles that contain more than one CMake package and didn’t
explicitly set separate_build_dir to false

This also applies partially to multiple iterations of the same software build (e.g. configopts)

Reported in https://github.com/easybuilders/easybuild-easyblocks/pull/1933#issuecomment-614642219

As the build directory is unique to EasyBuild it is safe to remove

### DIFF
--- a/easybuild/easyblocks/generic/cmakemake.py
+++ b/easybuild/easyblocks/generic/cmakemake.py
@@ -39,7 +39,7 @@ from easybuild.easyblocks.generic.configuremake import ConfigureMake
 from easybuild.framework.easyconfig import CUSTOM
 from easybuild.tools.build_log import print_warning
 from easybuild.tools.config import build_option
-from easybuild.tools.filetools import change_dir, mkdir, which
+from easybuild.tools.filetools import change_dir, mkdir, which, remove_dir
 from easybuild.tools.environment import setvar
 from easybuild.tools.modules import get_software_root
 from easybuild.tools.run import run_cmd
@@ -120,6 +120,12 @@ class CMakeMake(ConfigureMake):
 
         if builddir is None and self.cfg.get('separate_build_dir', True):
             builddir = os.path.join(self.builddir, 'easybuild_obj')
+            # For separate_build_dir we want a clean folder. So remove if it exists
+            # This can happen when multiple iterations are done (e.g. shared, static, ...)
+            if os.path.exists(builddir):
+                self.log.warning('Build directory %s already exists (from previous iterations?). Removing...',
+                                 builddir)
+                remove_dir(builddir)
 
         if builddir:
             mkdir(builddir, parents=True)


easybuild/easyblocks/generic/cmakemake.py
```diff
@@ -120,6 +120,12 @@ def configure_step(self, srcdir=None, builddir=None):
 
         if builddir is None and self.cfg.get('separate_build_dir', True):
             builddir = os.path.join(self.builddir, 'easybuild_obj')
```
boegel, line 122: @Flamefire Why not try and ensure a unique subdir here instead?

By using `self.name`, `self.version`, and may even an additional part with random chars (look for "salt", we do this in other places)?

Flamefire: Because I'm lazy and this is easier.
And also because some easyconfigs rely on this name already. See SimpleElastix, SImpleITK, VV, xtb, pyGIMLi (grep for the folder name, not checked that those use CMake)

But anyway: Some ECs may need the actual build folder name in their pre/post commands so having it deterministic is a good thing. Randomizing the name would require at least a new template for that name to be used in ECs

Flamefire:
@boegel 
Analysis of solutions:

1. Remove build directory if exists: Impossible to inspect build files after another package reused it, e.g in bundles. But this is rare
1. Use random name: Breaks ECs relying on `easybuild_obj` 
1. Encode name and other things (e.g. version) into the folder: Breaks ECs relying on `easybuild_obj` as the name, see https://github.com/easybuilders/easybuild-easyblocks/pull/2032#discussion_r409557691
1. Add add an additional EC parameter: `unique_build_dir`. It will start with `%(builddir)s/build_%(name)s` and append increasing numbers until it finds one that doesn't exist yet. We don't need to append the version number as you won't build 2 softwares with the same name in 1 Bundle, will you? I don't know how to get the current iteration number but the above will "usually" correlate well enough with the iteration number.
1. Build in subfolder of the source directory: Usually done in "generic workflow": `mkdir build && cd build; cmake .. <other opts>; make && make install`, but also breaks existing ECs, can lead to long folder names and folder must be unique to (but could be `easybuild_obj` which should be unique enough)

easybuild/easyblocks/generic/cmakemake.py
```diff
@@ -120,6 +120,12 @@ def configure_step(self, srcdir=None, builddir=None):
...
+            # For separate_build_dir we want a clean folder. So remove if it exists
```
LittleUker, line 123:
```suggestion
            # For separate_build_dir we want a clean folder. So remove if it exists
```

boegel:
Merging this for inclusion in EasyBuild v4.2.1.

As discussed with @Flamefire, this improves the current situation (it fixes some use cases), even though it may not be perfect and cover every possible use case.

We can follow up in another PR (cfr. https://github.com/easybuilders/easybuild-easyblocks/issues/2036#issuecomment-630910979).

PR #2033, opened 2020-04-16
Use config.guess for GCC builds

Flamefire:
Building of some older GCCs (tried 4.4) fails on e.g. Power due to config.guess being to old.

This is a solved problem in ConfigureMake by using an up-to-date one but GCC EB didn't use it yet as it doesn't call the super-configure method.

Additonally it ignores the build_type and host_type EC variables.

This is fixed with this PR by factoring out a method from ConfigureMake which can then be called by GCC

### DIFF
--- a/easybuild/easyblocks/g/gcc.py
+++ b/easybuild/easyblocks/g/gcc.py
@@ -392,7 +392,7 @@ class EB_GCC(ConfigureMake):
             config_guess = self.config_guess or '../config.guess'
             out, ec = run_cmd("../config.guess")
             if ec != 0:
-                raise EasyBuildError('Failed to optain platform_lib value from %s', config_guess)
+                raise EasyBuildError('Failed to obtain platform_lib value from %s', config_guess)
             self.platform_lib = out.strip()
 
         # IV) actual configure, but not on default path
--- a/easybuild/easyblocks/g/gcc.py
+++ b/easybuild/easyblocks/g/gcc.py
@@ -381,18 +381,20 @@ class EB_GCC(ConfigureMake):
         else:
             objdir = self.create_dir("obj")
 
-        build_and_host_options = self.determine_build_and_host_options()
-        configopts += ' ' + ' '.join(build_and_host_options)
+        build_type, host_type = self.determine_build_and_host_type()
+        if build_type:
+            configopts += ' --build=' + build_type
+        if host_type:
+            configopts += ' --host=' + host_type
 
         # instead of relying on uname, we use the actual --build option
-        if build_and_host_options and build_and_host_options[0].startswith('--build='):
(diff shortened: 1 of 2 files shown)


easybuild/easyblocks/g/gcc.py
```diff
@@ -383,14 +381,25 @@ def configure_step(self):
         else:
             objdir = self.create_dir("obj")
 
+        build_and_host_options = self.determine_build_and_host_options()
+        configopts += ' ' + ' '.join(build_and_host_options)
+
+        # instead of relying on uname, we use the actual --build option
+        if build_and_host_options and build_and_host_options[0].startswith('--build='):
+            self.platform_lib = build_and_host_options[0].split('=', 1)[1]
```
smoors, line 389:
```suggestion
        build_type_option = [x for x in build_and_host_options if x.startswith('--build=')]
        if len(build_type_option) == 1:
            self.platform_lib = build_type_option[0].split('=', 0)
```
make this a bit more robust? or alternatively, let determine_build_and_host_options() return both build and host type options as separate variables?

easybuild/easyblocks/g/gcc.py
```diff
@@ -383,14 +381,25 @@ def configure_step(self):
...
+        else:
+            # Fallback
+            config_guess = self.config_guess or '../config.guess'
+            out, ec = run_cmd("../config.guess")
+            if ec != 0:
+                raise EasyBuildError('Failed to optain platform_lib value from %s', config_guess)
```
smoors, line 395:
```suggestion
                raise EasyBuildError('Failed to obtain platform_lib value from %s', config_guess)
```

easybuild/easyblocks/g/gcc.py
Flamefire: Good point, it was meant to be implicit as suggested by the name but explicit is better: https://github.com/easybuilders/easybuild-easyblocks/pull/2033/commits/7dc3d2cf275e3099e788651d818141b2bd673543

easybuild/easyblocks/g/gcc.py
```diff
@@ -383,14 +381,27 @@ def configure_step(self):
...
+            out, ec = run_cmd("../config.guess")
+            if ec != 0:
+                raise EasyBuildError('Failed to obtain platform_lib value from config.guess')
+            self.platform_lib = out.strip()
+
         # IV) actual configure, but not on default path
         cmd = "../configure  %s %s" % (self.configopts, configopts)
 
         # instead of relying on uname, we run the same command GCC uses to
         # determine the platform
```
akesandgren, line 404: This comment seem misplaced now

Flamefire: Ah, right. Removed

akesandgren:
Going in, thanks @Flamefire!

PR #2034, opened 2020-04-16
update ABAQUS easyblock to support installation of v2020

branfosj:
There are some differences to where some items end up being installed. The items that were in `sim` are now in `cae` and I could not see a way to separate them from the items that are installed into `cae`. (This is compared to an ABAQUS 2017 installation - we do not have a newer version installed.)


easybuild/easyblocks/a/abaqus.py
```diff
@@ -90,26 +91,46 @@ def install_step(self):
...
-                r"location of your existing ANSA installation.*(\n.*){8}:": '',
-                r"FLUENT Path.*(\n.*){7}:": '',
-            }
-            run_cmd_qa('./StartTUI.sh', qa, no_qa=no_qa, std_qa=std_qa, log_all=True, simple=True, maxhits=100)
+            if LooseVersion(self.version) >= LooseVersion('2020'):
+                std_qa = {
+                    # disable installation of Isight (7) and enable installation of documentation (3)
+                    r"Isight\n\nEnter selection \(default: Next\):": '3\n7\n\n',
+                    # No Tosca (6 and 7), but do not trigger answering a later question that has no Tosca options
+                    r"21 \[\*\] Abaqus/CFD Solver\n\nEnter selection \(default: Next\):": '7\n8\n\n',
```
smoors, line 99: @branfosj the comment says 6 and 7, but the input is 7 and 8 ?

branfosj: The comment is wrong - I'll correct it.

For reference, this is the list in question:
```
 -1 Select All
  0 Select None
  1 [*] Abaqus/Standard Solver
  2 [*] Abaqus/Explicit Solver
  3 [*] Cosimulation Services
  4 [*] Abaqus ODB API Services
  5 [*] Abaqus CAE
  6 [*] Abaqus Samples
  7 [*] Tosca Structure
  8 [*] Tosca Fluid
  9 [*] fe-safe
 10 [*] fe-safe interface to Abaqus 2019 ODBs
 11 [*] fe-safe interface to Abaqus 2018 ODBs
 12 [*] fe-safe interface to Abaqus 2017 ODBs
 13 [*] fe-safe interface to Abaqus 2016 ODBs
 14 [*] fe-safe materials and surface finish specifications
 15 [*] Material database server for fe-safe
 16 [*] fe-safe tutorial and sample data
 17 [*] fe-safe tutorial models for Abaqus
 18 [*] fe-safe tutorial models for I-DEAS
 19 [*] fe-safe tutorial models for Ansys
 20 [*] fe-safe tutorial models for NASTRAN
 21 [*] Abaqus/CFD Solver
```

branfosj:
I've uploaded https://gist.github.com/branfosj/f0bfa2410f7279e5c2d0dfb58bc3c2e6 which is the output of the ABAQUS installer for one of my test runs. This means that we have a record of the path through the installer.

easybuild/easyblocks/a/abaqus.py
```diff
@@ -90,26 +91,46 @@ def install_step(self):
...
+                    r"Isight\nEnter selection \(default: Next\):": '6\n7\n\n',
+                    r"(?<!Isight)\nEnter selection \(default: Next\):": '',
+                    r"SIMULIA[0-9]*doc.*:": os.path.join(self.installdir, 'doc'),
+                    r"SimulationServices.*:": os.path.join(self.installdir, 'sim'),
+                    r"Choose the CODE installation directory.*:\n.*\n\n.*:": os.path.join(self.installdir, 'sim'),
+                    r"SIMULIA/CAE.*:": os.path.join(self.installdir, 'cae'),
+                    r"location of your Abaqus services \(solvers\).*(\n.*){8}:\s*": os.path.join(self.installdir,
+                                                                                                 'sim'),
+                    r"Default.*SIMULIA/Commands\]:\s*": os.path.join(self.installdir, 'Commands'),
+                    r"Default.*SIMULIA/CAE/plugins.*:\s*": os.path.join(self.installdir, 'plugins'),
```
boegel, line 124: @branfosj A lot of these patterns are common, can we flesh out those outside of the `if`/`else`?

easybuild/easyblocks/a/abaqus.py
boegel: We should include the `7 .*Tosca Structure` and `8 .*Tosca Fluid` in the question pattern, to be robust against changing numbering going forward?

You can even grab then number from the question, like we do in the WRF easyblock

branfosj:
https://github.com/easybuilders/easybuild-easyblocks/pull/2034/commits/df50973ed449bc1be0838bd0128df0c4f693b0b1 combines the common questions while keeping the `if`/`else`.

https://github.com/easybuilders/easybuild-easyblocks/pull/2034/commits/ea3cc122b8fa8a673a35b1099042bc606a87b650 and https://github.com/easybuilders/easybuild-easyblocks/pull/2034/commits/24afc94fec8ce4404b3ad142611e921c2b785acd pull the answer from parsing the question. This proved more complicated as we have to avoid matching the first view of a question. I spent a good while with `std_qa` attempts where I got stuck turning options on and then off repeatedly.

easybuild/easyblocks/a/abaqus.py
branfosj: I've changed it so that we have a more robust method that does not require an `if`/`else` block at all.

easybuild/easyblocks/a/abaqus.py
branfosj: Yes. I've changed it so that we pull the answer from the question for these blocks.

branfosj:
Test report using the latest version of the block is at https://github.com/easybuilders/easybuild-easyconfigs/pull/10423#issuecomment-636454266

boegel:
close/re-open to refresh CI (we now have a stricter code style check)

boegel:
@branfosj This breaks the installation of `ABAQUS-2018-hotfix-1806.eb`:

It looks like it chokes on this now, not being able to answer this question:

```
_______________________________________________________________________________

   SIMULIA fe-safe 2018 fe-safe
_______________________________________________________________________________

Choose the installation directory:
Input !c to clear the default value. Input !q and hit Enter to quit. Just press Enter to continue.                                                                                                                                                                                                                                                                                                                                                                                                                                                              
Default [/usr/SIMULIA/fe-safe/2018]:
```

easybuild/easyblocks/a/abaqus.py
```diff
@@ -102,12 +121,17 @@ def install_step(self):
...
                 r"location of your existing ANSA installation.*(\n.*){8}:": '',
                 r"FLUENT Path.*(\n.*){7}:": '',
+
+                # License server
+                r"License Server [0-9]+\s*(\n.*){3}:": 'abaqusfea',  # bypass value for license server
+                r"License Server . \(redundant\)\s*(\n.*){3}:": '',
+                r"Please choose an action:": '1',
+
+                # Continue
+                nextstr: '',
```
boegel, line 134: @branfosj One thing to be aware of here is that there's no order over these patterns (because it's a `dict` value).

So if the pattern in `nextstr` is matched, you won't match any other more specific patterns that also include the `nextstr` pattern...

mboisson:
What is the status of this PR ?

branfosj:
> What is the status of this PR ?

It works for me for both ABAQUS 2017 and 2020 (these are the versions I have sources for). However, it is broken installing ABAQUS 2018 hotfix 1806 (and maybe other versions). The broken installs may be because of the issue in https://github.com/easybuilders/easybuild-easyblocks/pull/2034#discussion_r484750829

Currently I do not have the time to investigate further.

PR #2035, opened 2020-04-19
LAMMPS: add support to handle Atomistica

lexming:
Adding the inter-atomic potentials from [Atomistica](https://github.com/Atomistica/atomistica) into LAMMPS requires not only linking to its library `libatomistica` but also building LAMMPS with additional source code files.
This PR updates the easyblock of `lammps` to detect the dependency on Atomistica and if it's present perform all needed steps to build LAMMPS including its potentials.

Tests can be done with `LAMMPS-7Aug2019-foss-2019b-Python-3.7.4-kokkos-atomistica.eb` from https://github.com/easybuilders/easybuild-easyconfigs/pull/10433, which is basically a regular easyconfig of LAMMPS with the added dependency on Atomistica. All needed files and libraries are provided by `Atomistica-0.10.1-foss-2019b-Python-3.7.4.eb` (also available in https://github.com/easybuilders/easybuild-easyconfigs/pull/10433).

### DIFF
--- a/easybuild/easyblocks/l/lammps.py
+++ b/easybuild/easyblocks/l/lammps.py
@@ -123,7 +123,7 @@ class EB_LAMMPS(CMakeMake):
 
         # Add files from Atomistica
         atomistica = get_software_root('Atomistica')
-        if atomistica:
+        if atomistica and 'ATOMISTICA' in self.cfg['user_packages']:
             # Package files
             atomistica_files = ['pair_atomistica.%s' % e for e in ['cpp', 'h']]
             self.log.info("Adding Atomistica package files into LAMMPS: %s" % ', '.join(atomistica_files))
@@ -273,7 +273,7 @@ class EB_LAMMPS(CMakeMake):
 
         # Atomistica custom sanity checks
         atomistica = get_software_root('Atomistica')
-        if atomistica:
+        if atomistica and 'ATOMISTICA' in self.cfg['user_packages']:
             custom_commands += [
                 "cd %s && " % os.path.join(self.start_dir, 'examples/atomistica') +
                 """python -c 'from lammps import lammps; l=lammps(); l.file("atomistica.in")'"""
--- a/easybuild/easyblocks/l/lammps.py
+++ b/easybuild/easyblocks/l/lammps.py
@@ -41,7 +41,7 @@ from easybuild.tools.config import build_option
 from easybuild.tools.modules import get_software_root, get_software_version
 from easybuild.tools.run import run_cmd
 from easybuild.tools.systemtools import get_shared_lib_ext
-from easybuild.tools.filetools import copy_file
+from easybuild.tools.filetools import copy_file, copy_dir
 
 from easybuild.easyblocks.generic.cmakemake import CMakeMake
 
@@ -121,13 +121,16 @@ class EB_LAMMPS(CMakeMake):
         if self.cfg['kokkos'] and get_software_root('CUDA'):
             env.unset_env_vars(['LIBS'])
 
-        # Add package files from Atomistica
-        if get_software_root('Atomistica'):
-            atomistica = get_software_root('Atomistica')
+        # Add files from Atomistica
+        atomistica = get_software_root('Atomistica')
+        if atomistica:
+            # Package files
             atomistica_files = ['pair_atomistica.%s' % e for e in ['cpp', 'h']]
-            self.log.info("Adding Atomistica package files into LAMMPS source: %s" % ','.join(atomistica_files))
+            self.log.info("Adding Atomistica package files into LAMMPS: %s" % ', '.join(atomistica_files))
             for pkgfile in atomistica_files:
                 copy_file(os.path.join(atomistica, 'share/lammps', pkgfile), os.path.join('src', pkgfile))
+            # Test for sanity checks
+            copy_dir(os.path.join(atomistica, 'share/examples/liquid_silicon'), 'examples/atomistica')
 
     def configure_step(self, **kwargs):
         """Custom configuration procedure for LAMMPS."""
@@ -280,10 +283,18 @@ class EB_LAMMPS(CMakeMake):
             # LAMMPS test - you need to call specific test file on path
             """python -c 'from lammps import lammps; l=lammps(); l.file("%s")'""" %
             # The path is joined by "build_dir" (start_dir)/examples/filename/in.filename
-            os.path.join(self.start_dir, "examples", "%s" % check_file, "in.%s" % check_file)
+            os.path.join(self.start_dir, 'examples', '%s' % check_file, 'in.%s' % check_file)
             # And this should be done for every file specified above
             for check_file in check_files
         ]
+
+        # Atomistica custom sanity checks
+        atomistica = get_software_root('Atomistica')
+        if atomistica:
+            custom_commands += [
+                "cd %s && " % os.path.join(self.start_dir, 'examples/atomistica') +
+                """python -c 'from lammps import lammps; l=lammps(); l.file("atomistica.in")'"""
+            ]
 
         shlib_ext = get_shared_lib_ext()
         custom_paths = {
--- a/easybuild/easyblocks/l/lammps.py
+++ b/easybuild/easyblocks/l/lammps.py
@@ -225,21 +225,29 @@ class EB_LAMMPS(CMakeMake):
                 self.cfg.update('configopts', '-DFFT_PACK=array')
 
         # Additional linker flags
-        lammps_link_libs = []
+        lammps_linklibs = []
 
         # Atomistica
         if get_software_root('Atomistica'):
-            lammps_link_libs += ['-latomistica']
+            # link with libatomistica
+            lammps_linklibs += ['-latomistica']
+            # Atomistica needs the MPI Fortran bindings, which are not added
+            # by default in LAMMPS as it uses a C++ linker
             if self.toolchain.options.get('usempi', None):
                 if self.toolchain.mpi_family() == toolchain.OPENMPI:
-                    lammps_link_libs += ['-lmpi_usempif08', '-lmpi_usempi_ignore_tkr', '-lmpi_mpifh', '-lmpi']
+                    # retrieve libraries used to compile OpenMPI
+                    out, ec = run_cmd("mpifort --showme:libs", simple=False)
+                    if ec:
+                        raise EasyBuildError("Failed to determine OpenMPI libraries: %s", out)
+                    lammps_linklibs += ["-l%s" % lib for lib in out.split()]
                 elif self.toolchain.mpi_family() == toolchain.INTELMPI:
-                    lammps_link_libs += ['-lmpifort', '-lmpi', '-lmpigi']
+                    # add Fortran libraries used by Intel MPI
+                    lammps_linklibs += ['-lmpifort', '-lmpi']
 
         # Set additional linker flags
-        if len(lammps_link_libs) > 0:
-            cmake_lammps_link_libs = '-DLAMMPS_LINK_LIBS="%s"' % ' '.join(lammps_link_libs)
-            self.cfg.update('configopts', cmake_lammps_link_libs)
+        if len(lammps_linklibs) > 0:
+            cmake_lammps_linklibs = '-DLAMMPS_LINK_LIBS="%s"' % ' '.join(lammps_linklibs)
+            self.cfg.update('configopts', cmake_lammps_linklibs)
 
         # https://lammps.sandia.gov/doc/Build_extras.html
         # KOKKOS
--- a/easybuild/easyblocks/l/lammps.py
+++ b/easybuild/easyblocks/l/lammps.py
@@ -27,6 +27,7 @@
 @author: Pavel Grochal (INUITS)
 @author: Kenneth Hoste (Ghent University)
 @author: Alan O'Cais (Juelich Supercomputing Centre)
+@author: Alex Domingo (Vrije Universiteit Brussel)
 """
 
 import os
@@ -40,6 +41,7 @@ from easybuild.tools.config import build_option
 from easybuild.tools.modules import get_software_root, get_software_version
 from easybuild.tools.run import run_cmd
 from easybuild.tools.systemtools import get_shared_lib_ext
+from easybuild.tools.filetools import copy_file
 
 from easybuild.easyblocks.generic.cmakemake import CMakeMake
 
@@ -118,6 +120,14 @@ class EB_LAMMPS(CMakeMake):
         # Unset LIBS when using both KOKKOS and CUDA - it will mix lib paths otherwise
         if self.cfg['kokkos'] and get_software_root('CUDA'):
             env.unset_env_vars(['LIBS'])
+
+        # Add package files from Atomistica
+        if get_software_root('Atomistica'):
+            atomistica = get_software_root('Atomistica')
+            atomistica_files = ['pair_atomistica.%s' % e for e in ['cpp', 'h']]
+            self.log.info("Adding Atomistica package files into LAMMPS source: %s" % ','.join(atomistica_files))
+            for pkgfile in atomistica_files:
+                copy_file(os.path.join(atomistica, 'share/lammps', pkgfile), os.path.join('src', pkgfile))
 
     def configure_step(self, **kwargs):
         """Custom configuration procedure for LAMMPS."""
@@ -210,6 +220,23 @@ class EB_LAMMPS(CMakeMake):
                     self.cfg.update('configopts', '-DFFT=FFTW3')
             if '-DFFT_PACK=' not in self.cfg['configopts']:
                 self.cfg.update('configopts', '-DFFT_PACK=array')
+
+        # Additional linker flags
+        lammps_link_libs = []
+
+        # Atomistica
+        if get_software_root('Atomistica'):
+            lammps_link_libs += ['-latomistica']
+            if self.toolchain.options.get('usempi', None):
+                if self.toolchain.mpi_family() == toolchain.OPENMPI:
+                    lammps_link_libs += ['-lmpi_usempif08', '-lmpi_usempi_ignore_tkr', '-lmpi_mpifh', '-lmpi']
+                elif self.toolchain.mpi_family() == toolchain.INTELMPI:
+                    lammps_link_libs += ['-lmpifort', '-lmpi', '-lmpigi']
+
+        # Set additional linker flags
+        if len(lammps_link_libs) > 0:
+            cmake_lammps_link_libs = '-DLAMMPS_LINK_LIBS="%s"' % ' '.join(lammps_link_libs)
+            self.cfg.update('configopts', cmake_lammps_link_libs)
 
         # https://lammps.sandia.gov/doc/Build_extras.html
         # KOKKOS
--- a/easybuild/easyblocks/l/lammps.py
+++ b/easybuild/easyblocks/l/lammps.py
@@ -130,7 +130,7 @@ class EB_LAMMPS(CMakeMake):
             for pkgfile in atomistica_files:
                 copy_file(os.path.join(atomistica, 'share/lammps', pkgfile), os.path.join('src', pkgfile))
             # Test for sanity checks
-            copy_dir(os.path.join(atomistica, 'share/examples/liquid_silicon'), 'examples/atomistica')
+            copy_dir(os.path.join(atomistica, 'share/examples/LAMMPS/liquid_silicon'), 'examples/atomistica')
 
     def configure_step(self, **kwargs):
         """Custom configuration procedure for LAMMPS."""
--- a/easybuild/easyblocks/l/lammps.py
+++ b/easybuild/easyblocks/l/lammps.py
@@ -224,31 +224,6 @@ class EB_LAMMPS(CMakeMake):
             if '-DFFT_PACK=' not in self.cfg['configopts']:
                 self.cfg.update('configopts', '-DFFT_PACK=array')
 
-        # Additional linker flags
-        lammps_linklibs = []
-
-        # Atomistica
-        if get_software_root('Atomistica'):
-            # link with libatomistica
-            lammps_linklibs += ['-latomistica']
-            # Atomistica needs the MPI Fortran bindings, which are not added
-            # by default in LAMMPS as it uses a C++ linker
-            if self.toolchain.options.get('usempi', None):
-                if self.toolchain.mpi_family() == toolchain.OPENMPI:
-                    # retrieve libraries used to compile OpenMPI
-                    out, ec = run_cmd("mpifort --showme:libs", simple=False)
-                    if ec:
-                        raise EasyBuildError("Failed to determine OpenMPI libraries: %s", out)
-                    lammps_linklibs += ["-l%s" % lib for lib in out.split()]
-                elif self.toolchain.mpi_family() == toolchain.INTELMPI:
-                    # add Fortran libraries used by Intel MPI
-                    lammps_linklibs += ['-lmpifort', '-lmpi']
-
-        # Set additional linker flags
-        if len(lammps_linklibs) > 0:
-            cmake_lammps_linklibs = '-DLAMMPS_LINK_LIBS="%s"' % ' '.join(lammps_linklibs)
-            self.cfg.update('configopts', cmake_lammps_linklibs)
-
         # https://lammps.sandia.gov/doc/Build_extras.html
         # KOKKOS
         if self.cfg['kokkos']:


easybuild/easyblocks/l/lammps.py
```diff
@@ -211,6 +224,23 @@ def configure_step(self, **kwargs):
...
+        lammps_link_libs = []
+
+        # Atomistica
+        if get_software_root('Atomistica'):
+            lammps_link_libs += ['-latomistica']
+            if self.toolchain.options.get('usempi', None):
+                if self.toolchain.mpi_family() == toolchain.OPENMPI:
+                    lammps_link_libs += ['-lmpi_usempif08', '-lmpi_usempi_ignore_tkr', '-lmpi_mpifh', '-lmpi']
+                elif self.toolchain.mpi_family() == toolchain.INTELMPI:
+                    lammps_link_libs += ['-lmpifort', '-lmpi', '-lmpigi']
```
akesandgren, line 237: This feels wrong. The MPI compiler wrappers should add these, and especially for OpenMPI the names of the libraries depend on how OpenMPI was built.

lexming: I share the feeling. However, even using the MPI compiler wrappers hits `undefined reference` errors to MPI symbols at the very last link stage

Full log at https://gist.github.com/lexming/40d78c44817dd7def3c6a260351fea92
>[100%] Linking CXX executable lmp
/theia/home/apps/CO7/ivybridge-ib/software/CMake/3.15.3-GCCcore-8.3.0/bin/cmake -E cmake_link_script CMakeFiles/lmp.dir/link.txt --verbose=1
/apps/brussel/CO7/ivybridge-ib/software/OpenMPI/3.1.4-GCC-8.3.0/bin/mpicxx  -O2 -ftree-vectorize -march=native -fno-math-errno -fopenmp -std=c++11 -fopenmp -O3 -DNDEBUG  -L/user/brussel/101/vsc10122/.local/easybuild-ivybridge/software/Atomistica/0.10.1-foss-2019b-Python-3.7.4/lib -L/apps/brussel/CO7/ivybridge-ib/software/PLUMED/2.5.3-foss-2019b-Python-3.7.4/lib -L/apps/brussel/CO7/ivybridge-ib/software/yaff/1.6.0-foss-2019b-Python-3.7.4/lib -L/apps/brussel/CO7/ivybridge-ib/software/kim-api/2.1.3-foss-2019b/lib64 -L/apps/brussel/CO7/ivybridge-ib/software/FFTW/3.3.8-gompi-2019b/lib -L/apps/brussel/CO7/ivybridge-ib/software/Voro++/0.4.6-foss-2019b/lib -L/apps/brussel/CO7/ivybridge-ib/software/FFmpeg/4.2.1-GCCcore-8.3.0/lib -L/apps/brussel/CO7/ivybridge-ib/software/libxml2/2.9.9-GCCcore-8.3.0/lib -L/apps/brussel/CO7/ivybridge-ib/software/PCRE/8.43-GCCcore-8.3.0/lib -L/apps/brussel/CO7/ivybridge-ib/software/HDF5/1.10.5-gompi-2019b/lib -L/apps/brussel/CO7/ivybridge-ib/software/zlib/1.2.11-GCCcore-8.3.0/lib -L/apps/brussel/CO7/ivybridge-ib/software/GSL/2.6-GCC-8.3.0/lib -L/apps/brussel/CO7/ivybridge-ib/software/netCDF/4.7.1-gompi-2019b/lib64 -L/apps/brussel/CO7/ivybridge-ib/software/libjpeg-turbo/2.0.3-GCCcore-8.3.0/lib64 -L/apps/brussel/CO7/ivybridge-ib/software/libpng/1.6.37-GCCcore-8.3.0/lib -L/apps/brussel/CO7/ivybridge-ib/software/Python/3.7.4-GCCcore-8.3.0/lib -L/apps/brussel/CO7/ivybridge-ib/software/archspec/0.1.0-GCCcore-8.3.0-Python-3.7.4/lib -L/apps/brussel/CO7/ivybridge-ib/software/ScaLAPACK/2.0.2-gompi-2019b/lib -L/apps/brussel/CO7/ivybridge-ib/software/OpenBLAS/0.3.7-GCC-8.3.0/lib -L/apps/brussel/CO7/ivybridge-ib/software/GCCcore/8.3.0/lib64 -L/apps/brussel/CO7/ivybridge-ib/software/GCCcore/8.3.0/lib -rdynamic CMakeFiles/lmp.dir/theia/scratch/brussel/101/vsc10122/easybuild-ivb/build/LAMMPS/7Aug2019/foss-2019b-Python-3.7.4-kokkos-atomistica/lammps-stable_7Aug2019/src/main.cpp.o  -o lmp -Wl,-rpath,/theia/scratch/brussel/101/vsc10122/easybuild-ivb/build/LAMMPS/7Aug2019/foss-2019b-Python-3.7.4-kokkos-atomistica/easybuild_obj:/theia/scratch/brussel/101/vsc10122/easybuild-ivb/build/LAMMPS/7Aug2019/foss-2019b-Python-3.7.4-kokkos-atomistica/easybuild_obj/lib/kokkos: liblammps.so.0 -latomistica -ljpeg -lpng -lfftw3 -lpython3.7m -Wl,-Bstatic -lvoro++ -lplumed -Wl,-Bdynamic -ldl -lstdc++ -lfftw3 -lgsl -lz -lopenblas -lgfortran -lm -lpthread -L/apps/brussel/CO7/ivybridge-ib/software/SciPy-bundle/2019.10-foss-2019b-Python-3.7.4/lib -L/apps/brussel/CO7/ivybridge-ib/software/Python/3.7.4-GCCcore-8.3.0/lib -L/apps/brussel/CO7/ivybridge-ib/software/libmatheval/1.1.11-GCCcore-8.3.0/lib -L/apps/brussel/CO7/ivybridge-ib/software/GSL/2.6-GCC-8.3.0/lib -L/apps/brussel/CO7/ivybridge-ib/software/zlib/1.2.11-GCCcore-8.3.0/lib -L/apps/brussel/CO7/ivybridge-ib/software/FFTW/3.3.8-gompi-2019b/lib -L/apps/brussel/CO7/ivybridge-ib/software/ScaLAPACK/2.0.2-gompi-2019b/lib -L/apps/brussel/CO7/ivybridge-ib/software/OpenBLAS/0.3.7-GCC-8.3.0/lib -L/apps/brussel/CO7/ivybridge-ib/software/GCCcore/8.3.0/lib64 -L/apps/brussel/CO7/ivybridge-ib/software/GCCcore/8.3.0/lib -rdynamic -Wl,-Bsymbolic -ldl -lnetcdf -lcurl -L/apps/brussel/CO7/ivybridge-ib/software/kim-api/2.1.3-foss-2019b/lib64 -Wl,-rpath,/apps/brussel/CO7/ivybridge-ib/software/kim-api/2.1.3-foss-2019b/lib64 -lkim-api libcsmpi.a -lm libpoems.a libatc.a -lopenblas libcolvars.a libh5md.a -lhdf5 -lsz -lz -lm -lpthread lib/kokkos/libkokkos.so -fopenmp -ldl -lrt -ltbbmalloc 
liblammps.so.0: error: undefined reference to 'mpi_comm_set_errhandler_'
liblammps.so.0: error: undefined reference to 'mpi_cart_create_'
liblammps.so.0: error: undefined reference to 'mpi_finalize_'
liblammps.so.0: error: undefined reference to 'mpi_comm_split_'
liblammps.so.0: error: undefined reference to 'mpi_init_'
liblammps.so.0: error: undefined reference to 'mpi_cart_coords_'
liblammps.so.0: error: undefined reference to 'mpi_initialized_'
liblammps.so.0: error: undefined reference to 'mpi_comm_free_'
liblammps.so.0: error: undefined reference to 'mpi_allreduce_'
liblammps.so.0: error: undefined reference to 'mpi_scan_'
liblammps.so.0: error: undefined reference to 'mpi_bcast_'
liblammps.so.0: error: undefined reference to 'mpi_allgatherv_'
liblammps.so.0: error: undefined reference to 'mpi_allgather_'
liblammps.so.0: error: undefined reference to 'mpi_barrier_'
liblammps.so.0: error: undefined reference to 'mpi_cart_shift_'
liblammps.so.0: error: undefined reference to 'mpi_get_count_'
liblammps.so.0: error: undefined reference to 'mpi_sendrecv_'
liblammps.so.0: error: undefined reference to 'mpi_error_string_'
liblammps.so.0: error: undefined reference to 'mpi_comm_rank_'
liblammps.so.0: error: undefined reference to 'mpi_comm_size_'
liblammps.so.0: error: undefined reference to 'mpi_abort_'
collect2: error: ld returned 1 exit status

I do not know if there is a better way to fix this link errors. I'll check if the list of flags can be stripped down.

akesandgren: Was liblammps.so created with the mpi wrappers? My guess is it wasn't, it needs to be.

lexming: `liblammps.so` was created with `mpicxx` and had no errors or warnings...
>[100%] Linking CXX shared library liblammps.so
/theia/home/apps/CO7/ivybridge-ib/software/CMake/3.15.3-GCCcore-8.3.0/bin/cmake -E cmake_link_script CMakeFiles/lammps.dir/link.txt --verbose=1
/apps/brussel/CO7/ivybridge-ib/software/OpenMPI/3.1.4-GCC-8.3.0/bin/mpicxx -fPIC -O2 -ftree-vectorize -march=native -fno-math-errno -fopenmp -std=c++11 -fopenmp -O3 -DNDEBUG -L/user/brussel/101/vsc10122/.local/easybuild-ivybridge/software/Atomistica/0.10.1-foss-2019b-Python-3.7.4/lib -L/apps/brussel/CO7/ivybridge-ib/software/PLUMED/2.5.3-foss-2019b-Python-3.7.4/lib -L/apps/brussel/CO7/ivybridge-ib/software/yaff/1.6.0-foss-2019b-Python-3.7.4/lib -L/apps/brussel/CO7/ivybridge-ib/software/kim-api/2.1.3-foss-2019b/lib64 -L/apps/brussel/CO7/ivybridge-ib/software/FFTW/3.3.8-gompi-2019b/lib -L/apps/brussel/CO7/ivybridge-ib/software/Voro++/0.4.6-foss-2019b/lib -L/apps/brussel/CO7/ivybridge-ib/software/FFmpeg/4.2.1-GCCcore-8.3.0/lib -L/apps/brussel/CO7/ivybridge-ib/software/libxml2/2.9.9-GCCcore-8.3.0/lib -L/apps/brussel/CO7/ivybridge-ib/software/PCRE/8.43-GCCcore-8.3.0/lib -L/apps/brussel/CO7/ivybridge-ib/software/HDF5/1.10.5-gompi-2019b/lib -L/apps/brussel/CO7/ivybridge-ib/software/zlib/1.2.11-GCCcore-8.3.0/lib -L/apps/brussel/CO7/ivybridge-ib/software/GSL/2.6-GCC-8.3.0/lib -L/apps/brussel/CO7/ivybridge-ib/software/netCDF/4.7.1-gompi-2019b/lib64 -L/apps/brussel/CO7/ivybridge-ib/software/libjpeg-turbo/2.0.3-GCCcore-8.3.0/lib64 -L/apps/brussel/CO7/ivybridge-ib/software/libpng/1.6.37-GCCcore-8.3.0/lib -L/apps/brussel/CO7/ivybridge-ib/software/Python/3.7.4-GCCcore-8.3.0/lib -L/apps/brussel/CO7/ivybridge-ib/software/archspec/0.1.0-GCCcore-8.3.0-Python-3.7.4/lib -L/apps/brussel/CO7/ivybridge-ib/software/ScaLAPACK/2.0.2-gompi-2019b/lib -L/apps/brussel/CO7/ivybridge-ib/software/OpenBLAS/0.3.7-GCC-8.3.0/lib -L/apps/brussel/CO7/ivybridge-ib/software/GCCcore/8.3.0/lib64 -L/apps/brussel/CO7/ivybridge-ib/software/GCCcore/8.3.0/lib -shared -Wl,-soname,liblammps.so.0 -o liblammps.so.0 @CMakeFiles/lammps.dir/objects1.rsp @CMakeFiles/lammps.dir/objects2.rsp @CMakeFiles/lammps.dir/objects3.rsp -Wl,-rpath,/theia/scratch/brussel/101/vsc10122/easybuild-ivb/build/LAMMPS/7Aug2019/foss-2019b-Python-3.7.4-kokkos-atomistica/easybuild_obj/lib/kokkos: -latomistica -ljpeg -lpng -lz -lfftw3 -lpython3.7m -Wl,-Bstatic -lvoro++ -lplumed -Wl,-Bdynamic -ldl -lstdc++ -lfftw3 -lgsl -lz -lopenblas -lgfortran -lm -lpthread -L/apps/brussel/CO7/ivybridge-ib/software/SciPy-bundle/2019.10-foss-2019b-Python-3.7.4/lib -L/apps/brussel/CO7/ivybridge-ib/software/Python/3.7.4-GCCcore-8.3.0/lib -L/apps/brussel/CO7/ivybridge-ib/software/libmatheval/1.1.11-GCCcore-8.3.0/lib -L/apps/brussel/CO7/ivybridge-ib/software/GSL/2.6-GCC-8.3.0/lib -L/apps/brussel/CO7/ivybridge-ib/software/zlib/1.2.11-GCCcore-8.3.0/lib -L/apps/brussel/CO7/ivybridge-ib/software/FFTW/3.3.8-gompi-2019b/lib -L/apps/brussel/CO7/ivybridge-ib/software/ScaLAPACK/2.0.2-gompi-2019b/lib -L/apps/brussel/CO7/ivybridge-ib/software/OpenBLAS/0.3.7-GCC-8.3.0/lib -L/apps/brussel/CO7/ivybridge-ib/software/GCCcore/8.3.0/lib64 -L/apps/brussel/CO7/ivybridge-ib/software/GCCcore/8.3.0/lib -rdynamic -Wl,-Bsymbolic -fopenmp -lnetcdf -lcurl -L/apps/brussel/CO7/ivybridge-ib/software/kim-api/2.1.3-foss-2019b/lib64 -Wl,-rpath,/apps/brussel/CO7/ivybridge-ib/software/kim-api/2.1.3-foss-2019b/lib64 -lkim-api libcsmpi.a -lm libpoems.a libatc.a libcolvars.a libh5md.a lib/kokkos/libkokkos.so -ltbbmalloc -ldl -lopenblas -lhdf5 -lsz -lz -lm -lpthread -fopenmp -ldl -lrt 
/theia/home/apps/CO7/ivybridge-ib/software/CMake/3.15.3-GCCcore-8.3.0/bin/cmake -E cmake_symlink_library liblammps.so.0 liblammps.so.0 liblammps.so

lexming: @akesandgren I think that the issue is that `liblammps` is made with `mpicxx` and `libatomistica` with `mpifort`. Then when both are linked together `liblammps` is missing all fortran symbols from MPI. The only flags needed (and that I manually added to the easyblock) are MPI libs in fortran.

ocaisa: I found https://github.com/LLNL/scr/pull/132 and that was added because dependencies on deprecated C++ bindings were introduced somehow (see issue https://github.com/LLNL/scr/issues/130). Maybe a similar looking patch might do the job?

akesandgren: Ah.
The "correct" solution for OpenMPI is then to use
"mpifort --showme:libs" to find out which libs it needs instead of hard coding it.
Unfortunately there is no such option for mpiifort

akesandgren: For IMPI you should also only need -lmpifort the rest should be added bu mpiicpc

lexming: The ideal solution is perform the link step with the MPI Fortran compiler instead of the CXX one. Then all the fortran bindings are automatically added and it does not depend on the toolchain in use. The problem is that this needs adding a `LINKER_LANGUAGE` variable to a specific  `set_target_properties` in CMakeLists.txt.

easybuild/easyblocks/l/lammps.py
```diff
@@ -119,6 +121,17 @@ def prepare_step(self, *args, **kwargs):
         if self.cfg['kokkos'] and get_software_root('CUDA'):
             env.unset_env_vars(['LIBS'])
 
+        # Add files from Atomistica
+        atomistica = get_software_root('Atomistica')
```
Flamefire, line 125: Shouldn't this look at the dependencies? I think it was `self.dependencies()`

ocaisa: @Flamefire You are a bit of a CMake guru, is there a way to trigger linking against the Fortran MPI libraries? The think is I believe it should already be there  from https://github.com/lammps/lammps/blob/master/cmake/CMakeLists.txt#L706 so I wonder is CMake not correctly detecting MPI?

easybuild/easyblocks/l/lammps.py
```diff
@@ -211,6 +224,31 @@ def configure_step(self, **kwargs):
...
+                    if ec:
+                        raise EasyBuildError("Failed to determine OpenMPI libraries: %s", out)
+                    lammps_linklibs += ["-l%s" % lib for lib in out.split()]
+                elif self.toolchain.mpi_family() == toolchain.INTELMPI:
+                    # add Fortran libraries used by Intel MPI
+                    lammps_linklibs += ['-lmpifort', '-lmpi']
+
+        # Set additional linker flags
+        if len(lammps_linklibs) > 0:
+            cmake_lammps_linklibs = '-DLAMMPS_LINK_LIBS="%s"' % ' '.join(lammps_linklibs)
```
Flamefire, line 249: Good idea, but won't work for newer versions: https://github.com/lammps/lammps/commit/5828815b3ec50845be1a312056de559c05bbff1a

Flamefire:
FWIW: I added a feature request to lammps: https://github.com/lammps/lammps/issues/2039

ocaisa:
I'm concerned that if we include Atomistica support by default in the shipped easyconfigs then those easyconfigs will only work with `foss` and `intel` since we are pretty confident the link step will fail in other cases (e.g., `try_toolchain` will likely fail for other MPIs). Since we know we can't support other cases we should add a final `else` clause when checking the MPI family and give some advice about what to do.

easybuild/easyblocks/l/lammps.py
Flamefire: It's hard. Let me give you some information on what's happening (If I'm not mistaken and they e.g. use a custom FindMPI in CMake [e.g. https://github.com/lammps/lammps/blob/b38e95f88c001b6cad1097f0e8a4d65508582ca8/cmake/CMakeLists.txt#L131]):

- They search for MPI https://github.com/lammps/lammps/blob/b38e95f88c001b6cad1097f0e8a4d65508582ca8/cmake/CMakeLists.txt#L155
- This creates imported targets containing the MPI libs, IIRC this is done by invoking the mpi wrapper and querying the libraries to use.
- This imported library is then linked to lammps: https://github.com/lammps/lammps/blob/b38e95f88c001b6cad1097f0e8a4d65508582ca8/cmake/CMakeLists.txt#L156
- Some information about found libraries is given as a message only: https://github.com/lammps/lammps/blob/b38e95f88c001b6cad1097f0e8a4d65508582ca8/cmake/CMakeLists.txt#L707-L709

Note how they explicitly link against `MPI::MPI_CXX`, so no fortran there.

Question I have: If they use MPI from C++ and Atomistica uses MPI from Fortran, don't you need to link against the MPI C++ and the MPI Fortran lib?

If so, then it would be best to write a simple CMake find script:

```
find_library(ATOMISITICA_LIB atomistica)
find_package(MPI COMPONENTS Fortran)
add_library(Atomistica::Atomistica STATIC IMPORTED)
set_target_properties(Atomistica::Atomistica
  IMPORTED LOCATION ${ATOMISITICA_LIB}
  INTERFACE_LINK_LIBRARIES MPI::MPI_Fortran
  IMPORTED_LINK_INTERFACE_LANGUAGES Fortran
)
include(FindPackageHandleStandardArgs)
find_package_handle_standard_args(Atomistica REQUIRED_VARS ATOMISITICA_LIB MPI_Fortran_FOUND)
```

The created library can then be linked against the lammps library:
- Save above to `FindAtomistica.cmake` in cmake/Modules of lammps
- Add call to `find_package(Atomistica REQUIRED)` to the CML
- Add `target_link_libraries(lammps PUBLIC Atomistica::Atomistica)`

If that works, the `FindAtomistica.cmake` can be contributed to LAMMPS so they can properly do it and make our job in the future easier, see https://github.com/lammps/lammps/issues/2039

easybuild/easyblocks/l/lammps.py
lexming: Setting `LINKER_LANGUAGE` is not easy to apply because it is not enough to link the final `lmp` executable with `mpifort`, other shared libraries also have references to MPI fortran bindings such as `liblammps.so`. So there is no simple command that can be done from this easyblock to fix the build of LAMMPS with Atomistica. The solution is to patch `CMakeLists.txt`.

easybuild/easyblocks/l/lammps.py
lexming: @Flamefire Thanks for going through the cmake build of LAMMPS and pointing out the required changes to handle a build linking with Atomistica. I've adapted them into a patch that works perfectly well :+1:

easybuild/easyblocks/l/lammps.py
lexming: This is very very sad :cry:  I decided to use `LAMMPS_LINK_LIBS` because I thought that this was a version agnostic approach that would not need a new revision on each release. If that is not the case, then it is just better to patch `CMakeLists.txt` instead of hacking the build from this easyblock.

lexming:
> LGTM, but I still think making sure Atomistica's library is a correctly built shared library would be the better solution since all this would go away.

I checked if `libatomisica` could be easily build as a shared object. But unfortunately it is not the case. Its makefiles are not prepared at to do a shared build and manually adding `-shared` to the compilation hits unexpected errors. I won't waste time on it.

easybuild/easyblocks/l/lammps.py
lexming: Patch is in https://github.com/easybuilders/easybuild-easyconfigs/pull/10433/commits/72b8adeb21dcc8dc10ccd5eed27737197f5a3455

lexming:
@akesandgren @Flamefire @ocaisa I removed the code messing with the library links of LAMMPS because `LAMMPS_LINK_LIBS` will be removed in newer releases. The only reason to go with that approach was to have something that is version agnostic. If that's not the case, we can just patch the CMake files. The advantage of patching is that it can be made independent from the toochain in use (wrt @ocaisa comment).

I'm not attempting to do any patching from the easyblock as the easyconfigs are much better suited for that task. I already updated PR https://github.com/easybuilders/easybuild-easyconfigs/pull/10433 and tests with this easyblock can be found in there.

Flamefire:
The patch looks good so far. I'm wondering: Could we make it more complete? What is missing is that `<atomisitica_root>/share/lammps/pair_atomistica.cpp` needs to be added to the lammps library. This can be done with `find_file(ATOMISTICA_PAIR_ATOMISTICA pair_atomistica.cpp PATH_SUFFIXES share/lammps)` in the Find*.cmake and `target_sources(lammps PRIVATE ${ATOMISTICA_PAIR_ATOMISTICA})`   
Maybe better alternative for the find: `find_path(ATOMISTICA_ROOT share/lammps/pair_atomistica.cpp)` and `set(ATOMISTICA_ADDITIONAL_FILES ${ATOMISTICA_ROOT}/share/lammps/pair_atomistica.cpp)` and use ATOMISTICA_ADDITIONAL_FILES in the CMakeLists. This has the advantage that if they decide to use an additional file one can simply add it to the list. Also the variable can be set by the invoker of CMake via `-D` and `ATOMISTICA_ROOT` is easier to understand.   

My motivation for that: This patch can be contributed as a PR to LAMMPS. This means we don't need to patch in the future.

I've seen you copy some example folder. Is that required or can that stay in the atomistica folder and be tested in there? You don't install it (again) so I guess yes.

Missing from this PR: change `get_software_root` to `dependencies`. I know it is minor but it makes the intend clear: Build with Atomistica when it was specified as a dependency as opposed to "do it if it happens to be loaded". This avoids also issues with EB_IGNORE_LOADED_MODULES (or what it was)

lexming:
@Flamefire I think that you are going too far with this issue now. The devs of LAMMPS showed no interest in your feature request and I am, personally, not interested in going through the ordeal of making a sane build system for Atomistica that can contributed upstream. Nonetheless, thanks a lot for your help, I really appreciate it.

Regarding the other remarks
* the files copied from examples is a set of tests for Atomistica that are run as part of the sanity checks in LAMMPS. Those are not installed with LAMMPS but with Atomistica.
* I use `get_software_root` not only as means to detect the presence of the module, but also to get the path to the installation, as it is needed to copy all those files. I agree that I could also add a check on `dependencies` but it is not done for the other deps in LAMMPS and it feels redundant given that I need to get the root anyhow.

Flamefire:
> The devs of LAMMPS showed no interest in your feature request 

No, they just said they won't do it themselves:
> We'll be more than happy to include a submission from the Atomistica developers in a similar vain.

Don't want you to do too much, but with the current approach the patch is almost enough and the change required is very small. That's why I suggested doing it. I expect also our lives to be easier if that is upstream. But in the end it is your decision of course.

> Those are not installed with LAMMPS but with Atomistica.

Ok, but why do you need to copy them to LAMMPS? Can't you just run them from where they are?

>  I could also add a check on dependencies but it is not done for the other deps in LAMMPS and it feels redundant 

It is not redundant as generally it could be a dependency of a dependency or just left loaded in the environment. IMO in that case it shouldn't get build with the rest. But given that we don't even have an EC for Atomistica but make it an extension of LAMMPS it should be fine. Didn't know that earlier

lexming:
> Ok, but why do you need to copy them to LAMMPS? Can't you just run them from where they are?

Yes I could. But the reason to not do it is to not pollute the installation directory of Atomistica with the files generated by the execution of the tests. I prefer to copy the necessary files to the build directory so that they are deleted at the end of the installation.

> It is not redundant as generally it could be a dependency of a dependency or just left loaded in the environment. IMO in that case it shouldn't get build with the rest. But given that we don't even have an EC for Atomistica but make it an extension of LAMMPS it should be fine. Didn't know that earlier

I would argue that it is most important to check for `software_root` than for dependency because the module might fail to properly set the environment even if it is declared as dependency. In any case, the best approach would be of course to check for both, no discussion about that. However, for this particular case, I think that we can settle with a check of `software_root` and a check of Atomistica in the list of `user_packages` defined in the easyconfig. Then, the copy of files from Atomistica only happens if 1) its installation path is available and 2) its package was explicitly requested to be installed in LAMMPS.

lexming:
Test with last commit in https://github.com/easybuilders/easybuild-easyconfigs/pull/10433#issuecomment-621987478

akesandgren:
@ocaisa how did the testing go?

ocaisa:
Not well unfortunately, see https://github.com/easybuilders/easybuild-easyconfigs/pull/10433#issuecomment-622393219

PR #2040, opened 2020-04-23
support having pyqt installed as part of Qt5

mboisson:
In our Qt installation, we install PyQt as part of Qt, rather than as a separate module. This change supports having that. 

It requires this framework PR : 
~~https://github.com/easybuilders/easybuild-framework/pull/3297~~


mboisson:
This PR is relevant in the context of this issue (which we currently already do within Compute Canada) 

https://github.com/easybuilders/easybuild-easyblocks/issues/2041

Micket:
@boegel suggested during the last conference call that you might be able to simply define EBROOTPYQT5 inside the Qt5 module, and this should work without having to hardcode a special case.

mboisson:
Mmm, that could solve the `elif qt5` part, but not the `sip_incdir` part.

Micket:
I agree, that fix looks like it's needed regardless, for python 3.8

lexming:
Part of the functionality of this PR has been already included with #2077

akesandgren:
@mboisson conflict resolution needed

akesandgren:
This PR now looks strange.
If I look in "Files changed" at the top I only see the "elif qt5" changes.
But if I look at the last "resolved conflics" commit, I only see changes to "pyqt_sipdir".

lexming:
@akesandgren I updated the title, which was indeed confusing. Now this PR just adds `Qt5` as a possible source of `pyqt`.

akesandgren:
In the case of only havaing a Qt5 install (with pyqt in it) shouldn't it somehow verify that it is there at all before going forward? Or is that done later on?

lexming:
There is no further check, the calls later on to `find_glob_pattern` could indeed be done with `fail_on_no_match=True` to handle such situation. Right now this PR works under the assumption that if you are installing Qscintilla with Qt5 loaded but not PyQt5, then Qt5 is providing `pyqt`.

akesandgren:
Okidoki...

akesandgren (approved):
LGTM

akesandgren:
Going in, thanks @mboisson!

PR #2042, opened 2020-04-24
Add GoPackage EasyBlock

Darkless012:


Darkless012:
Yet another take on reproducible go package easyblock.
Related: https://github.com/easybuilders/easybuild-easyblocks/pull/1629 https://github.com/easybuilders/easybuild-easyblocks/issues/1466

This time GoPackage easyblock is using `go modules` which are currently "the way to go" with managing dependencies in go.  

Setting only `GOBIN` to installdir/bin will ensure the binaries are put into correct place.
Apparently `GOPATH` is not a thing anymore...

**How it works?**

**Option a)** There is `go.mod` and `go.sum` in the downloaded package.
`go install` will download all the dependencies while `go.sum` ensures that those dependencies are always the same.

**Option b)** There is no `go.mod` and `go.sum`.
Go allows to pretty easy migration to native modules from various old "packaging" tools (https://blog.golang.org/migrating-to-go-modules)
In this case the bundle tries to do exactly that.

However "**modulename**" is required in this case. So that go knows how to call it.

Sometimes it also fails to pinpoint exact version of dependency, so "**force_deps**" is used to say go which version of that dependency is used.

Then it does all the magic.
As part of output there is generated `go.mod` and `go.sum` files in the log. 

Ideally you should make patchfile with those 2 new files, and include that into build. And this should truly ensure reproducibility of the build.

easybuild/easyblocks/generic/gopackage.py
```diff
@@ -0,0 +1,169 @@
...
+        # enforce use of go modules
+        env.setvar('GO111MODULE', 'on', verbose=False)
+        # set bin folder
+        env.setvar('GOBIN', os.path.join(self.installdir, 'bin'), verbose=False)
+
+    def configure_step(self):
+        """Configure Go package build/install."""
+
+        if self.go_cmd is None:
+            self.prepare_go()
```
boegel, line 90: You can drop this if you use `prepare_step` above, and just assume `self.go_cmd` is set by now

easybuild/easyblocks/generic/gopackage.py
```diff
@@ -0,0 +1,169 @@
...
+
+        # creates log entries for go being used, for debugging
+        run_cmd("%s version" % self.go_cmd, verbose=False, trace=False)
+        run_cmd("%s env" % self.go_cmd, verbose=False, trace=False)
+
+    def build_step(self):
+        """If Go package is not native go module, lets try to make the module."""
+
+        go_mod_file = os.path.join('./', 'go.mod')
+        go_sum_file = os.path.join('./', 'go.sum')
```
boegel, line 100: No need for the `./`:

```python
go_mod_file = 'go.mod'
go_sum_file = 'go.sum'
```

or

```python
go_mod_file, go_sum_file = 'go.mod', 'go.sum'

File: easybuild/easyblocks/generic/gopackage.py [hunk 3]
```diff
@@ -0,0 +1,169 @@
...
+            cmd = ' '.join([self.go_cmd, 'mod init', self.cfg['modulename']])
+            run_cmd(cmd, log_all=True, simple=True)
+
+            if self.cfg['forced_deps']:
+                for dep in self.cfg['forced_deps']:
+                    # go get specific dependencies which locks them in go.mod
+                    cmd = ' '.join([self.go_cmd, 'get %s@%s' % dep])
+                    run_cmd(cmd, log_all=True, simple=True)
+
+            # go build ./...
```
boegel, line 121: wtf is this triple dot thing? :D

easybuild/easyblocks/generic/gopackage.py
```diff
@@ -0,0 +1,169 @@
...
+            self.log.warn("go.mod not found! This is not natively supported go module. Trying to init module.")
+
+            if self.cfg['modulename'] is None:
+                raise EasyBuildError("Installing non-native go module. You need to specify 'modulename' in easyconfig")
+
+            # for more information about migrating to go modules
+            # see: https://blog.golang.org/migrating-to-go-modules
+
+            # go mod init
+            cmd = ' '.join([self.go_cmd, 'mod init', self.cfg['modulename']])
```
boegel, line 112:
```python
cmd = ' '.join([self.go_cmd, 'mod', 'init', self.cfg['modulename']])
```

easybuild/easyblocks/generic/gopackage.py
```diff
@@ -0,0 +1,169 @@
...
+            run_cmd(cmd, log_all=True, simple=True)
+
+            if self.cfg['forced_deps']:
+                for dep in self.cfg['forced_deps']:
+                    # go get specific dependencies which locks them in go.mod
+                    cmd = ' '.join([self.go_cmd, 'get %s@%s' % dep])
+                    run_cmd(cmd, log_all=True, simple=True)
+
+            # go build ./...
+            cmd = ' '.join([self.go_cmd, 'build ./...'])
```
boegel, line 122:
```python
cmd = ' '.join([self.go_cmd, 'build', './...'])
```

same below

(bonus points if you can add a comment to explain the triple dot, it looks like a typo...)

easybuild/easyblocks/generic/gopackage.py
```diff
@@ -0,0 +1,169 @@
...
+            run_cmd(cmd, log_all=True, simple=True)
+
+            # go test ./... again
+            cmd = ' '.join([self.go_cmd, 'test ./...'])
+            run_cmd(cmd, log_all=True, simple=True)
+
+            self.log.warn('Include generated go.mod and go.sum via patch to ensure locked dependencies '
+                          'and run this easyconfig again.')
+            cmd = 'cat go.mod'
+            run_cmd(cmd, log_all=True, simple=True)
```
boegel, line 144: No reason to use `cmd` here?

```python
run_cmd("cat go.mod", log_all=True, simple=True)

File: easybuild/easyblocks/generic/gopackage.py [hunk 7]
```diff
@@ -0,0 +1,169 @@
...
+
+        if not os.path.exists(go_sum_file) or not os.path.isfile(go_sum_file):
+            raise EasyBuildError("go.sum not found! This module has no locked dependency versions.")
+
+    def install_step(self):
+        """Install Go package to a custom path"""
+
+        # actually install Go package
+        cmd = ' '.join(
+            [self.cfg['preinstallopts'], self.go_cmd, 'install', self.cfg['installopts']])
```
boegel, line 157:
```python
cmd = ' '.join([
    self.cfg['preinstallopts'],
    self.go_cmd,
    'install',
    self.cfg['installopts'],
])

File: easybuild/easyblocks/generic/gopackage.py [hunk 8]
```diff
@@ -0,0 +1,169 @@
...
+        run_cmd(cmd, log_all=True, log_ok=True, simple=True)
+
+    def sanity_check_step(self):
+        """Custom sanity check for Go package."""
+
+        # Go Package should produce something into bin directory
+        custom_paths = {
+            'files': [],
+            'dirs': ['bin'],
```
boegel, line 166: This is very generic... We can't derive anything from software name?

easybuild/easyblocks/generic/gopackage.py
Darkless012: From what I've learned its, build "whatever file you can find (and build)" in this folder. And it's probably just a Go thing.

Darkless012: https://stackoverflow.com/a/28031651/2047157

easybuild/easyblocks/generic/gopackage.py
Darkless012: I'm not very confident in naming in go. But it should match last part of Go module name, considering that main.go is in root. Otherwise it could be any name from any sub-folder producing binary.

The main reason for this "basic" check was that for example compiling gonum doesn't produce any binary, since it is intended to be used as dependency in go source. So you are able to build and install this package, but no binary will be dropped in bin and nothing you can use on easybuild will be installed.

Should we check the name (namelower)?
Consider this: https://github.com/evolbioinfo/gotree
They have 3 different names in single paragraph: 
`Gotree` in heading
`GoTree` just below in description
`gotree` as cli command

boegel: I think assuming that there will be a `bin/%(namelower)s` makes sense, since this is only a default (it's easy to override in the easyconfig file).

This would work for both `goalign` and `gotree`.

Maybe also add `%(namelower)s --help` as default sanity check commands (via `custom_commands` here)?

Also easy to override in easyconfig file via `sanity_check_commands`.

PR #2043, opened 2020-04-24
add support for building NWChem on top of external GlobalArrays + also define $LAPACK_LIB (required for NWChem 7.x)

Micket:
WIP

It is now required to set `LAPACK_LIB` as well. Shouldn't hurt to set it for older NWchem either, so I skipped the messy if-version stuff.

External GlobalArrays is a big benefit, so that it doesn't download and builds it's own (sometimes buggy) version.
I'm not sure what to do with
```python
            elif mpi_family in [toolchain.INTELMPI]:
                if self.cfg['armci_network'] in ["MPI-MT"]:
                    libmpi = "-lmpigf -lmpigi -lmpi_ilp64 -lmpi_mt"
                else:
                    libmpi = "-lmpigf -lmpigi -lmpi_ilp64 -lmpi"
```

### DIFF
--- a/easybuild/easyblocks/n/nwchem.py
+++ b/easybuild/easyblocks/n/nwchem.py
@@ -236,8 +236,8 @@ class EB_NWChem(ConfigureMake):
         self.setvar_env_makeopt('FOPTIMIZE', os.getenv('FFLAGS'))
 
         # BLAS and ScaLAPACK
-        MPI_LIB_DIRS = ' '.join('-L' + d for d in os.getenv('MPI_LIB_DIR').split())
-        self.setvar_env_makeopt('BLASOPT', ' '.join([os.getenv('LDFLAGS'), MPI_LIB_DIRS,
+        mpi_lib_dirs = ' '.join('-L' + d for d in os.getenv('MPI_LIB_DIR').split())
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/n/nwchem.py
```diff
@@ -222,16 +227,21 @@ def configure_step(self):
...
+                libmpi += " -libumad -libverbs -lpthread"
 
         # compiler optimization flags: set environment variables _and_ add them to list of make options
         self.setvar_env_makeopt('COPTIMIZE', os.getenv('CFLAGS'))
         self.setvar_env_makeopt('FOPTIMIZE', os.getenv('FFLAGS'))
 
         # BLAS and ScaLAPACK
-        self.setvar_env_makeopt('BLASOPT', '%s -L%s %s %s' % (os.getenv('LDFLAGS'), os.getenv('MPI_LIB_DIR'),
-                                                              os.getenv('LIBSCALAPACK_MT'), libmpi))
+        MPI_LIB_DIRS = ' '.join('-L' + d for d in os.getenv('MPI_LIB_DIR').split())
```
boegel, line 239: @Micket Style-wise, this should be lower case since it's not a constant, but a variable.

```python
mpi_lib_dirs = ' '.join('-L' + d for d in os.getenv('MPI_LIB_DIR').split())

PR #2045, opened 2020-04-27
impi: execute minimal test in sanity check commands

lexming:
This PR improves the sanity checks of `impi` easyblock by building the minimal MPI test program `test.c` and adding its execution to the sanity check commands.

I tested it with latest impi-2019.7.217-iccifort-2020.1.217.eb
* broken build with default easyconfig properly **fails** on sanity checks: https://gist.github.com/lexming/5ab5be4624fa0786547973a65897f40f
* sane build setting `'I_MPI_HYDRA_TOPOLIB': 'ipl'` in easyconfig (which we need in our site) properly **passes** all sanity checks: https://gist.github.com/lexming/29ac434c5e7ad799cca986541f214a00

The tests are currently enabled for all versions as this test program is available in all non-archived releases of `impi`.

### DIFF
--- a/easybuild/easyblocks/i/impi.py
+++ b/easybuild/easyblocks/i/impi.py
@@ -212,20 +212,14 @@ EULA=accept
         }
 
         # Add minimal test program to sanity checks
-        try:
-            fake_mod_data = self.load_fake_module()
-        except EasyBuildError as err:
-            self.log.debug("Loading fake module failed: %s" % err)
-
         impi_testsrc = os.path.join(self.installdir, 'test/test.c')
         impi_testexe = os.path.join(self.builddir, 'mpi_test')
-        self.log.info("Building minimal MPI test program: %s", impi_testsrc)
-        build_test = "mpiicc %s -o %s" % (impi_testsrc, impi_testexe)
-        run_cmd(build_test, log_all=True, simple=True)
+        self.log.info("Adding minimal MPI test program to sanity checks: %s", impi_testsrc)
 
-        self.clean_up_fake_module(fake_mod_data)  # unload build environment
-
-        custom_commands = ['mpirun -n %s %s' % (self.cfg['parallel'], impi_testexe)]
+        custom_commands = [
+            "mpiicc %s -o %s" % (impi_testsrc, impi_testexe),  # build test program
+            'mpirun -n %s %s' % (self.cfg['parallel'], impi_testexe),  # run test program
+        ]
 
         super(EB_impi, self).sanity_check_step(custom_paths=custom_paths, custom_commands=custom_commands)
 
--- a/easybuild/easyblocks/i/impi.py
+++ b/easybuild/easyblocks/i/impi.py
@@ -213,27 +213,30 @@ EULA=accept
             'dirs': [],
         }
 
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/i/impi.py
```diff
@@ -210,7 +211,23 @@ def sanity_check_step(self):
...
+        try:
+            fake_mod_data = self.load_fake_module()
+        except EasyBuildError as err:
+            self.log.debug("Loading fake module failed: %s" % err)
+
+        impi_testsrc = os.path.join(self.installdir, 'test/test.c')
+        impi_testexe = os.path.join(self.builddir, 'mpi_test')
+        self.log.info("Building minimal MPI test program: %s", impi_testsrc)
+        build_test = "mpiicc %s -o %s" % (impi_testsrc, impi_testexe)
+        run_cmd(build_test, log_all=True, simple=True)
```
boegel, line 224: @lexming Why not make the compilation command part of the sanity check command?

That way, you can avoid having to load the (fake) module above?

lexming: good point, fixed :+1:

lexming:
Positive test with last commit is available at https://gist.github.com/lexming/638c1af52b057f56fa90f55b7ee37c05

easybuild/easyblocks/i/impi.py
```diff
@@ -210,7 +211,17 @@ def sanity_check_step(self):
...
 
-        super(EB_impi, self).sanity_check_step(custom_paths=custom_paths)
+        # Add minimal test program to sanity checks
+        impi_testsrc = os.path.join(self.installdir, 'test/test.c')
+        impi_testexe = os.path.join(self.builddir, 'mpi_test')
+        self.log.info("Adding minimal MPI test program to sanity checks: %s", impi_testsrc)
+
+        custom_commands = [
+            "mpiicc %s -o %s" % (impi_testsrc, impi_testexe),  # build test program
+            'mpirun -n %s %s' % (self.cfg['parallel'], impi_testexe),  # run test program
```
boegel, line 221: @lexming We should avoid hardcoding `mpirun` here, see https://github.com/lexming/easybuild-easyblocks/pull/2

akesandgren: And mpiicc too since this will also be used by gimpi toolchain

lexming: Good points! (I was not even aware that `gimpi` was a thing :D)

lexming:
SUCCESSFUL test with `impi-2019.7.217-iccifort-2020.1.217.eb` at:  https://gist.github.com/lexming/0b9c92be1146d7c2f9369508a2910515

lexming:
SUCCESSFUL test with `impi-2018.3.222-GCC-7.3.0-2.30.eb` at: https://gist.github.com/lexming/d8171b4d9dd89bf6fed6d132f67a4d3f

boegel:
@lexming The oldest `impi` versions we still have (4.x, 5.x) don't have `test/test.c` available, so I've added a commit to only build & run the test command for recent versions (>= 2017.x)...

PR #2047, opened 2020-04-28
clean $CPATH before building ROOT to avoid clash between external LLVM loaded as a dependency and internal LLVM used by ROOT

boegel:
fix for problem reported by @arielzn in https://sft.its.cern.ch/jira/browse/ROOT-10634 when building ROOT 6.20.04 with `foss/2019b`, where header files of `LLVM` 9.0.0 that is loaded as a dep for `Mesa` 19.1.7 (a direct dep for `ROOT`) are picked up, leading to errors like:

```
/tmp/root-6.20.04/interpreter/llvm/src/tools/clang/include/clang/Sema/Ownership.h:123:10: error: 'isPodLike' is not a class template
--
struct isPodLike<clang::OpaquePtr<T> > { static const bool value = true; }
;
^~~~~~~~~
/tmp/root-6.20.04/interpreter/llvm/src/tools/clang/include/clang/Sema/Ownership.h:123:40: error: 'llvm::isPodLike' is not a template
struct isPodLike<clang::OpaquePtr<T> >
{ static const bool value = true; };


```

### DIFF
--- a/easybuild/easyblocks/r/root.py
+++ b/easybuild/easyblocks/r/root.py
@@ -35,6 +35,7 @@ import os
 from easybuild.framework.easyconfig import CUSTOM
 from easybuild.easyblocks.generic.cmakemake import CMakeMake
 from easybuild.tools.build_log import EasyBuildError
+from easybuild.tools.environment import setvar
 from easybuild.tools.modules import get_software_root, get_software_version
 from easybuild.tools.run import run_cmd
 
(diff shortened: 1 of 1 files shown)


boegel:
Also tested with existing easyconfig `ROOT-6.14.06-foss-2018b-Python-3.6.6.eb`, should be good to go...

akesandgren:
Going in, thanks @boegel!

PR #2048, opened 2020-04-29
fixes/enhancements for PETSc easyblock

boegel:
* allow disabling `--with-mpi4py` option (enabling it seems totally broken, yields errors like `No rule to make target `mpi4py-build'`)
* add `SciPy-bundle` to dep filter (not really necessary since unknown options to `configure` command are silently ignored, but fine...)
* add path to PETSc Python scripts to `$PYTHONPATH` (see https://www.mcs.anl.gov/petsc/documentation/faq.html#sparse-matrix-ascii-format)
* add sanity check command for `PetscBinaryIO` Python module

requires for https://github.com/easybuilders/easybuild-easyconfigs/pull/10532


akesandgren:
Going in, thanks @boegel!

PR #2050, opened 2020-04-30
gitignore .mypy_cache dirs

willfurnass:


boegel:
OK I guess, but same question as in https://github.com/easybuilders/easybuild-framework/pull/3313#issuecomment-621942847...

willfurnass:
I didn't realise until you pointed it out that one could have global `.gitignore`s so feel free to reject and close.

boegel:
Clarified in https://github.com/easybuilders/easybuild-framework/pull/3313, good to go in, doesn't hurt...

PR #2051, opened 2020-05-01
don't let extract_file change directory in various easyblocks

boegel:
requires https://github.com/easybuilders/easybuild-framework/pull/3292 where implicit changing of directory in `extract_file` was deprecated


akesandgren:
Going in, thanks @boegel!

PR #2052, opened 2020-05-01
make postinstallcmds available to Mathematica easyconfigs + add bin/Executables to $PATH

smoors:


easybuild/easyblocks/m/mathematica.py
```diff
@@ -112,10 +114,14 @@ def post_install_step(self):
         else:
             self.log.info("No activation key provided, so skipping activation of the installation.")
 
+        super(EB_Mathematica, self).post_install_step()
```
boegel, line 117: I don't having this here, but you should do the symlinking you do in the easyconfig here instead (via the `symlink` function from `filetools`)?

smoors: I decided to just add `Executables` to the PATH, since everything in `bin` is symlinking to `Executables` anyway.

PR #2053, opened 2020-05-04
FreeSurfer: fix the extra dirs added to PATH

branfosj:
Only `bin` was ending up in the `PATH` for us. I think this is because `make_module_guesses` is not the right function and never gets called and then `bin` ends up there by default.

The change to line 47 is to fix:
```
ERROR EasyBuild crashed with an error (at easybuild/base/exceptions.py:124 in __init__): Use of unknown easyconfig parameter 'preinstall_cmd' when getting parameter value (at easybuild/framework/easyconfig/easyconfig.py:1638 in __getitem__)
```

### DIFF
--- a/easybuild/easyblocks/f/freesurfer.py
+++ b/easybuild/easyblocks/f/freesurfer.py
@@ -55,7 +55,7 @@ class EB_FreeSurfer(Tarball):
         """Include correct subdirectories to $PATH for FreeSurfer."""
         guesses = super(EB_FreeSurfer, self).make_module_req_guess()
 
-        guesses.update({'PATH': ['bin', 'fsfast/bin', 'mni/bin', 'tktools']})
+        guesses['PATH'].extend([os.path.join('fsfast', 'bin'), os.path.join('mni', 'bin'), 'tktools'])
 
         return guesses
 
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/f/freesurfer.py
```diff
@@ -44,18 +44,18 @@ def extra_options():
...
         write_file(os.path.join(self.installdir, '.license'), self.cfg['license_text'])
 
-    def make_module_guesses(self):
+    def make_module_req_guess(self):
         """Include correct subdirectories to $PATH for FreeSurfer."""
-        guesses = super(EB_FreeSurfer, self).make_module_guesses()
+        guesses = super(EB_FreeSurfer, self).make_module_req_guess()
 
-        guesses['PATH'] = ['bin', 'fsfast/bin', 'mni/bin', 'tktools']
+        guesses.update({'PATH': ['bin', 'fsfast/bin', 'mni/bin', 'tktools']})
```
akesandgren, line 58: bin comes from EasyBlock (via Tarball) and should not be needed here.

Can you check that?

branfosj: Yes, so long as we use `guesses['PATH'].extend()`. I've updated the PR to do that.

akesandgren:
A test report for an easyconfig using this last version of the easyblock?

branfosj:
I couldn't work out a way to get a test report directly here - so I used the PR for the most recent FreeSurfer EasyConfig. The test report is at https://gist.github.com/branfosj/aec22ab46d698ee30f10e780db0ec9a9

akesandgren:
That's the way to do it and then reference the PR with the test report.

akesandgren:
Going in, thanks @branfosj!

PR #2058, opened 2020-05-15
Use new MATLAB binary installer for versions >= 2020a

orbsmiv:
MATLAB will be deprecating the script-based installation method from 2020b onwards.

### DIFF
--- a/easybuild/easyblocks/m/matlab.py
+++ b/easybuild/easyblocks/m/matlab.py
@@ -170,8 +170,6 @@ class EB_MATLAB(PackedBinary):
             except IOError as err:
                 raise EasyBuildError("Failed to update config file %s: %s", self.configfile, err)
 
-            self.log.debug('configuration file updated with installation key:\n %s', config)
-
             (out, _) = run_cmd(cmd, log_all=True, simple=False)
 
             # check installer output for known signs of trouble
--- a/easybuild/easyblocks/m/matlab.py
+++ b/easybuild/easyblocks/m/matlab.py
@@ -140,6 +140,9 @@ class EB_MATLAB(PackedBinary):
         if isinstance(keys, string_type):
             keys = keys.split(',')
 
+        # Compile the installation key regex outside of the loop
+        regkey = re.compile(r"^(# )?fileInstallationKey=.*", re.M)
+
         # Make one install for each key
         for i, key in enumerate(keys):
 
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/m/matlab.py
```diff
@@ -141,24 +141,56 @@ def install_step(self):
...
+            self.log.debug('Installing with key %s of %s', i + 1, len(keys))
+
+            if LooseVersion(self.version) >= LooseVersion('2020a'):
+                self.log.debug('Version is %s - using binary installer method', self.version)
+                try:
+                    with tempfile.NamedTemporaryFile() as fd:
+                        tmp_configfile = fd.name
+                        with open(self.configfile) as template_fd:
+                            tmp_config = template_fd.read()
+                        tmp_config = tmp_config.replace('# fileInstallationKey=', 'fileInstallationKey=%s' % key)
```
orbsmiv, line 155: We had an internal discussion about whether to use a string replacement here, or maintain uniformity with the config section earlier in the block and use regex substitution.

akesandgren: A regexp is much safer to use. Please change to that.

easybuild/easyblocks/m/matlab.py
```diff
@@ -141,24 +141,56 @@ def install_step(self):
...
-            ])
-            (out, _) = run_cmd(cmd, log_all=True, simple=False)
+        for i, key in enumerate(keys):
+
+            self.log.debug('Installing with key %s of %s', i + 1, len(keys))
+
+            if LooseVersion(self.version) >= LooseVersion('2020a'):
+                self.log.debug('Version is %s - using binary installer method', self.version)
+                try:
+                    with tempfile.NamedTemporaryFile() as fd:
```
akesandgren, line 151: We might just as well move the whole section that creates self.configfile down into the installation step, and always regex in the file installation key into self.configfile.
That avoids creating a bunch of temp files.
But, with the file installation key written to a file, it must be protected from prying eyes and properly removed afterwards!

orbsmiv: @akesandgren Looking for a neat way to do this but in the case of versions < 2020a there's no requirement to regex the file installation key and the existing config file creation step is sufficient. (Earlier versions' installers take the `-fileInstallationKey` option.)
I could therefore keep the config file creation as is but make a regex for the _key_ line `^(# )?fileInstallationKey=.*` in the install step so that it can substitute and overwrite the installation key in place and avoid the temp file creation.

orbsmiv:
Looks like there was an issue with `apt-get` for some of the GItHub action steps, hence the failures on the checks for cc92b12

orbsmiv:
Running a close/open on this PR to retrigger the GitHub Actions

easybuild/easyblocks/m/matlab.py
```diff
@@ -158,48 +158,22 @@ def install_step(self):
...
+
+            try:
+                config = read_file(self.configfile)
+                config = regkey.sub("fileInstallationKey=%s" % key, config)
+                write_file(self.configfile, config)
+
+            except IOError as err:
+                raise EasyBuildError("Failed to update config file %s: %s", self.configfile, err)
+
+            self.log.debug('configuration file updated with installation key:\n %s', config)
```
akesandgren, line 173: This would print the keys in the log which you don't want to do since they should be protected from prying eyes.

Just drop the whole log.debug line. There's enough info in the self.log.info line above in my opinion.

orbsmiv: Dropped in 8884b60

akesandgren:
Going in, thanks @orbsmiv!

PR #2059, opened 2020-05-17
fix failing sanity check for GCCcore 5.x + sanity check under --module-only for all GCCcore versions

### LINKED ISSUE #1588: GCCcore sanity check fails under --module-only
Example:

```
$ eb GCCcore-7.3.0.eb --module-only
...
== FAILED: Installation ended unsuccessfully (build directory: /home/example/easybuild/build/GCCcore/7.3.0/dummy-): build failed (first 300 chars): Sanity check failed: no file found at 'libexec/gcc/x86_64-unknown-linux-gnu/7.3.0/lto1' or 'lib/gcc/x86_64-unknown-linux-gnu/7.3.0/lto1' in /home/example/easybuild/software/GCCcore/7.3.0
```

```
$ ls /home/example/easybuild/software/GCCcore/7.3.0/libexec/gcc
x86_64-pc-linux-gnu
```

Problem is that `self.platform_lib` in the `GCC` easyblock is set based on the result of the `get_platform_name` from `easybuild.tools.systemtools`. This value gets overwritten with the correct value in `configure_step` (based on the `config.guess` script included in the build directory), but this is skipped entirely under `--module-only`...

This can be fixed by simply checking the name of the subdirectory of `libexec/gcc` and `lib/gcc`, rather than deriving it from either `config.guess` or `get_platform_name`.

boegel:
The changes in #2033 broken the sanity check for `GCCcore-5.3.0.eb`, `GCCcore-5.4.0.eb`, `GCCcore-5.5.0.eb` (while older versions like `GCCcore-4.9.3.eb` and newer versions are still fine): 

```
== sanity checking...                                                                                                                                                                                         ...
  >> file 'libexec/gcc/x86_64-pc-linux-gnu/5.4.0/lto1' or 'lib/gcc/x86_64-pc-linux-gnu/5.4.0/lto1' found: FAILED
  >> file 'libexec/gcc/x86_64-pc-linux-gnu/5.4.0/lto-wrapper' or 'lib/gcc/x86_64-pc-linux-gnu/5.4.0/lto-wrapper' found: FAILED
  >> file 'libexec/gcc/x86_64-pc-linux-gnu/5.4.0/liblto_plugin.so' or 'lib/gcc/x86_64-pc-linux-gnu/5.4.0/liblto_plugin.so' found: FAILED
...
  >> (non-empty) directory 'lib/gcc/x86_64-pc-linux-gnu/5.4.0' found: FAILED                                                                                                                                 
```

The issue is that even though the updated `config.guess` suggests that the `x86_64-pc-linux-gnu` should be used, the `GCCcore` installation still has `x86_64-unknown-linux-gnu`.

In addition, it made the problem with `--module-only` worse, since rather than just having a failing sanity check (see #1588), the GCC easyblock now produces a hard crash because `self.platform_lib` is never defined when using `--module-only`:

```
$ eb GCCcore-5.4.0.eb --module-only
== sanity checking...
  File "/lib/python2.7/site-packages/easybuild_framework-4.2.1.dev0-py2.7.egg/easybuild/main.py", line 114, in build_and_install_software
  File "/lib/python2.7/site-packages/easybuild_framework-4.2.1.dev0-py2.7.egg/easybuild/framework/easyblock.py", line 3234, in build_and_install_one
  File "/lib/python2.7/site-packages/easybuild_framework-4.2.1.dev0-py2.7.egg/easybuild/framework/easyblock.py", line 3139, in run_all_steps
  File "/lib/python2.7/site-packages/easybuild_framework-4.2.1.dev0-py2.7.egg/easybuild/framework/easyblock.py", line 2994, in run_step
    step_method(self)()
  File "/lib/python2.7/site-packages/easybuild_easyblocks-4.2.1.dev0-py2.7.egg/easybuild/easyblocks/g/gcc.py", line 635, in sanity_check_step
    common_infix = os.path.join('gcc', self.platform_lib, self.version)
AttributeError: 'EB_GCC' object has no attribute 'platform_lib'
```

I've fixed both issues by slightly loosening up the sanity check in the `GCC` easyblock w.r.t. the "configuration name" subdirectory, which was already the proposed solution in #1588 .

We don't really care what this directory is called, just that it's there.

This allows us to drop the use of `self.platform_lib` entirely.

The updated `config.guess` (result of the changes in #2033) is still used by the GCC build process internally, so that's basically left untouched (we just don't use `config.guess` ourselves anymore now).

fixes #1588

### DIFF
--- a/easybuild/easyblocks/g/gcc.py
+++ b/easybuild/easyblocks/g/gcc.py
@@ -48,9 +48,10 @@ from easybuild.tools.config import build_option
 from easybuild.tools.filetools import apply_regex_substitutions, symlink, write_file
 from easybuild.tools.modules import get_software_root
 from easybuild.tools.run import run_cmd
-from easybuild.tools.systemtools import check_os_dependency, get_os_name, get_os_type, get_platform_name
+from easybuild.tools.systemtools import check_os_dependency, get_os_name, get_os_type
 from easybuild.tools.systemtools import get_gcc_version, get_shared_lib_ext
 from easybuild.tools.toolchain.compiler import OPTARCH_GENERIC
+from easybuild.tools.utilities import nub
 
 
 COMP_CMD_SYMLINKS = {
@@ -381,21 +382,13 @@ class EB_GCC(ConfigureMake):
         else:
             objdir = self.create_dir("obj")
 
+        # note: this also triggers the use of an updated config.guess script
+        # (unless both the 'build_type' and 'host_type' easyconfig parameters are specified)
         build_type, host_type = self.determine_build_and_host_type()
         if build_type:
             configopts += ' --build=' + build_type
         if host_type:
             configopts += ' --host=' + host_type
-
-        # instead of relying on uname, we use the actual --build option
-        if build_type:
-            self.platform_lib = build_type
-        else:
-            # Fallback
-            out, ec = run_cmd("../config.guess")
-            if ec != 0:
-                raise EasyBuildError('Failed to obtain platform_lib value from config.guess')
-            self.platform_lib = out.strip()
 
         # IV) actual configure, but not on default path
         cmd = "../configure  %s %s" % (self.configopts, configopts)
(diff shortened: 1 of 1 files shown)


boegel:
This makes `eb --module-only` works for all 22 `GCCcore` easyconfigs we have.

verdurin:
Going in, thanks @boegel!

PR #2064, opened 2020-05-21
added option use_pip_requirement to allow providing a requirements file to the pip command, as a source

mboisson:


boegel:
@mboisson Do you have a specific example use case for this?

mboisson:
In our case, installing our bundle modules which we install from wheels we compile independently. It is much easier to specify varying python dependencies with multi python using a requirements file than using the exts_list (plus the exts_list does not take from our wheel house).
  

  
  
  

  
  
>

mboisson:
Consider what we had to do in the past : 
https://github.com/ComputeCanada/easybuild-easyconfigs/blob/computecanada-master/easybuild/easyconfigs/s/Scipy-Stack/SciPy-Stack-2019b-dummy.eb

to what it looks like now : 
https://github.com/ComputeCanada/easybuild-easyconfigs/blob/computecanada-master/easybuild/easyconfigs/s/Scipy-Stack/SciPy-Stack-2020a-GCCcore-9.3.0.eb
plus
https://github.com/ComputeCanada/easybuild-easyconfigs/blob/computecanada-master/easybuild/easyconfigs/s/Scipy-Stack/requirements-2020a.txt

mboisson:
In general, the `exts_list` framework is not flexible enough to really support multi-python for python bundles, because the dependency graph can vary for different versions of python (in particular between python 2.7 and 3.x, but also sometimes within 3.x for different x).

mboisson:
This is another example : 
https://github.com/ComputeCanada/easybuild-easyconfigs/blob/computecanada-master/easybuild/easyconfigs/i/ipykernel/ipykernel-2020a-GCCcore-9.3.0.eb
https://github.com/ComputeCanada/easybuild-easyconfigs/blob/computecanada-master/easybuild/easyconfigs/i/ipykernel/requirements-2020a.txt

mboisson:
Any way to move this forward ?

akesandgren:
Doesn't --requirement take an argument? If so, where does that come from ?

mboisson:
Yes, the requirement file would be the source file in this case.

akesandgren:
Ok, then this should perhaps be noted in the help text for use_pip_requirement. I.e. that it takes the req-file from the sources argument.

mboisson:
> Ok, then this should perhaps be noted in the help text for use_pip_requirement. I.e. that it takes the req-file from the sources argument.

done

akesandgren (approved):
LGTM

akesandgren:
Going in, thanks @mboisson!

PR #2075, opened 2020-06-08
make sure PythonBundle uses correct 'python' command in extensions filter + also add lib64/python*/site-packages to $PYTHONPATH (if it exists)

boegel:
This is only required for installations that don't use a proper Python dependency, where `python` was not the command used for installing it (but `python3` was used instead), see for example https://github.com/easybuilders/easybuild-easyconfigs/pull/10754.

cc @teojgo

### DIFF
--- a/easybuild/easyblocks/generic/pythonbundle.py
+++ b/easybuild/easyblocks/generic/pythonbundle.py
@@ -112,6 +112,12 @@ class PythonBundle(Bundle):
 
         self.pylibdir = det_pylibdir(python_cmd=python_cmd)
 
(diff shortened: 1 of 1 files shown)


wpoely86:
I'm missing a test for this?

boegel:
@wpoely86 see https://github.com/easybuilders/easybuild-easyconfigs/pull/10754

boegel:
Fixes confirmed with test reports in https://github.com/easybuilders/easybuild-easyconfigs/pull/10754, @wpoely86 this should be good to go now...

PR #2079, opened 2020-06-14
fix build environment for WRF by patching shebang in compile script

lexming:
Currently the easyblock for WRF executes the `compile` script within its own `tcsh` shell. This completely resets the environment and the build proceeds exclusively with what is set in `configure.wrf`. Therefore, all executables and libraries set by name are taken from the system.

This PR fixes the shebang of the `compile` script so that it can be executed in the EB environment and also removes the hardcoded path for the preprocessor.

Should fix https://github.com/easybuilders/easybuild-easyconfigs/issues/10409

### DIFF
--- a/easybuild/easyblocks/w/wrf.py
+++ b/easybuild/easyblocks/w/wrf.py
@@ -256,7 +256,7 @@ class EB_WRF(EasyBlock):
         cmpscript = os.path.join(self.start_dir, 'compile')
         cmpsh_root = get_software_root('tcsh')
         if cmpsh_root:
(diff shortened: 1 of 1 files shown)


lexming:
* Positive test report with `WRF-4.0.2-foss-2018b-dmpar.eb`
https://gist.github.com/lexming/e197ae9e6a832472e90ac4f3924ce935

* Positive test report with `WRF-4.1.3-intel-2019b-dmpar.eb`
https://gist.github.com/lexming/179cb55e0a04436cd746d28a0f5e9a01

PR #2080, opened 2020-06-17
fix order of intel64/lib paths in generated module for impi (and drop intel64/lib/release_mt)

ocaisa:
This reverts https://github.com/easybuilders/easybuild-easyblocks/commit/0874dafb46550cbedc854d2f92d26422ac206121

### DIFF
--- a/easybuild/easyblocks/i/impi.py
+++ b/easybuild/easyblocks/i/impi.py
@@ -261,7 +261,7 @@ EULA=accept
             if LooseVersion(self.version) >= LooseVersion('2019'):
                 # The "release" library is default in v2019. Give it precedence over intel64/lib.
                 # (remember paths are *prepended*, so the last path in the list has highest priority)
-                lib_dirs = ['intel64/%s' % x for x in ['lib/release_mt', 'lib/release', 'lib']]
+                lib_dirs = ['intel64/%s' % x for x in ['lib', 'lib/release']]
                 include_dirs = ['intel64/include']
                 path_dirs = ['intel64/bin']
                 if self.cfg['ofi_internal']:
--- a/easybuild/easyblocks/i/impi.py
+++ b/easybuild/easyblocks/i/impi.py
@@ -259,7 +259,8 @@ EULA=accept
         else:
             guesses = {}
             if LooseVersion(self.version) >= LooseVersion('2019'):
-                # Keep release_mt and release in front, to give them priority over the possible symlinks in intel64/lib.
+                # Keep release_mt and release in front, to give priority to the possible symlinks in intel64/lib.
+                # (remember paths are *prepended*, so the last path in the list has highest priority)
                 # IntelMPI 2019 changed the default library to be the non-mt version.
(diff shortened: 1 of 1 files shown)


lexming:
(repeating my comment in the related issue)

The current comment is correct, at least regarding the goal of the list in `lib_dirs`, `release` or `release_mt` should precede `lib`. That's what Intel does.

Please check the script `$EBROOTIMPI/intel64/bin/mpivars.sh` in the installation of Intel MPI. This scripts is used to switch from the different release libraries. In line 128, you can see that the path to `release` or `relase_mt` precedes `lib`.

easybuild/easyblocks/i/impi.py
```diff
@@ -259,7 +259,8 @@ def make_module_req_guess(self):
         else:
             guesses = {}
             if LooseVersion(self.version) >= LooseVersion('2019'):
-                # Keep release_mt and release in front, to give them priority over the possible symlinks in intel64/lib.
+                # Keep release_mt and release in front, to give priority to the possible symlinks in intel64/lib.
+                # (remember paths are *prepended*, so the last path in the list has highest priority)
                 # IntelMPI 2019 changed the default library to be the non-mt version.
                 lib_dirs = ['intel64/%s' % x for x in ['lib/release_mt', 'lib/release', 'lib']]
```
lexming, line 265: Once https://github.com/easybuilders/easybuild-framework/pull/3367 is merged and order is consistently reversed, this should be the order for `lib_dirs` (`release_mt` should be removed, only one of `release` or `release_mt` is needed)
```suggestion
                lib_dirs = ['intel64/%s' % x for x in ['lib', 'lib/release']]
```

easybuild/easyblocks/i/impi.py
```diff
@@ -259,7 +259,8 @@ def make_module_req_guess(self):
         else:
             guesses = {}
             if LooseVersion(self.version) >= LooseVersion('2019'):
-                # Keep release_mt and release in front, to give them priority over the possible symlinks in intel64/lib.
+                # Keep release_mt and release in front, to give priority to the possible symlinks in intel64/lib.
+                # (remember paths are *prepended*, so the last path in the list has highest priority)
                 # IntelMPI 2019 changed the default library to be the non-mt version.
```
lexming, line 264:
```suggestion
                # The "release" library is default in v2019. Give it precedence over intel64/lib.
                # (remember paths are *prepended*, so the last path in the list has highest priority)
```

ocaisa:
@damianam Can you comment on this requested change? I know that it would have an impact on the `impi` installations at JSC because of what we do in `postinstallcmds` (but that is a pretty easy fix)

lexming:
Quick test with `impi-2019.7.217-iccifort-2020.1.217.eb` is fine.
These are the resulting `prepend_path` instructions in the `impi` module
```
[...]
prepend_path("LD_LIBRARY_PATH", pathJoin(root, "intel64/lib"))
prepend_path("LD_LIBRARY_PATH", pathJoin(root, "intel64/lib/release"))
prepend_path("LD_LIBRARY_PATH", pathJoin(root, "intel64/libfabric/lib"))
[...]
```

PR #2081, opened 2020-06-17
fix bug in PythonBundle: always add */site-packages subdirs to $PYTHONPATH when installing extensions

boegel:
This fixes a bug that was introduced in #2075.

Installing `SciPy-bundle` is broken currently for example, the installation of the `scipy` extensions fails because `numpy` can not be found, even though it is installed correctly.

### DIFF
--- a/easybuild/easyblocks/generic/pythonbundle.py
+++ b/easybuild/easyblocks/generic/pythonbundle.py
@@ -134,9 +134,17 @@ class PythonBundle(Bundle):
         if self.multi_python:
             txt += self.module_generator.prepend_paths(EBPYTHONPREFIXES, '')
         else:
-            for pylibdir in self.all_pylibdirs:
-                if os.path.exists(os.path.join(self.installdir, pylibdir)):
+
+            # the temporary module file that is generated before installing extensions
+            # must add all subdirectories to $PYTHONPATH without checking existence,
+            # otherwise paths will be missing since nothing is there initially
+            if self.current_step == 'extensions':
+                for pylibdir in self.all_pylibdirs:
                     txt += self.module_generator.prepend_paths('PYTHONPATH', pylibdir)
+            else:
+                for pylibdir in self.all_pylibdirs:
+                    if os.path.exists(os.path.join(self.installdir, pylibdir)):
+                        txt += self.module_generator.prepend_paths('PYTHONPATH', pylibdir)
 
         return txt
 
--- a/easybuild/easyblocks/generic/pythonbundle.py
+++ b/easybuild/easyblocks/generic/pythonbundle.py
@@ -139,12 +139,12 @@ class PythonBundle(Bundle):
             # must add all subdirectories to $PYTHONPATH without checking existence,
             # otherwise paths will be missing since nothing is there initially
             if self.current_step == 'extensions':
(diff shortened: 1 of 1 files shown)


ocaisa:
Now that I see this, I'm a bit concerned about https://github.com/easybuilders/easybuild-framework/blob/develop/easybuild/framework/easyblock.py#L1405
Are there cases (such as a bundle) where this will come back to bite us?

easybuild/easyblocks/generic/pythonbundle.py
```diff
@@ -134,9 +134,17 @@ def make_module_extra(self, *args, **kwargs):
...
+            # the temporary module file that is generated before installing extensions
+            # must add all subdirectories to $PYTHONPATH without checking existence,
+            # otherwise paths will be missing since nothing is there initially
+            if self.current_step == 'extensions':
+                for pylibdir in self.all_pylibdirs:
                     txt += self.module_generator.prepend_paths('PYTHONPATH', pylibdir)
+            else:
+                for pylibdir in self.all_pylibdirs:
+                    if os.path.exists(os.path.join(self.installdir, pylibdir)):
+                        txt += self.module_generator.prepend_paths('PYTHONPATH', pylibdir)
```
lexming, line 147: Minor suggestion to reduce the amount of duplicated lines of code
```suggestion
            if self.current_step == 'extensions':
                new_pylibdirs = self.all_pylibdirs
            else:
                new_pylibdirs = [l for l in self.all_pylibdirs if os.path.exists(os.path.join(self.installdir, l))]

            for pylibdir in new_pylibdirs:
                txt += self.module_generator.prepend_paths('PYTHONPATH', pylibdir)
```

lexming:
* Failed build of Scipy-bundle using current `develop`
https://gist.github.com/lexming/5808a464c1a38771fd9d697df71f9d6c
* Successful build of Scipy-bundle using this PR
https://gist.github.com/lexming/53fc1fb229630f6c1eda0ebeb996e099

lexming:
* Successful build of `matplotlib` v3.2.1 using this PR
https://gist.github.com/lexming/3dc52f3a0b51a91add1591b221fad29c
* Successful build of `IPython` v7.9.0 using this PR
https://gist.github.com/lexming/b4dde86e2d6cbef9d94645d54037eb50

boegel:
> Now that I see this, I'm a bit concerned about https://github.com/easybuilders/easybuild-framework/blob/develop/easybuild/framework/easyblock.py#L1405
> Are there cases (such as a bundle) where this will come back to bite us?

If that's a problem then it should make the sanity check fail due to missing stuff in the module file?

In this particular case, we're only making sure that the module file that is used to set up the environment before installing extensions is correct.

The module file used during the sanity check is a fresh one, which should have the same contents as the final module file (since nothing is changed anymore when the sanity check kicks off, that's after (post)install and extensions steps).

boegel:
Fix confirmed with `eb SciPy-bundle-2020.03-foss-2020a-Python-3.8.2.eb -f --include-easyblocks-from-pr 2081` after committing @lexming's suggestion, so this should be good to go...

lexming:
Going in, thanks @boegel !

PR #2082, opened 2020-06-25
set $XDG_CACHE_HOME to avoid abuse of /Users/kehoste/.cache/pip when installing DOLFIN

boegel:


akesandgren:
Going in, thanks @boegel!

PR #2087, opened 2020-06-29
rename include-fixed subdirectory in GCC installation to avoid problems after OS upgrade

boegel:
`limits.h` and `syslimits.h` are copied first since those are required, see also https://git.centos.org/rpms/gcc-toolset-10-gcc/blob/c8s/f/SPECS/gcc.spec#_1742

fixes https://github.com/easybuilders/easybuild-easyconfigs/issues/10666

### DIFF
--- a/easybuild/easyblocks/g/gcc.py
+++ b/easybuild/easyblocks/g/gcc.py
@@ -660,7 +660,8 @@ class EB_GCC(ConfigureMake):
                 move_file(include_fixed_path, include_fixed_renamed)
                 self.log.info("%s renamed to %s", include_fixed_path, include_fixed_renamed)
             else:
-                raise EasyBuildError("Exactly one 'include-fixed' directory expected, found %d: %s", len(matches), matches)
+                raise EasyBuildError("Exactly one 'include-fixed' directory expected, found %d: %s",
+                                     len(matches), matches)
         else:
             self.log.info("No include-fixed subdirectory found at %s", glob_pattern)
 
--- a/easybuild/easyblocks/g/gcc.py
+++ b/easybuild/easyblocks/g/gcc.py
@@ -45,7 +45,7 @@ from easybuild.easyblocks.generic.configuremake import ConfigureMake
 from easybuild.framework.easyconfig import CUSTOM
 from easybuild.tools.build_log import EasyBuildError
 from easybuild.tools.config import build_option
-from easybuild.tools.filetools import apply_regex_substitutions, symlink, write_file
+from easybuild.tools.filetools import apply_regex_substitutions, copy_file, move_file, symlink, write_file
 from easybuild.tools.modules import get_software_root
 from easybuild.tools.run import run_cmd
 from easybuild.tools.systemtools import check_os_dependency, get_os_name, get_os_type
@@ -606,6 +606,7 @@ class EB_GCC(ConfigureMake):
         """
         super(EB_GCC, self).post_install_step(*args, **kwargs)
 
+        # Add symlinks for cc/c++/f77/f95.
         bindir = os.path.join(self.installdir, 'bin')
         for key in COMP_CMD_SYMLINKS:
             src = COMP_CMD_SYMLINKS[key]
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/g/gcc.py
```diff
@@ -618,6 +619,52 @@ def post_install_step(self, *args, **kwargs):
             else:
                 raise EasyBuildError("Can't link '%s' to non-existing location %s", target, os.path.join(bindir, src))
 
+        # Rename include-fixed directory which includes system header files that were processed by fixincludes,
```
terjekv, line 622: Should we make this block specific to certain versions of GCC, or are we expecting them all to work okay with this level of modification?

easybuild/easyblocks/g/gcc.py
```diff
@@ -618,6 +619,52 @@ def post_install_step(self, *args, **kwargs):
...
+                retained_header_files = ['limits.h', 'syslimits.h']
+                for fn in retained_header_files:
+                    from_path = os.path.join(include_fixed_path, fn)
+                    to_path = os.path.join(include_path, fn)
+                    if os.path.exists(from_path):
+                        if os.path.exists(to_path):
+                            raise EasyBuildError("%s already exists, not overwriting it with %s!", to_path, from_path)
+                        else:
+                            copy_file(from_path, to_path)
+                            self.log.info("%s copied to %s", from_path, to_path)
```
terjekv, line 643: Add the context to the log with regards? Something like "GCC include modifications:"?

easybuild/easyblocks/g/gcc.py
```diff
@@ -618,6 +619,52 @@ def post_install_step(self, *args, **kwargs):
...
+                    else:
+                        self.log.warning("Can't copy non-existing file %s to %s", from_path, to_path)
+
+                readme = os.path.join(include_fixed_path, 'README.easybuild')
+                readme_txt = '\n'.join([
+                    "This directory was renamed by EasyBuild to avoid that the header files in it are picked up,",
+                    "since they may cause problems when the OS is upgraded to a new (minor) version.",
+                    '',
+                    "A couple of files were copied to %s first: %s" % (include_path, ', '.join(retained_header_files)),
```
terjekv, line 652: "These files were copied to %s first: %s"?

easybuild/easyblocks/g/gcc.py
```diff
@@ -618,6 +619,52 @@ def post_install_step(self, *args, **kwargs):
...
+                    '',
+                    "See https://github.com/easybuilders/easybuild-easyconfigs/issues/10666 for more information.",
+                    '',
+                ])
+                write_file(readme, readme_txt)
+
+                include_fixed_renamed = include_fixed_path + '.renamed-by-easybuild'
+                move_file(include_fixed_path, include_fixed_renamed)
+                self.log.info("%s renamed to %s", include_fixed_path, include_fixed_renamed)
```
terjekv, line 661: Context again? This would seem terrifying if I read it in a log without knowing why, and knowing why shouldn't require tribal knowledge.  :-)

easybuild/easyblocks/g/gcc.py
```diff
@@ -618,6 +619,52 @@ def post_install_step(self, *args, **kwargs):
...
+            else:
+                raise EasyBuildError("Exactly one 'include-fixed' directory expected, found %d: %s",
+                                     len(matches), matches)
+        else:
+            self.log.info("No include-fixed subdirectory found at %s", glob_pattern)
+
     def sanity_check_step(self):
```
terjekv, line 668: I don't think we can realistically expect GCC to ever pass a sanity check any more...

easybuild/easyblocks/g/gcc.py
boegel: To which versions should it be specific then? It applies to any version, no?

terjekv: I have no idea. I was more concerned with limiting scope for such a change. Do we know for what versions the change was done by Red Hat? The package we got linked was created for 9.*, so that's fairly recent. We know it's a problem on 8.3, but do we know if 7.* is affected?

boegel: I see `include-fixed` even as far back as GCC 4.9.2, so I don't think there's a point in adding a version check...

Also, looks like Red Hat has been trashing the stuff in `include-fixed` for a while, see https://git.centos.org/rpms/devtoolset-3-gcc/blob/611c22b0e52a80f78ab2682e7ab33b1e9b908f46/f/SPECS/gcc.spec#_1531, which goes back as far as Oct'14 (GCC 4.9.2)

terjekv: Well, that's terrifying. No version check it is. :-)

PR #2088, opened 2020-07-01
use 'wclean -platform' instead of 'wcleanPlatform' for OpenFOAM v2006 & newer

boegel:
see also https://develop.openfoam.com/Development/openfoam/-/wikis/guides/upgrade/v2006-User-Upgrade-Guide#additional-wmake-subcommands


olesenm:
Thanks for catching this. Didn't realize anyone (other than us) actually used this command.

PR #2090, opened 2020-07-06
Code Saturne CFD

surak:
https://www.code-saturne.org/cms/

Easyblock for it

### DIFF
--- a/easybuild/easyblocks/c/code_saturne.py
+++ b/easybuild/easyblocks/c/code_saturne.py
@@ -1,0 +1,167 @@
+##
+# Copyright 2009-2018 Ghent University
+#
+# This file is part of EasyBuild,
+# originally created by the HPC team of Ghent University (http://ugent.be/hpc/en),
+# with support of Ghent University (http://ugent.be/hpc),
+# the Flemish Supercomputer Centre (VSC) (https://www.vscentrum.be),
+# Flemish Research Foundation (FWO) (http://www.fwo.be/en)
+# and the Department of Economy, Science and Innovation (EWI) (http://www.ewi-vlaanderen.be/en).
+#
+# https://github.com/easybuilders/easybuild
+#
+# EasyBuild is free software: you can redistribute it and/or modify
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/c/code_saturne.py
```diff
@@ -0,0 +1,167 @@
...
+from easybuild.framework.easyblock import EasyBlock
+from easybuild.framework.easyconfig.default import CUSTOM
+from easybuild.tools.filetools import apply_regex_substitutions
+from easybuild.tools.run import run_cmd
+from easybuild.tools.systemtools import get_shared_lib_ext
+
+import easybuild.tools.environment as env
+
+
+class EB_Code_underscore_Saturne(EasyBlock):
```
boegel, line 43: @surak Wouldn't it be better to derive from `ConfigureMake`, that would simplify things below quite a bit (there's no need to customize `build_step` and `install_step` for example in that case...

easybuild/easyblocks/c/code_saturne.py
```diff
@@ -0,0 +1,167 @@
...
+        self.log.info("Running ./sbin/bootstrap ...")
+        cmd = './sbin/bootstrap'
+        run_cmd(cmd, log_all=True, simple=True, log_output=True)
+
+        cmd = ' '.join([
+            self.cfg['preconfigopts'],
+            './configure',
+            '--prefix=' + self.installdir,
+            '--without-modules',
+            self.cfg['configopts'],
```
boegel, line 78: It would be better to update `configopts` (add `--without-modules`, conditionally add `--enable-debug`) and call out to the parent when deriving from `ConfigureMake`).

Something like this:

```python
# add comment here to explain why --without-modules is added
self.cfg.update('configopts', '--without-modules')

if self.cfg['debug']:
    self.cfg.update('configopts', '--enable-debug')

super(EB_Code_underscore_Saturne, self).configure_step()
```

There's no point in returning the output (since there's nothing to consume it)

Issue #2092, opened 2020-07-07
Bazel & TensorFlow easyblocks expect to find include-fixed subdirectory in GCC installation


boegel:
```
$ eb Bazel-2.0.0-GCCcore-8.3.0.eb
...
== FAILED: Installation ended unsuccessfully (build directory: /tmp/Bazel/2.0.0/GCCcore-8.3.0):
build failed (first 300 chars): Derived directory /software/GCCcore/8.3.0/lib/gcc/x86_64-pc-linux-gnu/8.3.0/include-fixed does not exist (took 9 sec)
```

This issue is caused by the changes in #2087.

boegel:
fixed in #2093 and #2094

PR #2093, opened 2020-07-07
don't require include-fixed subdirectory in GCC installation for building TensorFlow, only use it when it's there

boegel:
(partial fix for #2092)

### DIFF
--- a/easybuild/easyblocks/t/tensorflow.py
+++ b/easybuild/easyblocks/t/tensorflow.py
@@ -360,32 +360,38 @@ class EB_TensorFlow(PythonPackage):
         else:
             raise EasyBuildError("Failed to determine installation prefix for binutils")
 
+        inc_paths, lib_paths = [], []
+
         gcc_root = get_software_root('GCCcore') or get_software_root('GCC')
         if gcc_root:
             gcc_lib64 = os.path.join(gcc_root, 'lib64')
+            lib_paths.append(gcc_lib64)
+
             gcc_ver = get_software_version('GCCcore') or get_software_version('GCC')
 
             # figure out location of GCC include files
             res = glob.glob(os.path.join(gcc_root, 'lib', 'gcc', '*', gcc_ver, 'include'))
             if res and len(res) == 1:
                 gcc_lib_inc = res[0]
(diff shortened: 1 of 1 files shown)


boegel:
Confirmed fix for

```
== FAILED: Installation ended unsuccessfully (build directory: /tmp/TensorFlow/1.15.0/foss-2019b-Python-3.7.4):
build failed (first 300 chars): Derived directory /software/GCCcore/8.3.0/lib/gcc/x86_64-pc-linux-gnu/8.3.0/include-fixed does not exist
```

easybuild/easyblocks/t/tensorflow.py
```diff
@@ -360,33 +360,39 @@ def build_step(self):
...
+            else:
+                self.log.info("Derived directory %s does not exist, so discarding it", gcc_lib_inc_fixed)
 
             # also check on location of include/c++/<gcc version> directory
             gcc_cplusplus_inc = os.path.join(gcc_root, 'include', 'c++', gcc_ver)
-            if not os.path.exists(gcc_cplusplus_inc):
+            if os.path.exists(gcc_cplusplus_inc):
+                inc_paths.append(gcc_cplusplus_inc)
+            else:
                 raise EasyBuildError("Derived directory %s does not exist", gcc_cplusplus_inc)
```
wpoely86, line 392: because it's just a info message, here it's fatal?

boegel: It was fatal before, I don't see a reason to change it?

We now know that `include-fixed` may not be there, and that it's acceptable. It's not acceptable for this path though, so if it's not there, we want to know about it (in case the path changes or something in newer GCC versions).

lexming:
Positive test with `TensorFlow-1.15.0-foss-2019b-Python-3.7.4.eb`
https://gist.github.com/lexming/45166ce3f27972c6129cea6bbf7a3600

Positive test with `TensorFlow-2.2.0-foss-2019b-Python-3.7.4.eb`
https://gist.github.com/lexming/c45eca94796804936e9badac80158254

lexming:
Going in, thanks @boegel !

PR #2097, opened 2020-07-07
removed version check against the version.txt file, since it is no lo…

mboisson:
…nger honoured in 11.0.x


mboisson:
@Micket for your review

PR #2100, opened 2020-07-12
add support for v1.3.2 to motioncor2

lexming:


akesandgren:
Going in, thanks @lexming!

PR #2101, opened 2020-07-15
filter out Python in SLEPc configure

lexming:
Filter out Python from the dependency check in the configure step of `SLEPc`.
This is needed for https://github.com/easybuilders/easybuild-easyconfigs/pull/10908 to fix issue https://github.com/easybuilders/easybuild-easyconfigs/issues/10865

### DIFF
--- a/easybuild/easyblocks/s/slepc.py
+++ b/easybuild/easyblocks/s/slepc.py
@@ -78,7 +78,7 @@ class EB_SLEPc(ConfigureMake):
         self.log.debug('SLEPC_DIR: %s' % os.getenv('SLEPC_DIR'))
 
         # optional dependencies
-        depfilter = self.cfg.builddependencies() + ["PETSc"]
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/s/slepc.py
```diff
@@ -78,7 +78,7 @@ def configure_step(self):
         self.log.debug('SLEPC_DIR: %s' % os.getenv('SLEPC_DIR'))
 
         # optional dependencies
-        depfilter = self.cfg.builddependencies() + ["PETSc"]
+        depfilter = self.cfg.builddependencies() + ['PETSc', 'Python']
         deps = [dep['name'] for dep in self.cfg.dependencies() if not dep['name'] in depfilter]
```
boegel, line 82: @lexming Please change this to `if dep['name'] not in dep_filter` while you're at it

lexming: fixed :+1:

PR #2103, opened 2020-07-17
add 'lib' symlink in tbb installation directory when building from source

mboisson:
Otherwise, embree does not find tbb

### DIFF
--- a/easybuild/easyblocks/t/tbb.py
+++ b/easybuild/easyblocks/t/tbb.py
@@ -164,13 +164,14 @@ class EB_tbb(IntelBase, ConfigureMake):
             # applications go looking into tbb/lib so we move what's in there to libs
             install_libpath = os.path.join(self.installdir, 'tbb', 'lib')
             shutil.move(install_libpath, os.path.join(self.installdir, 'tbb', 'libs'))
+            # and symlink the right lib from /tbb/libs/intel64/... to lib
             os.symlink(os.path.join(self.installdir, self.libpath), install_libpath)
         else:
             # no custom install step when building from source (building is done in install directory)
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/t/tbb.py
```diff
@@ -162,7 +162,6 @@ def install_step(self):
             self.libpath = os.path.join(self.libpath, libdir)
             self.log.debug("self.libpath: %s" % self.libpath)
             # applications go looking into tbb/lib so we move what's in there to libs
-            # and symlink the right lib from /tbb/libs/intel64/... to lib
```
boegel, line 165: @mboisson Why remove this comment here, there's symlinking being done below?

The comment should state it's creating a `tbb/lib` symlink though?

easybuild/easyblocks/t/tbb.py
```diff
@@ -171,9 +170,13 @@ def install_step(self):
             cand_lib_paths = glob.glob(os.path.join(self.installdir, 'build', '*_release'))
             if len(cand_lib_paths) == 1:
                 self.libpath = os.path.join('build', os.path.basename(cand_lib_paths[0]))
+                # and symlink the right lib from /tbb/libs/intel64/... to lib
```
boegel, line 173: This comment is incorrect: the `and` start doesn't make sense, and it symlinks to a subdirectory of `build`?

mboisson:
@boegel, see if the comments make more sense now.

akesandgren:
@boegel ping...

boegel (approved):

boegel:
tested & approved, thanks @mboisson!

PR #2104, opened 2020-07-21
add custom easyblock for PyTorch

Flamefire:
This makes writing ECs for PyTorch much easier through the following features:

- Support for `cuda_compute_capabilities` just like TF
- Auto setting of `modulename`
- Test on (dummy) installation (there have been reports that tests fail in build tree but pass on installed module)
- Auto setting of `PYTORCH_BUILD_VERSION/NUMBER`
- Auto detection of dependencies and setting of setup config variables for both build and install step (no more duplication required)
- Auto exclude *NNPACK on power
- Avoid using the users home during test
- Detection of build time downloaded submodules
- Arch specific test exclusion (`-x` with no test is invalid, hence need to do this in an EasyBlock!)
- Set `CMAKE_PREFIX_PATH` for `find_package(Torch)` to work for user apps build against the PyTorch C++ API

Example: https://github.com/easybuilders/easybuild-easyconfigs/pull/11000

### DIFF
--- a/easybuild/easyblocks/p/pytorch.py
+++ b/easybuild/easyblocks/p/pytorch.py
@@ -123,6 +123,11 @@ class EB_PyTorch(PythonPackage):
 
         # Gather default options. Will be checked against (and can be overwritten by) custom_opts
         options = ['PYTORCH_BUILD_VERSION=' + self.version, 'PYTORCH_BUILD_NUMBER=1']
+
+        # enable verbose mode when --debug is used (to show compiler commands)
+        if build_option('debug'):
+            options.append('VERBOSE=1')
+
         # Restrict parallelism
         options.append('MAX_JOBS=%s' % self.cfg['parallel'])
 
--- a/easybuild/easyblocks/p/pytorch.py
+++ b/easybuild/easyblocks/p/pytorch.py
@@ -1,0 +1,232 @@
+##
+# Copyright 2020-2020 Ghent University
+#
+# This file is part of EasyBuild,
+# originally created by the HPC team of Ghent University (http://ugent.be/hpc/en),
+# with support of Ghent University (http://ugent.be/hpc),
+# the Flemish Supercomputer Centre (VSC) (https://www.vscentrum.be),
+# Flemish Research Foundation (FWO) (http://www.fwo.be/en)
+# and the Department of Economy, Science and Innovation (EWI) (http://www.ewi-vlaanderen.be/en).
+#
+# https://github.com/easybuilders/easybuild
+#
+# EasyBuild is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation v2.
+#
+# EasyBuild is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with EasyBuild.  If not, see <http://www.gnu.org/licenses/>.
+##
+"""
+EasyBuild support for building and installing PyTorch, implemented as an easyblock
+
+@author: Alexander Grund (TU Dresden)
+"""
+
+import os
+import re
+import tempfile
+from distutils.version import LooseVersion
+from easybuild.easyblocks.generic.pythonpackage import PythonPackage
+from easybuild.framework.easyconfig import CUSTOM
+from easybuild.tools.build_log import EasyBuildError
+from easybuild.tools.config import build_option
+import easybuild.tools.environment as env
+from easybuild.tools.modules import get_software_root
+from easybuild.tools.systemtools import POWER, get_cpu_architecture
+
+
+class EB_PyTorch(PythonPackage):
+    """Support for building/installing TensorFlow."""
+
+    @staticmethod
+    def extra_options():
+        extra_vars = PythonPackage.extra_options()
+        extra_vars.update({
+            'excluded_tests': [{}, 'Mapping of architecture strings to list of tests to be excluded', CUSTOM],
+            'custom_opts': [[], 'List of options for the build/install command. Can be used to change the defaults ' +
+                                'set by the PyTorch EasyBlock, for example ["USE_MKLDNN=0"].', CUSTOM]
+        })
+        extra_vars['download_dep_fail'][0] = True
+        extra_vars['sanity_pip_check'][0] = True
+
+        return extra_vars
+
+    def __init__(self, *args, **kwargs):
+        """Constructor for PyTorch easyblock."""
+        super(EB_PyTorch, self).__init__(*args, **kwargs)
+        self.options['modulename'] = 'torch'
+        # Test as-if pytorch was installed
+        self.testinstall = True
+        self.tmpdir = tempfile.mkdtemp(suffix='-pytorch-build')
+
+    def fetch_step(self, skip_checksums=False):
+        """Fetch sources for installing PyTorch, including those for tests."""
+        super(EB_PyTorch, self).fetch_step(skip_checksums)
+        # Resolve tests early to avoid failures later. Use obtain_file if path is not absolute
+        tests = [test if os.path.isabs(test) else self.obtain_file(test) for test in self.cfg['tests']]
+        self.cfg['tests'] = tests
+
+    @staticmethod
+    def get_dependency_options_for_version(pytorch_version):
+        """
+        PyTorch can enable some functionality based on available software or use system software instead of a submodule
+        This returns EasyBuild names of that and the flag that should be used when the dependency is found
+
+        The result is a list of tuples (enable_flag, eb_name)
+        """
+        pytorch_version = LooseVersion(pytorch_version)
+
+        def is_version_ok(version_range):
+            """Return True if the PyTorch version to be installed matches the version_range"""
+            min_version, max_version = version_range.split(':')
+            result = True
+            if min_version and pytorch_version < LooseVersion(min_version):
+                result = False
+            if max_version and pytorch_version >= LooseVersion(max_version):
+                result = False
+            return result
+
+        available_libs = (
+            # Format: (PyTorch flag to enable, EB name, '<min version>:<exclusive max version>')
+            # Use `None` for the EB name if no known EC exists
+            ('USE_FFMPEG=1', 'FFmpeg', '1.0.0:'),
+            ('USE_GFLAGS=1', 'gflags', '1.0.0:'),
+            ('USE_GLOG=1', 'glog', '1.0.0:'),
+
+            # For system libs check CMakeLists.txt, below `if(USE_SYSTEM_LIBS)`, order kept here
+            # NCCL handled specially as other env variables are requires for it
+            ('USE_SYSTEM_CPUINFO=1', None, '1.6.0:'),
+            ('USE_SYSTEM_SLEEF=1', None, '1.6.0:'),
+            ('USE_SYSTEM_GLOO=1', None, '1.6.0:'),
+            ('BUILD_CUSTOM_PROTOBUF=0', 'protobuf', '1.2.0:'),
+            ('USE_SYSTEM_EIGEN_INSTALL=1', 'Eigen', '1.0.0:'),
+            ('USE_SYSTEM_FP16=1', None, '1.6.0:'),
+            ('USE_SYSTEM_PTHREADPOOL=1', None, '1.6.0:'),
+            ('USE_SYSTEM_PSIMD=1', None, '1.6.0:'),
+            ('USE_SYSTEM_FXDIV=1', None, '1.6.0:'),
+            ('USE_SYSTEM_BENCHMARK=1', None, '1.6.0:'),  # Google Benchmark
+            ('USE_SYSTEM_ONNX=1', None, '1.6.0:'),
+            ('USE_SYSTEM_XNNPACK=1', None, '1.6.0:'),
+        )
+        return [(enable_opt, dep_name) for enable_opt, dep_name, version_range in available_libs
+                if is_version_ok(version_range)]
+
+    def configure_step(self):
+        """Custom configure procedure for PyTorch."""
+        super(EB_PyTorch, self).configure_step()
+
+        # Gather default options. Will be checked against (and can be overwritten by) custom_opts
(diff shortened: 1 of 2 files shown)


easybuild/easyblocks/p/pytorch.py
```diff
@@ -0,0 +1,176 @@
...
+    """Support for building/installing TensorFlow."""
+
+    @staticmethod
+    def extra_options():
+        extra_vars = PythonPackage.extra_options()
+        extra_vars.update({
+            # see https://developer.nvidia.com/cuda-gpus
+            'cuda_compute_capabilities': [[], 'List of CUDA compute capabilities to build with', CUSTOM],
+            'excluded_tests': [{}, 'Mapping of architecture strings to list of tests to be excluded', CUSTOM],
+            'extra_opts': [[], 'List of extra options, overwrites defaults from EasyBlock, e.g. USE_MKLDNN=0', CUSTOM]
```
boegel, line 53: There is no `extra_opts` default easyconfig parameter, so the "overwrites defaults from EasyBlock" doesn't make sense?

```suggestion
            'extra_opts': [[], 'List of extra options for build/install command, for example ['USE_MKLDNN=0']", CUSTOM]
```

easybuild/easyblocks/p/pytorch.py
```diff
@@ -0,0 +1,176 @@
...
+
+        known_dependencies = ('FFmpeg', 'gflags', 'glog')
+        dependency_names = set(dep['name'] for dep in self.cfg.dependencies())
+        for dep in known_dependencies:
+            if dep in dependency_names:
+                options.append('USE_%s=1' % dep.upper())
+
+        # Some dependencies that are bundled with PyTorch can be used from the system
+        # No known ECs: 'cpuinfo', 'sleef', 'gloo', 'FP16', 'pthreadpool', 'psimd', 'FXDiv', 'ONNX', 'XNNPack'
+        #               'Eigen', 'GoogleBenchmark'
```
boegel, line 96: There are easyconfigs for `Eigen`, so that one belongs in `known_dependencies`?

easybuild/easyblocks/p/pytorch.py
```diff
@@ -0,0 +1,176 @@
...
+        if 'protobuf' in dependency_names:
+            options.append('BUILD_CUSTOM_PROTOBUF=0')
+        # Always use Infiniband
+        options.append('USE_IBVERBS=1')
```
boegel, line 100: Oof, it doesn't seem smart to blindly hardcode that...

Two options here:

* Do it by default, but provide a way out. For example check whether `USE_IBVERBS=` is already specified in `options`, and only add `USE_IBVERBS=1` if it's not. That way, you can opt out by using `extra_opts = ['USE_IBVERBS=0']` in the easyconfig file. That does require moving up the `options.extend(self.cfg['extra_opts'])` to above this part, of course.
  That's probably a good idea to do for all stuff you add to `options`, so introducing a (local) utility function like `add_option('USE_IBVERBS', '1')` probably makes sense?

* Don't do this at all in the easyblock, and leave it up to the easyconfig file to set this via `extra_opts`.

easybuild/easyblocks/p/pytorch.py
```diff
@@ -0,0 +1,176 @@
...
+
+        cudnn_root = get_software_root('cuDNN')
+        if cudnn_root:
+            options.append('CUDNN_LIB_DIR=%s/lib64' % cudnn_root)
```
boegel, line 104: please use `os.path.join` when composing file paths:

```python
options.append('CUDNN_LIB_DIR=' + os.path.join(cudnn_root, 'lib64'))
```

(same below)

easybuild/easyblocks/p/pytorch.py
Flamefire: Done slightly differently: I'm post-processing the options list into a unique_options list at the end. Makes for more straight-forward code.

easybuild/easyblocks/p/pytorch.py
Flamefire: Good catch, thanks! I've reworked that a bit similar to what I do in TensorFlow by using a list with PyTorch versions for each of such EC dependency options. This way the code is kinda self-documenting and easier to extent.

easybuild/easyblocks/p/pytorch.py
Flamefire: Changed to `'List of extra options for build/install command. Can be used to change the defaults set by the PyTorch EasyBlock, for example ["USE_MKLDNN=0"].'`

easybuild/easyblocks/p/pytorch.py
```diff
@@ -0,0 +1,229 @@
...
+        if nccl_root:
+            options.append('USE_SYSTEM_NCCL=1')
+            options.append('NCCL_INCLUDE_DIR=' + os.path.join(nccl_root, 'include'))
+
+        # list of CUDA compute capabilities to use can be specifed in two ways (where (2) overrules (1)):
+        # (1) in the easyconfig file, via the custom cuda_compute_capabilities;
+        # (2) in the EasyBuild configuration, via --cuda-compute-capabilities configuration option;
+        cuda_cc = build_option('cuda_compute_capabilities') or self.cfg['cuda_compute_capabilities']
+        if not cuda_cc:
+            raise EasyBuildError('cuda_compute_capabilities must be specified')
```
boegel, line 166:
```suggestion
            error_msg = "List of CUDA compute capabilities must be specified, either via "
            error_msg += "cuda_compute_capabilities easyconfig parameter or via --cuda-compute-capabilities")
            raise EasyBuildError(error_msg)
```

easybuild/easyblocks/p/pytorch.py
```diff
@@ -0,0 +1,229 @@
...
+    """Support for building/installing TensorFlow."""
+
+    @staticmethod
+    def extra_options():
+        extra_vars = PythonPackage.extra_options()
+        extra_vars.update({
+            # see https://developer.nvidia.com/cuda-gpus
+            'cuda_compute_capabilities': [[], 'List of CUDA compute capabilities to build with', CUSTOM],
+            'excluded_tests': [{}, 'Mapping of architecture strings to list of tests to be excluded', CUSTOM],
+            'extra_opts': [[], 'List of extra options for build/install command. Can be used to change the defaults ' +
```
boegel, line 54: Since the value for `extra_opts` allows to control some of the options set by default by the easyblock, maybe `custom_opts` is a better name?

Since they're not really *extra* options if the easyblock is also settings them by default with a different value...

easybuild/easyblocks/p/pytorch.py
```diff
@@ -0,0 +1,231 @@
+##
```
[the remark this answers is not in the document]
boegel: Go home @hound, you're drunk?

easybuild/easyblocks/p/pytorch.py
Flamefire: Done

akesandgren:
@boegel Is your change request done? If so clear state please.

easybuild/easyblocks/p/pytorch.py
```diff
@@ -0,0 +1,230 @@
...
+
+class EB_PyTorch(PythonPackage):
+    """Support for building/installing TensorFlow."""
+
+    @staticmethod
+    def extra_options():
+        extra_vars = PythonPackage.extra_options()
+        extra_vars.update({
+            # see https://developer.nvidia.com/cuda-gpus
+            'cuda_compute_capabilities': [[], 'List of CUDA compute capabilities to build with', CUSTOM],
```
boegel, line 52: This line can be removed thanks to https://github.com/easybuilders/easybuild-framework/pull/3382?

easybuild/easyblocks/p/pytorch.py
```diff
@@ -0,0 +1,230 @@
...
+            'excluded_tests': [{}, 'Mapping of architecture strings to list of tests to be excluded', CUSTOM],
+            'custom_opts': [[], 'List of options for the build/install command. Can be used to change the defaults ' +
+                                'set by the PyTorch EasyBlock, for example ["USE_MKLDNN=0"].', CUSTOM]
+        })
+        extra_vars['download_dep_fail'][0] = True
+        extra_vars['sanity_pip_check'][0] = True
+
+        return extra_vars
+
+    def __init__(self, *args, **kwargs):
```
boegel, line 62: Please add a docstring `"""Constructor for PyTorch easyblock."""`

easybuild/easyblocks/p/pytorch.py
```diff
@@ -0,0 +1,230 @@
...
+        super(EB_PyTorch, self).__init__(*args, **kwargs)
+        self.options['modulename'] = 'torch'
+        # Test as-if pytorch was installed
+        self.testinstall = True
+        self.tmpdir = tempfile.mkdtemp(suffix='-pytorch-build')
+
+    def fetch_step(self, skip_checksums=False):
```
boegel, line 69: Please add a docstring: `"""Fetch sources for installing PyTorch, including those for tests."'"

easybuild/easyblocks/p/pytorch.py
```diff
@@ -0,0 +1,230 @@
...
+        super(EB_PyTorch, self).fetch_step(skip_checksums)
+        # Resolve tests early to avoid failures later. Use obtain_file if path is not absolute
+        tests = [test if os.path.isabs(test) else self.obtain_file(test) for test in self.cfg['tests']]
```
boegel, line 72: Shouldn't we fail here if the absolute path does not exist?

easybuild/easyblocks/p/pytorch.py
```diff
@@ -0,0 +1,230 @@
...
+            ('USE_SYSTEM_XNNPACK=1', None, '1.6.0:'),
+        )
+        return [(enable_opt, dep_name) for enable_opt, dep_name, version_range in available_libs
+                if is_version_ok(version_range)]
+
+    def configure_step(self):
+        """Custom configure procedure for PyTorch."""
+        super(EB_PyTorch, self).configure_step()
+
+        options = ['PYTORCH_BUILD_VERSION=' + self.version, 'PYTORCH_BUILD_NUMBER=1']
```
boegel, line 124: Should we rename `options` to `default_opts`, to highlight that you can override them via the `custom_opts` easyconfig parameter?

Also add `VERBOSE=1`, perhaps only when `--debug` is used?

easybuild/easyblocks/p/pytorch.py
```diff
@@ -0,0 +1,230 @@
...
+        # Restrict parallelism
+        options.append('MAX_JOBS=%(parallel)s')
```
boegel, line 126: Better just grab the value for `'parallel` directly here (it's always properly set by this point)
```suggestion
        options.append('MAX_JOBS=' + self.cfg['parallel'])

File: easybuild/easyblocks/p/pytorch.py [hunk 14]
```diff
@@ -0,0 +1,230 @@
...
+
+        # BLAS Interface
+        if get_software_root('imkl'):
+            options.append('BLAS=MKL')
+        else:
+            # This is what PyTorch defaults to if no MKL is found. Make this explicit here
+            options.append('BLAS=Eigen')
```
boegel, line 133: Shouldn't this imply that `Eigen` is included as a dependency, so check the value of `get_software_root('Eigen')`?

easybuild/easyblocks/p/pytorch.py
```diff
@@ -0,0 +1,230 @@
...
+            if dep_name is None:
+                continue
+            if dep_name in dependency_names:
+                options.append(enable_opt)
+            else:
+                not_used_dep_names.append(dep_name)
+        self.log.info('Did not enable options for the following dependencies as they are not used in the EC: %s',
+                      not_used_dep_names)
+
+        # Always use Infiniband
```
boegel, line 148: Clarify the comment how to opt out?

```
# always use Infiniband by default
# you can opt out of this by defining a different value via 'custom_opts' in the easyconfig file:
# custom_opts = "USE_IBVERBS=0"

boegel (changes requested):

easybuild/easyblocks/p/pytorch.py
Flamefire: What if the test will be extracted later? If an absolute path is used I wouldn't assume much at this point. Existence is checked later again anyway,

easybuild/easyblocks/p/pytorch.py
Flamefire: Adding a comment above: `Gather default options. Will be checked against (and can be overwritten by) custom_opts`. IMO it isn't worth highlighting that those are default_opts as it reduces readability.

Where is "VERBOSE=1" coming from? Is this supported by PyTorch buildsystem?

easybuild/easyblocks/p/pytorch.py
Flamefire: Just checked: They download eigen as a submodule and use that (or better: we do the download)
However we might be able to skip the download if we use Eigen as a dependency. Will try that. But for this it doesn't matter: It is either the system version or a silent fallback to submodule

easybuild/easyblocks/p/pytorch.py
boegel: OK, but it kind of defeats the purpose of early checking. :)

But you're right, we shouldn't be too strict either here.

easybuild/easyblocks/p/pytorch.py
boegel: If it doesn't seem to do anything, ignore it, it may be an attempt at debugging to was left in place.

It was included in the PyTorch 1.4 easyconfig, see https://github.com/easybuilders/easybuild-easyconfigs/pull/11000/files#diff-5e116091061fb8243f4e109886113febL273

easybuild/easyblocks/p/pytorch.py
boegel: OK, fine then

easybuild/easyblocks/p/pytorch.py
Flamefire: I can try and see if it does anything

boegel: Using `VERBOSE=1` results in compiler commands actually being shown, I checked that.

So adding `VERBOSE=1` when `--debug` is used via a3a8d7130

boegel (approved):

boegel:
Tested with https://github.com/easybuilders/easybuild-easyconfigs/pull/11000, good to go now, thanks @Flamefire!

PR #2106, opened 2020-07-22
Fix MPI in netcdf4_python easyblock

Darkless012:
netcdf4 crashes if run without MPI (inlcluding tests).

If the MPI is used, wrap python commands.


smoors:
Going in, thanks @Darkless012!

PR #2107, opened 2020-07-22
patch ELPA's manual_cpp script to fix hardcoded /usr/bin/python

boegel:
ELPA requires Python as a build dep, but one of the scripts hardcodes `/usr/bin/python`...


lexming:
Tests in https://github.com/easybuilders/easybuild-easyconfigs/pull/11011#issuecomment-664339488

lexming:
Going in, thanks @boegel !

PR #2108, opened 2020-07-22
handle installation of pip & setuptools in Python 3.4+

Flamefire:
Python 3.4 comes with pip and setuptools. This is even required to install (the newer) pip in the EC so not having it is an error.

Furthermore installation of pip/setuptools fails if a distutils user config sets the `prefix` to a different directory as that would be where pip is then installed. See https://github.com/easybuilders/easybuild-easyconfigs/issues/11009
Hence check for that and report a (verbose) warning.

### DIFF
--- a/easybuild/easyblocks/p/python.py
+++ b/easybuild/easyblocks/p/python.py
@@ -200,7 +200,7 @@ class EB_Python(ConfigureMake):
 
         # Pip is included since 3.4 via ensurepip https://docs.python.org/3.4/whatsnew/changelog.html
         if LooseVersion(self.version) >= LooseVersion('3.4.0'):
-            # Default, but do it explicitely
+            # Default, but do it explicitly
             self.cfg.update('configopts', "--with-ensurepip=upgrade")
 
         modules_setup = os.path.join(self.cfg['start_dir'], 'Modules', 'Setup')
@@ -251,7 +251,7 @@ class EB_Python(ConfigureMake):
             self.cfg.update('configopts', "--with-tcltk-libs='%s'" % tcltk_libs)
 
         # don't add user site directory to sys.path (equivalent to python -s)
-        # This matters e.g. when pythong installs the bundled pip & setuptools (for >= 3.4)
+        # This matters e.g. when python installs the bundled pip & setuptools (for >= 3.4)
         env.setvar('PYTHONNOUSERSITE', '1', verbose=False)
 
         super(EB_Python, self).configure_step()
--- a/easybuild/easyblocks/p/python.py
+++ b/easybuild/easyblocks/p/python.py
@@ -161,13 +161,7 @@ class EB_Python(ConfigureMake):
         # See https://github.com/easybuilders/easybuild-easyconfigs/issues/11009
         for cfg in [os.path.join(os.path.expanduser('~'), name) for name in ('.pydistutils.cfg', 'pydistutils.cfg')]:
             if os.path.exists(cfg):
-                self.log.debug('Found distutils user config file at %s', cfg)
-                contents = read_file(cfg)
-                if any(line.startswith('prefix=') for line in contents.splitlines()):
-                    print_warning('Distutils user config found at %s containing a "prefix=" setting. '
-                                  'Installation might fail due to Python picking up that prefix. '
-                                  'Remove the file if this happens!',
-                                  log=self.log)
+                raise EasyBuildError("Legacy distutils user configuration file found at %s. Aborting.", cfg)
 
         self.cfg.update('configopts', "--enable-shared")
 
--- a/easybuild/easyblocks/p/python.py
+++ b/easybuild/easyblocks/p/python.py
@@ -46,7 +46,7 @@ from easybuild.framework.easyconfig import CUSTOM
 from easybuild.tools.build_log import EasyBuildError, print_warning
 from easybuild.tools.config import log_path
 from easybuild.tools.modules import get_software_libdir, get_software_root, get_software_version
-from easybuild.tools.filetools import change_dir, mkdir, symlink, write_file, read_file
+from easybuild.tools.filetools import change_dir, mkdir, symlink, write_file
 from easybuild.tools.run import run_cmd
 from easybuild.tools.systemtools import get_shared_lib_ext
 import easybuild.tools.toolchain as toolchain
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/p/python.py
```diff
@@ -185,6 +198,11 @@ def configure_step(self):
         if self.cfg['optimized'] and LooseVersion(self.version) >= LooseVersion('3.5.4'):
             self.cfg.update('configopts', "--enable-optimizations")
 
+        # Pip is included since 3.4 via ensurepip https://docs.python.org/3.4/whatsnew/changelog.html
+        if LooseVersion(self.version) >= LooseVersion('3.4.0'):
+            # Default, but do it explicitely
```
ocaisa, line 203: Typo `explicitly`

easybuild/easyblocks/p/python.py
```diff
@@ -232,6 +250,10 @@ def configure_step(self):
             }
             self.cfg.update('configopts', "--with-tcltk-libs='%s'" % tcltk_libs)
 
+        # don't add user site directory to sys.path (equivalent to python -s)
+        # This matters e.g. when pythong installs the bundled pip & setuptools (for >= 3.4)
```
ocaisa, line 254: Typo `python`

easybuild/easyblocks/p/python.py
Flamefire: Thanks

easybuild/easyblocks/p/python.py
Flamefire: Fixed

easybuild/easyblocks/p/python.py
```diff
@@ -156,6 +156,19 @@ def auto_detect_lto_support(self):
...
+        # See https://github.com/easybuilders/easybuild-easyconfigs/issues/11009
+        for cfg in [os.path.join(os.path.expanduser('~'), name) for name in ('.pydistutils.cfg', 'pydistutils.cfg')]:
+            if os.path.exists(cfg):
+                self.log.debug('Found distutils user config file at %s', cfg)
+                contents = read_file(cfg)
+                if any(line.startswith('prefix=') for line in contents.splitlines()):
+                    print_warning('Distutils user config found at %s containing a "prefix=" setting. '
+                                  'Installation might fail due to Python picking up that prefix. '
+                                  'Remove the file if this happens!',
+                                  log=self.log)
```
lexming, line 170: AFAIK `pydistutils` has been deprecated for a while in favour of virtual environments. Therefore, it might safer to just error out if its presence can cause issues.
```suggestion
                raise EasyBuildError("Legacy distutils user configuration file found at %s. Aborting.", cfg)
```

easybuild/easyblocks/p/python.py
```diff
@@ -185,6 +198,11 @@ def configure_step(self):
         if self.cfg['optimized'] and LooseVersion(self.version) >= LooseVersion('3.5.4'):
             self.cfg.update('configopts', "--enable-optimizations")
 
+        # Pip is included since 3.4 via ensurepip https://docs.python.org/3.4/whatsnew/changelog.html
+        if LooseVersion(self.version) >= LooseVersion('3.4.0'):
+            # Default, but do it explicitly
+            self.cfg.update('configopts', "--with-ensurepip=upgrade")
```
lexming, line 204: Instead of setting the default `upgrade`, I would just force the installation of `ensurepip` so that it is installed regardless of what is available in the system or the user distribution
```suggestion
            self.cfg.update('configopts', "--with-ensurepip=install")
```

easybuild/easyblocks/p/python.py
```diff
@@ -232,6 +250,10 @@ def configure_step(self):
             }
             self.cfg.update('configopts', "--with-tcltk-libs='%s'" % tcltk_libs)
 
+        # don't add user site directory to sys.path (equivalent to python -s)
+        # This matters e.g. when python installs the bundled pip & setuptools (for >= 3.4)
+        env.setvar('PYTHONNOUSERSITE', '1', verbose=False)
+
```
lexming, line 256: I have several dependencies of EasyBuild in my user site directory. Forcing the installation of `ensurepip` should be more reliable than completely disabling the user's site.

easybuild/easyblocks/p/python.py
Flamefire: I didn't want to be too restrictive. But sure that is easier

easybuild/easyblocks/p/python.py
Flamefire: Iirc this is the same as the upgrade when using pip: I.e. Force installation even if already installed. I think install is less "pressing"

easybuild/easyblocks/p/python.py
Flamefire: Do you have any reference on how the above is "forcing"? For me it wasn't iirc, but might have missed something. 

We do this for all pip packages and it didn't cause any problems yet, did it? So imo this is very reliable already and eb dependencies should not be affected as this env var only takes effect for new processes

easybuild/easyblocks/p/python.py
lexming: You are right, I misread the docs. `--with-ensurepip=install` just does a regular install, so if any version of the package is already installed it won't do anything. `upgrade` at least will match the version shipped with Python.

easybuild/easyblocks/p/python.py
lexming: Yep, `PYTHONNOUSERSITE` is needed. I thought that `--with-ensurepip=install` would force the installation but it's not the case.

easybuild/easyblocks/p/python.py
```diff
@@ -46,7 +46,7 @@
 from easybuild.tools.build_log import EasyBuildError, print_warning
 from easybuild.tools.config import log_path
 from easybuild.tools.modules import get_software_libdir, get_software_root, get_software_version
-from easybuild.tools.filetools import change_dir, mkdir, symlink, write_file
+from easybuild.tools.filetools import change_dir, mkdir, symlink, write_file, read_file
```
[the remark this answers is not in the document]
Flamefire:
```suggestion
from easybuild.tools.filetools import change_dir, mkdir, symlink, write_file
```

Flamefire:
@lexming @ocaisa Can this be merged? Would be great to have it in the upcoming release

ocaisa (approved):
Just waiting on the install to finish to confirm the additional sanity checks are used

ocaisa:
Going in, thanks @Flamefire

PR #2109, opened 2020-07-28
set $EB_INSTALLPYTHON in module generated for EasyBuild rather than setting $EB_PYTHON, to allow overriding Python command to be used for running EasyBuild with $EB_PYTHON

### LINKED ISSUE #2078: EB_PYTHON startup behaviour
The documentation at https://easybuild.readthedocs.io/en/latest/Python-2-3-compatibility.html?#controlling-which-python-command-easybuild-will-use-via-eb-python suggests that one can control which version of Python EasyBuild will use by setting EB_PYTHON.

Since https://github.com/easybuilders/easybuild-easyblocks/commit/40b76bef5a1f4f149e0c13cce913c051b54da5a3 running `module load EasyBuild` will now overwrite any existing EB_PYTHON environment variable.

While it is still possible to configure EB_PYTHON after running `module load EasyBuild`, it is no longer possible to configure EB_PYTHON first (e.g. in one's shell startup), since it is overwritten when EasyBuild loads.

If this is the desired behaviour then it should be documented. If this is not the desired behaviour, then it is a bug.

ocaisa:
Closes #2078
Requires ~~https://github.com/easybuilders/easybuild-framework/pull/3428~~ (which replaced ~~https://github.com/easybuilders/easybuild-framework/pull/3394~~)

### DIFF
--- a/easybuild/easyblocks/e/easybuildmeta.py
+++ b/easybuild/easyblocks/e/easybuildmeta.py
@@ -219,7 +219,8 @@ class EB_EasyBuildMeta(PythonPackage):
 
     def make_module_extra(self):
         """
-        Set $EB_INSTALLPYTHON to ensure that this EasyBuild installation uses the same Python executable it was installed with (unless overridden by $EB_PYTHON).
+        Set $EB_INSTALLPYTHON to ensure that this EasyBuild installation uses the same Python executable it was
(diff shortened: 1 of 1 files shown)


boegel (approved):
lgtm

akesandgren (approved):
LGTM

PR #2110, opened 2020-07-29
make TensorFlow easyblock also set $GCC_HOST_COMPILER_PREFIX to specify binutils location

Flamefire:
This avoids the need for the patch removing -B/usr/bin as TF will add GCC_HOST_COMPILER_PREFIX instead which is the correct path
Also some simplification of the used lists

This is required to be able to install TF 2.3.0: https://github.com/tensorflow/tensorflow/issues/41856 and allows to remove the patch from previous TF ECs (TODO: Do this after this is merged)

### DIFF
--- a/easybuild/easyblocks/t/tensorflow.py
+++ b/easybuild/easyblocks/t/tensorflow.py
@@ -155,6 +155,11 @@ class EB_TensorFlow(PythonPackage):
     def configure_step(self):
         """Custom configuration procedure for TensorFlow."""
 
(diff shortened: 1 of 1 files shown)


boegel:
I've tested this with various existing TensorFlow easyconfigs, both with the `TensorFlow-*_remove_usrbin_from_linker_bin_path_flag.patch` still included, and after removing it from the easyconfig, works like a charm, thanks @Flamefire!

PR #2113, opened 2020-08-11
Allow NAMD to be built on POWER, and also use the CUDA arch info

edmondac:
An easyconfigs PR for NAMD 2.14 will follow this PR

### DIFF
--- a/easybuild/easyblocks/n/namd.py
+++ b/easybuild/easyblocks/n/namd.py
@@ -27,6 +27,7 @@ from easybuild.tools.config import build_option
 from easybuild.tools.filetools import apply_regex_substitutions, change_dir, extract_file
 from easybuild.tools.modules import get_software_root, get_software_version
 from easybuild.tools.run import run_cmd
+from easybuild.tools.systemtools import POWER, X86_64, get_cpu_architecture
 
 
 class EB_NAMD(MakeCp):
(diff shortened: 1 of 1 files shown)


edmondac:
Test reports available in https://github.com/easybuilders/easybuild-easyconfigs/pull/11090

migueldiascosta:
Going in, thanks @edmondac!

easybuild/easyblocks/n/namd.py
```diff
@@ -43,10 +44,21 @@ def extra_options():
             # see http://charm.cs.illinois.edu/manuals/html/charm++/A.html
             'charm_arch': [None, "Charm++ target architecture", MANDATORY],
             'charm_opts': ['--with-production', "Charm++ build options", CUSTOM],
-            'namd_basearch': ['Linux-x86_64', "NAMD base target architecture (compiler family is appended", CUSTOM],
```
migueldiascosta, line 46: actually, if someone out there has been using `namd_basearch` in easyconfigs, this will break them, wouldn't it have been safer to keep the extra option with 'None' as default and then set it in the easyblock only when it's not set in the easyconfig?

boegel: Agreed, we shouldn't remove custom easyconfig parameters, that's not a backwards compatible change (especially since in EasyBuild 4.x we complain about unknown easyconfig parameters).

I'll open a PR to fix this.

easybuild/easyblocks/n/namd.py
```diff
@@ -43,10 +44,21 @@ def extra_options():
...
             'namd_cfg_opts': ['', "NAMD configure options", CUSTOM],
             'runtest': [True, "Run NAMD test case after building", CUSTOM],
         })
+        arch = get_cpu_architecture()
+        if arch == X86_64:
+            basearch = 'Linux-x86_64'
+        elif arch == POWER:
+            basearch = 'Linux-POWER'
+
+        cuda = get_software_root('CUDA')
```
boegel, line 56: This won't work as expected, since dependencies are not loaded yet at this point, this should be done in `prepare_step`...

easybuild/easyblocks/n/namd.py
boegel: see #2123

PR #2115, opened 2020-08-11
Add EasyBlock for Pybind11

Flamefire:
This makes sure that the installed Pybind11 can be found by pip in addition to CMake which avoids a redownload by e.g. SciPy

### DIFF
--- a/easybuild/easyblocks/p/pybind11.py
+++ b/easybuild/easyblocks/p/pybind11.py
@@ -1,0 +1,94 @@
+##
+# Copyright 2009-2020 Ghent University
+#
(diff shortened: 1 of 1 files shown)


smoors:
why not just install it with pip and add a postinstallcmds with a symbolic link from `site-packages/pybind11/include` to `%(installdir)s/include`?

Flamefire:
Because then it would be missing the CMake config files we currently have. Also the location of the headers vary. For <2.5 they are at `installdir/include/python3.7m` and for 2.5 they are in site-packages

migueldiascosta:
Going in, thanks @Flamefire!

PR #2117, opened 2020-08-13
Make TensorFlow use system/EasyBuild installed libraries

### LINKED ISSUE #1996: TensorFlow ignores system dependencies
TensorFlow has some bundled stuff that is built alongside the main build. We should use system libs where possible. There is a list of stuff that we can replace: https://github.com/tensorflow/tensorflow/blob/f64e839687b6129dfe2a9e21ef965593c7f3a3c2/third_party/systemlibs/syslibs_configure.bzl#L11-L46

I did a grep for compilations done during TF 2.1.0, grouped and sorted them (should be the number of files compiled basically as a metric how expensive a dependency is):

      2 @org_sqlite
      3 @farmhash_archive
      4 @lmdb
      6 @com_github_nanopb_nanopb
      6 @fft2d
      6 @highwayhash
      6 @jsoncpp_git
      6 @kissfft
      9 @snappy
     21 @pcre
     24 @double_conversion
     24 @gif
     30 @png
     44 @flatbuffers
     45 @zlib_archive
     48 @swig
     51 @nsync
     54 @hwloc
     63 @com_googlesource_code_re2
     78 @nasm
    150 @libjpeg_turbo
    178 @curl
    223 @com_google_absl
    252 @local_config_mlir
    290 @mkl_dnn
    347 @com_google_protobuf
    382 @icu
    552 @boringssl
    676 @aws
    831 @grpc
    840 @llvm

So judging from that protobuf is the obvious #1 candidate: We already include it as a dependency. 

llvm would be great to move out but not sure if this is worth.

aws is strange. We set TF_NEED_S3 and TF_NEED_AWS to 0 but it is still built. From the source I judge it is not supported (anymore). But there is `--config=noaws` that we can pass to configure

Curl is another candidate as we already have ECs for that. Same for double-conversion.

Generally: We should comb through the list and use explicit dependencies where possible.

Flamefire:
TensorFlow allows using already installed libs. This avoids problems with the bundled TF libs (e.g. in 2.3.0) and speeds up the build process.

closes #1996

edit: now requires ~~https://github.com/easybuilders/easybuild-framework/pull/3426~~

### DIFF
--- a/easybuild/easyblocks/t/tensorflow.py
+++ b/easybuild/easyblocks/t/tensorflow.py
@@ -102,6 +102,7 @@ class EB_TensorFlow(PythonPackage):
             'use_pip': True,
         }
         self.cfg['exts_filter'] = EXTS_FILTER_PYTHON_PACKAGES
+        self.system_libs_info = None
 
         self.test_script = None
 
@@ -152,85 +153,133 @@ class EB_TensorFlow(PythonPackage):
             adjust_permissions(wrapper, stat.S_IXUSR)
             self.log.info("Using wrapper script for '%s': %s", compiler, which(compiler))
 
-    def get_system_libs(self):
+    @staticmethod
+    def get_system_libs_for_version(tf_version, as_VALID_LIBS=False):
         """
-        Get list of dependencies for TF_SYSTEM_LIBS
+        Return a list of valid values for $TF_SYSTEM_LIBS for the given TF version
 
-        Returns a tuple of lists: TF_SYSTEM_LIBS names, include paths, library paths
+        The result is a list of tuples (tf_name, eb_name) where eb_name can be None if no EC exists or True to assume
+        the dependency is always installed (e.g. via pip in the TF EC)
+
+        Note: Can be used to check against third_party/systemlibs/syslibs_configure.bzl by running:
+              from easybuild.easyblocks.tensorflow import EB_TensorFlow
+              print(EB_TensorFlow.get_system_libs_for_version('2.1.0', as_VALID_LIBS=True))
         """
-        def is_version_ok(dep_version):
-            version = LooseVersion(self.version)
-            min_version = LooseVersion(dep_version[0])
+        tf_version = LooseVersion(tf_version)
+
+        def is_version_ok(version_requirement):
+            """Return True if the TF version to be installed matches the version_requirement"""
+            min_version = LooseVersion(version_requirement[0])
             result = False
-            if version >= min_version:
+            if tf_version >= min_version:
                 # min. version matches, so OK if max version matches or not present
-                if len(dep_version) == 1 or version < LooseVersion(dep_version[1]):
+                if len(version_requirement) == 1 or tf_version < LooseVersion(version_requirement[1]):
                     result = True
             return result
-
-        def make_tf_name(eb_dep_name, tf_dep_name):
-            return tf_dep_name or eb_dep_name.lower().replace('-', '_')
 
         # For this list check third_party/systemlibs/syslibs_configure.bzl --> VALID_LIBS
         # Also verify third_party/systemlibs/<name>.BUILD if it does something "strange" (e.g. link hardcoded headers)
         available_system_libs = (
-            # Format: 1. dependency name in EB
-            #         2. name in TF or None to lowercase EB name with - replaced by _
+            # Format: 1. name in TF
+            #         2. dependency name in EB, can use None if unknown or True to assume installed
             #         3. min. (incl), max. (excl) TF version required
-            ('cURL', None, ('2.1.0',)),
-            ('double-conversion', None, ('2.1.0',)),
-            ('giflib', 'gif', ('2.1.0',)),
-            ('hwloc', None, ('2.1.0',)),
-            ('ICU', None, ('2.1.0',)),
-            ('JsonCpp', 'jsoncpp_git', ('2.1.0',)),
-            ('libjpeg-turbo', 'jpeg', ('2.1.0', '2.2.0')),
-            ('libjpeg-turbo', 'libjpeg_turbo', ('2.2.0', )),
-            ('LMDB', None, ('2.1.0',)),
-            ('NASM', None, ('2.1.0',)),
-            ('SQLite', 'org_sqlite', ('2.1.0',)),
-            ('PCRE', None, ('2.1.0',)),
-            ('libpng', 'png', ('2.1.0',)),
-            ('pybind11', None, ('2.2.0',)),
-            ('snappy', None, ('2.1.0',)),
-            ('SWIG', None, ('2.1.0',)),
-            ('zlib', 'zlib_archive', ('2.1.0', '2.2.0')),
+            ('absl_py', True, ('2.1.0',)),
+            ('astor_archive', True, ('2.1.0',)),
+            ('astunparse_archive', True, ('2.2.0',)),
+            ('boringssl', None, ('2.1.0',)),
+            ('com_github_googleapis_googleapis', None, ('2.1.0',)),
+            ('com_github_googlecloudplatform_google_cloud_cpp', None, ('2.1.0',)),  # Not used due to $TF_NEED_GCP=0
+            ('com_github_grpc_grpc', None, ('2.2.0',)),
+            ('com_google_protobuf', 'protobuf-python', ('2.1.0',)),
+            ('com_googlesource_code_re2', None, ('2.1.0',)),
+            ('curl', 'cURL', ('2.1.0',)),
+            ('cython', 'Python', ('2.1.0',)),
+            ('double_conversion', 'double-conversion', ('2.1.0',)),
+            ('enum34_archive', True, ('2.1.0',)),  # Part of Python3
+            ('flatbuffers', 'flatbuffers', ('2.1.0',)),
+            ('functools32_archive', True, ('2.1.0',)),  # Part of Python3
+            ('gast_archive', True, ('2.1.0',)),
+            ('gif', 'giflib', ('2.1.0',)),
+            ('grpc', None, ('2.1.0', '2.2.0')),
+            ('hwloc', 'hwloc', ('2.1.0',)),
+            ('icu', 'ICU', ('2.1.0',)),
+            ('jpeg', 'libjpeg-turbo', ('2.1.0', '2.2.0')),
+            ('jsoncpp_git', 'JsonCpp', ('2.1.0',)),
+            ('keras_applications_archive', True, ('2.1.0', '2.2.0')),
+            ('libjpeg_turbo', 'libjpeg-turbo', ('2.2.0', )),
+            ('lmdb', 'LMDB', ('2.1.0',)),
+            ('nasm', 'NASM', ('2.1.0',)),
+            ('nsync', 'nsync', ('2.1.0',)),
+            ('opt_einsum_archive', True, ('2.1.0',)),
+            ('org_sqlite', 'SQLite', ('2.1.0',)),
+            ('pasta', True, ('2.1.0',)),
+            ('pcre', 'PCRE', ('2.1.0',)),
+            ('png', 'libpng', ('2.1.0',)),
+            ('pybind11', 'pybind11', ('2.2.0',)),
+            ('six_archive', 'Python', ('2.1.0',)),
+            ('snappy', 'snappy', ('2.1.0',)),
+            ('swig', 'SWIG', ('2.1.0',)),
+            ('termcolor_archive', True, ('2.1.0',)),
+            ('wrapt', True, ('2.1.0',)),
+            ('zlib_archive', 'zlib', ('2.1.0', '2.2.0')),
             ('zlib', 'zlib', ('2.2.0',)),
         )
-        # Same as above but installed as python extensions
-        python_system_libs = (
-            ('absl-py', None, ('2.1.0',)),
-            ('astor', 'astor_archive', ('2.1.0',)),
-            ('astunparse', 'astunparse_archive', ('2.2.0',)),
-            ('gast', 'gast_archive', ('2.1.0',)),
-            ('Keras-Applications', 'keras_applications_archive', ('2.1.0', '2.2.0')),
-            ('opt-einsum', 'opt_einsum_archive', ('2.1.0',)),
-            ('google-pasta', 'pasta', ('2.1.0',)),
-            ('termcolor', 'termcolor_archive', ('2.1.0',)),
-        )
+        result = [(tf_dep_name, dep_name) for tf_dep_name, dep_name, version_requirement in available_system_libs
+                  if is_version_ok(version_requirement)]
+        if as_VALID_LIBS:
+            result = '\n'.join(['    "%s",' % tf_dep_name for tf_dep_name, dep_name in result])
+        return result
 
-        dependencies = set(dep['name'] for dep in self.cfg.dependencies())
+    def get_system_libs(self):
+        """
+        Get list of dependencies for $TF_SYSTEM_LIBS
+
+        Returns a tuple of lists: $TF_SYSTEM_LIBS names, include paths, library paths
+        """
+        available_system_libs = EB_TensorFlow.get_system_libs_for_version(self.version)
+
+        dep_names = set(dep['name'] for dep in self.cfg.dependencies())
 
         system_libs = []
         cpaths = []
         libpaths = []
-        for dep_name, tf_dep_name, version_requirement in available_system_libs:
-            if dep_name in dependencies and is_version_ok(version_requirement):
-                system_libs.append(make_tf_name(dep_name, tf_dep_name))
+        ignored_system_deps = []
+        for tf_dep_name, dep_name in available_system_libs:
+            if dep_name is None:
+                continue
+            if dep_name is True:
+                system_libs.append(tf_dep_name)
+            elif dep_name in dep_names:
+                system_libs.append(tf_dep_name)
+                # For protobuf we need protobuf and protobuf-python where the latter depends on the former
+                # For includes etc. we need to get the values from protobuf
+                if dep_name == 'protobuf-python':
+                    dep_name = 'protobuf'
                 sw_root = get_software_root(dep_name)
+                # Dependency might be filtered via --filter-deps. In that case assume globally installed version
+                if not sw_root:
+                    continue
                 incpath = os.path.join(sw_root, 'include')
                 if dep_name == 'JsonCpp':
                     # Need to use the install prefix instead: https://github.com/tensorflow/tensorflow/issues/42303
                     incpath = sw_root
                 if os.path.exists(incpath):
                     cpaths.append(incpath)
+                    if dep_name == 'protobuf':
+                        # Need to set INCLUDEDIR as TF wants to symlink headers from there:
+                        # https://github.com/tensorflow/tensorflow/issues/37835
+                        env.setvar('INCLUDEDIR', incpath)
                 libpath = get_software_libdir(dep_name)
                 if libpath:
                     libpaths.append(os.path.join(sw_root, libpath))
+            else:
+                ignored_system_deps.append('%s (%s)' % (tf_dep_name, dep_name))
 
-        for dep_name, tf_dep_name, version_requirement in python_system_libs:
-            # We just assume they are present as otherwise pip check would fail
-            if is_version_ok(version_requirement):
-                system_libs.append(make_tf_name(dep_name, tf_dep_name))
+        if ignored_system_deps:
+            self.log.info('For the following dependencies TensorFlow will download a copy as EB dependency was not' +
+                          'found: \n%s',
+                          ', '.join(ignored_system_deps))
+
         return system_libs, cpaths, libpaths
 
     def configure_step(self):
@@ -286,8 +335,8 @@ class EB_TensorFlow(PythonPackage):
             env.setvar('PATH', os.pathsep.join([wrapper_dir, os.getenv('PATH')]))
 
         self.prepare_python()
-
         self.handle_jemalloc()
+        self.system_libs_info = self.get_system_libs()
 
         cuda_root = get_software_root('CUDA')
         cudnn_root = get_software_root('cuDNN')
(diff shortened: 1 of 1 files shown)


boegel:
Tested with:

* [x] `TensorFlow-1.13.1-foss-2019a-Python-3.7.2.eb`
* [ ] `TensorFlow-1.15.2-fosscuda-2019b-Python-3.7.4.eb`
* [x] `TensorFlow-2.1.0-foss-2019b-Python-3.7.4.eb`
* [ ] `TensorFlow-2.2.0-foss-2019b-Python-3.7.4.eb`
* [ ] `TensorFlow-2.2.0-fosscuda-2019b-Python-3.7.4.eb`

Flamefire:
Thanks for the testing. This change should currently have no effect as the dependencies are missing. I added them to the 2.1.0 ECs in https://github.com/easybuilders/easybuild-easyconfigs/pull/11109 so you can add those to your tests too if you like.

Flamefire:
Also another idea: As TF likes to add/remove/rename packages, maybe we should add all names to the EasyBlock and verify the list. Even when we don't have a corresponding EC yet.

easybuild/easyblocks/t/tensorflow.py
```diff
@@ -152,6 +152,87 @@ def write_wrapper(self, wrapper_dir, compiler, i_mpi_root):
...
+            ('hwloc', None, ('2.1.0',)),
+            ('ICU', None, ('2.1.0',)),
+            ('JsonCpp', 'jsoncpp_git', ('2.1.0',)),
+            ('libjpeg-turbo', 'jpeg', ('2.1.0', '2.2.0')),
+            ('libjpeg-turbo', 'libjpeg_turbo', ('2.2.0', )),
+            ('LMDB', None, ('2.1.0',)),
+            ('NASM', None, ('2.1.0',)),
+            ('SQLite', 'org_sqlite', ('2.1.0',)),
+            ('PCRE', None, ('2.1.0',)),
+            ('libpng', 'png', ('2.1.0',)),
```
boegel, line 192: for our own sanity, we should keep this sorted alphabetically?

easybuild/easyblocks/t/tensorflow.py
```diff
@@ -152,6 +152,87 @@ def write_wrapper(self, wrapper_dir, compiler, i_mpi_root):
...
+        )
+        # Same as above but installed as python extensions
+        python_system_libs = (
+            ('absl-py', None, ('2.1.0',)),
+            ('astor', 'astor_archive', ('2.1.0',)),
+            ('astunparse', 'astunparse_archive', ('2.2.0',)),
+            ('gast', 'gast_archive', ('2.1.0',)),
+            ('Keras-Applications', 'keras_applications_archive', ('2.1.0', '2.2.0')),
+            ('opt-einsum', 'opt_einsum_archive', ('2.1.0',)),
+            ('google-pasta', 'pasta', ('2.1.0',)),
```
boegel, line 207: same here, please sort alphabetically

easybuild/easyblocks/t/tensorflow.py
```diff
@@ -152,6 +152,87 @@ def write_wrapper(self, wrapper_dir, compiler, i_mpi_root):
...
+            ('termcolor', 'termcolor_archive', ('2.1.0',)),
+        )
+
+        dependencies = set(dep['name'] for dep in self.cfg.dependencies())
```
boegel, line 211: `dependencies` -> `dep_names`

easybuild/easyblocks/t/tensorflow.py
```diff
@@ -152,6 +152,87 @@ def write_wrapper(self, wrapper_dir, compiler, i_mpi_root):
...
+
+        system_libs = []
+        cpaths = []
+        libpaths = []
+        for dep_name, tf_dep_name, version_requirement in available_system_libs:
```
boegel, line 216: rename `version_requirement` to `version_range`?

easybuild/easyblocks/t/tensorflow.py
```diff
@@ -152,6 +152,87 @@ def write_wrapper(self, wrapper_dir, compiler, i_mpi_root):
...
+            if dep_name in dependencies and is_version_ok(version_requirement):
+                system_libs.append(make_tf_name(dep_name, tf_dep_name))
+                sw_root = get_software_root(dep_name)
```
boegel, line 219: We should probably verify here if this is different from `None`, since due some dependencies may be being filtered (via `--filter-deps`). See also https://github.com/easybuilders/easybuild-framework/issues/1494

easybuild/easyblocks/t/tensorflow.py
```diff
@@ -152,6 +152,87 @@ def write_wrapper(self, wrapper_dir, compiler, i_mpi_root):
...
+                if dep_name == 'JsonCpp':
+                    # Need to use the install prefix instead: https://github.com/tensorflow/tensorflow/issues/42303
+                    incpath = sw_root
+                if os.path.exists(incpath):
+                    cpaths.append(incpath)
+                libpath = get_software_libdir(dep_name)
+                if libpath:
+                    libpaths.append(os.path.join(sw_root, libpath))
+
+        for dep_name, tf_dep_name, version_requirement in python_system_libs:
```
boegel, line 230: also rename `version_requirement` to `version_range` (if you do so above)

easybuild/easyblocks/t/tensorflow.py
Flamefire: I decided to keep it like that as it is not necessarily a range but can be a single element tuple (hopefully almost always)

Flamefire:
I addressed your comments and also some changes:

- introduce function get_system_libs_for_version which does the version filtering and can be used to render a list that can be diffed against the TF included list on updates. IMO crucial to keep up to date. We might even want to do this automatically during the build, what do you think?
- Add ALL TF known libs to a) allow for the above and b) see which are missing to potentially add new ECs or find the correct ones
- Combined the 2 lists into 1 and use values of `None` and `True` to denote missing ECs and "assume installed"
- Change order of tf_name and eb_name and sort lines. There are benefits in doing it this way (e.g. check against the tf list) but maybe the other way round is better as it allows to see name changes more easily as EC names stay (usually) constant (see jpeg). I tend to revert the order again but I'm still not 100% happy with the list anyway. Not great to read. Any ideas?

Flamefire:
Ok, still need to do a few adjustments. I'm adding and testing more ECs to be used as dependencies. To check I added a log which dependencies that could be used by TF are not.

The main dependency I'd like to include is protobuf but that needs another patch and a test. Checking now if that works...

easybuild/easyblocks/t/tensorflow.py
```diff
@@ -152,6 +153,135 @@ def write_wrapper(self, wrapper_dir, compiler, i_mpi_root):
...
+            ('functools32_archive', True, ('2.1.0',)),  # Part of Python3
+            ('gast_archive', True, ('2.1.0',)),
+            ('gif', 'giflib', ('2.1.0',)),
+            ('grpc', None, ('2.1.0', '2.2.0')),
+            ('hwloc', 'hwloc', ('2.1.0',)),
+            ('icu', 'ICU', ('2.1.0',)),
+            ('jpeg', 'libjpeg-turbo', ('2.1.0', '2.2.0')),
+            ('jsoncpp_git', 'JsonCpp', ('2.1.0',)),
+            ('keras_applications_archive', True, ('2.1.0', '2.2.0')),
+            ('libjpeg_turbo', 'libjpeg-turbo', ('2.2.0', )),
```
boegel, line 209: @Flamefire Ordering by EasyBuild software name, like it was before, makes more sense to me imho, since we're mapping from EB name (key) to TF name (value).

Maybe it should even be a proper dictionary (with dict values), where you filter the values based on the version range?

```python
available_system_libs = {
    'libjpeg-turbo': {
        ('2.1.0', '2.2.0'): 'jpeg',
        ('2.2.0',): 'libjpeg-turbo',
    },
    ...
    'SWIG': {('2.1.0',): 'swig'},
}
```

or even:

```python
available_system_libs = {
    'libjpeg-turbo': {
        '2.1.0:2.2.0': 'jpeg',
        '2.2.0:': 'libjpeg-turbo',
    },
    ...
    'SWIG': {'2.1.0:': 'swig'},
}
```

easybuild/easyblocks/t/tensorflow.py
```diff
@@ -152,6 +153,135 @@ def write_wrapper(self, wrapper_dir, compiler, i_mpi_root):
             adjust_permissions(wrapper, stat.S_IXUSR)
             self.log.info("Using wrapper script for '%s': %s", compiler, which(compiler))
 
+    @staticmethod
+    def get_system_libs_for_version(tf_version, as_VALID_LIBS=False):
```
boegel, line 157: `as_VALID_LIBS` -> `as_valid_libs`

easybuild/easyblocks/t/tensorflow.py
Flamefire: The dict won't work due to the `True` and `None` values :/ And I'd like to keep them

easybuild/easyblocks/t/tensorflow.py
Flamefire: That is intentional: In TF the name of the variable is `VALID_LIBS` and the format of the output is the same as that constant so you can diff them

easybuild/easyblocks/t/tensorflow.py
```diff
@@ -152,6 +278,103 @@ def write_wrapper(self, wrapper_dir, compiler, i_mpi_root):
...
+        for dep_name, tf_name in dependency_mapping.items():
+            if dep_name in dep_names:
+                system_libs.append(tf_name)
+                # For protobuf we need protobuf and protobuf-python where the latter depends on the former
+                # For includes etc. we need to get the values from protobuf
+                if dep_name == 'protobuf-python':
+                    dep_name = 'protobuf'
+                sw_root = get_software_root(dep_name)
+                # Dependency might be filtered via --filter-deps. In that case assume globally installed version
+                if not sw_root:
```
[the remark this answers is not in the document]
Flamefire: This code is only executed if `dep_name is in self.ec.dependencies()` and you told me to add the check for a defined `sw_root` to cater for this bug. So IMO all is handled, isn't it?

easybuild/easyblocks/t/tensorflow.py
```diff
@@ -152,6 +278,103 @@ def write_wrapper(self, wrapper_dir, compiler, i_mpi_root):
...
+                incpath = os.path.join(sw_root, 'include')
+                if dep_name == 'JsonCpp':
+                    # Need to use the install prefix instead: https://github.com/tensorflow/tensorflow/issues/42303
+                    incpath = sw_root
+                if os.path.exists(incpath):
+                    cpaths.append(incpath)
+                    if dep_name == 'protobuf':
+                        # Need to set INCLUDEDIR as TF wants to symlink headers from there:
+                        # https://github.com/tensorflow/tensorflow/issues/37835
+                        env.setvar('INCLUDEDIR', incpath)
```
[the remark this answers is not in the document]
Flamefire: > is that what we want to do?

Unfortunately yes. With a patch for TF I have `INCLUDEDIR` is only used by the Bazel code for protobuf (without it IIRC JsonCpp does use it too) so it is fine.

> Wouldn't it be more future-proof to take into account the existing $INCLUDEDIR value, 

The point is: And do what? It is an environment variable which TF expects to be set to 1 (and only 1) folder so there is nothing we could do with an existing value. The way they wrote that stuff is that ALL software is installed into 1 prefix, e.g. `/usr` -.-

easybuild/easyblocks/t/tensorflow.py
```diff
@@ -152,6 +278,103 @@ def write_wrapper(self, wrapper_dir, compiler, i_mpi_root):
...
+        dependency_mapping, python_mapping = get_system_libs_for_version(self.version)
+
+        system_libs = []
+        cpaths = []
+        libpaths = []
+        ignored_system_deps = []
+
+        # Check direct dependencies
+        dep_names = set(dep['name'] for dep in self.cfg.dependencies())
+        for dep_name, tf_name in dependency_mapping.items():
```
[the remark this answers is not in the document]
Flamefire: IIRC we are doing similar things at other places. I don't think there is any harm in the "unspecified" order, or do you see any issues?

easybuild/easyblocks/t/tensorflow.py
```diff
@@ -237,6 +462,7 @@ def configure_step(self):
             'TF_NEED_KAFKA': '0',  # Amazon Kafka Platform
             'TF_SET_ANDROID_WORKSPACE': '0',
             'TF_DOWNLOAD_CLANG': '0',  # Still experimental in TF 2.1.0
+            'TF_SYSTEM_LIBS': ','.join(self.system_libs_info[0]),
```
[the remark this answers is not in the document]
Flamefire: I see no harm in being explicit here in NOT using any $TF_SYSTEM_LIBS. It will appear in the log as setting that to empty which is good IMO

easybuild/easyblocks/t/tensorflow.py
```diff
@@ -152,6 +278,103 @@ def write_wrapper(self, wrapper_dir, compiler, i_mpi_root):
...
+                ignored_system_deps.append('%s (%s)' % (tf_name, dep_name))
+
+        # Check installed python packages but only check stdout, not stderr which might contain user facing warnings
+        cmd_list = [self.python_cmd, '-m', 'pip', 'list', '--isolated', '--disable-pip-version-check',
+                    '--format', 'json']
+        full_cmd = ' '.join(cmd_list)
+        self.log.info("Running command '%s'" % full_cmd)
+        proc = subprocess_popen_text(cmd_list, env=os.environ)
+        (stdout, stderr) = proc.communicate()
+        ec = proc.returncode
```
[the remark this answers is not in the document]
Flamefire: As discussed in slack this is postponed until implemented in framework

easybuild/easyblocks/t/tensorflow.py
```diff
@@ -152,6 +278,103 @@ def write_wrapper(self, wrapper_dir, compiler, i_mpi_root):
...
+            else:
+                msg = 'Values for $TF_SYSTEM_LIBS in the TensorFlow EasyBlock are incomplete.\n'
+                if missing_libs:
+                    msg += 'Missing $TF_SYSTEM_LIBS=%s\n' % missing_libs
+                if missing_libs:
+                    msg += 'Unrecognized $TF_SYSTEM_LIBS=%s\n' % unknown_libs
+            if build_option('strict') == run.ERROR:
+                raise EasyBuildError(msg)
+            else:
+                print_warning(msg)
```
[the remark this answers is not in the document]
Flamefire: I added `The EasyBlock needs to be updated to fully work with TF version %s` to the error and some comments above

easybuild/easyblocks/t/tensorflow.py
```diff
@@ -152,6 +278,103 @@ def write_wrapper(self, wrapper_dir, compiler, i_mpi_root):
...
+        cmd_list = [self.python_cmd, '-m', 'pip', 'list', '--isolated', '--disable-pip-version-check',
+                    '--format', 'json']
+        full_cmd = ' '.join(cmd_list)
+        self.log.info("Running command '%s'" % full_cmd)
+        proc = subprocess_popen_text(cmd_list, env=os.environ)
+        (stdout, stderr) = proc.communicate()
+        ec = proc.returncode
+        self.log.info("Command '%s' returned with %s: stdout: %s; stderr: %s" % (full_cmd, ec, stdout, stderr))
+        if ec:
+            raise EasyBuildError('Failed to determine installed python packages')
```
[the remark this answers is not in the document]
Flamefire: Added to the end

Flamefire:
@boegel needed to rebase to include the ninja fix to test this. New commits are starting at "Address review"

easybuild/easyblocks/t/tensorflow.py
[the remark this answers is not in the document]
boegel: Agreed.

easybuild/easyblocks/t/tensorflow.py
```diff
@@ -152,6 +280,106 @@ def write_wrapper(self, wrapper_dir, compiler, i_mpi_root):
...
+    def verify_system_libs_info(self):
+        """Verifies that the stored info about $TF_SYSTEM_LIBS is complete"""
+        available_libs_src = set(get_system_libs_from_tf(self.start_dir or self.ext_dir))
+        available_libs_eb = set(split_tf_libs_txt(get_system_libs_for_version(self.version, as_valid_libs=True)))
+        # If available_libs_eb is empty it is not an error e.g. it is not worth trying to make all old ECs work
+        # So we just log it so it can be verified manually if required
+        if not available_libs_eb:
+            self.log.warning('TensorFlow EasyBlock does not have any information for $TF_SYSTEM_LIBS stored. ' +
+                             'This means most dependencies will be downloaded at build time by TensorFlow.\n' +
+                             'Available $TF_SYSTEM_LIBS according to the TF sources: %s', sorted(available_libs_src))
```
boegel, line 292: @Flamefire `TF sources` -> `TensorFlow sources` (it's fine to use `TF` in comments, but I wouldn't do that in warnings/errors)

easybuild/easyblocks/t/tensorflow.py
```diff
@@ -152,6 +280,106 @@ def write_wrapper(self, wrapper_dir, compiler, i_mpi_root):
...
+                msg = 'Failed to determine available $TF_SYSTEM_LIBS from the source'
+            else:
+                msg = 'Values for $TF_SYSTEM_LIBS in the TensorFlow EasyBlock are incomplete.\n'
+                if missing_libs:
+                    # Libs available according to TF sources but not listed in this EasyBlock
+                    msg += 'Missing entries for $TF_SYSTEM_LIBS: %s\n' % missing_libs
+                if unknown_libs:
+                    # Libs listed in this EasyBlock but not present in the TF sources -> Removed?
+                    msg += 'Unrecognized entries for $TF_SYSTEM_LIBS: %s\n' % unknown_libs
+                msg += 'The EasyBlock needs to be updated to fully work with TF version %s' % self.version
```
boegel, line 308: `TF version` -> `TensorFlow version`

easybuild/easyblocks/t/tensorflow.py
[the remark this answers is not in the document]
boegel: Yes, I overlooked that this was already using `self.ec.dependencies()` higher up

easybuild/easyblocks/t/tensorflow.py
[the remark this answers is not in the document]
boegel: Right ok... It's not a `$PATH`-like variable, so little else we can do to overwrite it.

easybuild/easyblocks/t/tensorflow.py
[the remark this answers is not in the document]
boegel: No real issues, it's just easier for humans staring at sorted things rather than having a random order (which may change between runs...).

easybuild/easyblocks/t/tensorflow.py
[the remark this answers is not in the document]
boegel: You're assuming here that every TensorFlow version won't trip over having `$TF_SYSTEM_LIBS` set to an empty value, which is technically not the same as not setting it (although in practice there's little difference).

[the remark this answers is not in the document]
Flamefire: I'd argue that this assumption is safe. In fact current TF code contains `syslibs = environ_cp.get('TF_SYSTEM_LIBS', '')` so for that not setting it and setting it to empty is exactly the same. I verified that in 1.10.0 where the variable first appears the behavior is the same: empty and unset are treated equal

Therefore I'd like to go with the current solution which is cleaner in code (1 line less, no repetition of the dict var name) and more explicit (shows in the log as being set to empty)

easybuild/easyblocks/t/tensorflow.py
```diff
@@ -73,6 +76,130 @@
...
+
+def get_system_libs_from_tf(source_dir):
+    """Return the valid values for TF_SYSTEM_LIBS from the TensorFlow source directory"""
+    syslibs_path = os.path.join(source_dir, 'third_party', 'systemlibs', 'syslibs_configure.bzl')
+    result = []
+    if os.path.exists(syslibs_path):
+        txt = read_file(syslibs_path)
+        m = re.search(r'VALID_LIBS = \[(.*?)\]', txt, re.DOTALL)
+        if not m:
+            raise EasyBuildError('VALID_LIBS definition not found in %s', syslibs_path)
```
boegel, line 98: @Flamefire This is too strict, it breaks the installation of `TensorFlow-1.10.0-fosscuda-2018b-Python-2.7.15.eb` for example.

It seems OK for 1.13.1 and newer, so I would only error out for sufficiently recent versions?

Also, we shouldn't use a single-letter variable outside of a list comprehension, so please fix that too while you're at it (`valid_libs = re.search(...)`)

boegel (changes requested):

easybuild/easyblocks/t/tensorflow.py
Flamefire: Found that 1.10.0 (where this was introduced) is missing spaces. Adapted the regex.

It is actually only done for "sufficiently recent versions" where I defined this as "has the `syslibs_configure.bzl` file", which should be ok enough.

Fixed in https://github.com/easybuilders/easybuild-easyblocks/pull/2117/commits/969fcf245727d88b7dd7cb1cfd7154e380e98f22

easybuild/easyblocks/t/tensorflow.py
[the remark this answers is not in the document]
boegel: OK!

boegel (approved):

boegel:
Doing a last round of testing, but this should be good to go now... Thanks a lot for your efforts on this @Flamefire!

boegel:
This has been tested extensively, both with a wide range of existing TensorFlow easyconfigs and the updated ones for TensorFlow 2.1.0 from https://github.com/easybuilders/easybuild-easyconfigs/pull/11109 that require these changes, so going in!

PR #2118, opened 2020-08-14
add support in NAMD easyblock to pass additional C++ compiler options for building Charm++ component with + wrap them in single quotes

boegel:
Building NAMD 2.14 with `foss/2019b` or `foss/2020a` requires passing `-fpermissive` when building Charm++, so we need a custom easyconfig parameter to allow for that...

### DIFF
--- a/easybuild/easyblocks/n/namd.py
+++ b/easybuild/easyblocks/n/namd.py
@@ -43,6 +43,7 @@ class EB_NAMD(MakeCp):
         extra.update({
             # see http://charm.cs.illinois.edu/manuals/html/charm++/A.html
             'charm_arch': [None, "Charm++ target architecture", MANDATORY],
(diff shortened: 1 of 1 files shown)


migueldiascosta:
just wondering, if the issue is specific to the g++ version, why not checking that and adding `-fpermissive` automatically? 

are you anticipating the nedd to pass other `cxxflags` to  `./build charm++`?

boegel:
@migueldiascosta I could, but i) I don't know which range of g++ versions the flag is required, ii) this problem will probably disappear in future NAMD versions (so the check in the easyblock would have to be both compiler and NAMD version specific, which seems a bit much).

There may arise a use case for passing other C++ compiler flags to the `./build charm++` command, so I figured this was the best way forward...

migueldiascosta:
Going in, thanks @boegel!

PR #2125, opened 2020-08-18
make sure self.start_dir is set to a full path before constructing installation command in RPackage easyblock

smoors:
currently `start_dir` is not prefixed for RPackage extensions, requiring the full path to be specified.
this is fixed by prepending `start_dir` with `ext_dir`.

### DIFF
--- a/easybuild/easyblocks/generic/rpackage.py
+++ b/easybuild/easyblocks/generic/rpackage.py
@@ -132,7 +132,10 @@ class RPackage(ExtensionEasyBlock):
         else:
             prefix = ''
 
-        loc = self.start_dir or self.ext_dir or self.ext_src
+        if self.start_dir:
+            loc = os.path.join(self.ext_dir or os.path.sep, self.start_dir)
+        else:
+            loc = self.ext_dir or self.ext_src
 
         cmd = ' '.join([
             self.cfg['preinstallopts'],


easybuild/easyblocks/generic/rpackage.py
```diff
@@ -132,7 +132,10 @@ def make_cmdline_cmd(self, prefix=None):
         else:
             prefix = ''
 
-        loc = self.start_dir or self.ext_dir or self.ext_src
+        if self.start_dir:
+            loc = os.path.join(self.ext_dir or os.path.sep, self.start_dir)
+        else:
+            loc = self.ext_dir or self.ext_src
```
boegel, line 138: As a side effect of the changes in https://github.com/easybuilders/easybuild-framework/pull/3426, this doesn't work as intended anymore...

By the time you reach this code, `self.start_dir` has already been changed to the value of `self.ext_dir`, because the existence check done in `_set_start_dir` on the relative path in `self.start_dir` has failed...

boegel: This was actually a bug introduced with https://github.com/easybuilders/easybuild-framework/pull/3426, should be fixed with https://github.com/easybuilders/easybuild-framework/pull/3435...

boegel:
Tested https://github.com/easybuilders/easybuild-easyconfigs/pull/11136 on top of this + the fix in https://github.com/easybuilders/easybuild-framework/pull/3435, works like a charm!

boegel (approved):
lgtm

PR #2129, opened 2020-08-20
fix code style issues in b*/.py and c/*.py

Flamefire:


easybuild/easyblocks/c/cp2k.py
```diff
@@ -267,10 +267,10 @@ def prepmodinc(self):
             # get list of modinc source files
             modincdir = os.path.join(imkl, self.cfg["modincprefix"], 'include')
 
-            if type(self.cfg["modinc"]) == list:
+            if isinstance(self.cfg["modinc"], list):
                 modfiles = [os.path.join(modincdir, x) for x in self.cfg["modinc"]]
 
-            elif type(self.cfg["modinc"]) == bool and type(self.cfg["modinc"]):
+            elif isinstance(self.cfg["modinc"], bool) and type(self.cfg["modinc"]):
```
boegel, line 273: @Flamefire This doesn't look right... Should probably be:

```python
elif isinstance(self.cfg["modinc"], bool) and self.cfg["modinc"):
```

please open an issue to follow up?

boegel (approved):
lgtm

easybuild/easyblocks/c/cp2k.py
Flamefire: https://github.com/easybuilders/easybuild-easyblocks/pull/2138

PR #2131, opened 2020-08-20
fix code style issues in e/*.py through g/*.py

Flamefire:


easybuild/easyblocks/f/foldx.py
```diff
@@ -58,7 +58,8 @@ def install_step(self):
                     if item in binaries:
                         shutil.copy2(os.path.join(self.cfg['start_dir'], item), bindir)
                         # make sure binary has executable permissions
-                        adjust_permissions(os.path.join(bindir, item), stat.S_IXUSR|stat.S_IXGRP|stat.S_IXOTH, add=True)
+                        adjust_permissions(os.path.join(bindir, item), stat.S_IXUSR |
+                                           stat.S_IXGRP | stat.S_IXOTH, add=True)
```
boegel, line 62:
```suggestion
                        perms = stat.S_IXUSR | stat.S_IXGRP | stat.S_IXOTH
                        adjust_permissions(os.path.join(bindir, item), perms, add=True)
```

easybuild/easyblocks/g/gamess_us.py
```diff
@@ -78,7 +77,8 @@ def __init__(self, *args, **kwargs):
             self.testdir = tempfile.mkdtemp()
             # make sure test dir doesn't contain [ or ], rungms csh script doesn't handle that well ("set: No match")
             if re.search(r'[\[\]]', self.testdir):
-                raise EasyBuildError("Temporary dir for tests '%s' will cause problems with rungms csh script", self.testdir)
+                raise EasyBuildError(
+                    "Temporary dir for tests '%s' will cause problems with rungms csh script", self.testdir)
```
boegel, line 81:
```suggestion
                error_msg = "Temporary dir for tests '%s' will cause problems with rungms csh script"
                raise EasyBuildError(error_msg, self.testdir)
```

easybuild/easyblocks/g/geant4.py
```diff
@@ -236,16 +236,18 @@ def configure_step(self):
...
                 self.g4system = filelist[0]
-    
+
             self.scriptdir = os.path.join(scriptdirbase, self.g4system)
             if not os.path.isdir(self.scriptdir):
                 raise EasyBuildError("Something went wrong. Dir: %s doesn't exist.", self.scriptdir)
-            self.log.info("The directory containing several important scripts to be copied was found: %s" % self.scriptdir)
+            self.log.info(
+                "The directory containing several important scripts to be copied was found: %s" %
+                self.scriptdir)
```
boegel, line 250:
```suggestion
            self.log.info("The directory containing several important scripts to be copied was found: %s",
                          self.scriptdir)

File: easybuild/easyblocks/g/geant4.py [hunk 4]
```diff
@@ -298,7 +300,7 @@ def install_step(self):
                         'G4NDL%s' % self.cfg['G4NDLVersion'],
                         'PhotonEvaporation%s' % self.cfg['PhotonEvaporationVersion'],
                         'RadioactiveDecay%s' % self.cfg['G4RadioactiveDecayVersion'],
-                       ]
+                        ]
```
boegel, line 303: I would replace this with:

```python
            datalist = [
                'G4ABLA%s' % self.cfg['G4ABLAVersion'],
                'G4EMLOW%s' % self.cfg['G4EMLOWVersion'],
                'G4NDL%s' % self.cfg['G4NDLVersion'],
                'PhotonEvaporation%s' % self.cfg['PhotonEvaporationVersion'],
                'RadioactiveDecay%s' % self.cfg['G4RadioactiveDecayVersion'],
            ]

boegel (changes requested):

easybuild/easyblocks/g/geant4.py
Flamefire: Good idea

boegel (approved):

PR #2133, opened 2020-08-20
fix code style issues in h/*.py through l/*.py

Flamefire:


easybuild/easyblocks/i/itac.py
```diff
@@ -69,7 +70,7 @@ def install_step(self):
                 super(EB_itac, self).move_after_install()
         else:
             silent = \
-"""
+                """
 [itac]
```
boegel, line 74:
```suggestion
            silent = """
[itac]

boegel (approved):
lgtm, retested HDF5 easyblock (with) `HDF5-1.10.6-gompi-2020a.eb` just to make sure, same for `ifort` easyblock

easybuild/easyblocks/i/itac.py
Flamefire: Done

boegel (approved):

PR #2138, opened 2020-08-20
correctly check whether 'modinc' easyconfig parameter is set to True in CP2K easyblock

Flamefire:
In the current version the result is always `True` now it honors when it is set to `False`

It is possible that this worked though, because IIRC True/False are singletons of their own type so `type(False) == False`


boegel:
@Flamefire Definitely did not work as intended:

```python
>>> bool(False)
False
>>> type(False)
<type 'bool'>
>>> bool(type(False))
True
```

boegel (approved):
lgtm

PR #2136, opened 2020-08-20
fix code style issues in n/*.py

Flamefire:

PR #2145, opened 2020-08-28
fix call to os.path.join in sanity checks of netcdf easyblock

lexming:


ocaisa (approved):

ocaisa:
Thanks for catching this @lexming

boegel:
Looks like this got introduced in https://github.com/easybuilders/easybuild-easyblocks/pull/2136/files#diff-4575cb4e99ad507d89d00e610af706b5R139, ouch...

Thanks for the fix @lexming!

PR #2158, opened 2020-09-02
Solve issue where pybind11 picks up on the system Python instead of one provided by a module

casparvl:
**Original issue**

The system python (or in our case: one from a custom sysroot) got picked up when trying to install `pybind11-2.4.3-GCCcore-8.3.0-Python-3.7.4`. This can be solved by setting `-DPYTHON_EXECUTABLE=$EBROOTPYTHON/bin/python` in the `configopts`. Apparently, the `pybind11` for the `GCCcore-9.3.0` toolchain indeed already did that.

Here, we solve it more generically by doing it in the EasyBlock. If a Python is loaded (i.e. get_software_root is not `None`), you'll always want to use that Python. Thus, a generic solution like this makes sense in my view.


easybuild/easyblocks/p/pybind11.py
```diff
@@ -54,6 +55,13 @@ def extra_options(extra_vars=None):
         extra_vars['download_dep_fail'][0] = True
         return extra_vars
 
+    def configure_step(self):
+        """Avoid that a system Python is picked up when a Python module is loaded"""
+        python_root = get_software_root('Python')
+        if python_root:
+            self.cfg.update('configopts', "-DPYTHON_EXECUTABLE=%s/bin/python" % python_root)
```
ocaisa, line 62: We should really check if the value already been set beforehand, this is how it's done in the LAMMPS easyblock:
```
option = 'PYTHON_EXECUTABLE'
if "-D%s=" % option not in self.cfg['configopts']:
  self.cfg.update('configopts', '-D%s=%s/bin/python' % (option, python_root))
```

ocaisa: We should also log the update

boegel: fixed in https://github.com/casparvl/easybuild-easyblocks/pull/2

casparvl: Thanks @boegel , I merged it in my branch so this PR is updated with the changes :)

ocaisa (approved):
LGTM

ocaisa:
Going in, thanks @casparvl !

PR #2161, opened 2020-09-03
OpenFOAM-support for arm64/aarch64

terjekv:

PR #2162, opened 2020-09-03
fix OpenFOAM easyblock to support installation on Arm/aarch64 systems

terjekv:
- More aarch64 compatibility fixes.
  - Different versions of OpenFOAM use different abbreviations for ARM64, or Arm64...


ocaisa:
@terjekv Can you close #2161 since this one replaces it?

terjekv:
Sorted. They were supposed to be compatible, but the develop merging seems to have gone awry. Sorry.  :-(

terjekv:
A test report is on its way, it'll take a few hours due to both OpenFOAM variants needing a full build.

terjekv:
This is odd, I ran a test report with the easyblock, but I'm not sure where the test result ended up?

boegel:
> This is odd, I ran a test report with the easyblock, but I'm not sure where the test result ended up?

You hit a known bug that we haven't fixed yet, see https://github.com/easybuilders/easybuild-framework/issues/3289

Your test report is there though, there's just no corresponding comment for it: https://gist.github.com/terjekv/993437e9e5b14f67e68a33dc57f54809

boegel (approved):
lgtm

PR #2163, opened 2020-09-04
fix Python package detection in TensorFlow easyblock and improve warning message

Flamefire:
Need to check against imported package not the pip names

### DIFF
--- a/easybuild/easyblocks/t/tensorflow.py
+++ b/easybuild/easyblocks/t/tensorflow.py
@@ -389,7 +389,7 @@ class EB_TensorFlow(PythonPackage):
         if ignored_system_deps:
             self.log.warning('For the following dependencies TensorFlow will download a copy because an ' +
                              'EB dependency was not found: \n%s\n' +
-                             'Dependencies: %s\n',
+                             'Dependencies: %s\n' +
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/t/tensorflow.py
```diff
@@ -394,6 +394,8 @@ def get_system_libs(self):
                              ', '.join(ignored_system_deps),
                              ', '.join(dep_names),
                              ', '.join(self.get_installed_python_packages()))
+        else:
+            self.log.info("All TensorFlow dependencies resolved via EasyBuild!")
```
Flamefire, line 398: @boegel nit: 
```suggestion
            self.log.info("All known TensorFlow dependencies resolved via EasyBuild!")
```

or something like that. We don't check for all TF dependencies (some don't have any ECs yet, see the dict) and some can't be provided via EB but are still downloaded

boegel: OK, thanks for clarifying! :+1:

boegel (approved):
lgtm, tested with https://github.com/easybuilders/easybuild-easyconfigs/pull/11109

PR #2164, opened 2020-09-07
explicitely enable/disable CUDA options in PyTorch easyblock + fix download check in sanity check + fix for disabling of *NNPACK on POWER systems

Flamefire:
The foss toolchain PyTorch fails because PyTorch tries to find CUDA libs and fails. This can be avoided by disabling CUDA.

This is now done.

Additionally I noticed double `USE_USE_` prefixes which I also fixed. Question here: Maybe don't add the USE_ prefix at all to make the EB searchable for Pytorch config options?

### DIFF
--- a/easybuild/easyblocks/p/pytorch.py
+++ b/easybuild/easyblocks/p/pytorch.py
@@ -177,7 +177,7 @@ class EB_PyTorch(PythonPackage):
 
         if get_cpu_architecture() == POWER:
             # *NNPACK is not supported on Power, disable to avoid warnings
-            nnpacks = ('NNPACK', 'QNNPACK', 'USE_PYTORCH_QNNPACK', 'USE_XNNPACK')
+            nnpacks = ('NNPACK', 'QNNPACK', 'PYTORCH_QNNPACK', 'XNNPACK')
             options.extend('USE_%s=0' % nnpack for nnpack in nnpacks)
 
         # Metal only supported on IOS which likely doesn't work with EB, so disabled
--- a/easybuild/easyblocks/p/pytorch.py
+++ b/easybuild/easyblocks/p/pytorch.py
@@ -155,25 +155,27 @@ class EB_PyTorch(PythonPackage):
         # you can disable this by including 'USE_IBVERBS=0' in 'custom_opts' in the easyconfig file
         options.append('USE_IBVERBS=1')
 
-        cudnn_root = get_software_root('cuDNN')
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/p/pytorch.py
```diff
@@ -155,29 +155,35 @@ def configure_step(self):
...
+            options.append('TORCH_CUDA_ARCH_LIST="%s"' % ';'.join(cuda_cc))
+        else:
+            # Disable CUDA
+            options.append('USE_CUDA=0')
 
         if get_cpu_architecture() == POWER:
             # *NNPACK is not supported on Power, disable to avoid warnings
-            nnpacks = ('NNPACK', 'QNNPACK', 'USE_PYTORCH_QNNPACK', 'USE_XNNPACK')
+            nnpacks = ('NNPACK', 'QNNPACK', 'PYTORCH_QNNPACK', 'XNNPACK')
             options.extend('USE_%s=0' % nnpack for nnpack in nnpacks)
```
boegel, line 187: @Flamefire As you suggested, it's probably better to not use the prefix approach, but make it something more straightforward that is easy to `grep`, like:

```python
options.extend(['USE_NNPACK=0', 'USE_QNNPACK=0', 'USE_PYTORCH_QNNPACK=0', 'USE_XNNPACK=0'])

boegel (changes requested):

easybuild/easyblocks/p/pytorch.py
Flamefire: Done https://github.com/easybuilders/easybuild-easyblocks/pull/2164/commits/ae7f815252333b68c5955434a23bb09975f6078e

Flamefire:
I added another bugfix which makes https://github.com/easybuilders/easybuild-easyconfigs/pull/11219 required for the next release: https://github.com/easybuilders/easybuild-easyblocks/pull/2164/commits/7e3eec833cce68834a5e3441766289cd088e0d0a

Seems I misunderstood when the `sanity_check_fail_msgs` list is checked so that was ignored

boegel:
EB_ARGS="--include-easyblocks-from-pr 2164"

boegel (approved):
lgtm

boegel:
Tested with https://github.com/easybuilders/easybuild-easyconfigs/pull/11219, good to go, thanks @Flamefire!

PR #2165, opened 2020-09-08
update TensorFlow easyblock to support installing TensorFlow 2.0 with dependencies provided through EasyBuild

Flamefire:
This adds $TF_SYSTEM_LIBS for TF 2.0 (verified with 2.0.0/fosscuda) and improves the logs. Most notably it mentions "$TF_SYSTEM_LIBS" so the log can be searched for that. I also slightly clarified the meaning of the dependency lists

As can be verified by reviewing the changes only TF 2.0 ECs are affected by this change so others don't need to be tested

This should be tested with https://github.com/easybuilders/easybuild-easyconfigs/pull/11233


boegel (approved):
lgtm, will test it too before merging

boegel:
tested with https://github.com/easybuilders/easybuild-easyconfigs/pull/11233, good to go, thanks @Flamefire!

PR #2166, opened 2020-09-09
update TensorFlow easyblock to put Bazel build files in build directory + avoid unnecessary runtime patching

Flamefire:
Currently we used /tmp (or better: tempdir as set by EB options) to put bazel build files in. However often users set builddir to some fast storage (e.g. /dev/shm) so using that instead can speed up the build of TensorFlow. This PR solves that and also  unifies some settings of various directories used

Additionally it removes some patches:
- The configure.py patch doesn't apply since 1.12.0: https://github.com/tensorflow/tensorflow/commit/56a99d6c28fd0ff238bd834bd32c2d3fc894b86c As this hasn't caused any issues yet it can be removed. It tried building some TF 2.x and 1.15 versions and no .cache/bazel folder is created
- Also I was told that TF is much more friendly to new toolchains. I hence tested removing our various, extensive patching of CROSSTOOL files with some TF 2.x ECs and it seems to be indeed no longer required. Bazel/TF seem to detect the built-in directories just well, so no need to manually specify them. I don't know when that happened (and am unsure how to find that) so I simply used 2.0 as the cutoff point. We've got 1.13.1 for 2019a, so I could test and use that as well, if wanted. Factoring this out into a separate function greatly simplifies the code making it more readable.

### DIFF
--- a/easybuild/easyblocks/t/tensorflow.py
+++ b/easybuild/easyblocks/t/tensorflow.py
@@ -404,11 +404,13 @@ class EB_TensorFlow(PythonPackage):
 
     def setup_build_dirs(self):
         """Setup temporary build directories"""
-        tmpdir = tempfile.mkdtemp(suffix='-bazel-tf')
+        # Tensorflow/Bazel needs a couple of directories where it stores build cache and artefacts
+        tmpdir = tempfile.mkdtemp(suffix='-bazel-tf', dir=self.builddir)
         self.output_root_dir = os.path.join(tmpdir, 'output_root')
         self.output_base_dir = os.path.join(tmpdir, 'output_base')
         self.output_user_root_dir = os.path.join(tmpdir, 'output_user_root')
         self.wrapper_dir = os.path.join(tmpdir, 'wrapper_bin')
+        # This (likely) needs to be a subdir of output_base
         self.install_base_dir = os.path.join(self.output_base_dir, 'inst_base')
 
     def configure_step(self):
--- a/easybuild/easyblocks/t/tensorflow.py
+++ b/easybuild/easyblocks/t/tensorflow.py
@@ -608,10 +608,13 @@ class EB_TensorFlow(PythonPackage):
         for (key, val) in sorted(config_env_vars.items()):
             env.setvar(key, val)
 
-        # patch configure.py (called by configure script) to avoid that Bazel abuses $HOME/.cache/bazel
-        regex_subs = [(r"(run_shell\(\['bazel')",
-                       r"\1, '--output_base=%s', '--install_base=%s'" % (tmpdir, os.path.join(tmpdir, 'inst_base')))]
-        apply_regex_substitutions('configure.py', regex_subs)
+        # Does no longer apply (and might not be required at all) since 1.12.0
+        if LooseVersion(self.version) < LooseVersion('1.12.0'):
+            # patch configure.py (called by configure script) to avoid that Bazel abuses $HOME/.cache/bazel
+            regex_subs = [(r"(run_shell\(\['bazel')",
+                           r"\1, '--output_base=%s', '--install_base=%s'" % (self.output_base_dir,
+                                                                             self.install_base_dir))]
+            apply_regex_substitutions('configure.py', regex_subs)
 
         # Tell Bazel to not use $HOME/.cache/bazel at all
         # See https://docs.bazel.build/versions/master/output_directories.html
@@ -619,12 +622,8 @@ class EB_TensorFlow(PythonPackage):
         cmd = self.cfg['preconfigopts'] + './configure ' + self.cfg['configopts']
         run_cmd(cmd, log_all=True, simple=True)
 
-    def build_step(self):
-        """Custom build procedure for TensorFlow."""
-
-        # pre-create target installation directory
-        mkdir(os.path.join(self.installdir, self.pylibdir), parents=True)
-
+    def patch_crosstool_files(self):
+        """Patches the CROSSTOOL files to include EasyBuild provided compiler paths"""
         inc_paths, lib_paths = [], []
 
         gcc_root = get_software_root('GCCcore') or get_software_root('GCC')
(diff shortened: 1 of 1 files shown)


Flamefire:
Test:

### x86

- [ ] TensorFlow-1.13.1-foss-2019a-Python-3.7.2.eb
- [ ] TensorFlow-1.13.1-fosscuda-2019a-Python-3.7.2.eb
- [ ] TensorFlow-1.14.0-foss-2019a-Python-3.7.2.eb
- [ ] TensorFlow-1.14.0-fosscuda-2019a-Python-3.7.2.eb
- [ ] TensorFlow-1.15.0-foss-2019b-Python-3.7.4.eb
- [ ] TensorFlow-1.15.0-fosscuda-2019b-Python-3.7.4.eb
- [ ] TensorFlow-1.15.2-foss-2019b-Python-3.7.4.eb
- [ ] TensorFlow-1.15.2-fosscuda-2019b-Python-3.7.4.eb
- [ ] TensorFlow-2.0.0-foss-2019a-Python-3.7.2.eb
- [ ] TensorFlow-2.0.0-fosscuda-2019b-Python-3.7.4.eb
- [ ] TensorFlow-2.0.1-foss-2019a-Python-3.7.2.eb
- [ ] TensorFlow-2.0.1-fosscuda-2019b-Python-3.7.4.eb
- [ ] TensorFlow-2.1.0-foss-2019b-Python-3.7.4.eb
- [ ] TensorFlow-2.1.0-fosscuda-2019b-Python-3.7.4.eb
- [ ] TensorFlow-2.2.0-foss-2019b-Python-3.7.4.eb
- [ ] TensorFlow-2.2.0-fosscuda-2019b-Python-3.7.4.eb

### POWER

- [ ] TensorFlow-1.15.0-foss-2019b-Python-3.7.4.eb
- [ ] TensorFlow-1.15.0-fosscuda-2019b-Python-3.7.4.eb
- [ ] TensorFlow-1.15.2-foss-2019b-Python-3.7.4.eb
- [ ] TensorFlow-1.15.2-fosscuda-2019b-Python-3.7.4.eb
- [ ] TensorFlow-2.0.0-fosscuda-2019b-Python-3.7.4.eb
- [ ] TensorFlow-2.0.1-fosscuda-2019b-Python-3.7.4.eb
- [ ] TensorFlow-2.1.0-foss-2019b-Python-3.7.4.eb
- [ ] TensorFlow-2.1.0-fosscuda-2019b-Python-3.7.4.eb
- [ ] TensorFlow-2.2.0-foss-2019b-Python-3.7.4.eb
- [ ] TensorFlow-2.2.0-fosscuda-2019b-Python-3.7.4.eb
- [ ] TensorFlow-2.3.0-fosscuda-2019b-Python-3.7.4.eb

Flamefire:
@boegel All tests finished successfully. The 1.13.1 fails due to https://github.com/easybuilders/easybuild-easyconfigs/issues/11253

So IMO good to go.

easybuild/easyblocks/t/tensorflow.py
```diff
@@ -423,14 +432,14 @@ def configure_step(self):
                 filtered_path = os.pathsep.join([p for fil in path_filter for p in path if fil not in p])
                 env.setvar(var, filtered_path)
 
-        wrapper_dir = os.path.join(tmpdir, 'bin')
-        use_wrapper = False
+        self.setup_build_dirs()
```
boegel, line 435: @Flamefire Why not do this in the easyblock constructor, since it doesn't require that the build environment is set up yet?

easybuild/easyblocks/t/tensorflow.py
```diff
@@ -681,13 +689,24 @@ def build_step(self):
...
-        cmd = [self.cfg['prebuildopts'], 'bazel', '--output_base=%s' % tmpdir,
-               '--install_base=%s' % os.path.join(tmpdir, 'inst_base'),
-               '--output_user_root=%s' % user_root_tmpdir, 'build']
+        cmd = [self.cfg['prebuildopts'],
+               'bazel',
+               '--output_base=%s' % self.output_base_dir,
+               '--install_base=%s' % self.install_base_dir,
+               '--output_user_root=%s' % self.output_user_root_dir,
+               'build'
+               ]
```
boegel, line 709:
```suggestion
        cmd = [
            self.cfg['prebuildopts'],
            'bazel',
            '--output_base=%s' % self.output_base_dir,
            '--install_base=%s' % self.install_base_dir,
            '--output_user_root=%s' % self.output_user_root_dir,
            'build',
        ]

File: easybuild/easyblocks/t/tensorflow.py [hunk 3]
```diff
@@ -778,12 +797,6 @@ def test_step(self):
 
     def install_step(self):
         """Custom install procedure for TensorFlow."""
-
-        # avoid that pip (ab)uses $HOME/.cache/pip
-        # cfr. https://pip.pypa.io/en/stable/reference/pip_install/#caching
-        env.setvar('XDG_CACHE_HOME', tempfile.gettempdir())
-        self.log.info("Using %s as pip cache directory", os.environ['XDG_CACHE_HOME'])
```
boegel, line 785: This was added for a good reason, I don't think we should stop setting `$XDG_CACHE_HOME`?

boegel (changes requested):

easybuild/easyblocks/t/tensorflow.py
Flamefire: Yes. See the commit message: `Remove setting of XDG_CACHE_HOME (done in PythonPackage)` ;)

So seems this was moved up one EasyBlock. I can rebuild again just to double-check this is actually used

easybuild/easyblocks/t/tensorflow.py
Flamefire: Because it is creating folders (in /tmp) which I see more fit for the configure step. Can't really avoid that as we need that unique folder

easybuild/easyblocks/t/tensorflow.py
Flamefire: From the log send in slack:
```
== 2020-09-15 17:31:46,045 pythonpackage.py:260 INFO Using default value for expected module name (lowercase software name): 'tensorflow'
== 2020-09-15 17:31:46,045 pythonpackage.py:296 INFO Using pip with --no-deps option
== 2020-09-15 17:31:46,046 environment.py:97 INFO Environment variable XDG_CACHE_HOME set to /import/ebswbuild/home/easybuild/builddirs/builddir.bdw/zao/tmpdir/9983760/eb-4s0xF9 (previous value: '/import/ebswbuild/home/easybuild/builddirs/builddir.bdw/zao/tmpdir/9983760/eb-4s0xF9')
```

So I'd say this verifies this

boegel (approved):

boegel:
I've also tested this with a wide range of TensorFlow easyconfigs, no problems encountered whatsoever, so good to go.

Thanks a lot for your efforts on this @Flamefire!

PR #2169, opened 2020-09-10
unset $COLUMNS if it is set to '0' before running Perl's configure script

boegel:
When doing the following:

```shell
export COLUMNS=0; eb Perl-5.30.2-GCCcore-9.3.0.eb
```

`Perl`'s `Configure` script dies with:

```
THIS PACKAGE SEEMS TO BE INCOMPLETE.
```

because `ls` produces a warning:

```
ls: ignoring invalid width in environment variable COLUMNS: 0
```

I'm not sure *why* or how `$COLUMNS` got set to zero (I wasn't doing that myself), but simply undefining it when that's the case fixes the issue...


fgeorgatos (approved):
lgtm!

zao (approved):
LGTM

Micket (approved):
lgtm

PR #2172, opened 2020-09-11
improve workaround for JsonCpp dependency in TensorFlow easyblock

Flamefire:
The TensorFlow sources include the json sources as "include/json/..." which is unusual
As a workaround the root folder to JsonCpp is added to CPATH.
However for fixed versions this is not required and the include folder should be added instead.
A patch for 2.3 exists and will be part of all future releases. 2.3+ ECs should hence rather use the patch instead.
To avoid having to change existing ECs not using the patch but allow them to use the patch both folders are added.


boegel (approved):
lgtm, tested with a variety of `TensorFlow` easyconfigs...

PR #2175, opened 2020-09-14
updated CBLAS easyblock for toolchains that include imkl

Darkless012:


easybuild/easyblocks/c/cblas.py
```diff
@@ -56,8 +56,11 @@ def configure_step(self):
         self.cfg.update('buildopts', 'FC="%s"' % os.getenv('F77'))
         self.cfg.update('buildopts', 'CFLAGS="%s -DADD_"' % os.getenv('CFLAGS'))
         self.cfg.update('buildopts', 'FFLAGS="%s -DADD_"' % os.getenv('FFLAGS'))
+
+        extra_blas_libs = '-lmkl_intel_thread -lmkl_lapack95_lp64 -lmkl_intel_lp64 -lmkl_core'
```
boegel, line 60: @Darkless012 This should be done conditionally, only when installing `CBLAS` with a toolchain that includes `imkl`?

boegel (changes requested):

boegel:
@Darkless012 Also needs a sync with `develop`, it seems like you forked your branch from an outdated `develop` branch?

We've done some code style cleanup recently (see https://github.com/easybuilders/easybuild-easyblocks/pull/2129/files#diff-701f48385de9e1dfd0c15d2874629964), so the conflict should be trivial to resolve...

easybuild/easyblocks/c/cblas.py
Darkless012: oki, will fix that. Thanks :)

easybuild/easyblocks/c/cblas.py
```diff
@@ -55,12 +56,18 @@ def configure_step(self):
...
         blas_lib_dir = os.getenv('BLAS_LIB_DIR')
         blas_libs = []
         for blas_lib in os.getenv('BLAS_STATIC_LIBS').split(','):
             blas_lib = os.path.join(blas_lib_dir, blas_lib)
             if os.path.exists(blas_lib):
                 blas_libs.append(blas_lib)
+
+        if get_software_root("imkl"):
+            extra_blas_libs = '-lmkl_intel_thread -lmkl_lapack95_lp64 -lmkl_intel_lp64 -lmkl_core'
+            blas_libs += extra_blas_libs.split()
```
Flamefire, line 69: This will include the `-l` which I think is wrong. Are you sure? Why not simply `extra_blas_libs = ['mkl_intel_thread', 'lmkl_lapack95_lp64' , ...]`

Darkless012: thanks, I guess I got carried away :)

boegel (approved):
lgtm

Tested with existing `CBLAS-20110120-foss-2016b.eb` + the easyconfigs using a recent `intel` toolchain from https://github.com/easybuilders/easybuild-easyconfigs/pull/11265, works like a charm! :+1:

PR #2176, opened 2020-09-15
Workaround duplicate prefix path in Eigen CMake config

Flamefire:
The current EasyBlock causes a duplicate (install) prefix paths in the CMake files leading to /foo/Eigen/foo/Eigen/include as the paths causing dependencies to choke on that, also mostly minor as we have the full/correct path in CPATH

See https://gitlab.com/libeigen/eigen/-/issues/1990


akesandgren (approved):
LGTM

akesandgren:
Going in, thanks @Flamefire!

PR #2177, opened 2020-09-15
update CP2K easyblock to add support of ELPA and spglib libraries

schulkov:
introduce EB_Libint_Cmake easyblock for Libint >= 2.6.0


boegel:
replaced with #2178

PR #2116, opened 2020-08-12
update Libint easyblock for versions >= 2.6.0 + add custom easyconfig parameter to enable Fortran support

boegel:
(created using `eb --new-pr`)

PR #2178, opened 2020-09-15
update CP2K easyblock to add support of ELPA and spglib libraries

schulkov:
introduce EB_Libint_Cmake easyblock for Libint >= 2.6.0


boegel:
@schulkov Thanks a lot for your contribution!

I know that the custom Libint easyblock is related to CP2K, but it may be better to tackle that in a separate pull request from the changes you're making to the CP2K easyblock?

Especially since we already have an easyconfig file for Libint 2.6.0 that serves as a dependency for CP2K 7.1, see https://github.com/easybuilders/easybuild-easyconfigs/pull/11099 .

The existing Libint easyblock was updated to support this (see https://github.com/easybuilders/easybuild-easyblocks/pull/2116), which is an approach I strongly prefer to adding another custom easyblock for Libint.

Also, we're still using the non-CMake build procedure there, since although using CMake is recommended it leads to a broken  Fortran interface in Libint 2.6.0 (see https://github.com/evaleev/libint/pull/142) which is required for CP2K.

PR #2180, opened 2020-09-16
add easyblock for RAxML

lexming:
This easyblock automatically builds those binaries of RAxML that 
1. use optimizations available in the CPU micro-architecture of the host system
2. have features compatible with the toolchain used in the build

Current easyconfigs of RAxML build one arbitrary binary of RAxML and create a symlink with the name `raxmlHPC`. This is a bad approach because the different binaries of RAxML have different requirements and features. For instance, binaries using MPI have a minimum number of trees to process

> This is RAxML MPI Process Number: 0
>
>WARNING: The number of threads is currently set to 0
>You can specify the number of threads to run via -T numberOfThreads
>NumberOfThreads must be set to an integer value greater than 1
>
>RAxML, will now set the number of threads automatically to 2 !
>
>Error: you are running the parallel MPI program but only want to compute one tree
>For the MPI version you must specify a number of trees greater than 1 with the -# or -N option

### DIFF
--- a/easybuild/easyblocks/r/raxml.py
+++ b/easybuild/easyblocks/r/raxml.py
@@ -1,0 +1,153 @@
+##
+# Copyright 2009-2020 Ghent University
+#
+# This file is part of EasyBuild,
+# originally created by the HPC team of Ghent University (http://ugent.be/hpc/en),
+# with support of Ghent University (http://ugent.be/hpc),
+# the Flemish Supercomputer Centre (VSC) (https://www.vscentrum.be),
+# Flemish Research Foundation (FWO) (http://www.fwo.be/en)
+# and the Department of Economy, Science and Innovation (EWI) (http://www.ewi-vlaanderen.be/en).
+#
+# https://github.com/easybuilders/easybuild
+#
+# EasyBuild is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation v2.
+#
+# EasyBuild is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with EasyBuild.  If not, see <http://www.gnu.org/licenses/>.
+##
+"""
+EasyBuild support for building and installing RAxML, implemented as an easyblock
+@author: Alex Domingo (Vrije Universiteit Brussel)
+"""
+import os
+
+import easybuild.tools.environment as env
+from easybuild.easyblocks.generic.makecp import MakeCp
+from easybuild.framework.easyconfig import CUSTOM
+from easybuild.tools.systemtools import get_cpu_features
+
+
+class EB_RAxML(MakeCp):
+    """Support for building and installing MSM."""
+    @staticmethod
+    def extra_options(extra_vars=None):
+        """Change default values of options"""
+        extra = MakeCp.extra_options()
+        # files_to_copy is not mandatory here
+        extra['files_to_copy'][2] = CUSTOM
+        return extra
+
+    def __init__(self, *args, **kwargs):
+        """RAxML easyblock constructor, define class variables."""
+        super(EB_RAxML, self).__init__(*args, **kwargs)
+
+        # Set optmization of RAxML build for host micro-architecture
+        cpuopt_label = {
+            'sse3': 'SSE3',
+            'sse4_1': 'SSE3',
+            'sse4_2': 'SSE3',
+            'avx': 'AVX',
+            'avx1.0': 'AVX',  # on macOS, AVX is indicated with 'avx1.0' rather than 'avx'
+            'avx2': 'AVX2',
+        }
+        cpu_features = set([label for feat, label in cpuopt_label.items() if feat in get_cpu_features()])
+        self.log.debug("Enabling the following CPU optimizations for RAxML: %s", ', '.join(cpu_features))
+        cpu_features.update([None])  # add generic build
+
+        # Build features supported by RAxML, grouped in non-MPI and MPI
+        build_labels = {'nompi': [None, 'PTHREADS'], 'mpi': ['MPI', 'HYBRID']}
+
+        # List of Makefiles in the build
+        makefile = ('Makefile', 'gcc', '.')
+        # Always build non-MPI variants
+        self.target_makefiles = {
+            'nompi': self.make_filename_variants(cpu_features, build_labels['nompi'], *makefile),
+        }
+        if self.toolchain.options.get('usempi', None):
+            # Add MPI variants in their own group
+            self.target_makefiles.update({
+                'mpi': self.make_filename_variants(cpu_features, build_labels['mpi'], *makefile),
+            })
+
+        # List of binaries in the installation
+        binary = ('raxmlHPC', None, '-')
+        self.target_bin = self.make_filename_variants(build_labels['nompi'], cpu_features,  *binary)
+        if 'mpi' in self.target_makefiles:
+            self.target_bin.extend(self.make_filename_variants(build_labels['mpi'], cpu_features, *binary))
+
+    def make_filename_variants(self, main_feature, extra_feature, prefix, suffix, divider):
+        """Returns list of RAxML filenames for the combination of all given features"""
+
+        # Features are expected as lists
+        if not isinstance(main_feature, list):
+            main_feature = list(main_feature)
+        if not isinstance(extra_feature, list):
+            extra_feature = list(extra_feature)
+
+        # Permutations of features
+        if extra_feature:
+            raxml_variants = [(mf, xf) for mf in main_feature for xf in extra_feature]
+        else:
+            raxml_variants = main_feature
+
+        # Prepend/Append prefix/suffix
+        for n, variant in enumerate(raxml_variants):
+            full_filename = (prefix,) + variant + (suffix,)
+            full_filename = tuple(filter(None, full_filename))  # avoid doubling the divider
+            raxml_variants[n] = divider.join(full_filename)
+
+        return raxml_variants
+
+    def configure_step(self):
+        """No custom configuration step for RAxML"""
+        pass
+
+    def build_step(self):
+        """Build all binaries of RAxML compatible with host CPU architecture"""
+
+        # Compiler is manually set through 'buildopts'
+        cc = os.getenv('CC')
+        # Always use non-MPI compiler for non-MPI Makefiles
+        if 'mpi' in self.target_makefiles:
+            cc_seq = os.getenv('CC_SEQ')
+        else:
+            cc_seq = os.getenv('CC')
+
+        # Build selected RAxML makefiles
+        user_buildopts = self.cfg['buildopts']
+
+        self.log.debug("Building makefiles of RAxML with %s: %s", cc_seq, ', '.join(self.target_makefiles['nompi']))
+        for mf in self.target_makefiles['nompi']:
+            self.cfg['buildopts'] = '-f %s CC="%s" %s' % (mf, cc_seq, user_buildopts)
+            super(EB_RAxML, self).build_step()
+
+        if 'mpi' in self.target_makefiles:
+            self.log.debug("Building makefiles of RAxML with %s: %s", cc, ', '.join(self.target_makefiles['mpi']))
+            for mf in self.target_makefiles['mpi']:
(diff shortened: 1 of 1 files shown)


lexming:
Tests in https://github.com/easybuilders/easybuild-easyconfigs/pull/11294

easybuild/easyblocks/r/raxml.py
```diff
@@ -0,0 +1,152 @@
...
+"""
+import os
+
+from easybuild.easyblocks.generic.makecp import MakeCp
+from easybuild.framework.easyconfig import CUSTOM
+from easybuild.tools.systemtools import get_cpu_features
+
+
+class EB_RAxML(MakeCp):
+    """Support for building and installing MSM."""
```
boegel, line 37: `MSM` -> `RAxML`?

easybuild/easyblocks/r/raxml.py
```diff
@@ -0,0 +1,152 @@
...
+            'avx': 'AVX',
+            'avx1.0': 'AVX',  # on macOS, AVX is indicated with 'avx1.0' rather than 'avx'
+            'avx2': 'AVX2',
+        }
+        cpu_features = set([label for feat, label in cpuopt_label.items() if feat in get_cpu_features()])
+        self.log.debug("Enabling the following CPU optimizations for RAxML: %s", ', '.join(cpu_features))
+        cpu_features.update([None])  # add generic build
+
+        # Build features supported by RAxML, grouped in non-MPI and MPI
+        build_labels = {'nompi': [None, 'PTHREADS'], 'mpi': ['MPI', 'HYBRID']}
```
boegel, line 64: please reformat:

```suggestion
        build_labels = {
            'nompi': [None, 'PTHREADS'],
            'mpi': ['MPI', 'HYBRID'],
        }

File: easybuild/easyblocks/r/raxml.py [hunk 3]
```diff
@@ -0,0 +1,152 @@
...
+        # Set optmization of RAxML build for host micro-architecture
+        cpuopt_label = {
+            'sse3': 'SSE3',
+            'sse4_1': 'SSE3',
+            'sse4_2': 'SSE3',
+            'avx': 'AVX',
+            'avx1.0': 'AVX',  # on macOS, AVX is indicated with 'avx1.0' rather than 'avx'
+            'avx2': 'AVX2',
+        }
+        cpu_features = set([label for feat, label in cpuopt_label.items() if feat in get_cpu_features()])
```
boegel, line 59: @lexming Should we also take the use of `optarch=GENERIC` into account here?

And what do we do when `optarch` is customized (see https://easybuild.readthedocs.io/en/latest/Controlling_compiler_optimization_flags.html).

A better approach could be to talk to `archspec` (like @terjekv is doing in https://github.com/easybuilders/easybuild-easyblocks/pull/1946)?

boegel (changes requested):

easybuild/easyblocks/r/raxml.py
verdurin: @lexming any chance you could address the comments from @boegel ?

easybuild/easyblocks/r/raxml.py
lexming: fixed in https://github.com/easybuilders/easybuild-easyblocks/pull/2180/commits/f7afebdd183da07959f1fa9300501ee4923e1fad

easybuild/easyblocks/r/raxml.py
lexming: fixed in https://github.com/easybuilders/easybuild-easyblocks/pull/2180/commits/f7afebdd183da07959f1fa9300501ee4923e1fad

easybuild/easyblocks/r/raxml.py
lexming: it's been a while, but I have finally refactored this easyblock to add your requests:
* respect optimization settings from `--optarch` or `toolchainopts.optarch`
* with default `optarch`, host CPU features are auto-detected
* CPU features auto-detection is done with `archspec` preferably and it fallbacks to `get_cpu_features` if not installed

lexming:
## Test reports

* Auto-detection
  * [Skylake - AVX512](https://github.com/easybuilders/easybuild-easyconfigs/pull/17433#issuecomment-1445530280)
  * [Broadwell - AVX2](https://github.com/easybuilders/easybuild-easyconfigs/pull/17433#issuecomment-1445530432)
  * [Zen2 - AVX2](https://github.com/easybuilders/easybuild-easyconfigs/pull/17433#issuecomment-1445710575)
  * [Ivybridge - AVX](https://github.com/easybuilders/easybuild-easyconfigs/pull/17433#issuecomment-1445529914)

lexming:
Test report by @lexming
 * **SUCCESS** _RAxML-8.2.12-GCC-11.3.0.eb_ 
 * **SUCCESS** _RAxML-8.2.12-gompi-2022a.eb_ 

node300.hydra.os - Linux CentOS Linux 7.9.2009, x86_64, Intel(R) Xeon(R) Gold 6148 CPU @ 2.40GHz (skylake_avx512), Python 3.6.8
See https://gist.github.com/c11b4d9e09219333dbd36fcf85a32ba3 for a full test report.

boegel:
@lexming Some existing `RAxML` easyconfigs have `raxmlHPC -v` and `raxmlHPC -h` as sanity check commands, we can let this easyblock run those too?

easybuild/easyblocks/r/raxml.py
```diff
@@ -0,0 +1,194 @@
...
+import os
+
+from easybuild.easyblocks.generic.makecp import MakeCp
+from easybuild.framework.easyconfig import CUSTOM
+from easybuild.tools.build_log import EasyBuildError, print_warning
+from easybuild.tools.config import build_option
+from easybuild.tools.systemtools import get_cpu_features, HAVE_ARCHSPEC
+from easybuild.tools.toolchain.compiler import OPTARCH_GENERIC
+
+RAXML_BINARY_NAME = "raxmlHPC"
```
boegel, line 39: Not sure `RAXML_BINARY_NAME` is useful, since it's only used in one place currently?

easybuild/easyblocks/r/raxml.py
```diff
@@ -0,0 +1,194 @@
...
+# Supported instruction sets ordered by priority (high to low) and related CPU features
+RAXML_INSTRUCT_SETS = ["AVX2", "AVX", "SSE3"]
```
boegel, line 41: What about AVX-512? If there's a reason no to include that, add a comment on it?
Where can the list of supported instruction sets be found (docs link? README?)

Also, out of curiosity, does it also support NEON (for `aarch64`)?
If someone cares enough the easyblock can be updated later to also cover that, doesn't need to be done here.

easybuild/easyblocks/r/raxml.py
```diff
@@ -0,0 +1,194 @@
...
+RAXML_CPU_FEATURES = {
+    "SSE3": ["sse3", "see4_1", "sse4_2"],
```
boegel, line 43: typo:
```suggestion
    "SSE3": ["sse3", "sse4_1", "sse4_2"],

File: easybuild/easyblocks/r/raxml.py [hunk 7]
```diff
@@ -0,0 +1,194 @@
...
+        extra = MakeCp.extra_options()
+        # files_to_copy is not mandatory
+        extra["files_to_copy"][2] = CUSTOM
+        return extra
+
+    def __init__(self, *args, **kwargs):
+        """RAxML easyblock constructor, define class variables."""
+        super(EB_RAxML, self).__init__(*args, **kwargs)
+
+        def filter_optarch_features(support_features):
```
boegel, line 69: Make this a function outside of the easyblock? Same for others below.

Advantage is that tests can be added easily to CI for these, and code becomes quite a bit easier to parse for humans

easybuild/easyblocks/r/raxml.py
```diff
@@ -0,0 +1,194 @@
...
+            Check if host CPU architecture has given feature
+            """
+            if HAVE_ARCHSPEC:
+                import archspec.cpu
+
+                host = archspec.cpu.host()
+                return feature.lower() in host
+
+            try:
+                return any(f in get_cpu_features() for f in RAXML_CPU_FEATURES[feature])
```
boegel, line 116: Would be nicer with a single `return` as the end, I'm not a big fan of inline returns, see also https://realpython.com/python-return-statement/#remembering-the-return-value (but that's partially personal, I guess)

boegel: Maybe this function should be moved into framework, it's pretty generic... Same for `filter_optarch_features`

easybuild/easyblocks/r/raxml.py
```diff
@@ -0,0 +1,194 @@
...
+            except KeyError:
+                raise EasyBuildError("Unknown CPU feature level for RAxML: %s", feature)
+
+        def list_filename_variants(main_features, extra_features, prefix, suffix, divider):
```
boegel, line 120: This should also go outside of easyblock, maybe rename to `list_raxml_filename_variants` to indicate it's specific to `RAxML`

easybuild/easyblocks/r/raxml.py
```diff
@@ -0,0 +1,194 @@
...
+            return file_variants
+
+        # Filter target CPU instructions sets according to optarch
+        cpu_features = filter_optarch_features(RAXML_INSTRUCT_SETS)
+
+        if len(cpu_features) > 1:
+            # unrestricted optimization settings, set optimization level for host micro-architecture
+            cpu_features = [feat for feat in cpu_features if has_cpu_feature(feat)]
+            dbg_msg = "Enabling the following CPU optimizations for RAxML by autodetection: %s"
+            self.log.debug(dbg_msg, ", ".join(cpu_features))
```
boegel, line 145: Make this `info` log message?

easybuild/easyblocks/r/raxml.py
```diff
@@ -0,0 +1,194 @@
...
+            # add generic build
+            cpu_features.append(None)
```
boegel, line 147: Can you clarify (in the comment) why this is done?

easybuild/easyblocks/r/raxml.py
```diff
@@ -0,0 +1,194 @@
...
+
+        # Set parallelization level of RAxML for current toolchain
+        parallel_features = RAXML_PARALLEL_FEATURES["nompi"]
+        if self.toolchain.options.get("usempi", None):
+            parallel_features.extend(RAXML_PARALLEL_FEATURES["mpi"])
+
+        # List of builds to carry out
+        self.target_makefiles = list_filename_variants(cpu_features, parallel_features, "Makefile", "gcc", ".")
```
boegel, line 155: Is always using `gcc` OK? What if we're installing `RAxML` with Intel compilers?

easybuild/easyblocks/r/raxml.py
```diff
@@ -0,0 +1,194 @@
...
+
+        # Build selected RAxML makefiles
+        user_buildopts = self.cfg["buildopts"]
+
+        for mf in self.target_makefiles:
+            cc_opt = compiler
+            if not any(feature in mf for feature in RAXML_PARALLEL_FEATURES["mpi"]):
+                cc_opt = compiler_nompi
+            self.cfg["buildopts"] = '-f %s CC="%s" %s' % (mf, cc_opt, user_buildopts)
+            self.log.debug("Building RAxML makefile with %s: %s", cc_opt, mf)
```
boegel, line 175: `log.info`?

easybuild/easyblocks/r/raxml.py
```diff
@@ -0,0 +1,194 @@
...
+        super(EB_RAxML, self).install_step()
+
+    def sanity_check_step(self):
+        """Custom sanity check for RAxML."""
+
+        custom_paths = {
+            "files": [os.path.join("bin", x) for x in self.target_bins],
+            "dirs": ["share/manual", "share/usefulScripts"],
+        }
+        super(EB_RAxML, self).sanity_check_step(custom_paths=custom_paths)
```
boegel, line 194: Also include a custom sanity check command like `raxmlHPC -h`?

akesandgren:
@lexming this easyblock needs some attention...

easybuild/easyblocks/r/raxml.py
ocaisa: They don't support that in their makefiles, this choice reflects what they actually set in `CFLAGS` (they do not respect the `CFLAGS` in the environment)

boegel:
Just to have a cross-reference: see also:
* https://github.com/easybuilders/easybuild-easyconfigs/pull/23975

lexming:
@ocaisa I fully agree, closing this in favour of https://github.com/easybuilders/easybuild-easyconfigs/pull/23975

PR #2190, opened 2020-09-28
add custom easyblock for NVHPC (aka PGI)

AndiH:
PGI beyond 20.4 is not called PGI any more but _NVIDIA HPC SDK_ (NVHPC), with 20.7 being the first version; https://developer.nvidia.com/hpc-sdk.

This pull request is part of a set of three pull requests to introduce NVHPC to EasyBuild ([1](https://github.com/easybuilders/easybuild-framework/pull/3454), [2](https://github.com/easybuilders/easybuild-easyblocks/pull/2190), [3](https://github.com/easybuilders/easybuild-easyconfigs/pull/11391)). The stack has been successfully deployed at JSC and is under use by our HPC users.  
The EasyBuild files are created after discussions with the NVHPC product owners, taking into account HPC practices and EasyBuild conventions.

This request is for pulling the NVHPC EasyBlock. It is based on the PGI EasyBlock but cleaned-up and extended for NVHPC.

Additional options are included; for example `default_cuda_version`, `compute_capability` and several `module_*` options.  The former two are passed to the NVHPC's `install` script which _configures_ NVHPC for usage with a certain CUDA or device Compute Capability.  
The `module_*` options refer to options being passed to the LMod module where additional directories beyond the default directories are included into the environment. As an example, `nvhpc_nvshmem_basedir` adds the directory of NVHPC in which NVSHMEM is included into `CPATH` and `LD_LIBRARY_PATH` with their corresponding sub-directories. The options exist to enable HPC centres to rather provide their own installations of the packaged libraries.

### DIFF
--- a/easybuild/easyblocks/n/nvhpc.py
+++ b/easybuild/easyblocks/n/nvhpc.py
@@ -180,12 +180,15 @@ class EB_NVHPC(PackedBinary):
             self.log.info("Removing $CPATH entry: %s", dirs['CPATH'])
             del dirs['CPATH']
 
-        # EasyBlock Options:
-        ## BYO Compilers allows for using NVHPC's libraries and tools with other, external compilers
+        # EasyBlock option parsing follows:
+        # BYO Compilers:
+        # Use NVHPC's libraries and tools with other, external compilers
         if self.cfg['module_byo_compilers']:
             if 'PATH' in dirs:
                 del dirs["PATH"]
-        ## NVHPC is shipped with a compiled OpenMPI installation; enable it by setting the according environment variables
+        # Own MPI:
+        # NVHPC is shipped with a compiled OpenMPI installation
+        # Enable it by setting according environment variables
         if self.cfg['module_nvhpc_own_mpi']:
             self.nvhpc_mpi_basedir = os.path.join(self.nvhpc_install_subdir, "comm_libs", "mpi")
             env_vars_dirs = {
@@ -194,7 +197,9 @@ class EB_NVHPC(PackedBinary):
                 'LD_LIBRARY_PATH': 'lib'
             }
             self._nvhpc_extended_components(dirs, self.nvhpc_mpi_basedir, env_vars_dirs)
-        ## NVHPC is shipped with math libraries in a dedicated folder; enable them by setting the according environment variables
+        # Math Libraries:
+        # NVHPC is shipped with math libraries (in a dedicated folder)
+        # Enable them by setting according environment variables
         if self.cfg['module_add_math_libs']:
             self.nvhpc_math_basedir = os.path.join(self.nvhpc_install_subdir, "math_libs")
             env_vars_dirs = {
@@ -202,14 +207,18 @@ class EB_NVHPC(PackedBinary):
                 'LD_LIBRARY_PATH': 'lib64'
             }
             self._nvhpc_extended_components(dirs, self.nvhpc_math_basedir, env_vars_dirs)
-        ## NVHPC is shipped with NVIDIA's GPU profilers; enable them by setting the according environment variables
+        # GPU Profilers:
+        # NVHPC is shipped with NVIDIA's GPU profilers (Nsight Compute/Nsight Systems)
+        # Enable them by setting the according environment variables
         if self.cfg['module_add_profilers']:
             self.nvhpc_profilers_basedir = os.path.join(self.nvhpc_install_subdir, "profilers")
             env_vars_dirs = {
                 'PATH': ['Nsight_Compute', 'Nsight_Systems/bin']
             }
             self._nvhpc_extended_components(dirs, self.nvhpc_profilers_basedir, env_vars_dirs)
-        ## NVHPC is shipped with NCCL; enable it by setting the according environment variables
+        # NCCL:
+        # NVHPC is shipped with NCCL
+        # Enable it by setting the according environment variables
         if self.cfg['module_add_nccl']:
             self.nvhpc_nccl_basedir = os.path.join(self.nvhpc_install_subdir, "comm_libs", "nccl")
             env_vars_dirs = {
@@ -217,7 +226,9 @@ class EB_NVHPC(PackedBinary):
                 'LD_LIBRARY_PATH': 'lib'
             }
             self._nvhpc_extended_components(dirs, self.nvhpc_nccl_basedir, env_vars_dirs)
-        ## NVHPC is shipped with NVSHMEM; enable it by setting the according environment variables
+        # NVSHMEM:
+        # NVHPC is shipped with NVSHMEM
+        # Enable it by setting the according environment variables
         if self.cfg['module_add_nvshmem']:
             self.nvhpc_nvshmem_basedir = os.path.join(self.nvhpc_install_subdir, "comm_libs", "nvshmem")
             env_vars_dirs = {
@@ -225,7 +236,9 @@ class EB_NVHPC(PackedBinary):
                 'LD_LIBRARY_PATH': 'lib'
             }
             self._nvhpc_extended_components(dirs, self.nvhpc_nvshmem_basedir, env_vars_dirs)
-        ## NVHPC is shipped with CUDA; rather use this CUDA than an external CUDA module (via $CUDA_HOME) by setting the according environment variables
+        # CUDA:
+        # NVHPC is shipped with CUDA (possibly multiple versions)
+        # Rather use this CUDA than an external CUDA (via $CUDA_HOME) by setting according environment variables
         if self.cfg['module_add_cuda']:
             self.nvhpc_cuda_basedir = os.path.join(self.nvhpc_install_subdir, "cuda")
             env_vars_dirs = {
--- a/easybuild/easyblocks/n/nvhpc.py
+++ b/easybuild/easyblocks/n/nvhpc.py
@@ -113,11 +113,11 @@ class EB_NVHPC(PackedBinary):
         cfg_default_compute_capability = build_option('cuda_compute_capabilities')
         if cfg_default_compute_capability is not None:
             default_compute_capability = cfg_default_compute_capability
-        elif ec_default_compute_capability is not None:
+        elif ec_default_compute_capability and ec_default_compute_capability is not None:
             default_compute_capability = ec_default_compute_capability
         else:
             error_msg = "A default Compute Capability is needed for installation of NVHPC."
-            error_msg += "Please provide it either in the easyconfig file like 'cuda_compute_capabilities=7.0',"
+            error_msg += "Please provide it either in the easyconfig file like 'cuda_compute_capabilities=\"7.0\"',"
             error_msg += "or use 'eb --cuda-compute-capabilities=7.0' from the command line."
             raise EasyBuildError(error_msg)
 
--- a/easybuild/easyblocks/n/nvhpc.py
+++ b/easybuild/easyblocks/n/nvhpc.py
@@ -44,7 +44,8 @@ from easybuild.easyblocks.generic.packedbinary import PackedBinary
 from easybuild.framework.easyconfig import CUSTOM
 from easybuild.tools.filetools import adjust_permissions, write_file
 from easybuild.tools.run import run_cmd
-from easybuild.tools.modules import get_software_root
+from easybuild.tools.modules import get_software_root, get_software_version
+from easybuild.tools.build_log import EasyBuildError
 
 
 # contents for siterc file to make PGI/NVHPC pick up $LIBRARY_PATH
@@ -74,7 +75,7 @@ class EB_NVHPC(PackedBinary):
     @staticmethod
     def extra_options():
         extra_vars = {
-            'default_cuda_version': ["10.1", "CUDA Version to be used as default (10.2, 11.0, ...)", CUSTOM],
+            'default_cuda_version': [None, "CUDA Version to be used as default (10.2, 11.0, ...)", CUSTOM],
             'compute_capability':   ["70", "Compute Capability (70, 80, ...)", CUSTOM],
             'module_byo_compilers': [False, "BYO Compilers: Remove compilers from module", CUSTOM],
             'module_nvhpc_own_mpi': [False, "Add NVHPC's packaged OpenMPI to module", CUSTOM],
@@ -95,10 +96,23 @@ class EB_NVHPC(PackedBinary):
     def install_step(self):
         """Install by running install command."""
 
+        default_cuda_version = self.cfg['default_cuda_version']
+        if default_cuda_version is None:
+            module_cuda_version_full = get_software_version('CUDA')
+            if module_cuda_version_full is not None:
+                default_cuda_version = '.'.join(module_cuda_version_full.split('.')[:2])
+            else:
+                error_msg = "A default CUDA version is needed for installation of NVHPC. "
+                error_msg += "It can not be determined automatically and needs to be added manually. "
+                error_msg += "You can edit the easyconfig file, "
+                error_msg += "or use 'eb --try-amend=default_cuda_version=<version>'."
+                raise EasyBuildError(error_msg)
+
+
         nvhpc_env_vars = {
             'NVHPC_INSTALL_DIR': self.installdir,
             'NVHPC_SILENT': 'true',
-            'NVHPC_DEFAULT_CUDA': str(self.cfg['default_cuda_version']),  # 10.2, 11.0
+            'NVHPC_DEFAULT_CUDA': str(default_cuda_version),  # 10.2, 11.0
             'NVHPC_STDPAR_CUDACC': str(self.cfg['compute_capability']),  # 70, 80
             }
         cmd = "%s ./install" % ' '.join(['%s=%s' % x for x in sorted(nvhpc_env_vars.items())])
--- a/easybuild/easyblocks/n/nvhpc.py
+++ b/easybuild/easyblocks/n/nvhpc.py
@@ -44,6 +44,7 @@ from easybuild.framework.easyconfig import CUSTOM
 from easybuild.tools.filetools import adjust_permissions, write_file
 from easybuild.tools.run import run_cmd
 from easybuild.tools.modules import get_software_root, get_software_version
+from easybuild.tools.config import build_option
 from easybuild.tools.build_log import EasyBuildError
 
 
@@ -74,15 +75,15 @@ class EB_NVHPC(PackedBinary):
     @staticmethod
     def extra_options():
         extra_vars = {
-            'default_cuda_version': [None, "CUDA Version to be used as default (10.2, 11.0, ...)", CUSTOM],
-            'compute_capability':   ["70", "Compute Capability (70, 80, ...)", CUSTOM],
-            'module_byo_compilers': [False, "BYO Compilers: Remove compilers from module", CUSTOM],
-            'module_nvhpc_own_mpi': [False, "Add NVHPC's packaged OpenMPI to module", CUSTOM],
-            'module_add_cuda':      [False, "Add NVHPC's CUDA to module", CUSTOM],
-            'module_add_math_libs': [False, "Add NVHPC's math libraries to module", CUSTOM],
-            'module_add_nccl':      [False, "Add NVHPC's NCCL library to module", CUSTOM],
-            'module_add_nvshmem':   [False, "Add NVHPC's NVSHMEM library to module", CUSTOM],
-            'module_add_profilers': [False, "Add NVHPC's NVIDIA Profilers to module", CUSTOM]
+            'default_cuda_version':      [None, "CUDA Version to be used as default (10.2 or 11.0 or ...)", CUSTOM],
+            'cuda_compute_capabilities': [None, "Compute Capability (7.0 or 8.0 or ...)", CUSTOM],
+            'module_byo_compilers':      [False, "BYO Compilers: Remove compilers from module", CUSTOM],
+            'module_nvhpc_own_mpi':      [False, "Add NVHPC's packaged OpenMPI to module", CUSTOM],
+            'module_add_cuda':           [False, "Add NVHPC's CUDA to module", CUSTOM],
+            'module_add_math_libs':      [False, "Add NVHPC's math libraries to module", CUSTOM],
+            'module_add_nccl':           [False, "Add NVHPC's NCCL library to module", CUSTOM],
+            'module_add_nvshmem':        [False, "Add NVHPC's NVSHMEM library to module", CUSTOM],
+            'module_add_profilers':      [False, "Add NVHPC's NVIDIA Profilers to module", CUSTOM]
         }
         return PackedBinary.extra_options(extra_vars)
 
@@ -107,11 +108,23 @@ class EB_NVHPC(PackedBinary):
                 error_msg += "or use 'eb --try-amend=default_cuda_version=<version>'."
                 raise EasyBuildError(error_msg)
 
+        ec_default_compute_capability = self.cfg['cuda_compute_capabilities']
+        cfg_default_compute_capability = build_option('cuda_compute_capabilities')[0]  # only single values supported
+        if ec_default_compute_capability is not None:
+            default_compute_capability = ec_default_compute_capability
+        elif cfg_default_compute_capability is not None:
+            default_compute_capability = cfg_default_compute_capability.replace(".", "")
+        else:
+            error_msg = "A default Compute Capability is needed for installation of NVHPC."
+            error_msg += "Please provide it either in the easyconfig file like 'cuda_compute_capabilits=7.0',"
+            error_msg += "or use 'eb --cuda-compute-capabilities=7.0' from the command line."
+            raise EasyBuildError(error_msg)
+
         nvhpc_env_vars = {
             'NVHPC_INSTALL_DIR': self.installdir,
             'NVHPC_SILENT': 'true',
             'NVHPC_DEFAULT_CUDA': str(default_cuda_version),  # 10.2, 11.0
-            'NVHPC_STDPAR_CUDACC': str(self.cfg['compute_capability']),  # 70, 80
+            'NVHPC_STDPAR_CUDACC': str(default_compute_capability),  # 70, 80; single values, no list!
             }
         cmd = "%s ./install" % ' '.join(['%s=%s' % x for x in sorted(nvhpc_env_vars.items())])
         run_cmd(cmd, log_all=True, simple=True)
--- a/easybuild/easyblocks/n/nvhpc.py
+++ b/easybuild/easyblocks/n/nvhpc.py
@@ -78,11 +78,11 @@ class EB_NVHPC(PackedBinary):
             'compute_capability':   ["70", "Compute Capability (70, 80, ...)", CUSTOM],
             'module_byo_compilers': [False, "BYO Compilers: Remove compilers from module", CUSTOM],
             'module_nvhpc_own_mpi': [False, "Add NVHPC's packaged OpenMPI to module", CUSTOM],
+            'module_add_cuda':      [False, "Add NVHPC's CUDA to module", CUSTOM],
             'module_add_math_libs': [False, "Add NVHPC's math libraries to module", CUSTOM],
-            'module_add_profilers': [False, "Add NVHPC's NVIDIA Profilers to module", CUSTOM],
             'module_add_nccl':      [False, "Add NVHPC's NCCL library to module", CUSTOM],
             'module_add_nvshmem':   [False, "Add NVHPC's NVSHMEM library to module", CUSTOM],
-            'module_add_cuda':      [False, "Add NVHPC's CUDA to module", CUSTOM]
+            'module_add_profilers': [False, "Add NVHPC's NVIDIA Profilers to module", CUSTOM]
         }
         return PackedBinary.extra_options(extra_vars)
 
@@ -123,7 +123,7 @@ class EB_NVHPC(PackedBinary):
                 if os.path.islink(path):
                     os.remove(path)
 
-        # install (or update) siterc file to make NVHPC consider $LIBRARY_PATHT
+        # install (or update) siterc file to make NVHPC consider $LIBRARY_PATH
         siterc_path = os.path.join(compilers_subdir, 'bin', 'siterc')
         write_file(siterc_path, SITERC_LIBRARY_PATH, append=True)
         self.log.info("Appended instructions to pick up $LIBRARY_PATH to siterc file at %s: %s",
@@ -142,11 +142,19 @@ class EB_NVHPC(PackedBinary):
         }
         super(EB_NVHPC, self).sanity_check_step(custom_paths=custom_paths)
 
-    def _nvhpc_extended_components(self, dirs, basepath, env_vars__folders):
-        for env_var, folder in env_vars__folders.items():
+    def _nvhpc_extended_components(self, dirs, basepath, env_vars_dirs):
+        """
+        Extends the `dirs` dictionary of key:environment_variables, value:list_of_directories with additional vars and dirs.
+        The dictionary key for a new env var will be created if it doesn't exist.
+        Also, the relative path specified in the `env_vars_dirs` dict is absolutized with the `basepath` prefix.
+        """
+        for env_var, folders in sorted(env_vars_dirs.items()):
             if env_var not in dirs:
                 dirs[env_var] = []
-            dirs[env_var].append(os.path.join(basepath, folder))
+            if isinstance(folders, list) is False:
+                folders = [folders]
+            for folder in folders:
+                dirs[env_var].append(os.path.join(basepath, folder))
 
     def make_module_req_guess(self):
         """Prefix subdirectories in NVHPC install dir considered for environment variables defined in module file."""
@@ -176,8 +184,7 @@ class EB_NVHPC(PackedBinary):
         ## NVHPC is shipped with NVIDIA's GPU profilers; enable them by setting the according environment variables
         if self.cfg['module_add_profilers'] is True:
             self.nvhpc_profilers_basedir = os.path.join(self.nvhpc_install_subdir, "profilers")
-            self._nvhpc_extended_components(dirs, self.nvhpc_profilers_basedir, {'PATH': 'Nsight_Compute'})
-            self._nvhpc_extended_components(dirs, self.nvhpc_profilers_basedir, {'PATH': 'Nsight_Systems/bin'})
+            self._nvhpc_extended_components(dirs, self.nvhpc_profilers_basedir, {'PATH': ['Nsight_Compute', 'Nsight_Systems/bin']})
         ## NVHPC is shipped with NCCL; enable it by setting the according environment variables
         if self.cfg['module_add_nccl'] is True:
             self.nvhpc_nccl_basedir = os.path.join(self.nvhpc_install_subdir, "comm_libs", "nccl")
--- a/easybuild/easyblocks/n/nvhpc.py
+++ b/easybuild/easyblocks/n/nvhpc.py
@@ -109,11 +109,11 @@ class EB_NVHPC(PackedBinary):
                 raise EasyBuildError(error_msg)
 
         ec_default_compute_capability = self.cfg['cuda_compute_capabilities']
-        cfg_default_compute_capability = build_option('cuda_compute_capabilities')[0]  # only single values supported
-        if ec_default_compute_capability is not None:
+        cfg_default_compute_capability = build_option('cuda_compute_capabilities')
+        if cfg_default_compute_capability is not None:
+            default_compute_capability = cfg_default_compute_capability[0].replace(".", "")
+        elif ec_default_compute_capability is not None:
             default_compute_capability = ec_default_compute_capability
-        elif cfg_default_compute_capability is not None:
-            default_compute_capability = cfg_default_compute_capability.replace(".", "")
         else:
             error_msg = "A default Compute Capability is needed for installation of NVHPC."
             error_msg += "Please provide it either in the easyconfig file like 'cuda_compute_capabilits=7.0',"
--- a/easybuild/easyblocks/n/nvhpc.py
+++ b/easybuild/easyblocks/n/nvhpc.py
@@ -176,7 +176,7 @@ class EB_NVHPC(PackedBinary):
         for env_var, folders in sorted(env_vars_dirs.items()):
             if env_var not in dirs:
                 dirs[env_var] = []
-            if isinstance(folders, list) is False:
+            if not isinstance(folders, list):
                 folders = [folders]
             for folder in folders:
                 dirs[env_var].append(os.path.join(basepath, folder))
--- a/easybuild/easyblocks/n/nvhpc.py
+++ b/easybuild/easyblocks/n/nvhpc.py
@@ -39,7 +39,6 @@ import re
 import stat
 import sys
 
-import easybuild.tools.environment as env
 from easybuild.easyblocks.generic.packedbinary import PackedBinary
 from easybuild.framework.easyconfig import CUSTOM
 from easybuild.tools.filetools import adjust_permissions, write_file
@@ -157,7 +156,7 @@ class EB_NVHPC(PackedBinary):
 
     def _nvhpc_extended_components(self, dirs, basepath, env_vars_dirs):
         """
-        Extends the `dirs` dictionary of key:environment_variables, value:list_of_directories with additional vars and dirs.
+        Extends `dirs` dict of key:environment_variables, value:list_of_directories with additional vars and dirs.
         The dictionary key for a new env var will be created if it doesn't exist.
         Also, the relative path specified in the `env_vars_dirs` dict is absolutized with the `basepath` prefix.
         """
--- a/easybuild/easyblocks/n/nvhpc.py
+++ b/easybuild/easyblocks/n/nvhpc.py
@@ -108,7 +108,6 @@ class EB_NVHPC(PackedBinary):
                 error_msg += "or use 'eb --try-amend=default_cuda_version=<version>'."
                 raise EasyBuildError(error_msg)
 
-
         nvhpc_env_vars = {
             'NVHPC_INSTALL_DIR': self.installdir,
             'NVHPC_SILENT': 'true',
@@ -191,8 +190,8 @@ class EB_NVHPC(PackedBinary):
         if self.cfg['module_nvhpc_own_mpi']:
             self.nvhpc_mpi_basedir = os.path.join(self.nvhpc_install_subdir, "comm_libs", "mpi")
             env_vars_dirs = {
-                'PATH': 'bin', 
-                'CPATH': 'include', 
+                'PATH': 'bin',
+                'CPATH': 'include',
                 'LD_LIBRARY_PATH': 'lib'
             }
             self._nvhpc_extended_components(dirs, self.nvhpc_mpi_basedir, env_vars_dirs)
--- a/easybuild/easyblocks/n/nvhpc.py
+++ b/easybuild/easyblocks/n/nvhpc.py
@@ -124,14 +124,16 @@ class EB_NVHPC(PackedBinary):
         if isinstance(default_compute_capability, list):
             _before_default_compute_capability = default_compute_capability
             default_compute_capability = _before_default_compute_capability[0]
-            print_warning(
-                "Replaced list of compute capabilities {before} with first element of list {after}"\
-                .format(
-                    before=_before_default_compute_capability,
-                    after=default_compute_capability
-                )
-            )
-        default_compute_capability.replace(".", "")
+            warning_msg = "Replaced list of compute capabilities {} ".format(_before_default_compute_capability)
+            warning_msg += "with first element of list {}".format(default_compute_capability)
+            print_warning(warning_msg)
+
+        # Remove dot-divider for CC; error out if it is not a string
+        if isinstance(default_compute_capability, str):
+            default_compute_capability = default_compute_capability.replace('.', '')
+        else:
+            raise EasyBuildError("Unexpected non-string value encountered for compute capability: %s",
+                                 default_compute_capability)
 
         nvhpc_env_vars = {
             'NVHPC_INSTALL_DIR': self.installdir,
--- a/easybuild/easyblocks/n/nvhpc.py
+++ b/easybuild/easyblocks/n/nvhpc.py
@@ -40,7 +40,6 @@ import stat
 import sys
 
 import easybuild.tools.environment as env
-from distutils.version import LooseVersion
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/n/nvhpc.py
```diff
@@ -0,0 +1,201 @@
...
+class EB_NVHPC(PackedBinary):
+    """
+    Support for installing the NVIDIA HPC SDK (NVHPC) compilers
+    """
+
+    @staticmethod
+    def extra_options():
+        extra_vars = {
+            'default_cuda_version': ["10.1", "CUDA Version to be used as default (10.2, 11.0, ...)", CUSTOM],
+            'compute_capability':   ["70", "Compute Capability (70, 80, ...)", CUSTOM],
```
boegel, line 79: Does NVHPC support installing for a set compute capabilities too?

If so, we should leverage the top-level `cuda_compute_capabilities` easyconfig parameter instead (introduced in #3382)

easybuild/easyblocks/n/nvhpc.py
```diff
@@ -0,0 +1,201 @@
...
+
+class EB_NVHPC(PackedBinary):
+    """
+    Support for installing the NVIDIA HPC SDK (NVHPC) compilers
+    """
+
+    @staticmethod
+    def extra_options():
+        extra_vars = {
+            'default_cuda_version': ["10.1", "CUDA Version to be used as default (10.2, 11.0, ...)", CUSTOM],
```
boegel, line 78: Does it make sense to hardcode any default version here?

How about using `None` as a default, and complaining when no specific version is specified?

At installation time, you can specify a value to be used using `eb --try-amend=default_cuda_version=10.1` for example (the error message produced when the value is still `None` could mention that, we do something like this in the [XALT easyblock](https://github.com/easybuilders/easybuild-easyblocks/blob/5e2a7577f2ddcc04b0364b577b9d84b520842540/easybuild/easyblocks/x/xalt.py#L71-L81)).

The problem with using a default version here is that it'll go irrelevant in a heart beat (well, ok, months or years) since CUDA is updated regularly.

easybuild/easyblocks/n/nvhpc.py
```diff
@@ -0,0 +1,201 @@
...
+            'compute_capability':   ["70", "Compute Capability (70, 80, ...)", CUSTOM],
+            'module_byo_compilers': [False, "BYO Compilers: Remove compilers from module", CUSTOM],
+            'module_nvhpc_own_mpi': [False, "Add NVHPC's packaged OpenMPI to module", CUSTOM],
+            'module_add_math_libs': [False, "Add NVHPC's math libraries to module", CUSTOM],
```
boegel, line 82: Please keep these sorted alphabetically here (easier for humans)

easybuild/easyblocks/n/nvhpc.py
```diff
@@ -0,0 +1,201 @@
...
+            'module_add_nvshmem':   [False, "Add NVHPC's NVSHMEM library to module", CUSTOM],
+            'module_add_cuda':      [False, "Add NVHPC's CUDA to module", CUSTOM]
+        }
+        return PackedBinary.extra_options(extra_vars)
+
+    def __init__(self, *args, **kwargs):
+        """Easyblock constructor, define custom class variables specific to NVHPC."""
+        super(EB_NVHPC, self).__init__(*args, **kwargs)
+
+        self.nvhpc_install_subdir = os.path.join('Linux_x86_64', self.version)
```
boegel, line 94: Does NVHPC also support non-x86_64 systems?

If so, we should use `get_cpu_architecture()` from `easybuild.tools.systemtools` here (and potentially do a mapping), like we do in the [CUDA easyblock](https://github.com/easybuilders/easybuild-easyblocks/blob/5e2a7577f2ddcc04b0364b577b9d84b520842540/easybuild/easyblocks/c/cuda.py#L64).

easybuild/easyblocks/n/nvhpc.py
```diff
@@ -0,0 +1,201 @@
...
+
+    def install_step(self):
+        """Install by running install command."""
+
+        nvhpc_env_vars = {
+            'NVHPC_INSTALL_DIR': self.installdir,
+            'NVHPC_SILENT': 'true',
+            'NVHPC_DEFAULT_CUDA': str(self.cfg['default_cuda_version']).lower(), #10.2, 11.0
```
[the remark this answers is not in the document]
boegel: How does `.lower()` make sense, since the value is a string of digits?

easybuild/easyblocks/n/nvhpc.py
```diff
@@ -0,0 +1,201 @@
...
+# split this value at colons, separate by -L, prepend 1st one by -L
+variable library_path is
+default($if($LIBRARY_PATH,-L$replace($LIBRARY_PATH,":", -L)));
+
+# add the -L arguments to the link line
+append LDLIBARGS=$library_path;
+
+# also include the location where libm & co live on Debian-based systems
+# cfr. https://github.com/easybuilders/easybuild-easyblocks/pull/919
+append LDLIBARGS=-L/usr/lib/x86_64-linux-gnu;
```
boegel, line 66: Should we also take into account `get_cpu_architecture()` here (see also below)?

easybuild/easyblocks/n/nvhpc.py
```diff
@@ -0,0 +1,201 @@
...
+        nvhpc_env_vars = {
+            'NVHPC_INSTALL_DIR': self.installdir,
+            'NVHPC_SILENT': 'true',
+            'NVHPC_DEFAULT_CUDA': str(self.cfg['default_cuda_version']).lower(), #10.2, 11.0
+            'NVHPC_STDPAR_CUDACC': str(self.cfg['compute_capability']).lower(), #70, #80
+            }
+        cmd = "%s ./install" % ' '.join(['%s=%s' % x for x in sorted(nvhpc_env_vars.items())])
+        run_cmd(cmd, log_all=True, simple=True)
+
+        # make sure localrc uses GCC in PATH, not always the system GCC, and does not use a system g77 but gfortran
```
boegel, line 108: I this comment still accurate? I need no mention of `gfortran` below?

easybuild/easyblocks/n/nvhpc.py
```diff
@@ -0,0 +1,201 @@
...
+
+        # If an OS libnuma is NOT found, makelocalrc creates symbolic links to libpgnuma.so
+        # If we use the EB libnuma, delete those symbolic links to ensure they are not used
+        if get_software_root("numactl"):
+            for filename in ["libnuma.so", "libnuma.so.1"]:
+                path = os.path.join(compilers_subdir, "lib", filename)
+                if os.path.islink(path):
+                    os.remove(path)
+
+        # install (or update) siterc file to make NVHPC consider $LIBRARY_PATHT
```
boegel, line 127: typo: `PATHT` -> `PATH`

easybuild/easyblocks/n/nvhpc.py
```diff
@@ -0,0 +1,201 @@
...
+        prefix = self.nvhpc_install_subdir
+        custom_paths = {
+            'files': [os.path.join(prefix, 'compilers', 'bin', x) for x in ['nvc', 'nvc++', 'nvfortran', 'siterc']],
+            'dirs': [os.path.join(prefix, 'compilers', 'bin'), os.path.join(prefix, 'compilers', 'lib'),
+                     os.path.join(prefix, 'compilers', 'include'), os.path.join(prefix, 'compilers', 'man')]
+        }
+        super(EB_NVHPC, self).sanity_check_step(custom_paths=custom_paths)
+
+    def _nvhpc_extended_components(self, dirs, basepath, env_vars__folders):
+        for env_var, folder in env_vars__folders.items():
```
boegel, line 147: Use `sorted(...items())`, which is helpful when humans need to debug stuff?

easybuild/easyblocks/n/nvhpc.py
```diff
@@ -0,0 +1,201 @@
...
+        """Custom sanity check for NVHPC"""
+        prefix = self.nvhpc_install_subdir
+        custom_paths = {
+            'files': [os.path.join(prefix, 'compilers', 'bin', x) for x in ['nvc', 'nvc++', 'nvfortran', 'siterc']],
+            'dirs': [os.path.join(prefix, 'compilers', 'bin'), os.path.join(prefix, 'compilers', 'lib'),
+                     os.path.join(prefix, 'compilers', 'include'), os.path.join(prefix, 'compilers', 'man')]
+        }
+        super(EB_NVHPC, self).sanity_check_step(custom_paths=custom_paths)
+
+    def _nvhpc_extended_components(self, dirs, basepath, env_vars__folders):
```
boegel, line 146: Please add a docstring for this method to describe what it does

Oh, and please just use `env_vars_folders` (no double underscore), or `env_vars_dirs` (dirs and folders are the same, no need to cause confusion?)

easybuild/easyblocks/n/nvhpc.py
```diff
@@ -0,0 +1,201 @@
...
+
+    def make_module_req_guess(self):
+        """Prefix subdirectories in NVHPC install dir considered for environment variables defined in module file."""
+        dirs = super(EB_NVHPC, self).make_module_req_guess()
+        for key in dirs:
+            dirs[key] = [os.path.join(self.nvhpc_install_subdir, 'compilers', d) for d in dirs[key]]
+
+        # $CPATH should not be defined in module for NVHPC, it causes problems
+        # cfr. https://github.com/easybuilders/easybuild-easyblocks/issues/830
+        if 'CPATH' in dirs:
```
boegel, line 160: Why not handle this above in the `for key in ...`?

easybuild/easyblocks/n/nvhpc.py
```diff
@@ -0,0 +1,201 @@
...
+            self.log.info("Removing $CPATH entry: %s", dirs['CPATH'])
+            del dirs['CPATH']
+
+        # EasyBlock Options:
+        ## BYO Compilers allows for using NVHPC's libraries and tools with other, external compilers
+        if self.cfg['module_byo_compilers'] is True:
```
boegel, line 166: Just use `if self.cfg['module_byo_compilers']:`? (same below)

easybuild/easyblocks/n/nvhpc.py
```diff
@@ -0,0 +1,201 @@
...
+            if 'PATH' in dirs:
+                del dirs["PATH"]
+        ## NVHPC is shipped with a compiled OpenMPI installation; enable it by setting the according environment variables
+        if self.cfg['module_nvhpc_own_mpi'] is True:
+            self.nvhpc_mpi_basedir = os.path.join(self.nvhpc_install_subdir, "comm_libs", "mpi")
+            self._nvhpc_extended_components(dirs, self.nvhpc_mpi_basedir, {'PATH': 'bin', 'CPATH': 'include', 'LD_LIBRARY_PATH': 'lib'})
```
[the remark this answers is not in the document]
boegel:
```suggestion
            env_vars_dirs = {
                'CPATH': 'include',
                'LD_LIBRARY_PATH': 'lib',
                'PATH': 'bin',
            }
            self._nvhpc_extended_components(dirs, self.nvhpc_mpi_basedir, env_vars_dirs)
```

(same pattern can be applied below)

easybuild/easyblocks/n/nvhpc.py
```diff
@@ -0,0 +1,201 @@
...
+        ## NVHPC is shipped with math libraries in a dedicated folder; enable them by setting the according environment variables
+        if self.cfg['module_add_math_libs'] is True:
+            self.nvhpc_math_basedir = os.path.join(self.nvhpc_install_subdir, "math_libs")
+            self._nvhpc_extended_components(dirs, self.nvhpc_math_basedir, {'CPATH': 'include', 'LD_LIBRARY_PATH': 'lib64'})
+        ## NVHPC is shipped with NVIDIA's GPU profilers; enable them by setting the according environment variables
+        if self.cfg['module_add_profilers'] is True:
+            self.nvhpc_profilers_basedir = os.path.join(self.nvhpc_install_subdir, "profilers")
+            self._nvhpc_extended_components(dirs, self.nvhpc_profilers_basedir, {'PATH': 'Nsight_Compute'})
+            self._nvhpc_extended_components(dirs, self.nvhpc_profilers_basedir, {'PATH': 'Nsight_Systems/bin'})
```
boegel, line 181: Enhance `_nvhpc_extended_components` so you can use `{'PATH': ['Nsight_Systems', 'Nsight_Systems/bin']}` ?

boegel (changes requested):
Please also take care of the code style suggestions.

Don't hesitate to reach out if you need any help!

easybuild/easyblocks/n/nvhpc.py
boegel: Actually there a better option than throwing an error: use the CUDA version of the loaded module (since you include `CUDA` as a dep)?

You can obtain it via `get_software_version('CUDA')` (could be `None` if `CUDA` is not loaded/not included as a dep)

easybuild/easyblocks/n/nvhpc.py
```diff
@@ -0,0 +1,201 @@
...
+            self.nvhpc_nccl_basedir = os.path.join(self.nvhpc_install_subdir, "comm_libs", "nccl")
+            self._nvhpc_extended_components(dirs, self.nvhpc_nccl_basedir, {'CPATH': 'include', 'LD_LIBRARY_PATH': 'lib'})
+        ## NVHPC is shipped with NVSHMEM; enable it by setting the according environment variables
+        if self.cfg['module_add_nvshmem'] is True:
+            self.nvhpc_nvshmem_basedir = os.path.join(self.nvhpc_install_subdir, "comm_libs", "nvshmem")
+            self._nvhpc_extended_components(dirs, self.nvhpc_nvshmem_basedir, {'CPATH': 'include', 'LD_LIBRARY_PATH': 'lib'})
+        ## NVHPC is shipped with CUDA; rather use this CUDA than an external CUDA module (via $CUDA_HOME) by setting the according environment variables
+        if self.cfg['module_add_cuda'] is True:
+            self.nvhpc_cuda_basedir = os.path.join(self.nvhpc_install_subdir, "cuda")
+            self._nvhpc_extended_components(dirs, self.nvhpc_cuda_basedir, {'PATH': 'bin', 'LD_LIBRARY_PATH': 'lib64', 'CPATH': 'include'})
```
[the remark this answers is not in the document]
AndiH: @boegel How do you want me to resolve the line length in these kind of lines?
```
self._nvhpc_extended_components(dirs, self.nvhpc_cuda_basedir, {'PATH': 'bin', 'LD_LIBRARY_PATH': 'lib64', 'CPATH': 'include'})
```

Would something like
```
            self._nvhpc_extended_components(
                dirs, 
                self.nvhpc_mpi_basedir, 
                {
                    'PATH': 'bin', 
                    'CPATH': 'include', 
                    'LD_LIBRARY_PATH': 'lib'
                }
            )
```
really better?

easybuild/easyblocks/n/nvhpc.py
```diff
@@ -0,0 +1,201 @@
...
+            self._nvhpc_extended_components(dirs, self.nvhpc_profilers_basedir, {'PATH': 'Nsight_Systems/bin'})
+        ## NVHPC is shipped with NCCL; enable it by setting the according environment variables
+        if self.cfg['module_add_nccl'] is True:
+            self.nvhpc_nccl_basedir = os.path.join(self.nvhpc_install_subdir, "comm_libs", "nccl")
+            self._nvhpc_extended_components(dirs, self.nvhpc_nccl_basedir, {'CPATH': 'include', 'LD_LIBRARY_PATH': 'lib'})
+        ## NVHPC is shipped with NVSHMEM; enable it by setting the according environment variables
+        if self.cfg['module_add_nvshmem'] is True:
+            self.nvhpc_nvshmem_basedir = os.path.join(self.nvhpc_install_subdir, "comm_libs", "nvshmem")
+            self._nvhpc_extended_components(dirs, self.nvhpc_nvshmem_basedir, {'CPATH': 'include', 'LD_LIBRARY_PATH': 'lib'})
+        ## NVHPC is shipped with CUDA; rather use this CUDA than an external CUDA module (via $CUDA_HOME) by setting the according environment variables
```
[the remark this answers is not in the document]
AndiH: Re 2#: I'm using Markdown-styled two levels of headers. There are two hashes because I consider these comments a level below `# EasyBlock Options:`.
Yay/nay?

easybuild/easyblocks/n/nvhpc.py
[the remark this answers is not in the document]
AndiH: It does not. I didn't pay attention when copying from `pgi.py`.

easybuild/easyblocks/n/nvhpc.py
AndiH: The options are given further to the NVHPC `install` script which, in turn, writes it to a `cudarc` file, similarly to the old PGI RC files. The contents after calling the EB block with 11.0 and 70 are
```
set DEFCUDAVERSION=11.0;
set DEFSTDPARCOMPUTECAP=70;
```

I don't know if setting a set of compute capabilities is supported, but my guess would be: yes. 

If the overriding `$NVHPC_STDPAR_CUDACC` (which we use) is not set, `install` tries to determine the compute capability itself by calling `nvaccelinfo -dev 0 | grep -i "default target" | cut -d":" -f2 | tr -d 'c\040\015\012\011'`; resulting in `70` on a V100 system.

Finally, to your question: I guess we could use `cuda_compute_capabilities`? How would that be specified in the easyconfig?

easybuild/easyblocks/n/nvhpc.py
AndiH: Let's answer your second suggestion first.

*I* do include the CUDA module as a dependency, but it is not strictly necessary. That's what `module_add_cuda` is about. NVHPC is shipped with a CUDA toolkit installation which actually the NVHPC owners would like you to use – as it is tested with NVHPC. For Jülich, I don't like this, because we _do_ have a reference system CUDA installation which should be the one in use, and not a shadow-version introduced with NVHPC. That said, I don't want to keep anyone from doing things differently (and for us, opting to do so in the future as well).

Now, coming back to your first suggestion: Having `None` is a good idea. It can not only be overridden by the command line flag `--try-amend` but still via the easyconfig, right? I'd like to have as much in the easyconfig as possible to make it documented concisely (and so I remember next time I install a NVHPC version 😀 ).

I'd implement the following sketch then:

* If `default_cuda_version` is set (which ever way), use this. If it is not,
* if `get_software_version('CUDA')` returns something, use this. If it does not,
* error out, as in XALT.

Does that sound good? I'll dive into how to do that then.

easybuild/easyblocks/n/nvhpc.py
AndiH: I've sorted them now in first level by importance and in second level alphabetically.

```
default_cuda_version': 
compute_capability':   
module_byo_compilers': 
module_nvhpc_own_mpi': 
module_add_cuda':      
module_add_math_libs': 
module_add_nccl':      
module_add_nvshmem':   
module_add_profilers': 
```

Ok?

easybuild/easyblocks/n/nvhpc.py
AndiH: According to their website, they support "Linux x86_64", "Linux OpenPOWER", "Linux Arm Server", "Windows x86 (eventually)".

I can have a quick check if the sub-directories are named in a sensible manner, but I don't have time (or capabilitie) to test all versions…

easybuild/easyblocks/n/nvhpc.py
AndiH: I don't know. This comes from the PGI EasyBlock and I didn't feel comfortable just deleting it.

easybuild/easyblocks/n/nvhpc.py
AndiH: I just left it there from PGI, but I can have a look.

easybuild/easyblocks/n/nvhpc.py
AndiH: ✅

easybuild/easyblocks/n/nvhpc.py
AndiH: This part I took from `pgi.py`. Should I change? I'd need to change all the following lines…

easybuild/easyblocks/n/nvhpc.py
AndiH: Good idea. It makes `_nvhpc_extended_components` somewhat more complex, but I think it's a good addition.

AndiH:
(Snapshot at end of work day today. Will continue after additional input. Note: Not tested on system.)

easybuild/easyblocks/n/nvhpc.py
AndiH: Changed as suggested for all cases.

easybuild/easyblocks/n/nvhpc.py
AndiH: I've implemented it accordingly in 25c798a7f5126769d09ba9eab3aa3f12952b48b3; I tested as much as I can and it seems to work.

One thing to note: In the easyconfig file (yet to be updated), `default_cuda_version = None` needs to be specified. Otherwise, EasyBuild would not allow overwriting it from the command line with the following error:

```
ERROR: Unknown easyconfig parameter: default_cuda_version (known: ['allow_prepend_abs_path', 'allow_system_deps', 'bitbucket_account', 'block', 'builddependencies', 'buildininstalldir', 'buildopts', 'buildstats', 'checksums', 'cleanupoldbuild', 'cleanupoldinstall', 'configopts', 'cuda_compute_capabilities', 'dependencies', 'deprecated', 'description', 'docpaths', 'docurls', 'dontcreateinstalldir', 'easyblock', 'easybuild_version', 'enhance_sanity_check', 'examples', 'exts_classmap', 'exts_default_options', 'exts_defaultclass', 'exts_download_dep_fail', 'exts_filter', 'exts_list', 'fix_perl_shebang_for', 'fix_python_shebang_for', 'github_account', 'group', 'hidden', 'hiddendependencies', 'homepage', 'include_modpath_extensions', 'installopts', 'keeppreviousinstall', 'keepsymlinks', 'key', 'license_file', 'license_server', 'license_server_port', 'maxparallel', 'modaliases', 'modaltsoftname', 'moddependpaths', 'modextrapaths', 'modextravars', 'modloadmsg', 'modluafooter', 'modtclfooter', 'module_depends_on', 'moduleclass', 'moduleforceunload', 'moduleloadnoconflict', 'multi_deps', 'multi_deps_load_default', 'name', 'onlytcmod', 'osdependencies', 'parallel', 'patches', 'postinstallcmds', 'prebuildopts', 'preconfigopts', 'preinstallopts', 'pretestopts', 'recursive_module_unload', 'runtest', 'sanity_check_commands', 'sanity_check_paths', 'site_contacts', 'skip', 'skipsteps', 'software_license', 'software_license_urls', 'source_urls', 'sources', 'start_dir', 'stop', 'testopts', 'tests', 'toolchain', 'toolchainopts', 'unpack_options', 'unwanted_env_vars', 'upstream_contacts', 'usage', 'version', 'versionprefix', 'versionsuffix', 'whatis'])
```

This is the intended behaviour, yes?

AndiH:
Some more work done. @boegel can you have a look at the open questions?

easybuild/easyblocks/n/nvhpc.py
damianam: There are a few keys to be deleted and/or tweaked. Having them in the `for` could make the code less readable I guess?

easybuild/easyblocks/n/nvhpc.py
damianam: I would suggest to leave that as "room for improvement" We do not have any production system to test that, so I think it makes sense to get this merged, and then when someone needs it in one of these systems, it could be improved by people with access to those systems.

easybuild/easyblocks/n/nvhpc.py
damianam: That lgtm. @boegel ?

easybuild/easyblocks/n/nvhpc.py
damianam: In the easyconfigs it works like this: `cuda_compute_capabilities = ['3.0', '3.2', '3.5', '3.7', '5.0', '5.2', '5.3', '6.0', '6.1', '7.0']`

I guess one could get rid of the `compute_capability` parameter here, and simply use `cuda_compute_capabilities` in the easyblock, converted as `','.join(''.join(x.split('.')) for x in cuda_compute_capabilities)` or something similar? I think keeping it consistent with other packages makes sense.

Could we give it a quick try @AndiH ?

easybuild/easyblocks/n/nvhpc.py
boegel: Sorting by "importance" may be subjective, so my preference goes to sorting alphabetically.

Note that in the output of `eb -a -e NVHPC` they'll be sorting alphabeticlaly anyway.

easybuild/easyblocks/n/nvhpc.py
boegel: Indeed, it would be confusing to have both `cuda_compute_capabilities` (which is a general easyconfig parameter since https://github.com/easybuilders/easybuild-framework/pull/3382) and `compute_capability`, especially if the value set for `cuda_compute_capabilities` is blatantly ignored.

This easyblock should also take into account the value of the `--cuda-compute-capabilities` EasyBuild configuration option, like we do elsewhere (see https://github.com/easybuilders/easybuild-easyblocks/blob/5c8c30c5d9f86cf4fa82b7d5befd7af90aff74d9/easybuild/easyblocks/t/tensorflow.py#L518-L523 for example)

easybuild/easyblocks/n/nvhpc.py
boegel: OK, makes sense to keep this hardcoded now and revisit it later.

easybuild/easyblocks/n/nvhpc.py
boegel: It probably doesn't, so remove?

easybuild/easyblocks/n/nvhpc.py
boegel: Yeah, I guess so, it won't be cleaner to do it otherwise, especially with the additional logic below, so let's leave it as is.

easybuild/easyblocks/n/nvhpc.py
```diff
@@ -0,0 +1,201 @@
...
+            dirs[key] = [os.path.join(self.nvhpc_install_subdir, 'compilers', d) for d in dirs[key]]
+
+        # $CPATH should not be defined in module for NVHPC, it causes problems
+        # cfr. https://github.com/easybuilders/easybuild-easyblocks/issues/830
+        if 'CPATH' in dirs:
+            self.log.info("Removing $CPATH entry: %s", dirs['CPATH'])
+            del dirs['CPATH']
+
+        # EasyBlock Options:
+        ## BYO Compilers allows for using NVHPC's libraries and tools with other, external compilers
```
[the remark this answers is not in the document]
boegel: @AndiH Just use a single `#`:

```python
# BYO Compilers allows for using NVHPC's libraries and tools with other, external compilers
```

(same below)

easybuild/easyblocks/n/nvhpc.py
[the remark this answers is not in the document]
boegel: We consistently use a single hash for comments, otherwise our code style checks will keep failing...

easybuild/easyblocks/n/nvhpc.py
boegel: How about this (favor readability for humans):

```python
nvhpc_env_vars = {
    'CPATH': 'include',
    'LD_LIBRARY_PATH': 'lib64',
    'PATH': 'bin',
}
self._nvhpc_extended_components(dirs, self.nvhpc_cuda_basedir, nvhpc_env_vars)

File: easybuild/easyblocks/n/nvhpc.py (hunk 2 above)
Comment (boegel) [reply easybuild/easyblocks/n/nvhpc.py]: Yes, sounds good!

File: easybuild/easyblocks/n/nvhpc.py (hunk 6 above)
Comment (boegel) [reply easybuild/easyblocks/n/nvhpc.py:65]: OK to leave as is for now (with `x86_64` hardcoded), can be tweaked later for other architectures.

File: easybuild/easyblocks/n/nvhpc.py (hunk 12 above)
Comment (boegel) [reply easybuild/easyblocks/n/nvhpc.py]: Sometimes it makes sense, but here we don't except anything else than a boolean value, do we?

File: easybuild/easyblocks/n/nvhpc.py (hunk 15 above)
Comment (AndiH) [reply easybuild/easyblocks/n/nvhpc.py]: That's in already! The snippet was outdated. Did I not resolve this? Sorry :(

File: easybuild/easyblocks/n/nvhpc.py (hunk 1 above)
Comment (AndiH) [reply easybuild/easyblocks/n/nvhpc.py:78]: Alright. I'm not super convinced (because using `cuda_compute_capabilities` suggests that something is compiled for this Compute Capability, which it is not – it is merely written as the default for any NVHPC-based compilations for the future), but I'll implement.

Comment (AndiH) [reply easybuild/easyblocks/n/nvhpc.py:78]: I poked (well `rg`-grepped around) and found this note:
```
# Currently, stdpar is limited to a single compute capability, so it needs its
# own special case for detection and default. Once we can support multiple
# compute capabilites we can remove all of the *STDPARCOMPUTECAP* variables and
# only use COMPUTECAP like we do for all other GPU codes.
```

I still would implement the block such that it is using `cuda_compute_capabilities` but would make a note (where?) that only single values are supported.

Comment (AndiH) [reply easybuild/easyblocks/n/nvhpc.py:78]: (I've also learned something new. `DEFSTDPARCOMPUTECAP` is indeed only used for the pSTL (as the name suggests) and if one were to set a general default for the compute capability, one should rather choose `COMPUTECAP` (which, probably, also goes to  `cudarc`).)

Comment (AndiH) [reply easybuild/easyblocks/n/nvhpc.py:78]: `--cuda-compute-capabilities` accepts only values with a dot, for example `7.0`. But that is not something understood by NVHPC and also not how you would specify it for any of the compilers in the CLI, for example `nvc -ta=tesla:cc70`.
I'll add the dot-removal now…
(But I'm getting less and less convinced that this is the right thing to do.)

Comment (AndiH) [reply easybuild/easyblocks/n/nvhpc.py]: Implemented in 2356db58

File: easybuild/easyblocks/n/nvhpc.py [hunk 18]
```diff
@@ -0,0 +1,269 @@
...
+class EB_NVHPC(PackedBinary):
+    """
+    Support for installing the NVIDIA HPC SDK (NVHPC) compilers
+    """
+
+    @staticmethod
+    def extra_options():
+        extra_vars = {
+            'default_cuda_version':      [None, "CUDA Version to be used as default (10.2 or 11.0 or ...)", CUSTOM],
+            'cuda_compute_capabilities': [None, "Compute Capability (7.0 or 8.0 or ...)", CUSTOM],
```
boegel, line 79: You can simply remove this line, `cuda_compute_capabilities` not a custom easyconfig parameter here, it's always supported.

easybuild/easyblocks/n/nvhpc.py
```diff
@@ -0,0 +1,269 @@
...
+                error_msg = "A default CUDA version is needed for installation of NVHPC. "
+                error_msg += "It can not be determined automatically and needs to be added manually. "
+                error_msg += "You can edit the easyconfig file, "
+                error_msg += "or use 'eb --try-amend=default_cuda_version=<version>'."
+                raise EasyBuildError(error_msg)
+
+        ec_default_compute_capability = self.cfg['cuda_compute_capabilities']
+        cfg_default_compute_capability = build_option('cuda_compute_capabilities')
+        if cfg_default_compute_capability is not None:
+            default_compute_capability = cfg_default_compute_capability[0].replace(".", "")
```
boegel, line 114: I think it's better no to do this silently, but check if there are multiple values in `cfg_default_compute_capability`, and if so print a clear warning (via `print_warning`) that only the first value will be used.

In practice it's likely that the `--cuda-compute-capabilities` will be set to a single value.

easybuild/easyblocks/n/nvhpc.py
```diff
@@ -0,0 +1,269 @@
...
+        elif ec_default_compute_capability is not None:
+            default_compute_capability = ec_default_compute_capability
```
boegel, line 116: Since the name suggests a list, you should also do the same check here as above (so only do it once, below).

It's probably wise here to check whether the value is a list of a string value, and act accordingly.

A string value doesn't usually make sense for `cuda_compute_capabilities`, but it does here, so also accepting `'7.0'` rather than forcing `['7.0']` makes sense I think.

easybuild/easyblocks/n/nvhpc.py
```diff
@@ -0,0 +1,269 @@
...
+        else:
+            error_msg = "A default Compute Capability is needed for installation of NVHPC."
+            error_msg += "Please provide it either in the easyconfig file like 'cuda_compute_capabilits=7.0',"
```
boegel, line 119: Typo: `capabilits` -> `capabilities`

easybuild/easyblocks/n/nvhpc.py
```diff
@@ -0,0 +1,269 @@
...
+            error_msg += "or use 'eb --cuda-compute-capabilities=7.0' from the command line."
+            raise EasyBuildError(error_msg)
+
+        nvhpc_env_vars = {
+            'NVHPC_INSTALL_DIR': self.installdir,
+            'NVHPC_SILENT': 'true',
+            'NVHPC_DEFAULT_CUDA': str(default_cuda_version),  # 10.2, 11.0
+            'NVHPC_STDPAR_CUDACC': str(default_compute_capability),  # 70, 80; single values, no list!
```
boegel, line 127: (in comment)

`values` -> `value`

easybuild/easyblocks/n/nvhpc.py
boegel: Still needs to be cleaned up?

easybuild/easyblocks/n/nvhpc.py
```diff
@@ -0,0 +1,269 @@
...
+
+    def sanity_check_step(self):
+        """Custom sanity check for NVHPC"""
+        prefix = self.nvhpc_install_subdir
+        custom_paths = {
+            'files': [os.path.join(prefix, 'compilers', 'bin', x) for x in ['nvc', 'nvc++', 'nvfortran', 'siterc']],
+            'dirs': [os.path.join(prefix, 'compilers', 'bin'), os.path.join(prefix, 'compilers', 'lib'),
+                     os.path.join(prefix, 'compilers', 'include'), os.path.join(prefix, 'compilers', 'man')]
+        }
+        super(EB_NVHPC, self).sanity_check_step(custom_paths=custom_paths)
```
boegel, line 168: Is it worth also adding a trivial sanity check command like `nvc --help` and/or `nvc --version` to make sure it's "functional"?

easybuild/easyblocks/n/nvhpc.py
```diff
@@ -0,0 +1,269 @@
...
+    def _nvhpc_extended_components(self, dirs, basepath, env_vars_dirs):
+        """
+        Extends `dirs` dict of key:environment_variables, value:list_of_directories with additional vars and dirs.
+        The dictionary key for a new env var will be created if it doesn't exist.
+        Also, the relative path specified in the `env_vars_dirs` dict is absolutized with the `basepath` prefix.
+        """
+        for env_var, folders in sorted(env_vars_dirs.items()):
+            if env_var not in dirs:
+                dirs[env_var] = []
+            if isinstance(folders, list) is False:
```
boegel, line 179:
```python
if not isinstance(folders, list):
```

boegel (changes requested):

easybuild/easyblocks/n/nvhpc.py
AndiH: Fixed in b07ff629c75cf1899b5ff3b64897eef6a46fc04f

easybuild/easyblocks/n/nvhpc.py
```diff
@@ -0,0 +1,282 @@
...
+            error_msg += "Please provide it either in the easyconfig file like 'cuda_compute_capabilities=7.0',"
+            error_msg += "or use 'eb --cuda-compute-capabilities=7.0' from the command line."
+            raise EasyBuildError(error_msg)
+
+        # Extract first element of default_compute_capability list, if it is a list
+        if isinstance(default_compute_capability, list):
+            _before_default_compute_capability = default_compute_capability
+            default_compute_capability = _before_default_compute_capability[0]
+            print_warning(
+                "Replaced list of compute capabilities {before} with first element of list {after}"\
```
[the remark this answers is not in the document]
AndiH: What? This is an error, isn't it, @boegel?

easybuild/easyblocks/n/nvhpc.py
AndiH: I've added some checks and print out. I hope I cover all the cases (commit dd49b649a040105cf012c2cbfc9ee085366578b8)

easybuild/easyblocks/n/nvhpc.py
AndiH: See dd49b649a040105cf012c2cbfc9ee085366578b8

easybuild/easyblocks/n/nvhpc.py
AndiH: It's gone in dd49b649a040105cf012c2cbfc9ee085366578b8

easybuild/easyblocks/n/nvhpc.py
AndiH: `eb -a -e NVHPC` doesn't seem to work for me:

```
❯ EB_PYTHON=python3 eb -a -e NVHPC --include-toolchains='/p/project/cjsc/herten1/Projects/EB-Installations/JUWELS/easybuild-repository/Custom_Toolchains/*.py','/p/project/cjsc/herten1/Projects/EB-Installations/JUWELS/easybuild-repository/Custom_Toolchains/compiler/*.py',$EASYBUILD_INCLUDE_TOOLCHAINS --include-easyblocks='/p/project/cjsc/herten1/Projects/EB-Installations/JUWELS/easybuild-repository/Custom_EasyBlocks/*.py',$EASYBUILD_INCLUDE_EASYBLOCKS
ERROR: Failed to parse configuration options: "Failed to obtain class for NVHPC easyblock (not available?): No module named 'easybuild.easyblocks.generic.nvhpc'"
```

Do I need to implement something else for that to work?

easybuild/easyblocks/n/nvhpc.py
AndiH: What the regex does, it replaces 
```diff
29c29
< #PATH=/usr/bin:/bin:$PATH
---
> PATH=/usr/bin:/bin:$PATH
```
So I'm inclined to think this is still an issue as this places `/usr/bin` on top of everything else in `$PATH`. I tried to have a look if I could see any strange behavior when leaving the line in, but I don't know at which point in the chain NVHPC actually needs something from GCC and if takes the wrong one in that case…

easybuild/easyblocks/n/nvhpc.py
AndiH: Do you have a sketch of this in another EasyBlock?

AndiH: I found a hint in the [TensorFlow EasyBlock](https://github.com/easybuilders/easybuild-easyblocks/blob/master/easybuild/easyblocks/t/tensorflow.py#L843-L847) and implemented
```
custom_commands = ["%s -v" % custom_paths["files"][0]]
```
in d47eedc6. Is that alright?

easybuild/easyblocks/n/nvhpc.py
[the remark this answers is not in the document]
boegel: It's just telling you you can strip off the trailing `\`:

```
>>> x = ( "foo"
... "bar")
>>> x
'foobar'
```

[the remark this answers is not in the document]
boegel: Ah, I overlooked the `.format`.

I would go for something like this:

```python
warning_msg = "Replaced list of compute capabilities " + _before_default_compute_capability
warning_msg += " with first element of list: " + default_compute_capability
print_warning(warning_msg)
```

easybuild/easyblocks/n/nvhpc.py
```diff
@@ -0,0 +1,283 @@
...
+                .format(
+                    before=_before_default_compute_capability,
+                    after=default_compute_capability
+                )
+            )
+        default_compute_capability.replace(".", "")
```
boegel, line 134: Assignment is missing here, this can't work? `.replace` returns a new string value...

Also, I hate to be pedantic, but you're now just assuming you have a string value in `default_compute_capability`, which may not be the case (and if it's not, the crash will be pretty ugly).

So I'd add a small check:

```python
if isinstance(default_compute_capability, str):
    default_compute_capability = default_compute_capability.replace('.', '')
else:
    raise EasyBuildError("Unexpected non-string value encountered for compute capability: %s",
                         default_compute_capability)

File: easybuild/easyblocks/n/nvhpc.py [hunk 27]
```diff
@@ -0,0 +1,283 @@
...
+
+    def sanity_check_step(self):
+        """Custom sanity check for NVHPC"""
+        prefix = self.nvhpc_install_subdir
+        custom_paths = {
+            'files': [os.path.join(prefix, 'compilers', 'bin', x) for x in ['nvc', 'nvc++', 'nvfortran', 'siterc']],
+            'dirs': [os.path.join(prefix, 'compilers', 'bin'), os.path.join(prefix, 'compilers', 'lib'),
+                     os.path.join(prefix, 'compilers', 'include'), os.path.join(prefix, 'compilers', 'man')]
+        }
+        custom_commands = ["%s -v" % custom_paths["files"][0]]
```
boegel, line 181: I would check all commands, and just use the command name (not the path):

```python
compiler_cmds = ['nvc', 'nvc++', 'nvfortran']
custom_paths = {
    'files': [os.path.join(prefix, 'compilers', 'bin', x) for x in compiler_cmds + ['siterc']],
...
custom_commands = ["%s -v" % x for x in compiler_cmds]

File: easybuild/easyblocks/n/nvhpc.py (hunk 3 above)
Comment (boegel) [reply easybuild/easyblocks/n/nvhpc.py]: You should use the class name: `eb -a -e EB_NVHPC`, my bad.

File: easybuild/easyblocks/n/nvhpc.py (hunk 7 above)
Comment (boegel) [reply easybuild/easyblocks/n/nvhpc.py:145]: OK, fine, let's leave the comment in place.

File: easybuild/easyblocks/n/nvhpc.py (hunk 23 above)
Comment (boegel) [reply easybuild/easyblocks/n/nvhpc.py]: Close, see comment on running `-v` for all the compiler commands, using only the command name.

**Comment by boegel** [comment]
@AndiH Please also revisit https://github.com/easybuilders/easybuild-easyconfigs/pull/11391, even though the tests can't pass there as long as this easyblock isn't merged, since that's the test case for this easyblock (we'll merge both in quick succession).

We're getting there, I'm confident we can get this merged really soon now...

**Comment by AndiH** [comment]
@boegel Absolutely! I wanted to get the EasyBlock done first before I implement the easyconfig. I am working on that next!

File: easybuild/easyblocks/n/nvhpc.py (hunk 3 above)
Comment (AndiH) [reply easybuild/easyblocks/n/nvhpc.py]: That does work. I sorted them in 1ac891ee.

File: easybuild/easyblocks/n/nvhpc.py (hunk 25 above)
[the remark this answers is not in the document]
Comment (AndiH) [reply easybuild/easyblocks/n/nvhpc.py:128]: `_before_default_compute_capability` is a list; that doesn't work like that. But it works with `.format`. I'll do a hybrid of our ideas, but using`format`.

File: easybuild/easyblocks/n/nvhpc.py (hunk 4 above)
Comment (AndiH) [reply easybuild/easyblocks/n/nvhpc.py:94]: I've added a note to the code in 7937f130.

File: easybuild/easyblocks/n/nvhpc.py (hunk 26 above)
Comment (AndiH) [reply easybuild/easyblocks/n/nvhpc.py]: Implemented in 2c917098cb8c49ecff5888d1b17872fae6623d81

File: easybuild/easyblocks/n/nvhpc.py (hunk 27 above)
Comment (AndiH) [reply easybuild/easyblocks/n/nvhpc.py]: Implemented in c54b719788144c4a959613e23880a8d62b8d59fa.

**Comment by AndiH** [comment]
I think I've tackled all the remaining open issues!
I leave resolution of the open comments to you and will now start with the edits for the easyconfig.

**Comment by boegel** [comment]
@AndiH Tested this using a tweaked version of https://github.com/easybuilders/easybuild-easyconfigs/pull/11391 (only changed the `CUDA` dep), ran into this crash:

```
== installing...
  File "/user/boegel/easybuild-framework/easybuild/main.py", line 115, in build_and_install_software
  File "/user/boegel/easybuild-framework/easybuild/framework/easyblock.py", line 3301, in build_and_install_one
  File "/user/boegel/easybuild-framework/easybuild/framework/easyblock.py", line 3204, in run_all_steps
  File "/user/boegel/easybuild/framework/easyblock.py", line 3059, in run_step
    step_method(self)()
  File "/tmp/eb-30tldro0/included-easyblocks-h6qfu1hb/easybuild/easyblocks/nvhpc.py", line 127, in install_step
    default_compute_capability = _before_default_compute_capability[0]
IndexError: list index out of range
```

**Comment by AndiH** [comment]
@boegel Did you specify a Compute Capability (via CLI or via easyconfig)?

**Comment by AndiH** [comment]
Ok, this is a bit weird, but fixed in e4c6c34d.

If `cuda_compute_capabilities` is not set in the easyconfig file and not via CLI, the value is not `None`, but rather `[]`; for whatever reason. The added `if ec_default_compute_capability` checks for empty lists.

**Comment by boegel** [comment]
@AndiH Getting `[]` as default value for `cuda_compute_capabilities` makes sense, see:

```
$ eb -a | grep cuda_compute_capabilities
cuda_compute_capabilities   List of CUDA compute capabilities to build with (if supported) [default: []]
```

Thanks for fixing that issue!

I'm running some final tests now, but I think this PR and https://github.com/easybuilders/easybuild-easyconfigs/pull/11391 are good to go (I'll deal with the failing tests in the latter once this PR is merged).

AndiH:
Great! 🕺

I try to work soon on NVHPC 20.9, which has been released meanwhile!

boegel (approved):

boegel:
lgtm, tested with https://github.com/easybuilders/easybuild-easyconfigs/pull/11391

Thanks a lot for the effort and patience @AndiH !

PR #2193, opened 2020-10-01
Add sanity check commands to clang and print warning for missing ncurses

Flamefire:
Loading the clang module and trying to run it makes it error with a missing ncurses lib which isn't currently detected.

I added sanity check commands and a warning to avoid this.

Question: Do we want this here and update the ECs or repeat the sanity check commands in every EC to avoid breaking existing ones?

### DIFF
--- a/easybuild/easyblocks/c/clang.py
+++ b/easybuild/easyblocks/c/clang.py
@@ -42,7 +42,7 @@ from distutils.version import LooseVersion
 from easybuild.easyblocks.generic.cmakemake import CMakeMake
 from easybuild.framework.easyconfig import CUSTOM
 from easybuild.tools import run
-from easybuild.tools.build_log import EasyBuildError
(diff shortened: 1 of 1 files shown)


migueldiascosta:
well, in systems where the existing ECs work because `ncurses` is install at the system level, this doesn't break them, so it will only break installations that wouldn't work anyway... I'd say option a...

Flamefire:
I updated many Clang ECs to include the ncurses dependency.

Note that ncurses is autodetected and as we have cmake as a build dependency for clang it gets detected and used. I guess the issue didn't show because the Clang module was mostly loaded with something else (e.g. CMake) which loads ncurses.

migueldiascosta (approved):
lgtm

migueldiascosta:
Going in, thanks @Flamefire!

PR #2192, opened 2020-10-01
Update DIRAC easyblock for new easyconfigs

Darkless012:
remove test_step in dirac easyblock, small additional fixes

failing tests are now patched

PR #2198, opened 2020-10-12
remove unused custom easyblock for DIRAC

hajgato:
None of the current `DIRAC` easyconfigs use the `dirac` EasyBlock. All of them use `CMakeMake`, so it can be deleted.


hajgato:
BTW, it was unnecessary to merge the previous PR https://github.com/easybuilders/easybuild-easyblocks/pull/2192 since the corresponding easyconfig do not use the easyblock.

boegel (approved):
Only used in archived DIRAC easyconfigs, so can indeed be removed...

PR #2208, opened 2020-10-21
Only enable CUDA in LAMMPS if it is an explicit dep + add support for building for NVIDIA Ampere GPUs

### LINKED ISSUE #2073: Use of deprecated CMake options for Kokkos in LAMMPS easyblock
Recent versions of LAMMPS (at least version `2Jun20`) enforces camel casing for the Kokkos options (e.g., `-DKokkos_ENABLE_CUDA=yes`). Hopefully this naming was already supported in previous versions so we don't have to introduce version-related conditionals (which is tedious with LAMMPS versioning)

ocaisa:
Also fixes #2073

### DIFF
--- a/easybuild/easyblocks/l/lammps.py
+++ b/easybuild/easyblocks/l/lammps.py
@@ -116,19 +116,22 @@ class EB_LAMMPS(CMakeMake):
         super(EB_LAMMPS, self).prepare_step(*args, **kwargs)
 
         # Unset LIBS when using both KOKKOS and CUDA - it will mix lib paths otherwise
-        if self.cfg['kokkos'] and get_software_root('CUDA'):
+        if 'cuda' in [dep['name'].lower() for dep in self.cfg.dependencies()] and get_software_root('CUDA'):
+            self.cuda = True
+        else:
+            self.cuda = False
+        if self.cfg['kokkos'] and self.cuda:
             env.unset_env_vars(['LIBS'])
 
     def configure_step(self, **kwargs):
         """Custom configuration procedure for LAMMPS."""
 
-        cuda = get_software_root('CUDA')
         # list of CUDA compute capabilities to use can be specifed in two ways (where (2) overrules (1)):
         # (1) in the easyconfig file, via the custom cuda_compute_capabilities;
         # (2) in the EasyBuild configuration, via --cuda-compute-capabilities configuration option;
         ec_cuda_cc = self.cfg['cuda_compute_capabilities']
         cfg_cuda_cc = build_option('cuda_compute_capabilities')
-        cuda_cc = check_cuda_compute_capabilities(cfg_cuda_cc, ec_cuda_cc)
+        cuda_cc = check_cuda_compute_capabilities(cfg_cuda_cc, ec_cuda_cc, cuda=self.cuda)
 
         # cmake has its own folder
         self.cfg['srcdir'] = os.path.join(self.start_dir, 'cmake')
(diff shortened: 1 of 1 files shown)


ocaisa:
Also taking the opportunity to add support for Ampere

bartoldeman (approved):
lgtm

easybuild/easyblocks/l/lammps.py
```diff
@@ -116,19 +117,22 @@ def prepare_step(self, *args, **kwargs):
         super(EB_LAMMPS, self).prepare_step(*args, **kwargs)
 
         # Unset LIBS when using both KOKKOS and CUDA - it will mix lib paths otherwise
-        if self.cfg['kokkos'] and get_software_root('CUDA'):
+        if 'cuda' in [dep['name'].lower() for dep in self.cfg.dependencies()] and get_software_root('CUDA'):
```
bartoldeman, line 120: on second thought, this reads a little funny since it's a boolean you can just do:
`self.cuda = 'cuda' in [dep['name'].lower() for dep in self.cfg.dependencies()] and get_software_root('CUDA')`
Also need to check if `self.cfg.dependencies()` include cuda if the toolchain includes cuda as well, e.g. fosscuda?

bartoldeman (changes requested):

easybuild/easyblocks/l/lammps.py
```diff
@@ -116,19 +117,19 @@ def prepare_step(self, *args, **kwargs):
         super(EB_LAMMPS, self).prepare_step(*args, **kwargs)
 
         # Unset LIBS when using both KOKKOS and CUDA - it will mix lib paths otherwise
-        if self.cfg['kokkos'] and get_software_root('CUDA'):
+        self.cuda = 'cuda' in [dep['name'].lower() for dep in self.cfg.dependencies()] and get_software_root('CUDA')
```
ocaisa, line 120: Still need to handle CUDA-aware toolchain

easybuild/easyblocks/l/lammps.py
```diff
@@ -96,6 +97,11 @@ class EB_LAMMPS(CMakeMake):
     Support for building and installing LAMMPS
     """
 
+    def __init__(self, *args, **kwargs):
+        super(EB_LAMMPS, self).__init__(*args, **kwargs)
+        self.cuda = 'cuda' in [dep['name'].lower() for dep in self.cfg.dependencies()] or \
+                    hasattr(self.toolchain, 'COMPILER_CUDA_FAMILY')
```
akesandgren, line 103: What's wrong with just using get_software_root('CUDA') ?

ocaisa: There is no way to turn off CUDA right now, in our case CUDA leaks in because it is a dep of UCX way down the stack, but a CUDA build will not run on a node without GPUs so we need a way be explicit about whether we want CUDA or not

ocaisa:
Test report in https://gist.github.com/ocaisa/fbf071eecd693b63e16075868358b31b

ocaisa:
Another test report for foss-cuda in https://gist.github.com/8a4872eeb39cc20ed3aa301fe37568cb

easybuild/easyblocks/l/lammps.py
```diff
@@ -216,20 +224,21 @@ def configure_step(self, **kwargs):
         if self.cfg['kokkos']:
 
             if self.toolchain.options.get('openmp', None):
-                self.cfg.update('configopts', '-DKOKKOS_ENABLE_OPENMP=yes')
+                self.cfg.update('configopts', '-DKokkos_ENABLE_OPENMP=yes')
```
ocaisa, line 227: Latest versions of LAMMPS enforce the `Kokkos` in the cmake defs but they are backwards compatible so safe to swap out

easybuild/easyblocks/l/lammps.py
boegel: Indeed, this approach is better: only if we're building with CUDA included as a dep or with a CUDA-enabled toolchain should CUDA support be enabled.

It probably doesn't hurt to create an easy method for this in the `Toolchain` class in framework?

boegel (approved):
lgtm, test reports with existing easyconfigs coming up...

boegel:
Test report by @boegel
 * **SUCCESS** _LAMMPS-3Mar2020-foss-2019b-Python-3.7.4-kokkos.eb_ 
 * **SUCCESS** _LAMMPS-3Mar2020-foss-2020a-Python-3.8.2-kokkos.eb_ 
 * **SUCCESS** _LAMMPS-3Mar2020-intel-2019b-Python-3.7.4-kokkos.eb_ 
 * **SUCCESS** _LAMMPS-7Aug2019-foss-2019b-Python-3.7.4-kokkos.eb_ 
 * **SUCCESS** _LAMMPS-7Aug2019-intel-2019b-Python-3.7.4-kokkos.eb_ 
 * **SUCCESS** _LAMMPS-7Aug2019-intel-2019b-Python-3.7.4-kokkos-OCTP.eb_ 

node3411.kirlia.os - Linux centos linux 7.8.2003, x86_64, Intel(R) Xeon(R) Gold 6240 CPU @ 2.60GHz (cascadelake), Python 2.7.5
See https://gist.github.com/90791dbc5f9123e1e50ea99350816bfb for a full test report.

PR #2211, opened 2020-10-28
Revert "Only enable CUDA in LAMMPS if it is an explicit dep + add support for building for NVIDIA Ampere GPUs"

ocaisa:
Reverts easybuilders/easybuild-easyblocks#2208 as some options have changed in recent LAMMPS and as a result there is no Kokkos in the final build, need to tidy this up and then resubmit

### DIFF
--- a/easybuild/easyblocks/l/lammps.py
+++ b/easybuild/easyblocks/l/lammps.py
@@ -85,7 +85,6 @@ KOKKOS_GPU_ARCH_TABLE = {
     '7.0': 'Volta70',  # NVIDIA Volta generation CC 7.0
     '7.2': 'Volta72',  # NVIDIA Volta generation CC 7.2
     '7.5': 'Turing75',  # NVIDIA Turing generation CC 7.5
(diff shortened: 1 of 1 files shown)


ocaisa:
@boegel @migueldiascosta I would prioritise this before the release as it is a major issue

boegel (approved):

ocaisa:
A number of things have changed in how to configure Kokkos with LAMMPS. Most importantly, you should up with CMake config options like
```
-DPKG_KOKKOS=on  -DKokkos_ENABLE_OPENMP=yes  -DKokkos_ARCH_SKX=yes  -DKokkos_ENABLE_CUDA=yes  -DKokkos_ARCH_AMPERE80=yes  -DCMAKE_CXX_COMPILER="/tmp/ocais1/LAMMPS/22Oct2020/gpsmkl-2020-Python-3.8.5-CUDA/lammps-patch_22Oct2020/lib/kokkos/bin/nvcc_wrapper"
```
Note the `-DPKG_KOKKOS` but `-DKokkos_ENABLE_OPENMP` and also how architectures are provided has changed.

PR #2212, opened 2020-10-28
fix two bugs in GROMACS easyblock when using GCC & MKL for FFT and BLAS/LAPACK

ostueker:
First change eliminates the (wrong) variable `$EBROOTMKL` from the easyblock as per discussion of @boegel and @mboisson on EasyBuild Slack.

Secondly GROMACS 2020.4' cmake was refusing to use MKL as FFT library (using the `gomkl-2020a` toolchain), which was fixed by appending `-lpthread -lm -ldl` to `-DMKL_LIBRARIES`.

### DIFF
--- a/easybuild/easyblocks/g/gromacs.py
+++ b/easybuild/easyblocks/g/gromacs.py
@@ -329,7 +329,7 @@ class EB_GROMACS(CMakeMake):
             if get_software_root('imkl'):
                 # using MKL for FFT, so it will also be used for BLAS/LAPACK
                 self.cfg.update('configopts', '-DGMX_FFT_LIBRARY=mkl -DMKL_INCLUDE_DIR="%s" ' %
-                    os.path.join(os.getenv('MKLROOT'), 'include'))
+                                os.path.join(os.getenv('MKLROOT'), 'include'))
                 libs = os.getenv('LAPACK_STATIC_LIBS').split(',')
                 mkl_libs = [os.path.join(os.getenv('LAPACK_LIB_DIR'), lib) for lib in libs if lib != 'libgfortran.a']
                 mkl_libs = ['-Wl,--start-group'] + mkl_libs + ['-Wl,--end-group -lpthread -lm -ldl']
--- a/easybuild/easyblocks/g/gromacs.py
+++ b/easybuild/easyblocks/g/gromacs.py
@@ -328,10 +328,11 @@ class EB_GROMACS(CMakeMake):
 
             if get_software_root('imkl'):
                 # using MKL for FFT, so it will also be used for BLAS/LAPACK
-                self.cfg.update('configopts', '-DGMX_FFT_LIBRARY=mkl -DMKL_INCLUDE_DIR="$EBROOTMKL/mkl/include" ')
+                self.cfg.update('configopts', '-DGMX_FFT_LIBRARY=mkl -DMKL_INCLUDE_DIR="%s" ' %
+                    os.path.join(os.getenv('MKLROOT'), 'include'))
                 libs = os.getenv('LAPACK_STATIC_LIBS').split(',')
                 mkl_libs = [os.path.join(os.getenv('LAPACK_LIB_DIR'), lib) for lib in libs if lib != 'libgfortran.a']
-                mkl_libs = ['-Wl,--start-group'] + mkl_libs + ['-Wl,--end-group']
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/g/gromacs.py
```diff
@@ -328,10 +328,11 @@ def configure_step(self):
 
             if get_software_root('imkl'):
                 # using MKL for FFT, so it will also be used for BLAS/LAPACK
-                self.cfg.update('configopts', '-DGMX_FFT_LIBRARY=mkl -DMKL_INCLUDE_DIR="$EBROOTMKL/mkl/include" ')
+                self.cfg.update('configopts', '-DGMX_FFT_LIBRARY=mkl -DMKL_INCLUDE_DIR="%s" ' %
+                                os.path.join(os.getenv('MKLROOT'), 'include'))
```
akesandgren, line 332: Although the above works and is correct,
I still think that the correct way to do this from an easyblock perspective would be to do
```
imkl_root = get_software_root('imkl')
  if imkl_root:
    self.cfg.update('configopts', '-DGMX_FFT_LIBRARY=mkl -DMKL_INCLUDE_DIR="%s" ' % os.path.join(imkl_root, 'mkl', 'include'))
```

boegel: Yeah, I agree it's probably better to avoid relying on `$MKLROOT`.
A bit cleaner:

```python
imkl_root = get_software_root('imkl')
if imkl_root:
    # using MKL for FFT, so it will also be used for BLAS/LAPACK
    imkl_include = os.path.join(imkl_root, 'mkl', 'include')
    self.cfg.update('configopts', '-DGMX_FFT_LIBRARY=mkl -DMKL_INCLUDE_DIR="%s" ' % imkl_include)

File: easybuild/easyblocks/g/gromacs.py [hunk 2]
```diff
@@ -328,10 +328,11 @@ def configure_step(self):
...
                 libs = os.getenv('LAPACK_STATIC_LIBS').split(',')
                 mkl_libs = [os.path.join(os.getenv('LAPACK_LIB_DIR'), lib) for lib in libs if lib != 'libgfortran.a']
-                mkl_libs = ['-Wl,--start-group'] + mkl_libs + ['-Wl,--end-group']
+                mkl_libs = ['-Wl,--start-group'] + mkl_libs + ['-Wl,--end-group -lpthread -lm -ldl']
```
boegel, line 335: Is it OK to always include `-lpthread -lm -ldl` here?

Can you clarify which error this fixes exactly?

ostueker: When trying to compile GROMACS 2020.4 with the `gomkl-2020a` (GCC + MKL) toolchain (which we found to produce an mdrun that runs a bit faster than the foss-toolchain), cmake wasn't finding the FFT provided by MKL unless I added the `-lpthread -lm -ldl` flags:

```CMake Error at cmake/gmxManageFFTLibraries.cmake:173 (message):
  Linking with MKL was requested, but was not successful: The include path to
  mkl.h in MKL_INCLUDE_DIR, and the link libraries in
  MKL_LIBRARIES=-Wl,--start-group;/cvmfs/soft.computecanada.ca/easybuild/software/2020/Core/imkl/2020.1.217/mkl/lib/intel64/libmkl_gf_lp64.a;/cvmfs/soft.computecanada.ca/easybuild/software/2020/Core/imkl/2020.1.217/mkl/lib/intel64/libmkl_sequential.a;/cvmfs/soft.computecanada.ca/easybuild/software/2020/Core/imkl/2020.1.217/mkl/lib/intel64/libmkl_core.a;-Wl,--end-group
  need to match what the MKL documentation says you need for your system: Use
  full paths to library files, in the right order, and separated by
  semicolons.
Call Stack (most recent call first):
  CMakeLists.txt:716 (include)
```

I used the [MKL advisor](https://software.intel.com/content/www/us/en/develop/articles/intel-mkl-link-line-advisor.html) to suggest the linker-flags, which is how I came to try adding `-lpthread -lm -ldl`, which solved the above error.

easybuild/easyblocks/g/gromacs.py
ostueker: OK, I'll change that.

akesandgren (approved):
LGTM

akesandgren:
Going in, thanks @ostueker!

PR #2218, opened 2020-10-31
correctly determine path to active binutils in TensorFlow easyblock

boegel:
Fix for failing installation when EasyBuild was configured with `binutils` in  `--filter-deps`, and taking into account the use of `--rpath`


boegel:
Test report by @boegel
 * **SUCCESS** _TensorFlow-2.3.1-foss-2020a-Python-3.8.2.eb_ 

node3412.kirlia.os - Linux centos linux 7.8.2003, x86_64, Intel(R) Xeon(R) Gold 6240 CPU @ 2.60GHz (cascadelake), Python 2.7.5
See https://gist.github.com/e63490e52666055fc35fa7e6caa6da1d for a full test report.

boegel:
Marking this as WIP, since this change it not enough to get the `TensorFlow` installation working in an environment with a custom `sysroot` + use of RPATH... The latter is probably not correctly taken into account yet in the TensorFlow easyblock, and may to resolve the problem being fixed here a bit differently.

easybuild/easyblocks/t/tensorflow.py
```diff
@@ -421,10 +421,12 @@ def setup_build_dirs(self):
...
         """Custom configuration procedure for TensorFlow."""
 
-        binutils_root = get_software_root('binutils')
-        if not binutils_root:
-            raise EasyBuildError("Failed to determine installation prefix for binutils")
-        self.binutils_bin_path = os.path.join(binutils_root, 'bin')
+        # determine location where binutils' ld command is installed
+        # note that this may be an RPATH wrapper script (when EasyBuild is configured with --rpath)
+        ld_path = which('ld')
+        self.log.info("Found ld at %s", ld_path)
```
Flamefire, line 427: IIRC `which` already logs this
```suggestion
```

easybuild/easyblocks/t/tensorflow.py
```diff
@@ -421,10 +421,12 @@ def setup_build_dirs(self):
...
+        self.binutils_bin_path = os.path.dirname(ld_path)
+        self.log.info("Assuming that all binutils tools are located at %s", self.binutils_bin_path)
```
Flamefire, line 429: Couldn't we check instead of assume?

Flamefire (changes requested):

easybuild/easyblocks/t/tensorflow.py
boegel: Actually this log message is misleading, because when `--rpath` is used actually only `ld` (a wrapper script) will be located there, but that's enough...

I'll just remove the log message.

boegel:
Test report by @boegel
 * **SUCCESS** _TensorFlow-2.3.1-foss-2020a-Python-3.8.2.eb_ 

node3575.doduo.os - Linux RHEL 8.2, x86_64, AMD EPYC 7552 48-Core Processor (zen2), Python 3.6.8
See https://gist.github.com/0e79e34dd8c46eb0c7e4f7393ec04e30 for a full test report.

boegel:
Test report by @boegel
 * **SUCCESS** _TensorFlow-2.3.1-foss-2020a-Python-3.8.2.eb_ 

node3105.skitty.os - Linux centos linux 7.8.2003, x86_64, Intel(R) Xeon(R) Gold 6140 CPU @ 2.30GHz, Python 3.6.8
See https://gist.github.com/11e6a91cae5c13fc7600ef7d665ddf5b for a full test report.

easybuild/easyblocks/t/tensorflow.py
```diff
@@ -421,10 +421,10 @@ def setup_build_dirs(self):
     def configure_step(self):
         """Custom configuration procedure for TensorFlow."""
 
-        binutils_root = get_software_root('binutils')
-        if not binutils_root:
-            raise EasyBuildError("Failed to determine installation prefix for binutils")
-        self.binutils_bin_path = os.path.join(binutils_root, 'bin')
+        # determine location where binutils' ld command is installed
+        # note that this may be an RPATH wrapper script (when EasyBuild is configured with --rpath)
```
ocaisa, line 425: This comment makes me nervous, what if they expect to find more than just `ld`?

boegel: We can consider symlinking the other binutils tools in that some location (or in a different path, with `ld` a symlink to the wrapper script).

There seems to be no need for this though, it works as expected this way...

Flamefire: Can't fully remember but isn't this path what gets passed to `-B` to the compiler? The tools are also in PATH so maybe it does indeed work or we don't noticed that it doesn't...

akesandgren:
@boegel conflict resolution needed.

terjekv:
What is the long-term status of this PR? I'm looking at what --from-pr-fixes we can drop when building in [NE]ESSI. :)

boegel:
@terjekv Yeah, we should really get this merged...

Need to check back on the discussion, and see if any changes should be made before doing a broad round of testing (both in and outside of EESSI build environment)

ocaisa (approved):
This is in use for a long time in EESSI so ok from me.

PR #2219, opened 2020-11-05
add custom easyblock for Metagenome-Atlas

Darkless012:


easybuild/easyblocks/m/metagenome_atlas_bundle.py
```diff
@@ -0,0 +1,330 @@
+##
+# Copyright 2016-2020 Ghent University
```
boegel, line 2: New easyblock, so should be `2020-2020`

easybuild/easyblocks/m/metagenome_atlas_bundle.py
```diff
@@ -0,0 +1,330 @@
...
+"""
+
+CONFIG_FILE_TEXT = r"""
+restart-times: 0
+cluster-config: "%s" #abs path
+cluster: "scheduler.py" #
+cluster-status: "slurm_status.py" #
+max-jobs-per-second: 10
+max-status-checks-per-second: 10
+cores: 99 # how many jobs you want to submit to your cluster queue
```
boegel, line 89: Why 99? Is this just a default value? If so, use `1` as default?

Also `cores` vs "`how many jobs`" doesn't match?

easybuild/easyblocks/m/metagenome_atlas_bundle.py
```diff
@@ -0,0 +1,330 @@
...
+rerun-incomplete: true  # recomended for cluster submissions
+keep-going: false
+"""
+
+KEY_MAPPING_TEXT = r"""
+# only parameters defined in key_mapping (see below) are passed to the command in the order specified.
+system: "slurm" #check if system is defined below
+
+slurm:
+  command: "sbatch --parsable"
```
boegel, line 100: Where does this come from, did you have to compose this yourself?

boegel (changes requested):
@Darkless012 There's a lot of stuff in here that should probably be proper `*.py` files that come along with the easyconfig file rather than the easyblock (so they can be tweaked easily for specific versions of `Metagenome-Atlas`)...

It should also be clear where those scripts come from

easybuild/easyblocks/m/metagenome_atlas_bundle.py
Darkless012: understand

easybuild/easyblocks/m/metagenome_atlas_bundle.py
Darkless012: The script is plain copy from the template. Will change that to 1.

easybuild/easyblocks/m/metagenome_atlas_bundle.py
Darkless012: This is also taken straight from the template.
https://github.com/metagenome-atlas/clusterprofile/blob/master/{{cookiecutter.profile_name}}/key_mapping.yaml

boegel (changes requested):
Please include some comments above the constants to explain where this stuff comes from

Seems like we should be pointing to https://github.com/metagenome-atlas/clusterprofile/tree/master/%7B%7Bcookiecutter.profile_name%7D%7D, perhaps also to some documentation?

boegel (approved):

PR #2224, opened 2020-11-09
enhance extension filter for Python packages to ignore user site dir when for checking Python package extensions

Flamefire:
The `EXTS_FILTER_PYTHON_PACKAGES` is used as the default for checking whether to skip installing an extension. Now it excludes the user site dir and hence does not wrongly skip installing a package when it is present in the user folder.


akesandgren:
And this does not negatively affect users doing their own installations with EB I hope?

Flamefire:
No. It just aligns the check with the install (and pretty much any other) python command used here

akesandgren (approved):
LGTM

akesandgren:
Going in, thanks @Flamefire!

PR #2226, opened 2020-11-09
add support to Boost easyblock for providing extra C++ compiler flags

boegel:


easybuild/easyblocks/b/boost.py
```diff
@@ -76,9 +76,10 @@ def extra_options():
         extra_vars = {
             'boost_mpi': [False, "Build mpi boost module", CUSTOM],
             'boost_multi_thread': [False, "Build boost with multi-thread option", CUSTOM],
-            'toolset': [None, "Toolset to use for Boost configuration ('--with-toolset for bootstrap.sh')", CUSTOM],
+            'extra_cxxflags': [None, "Extra C++ compiler flags to use", CUSTOM],
```
bartoldeman, line 79: I wonder if an option like this would be better done as a toolchainopt in the framework?
Then EB just sets CXXFLAGS with the extra flags and it will be picked up with the existing logic in boost.py.

bartoldeman (changes requested):

easybuild/easyblocks/b/boost.py
boegel: Yeah, that's actually a good point...

Should probably also be an EasyBuild configuration option like `--cxxflags` (but perhaps as an override rather than to specify extra flags, so that would be a separate PR/feature).

bartoldeman: https://github.com/easybuilders/easybuild-framework/pull/2193
is a start...

boegel: It's actually exactly what's needed, thanks for reviving that PR!

boegel:
Superseded by better solution in https://github.com/easybuilders/easybuild-framework/pull/2193

PR #2229, opened 2020-11-11
Add support for building Clang with OpenMP offload support

nordmoen:
In the current setup, code with OpenMP will compile, but will encounter
a strange error at runtime. This update adds support for defining the
OpenMP CUDA capabilities which remedies the runtime error. This change
builds on the existing CUDA support already defined in the build
process.

### DIFF
--- a/easybuild/easyblocks/c/clang.py
+++ b/easybuild/easyblocks/c/clang.py
@@ -293,11 +293,11 @@ class EB_Clang(CMakeMake):
             cfg_cuda_cc = build_option('cuda_compute_capabilities')
             cuda_cc = cfg_cuda_cc or ec_cuda_cc or []
             if not cuda_cc:
-                raise EasyBuildError("Can't build Clang with CUDA support "\
+                raise EasyBuildError("Can't build Clang with CUDA support "
                                      "without specifying 'cuda-compute-capabilities'")
             default_cc = self.cfg['default_cuda_capability'] or min(cuda_cc)
             if not self.cfg['default_cuda_capability']:
-                print_warning("No default CUDA capability defined! "\
+                print_warning("No default CUDA capability defined! "
                               "Using '%s' taken as minimum from 'cuda_compute_capabilities'" % default_cc)
             cuda_cc = [cc.replace('.', '') for cc in cuda_cc]
             default_cc = default_cc.replace('.', '')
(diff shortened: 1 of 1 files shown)


bartoldeman (approved):
LGTM, thanks!

easybuild/easyblocks/c/clang.py
```diff
@@ -85,6 +85,8 @@ def extra_options():
             'skip_all_tests': [False, "Skip running of tests", CUSTOM],
             # The sanitizer tests often fail on HPC systems due to the 'weird' environment.
             'skip_sanitizer_tests': [True, "Do not run the sanitizer tests", CUSTOM],
+            'default_cuda_capability': [None, "Default CUDA capability specified for clang, e.g. '7.5'", CUSTOM],
+            'cuda_compute_capabilities': [[], "List of CUDA compute capabilities to build with", CUSTOM],
```
boegel, line 89: This is already a general easyconfig parameter (since https://github.com/easybuilders/easybuild-framework/pull/3382), so adding `cuda_compute_capabilities` here too shouldn't be needed...

@nordmoen Do you have a `Clang` easyconfig file that requires the enhancement you implemented here?

nordmoen: @boegel The default parameter is quite nice if a user has specified more than one compute compatibility and doesn't want to use the minimum.

I added the `cuda_compute_capabilities` based on [feedback from @Flamefire on Slack](https://easybuild.slack.com/archives/C34UA1HT7/p1605097924389300?thread_ts=1605082171.371900&cid=C34UA1HT7). In this way it resembles other similar `easyconfig` files like `TensorFlow`. I have used the built-in `--cuda-compute-capabilites` flag in EB and for me it should not be needed.

boegel: There's two different things here:

1) The `cuda_compute_capabilities` easyconfig parameter, which is a general easyconfig parameter (so you don't need to define it in `extra_options` for it to be defined & picked up), see "`eb -a | grep cuda`".

2) The `--cuda-compute-capabilities` EasyBuild configuration option, which is what you're picking up via `build_option('cuda_compute_capabilities')` (I know it can be confusing...).

I was merely pointing out that add `cuda_compute_capabilities` as a knwon easyconfig parameter is not needed at all, since it's already a known general easyconfig parameter. There's no real downside though (except for adding to the confusion, maybe).

PR #2233, opened 2020-11-12
fix hardcoded path in NVHPC easyblock to support multiple architectures

nordmoen:
The `nvhpc.py` file contained one hardcoded path which hindered it from
working on multiple architectures.

### DIFF
--- a/easybuild/easyblocks/n/nvhpc.py
+++ b/easybuild/easyblocks/n/nvhpc.py
@@ -91,6 +91,10 @@ class EB_NVHPC(PackedBinary):
         """Easyblock constructor, define custom class variables specific to NVHPC."""
         super(EB_NVHPC, self).__init__(*args, **kwargs)
 
+        # Ideally we should be using something like `easybuild.tools.systemtools.get_cpu_architecture` here, however,
+        # on `ppc64le` systems this function returns `POWER` instead of `ppc64le`. Since this path needs to reflect
+        # `arch` (https://easybuild.readthedocs.io/en/latest/version-specific/easyconfig_templates.html) the same
+        # procedure from `templates.py` was reused here:
         architecture = 'Linux_%s' % platform.uname()[4]
         self.nvhpc_install_subdir = os.path.join(architecture, self.version)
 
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/n/nvhpc.py
```diff
@@ -90,8 +91,8 @@ def __init__(self, *args, **kwargs):
         """Easyblock constructor, define custom class variables specific to NVHPC."""
         super(EB_NVHPC, self).__init__(*args, **kwargs)
 
-        # Potential improvement: get "Linux_x86_64" from easybuild.tools.systemtools' get_cpu_architecture()
-        self.nvhpc_install_subdir = os.path.join('Linux_x86_64', self.version)
+        architecture = 'Linux_%s' % platform.uname()[4]
```
Micket, line 94: Can't we do what the comment suggests instead?

```suggestion
        architecture = 'Linux_%s' % easybuild.tools.systemtools.get_cpu_architecture()
```
+ change of import statement

Micket (changes requested):

easybuild/easyblocks/n/nvhpc.py
nordmoen: Should have commented instead of running with it. I tested with `get_cpu_architecture()` and on our `ppc64le` system it simply returned `POWER`. Since I needed something to reflect [`arch`](https://easybuild.readthedocs.io/en/latest/version-specific/easyconfig_templates.html) I simply copied its definition.

I'll add a comment to this effect so that it is clearer why I did this.

nordmoen:
@Micket Any thoughts on this?

Micket (approved):
lgtm

boegel:
Test report by @boegel
 * **SUCCESS** _NVHPC-20.7.eb_ 

node2610.swalot.os - Linux centos linux 7.9.2009, x86_64, Intel(R) Xeon(R) CPU E5-2660 v3 @ 2.60GHz (haswell), Python 3.6.8
See https://gist.github.com/3de51b45983cbda33b46d7a8c2b947f0 for a full test report.

PR #2234, opened 2020-11-13
added support for preinstallopts and installopts to the ncl easyblock

mboisson:
This is needed for us because I need to pass CPATH for our installation of freetype in Gentoo.

### DIFF
--- a/easybuild/easyblocks/n/ncl.py
+++ b/easybuild/easyblocks/n/ncl.py
@@ -224,7 +224,7 @@ class EB_NCL(EasyBlock):
     def install_step(self):
         """Build in install dir using build_step."""
 
-        cmd = "make Everything"
+        cmd = "%s make Everything %s" % (self.cfg['preinstallopts'], self.cfg['installopts'])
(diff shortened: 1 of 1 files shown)


migueldiascosta:
although using (pre)installopts here makes sense since in this easyblock building is done in the install_step, I wonder if from the user/easyconfig point of view it wouldn't make more sense to use (pre)buildopts...?

mboisson:
Possibly. On one side, you would expect it to be in the build step. On the other side, if you put it in buildopts and you see it used in the install step when you look at the logs, it's just as puzzling.

migueldiascosta:
True, but the logs are already puzzling in that sense even without these opts, no? i.e., there should already be a log message about building being done in the install_step?

not a big deal though, so let me know if you prefer to leave it as it is

mboisson:
I would leave it as is.

migueldiascosta (approved):
lgtm

migueldiascosta:
Going in, thanks @mboisson!

PR #2238, opened 2020-11-17
Evaluate numpy test results

### LINKED ISSUE #605: also run numpy.test rather than only a simply numpy.dot performance test in EB_numpy.test_step

### LINKED ISSUE #2237: Numpy (and likely scipy) test failures ignored
The numpy EasyBlock runs the unittests as `"cd .. && %(python)s -c 'import numpy; numpy.test(verbose=2)'"`

However the `numpy.test` function returns True/False depending on success. As that is never checked the exit code is always 0 (unless a segfault or similar occurred). See https://github.com/numpy/numpy/blob/360ba0572483457837992d711a0a00580741fc88/numpy/_pytesttester.py#L213

This means we basically run the tests but ignore any failures other than hard crashes. And with 2019b numpy I see errors in the test suite (caused by having LDFLAGS set)
I don't think this is intentional and hence should be changed.

I assume scipy has the same problem as it is similar.

Flamefire:
This checks for any errors during the numpy test suite.
Supersedes #1744

Fixes #2237
Fixes #605

### DIFF
--- a/easybuild/easyblocks/n/numpy.py
+++ b/easybuild/easyblocks/n/numpy.py
@@ -66,7 +66,17 @@ class EB_numpy(FortranPythonPackage):
         self.sitecfg = None
         self.sitecfgfn = 'site.cfg'
         self.testinstall = True
-        self.testcmd = "cd .. && %(python)s -c 'import numpy; numpy.test(verbose=2)'"
+        if LooseVersion(self.version) >= LooseVersion('1.15'):
+            # Numpy 1.15+ returns a True on success. Hence invert to get a failure value
+            test_code = 'sys.exit(not numpy.test(verbose=2))'
+        else:
+            # Return value is a TextTestResult. Check the errors member for any error
+            test_code = 'sys.exit(len(numpy.test(verbose=2).errors) > 0)'
+        # Prepend imports
+        test_code = "import sys; import numpy; " + test_code
+        # LDFLAGS should not be set when testing numpy/scipy, because it overwrites whatever numpy/scipy sets
+        # see http://projects.scipy.org/numpy/ticket/182
+        self.testcmd = "unset LDFLAGS && cd .. && %%(python)s -c '%s'" % test_code
 
     def configure_step(self):
         """Configure numpy build by composing site.cfg contents."""


ocaisa:
@Flamefire Can you update the scipy easyblock as well for this change?

Flamefire:
Done. And with SciPy-bundle-2019.03-foss-2019a.eb I get a failure due to a precision issue -.- Same with

Those are very small failures, e.g. `-0.00077271310043404 !=          -0.000772713100434114  (rdiff          9.569216326469691e-14)` or `"Arrays are not almost equal to 5 decimals" ... Max absolute difference: 1.5497208e-05`

Numpy seems to be fine. Not sure what to do. I'd suggest I'll remove the scipy change so we have at least numpy tested

ocaisa:
Let's put it in a separate PR so we can still try to address it

ocaisa:
Was that test on POWER? Looks like this is not uncommon:
https://github.com/scipy/scipy/pull/11181

Flamefire:
No, this was on Intel and AMD x86 CPUs. Split out the SciPy stuff into #2241

Flamefire:
Test report by @Flamefire
 * **SUCCESS** _SciPy-bundle-2019.03-foss-2019a.eb_ 
 * **SUCCESS** _SciPy-bundle-2019.10-foss-2019b-Python-2.7.16.eb_ 
 * **SUCCESS** _SciPy-bundle-2019.10-foss-2019b-Python-3.7.4.eb_ 
 * **SUCCESS** _Python-3.6.6-foss-2018b.eb_ 
 * **SUCCESS** _Python-3.6.4-foss-2018a.eb_ 

taurusi6257.taurus.hrsk.tu-dresden.de - Linux RHEL 7.8, x86_64, Intel(R) Xeon(R) CPU E5-2680 v3 @ 2.50GHz, Python 2.7.5
See https://gist.github.com/0a047b845b27d654b8e0bbe381e00874 for a full test report.

migueldiascosta:
lgtm

I' wondering if there should be a log message about inverting the return value, so if a future version of numpy changes the behaviour (again) and the test_steps fails although all tests pass, at least it would be immediately clear why (?)

Flamefire:
I expect the shown error message to be clear enough for that case. I would not clutter the log in anticipation of stuff that may or may not change and may or may not be obvious what changed. I mean: The test command is pretty clear: Run a test function and exit with the inverse of the result. As long as you remember that exit takes an error value and not a success value of course.

migueldiascosta (approved):
lgtm

migueldiascosta:
Going in, thanks @Flamefire!

PR #2244, opened 2020-11-20
Also add bin subdir to PATH when installing a python package

Flamefire:
Some packages install scripts, and then show a warning like:

>   WARNING: The script markdown_py is installed in '/tmp/install/software/TensorFlow/2.4.0-fosscuda-2019b-Python-3.7.4/bin' which is not on PATH.
>   Consider adding this directory to PATH or, if you prefer to suppress this warning, use --no-warn-script-location.

This can be problematic for PythonBundle when a later package requires a binary installed by a prior package
which would then not be found.


akesandgren (approved):
LGTM

akesandgren:
Going in, thanks @Flamefire!

PR #2246, opened 2020-11-21
prepend all hardcoded /usr/* paths with sysroot in Python's setup.py installation script

boegel:
(created using `eb --new-pr`)
follow-up for #2148 and #2149, apparently that fix was incomplete

see also https://github.com/EESSI/software-layer/issues/41

PR #2248, opened 2020-11-22
patch CMake's UnixPaths.cmake script if --sysroot is set

boegel:
Alternative to https://github.com/easybuilders/easybuild-easyblocks/pull/2246, which isn't working out well...

@bartoldeman Is this more or less what you had in mind?

### DIFF
--- a/easybuild/easyblocks/c/cmake.py
+++ b/easybuild/easyblocks/c/cmake.py
@@ -29,7 +29,9 @@ import os
 
 from easybuild.easyblocks.generic.configuremake import ConfigureMake
 from easybuild.framework.easyconfig import CUSTOM
+from easybuild.tools.build_log import EasyBuildError
 from easybuild.tools.config import build_option
+from easybuild.tools.filetools import apply_regex_substitutions
 from easybuild.tools.modules import get_software_root, get_software_libdir
 import easybuild.tools.environment as env
 
@@ -46,6 +48,29 @@ class EB_CMake(ConfigureMake):
             'use_openssl': [True, "Use openSSL (if building CURL in CMake)", CUSTOM],
         })
         return extra_vars
+
+    def patch_step(self):
+        """
+        Apply patch files, and do runtime patching (if needed).
+
+        Tweak UnixPaths.cmake if EasyBuild is configured with --sysroot.
+        """
+        super(EB_CMake, self).patch_step()
+
+        sysroot = build_option('sysroot')
+        if sysroot:
+            # prepend custom sysroot to all hardcoded paths like /lib and /usr
+            # see also patch applied by Gentoo:
+            # https://gitweb.gentoo.org/repo/gentoo.git/tree/dev-util/cmake/files/cmake-3.14.0_rc3-prefix-dirs.patch
+            srcdir = os.path.join(self.builddir, 'cmake-%s' % self.version)
+            unixpaths_cmake = os.path.join(srcdir, 'Modules', 'Platform', 'UnixPaths.cmake')
+            if os.path.exists(unixpaths_cmake):
+                regex_subs = [
+                    (r' /([a-z])', r' %s/\1' % sysroot),
+                ]
+                apply_regex_substitutions(unixpaths_cmake, regex_subs)
+            else:
+                raise EasyBuildError("File to patch %s not found", unixpaths_cmake)
 
     def configure_step(self):
         """
@@ -95,6 +120,7 @@ class EB_CMake(ConfigureMake):
             self.log.info("Found sysroot '%s', adding it to $CMAKE_PREFIX_PATH and $CMAKE_LIBRARY_PATH", sysroot)
             cmake_prefix_path.append(sysroot)
             cmake_library_path.append(os.path.join(sysroot, 'usr', 'lib'))
+            cmake_library_path.append(os.path.join(sysroot, 'usr', 'lib64'))
             cmake_include_path.append(os.path.join(sysroot, 'usr', 'include'))
 
         cmake_path_env_vars = {


easybuild/easyblocks/c/cmake.py
```diff
@@ -95,6 +130,7 @@ def configure_step(self):
             self.log.info("Found sysroot '%s', adding it to $CMAKE_PREFIX_PATH and $CMAKE_LIBRARY_PATH", sysroot)
             cmake_prefix_path.append(sysroot)
             cmake_library_path.append(os.path.join(sysroot, 'usr', 'lib'))
+            cmake_library_path.append(os.path.join(sysroot, 'usr', 'lib64'))
```
bartoldeman, line 133: I wonder if this chunk is still needed at all: if CMake is patched itself do you still need those environment variables set?

easybuild/easyblocks/c/cmake.py
```diff
@@ -47,6 +49,39 @@ def extra_options():
...
+
+        Tweak UnixPaths.cmake if EasyBuild is configured with --sysroot.
+        """
+        super(EB_CMake, self).patch_step()
+
+        sysroot = build_option('sysroot')
+        if sysroot:
+            # prepend custom sysroot to all hardcoded paths like /lib and /usr
+            # see also patch applied by Gentoo:
+            # https://gitweb.gentoo.org/repo/gentoo.git/tree/dev-util/cmake/files/cmake-3.14.0_rc3-prefix-dirs.patch
```
bartoldeman, line 64: There are many small differences versus this patch. Those may not matter though but we should carefully check.

bartoldeman (changes requested):
I'll try out this block tomorrow morning and see if it's working ok.

easybuild/easyblocks/c/cmake.py
bartoldeman: I checked and all the small differences look inconsequential. Here '-' is the Gentoo-patched cmake and '+' the easyblock patch:
```
 list(APPEND CMAKE_SYSTEM_PREFIX_PATH
   # Standard
-  /cvmfs/soft.computecanada.ca/gentoo/2020/usr/local /cvmfs/soft.computecanada.ca/gentoo/2020/usr /cvmfs/soft.computecanada.ca/gentoo/2020/
+  /cvmfs/soft.computecanada.ca/gentoo/2020/usr/local /cvmfs/soft.computecanada.ca/gentoo/2020/usr /cvmfs/soft.computecanada.ca/gentoo/2020
```
this is just a slash

```
 # Non "standard" but common install prefixes
 list(APPEND CMAKE_SYSTEM_PREFIX_PATH
-  /usr/X11R6
-  /usr/pkg
-  /opt
+  /cvmfs/soft.computecanada.ca/gentoo/2020/usr/X11R6
+  /cvmfs/soft.computecanada.ca/gentoo/2020/usr/pkg
+  /cvmfs/soft.computecanada.ca/gentoo/2020/opt
```
these don't exist in the prefix, and most likely `/opt/bin`, `/opt/include` aren't used either so this looks harmless.

```
-list(APPEND CMAKE_SYSTEM_LIBRARY_PATH
-  /cvmfs/soft.computecanada.ca/gentoo/2020/usr/x86_64-pc-linux-gnu/lib//gcc
-  /cvmfs/soft.computecanada.ca/gentoo/2020/usr/x86_64-pc-linux-gnu/lib/
-  /cvmfs/soft.computecanada.ca/gentoo/2020/usr/lib64
-  /cvmfs/soft.computecanada.ca/gentoo/2020/usr/libx32
-  /cvmfs/soft.computecanada.ca/gentoo/2020/usr/lib32
-  /cvmfs/soft.computecanada.ca/gentoo/2020/usr/lib
-  /cvmfs/soft.computecanada.ca/gentoo/2020/lib
-  )
-
-list(APPEND CMAKE_SYSTEM_PROGRAM_PATH
-  /cvmfs/soft.computecanada.ca/gentoo/2020/usr/bin
-  /cvmfs/soft.computecanada.ca/gentoo/2020/bin
+list(APPEND CMAKE_SYSTEM_LIBRARY_PATH /cvmfs/soft.computecanada.ca/gentoo/2020/usr/lib64 /cvmfs/soft.computecanada.ca/gentoo/2020/usr/lib /cvmfs/soft.computecanada.ca/gentoo/2020/lib64 /cvmfs/soft.computecanada.ca/gentoo/2020/lib
+  # X11
+ 
   )
```
The GCC paths are Gentoo GCC so not relevant to EB GCC.
The X11 path is a historical artifact, it's no longer relevant.
Appending to CMAKE_SYSTEM_PROGRAM_PATH on the Gentoo side looks silly since this defaults to CMAKE_SYSTEM_PREFIX_PATH entries suffixed with /bin already.

```
 list(APPEND CMAKE_PLATFORM_IMPLICIT_LINK_DIRECTORIES
-  /cvmfs/soft.computecanada.ca/gentoo/2020/usr/x86_64-pc-linux-gnu/lib//gcc
-  /cvmfs/soft.computecanada.ca/gentoo/2020/usr/x86_64-pc-linux-gnu/lib/
-  /cvmfs/soft.computecanada.ca/gentoo/2020/usr/lib64
-  /cvmfs/soft.computecanada.ca/gentoo/2020/usr/libx32
-  /cvmfs/soft.computecanada.ca/gentoo/2020/usr/lib32
-  /cvmfs/soft.computecanada.ca/gentoo/2020/usr/lib
-  /cvmfs/soft.computecanada.ca/gentoo/2020/lib
+  /cvmfs/soft.computecanada.ca/gentoo/2020/lib /cvmfs/soft.computecanada.ca/gentoo/2020/lib32 /cvmfs/soft.computecanada.ca/gentoo/2020/lib64 /cvmfs/soft.computecanada.ca/gentoo/2020/usr/lib /cvmfs/soft.computecanada.ca/gentoo/2020/usr/lib32 /cvmfs/soft.computecanada.ca/gentoo/2020/usr/lib64
   )
```
same story as above for GCC


```
@@ -92,11 +77,11 @@
 # parsing the implicit directory information from compiler output.
 set(_CMAKE_C_IMPLICIT_INCLUDE_DIRECTORIES_INIT
   ${CMAKE_C_IMPLICIT_INCLUDE_DIRECTORIES}
-  /cvmfs/soft.computecanada.ca/gentoo/2020/usr/include
+  "${_cmake_sysroot_compile}/usr/include"
   )
 set(_CMAKE_CXX_IMPLICIT_INCLUDE_DIRECTORIES_INIT
   ${CMAKE_CXX_IMPLICIT_INCLUDE_DIRECTORIES}
-  /cvmfs/soft.computecanada.ca/gentoo/2020/usr/include
+  "${_cmake_sysroot_compile}/usr/include"
   )
 set(_CMAKE_CUDA_IMPLICIT_INCLUDE_DIRECTORIES_INIT
   ${CMAKE_CUDA_IMPLICIT_INCLUDE_DIRECTORIES}
```
not sure why Gentoo patches this. The implicit include directories are ok when they are initialized from the compiler output at CMake-compile-time, so we should not worry about this bit.

easybuild/easyblocks/c/cmake.py
bartoldeman: So I believe this chunk should be removed since these directories are already searched via the compiled-in CMAKE_SYSTEM_PREFIX_PATH.

Flamefire: They should actually never be required. CMake already reroots paths when the SYSPATH or FIND_ROOT_PATH is passed

Flamefire:
I don't think this is a good idea. Using sysroot from a toolchain file sounds much much better.
Counter example: Some software does` find_library(FOO foo.so PATHS /usr/foo)`
This won't be caught be the patch approach but by the sysroot approach
Better without a toolchain file (if --sysroot is not really required):
CMAKE_FIND_ROOT_PATH with CMAKE_FIND_ROOT_PATH_MODE_LIBRARY and the other 2

bartoldeman:
I may have misunderstood your point here @Flamefire but the question is here if `find_library(FOO foo.so PATHS /usr/foo)` should look in `$EPREFIX/usr/foo` or in `/usr/foo`.

In the EESSI (or CC CVMFS) Gentoo Prefix context the paths are absolute and not auto-prepended with any prefix, i.e. the output of `pkg-config` and friends *includes* the prefix, and if you then prepend the prefix to that you get a double prefix.

Flamefire:
@bartoldeman IIUC the whole point of sysroot is that you use `$EPREFIX` and **not** `/usr/foo`, is this correct?

The CMake `find_*` commands have logic which will prepend the `CMAKE_FIND_ROOT_PATH` (or `CMAKE_SYSROOT`) to ALL paths passed to the `find_*` command (explicit and implicit via e.g. `CMAKE_PREFIX_PATH`) And via some variables you can control if the non-prefixed paths should be searched too. This is useful so that  `find_library(FOO foo.so PATHS /usr/foo)` searches `$EPREFIX/usr/foo`.

I can see that failing though as e.g. modules set `CMAKE_PREFIX_PATH` and you don't want to search prepended paths here but you want (only) them for the paths passed to `PATHS` (above) and for `CMAKE_SYSTEM_PREFIX_PATH` (autogenerated)

This is tough... I guess then this is the right approach to get it right most of the time. Make sure to test this thoroughly with `--debug-find`

bartoldeman:
I think we are conflating two uses of sysroot here, one as used in cross-compilation strategies and one used as a prefix in a Gentoo Prefix installation.

I'll do some more research and get back on this.

boegel:
> I think we are conflating two uses of sysroot here, one as used in cross-compilation strategies and one used as a prefix in a Gentoo Prefix installation.
> 
> I'll do some more research and get back on this.

How's your research going on this? ;-)

boegel:
@bartoldeman I would like to get back to this, and get it merged ASAP.

Maybe we should set up a quick call to discuss what needs to change here?

We've been using this easyblock in EESSI for a while now, and haven't seen any problems with CMake itself, or stuff that involves CMake in the installation procedure...

bartoldeman (changes requested):
I think rereading my comments it's ok, but I would still remove the whole chunk from line 129 to line 136 since it looks superfluous.
```
sysroot = ...
if sysroot:
    ...
```

boegel:
Test report by @boegel
 * **SUCCESS** _CMake-3.20.1-GCCcore-10.3.0.eb_ 
 * **SUCCESS** _CMake-3.22.1-GCCcore-11.2.0.eb_ 

node3129.skitty.os - Linux Debian GNU/Linux 11 (bullseye), x86_64, Intel(R) Xeon(R) Gold 6140 CPU @ 2.30GHz (skylake_avx512), Python 3.11.4
See https://gist.github.com/boegel/4af87dbe3320d0100362d626e95f6d65 for a full test report.

bartoldeman (approved):
LGTM

PR #2254, opened 2020-11-27
filter out user packages in LAMMPS easyblock if corresponding dependency isn't included + only set -DUSER-INTEL on x86_64 systems

edmondac:


easybuild/easyblocks/l/lammps.py
```diff
@@ -121,6 +121,9 @@ def prepare_step(self, *args, **kwargs):
 
     def configure_step(self, **kwargs):
         """Custom configuration procedure for LAMMPS."""
+        vtk = get_software_root('VTK')
+        if not vtk:
```
akesandgren, line 125: If you don't use "vtk" any other place there is no reason to create a variable for the result just do
if not get_software_root('VTK')

akesandgren (changes requested):

easybuild/easyblocks/l/lammps.py
edmondac: yup

easybuild/easyblocks/l/lammps.py
```diff
@@ -121,6 +121,8 @@ def prepare_step(self, *args, **kwargs):
 
     def configure_step(self, **kwargs):
         """Custom configuration procedure for LAMMPS."""
+        if not get_software_root('VTK'):
+            self.cfg['user_packages'] = [x for x in self.cfg['user_packages'] if x != 'VTK']
```
akesandgren, line 125: This feels unsafe, but I'm not a Python expert.... @boegel ??

edmondac: I think it's safe, in that it's just removing 'VTK' from the list if it's present, and otherwise keeping everything. But happy for @boegel to decide :-)

akesandgren: I meant that user_packages is on both sides of the operation...

edmondac: Should be fine I think. It builds the list on the right then assigns it... IIRC.

```
>>> a = [1,2,3]
>>> a = [x for x in a if x > 1]
>>> a
[2, 3]
```

boegel: It's fine indeed: you're creating a new list first (based on `self.cfg['user_packages']`), and then do the assignment which overwrites the `self.cfg['user_packages']` value.

boegel:
Test report by @boegel
 * **SUCCESS** _LAMMPS-3Mar2020-foss-2019b-Python-3.7.4-kokkos.eb_ 
 * **SUCCESS** _LAMMPS-3Mar2020-foss-2020a-Python-3.8.2-kokkos.eb_ 
 * **SUCCESS** _LAMMPS-3Mar2020-intel-2019b-Python-3.7.4-kokkos.eb_ 
 * **SUCCESS** _LAMMPS-3Mar2020-intel-2020a-Python-3.8.2-kokkos.eb_ 
 * **SUCCESS** _LAMMPS-7Aug2019-foss-2019b-Python-3.7.4-kokkos.eb_ 
 * **SUCCESS** _LAMMPS-7Aug2019-intel-2019b-Python-3.7.4-kokkos-OCTP.eb_ 
 * **SUCCESS** _LAMMPS-7Aug2019-intel-2019b-Python-3.7.4-kokkos.eb_ 

Build succeeded for 7 out of 7 (7 easyconfigs in total)
node3104.skitty.os - Linux centos linux 7.9.2009, x86_64, Intel(R) Xeon(R) Gold 6140 CPU @ 2.30GHz, Python 3.6.8
See https://gist.github.com/2ea94b9786ef704a580683617ec283be for a full test report.

boegel (approved):
lgtm

PR #2255, opened 2020-11-27
add custom easyblock for code-server

ocaisa:


easybuild/easyblocks/c/code_server.py
```diff
@@ -0,0 +1,53 @@
...
+            mapped_arch = 'amd64'
+        elif myarch == AARCH64:
+            mapped_arch = 'arm64'
+        else:
+            raise EasyBuildError("Architecture %s is not supported for code-server on EasyBuild", myarch)
+
+        super(EB_code_minus_server, self).__init__(*args, **kwargs)
+
+        self.cfg.template_values['mapped_arch'] = mapped_arch
+        self.cfg.generate_template_values()
```
boegel, line 53: Also implement the installation procedure and sanity check in the easyblock?

See also https://github.com/easybuilders/easybuild-easyconfigs/pull/11778#pullrequestreview-541684454

boegel (dismissed):

easybuild/easyblocks/c/code_server.py
ocaisa: Done

Micket:
Hm, what's up with ` easybuild/easyblocks/c/chimera.py` the empty file?

ocaisa:
When I was in that dir I noticed it had execute permissions, so I removed them

Micket (approved):
lgtm

PR #2262, opened 2020-12-02
add custom easyblock for JAX

edmondac:
For https://github.com/easybuilders/easybuild-easyconfigs/pull/11595

### DIFF
--- a/easybuild/easyblocks/b/bazel.py
+++ b/easybuild/easyblocks/b/bazel.py
@@ -41,6 +41,7 @@ from easybuild.framework.easyconfig import CUSTOM
 class EB_Bazel(EasyBlock):
     """Support for building/installing Bazel."""
 
+    @staticmethod
     def extra_options():
         """Extra easyconfig parameters specific to EB_Bazel."""
         extra_vars = {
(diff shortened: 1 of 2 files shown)


edmondac:
See https://github.com/easybuilders/easybuild-easyconfigs/pull/11595 for successful test reports. The CI on there won't work until this PR is merged, I believe.

Flamefire:
@edmondac Could you factor out the Bazel EB? I'd like to see that merged fast as my TF stuff is blocked by that ****

edmondac:
> @edmondac Could you factor out the Bazel EB? I'd like to see that merged fast as my TF stuff is blocked by that ****

Sure - Bazel commits cherry-picked to a new branch in https://github.com/easybuilders/easybuild-easyblocks/pull/2272

branfosj:
easybuilders/easybuild-easyconfigs#11595 is closed as we are going with a different method for installing jax / jaxlib, so this is unneeded.

Flamefire:
We might revisit this though depending on how hard the EC turns out to be and if we want to move stuff out of it. But technically we don't need an easyblock, well besides patching the TF dependency

PR #2264, opened 2020-12-03
clang: introduce build_lldb to build debugger

bartoldeman:
This is completely analogous to build_lld which builds the linker


Micket (approved):
lgtm

Micket:
Shall we just close this since we opted for a standalone module instead?

bartoldeman:
ok, I'll close it then. With the Python dep it becomes very awkward.

PR #2272, opened 2020-12-07
add support for statically linking Bazel

edmondac:
Static Bazel changes cherry-picked from https://github.com/easybuilders/easybuild-easyblocks/pull/2262

### DIFF
--- a/easybuild/easyblocks/b/bazel.py
+++ b/easybuild/easyblocks/b/bazel.py
@@ -41,6 +41,7 @@ from easybuild.framework.easyconfig import CUSTOM
 class EB_Bazel(EasyBlock):
     """Support for building/installing Bazel."""
 
+    @staticmethod
     def extra_options():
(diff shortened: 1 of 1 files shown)


Flamefire:
Tested this and seems to work. Another approach might be to use the (recently created) upstream patch: https://github.com/bazelbuild/bazel/pull/12651/files
Maybe using `--action_env=LD_LIBRARY_PATH` to the build process of the dependency would also work. Testing that now so we at least know our options.

Flamefire:
Currently testing the official patch. Is there a preference for when there is a known patch for the next version but many versions require the patch? I.e. still do the workaround in the EasyBlock or add the patch to prior ECs?
Note that the change in this PR is opt-in so ECs would need to be changed too.

boegel:
superseded by #2285 (at least, once the static linking is also done opt-in there like it is here)

PR #2274, opened 2020-12-09
add easyblock for RELION

lexming:


migueldiascosta:
@lexming it seems https://github.com/easybuilders/easybuild-easyblocks/pull/2280 was not enough, there's a similar code block in `test/easyblocks/module.py`

lexming:
@migueldiascosta fix for those tests in https://github.com/easybuilders/easybuild-easyblocks/pull/2302

lexming:
Successful test report in https://github.com/easybuilders/easybuild-easyconfigs/pull/10965#issuecomment-755300982

JackPerdue:
It might be  useful to split out the part concerning batch/queue systems (i.e. Slurm/Torque/LSF(?)) and their particulars to its own easyblock for use by others that might want to use it.  I don't know what the EB framework uses for when it uses jobs to build things (I've always just built on the login nodes), but it might be able to use it too.  Just a thought.

dithwick:
Two other quick thoughts (although maybe best for a later update):

* ALTCPU=ON should only be enabled when TBB is in the dependency chain, maybe it's worth detecting this (although I note the corresponding easyconfig does have this).
* MKLFFT=ON maybe should be added when MKL is detected.

shahpnmlab (commented):
Line 171, results in a str float comparison which is not valid.

PR #2280, opened 2020-12-14
prefer default value for extra options in easyblock tests

lexming:
Use the default value for mandatory extra options which have one and fallback to `foo` only if if the default is not set (`None`, `False` or empty).

This allows having extra options with a limited number of valid options and check for those. Raising an error if the provided option is not one of the allowed values.

For instance, as it is done for `queue_cmd` in the relion easyblock (PR #2274 ) at https://github.com/easybuilders/easybuild-easyblocks/blob/59a50773183dc7b6d4579833d4bd7018af5d4c5a/easybuild/easyblocks/r/relion.py#L53

### DIFF
--- a/test/easyblocks/init_easyblocks.py
+++ b/test/easyblocks/init_easyblocks.py
@@ -151,11 +151,16 @@ def template_init_test(self, easyblock, name='foo', version='1.3.2'):
         extra_options = app_class.extra_options()
         check_extra_options_format(extra_options)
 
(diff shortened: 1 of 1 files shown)


migueldiascosta:
for a while I didn't see that this was in the tests and was very confused with that original `foo` value :)

lgtm

migueldiascosta (approved):
lgtm

migueldiascosta:
Going in, thanks @lexming!

PR #2287, opened 2020-12-17
rework module-only tests to use unique software name (rather than 'foo')

Flamefire:
I experienced a test failure of morthur finding artefacts from Bazel easyblock
This is caused by the test modules all having the same name so
a) Tests are incomplete as they may simply assert things fulfilled by
   previous subtests
b) Tests may wrongly fail as previous tests may add unexpected
   files/folders which are picked up by the next tests

Solution: Use unique names and workaround some limitations discovered by that


boegel (approved):
lgtm

PR #2302, opened 2021-01-06
prefer default value for extra options in easyblock module-only tests

lexming:
Follow-up to https://github.com/easybuilders/easybuild-easyblocks/pull/2280 for `--module-only` tests.

Use the default value for mandatory extra options which have one and fallback to `foo` only if the default is not set (`None`, `False` or empty).

This allows having extra options with a limited number of valid options and check for those. Raising an error if the provided option is not one of the allowed values.

For instance, as it is done for `queue_cmd` in the relion easyblock (PR #2274 ) at https://github.com/easybuilders/easybuild-easyblocks/blob/59a50773183dc7b6d4579833d4bd7018af5d4c5a/easybuild/easyblocks/r/relion.py#L53


boegel (approved):
lgtm

PR #2305, opened 2021-01-11
add custom easyblock to support installation of new Intel compilers (v2021.x, oneAPI)

boegel:
* enhance `IntelBase` to implement installation procedure for 2021.x versions of Intel tools
* add new custom easyblock for `intel-compilers`, to be used with Intel oneAPI HPC Base toolkit v2021 & newer
* + small bug fix in `Bundle` that I hit when working on this

requires
* ~~https://github.com/easybuilders/easybuild-framework/pull/3535~~ (support for `--accept-eula`)
* ~~https://github.com/easybuilders/easybuild-framework/pull/3546~~ to allow using custom name when calling `check_accepted_eula`

### DIFF
--- a/easybuild/easyblocks/i/intel_compilers.py
+++ b/easybuild/easyblocks/i/intel_compilers.py
@@ -31,7 +31,7 @@ import os
 from distutils.version import LooseVersion
 
 from easybuild.easyblocks.generic.intelbase import IntelBase
-from easybuild.tools.build_log import EasyBuildError
+from easybuild.tools.build_log import EasyBuildError, print_msg
 
 
 class EB_intel_minus_compilers(IntelBase):
@@ -76,6 +76,19 @@ class EB_intel_minus_compilers(IntelBase):
             # Fortran Compiler & Fortran Compiler Classic
             'intel.oneapi.lin.ifort-compiler',
         ]
+
+    def install_step(self):
+        """
+        Install step: install each 'source file' one by one.
+        To install a patch release of Intel oneAPI compilers, we need to install the HPC Toolkit first,
+        and then separate updates for the C++ and Fortran compilers...
+        """
+        srcs = self.src[:]
+        cnt = len(srcs)
+        for idx, src in enumerate(srcs):
+            print_msg("installing part %d/%s (%s)..." % (idx + 1, cnt, src['name']))
+            self.src = [src]
+            super(EB_intel_minus_compilers, self).install_step()
 
     def sanity_check_step(self):
         """
--- a/easybuild/easyblocks/i/intel_compilers.py
+++ b/easybuild/easyblocks/i/intel_compilers.py
@@ -1,0 +1,124 @@
+# #
+# Copyright 2021-2021 Ghent University
+#
+# This file is part of EasyBuild,
+# originally created by the HPC team of Ghent University (http://ugent.be/hpc/en),
+# with support of Ghent University (http://ugent.be/hpc),
+# the Flemish Supercomputer Centre (VSC) (https://www.vscentrum.be),
+# Flemish Research Foundation (FWO) (http://www.fwo.be/en)
+# and the Department of Economy, Science and Innovation (EWI) (http://www.ewi-vlaanderen.be/en).
+#
+# https://github.com/easybuilders/easybuild
+#
+# EasyBuild is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation v2.
+#
+# EasyBuild is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with EasyBuild.  If not, see <http://www.gnu.org/licenses/>.
+# #
+"""
+EasyBuild support for installing Intel compilers, implemented as an easyblock
+
+@author: Kenneth Hoste (Ghent University)
+"""
+import os
+from distutils.version import LooseVersion
+
+from easybuild.easyblocks.generic.intelbase import IntelBase
+from easybuild.tools.build_log import EasyBuildError
+
+
+class EB_intel_minus_compilers(IntelBase):
+    """
+    Support for installing Intel compilers, starting with verion 2021.x (oneAPI)
+    """
+
+    def __init__(self, *args, **kwargs):
+        """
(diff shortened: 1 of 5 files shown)


boegel:
@rscohn2 Any feedback on this?

rscohn2:
It looks good. Where is the code that constructs the URL for downloads?

ocaisa:
@rscohn2 You can find the URL for downloading in https://github.com/easybuilders/easybuild-easyconfigs/pull/11982/files

easybuild/easyblocks/i/intel_compilers.py
```diff
@@ -0,0 +1,124 @@
...
+        # avoid that IntelBase trips over not having license info specified
+        kwargs['requires_runtime_license'] = False
+
+        super(EB_intel_minus_compilers, self).prepare_step(*args, **kwargs)
+
+    def configure_step(self):
+        """Configure installation: specify components to install."""
+
+        # redefine $HOME, to avoid that anything is stored in $HOME/intel (like the 'installercache' database)
+        os.environ['HOME'] = self.builddir
```
ocaisa, line 69: This is problematic, it broke GitHub integration for me when there was a failure. It also doesn't correspond to what happens with `https://github.com/easybuilders/easybuild-easyblocks/pull/2305/files#diff-3dc6b8bfa78867742693cea33fdba081128e823f26ba25e177dd68d83adfec62R211`, this should be used instead?

ocaisa: If not then you will also need a custom cleanup step, or make the one in intelbase a bit more resilient

ocaisa: I think you should revert the redefinition of `HOME` as soon as you no longer need it, and make `clean_home_subdir()` only do something if the directory actually exists. That should fix the issues I saw.

scroogie:
Can I do anything to help this getting forward?

easybuild/easyblocks/i/intel_compilers.py
rscohn2: FYI: The installer team confirmed that setting the HOME variable is the best way to control where the files are dropped.

boegel: Thanks for the feedback @rscohn2!

boegel:
> Can I do anything to help this getting forward?

If you're up for it, you could look into the necessary changes w.r.t. redefining `$HOME` and rolling that back in time.

One way could be to adding `'HOME=%s' % self.builddir` to the `cmd` being constructed in `install_step_oneapi` (rather than setting `$HOME` globally via `os.environ`), and then check whether that works as expected.

You can test using `eb --from-pr 11982` (see https://github.com/easybuilders/easybuild-easyconfigs/pull/11982), but that PR also needs a little bit of updating.

I hope to pick up on this again myself soon, but finding time for it has not been easy in recent weeks... 😅

scroogie:
Okay, not sure I have the necessary knowledge yet, but I'll have a look. 

Regarding the variable, subprocess.Popen() actually allows passing an env, which we could just copy from os.environ. Then we dont run any risk of modifying the live env. Maybe that would be a sensible addition to run_cmd? 

I'll try prefixing the command as well though. What I don't know, does that actually work in all shells, i.e. tscsh or ksh, or are we always in bash (didnt find it directly)?

scroogie:
@boegel Just a short ping to ask which direction you think makes more sense.

boegel:
@scroogie `run_cmd` always runs commands in a bash subshell I think, so just prefixing the command with `HOME=...` should work

easybuild/easyblocks/i/intel_compilers.py
ocaisa: What about changing this line to
```
self.cfg['preinstallopts'] += " HOME=%s " % self.builddir
```

ocaisa:
The issues I saw with this are addressed in https://github.com/boegel/easybuild-easyblocks/pull/37

boegel:
Don't merge this just yet, may need some more love since to get the 2021.1.2 version of the compilers you need to install the HPC base kit + additional updates for C/C++ and Fortran compilers; see https://community.intel.com/t5/Intel-Fortran-Compiler/Announcing-Intel-Fortran-Compiler-Patch-Release-2021-1-2-Answers/m-p/1240262#M153503...

boegel:
@ocaisa I think this is good to go now...

easybuild/easyblocks/i/intel_compilers.py
```diff
@@ -0,0 +1,131 @@
...
+    def configure_step(self):
+        """Configure installation."""
+
+        # redefine $HOME for install step, to avoid that anything is stored in $HOME/intel
+        # (like the 'installercache' database)
+        self.cfg['preinstallopts'] += " HOME=%s " % self.builddir
+
+    def install_step(self):
+        """
+        Install step: install each 'source file' one by one.
```
ocaisa, line 74: I don't think this docstring is completely accurate any more

boegel: fixed in 0481c97b5

boegel:
Test report by @boegel
 * **SUCCESS** _icc-2016.3.210-GCC-5.4.0-2.26.eb_ 
 * **SUCCESS** _ifort-2016.3.210-GCC-5.4.0-2.26.eb_ 
 * **SUCCESS** _impi-5.1.3.181-iccifort-2016.3.210-GCC-5.4.0-2.26.eb_ 
 * **SUCCESS** _imkl-11.3.3.210-iimpi-2016b.eb_ 
 * **SUCCESS** _intel-2016b.eb_ 

node2610.swalot.os - Linux centos linux 7.9.2009, x86_64, Intel(R) Xeon(R) CPU E5-2660 v3 @ 2.60GHz (haswell), Python 3.6.8
See https://gist.github.com/69a657eb1ad0f2da5b0a4029325d4e0c for a full test report.

ocaisa (approved):
LGTM

boegel:
Test report by @boegel
 * **SUCCESS** _iccifort-2020.4.304.eb_ 
 * **SUCCESS** _impi-2019.9.304-iccifort-2020.4.304.eb_ 
 * **SUCCESS** _imkl-2020.4.304-iimpi-2020b.eb_ 
 * **SUCCESS** _intel-2020b.eb_ 

node3300.joltik.os - Linux centos linux 7.9.2009, x86_64, Intel(R) Xeon(R) Gold 6242 CPU @ 2.80GHz (cascadelake), Python 3.6.8
See https://gist.github.com/2d4bfdab8cf88cda0b58fdb7871e6c0a for a full test report.

PR #2313, opened 2021-01-20
update impi easyblock for impi 2021.x (oneAPI)

boegel:
requires ~~#2305~~

~~WIP, sanity check doesn't pass?~~ (fixed in cc6e3ee)


bartoldeman (approved):
lgtm, tested in https://github.com/easybuilders/easybuild-easyconfigs/pull/12026

PR #2349, opened 2021-02-26
enhance CUDA support in CP2K easyblock (WIP)

boegel:
(replacement for #1703)

I'm not sure I'm entirely happy with these changes, there's quite a bit of hardcoding going on, this probably deserves one or more additional custom easyconfig parameters?

### DIFF
--- a/easybuild/easyblocks/c/cp2k.py
+++ b/easybuild/easyblocks/c/cp2k.py
@@ -228,12 +228,30 @@ class EB_CP2K(EasyBlock):
             else:
                 options['DFLAGS'] += ' -D__ELPA3'
 
-        # CUDA
+        # CUDA support
+        # see https://github.com/cp2k/cp2k/blob/master/INSTALL.md#2j-cuda-optional-improved-performance-on-gpu-systems
         cuda = get_software_root('CUDA')
         if cuda:
-            options['DFLAGS'] += ' -D__ACC -D__DBCSR_ACC'
-            options['LIBS'] += ' -lcudart -lcublas -lcufft -lrt'
-            options['NVCC'] = ' nvcc'
+            # check GPU architecture(s) to build for
+            cuda_cc_sm_vals = self.cfg.template_values.get('cuda_sm_comma_sep')
+            if cuda_cc_sm_vals:
+                gpu_targets = '--gpu-code=%s' % cuda_cc_sm_vals
+            else:
+                gpu_targets = ''
+
+            options['DFLAGS'] += ' -D__ACC -D__DBCSR_ACC -D__PW_CUDA -D__GRID_CUDA'
+            options['LIBS'] += ' -lcudart -lnvrtc -lcuda -lcublas -lcufft -lrt'
+            options['NVCC'] = 'nvcc'
+            options['NVFLAGS'] = ' '.join([
+                options['DFLAGS'],
+                '-O3',
+                '--std=c++11',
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/c/cp2k.py
```diff
@@ -228,12 +228,30 @@ def configure_step(self):
...
         if cuda:
-            options['DFLAGS'] += ' -D__ACC -D__DBCSR_ACC'
-            options['LIBS'] += ' -lcudart -lcublas -lcufft -lrt'
-            options['NVCC'] = ' nvcc'
+            # check GPU architecture(s) to build for
+            cuda_cc_sm_vals = self.cfg.template_values.get('cuda_sm_comma_sep')
+            if cuda_cc_sm_vals:
+                gpu_targets = '--gpu-code=%s' % cuda_cc_sm_vals
+            else:
+                gpu_targets = ''
```
boegel, line 240: This needs a better default, and/or a warning if no CUDA compute capabilities were specified?

easybuild/easyblocks/c/cp2k.py
```diff
@@ -228,12 +228,30 @@ def configure_step(self):
...
+            options['NVFLAGS'] = ' '.join([
+                options['DFLAGS'],
+                '-O3',
+                '--std=c++11',
+                gpu_targets,
+                # control host compilers + options
+                '-ccbin="%s"' % os.getenv('CXX'),
+                '-Xcompiler="%s"' % os.getenv('CXXFLAGS'),
+            ])
+            options['GPUVER'] = 'V100'
```
boegel, line 254: This is obviously hardcoded, should we auto-detect what the correct value should be for this?

Not sure what the significance of this is, actually...

easybuild/easyblocks/c/cp2k.py
```diff
@@ -228,12 +228,30 @@ def configure_step(self):
...
-            options['LIBS'] += ' -lcudart -lcublas -lcufft -lrt'
-            options['NVCC'] = ' nvcc'
+            # check GPU architecture(s) to build for
+            cuda_cc_sm_vals = self.cfg.template_values.get('cuda_sm_comma_sep')
+            if cuda_cc_sm_vals:
+                gpu_targets = '--gpu-code=%s' % cuda_cc_sm_vals
+            else:
+                gpu_targets = ''
+
+            options['DFLAGS'] += ' -D__ACC -D__DBCSR_ACC -D__PW_CUDA -D__GRID_CUDA'
```
boegel, line 242: Do we need easyconfig parameters to opt-out of using `-D__PW_CUDA` and `-D__GRID_CUDA`?

easybuild/easyblocks/c/cp2k.py
Flamefire: Is this the point where you said in the call that only a single GPU arch is supported?

I just searched the source of CP2K and didn't find any place where this is actually used: https://github.com/cp2k/cp2k/search?q=GPUVER
All uses are either in the autotools build chain, which we don't use (I'd guess we could?) So I guess we don't need it at all. Or did I miss anything?

boegel: No, this is really required, I tried without it. It's used by the DBCSR component.

I have an updated version of this, which I apparently haven't pushed yet, doing that right now (see 68e15c56bcb13f685695a1315f27489ee1acdf3c)

PR #2356, opened 2021-03-10
enhance cuDNN and CUDA easyblocks to support aarch64

robert-mijakovic:
Adds armsbsa in cuDNN easyblock.

### DIFF
--- a/easybuild/easyblocks/c/cuda.py
+++ b/easybuild/easyblocks/c/cuda.py
@@ -1,14 +1,26 @@
 ##
-# This file is an EasyBuild reciPY as per https://github.com/easybuilders/easybuild
+# Copyright 2012-2021 Ghent University
 #
-# Copyright:: Copyright 2012-2019 Cyprus Institute / CaSToRC, Uni.Lu, NTUA, Ghent University,
-#             Forschungszentrum Juelich GmbH
-# Authors::   George Tsouloupas <g.tsouloupas@cyi.ac.cy>, Fotis Georgatos <fotis@cern.ch>, Kenneth Hoste, Damian Alvarez
-# License::   MIT/GPL
-# $Id$
+# This file is part of EasyBuild,
+# originally created by the HPC team of Ghent University (http://ugent.be/hpc/en),
+# with support of Ghent University (http://ugent.be/hpc),
+# the Flemish Supercomputer Centre (VSC) (https://www.vscentrum.be),
+# Flemish Research Foundation (FWO) (http://www.fwo.be/en)
+# and the Department of Economy, Science and Innovation (EWI) (http://www.ewi-vlaanderen.be/en).
 #
-# This work implements a part of the HPCBIOS project and is a component of the policy:
-# http://hpcbios.readthedocs.org/en/latest/HPCBIOS_2012-99.html
+# https://github.com/easybuilders/easybuild
+#
+# EasyBuild is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation v2.
+#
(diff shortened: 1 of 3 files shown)


easybuild/easyblocks/c/cuda.py
```diff
@@ -32,7 +45,7 @@
 from easybuild.tools.filetools import adjust_permissions, patch_perl_script_autoflush
 from easybuild.tools.filetools import remove_file, which, write_file
 from easybuild.tools.run import run_cmd, run_cmd_qa
-from easybuild.tools.systemtools import POWER, X86_64, get_cpu_architecture, get_shared_lib_ext
+from easybuild.tools.systemtools import POWER, X86_64, AARCH64, get_cpu_architecture, get_shared_lib_ext
```
boegel, line 48: Please sort alphabetically (`AARCH64` before `POWER`), same below

easybuild/easyblocks/j/java.py
```diff
@@ -36,13 +37,30 @@
 from easybuild.easyblocks.generic.packedbinary import PackedBinary
 from easybuild.tools.filetools import adjust_permissions, change_dir, copy_dir, copy_file, remove_dir
 from easybuild.tools.run import run_cmd
+from easybuild.tools.build_log import EasyBuildError
+from easybuild.tools.systemtools import X86_64, AARCH64, get_cpu_architecture
```
boegel, line 41: Please also retain alphabetical sorting here, also between the `from` lines:

```python
from easybuild.tools.systemtools import AARCH64, X86_64, get_cpu_architecture

File: easybuild/easyblocks/j/java.py [hunk 3]
```diff
@@ -36,13 +37,30 @@
...
 
+    def __init__(self, *args, **kwargs):
+        """ Init the Java easyblock adding a new javaarch template var """
+        myarch = get_cpu_architecture()
+        if myarch == X86_64:
+            javaarch = 'x64'
+        elif myarch == AARCH64:
+            javaarch = 'aarch64'
+        else:
+            raise EasyBuildError("Architecture %s is not supported for Java on EasyBuild", myarch)
```
boegel, line 57: What about POWER?
See https://github.com/easybuilders/easybuild-easyconfigs/blob/develop/easybuild/easyconfigs/j/Java/Java-11.0.6-ppc64le.eb

boegel (changes requested):

easybuild/easyblocks/j/java.py
```diff
@@ -36,13 +37,30 @@
...
+
+        super(EB_Java, self).__init__(*args, **kwargs)
+
+        self.cfg.template_values['javaarch'] = javaarch
```
boegel, line 61: In which context do you need this?

ocaisa: This also came up in the conf call today (raised by @Micket), currently Java is a special case and uses a modulerc, not sure what the advantage is of adding templating since the existing easyconfigs require architecture versionsuffixes (and it doesn't look like this will change).

boegel (approved):
lgtm

boegel:
Test report by @boegel
 * **SUCCESS** _CUDA-11.1.1-GCC-10.2.0.eb_ 
 * **SUCCESS** _cuDNN-8.0.4.30-CUDA-11.1.1.eb_ 

node3131.skitty.os - Linux centos linux 7.9.2009, x86_64, Intel(R) Xeon(R) Gold 6140 CPU @ 2.30GHz, Python 3.6.8
See https://gist.github.com/3efeb32d8cf4a24ded0375d59c5c0d7b for a full test report.

PR #2365, opened 2021-03-19
Make sure that path to eb is always found by tests

ocaisa:


ocaisa:
This was to fix https://github.com/easybuilders/easybuild/issues/684 for the easyblock repo, but actually (for me) the problem doesn't occur here

ocaisa:
I think it's better to close this rather than introduce unnecessary code

PR #2370, opened 2021-03-26
Fix accidental change of the default source_urls in the PythonPackage EB

Flamefire:
The original list was modified as a side effect, a copy is required here

edit: fix for bug introduced in #2364


boegel (approved):
lgtm

PR #2375, opened 2021-03-30
Ignore: Lumerical does fairly random things with paths ...

cwant:
This PR handles the following issues (while maintaining reverse compatibility):

* Unexpected name of RPM (e.g., `Lumerical-2021R1-2-dc09d2e.el6.x86_64.rpm` in tarball `Lumerical-2021-R1.2-2621-dc09d2e.tar.gz`)
* Unexpected base directory of install (e.g., `opt/lumerical/v221` instead of `opt/lumerical/2021R1` or such)

This PR doesn't handle the patch file though (pretty much have to hack a new patch each time).
To solve this, probably need to rework `apply_patch` method in the lumerical subclass:

 https://github.com/easybuilders/easybuild-framework/blob/ee83bb36ac0191d92ad58dbc391eb237c1cf855c/easybuild/framework/application.py#L574

(E.g., apply it in different directory with higher `-p` flag to `patch`)


cwant:
Sorry, this was meant to be for the Compute Canada repository. For whatever reason, github assumed that I wanted to issue a PR against your upstream repository.

boegel:
@cwant No problem, thanks for clarifying what went wrong! :)

PR #2377, opened 2021-04-01
Improve Python package version check and add unversioned_packages EC param

Flamefire:
This fixes a format argument confusion in the check for a 0.0.0 version and enhances the error message.
As an escape hatch I added an EC param `unversioned_packages` which contains Python package names which should not trigger a failure on a 0.0.0 version as a last resort, e.g. as of recently in the opencarp package: https://github.com/easybuilders/easybuild-easyconfigs/pull/12462

As those checks can be annoying as they run after a potentially long installation I bundled all error messages together.

As it is impossible to detect the package name used by `pip list` automatically, the `unversioned_packages` must be set even in ECs using other ECs which have such packages. I expect that to be very rare though.
As an explanation: EC "python-protobuf` has a modulename of "google.protobuf" and a pip name of "protobuf". The Keras-* packages are likely similar. Reason is that the name used by pip is in the meta data of the package and could theoretically be anything, IIUC

### DIFF
--- a/easybuild/easyblocks/generic/pythonpackage.py
+++ b/easybuild/easyblocks/generic/pythonpackage.py
@@ -829,7 +829,7 @@ class PythonPackage(ExtensionEasyBlock):
                                        'Check that the name from `pip list` is used which may be different than the '
                                        'module name.' % unversioned_package)
                             else:
-                                msg = ('Package %s in unversioned_packages Has a version of %s which is valid. '
+                                msg = ('Package %s in unversioned_packages has a version of %s which is valid. '
                                        'Please remove it from unversioned_packages.' % (unversioned_package, version))
                             pip_check_errors.append(msg)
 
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/generic/pythonpackage.py
```diff
@@ -796,23 +801,55 @@ def sanity_check_step(self, *args, **kwargs):
...
+                            self.log.debug('Excluding unversioned package %s from check', unversioned_package)
+                        except ValueError:
+                            try:
+                                version = next(pkg['version'] for pkg in pkgs if pkg['name'] == unversioned_package)
+                            except StopIteration:
+                                msg = ('Package %s in unversioned_packages was not found in the installed packages. '
+                                       'Check that the name from `pip list` is used which may be different than the '
+                                       'module name.' % unversioned_package)
+                            else:
+                                msg = ('Package %s in unversioned_packages Has a version of %s which is valid. '
```
boegel, line 832:
```suggestion
                                msg = ('Package %s in unversioned_packages has a version of %s which is valid. '
```

boegel (changes requested):
tiny typo to fix...

boegel (approved):
lgtm

Tested with both an easyconfig file that now fails due to a `0.0.0` version being installed (`carputils-20200915-foss-2020a-Python-3.8.2.eb`), and verifying that `unversioned_packages` works as expected, and with `PythonPackage` easyconfigs that are not affected.

PR #2381, opened 2021-04-08
add support for post install commands for python extensions

smoors:
post install commands can be done with a hack if using `python setup.py install`:
```
'installopts': ' && some_post_install_stuff',
```
with pip however, this hack is not possible.

this PR adds support for post install commands for each individual extension, so one can use it in the same way as the usual `postinstallcmds`:
```
'ext_postinstallcmds': [list of commands]',
```

### DIFF
--- a/easybuild/easyblocks/generic/pythonpackage.py
+++ b/easybuild/easyblocks/generic/pythonpackage.py
@@ -59,9 +59,15 @@ from easybuild.tools.hooks import CONFIGURE_STEP, BUILD_STEP, TEST_STEP, INSTALL
 # not 'easy_install' deliberately, to avoid that pkg installations listed in easy-install.pth get preference
 # '.' is required at the end when using easy_install/pip in unpacked source dir
 EASY_INSTALL_TARGET = "easy_install"
-EASY_INSTALL_INSTALL_CMD = "%(python)s setup.py " + EASY_INSTALL_TARGET + " --prefix=%(prefix)s %(installopts)s %(loc)s %(postextinstallcmds)s"
+EASY_INSTALL_INSTALL_CMD = (
+    "%(python)s setup.py " +
+    EASY_INSTALL_TARGET +
+    " --prefix=%(prefix)s %(installopts)s %(loc)s %(postextinstallcmds)s"
+)
 PIP_INSTALL_CMD = "pip install --prefix=%(prefix)s %(installopts)s %(loc)s %(postextinstallcmds)s"
-SETUP_PY_INSTALL_CMD = "%(python)s setup.py %(install_target)s --prefix=%(prefix)s %(installopts)s %(postextinstallcmds)s"
+SETUP_PY_INSTALL_CMD = (
+    "%(python)s setup.py %(install_target)s --prefix=%(prefix)s %(installopts)s %(postextinstallcmds)s"
+)
 SETUP_PY_DEVELOP_CMD = "%(python)s setup.py develop --prefix=%(prefix)s %(installopts)s %(postextinstallcmds)s"
 UNKNOWN = 'UNKNOWN'
 
--- a/easybuild/easyblocks/generic/pythonpackage.py
+++ b/easybuild/easyblocks/generic/pythonpackage.py
@@ -59,10 +59,10 @@ from easybuild.tools.hooks import CONFIGURE_STEP, BUILD_STEP, TEST_STEP, INSTALL
 # not 'easy_install' deliberately, to avoid that pkg installations listed in easy-install.pth get preference
 # '.' is required at the end when using easy_install/pip in unpacked source dir
 EASY_INSTALL_TARGET = "easy_install"
-EASY_INSTALL_INSTALL_CMD = "%(python)s setup.py " + EASY_INSTALL_TARGET + " --prefix=%(prefix)s %(installopts)s %(loc)s"
-PIP_INSTALL_CMD = "pip install --prefix=%(prefix)s %(installopts)s %(loc)s"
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/generic/pythonpackage.py
```diff
@@ -259,6 +265,8 @@ def extra_options(extra_vars=None):
                                            "to be the requirements file.", CUSTOM],
             'use_setup_py_develop': [False, "Install using '%s' (deprecated)" % SETUP_PY_DEVELOP_CMD, CUSTOM],
             'zipped_egg': [False, "Install as a zipped eggs (requires use_easy_install)", CUSTOM],
+            'postextinstallcmds': ['', "List of additional commands to execute "
```
ocaisa, line 268:
```suggestion
            'postextinstallcmds': [[], "List of additional commands to execute "
```
This is a list to I think this needs to initialised as one

ocaisa (changes requested):
Since `postinstallcmds` is already in framework, I think it would be better to name this option to something like `ext_postinstallcmds` so that it'll turn up when searching the repo

smoors:
@ocaisa thanks for the suggestions.
I've changed it to `ext_postinstallcmds` and made the code similar to how `postinstallcmds` was implemented, with added type checks.
I've also changed the default value to `None` instead of `[]` as this is also the case for `postinstallcmds`.

ocaisa (commented):
Code LGTM, have you got an easyconfig that uses it?

smoors:
@ocaisa see https://github.com/easybuilders/easybuild-easyconfigs/pull/12559

ocaisa:
Tested with https://github.com/easybuilders/easybuild-easyconfigs/pull/12559 and looks good, relevant output excerpt below, no impact on other extensions and does what it says on the tin:
```
== installing extension about 5.2 (5/7)...
  >> defining build environment for foss/2020b toolchain
  >> running command:
	[started at: 2021-04-10 08:02:04]
	[working dir: /users/ocaisa/.local/easybuild/build/DROP/1.0.3/foss-2020b-R-4.0.3/about]
	[output logged in /tmp/eb-7p6f46c9/easybuild-run_cmd-8lijbxou.log]
	tar xzf /shared/maintainers/sources/d/DROP/extensions/about-5.2.tar.gz
== 	configuring...
== 	building...
== 	testing...
== 	installing...
  >> running command:
	[started at: 2021-04-10 08:02:05]
	[working dir: /users/ocaisa/.local/easybuild/build/DROP/1.0.3/foss-2020b-R-4.0.3/about/about-5.2]
	[output logged in /tmp/eb-7p6f46c9/easybuild-run_cmd-jqxzxvve.log]
	pip install --prefix=/users/ocaisa/.local/easybuild/software/DROP/1.0.3-foss-2020b-R-4.0.3  --no-deps  --ignore-installed  --no-build-isolation  .
  >> command completed: exit 0, ran in 00h00m01s
== installing extension pandoc 1.0.2 (6/7)...
  >> defining build environment for foss/2020b toolchain
  >> running command:
	[started at: 2021-04-10 08:02:07]
	[working dir: /users/ocaisa/.local/easybuild/build/DROP/1.0.3/foss-2020b-R-4.0.3/pandoc]
	[output logged in /tmp/eb-7p6f46c9/easybuild-run_cmd-9tvs15xk.log]
	tar xzf /shared/maintainers/sources/d/DROP/extensions/v1.0.2.tar.gz
== 	configuring...
== 	building...
== 	testing...
== 	installing...
  >> running command:
	[started at: 2021-04-10 08:02:07]
	[working dir: /users/ocaisa/.local/easybuild/build/DROP/1.0.3/foss-2020b-R-4.0.3/pandoc/pandoc-1.0.2]
	[output logged in /tmp/eb-7p6f46c9/easybuild-run_cmd-u7trm40w.log]
	pip install --prefix=/users/ocaisa/.local/easybuild/software/DROP/1.0.3-foss-2020b-R-4.0.3  --no-deps  --ignore-installed  --no-build-isolation  .  && cp -a pandoc/definitions /users/ocaisa/.local/easybuild/software/DROP/1.0.3-foss-2020b-R-4.0.3/lib/python3.8/site-packages/pandoc/
  >> command completed: exit 0, ran in 00h00m01s
```

ocaisa (approved):

easybuild/easyblocks/generic/pythonpackage.py
```diff
@@ -514,6 +516,16 @@ def compose_install_command(self, prefix, extrapath=None, installopts=None):
             },
         ])
 
+        ext_postinstallcmds = self.cfg['ext_postinstallcmds']
```
boegel, line 519: Hmm, I wonder if it wouldn't have been better to use the existing `postinstallcmds` instead...

You can already specify `postinstallcmds` for a specific extension (just like any other common easyconfig parameter), but doing so won't have any effect, unless a change like this is made in easyblocks that are used to install extensions.

So, how about just renaming to `ext_postinstallcmds` to `postinstallcmds` here after all (and removing the support for picking up `ext_postinstallcmds`, while this is still only in `develop`?

cc @smoors, @ocaisa

smoors: @boegel that would indeed be better.
I went for another parameter to avoid the clash with the global `postinstallcmds`, but it's easy to fix this, by adding this line to the __init__ function in extensionblock.py (similar to the `start_dir` parameter):
```
self.cfg['postinstallcmds'] = self.ext.get('options', {}).get('postinstallcmds', None)
```
because of the clash I assumed this was intended, but now I understand it was just not yet implemented.
I can make a PR for this if that's the way to go.

Issue #2127, opened 2020-08-19
Set better defaults in PythonPackage


Flamefire:
We tend to set

```
download_dep_fail = True
use_pip = True
sanity_pip_check = True
```

for all PythonPackage ECs nowadays and pip should be the default to install PythonPackages as well as the sanity check and download_dep_fail.

Hence for EasyBuild 5.0 I propose to set this as the default which is a potentially breaking change for ECs not updated but likely to work for most packages anyway.

@boegel if agreed please add this to the 5.0 milestone

Flamefire:
Pinging in here after discovering an issue caused by not using pip: https://github.com/easybuilders/easybuild-easyblocks/issues/2389

boegel:
This will be fixed by #3022

PR #2388, opened 2021-04-14
Enhance Python easyblock: install pip2 when available, tweak defaults, create unversioned pip symlink

Flamefire:
Installing pip from the core Python allows easier installing of packages and avoids the fallback to using setup.py and egg files which can't be easily upgraded

I added an option `install_pip` for installing an initial pip+setuptools which may become the default in EB 5.0. However when that is used, then `pip` MUST be used to upgrade those 2 later in the EC or otherwise the initial version is not properly removed/updated but merely half-way patched (or with pip even kept and silently used)
Hence an additional check is added to catch this case.

This is a **breaking change**, however I see no viable alternative to fix the current bug. See https://github.com/easybuilders/easybuild-easyblocks/pull/2388#issuecomment-829118636 for more details.
Current ECs are fixed in https://github.com/easybuilders/easybuild-easyconfigs/pull/12650

The remaining changes are closely related:
- When installing pip or checking for extensions the user site dir should be ignored
- Creating the symlink `pip` to `pip2/3` spares the ECs to do that in `installopts`
- Set download_dep_fail by default. However it may fail in case we missed to install some deps in the EC which is the case for old ECs, hence I choose to only do it in recent Python 2 and in Python 3.
- Set sanity_pip_check by default, if the installed pip is new enough. If `pip check` doesn't succeed for the core Python package something is wrong... 
- **Disable** '--ignore-installed` by default which is important as we update pip and setuptools. For the other packages it has no effect, as there can't be an installed package which we would want to ignore when installing Python itself.

The last one is critical as `--ignore-installed` does not properly update but only add new files, not changing existing files leads to a mix of versions and issues such as https://github.com/easybuilders/easybuild-easyconfigs/pull/12757

### DIFF
--- a/easybuild/easyblocks/p/python.py
+++ b/easybuild/easyblocks/p/python.py
@@ -215,6 +215,13 @@ class EB_Python(ConfigureMake):
                 result = True
         return result
 
+    def _has_ensure_pip(self):
+        """Check if  this Python version has/should have the ensurepip package"""
+        # Pip is included since 3.4 via ensurepip https://docs.python.org/3.4/whatsnew/changelog.html
+        # And in 2.7.9+: https://docs.python.org/2.7/whatsnew/2.7.html#pep-477-backport-ensurepip-pep-453-to-python-2-7
+        version = LooseVersion(self.version)
+        return version >= LooseVersion('3.4.0') or (version < LooseVersion('3') and version >= LooseVersion('2.7.9'))
+
     def configure_step(self):
         """Set extra configure options."""
 
@@ -257,9 +264,8 @@ class EB_Python(ConfigureMake):
                 if LooseVersion(gcc_ver) >= LooseVersion('8.0'):
                     self.cfg.update('configopts', "--enable-optimizations")
 
-        # Pip is included since 3.4 via ensurepip https://docs.python.org/3.4/whatsnew/changelog.html
-        if LooseVersion(self.version) >= LooseVersion('3.4.0'):
-            # Default, but do it explicitly
+        if self._has_ensure_pip():
+            # Default in 3.4+, required in 2.7
             self.cfg.update('configopts', "--with-ensurepip=upgrade")
 
         modules_setup = os.path.join(self.cfg['start_dir'], 'Modules', 'Setup')
@@ -436,7 +442,7 @@ class EB_Python(ConfigureMake):
             "python -c 'import readline'",  # make sure readline support was built correctly
         ]
 
-        if LooseVersion(self.version) >= LooseVersion('3.4.0'):
+        if self._has_ensure_pip():
             # Check that pip and setuptools are installed
             custom_paths['files'].extend([
                 os.path.join('bin', pip) for pip in ('pip', 'pip3', 'pip' + self.pyshortver)
--- a/easybuild/easyblocks/p/python.py
+++ b/easybuild/easyblocks/p/python.py
@@ -224,6 +224,9 @@ class EB_Python(ConfigureMake):
         self.cfg['exts_defaultclass'] = "PythonPackage"
         self.cfg['exts_filter'] = EXTS_FILTER_PYTHON_PACKAGES
 
+        # don't add user site directory to sys.path (equivalent to python -s)
+        env.setvar('PYTHONNOUSERSITE', '1')
+
         # don't pass down any build/install options that may have been specified
         # 'make' options do not make sense for when building/installing Python libraries (usually via 'python setup.py')
         msg = "Unsetting '%s' easyconfig parameter before building/installing extensions: %s"
@@ -244,12 +247,6 @@ class EB_Python(ConfigureMake):
                                              "you must set 'use_pip=True' for pip&setuptools. "
                                              "Found 'use_pip=False' (maybe by default) for %s.",
                                              ext['name'])
-
-    def extensions_step(self, *args, **kwargs):
-        """Install extensions (PythonPackages)"""
-        # don't add user site directory to sys.path (equivalent to python -s)
-        env.setvar('PYTHONNOUSERSITE', '1')
-        super(EB_Python, self).extensions_step(*args, **kwargs)
 
     def auto_detect_lto_support(self):
         """Return True, if LTO should be enabled for current toolchain"""
--- a/easybuild/easyblocks/p/python.py
+++ b/easybuild/easyblocks/p/python.py
@@ -448,8 +448,9 @@ class EB_Python(ConfigureMake):
 
         if self._has_ensure_pip():
             # Check that pip and setuptools are installed
+            py_maj_version = self.version.split('.')[0]
             custom_paths['files'].extend([
-                os.path.join('bin', pip) for pip in ('pip', 'pip3', 'pip' + self.pyshortver)
+                os.path.join('bin', pip) for pip in ('pip', 'pip' + py_maj_version, 'pip' + self.pyshortver)
             ])
             custom_commands.extend([
                 "python -c 'import pip'",
--- a/easybuild/easyblocks/p/python.py
+++ b/easybuild/easyblocks/p/python.py
@@ -136,6 +136,10 @@ class EB_Python(ConfigureMake):
 
         super(EB_Python, self).patch_step(*args, **kwargs)
 
+        if self._has_ensure_pip():
+            # Ignore user site dir. -E ignores PYTHONNOUSERSITE, so we have to add -s
+            apply_regex_substitutions('configure', [(r"(PYTHON_FOR_BUILD=.*-E)'", r"\1 -s'")])
+
         # if we're installing Python with an alternate sysroot,
         # we need to patch setup.py which includes hardcoded paths like /usr/include and /lib64;
         # this fixes problems like not being able to build the _ssl module ("Could not build the ssl module")
--- a/easybuild/easyblocks/p/python.py
+++ b/easybuild/easyblocks/p/python.py
@@ -110,6 +110,10 @@ class EB_Python(ConfigureMake):
         """Add extra config options specific to Python."""
         extra_vars = {
             'ebpythonprefixes': [True, "Create sitecustomize.py and allow use of $EBPYTHONPREFIXES", CUSTOM],
+            'install_pip': [False,
+                            "Use the ensurepip module (Python 2.7.9+, 3.4+) to install the bundled versions "
+                            "of pip and setuptools into Python. You _must_ then use pip for upgrading pip&setuptools!",
+                            CUSTOM],
             'optimized': [True, "Build with expensive, stable optimizations (PGO, etc.) (version >= 3.5.4)", CUSTOM],
             'ulimit_unlimited': [False, "Ensure stack size limit is set to '%s' during build" % UNLIMITED, CUSTOM],
             'use_lto': [None, "Build with Link Time Optimization (>= v3.7.0, potentially unstable on some toolchains). "
@@ -142,6 +146,12 @@ class EB_Python(ConfigureMake):
                 exts_default_options[key] = ext_defaults[key]
         self.log.debug("exts_default_options: %s", self.cfg['exts_default_options'])
 
+        self.install_pip = self.cfg['install_pip']
+        if self.install_pip:
+            if not self._has_ensure_pip():
+                raise EasyBuildError("The ensurepip module required to install pip (requested by install_pip=True) "
+                                     "is not available in Python %s", self.version)
+
     def patch_step(self, *args, **kwargs):
         """
         Custom patch step for Python:
@@ -150,7 +160,7 @@ class EB_Python(ConfigureMake):
 
         super(EB_Python, self).patch_step(*args, **kwargs)
 
-        if self._has_ensure_pip():
+        if self.install_pip:
             # Ignore user site dir. -E ignores PYTHONNOUSERSITE, so we have to add -s
             apply_regex_substitutions('configure', [(r"(PYTHON_FOR_BUILD=.*-E)'", r"\1 -s'")])
 
@@ -249,6 +259,19 @@ class EB_Python(ConfigureMake):
     def configure_step(self):
         """Set extra configure options."""
 
+        if self.install_pip:
+            # When using ensurepip, then pip must be used to upgrade pip and setuptools
+            # Otherwise it will only copy new files leading to a combination of of files from the old and new version
+            use_pip_default = self.cfg['exts_default_options'].get('use_pip')
+            # self.exts is populated in fetch_step
+            for ext in self.exts:
+                if ext['name'] in ('pip', 'setuptools'):
+                    if not ext.get('options', {}).get('use_pip', use_pip_default):
+                        raise EasyBuildError("When using ensurepip to install pip (requested by install_pip=True) "
+                                             "you must set 'use_pip=True' for pip&setuptools. "
+                                             "Found 'use_pip=False' (maybe by default) for %s.",
+                                             ext['name'])
+
         # Check for and report distutils user configs which may make the installation fail
         # See https://github.com/easybuilders/easybuild-easyconfigs/issues/11009
         for cfg in [os.path.join(os.path.expanduser('~'), name) for name in ('.pydistutils.cfg', 'pydistutils.cfg')]:
@@ -288,7 +311,7 @@ class EB_Python(ConfigureMake):
                 if LooseVersion(gcc_ver) >= LooseVersion('8.0'):
                     self.cfg.update('configopts', "--enable-optimizations")
 
-        if self._has_ensure_pip():
+        if self.install_pip:
             # Default in 3.4+, required in 2.7
             self.cfg.update('configopts', "--with-ensurepip=upgrade")
 
@@ -400,7 +423,7 @@ class EB_Python(ConfigureMake):
         python_binary_path = os.path.join(self.installdir, 'bin', 'python')
         if not os.path.isfile(python_binary_path):
             symlink('python' + self.pyshortver, python_binary_path, use_abspath_source=False)
-        if self._has_ensure_pip():
+        if self.install_pip:
             pip_binary_path = os.path.join(self.installdir, 'bin', 'pip')
             if not os.path.isfile(pip_binary_path):
                 symlink('pip' + self.pyshortver, pip_binary_path, use_abspath_source=False)
@@ -471,7 +494,7 @@ class EB_Python(ConfigureMake):
             "python -c 'import readline'",  # make sure readline support was built correctly
         ]
 
-        if self._has_ensure_pip():
+        if self.install_pip:
             # Check that pip and setuptools are installed
             py_maj_version = self.version.split('.')[0]
             custom_paths['files'].extend([
--- a/easybuild/easyblocks/p/python.py
+++ b/easybuild/easyblocks/p/python.py
@@ -133,17 +133,13 @@ class EB_Python(ConfigureMake):
             self.pythonpath = os.path.join(easybuild_subdir, 'python')
 
         ext_defaults = {
-            # The only installed packages at this point are default installed ones, e.g. pip&setuptools
-            # And we want to upgrade them cleanly, i.e. uninstall them
+            # We should enable this (by default) for all extensions because the only installed packages at this point
+            # (i.e. those in the site-packages folder) are the default installed ones, e.g. pip & setuptools.
+            # And we must upgrade them cleanly, i.e. uninstall them first. This also applies to any other package
+            # which is voluntarily or accidentally installed multiple times.
+            # Example: Upgrading to a higher version after installing new dependencies.
             'pip_ignore_installed': False,
-            # Python installations must be clean. Requires pip >= 9
-            'sanity_pip_check': LooseVersion(self._get_pip_ext_version() or '0') >= LooseVersion('9'),
         }
-        # Some older ECs break if we enable download_dep_fail due to missing packages
-        # Since Python 2.7.14 (and all Python 3 versions) we usually enable download_dep_fail in the ECs
-        # already so this should not lead to failures
-        if LooseVersion(self.version) >= LooseVersion('2.7.14'):
-            ext_defaults['download_dep_fail'] = True
 
         exts_default_options = self.cfg.get_ref('exts_default_options')
         for key in ext_defaults:
@@ -156,14 +152,6 @@ class EB_Python(ConfigureMake):
             if not self._has_ensure_pip():
                 raise EasyBuildError("The ensurepip module required to install pip (requested by install_pip=True) "
                                      "is not available in Python %s", self.version)
-
-    def _get_pip_ext_version(self):
-        """Return the pip version from exts_list or None"""
-        for ext in self.cfg.get_ref('exts_list'):
-            # Must be (at least) a name-version tuple
-            if isinstance(ext, tuple) and len(ext) >= 2 and ext[0] == 'pip':
-                return ext[1]
-        return None
 
     def patch_step(self, *args, **kwargs):
         """
--- a/easybuild/easyblocks/p/python.py
+++ b/easybuild/easyblocks/p/python.py
@@ -248,7 +248,7 @@ class EB_Python(ConfigureMake):
     def extensions_step(self, *args, **kwargs):
         """Install extensions (PythonPackages)"""
         # don't add user site directory to sys.path (equivalent to python -s)
-        env.setvar('PYTHONNOUSERSITE', '1', verbose=False)
+        env.setvar('PYTHONNOUSERSITE', '1')
         super(EB_Python, self).extensions_step(*args, **kwargs)
 
     def auto_detect_lto_support(self):
@@ -363,7 +363,7 @@ class EB_Python(ConfigureMake):
 
         # don't add user site directory to sys.path (equivalent to python -s)
         # This matters e.g. when python installs the bundled pip & setuptools (for >= 3.4)
-        env.setvar('PYTHONNOUSERSITE', '1', verbose=False)
+        env.setvar('PYTHONNOUSERSITE', '1')
 
         super(EB_Python, self).configure_step()
 
--- a/easybuild/easyblocks/p/python.py
+++ b/easybuild/easyblocks/p/python.py
@@ -232,6 +232,19 @@ class EB_Python(ConfigureMake):
                 self.log.debug(msg, param, self.cfg[param])
             self.cfg[param] = ''
 
+        if self.install_pip:
+            # When using ensurepip, then pip must be used to upgrade pip and setuptools
+            # Otherwise it will only copy new files leading to a combination of of files from the old and new version
+            use_pip_default = self.cfg['exts_default_options'].get('use_pip')
+            # self.exts is populated in fetch_step
+            for ext in self.exts:
+                if ext['name'] in ('pip', 'setuptools'):
+                    if not ext.get('options', {}).get('use_pip', use_pip_default):
+                        raise EasyBuildError("When using ensurepip to install pip (requested by install_pip=True) "
+                                             "you must set 'use_pip=True' for pip&setuptools. "
+                                             "Found 'use_pip=False' (maybe by default) for %s.",
+                                             ext['name'])
+
     def extensions_step(self, *args, **kwargs):
         """Install extensions (PythonPackages)"""
         # don't add user site directory to sys.path (equivalent to python -s)
@@ -258,20 +271,6 @@ class EB_Python(ConfigureMake):
 
     def configure_step(self):
         """Set extra configure options."""
-
-        if self.install_pip:
-            # When using ensurepip, then pip must be used to upgrade pip and setuptools
-            # Otherwise it will only copy new files leading to a combination of of files from the old and new version
-            use_pip_default = self.cfg['exts_default_options'].get('use_pip')
-            # self.exts is populated in fetch_step
-            for ext in self.exts:
-                if ext['name'] in ('pip', 'setuptools'):
-                    if not ext.get('options', {}).get('use_pip', use_pip_default):
-                        raise EasyBuildError("When using ensurepip to install pip (requested by install_pip=True) "
-                                             "you must set 'use_pip=True' for pip&setuptools. "
-                                             "Found 'use_pip=False' (maybe by default) for %s.",
-                                             ext['name'])
-
         # Check for and report distutils user configs which may make the installation fail
         # See https://github.com/easybuilders/easybuild-easyconfigs/issues/11009
         for cfg in [os.path.join(os.path.expanduser('~'), name) for name in ('.pydistutils.cfg', 'pydistutils.cfg')]:
--- a/easybuild/easyblocks/p/python.py
+++ b/easybuild/easyblocks/p/python.py
@@ -137,8 +137,8 @@ class EB_Python(ConfigureMake):
             # The only installed packages at this point are default installed ones, e.g. pip&setuptools
             # And we want to upgrade them cleanly, i.e. uninstall them
             'pip_ignore_installed': False,
-            # Python installations must be clean
-            'sanity_pip_check': True,
+            # Python installations must be clean. Requires pip >= 9
+            'sanity_pip_check': LooseVersion(self._get_pip_ext_version() or '0') >= LooseVersion('9'),
         }
         exts_default_options = self.cfg.get_ref('exts_default_options')
         for key in ext_defaults:
@@ -151,6 +151,14 @@ class EB_Python(ConfigureMake):
             if not self._has_ensure_pip():
                 raise EasyBuildError("The ensurepip module required to install pip (requested by install_pip=True) "
                                      "is not available in Python %s", self.version)
+
+    def _get_pip_ext_version(self):
+        """Return the pip version from exts_list or None"""
+        for ext in self.cfg.get_ref('exts_list'):
+            # Must be (at least) a name-version tuple
+            if isinstance(ext, tuple) and len(ext) >= 2 and ext[0] == 'pip':
+                return ext[1]
+        return None
 
     def patch_step(self, *args, **kwargs):
         """
--- a/easybuild/easyblocks/p/python.py
+++ b/easybuild/easyblocks/p/python.py
@@ -200,10 +200,17 @@ class EB_Python(ConfigureMake):
         self.cfg['exts_defaultclass'] = "PythonPackage"
         self.cfg['exts_filter'] = EXTS_FILTER_PYTHON_PACKAGES
 
-        # The only installed packages at this point are default installed ones, e.g. pip&setuptools
-        # And we want to upgrade them cleanly, i.e. uninstall them
-        if 'pip_ignore_installed' not in self.cfg['exts_default_options']:
-            self.cfg['exts_default_options']['pip_ignore_installed'] = False
+        ext_defaults = {
+            'download_dep_fail': True,
+            # The only installed packages at this point are default installed ones, e.g. pip&setuptools
+            # And we want to upgrade them cleanly, i.e. uninstall them
+            'pip_ignore_installed': False,
+            # Python installations must be clean
+            'sanity_pip_check': True,
+        }
+        for key, value in ext_defaults.items():
+            if key not in self.cfg['exts_default_options']:
+                self.cfg['exts_default_options'][key] = value
 
         # don't pass down any build/install options that may have been specified
         # 'make' options do not make sense for when building/installing Python libraries (usually via 'python setup.py')
--- a/easybuild/easyblocks/p/python.py
+++ b/easybuild/easyblocks/p/python.py
@@ -43,6 +43,7 @@ from distutils.version import LooseVersion
 import easybuild.tools.environment as env
 from easybuild.easyblocks.generic.configuremake import ConfigureMake
 from easybuild.framework.easyconfig import CUSTOM
+from easybuild.framework.easyconfig.easyconfig import disable_templating
 from easybuild.tools.build_log import EasyBuildError, print_warning
 from easybuild.tools.config import build_option, log_path
 from easybuild.tools.modules import get_software_libdir, get_software_root, get_software_version
@@ -136,9 +137,12 @@ class EB_Python(ConfigureMake):
             # Python installations must be clean
             'sanity_pip_check': True,
         }
-        for key, value in ext_defaults.items():
-            if key not in self.cfg['exts_default_options']:
-                self.cfg['exts_default_options'][key] = value
+        # Disable templating to update the underlying values
+        with disable_templating(self.cfg):
+            for key, value in ext_defaults.items():
+                if key not in self.cfg['exts_default_options']:
+                    self.cfg['exts_default_options'][key] = value
+        self.log.info("exts_default_options: %s", self.cfg['exts_default_options'])
 
     def patch_step(self, *args, **kwargs):
         """
--- a/easybuild/easyblocks/p/python.py
+++ b/easybuild/easyblocks/p/python.py
@@ -128,6 +128,18 @@ class EB_Python(ConfigureMake):
             easybuild_subdir = log_path()
             self.pythonpath = os.path.join(easybuild_subdir, 'python')
 
+        ext_defaults = {
+            'download_dep_fail': True,
+            # The only installed packages at this point are default installed ones, e.g. pip&setuptools
+            # And we want to upgrade them cleanly, i.e. uninstall them
+            'pip_ignore_installed': False,
+            # Python installations must be clean
+            'sanity_pip_check': True,
+        }
+        for key, value in ext_defaults.items():
+            if key not in self.cfg['exts_default_options']:
+                self.cfg['exts_default_options'][key] = value
+
     def patch_step(self, *args, **kwargs):
         """
         Custom patch step for Python:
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/p/python.py
```diff
@@ -128,6 +129,21 @@ def __init__(self, *args, **kwargs):
...
+            # And we want to upgrade them cleanly, i.e. uninstall them
+            'pip_ignore_installed': False,
+            # Python installations must be clean
+            'sanity_pip_check': True,
+        }
+        # Disable templating to update the underlying values
+        with disable_templating(self.cfg):
+            for key, value in ext_defaults.items():
+                if key not in self.cfg['exts_default_options']:
+                    self.cfg['exts_default_options'][key] = value
```
boegel, line 144: You could simplify this a bit by using `get_ref`, you shouldn't need `disable_templating` anymore then:

```python
exts_default_options = self.cfg.get_ref('exts_default_options')
for key in ext_defaults:
    if key not in exts_default_options:
        exts_default_options[key] = ext_defaults[key]
```

easybuild/easyblocks/p/python.py
```diff
@@ -128,6 +129,21 @@ def __init__(self, *args, **kwargs):
...
+        self.log.info("exts_default_options: %s", self.cfg['exts_default_options'])
```
boegel, line 145: This should be `debug` unless you make it a more meaningful log message imho

easybuild/easyblocks/p/python.py
```diff
@@ -366,9 +398,14 @@ def install_step(self):
 
         super(EB_Python, self).install_step()
 
+        # Create non-versioned symlinks for python and pip
```
boegel, line 401:
```suggestion
        # Create non-versioned relative symlinks for python and pip
```

easybuild/easyblocks/p/python.py
```diff
@@ -366,9 +398,14 @@ def install_step(self):
...
         python_binary_path = os.path.join(self.installdir, 'bin', 'python')
         if not os.path.isfile(python_binary_path):
-            symlink(python_binary_path + self.pyshortver, python_binary_path)
+            symlink('python' + self.pyshortver, python_binary_path, use_abspath_source=False)
+        if self._has_ensure_pip():
```
boegel, line 405: Maybe make the check a bit more direct: verify that the `pip<version>` binary you're symlinking to exists?

boegel (changes requested):

easybuild/easyblocks/p/python.py
Flamefire: Could do, but this is actually done in the sanity check already. So same as assuming `python<version>` exists, we can assume the `pip<version>` exists. During installation I'd say we can make such assumptions as they are checked later, so no problem here (symlinking to non-existing files is ok)

easybuild/easyblocks/p/python.py
Flamefire: Copied from PythonBundle, but downgraded as suggested as that makes sense

easybuild/easyblocks/p/python.py
Flamefire: Good idea, done

easybuild/easyblocks/p/python.py
boegel: Yeah, makes sense, it was just a small suggestion.

boegel:
@Flamefire These changes seem to at least break `Python-2.7.16-GCCcore-8.3.0.eb`, perhaps more.

I also have a test report coming up from `generoso` with a reinstall of all installed `Python` modules we have there...

boegel:
Test report by @boegel
 * **FAIL (build issue)** _Python-2.7.16-GCCcore-8.3.0.eb_ (partial log available at https://gist.github.com/caacc4ac0082be04b8f8bc462bcb49f3)

node3143.skitty.os - Linux centos linux 7.9.2009, x86_64, Intel(R) Xeon(R) Gold 6140 CPU @ 2.30GHz, Python 3.6.8
See https://gist.github.com/fb48f60089595c4639fac9a3db5701d6 for a full test report.

Flamefire:
I may have an explanation here: That EasyConfig installs setuptools without pip. This may lead to the same issue as with pip itself: It will install the new setuptools without removing the old one and then the newer pip might use the older setuptools which leads to that issue.
Going to double-check.

Reading https://github.com/pypa/pip/issues/6264 and your comments there suggests that indeed the older setuptools (installed by the bootstrap/ensurepip) are used

Not sure how this can be solved if the EC tries to install pip/setuptools without pip while Python has already a pip and setuptools.

I'm running out of ideas how to fix the issues AND have it backward compat. E.g. current ECs install pip without pip but use pip for setuptools: https://github.com/easybuilders/easybuild-easyconfigs/blob/91e0e3809cc56ae6e95db8cef902093d2f806be0/easybuild/easyconfigs/p/Python/Python-3.8.2-GCCcore-9.3.0.eb#L53
This breaks pip in a similar way as it breaks setuptools now (your issue) as it may use the older pip, not the newer one.

So either we use ensurepip in Python base and use pip to install/upgrade BOTH setuptools and pip, or don't use ensurepip and install setuptools/pip via setup.py. Everything else is trouble... And we can't have that without breaking something

Flamefire:
Test report by @Flamefire
 * **SUCCESS** _Python-2.7.14-foss-2018a.eb_ 
 * **FAIL (build issue)** _Python-2.7.15-GCCcore-8.2.0.eb_ (partial log available at https://gist.github.com/73bfbb56fd62e82020cad92542c5b9b0)
 * **SUCCESS** _Python-2.7.18-GCCcore-10.2.0.eb_ 
 * **SUCCESS** _Python-3.6.4-foss-2018a.eb_ 
 * **SUCCESS** _Python-3.7.2-GCCcore-8.2.0.eb_ 
 * **SUCCESS** _Python-3.8.6-GCCcore-10.2.0.eb_ 
 * **SUCCESS** _bzip2-1.0.6-GCCcore-6.3.0.eb_ 
 * **SUCCESS** _libreadline-7.0-GCCcore-6.3.0.eb_ 
 * **SUCCESS** _Tcl-8.6.6-GCCcore-6.3.0.eb_ 
 * **SUCCESS** _flex-2.5.39.eb_ 
 * **SUCCESS** _SQLite-3.17.0-GCCcore-6.3.0.eb_ 
 * **SUCCESS** _binutils-2.25.eb_ 
 * **SUCCESS** _GCCcore-4.9.3.eb_ 
 * **SUCCESS** _M4-1.4.17-GCCcore-4.9.3.eb_ 
 * **SUCCESS** _flex-2.5.39-GCCcore-4.9.3.eb_ 
 * **SUCCESS** _zlib-1.2.8-GCCcore-4.9.3.eb_ 
 * **SUCCESS** _libffi-3.2.1-GCCcore-6.3.0.eb_ 
 * **SUCCESS** _XZ-5.2.3-GCCcore-6.3.0.eb_ 
 * **SUCCESS** _Bison-3.0.4-GCCcore-4.9.3.eb_ 
 * **SUCCESS** _binutils-2.25-GCCcore-4.9.3.eb_ 
 * **SUCCESS** _ncurses-6.0-GCCcore-4.9.3.eb_ 
 * **SUCCESS** _bzip2-1.0.6-GCCcore-4.9.3.eb_ 
 * **SUCCESS** _libreadline-6.3-GCCcore-4.9.3.eb_ 
 * **SUCCESS** _Autoconf-2.69-GCCcore-6.3.0.eb_ 
 * **SUCCESS** _GCC-4.9.3-2.25.eb_ 
 * **SUCCESS** _Automake-1.15-GCCcore-6.3.0.eb_ 
 * **SUCCESS** _GCC-6.3.0-2.27.eb_ 
 * **SUCCESS** _Tcl-8.6.4-GCC-4.9.3-2.25.eb_ 
 * **SUCCESS** _SQLite-3.9.2-GCC-4.9.3-2.25.eb_ 
 * **SUCCESS** _Tk-8.6.4-GCC-4.9.3-2.25-no-X11.eb_ 
 * **SUCCESS** _libtool-2.4.6-GCCcore-6.3.0.eb_ 
 * **SUCCESS** _Autotools-20150215-GCCcore-6.3.0.eb_ 
 * **SUCCESS** _GMP-6.1.2-GCCcore-6.3.0.eb_ 
 * **FAIL (build issue)** _OpenBLAS-0.2.19-GCC-6.3.0-2.27-LAPACK-3.7.0.eb_ (partial log available at https://gist.github.com/69e2db7f7a14e42d4e8515d6b683bbe9)
 * **SUCCESS** _Autoconf-2.69-GCCcore-4.9.3.eb_ 
 * **SUCCESS** _OpenBLAS-0.2.15-GCC-4.9.3-2.25-LAPACK-3.6.0.eb_ 
 * **SUCCESS** _Automake-1.15-GCCcore-4.9.3.eb_ 
 * **SUCCESS** _numactl-2.0.11-GCCcore-6.3.0.eb_ 
 * **SUCCESS** _hwloc-1.11.5-GCC-6.3.0-2.27.eb_ 
 * **FAIL (build issue)** _OpenMPI-2.0.2-GCC-6.3.0-2.27.eb_ (partial log available at https://gist.github.com/6e919f98290a86d14df2e3e072a3e6c3)
 * **SUCCESS** _libtool-2.4.6-GCCcore-4.9.3.eb_ 
 * **FAIL (build issue)** _gompi-2017a.eb_ (partial log available at https://gist.github.com/b91b34b573e7a0d95a372f1a64d9d684)
 * **SUCCESS** _Autotools-20150215-GCCcore-4.9.3.eb_ 
 * **FAIL (build issue)** _FFTW-3.3.6-gompi-2017a.eb_ (partial log available at https://gist.github.com/ac79eed625cef44eb0195b1250e40490)
 * **FAIL (build issue)** _ScaLAPACK-2.0.2-gompi-2017a-OpenBLAS-0.2.19-LAPACK-3.7.0.eb_ (partial log available at https://gist.github.com/e685cc27fafc6e1fb561e30e591e17ca)
 * **SUCCESS** _numactl-2.0.11-GCCcore-4.9.3.eb_ 
 * **FAIL (build issue)** _foss-2017a.eb_ (partial log available at https://gist.github.com/bb35ac05a1c8a779830584e81bd25090)
 * **SUCCESS** _GMP-6.1.0-GCC-4.9.3-2.25.eb_ 
 * **SUCCESS** _XZ-5.2.2-GCCcore-4.9.3.eb_ 
 * **SUCCESS** _hwloc-1.11.2-GCC-4.9.3-2.25.eb_ 
 * **FAIL (build issue)** _Python-2.7.13-foss-2017a.eb_ (partial log available at https://gist.github.com/4799010053a6e36a339210b83b0158cb)
 * **FAIL (build issue)** _Python-3.6.1-foss-2017a.eb_ (partial log available at https://gist.github.com/20c97db0a45a1632b825a657487b6bf4)
 * **FAIL (build issue)** _OpenMPI-1.10.2-GCC-4.9.3-2.25.eb_ (partial log available at https://gist.github.com/0cc90346b6d8a85558c12dcdd6a9d32c)
 * **FAIL (build issue)** _gompi-2016a.eb_ (partial log available at https://gist.github.com/3d8bcab749d2ce4409b98c2228ecd9d9)
 * **FAIL (build issue)** _FFTW-3.3.4-gompi-2016a.eb_ (partial log available at https://gist.github.com/168de8961fcc5101b2241c45892edb06)
 * **FAIL (build issue)** _ScaLAPACK-2.0.2-gompi-2016a-OpenBLAS-0.2.15-LAPACK-3.6.0.eb_ (partial log available at https://gist.github.com/d3146f7f95763e22c8f140a38ccf8265)
 * **FAIL (build issue)** _foss-2016a.eb_ (partial log available at https://gist.github.com/466389ddf9e48c33f7732b93243d6ffa)
 * **FAIL (build issue)** _Python-2.7.11-foss-2016a.eb_ (partial log available at https://gist.github.com/b856fc0fd07484598caf547737575a2d)
 * **FAIL (build issue)** _Python-3.5.1-foss-2016a.eb_ (partial log available at https://gist.github.com/518cc8ca11837da603f2facd4b78f87b)

Build succeeded for 43 out of 59 (10 easyconfigs in total)
taurusi6496.taurus.hrsk.tu-dresden.de - Linux RHEL 7.9, x86_64, Intel(R) Xeon(R) CPU E5-2680 v3 @ 2.50GHz (haswell), Python 2.7.5
See https://gist.github.com/e92376adb57bbbe858dc2c1fd8a961e4 for a full test report.

Flamefire:
Test report by @Flamefire
 * **SUCCESS** _Python-2.7.14-foss-2018a.eb_ 
 * **FAIL (build issue)** _Python-2.7.15-GCCcore-8.2.0.eb_ (partial log available at https://gist.github.com/bcfbbdd8e17f06ba982d55a144511f1d)
 * **SUCCESS** _Python-2.7.18-GCCcore-10.2.0.eb_ 
 * **SUCCESS** _Python-3.6.4-foss-2018a.eb_ 
 * **SUCCESS** _Python-3.7.2-GCCcore-8.2.0.eb_ 
 * **SUCCESS** _Python-3.8.6-GCCcore-10.2.0.eb_ 
 * **SUCCESS** _bzip2-1.0.6-GCCcore-6.3.0.eb_ 
 * **SUCCESS** _libreadline-7.0-GCCcore-6.3.0.eb_ 
 * **SUCCESS** _Tcl-8.6.6-GCCcore-6.3.0.eb_ 
 * **SUCCESS** _flex-2.5.39.eb_ 
 * **SUCCESS** _SQLite-3.17.0-GCCcore-6.3.0.eb_ 
 * **SUCCESS** _binutils-2.25.eb_ 
 * **SUCCESS** _GCCcore-4.9.3.eb_ 
 * **SUCCESS** _M4-1.4.17-GCCcore-4.9.3.eb_ 
 * **SUCCESS** _flex-2.5.39-GCCcore-4.9.3.eb_ 
 * **SUCCESS** _zlib-1.2.8-GCCcore-4.9.3.eb_ 
 * **SUCCESS** _libffi-3.2.1-GCCcore-6.3.0.eb_ 
 * **SUCCESS** _XZ-5.2.3-GCCcore-6.3.0.eb_ 
 * **SUCCESS** _Bison-3.0.4-GCCcore-4.9.3.eb_ 
 * **SUCCESS** _binutils-2.25-GCCcore-4.9.3.eb_ 
 * **SUCCESS** _ncurses-6.0-GCCcore-4.9.3.eb_ 
 * **SUCCESS** _bzip2-1.0.6-GCCcore-4.9.3.eb_ 
 * **SUCCESS** _libreadline-6.3-GCCcore-4.9.3.eb_ 
 * **SUCCESS** _Autoconf-2.69-GCCcore-6.3.0.eb_ 
 * **SUCCESS** _GCC-4.9.3-2.25.eb_ 
 * **SUCCESS** _Automake-1.15-GCCcore-6.3.0.eb_ 
 * **SUCCESS** _GCC-6.3.0-2.27.eb_ 
 * **SUCCESS** _Tcl-8.6.4-GCC-4.9.3-2.25.eb_ 
 * **SUCCESS** _SQLite-3.9.2-GCC-4.9.3-2.25.eb_ 
 * **SUCCESS** _Tk-8.6.4-GCC-4.9.3-2.25-no-X11.eb_ 
 * **SUCCESS** _libtool-2.4.6-GCCcore-6.3.0.eb_ 
 * **SUCCESS** _Autotools-20150215-GCCcore-6.3.0.eb_ 
 * **SUCCESS** _GMP-6.1.2-GCCcore-6.3.0.eb_ 
 * **SUCCESS** _OpenBLAS-0.2.19-GCC-6.3.0-2.27-LAPACK-3.7.0.eb_ 
 * **SUCCESS** _Autoconf-2.69-GCCcore-4.9.3.eb_ 
 * **SUCCESS** _OpenBLAS-0.2.15-GCC-4.9.3-2.25-LAPACK-3.6.0.eb_ 
 * **SUCCESS** _Automake-1.15-GCCcore-4.9.3.eb_ 
 * **SUCCESS** _numactl-2.0.11-GCCcore-6.3.0.eb_ 
 * **SUCCESS** _hwloc-1.11.5-GCC-6.3.0-2.27.eb_ 
 * **FAIL (build issue)** _OpenMPI-2.0.2-GCC-6.3.0-2.27.eb_ (partial log available at https://gist.github.com/b12c0182ba637c0eda347568315af6a8)
 * **SUCCESS** _libtool-2.4.6-GCCcore-4.9.3.eb_ 
 * **FAIL (build issue)** _gompi-2017a.eb_ (partial log available at https://gist.github.com/3246f29db6b81a744879e0060baa7e44)
 * **SUCCESS** _Autotools-20150215-GCCcore-4.9.3.eb_ 
 * **FAIL (build issue)** _FFTW-3.3.6-gompi-2017a.eb_ (partial log available at https://gist.github.com/30ac04791973bbd2f14ca49a570bbac8)
 * **FAIL (build issue)** _ScaLAPACK-2.0.2-gompi-2017a-OpenBLAS-0.2.19-LAPACK-3.7.0.eb_ (partial log available at https://gist.github.com/f548f4657624e06c9d8803dd83620df5)
 * **SUCCESS** _numactl-2.0.11-GCCcore-4.9.3.eb_ 
 * **FAIL (build issue)** _foss-2017a.eb_ (partial log available at https://gist.github.com/7328744c83bef5b816c8edbd9e75f7f4)
 * **SUCCESS** _GMP-6.1.0-GCC-4.9.3-2.25.eb_ 
 * **SUCCESS** _XZ-5.2.2-GCCcore-4.9.3.eb_ 
 * **SUCCESS** _hwloc-1.11.2-GCC-4.9.3-2.25.eb_ 
 * **FAIL (build issue)** _Python-2.7.13-foss-2017a.eb_ (partial log available at https://gist.github.com/1b20520ccd78633acd38fc561c2054a0)
 * **FAIL (build issue)** _Python-3.6.1-foss-2017a.eb_ (partial log available at https://gist.github.com/5934441df92d51f1663c5d9d62bd898e)
 * **FAIL (build issue)** _OpenMPI-1.10.2-GCC-4.9.3-2.25.eb_ (partial log available at https://gist.github.com/f8fbf79b4ddbc1f1dbd9380e6657e9ef)
 * **FAIL (build issue)** _gompi-2016a.eb_ (partial log available at https://gist.github.com/c0dea66da24437538e4fd92d3afe6c3b)
 * **FAIL (build issue)** _FFTW-3.3.4-gompi-2016a.eb_ (partial log available at https://gist.github.com/d8db87f38026a6b37772f5bbcb5e4f7e)
 * **FAIL (build issue)** _ScaLAPACK-2.0.2-gompi-2016a-OpenBLAS-0.2.15-LAPACK-3.6.0.eb_ (partial log available at https://gist.github.com/ad6cb698e98a5d19adb97373eadd8152)
 * **FAIL (build issue)** _foss-2016a.eb_ (partial log available at https://gist.github.com/817117ac53902a5714281de5462dfcaa)
 * **FAIL (build issue)** _Python-2.7.11-foss-2016a.eb_ (partial log available at https://gist.github.com/ed5fa0b0da7b733ac11e6218cfa16a09)
 * **FAIL (build issue)** _Python-3.5.1-foss-2016a.eb_ (partial log available at https://gist.github.com/438279be0b2bb2c93e06c4acff5411e1)

Build succeeded for 44 out of 59 (10 easyconfigs in total)
taurusi6601.taurus.hrsk.tu-dresden.de - Linux RHEL 7.9, x86_64, Intel(R) Xeon(R) CPU E5-2680 v3 @ 2.50GHz (haswell), Python 2.7.5
See https://gist.github.com/72e16c09db2d3646668cfbca7cb88f5c for a full test report.

boegel:
@Flamefire We're fixing bugs here, which means things don't have to be fully backwards-compatible. If some of the existing easyconfigs we have centrally need changes in combination with the changes being made here to ensure they still work, then so be it.

Flamefire:
Ok then my suggestion would be:
- Add an option to install pip (via ensurepip) which is enabled by default for Python 3.4+, i.e. not for Python 2 to avoid having to change countless ECs. While not using pip there to install pip may be suboptimal it seemingly works
- When that option is enabled, check that use_pip for pip and setuptools extensions is set to true, if not either force set it or error out
- Additionally move the "wheel" package to the front

This means for older pythons (e.g. Python 2) setuptools&pip will still be installed as eggs, which has worked so far. Real issues occur only when we want to update those 2 which was not (AFAIK) the case, until we used ensurepip. We can update recentish Python 2 ECs to set this option

Installing/upgrading setuptools via pip was likely never a problem (checked some versions, and if we used ensurepip of course), but installing pip via pip requires wheel for newer versions of pip (I think 20+ according to some quick experiments)
My guess is (assuming sane Python devs), that `wheel` can always be installed with any version of setuptools and pip, hence we can do that first.
Newer versions of pip use a new feature of setuptools, e.g. the `__legacy__` thing.
Hence I think it makes sense to either enforce (i.e. error out) or ensure (i.e. reorder exts) to install: wheels, setuptools, pip, in this order, when we have used ensurepip.

The error option would mean we need to fix a lot of ECs.

So maybe add that option but make it off by default. Then we only need to fix ECs which we have changed to rely on pip being there, which are far less, and maybe backport that for a few older ones (2019a+ I'd say)

--> Add option `use_ensurepip` (better name idea?), off by default, and if set check if `use_pip=True` is used for pip&setuptools and error out if not the case

I'll be back for further discussion via slack on Monday if required.

Flamefire:
Test report by @Flamefire
 * **SUCCESS** _Python-2.7.14-foss-2018a.eb_ 
 * **SUCCESS** _Python-2.7.15-GCCcore-8.2.0.eb_ 
 * **SUCCESS** _Python-2.7.18-GCCcore-10.2.0.eb_ 
 * **SUCCESS** _Python-3.6.4-foss-2018a.eb_ 
 * **SUCCESS** _Python-3.7.2-GCCcore-8.2.0.eb_ 
 * **SUCCESS** _bzip2-1.0.6-GCCcore-4.9.3.eb_ 
 * **SUCCESS** _bzip2-1.0.6-GCCcore-6.3.0.eb_ 
 * **SUCCESS** _libreadline-7.0-GCCcore-6.3.0.eb_ 
 * **SUCCESS** _ncurses-6.0-GCCcore-4.9.3.eb_ 
 * **SUCCESS** _libreadline-6.3-GCCcore-4.9.3.eb_ 
 * **SUCCESS** _Tcl-8.6.6-GCCcore-6.3.0.eb_ 
 * **SUCCESS** _SQLite-3.17.0-GCCcore-6.3.0.eb_ 
 * **SUCCESS** _libffi-3.2.1-GCCcore-6.3.0.eb_ 
 * **SUCCESS** _XZ-5.2.3-GCCcore-6.3.0.eb_ 
 * **SUCCESS** _Tcl-8.6.4-GCC-4.9.3-2.25.eb_ 
 * **SUCCESS** _SQLite-3.9.2-GCC-4.9.3-2.25.eb_ 
 * **SUCCESS** _Tk-8.6.4-GCC-4.9.3-2.25-no-X11.eb_ 
 * **SUCCESS** _GMP-6.1.0-GCC-4.9.3-2.25.eb_ 
 * **SUCCESS** _Autoconf-2.69-GCCcore-6.3.0.eb_ 
 * **FAIL (build issue)** _Python-2.7.11-foss-2016a.eb_ (partial log available at https://gist.github.com/b4fbc01feab09cf6d2ad1cc390a9dda7)
 * **SUCCESS** _XZ-5.2.2-GCCcore-4.9.3.eb_ 
 * **SUCCESS** _Automake-1.15-GCCcore-6.3.0.eb_ 
 * **SUCCESS** _GCC-6.3.0-2.27.eb_ 
 * **SUCCESS** _Python-3.5.1-foss-2016a.eb_ 
 * **SUCCESS** _libtool-2.4.6-GCCcore-6.3.0.eb_ 
 * **SUCCESS** _Autotools-20150215-GCCcore-6.3.0.eb_ 
 * **SUCCESS** _GMP-6.1.2-GCCcore-6.3.0.eb_ 
 * **SUCCESS** _OpenBLAS-0.2.19-GCC-6.3.0-2.27-LAPACK-3.7.0.eb_ 
 * **SUCCESS** _numactl-2.0.11-GCCcore-6.3.0.eb_ 
 * **SUCCESS** _hwloc-1.11.5-GCC-6.3.0-2.27.eb_ 
 * **SUCCESS** _OpenMPI-2.0.2-GCC-6.3.0-2.27.eb_ 
 * **SUCCESS** _gompi-2017a.eb_ 
 * **FAIL (build issue)** _FFTW-3.3.6-gompi-2017a.eb_ (partial log available at https://gist.github.com/beb197d53c9e7fee28122a955ec880e1)
 * **SUCCESS** _ScaLAPACK-2.0.2-gompi-2017a-OpenBLAS-0.2.19-LAPACK-3.7.0.eb_ 
 * **FAIL (build issue)** _foss-2017a.eb_ (partial log available at https://gist.github.com/66991e5f118c840863c4ca2c1fcbbe52)
 * **FAIL (build issue)** _Python-2.7.13-foss-2017a.eb_ (partial log available at https://gist.github.com/eca1f4d4a7a5066398a8ee5b551bdbb0)
 * **FAIL (build issue)** _Python-3.6.1-foss-2017a.eb_ (partial log available at https://gist.github.com/4709d3572be16949b8d9105621796ec9)

Build succeeded for 32 out of 37 (9 easyconfigs in total)
taurusi6327.taurus.hrsk.tu-dresden.de - Linux RHEL 7.9, x86_64, Intel(R) Xeon(R) CPU E5-2680 v3 @ 2.50GHz (haswell), Python 2.7.5
See https://gist.github.com/bc92ce419dc03616cc43ba384ea86683 for a full test report.

easybuild/easyblocks/p/python.py
```diff
@@ -204,6 +235,19 @@ def prepare_for_extensions(self):
                 self.log.debug(msg, param, self.cfg[param])
             self.cfg[param] = ''
 
+        if self.install_pip:
+            # When using ensurepip, then pip must be used to upgrade pip and setuptools
+            # Otherwise it will only copy new files leading to a combination of of files from the old and new version
```
ocaisa, line 240:
```suggestion
            # Otherwise it will only copy new files leading to a combination of files from the old and new version
```

Flamefire:
Enabling the default sanity_pip_check can only be done if we also install a recent pip (i.e. >= 9). I added https://github.com/easybuilders/easybuild-easyblocks/pull/2388/commits/6c7a1901607451186eacf638b870ff1693091832 for that.

Flamefire:
Enabling the download_dep_fail option by default does make sense to me but it shows "bugs" in older ECs, e.g. Python-2.7.13-foss-2017a.eb does not include ipaddress but that is a dependency for cryptography. Due to enabling the "disconnected" option of pip if "download_dep_fail" is set, with this change it will NOT download ipaddress from PyPi anymore and fail to install the EC.

@boegel I can PR the fix to the EC and keep the `download_dep_fail = True` default here. Or shall we wait with changing this default until EB 5?

Flamefire:
Looks like fixing the EC(s) is more work and likely some more are also affected. I choose to not set download_dep_fail for "old" Python versions. The remaining ones should already set that in the EC so it's safe there, I guess. Could still remove this and wait for EB 5.x with that change

Flamefire:
Test report by @Flamefire
 * **SUCCESS** _Python-2.7.11-foss-2016a.eb_ 
 * **SUCCESS** _Python-2.7.14-foss-2018a.eb_ 
 * **SUCCESS** _Python-2.7.15-GCCcore-8.2.0.eb_ 
 * **SUCCESS** _Python-2.7.18-GCCcore-10.2.0.eb_ 
 * **SUCCESS** _Python-3.6.4-foss-2018a.eb_ 
 * **SUCCESS** _Python-3.7.2-GCCcore-8.2.0.eb_ 
 * **SUCCESS** _bzip2-1.0.6-GCCcore-6.3.0.eb_ 
 * **SUCCESS** _XZ-5.2.2-GCCcore-4.9.3.eb_ 
 * **SUCCESS** _Python-3.5.1-foss-2016a.eb_ 
 * **SUCCESS** _libreadline-7.0-GCCcore-6.3.0.eb_ 
 * **SUCCESS** _GMP-6.1.2-GCCcore-6.3.0.eb_ 
 * **SUCCESS** _Tcl-8.6.6-GCCcore-6.3.0.eb_ 
 * **SUCCESS** _SQLite-3.17.0-GCCcore-6.3.0.eb_ 
 * **SUCCESS** _libffi-3.2.1-GCCcore-6.3.0.eb_ 
 * **SUCCESS** _XZ-5.2.3-GCCcore-6.3.0.eb_ 
 * **FAIL (build issue)** _Python-2.7.13-foss-2017a.eb_ (partial log available at https://gist.github.com/6f8088abebdf19c6077156e61873ec8f)
 * **FAIL (build issue)** _Python-3.6.1-foss-2017a.eb_ (partial log available at https://gist.github.com/e6df4900f4d4929bf3241740a3e37cd2)

Build succeeded for 15 out of 17 (9 easyconfigs in total)
taurusi6216.taurus.hrsk.tu-dresden.de - Linux RHEL 7.9, x86_64, Intel(R) Xeon(R) CPU E5-2680 v3 @ 2.50GHz (haswell), Python 2.7.5
See https://gist.github.com/d0a47e552f431e5185a2c426cca74961 for a full test report.

Flamefire:
Ok, it seems the old ECs can no longer be installed. I tested the failing 2 Python ECs without this patch and e.g. Python/3.6.1-foss-2017a/ does fail too with:

```
Processing dependencies for cffi==1.11.5
Searching for pycparser
Reading https://pypi.python.org/simple/pycparser/
Couldn't find index page for 'pycparser' (maybe misspelled?)
Scanning index of all packages (this may take a while)
Reading https://pypi.python.org/simple/
No local packages or working download links found for pycparser
error: Could not find suitable distribution for Requirement.parse('pycparser')
```

So it seems we missed a required package and PyPi has changed something so old pips can't use it anymore.
That would be a reason to enable download_dep_fail by default unconditionally to "weed out" such ECs and fix them if anyone still uses them.

But I'd also be fine with leaving this as-is and get this in rather sooner than later

boegel:
@Flamefire That failure is not related to the changes made in #2390, right?

Flamefire:
@boegel Yes they are not related. The failing command:
> == 2021-05-04 17:11:09,106 easyblock.py:3427 WARNING build failed (first 300 chars): cmd " /tmp/ebinstall/software/Python/3.6.1-foss-2017a/bin/python setup.py install --prefix=/tmp/ebinstall/software/Python/3.6.1-foss-2017a " exited with exit code 1 and output:

I.e. it is using setup.py not pip, hence this is unaffected by the change in #2390 which only affects pip

easybuild/easyblocks/p/python.py
```diff
@@ -128,6 +132,39 @@ def __init__(self, *args, **kwargs):
...
+            # And we want to upgrade them cleanly, i.e. uninstall them
+            'pip_ignore_installed': False,
+            # Python installations must be clean. Requires pip >= 9
+            'sanity_pip_check': LooseVersion(self._get_pip_ext_version() or '0') >= LooseVersion('9'),
+        }
+        # Some older ECs break if we enable download_dep_fail due to missing packages
+        # Since Python 2.7.14 (and all Python 3 versions) we usually enable download_dep_fail in the ECs
+        # already so this should not lead to failures
+        if LooseVersion(self.version) >= LooseVersion('2.7.14'):
+            ext_defaults['download_dep_fail'] = True
```
boegel, line 146: @Flamefire I don't think we should automatically enable this at this point, that's a rather intrusive change in the default behavior, which we should only make in EasyBuild v5.0 (cfr. https://github.com/easybuilders/easybuild-easyblocks/issues/2127).

I fully agree with you that we should do this, which is why the easyconfig test suite has been requiring this for a while now, and I'm looking forward to not having to worry about this anymore and enabling it by default, but we shouldn't for the time being.

We're not enabling `download_dep_fail` in the oldest Python 3 easyconfigs (and it's not worth spending time in trying to enable that there). Those easyconfigs are old enough that they'll be archived in EasyBuild v5.0, so they won't block enabling `download_dep_fail` by default.

If I understand correctly, just removing this part won't affect the rest of this PR, correct me if I'm wrong...

easybuild/easyblocks/p/python.py
```diff
@@ -128,6 +132,39 @@ def __init__(self, *args, **kwargs):
             easybuild_subdir = log_path()
             self.pythonpath = os.path.join(easybuild_subdir, 'python')
 
+        ext_defaults = {
+            # The only installed packages at this point are default installed ones, e.g. pip&setuptools
+            # And we want to upgrade them cleanly, i.e. uninstall them
+            'pip_ignore_installed': False,
+            # Python installations must be clean. Requires pip >= 9
+            'sanity_pip_check': LooseVersion(self._get_pip_ext_version() or '0') >= LooseVersion('9'),
```
boegel, line 140: This comment should be clarified a bit I think, and I would use `9.0` rather than `9` (just for clarity sake):

```suggestion
            # enable "pip check" in sanity check by default, to ensure that Python installations are clean;
            # requires pip v9.0 or newer
            'sanity_pip_check': LooseVersion(self._get_pip_ext_version() or '0') >= LooseVersion('9.0'),
```

That said, I don't think we should be enabling `sanity_pip_check` by default just yet. It'll result in broken `Python` easyconfigs that worked just fine up until now, and changes that break backwards compability should only be made in the next major EasyBuild release (v5.0), cfr. https://github.com/easybuilders/easybuild-easyblocks/issues/2127

easybuild/easyblocks/p/python.py
```diff
@@ -128,6 +132,39 @@ def __init__(self, *args, **kwargs):
             easybuild_subdir = log_path()
             self.pythonpath = os.path.join(easybuild_subdir, 'python')
 
+        ext_defaults = {
+            # The only installed packages at this point are default installed ones, e.g. pip&setuptools
+            # And we want to upgrade them cleanly, i.e. uninstall them
+            'pip_ignore_installed': False,
```
boegel, line 138: The intention here is to avoid that `--ignore-installed` is used when updating `pip` and `setuptools`, but you're effectively disabling it for *all* extensions being installed along with Python (unless otherwise specified explicitly in the easyconfig file via `exts_default_options`)?

easybuild/easyblocks/p/python.py
```diff
@@ -110,6 +110,10 @@ def extra_options():
         """Add extra config options specific to Python."""
         extra_vars = {
             'ebpythonprefixes': [True, "Create sitecustomize.py and allow use of $EBPYTHONPREFIXES", CUSTOM],
+            'install_pip': [False,
+                            "Use the ensurepip module (Python 2.7.9+, 3.4+) to install the bundled versions "
+                            "of pip and setuptools into Python. You _must_ then use pip for upgrading pip&setuptools!",
```
boegel, line 115:
```suggestion
                            "of pip and setuptools into Python. You _must_ then use pip for upgrading "
                            "pip & setuptools by installing newer versions as extensions!",

File: easybuild/easyblocks/p/python.py [hunk 10]
```diff
@@ -204,6 +248,19 @@ def prepare_for_extensions(self):
...
+        if self.install_pip:
+            # When using ensurepip, then pip must be used to upgrade pip and setuptools
+            # Otherwise it will only copy new files leading to a combination of files from the old and new version
+            use_pip_default = self.cfg['exts_default_options'].get('use_pip')
+            # self.exts is populated in fetch_step
+            for ext in self.exts:
+                if ext['name'] in ('pip', 'setuptools'):
+                    if not ext.get('options', {}).get('use_pip', use_pip_default):
+                        raise EasyBuildError("When using ensurepip to install pip (requested by install_pip=True) "
+                                             "you must set 'use_pip=True' for pip&setuptools. "
```
boegel, line 260:
```suggestion
                                             "you must set 'use_pip=True' for the pip & setuptools extensions. "

boegel (changes requested):

easybuild/easyblocks/p/python.py
Flamefire: Postponed to EB 5.x where this will be the default in the PythonPackage EasyBlock

easybuild/easyblocks/p/python.py
Flamefire: Agreed. Removed and will open a follow-up PR for this

easybuild/easyblocks/p/python.py
Flamefire: Enhanced the comment as suggested. 
I found another reason for that: Similar for pip+setuptools we MUST actually use this for any other extension that we install twice (for whatever reason) or we will have those Frankenstein-software again with half-updated files.
This could happen by accident, e.g. auto-installing a dependency of an extension and then the same package as an extension, or intentionally by e.g. installing an older version first and upgrading it later (e.g. when other packages were installed which are required for the newer version)

boegel: OK, thanks for clarifying the comment!

easybuild/easyblocks/p/python.py
Flamefire: Follow-up: https://github.com/easybuilders/easybuild-easyblocks/pull/2423

boegel:
Test report by @boegel
#### Overview of tested easyconfigs (in order)
 * **SUCCESS** _Python-2.7.10-GCC-4.9.3-2.25-bare.eb_ 
 * **SUCCESS** _Python-2.7.10-GNU-4.9.3-2.25-bare.eb_ 
 * **SUCCESS** _Python-2.7.11-foss-2016a.eb_ 
 * **SUCCESS** _Python-2.7.11-GCC-4.9.3-2.25-bare.eb_ 
 * **SUCCESS** _Python-2.7.11-intel-2016.02-GCC-4.9.eb_ 
 * **SUCCESS** _Python-2.7.11-intel-2016a.eb_ 
 * **SUCCESS** _Python-2.7.12-GCC-5.4.0-2.26-bare.eb_ 
 * **SUCCESS** _Python-2.7.12-GCCcore-4.9.3-bare.eb_ 
 * **SUCCESS** _Python-2.7.13-GCCcore-6.3.0-bare.eb_ 
 * **SUCCESS** _Python-2.7.14-foss-2017b.eb_ 
 * **SUCCESS** _Python-2.7.14-foss-2018a.eb_ 
 * **SUCCESS** _Python-2.7.14-fosscuda-2017b.eb_ 
 * **SUCCESS** _Python-2.7.14-fosscuda-2018a.eb_ 
 * **SUCCESS** _Python-2.7.14-GCCcore-6.4.0-bare.eb_ 
 * **SUCCESS** _Python-2.7.15-foss-2018b.eb_ 
 * **SUCCESS** _Python-2.7.15-fosscuda-2018b.eb_ 
 * **SUCCESS** _Python-2.7.15-GCCcore-7.2.0-bare.eb_ 
 * **SUCCESS** _Python-2.7.15-GCCcore-7.3.0-bare.eb_ 
 * **SUCCESS** _Python-2.7.15-GCCcore-8.2.0.eb_ 
 * **SUCCESS** _Python-2.7.15-intel-2018b.eb_ 
 * **SUCCESS** _Python-2.7.15-iomkl-2018b.eb_ 
 * **SUCCESS** _Python-2.7.16-GCCcore-8.3.0.eb_ 
 * **SUCCESS** _Python-2.7.18-GCCcore-10.2.0.eb_ 
 * **SUCCESS** _Python-2.7.18-GCCcore-9.3.0.eb_ 
 * **SUCCESS** _Python-2.7.9-GCC-4.8.4-bare.eb_ 
 * **SUCCESS** _Python-2.7.9-GCC-4.9.2-bare.eb_ 
 * **SUCCESS** _Python-3.5.1-foss-2016a.eb_ 
 * **FAIL (build issue)** _Python-3.5.1-intel-2016a.eb_ (partial log available at https://gist.github.com/b69cdc5ac6cccbb46b1d0069faab2aee)
 * **SUCCESS** _Python-3.6.2-foss-2017b.eb_ 
 * **FAIL (build issue)** _Python-3.6.3-foss-2017b.eb_ (partial log available at https://gist.github.com/58b19faa19690eda3e9838561f8e9902)
 * **SUCCESS** _Python-3.6.4-foss-2017a.eb_ 
 * **SUCCESS** _Python-3.6.4-foss-2018a.eb_ 
 * **SUCCESS** _Python-3.6.4-fosscuda-2018a.eb_ 
 * **SUCCESS** _Python-3.6.4-golf-2018a.eb_ 
 * **SUCCESS** _Python-3.6.4-iomkl-2018.02.eb_ 
 * **SUCCESS** _Python-3.6.6-foss-2018b.eb_ 
 * **SUCCESS** _Python-3.6.6-fosscuda-2018b.eb_ 
 * **SUCCESS** _Python-3.6.6-intel-2018b.eb_ 
 * **SUCCESS** _Python-3.6.6-iomkl-2018b.eb_ 
 * **SUCCESS** _Python-3.7.0-foss-2018b.eb_ 
 * **SUCCESS** _Python-3.7.0-intel-2018b.eb_ 
 * **SUCCESS** _Python-3.7.0-iomkl-2018b.eb_ 
 * **SUCCESS** _Python-3.7.2-GCCcore-8.2.0.eb_ 
 * **SUCCESS** _Python-3.7.4-GCCcore-8.3.0.eb_ 
 * **SUCCESS** _Python-3.8.2-GCCcore-9.3.0.eb_ 
 * **SUCCESS** _Python-3.8.6-GCCcore-10.2.0.eb_ 

Build succeeded for 44 out of 46 (46 easyconfigs in total)
node2648.swalot.os - Linux centos linux 7.9.2009, x86_64, Intel(R) Xeon(R) CPU E5-2660 v3 @ 2.60GHz (haswell), Python 3.6.8
See https://gist.github.com/ec0738dcc52efe09bf0cea050f846a84 for a full test report.

boegel (approved):
The `Python` easyconfigs that didn't work are due to problems not related to the changes in this PR (they're broken in the same way with the current `Python` easyblock), so I'll go ahead and merge this and follow up on those specific easyconfigs in a separate (easyconfigs) PR (with low priority though, since they're ancient...).

Thanks for your efforts on this @Flamefire!

PR #2429, opened 2021-05-16
add easyblock to install OpenSSL wrapper for OpenSSL installed in OS, or build and install OpenSSL from source if not available in OS

lexming:
Fixes issue https://github.com/easybuilders/easybuild-easyconfigs/issues/11895

This easyblock looks for the installation files of OpenSSL in the host system to wrap them in the EB installation directory with symlinks
* it searches the library in the host system by loading the respective `libssl.so` in Python and retrieve the path to the library file using `dlinfo` in Linux and `ctypes.macholib.dyld.dyld_find` in Mac OS.
* it searches the headers of the required version of OpenSSL  (v1.0.x or v1.1.x) in the include paths of `gcc` in the system.

This method does not need to crawl through any environment variables, is resilient to installations in exotic paths and integrates well within the EB environment.

Why symlinking?
* it provides a transparent way to update OpenSSL without breaking dynamic links and _probably_ without breaking rpaths either. Any packages build with the wrappers in place should use the paths to the installation directory of the wrapper, not the actual libraries in the system or EB.

If the host system lacks the libs or headers of the required version of OpenSSL, the easyblock builds and installs any component included in the easyconfig bundle (`EB_OpenSSL_wrapper` is based on `Bundle`). This approach has the advantage of automatically disabling the installation of a fallback OpenSSL in EB if the host system already has the required files. Moreover, if the fallback is installed, the result is identical to any other installation of OpenSSL in EB.

Easyconfigs can be found in https://github.com/easybuilders/easybuild-easyconfigs/pull/12858

edit: requires ~~https://github.com/easybuilders/easybuild-framework/pull/3682~~ + ~~https://github.com/easybuilders/easybuild-framework/pull/3683~~

### DIFF
--- a/easybuild/easyblocks/o/openssl_wrapper.py
+++ b/easybuild/easyblocks/o/openssl_wrapper.py
@@ -165,7 +165,7 @@ class EB_OpenSSL_wrapper(Bundle):
                               "Falling back to OpenSSL in EasyBuild")
 
         # Check system include paths for OpenSSL headers
-        cmd = "gcc -E -Wp,-v -xc /dev/null"
+        cmd = "LC_ALL=C gcc -E -Wp,-v -xc /dev/null"
         (out, ec) = run_cmd(cmd, log_all=True, simple=False, trace=False)
 
         sys_include_dirs = []
--- a/easybuild/easyblocks/o/openssl_wrapper.py
+++ b/easybuild/easyblocks/o/openssl_wrapper.py
@@ -33,10 +33,9 @@ import os
 import re
 
 from easybuild.easyblocks.generic.bundle import Bundle
-from easybuild.tools.filetools import expand_glob_paths, find_glob_pattern, mkdir, read_file, symlink, which
-from easybuild.tools.modules import get_software_root
+from easybuild.framework.easyconfig import CUSTOM
+from easybuild.tools.filetools import expand_glob_paths, mkdir, read_file, symlink, which
 from easybuild.tools.run import run_cmd
-from easybuild.tools.systemtools import get_shared_lib_ext
 from easybuild.tools.systemtools import DARWIN, LINUX, get_os_type, get_shared_lib_ext
 
 def locate_solib(libobj):
@@ -67,6 +66,15 @@ class EB_OpenSSL_wrapper(Bundle):
     Create a wrapper .modulerc file for OpenSSL
     """
 
+    @staticmethod
+    def extra_options(extra_vars=None):
+        """Easyconfig parameters specific to OpenSSL wrapper"""
+        extra_vars = Bundle.extra_options(extra_vars=extra_vars)
+        extra_vars.update({
+            'wrap_system_openssl': [True, 'Detect and wrap OpenSSL installation in host system', CUSTOM],
+        })
+        return extra_vars
+
     def __init__(self, *args, **kwargs):
         """Define the names of OpenSSL shared objects"""
         super(EB_OpenSSL_wrapper, self).__init__(*args, **kwargs)
@@ -74,7 +82,14 @@ class EB_OpenSSL_wrapper(Bundle):
         # Libraries packaged in OpenSSL
         self.openssl_libs = ['libssl', 'libcrypto']
 
-        # Check the system library of OpenSSL
+        # Paths to system libraries and headers of OpenSSL
+        self.ssl_syslib = None
+        self.ssl_syshead = None
+
+        if not self.cfg.get('wrap_system_openssl'):
+            return
+
+        # Check the system libraries of OpenSSL
         libssl = {
             '1.0': {LINUX: 'libssl.so.10', DARWIN: 'libssl.1.0.dylib'},
             '1.1': {LINUX: 'libssl.so.1.1', DARWIN: 'libssl.1.1.dylib'},
@@ -86,7 +101,6 @@ class EB_OpenSSL_wrapper(Bundle):
             libssl_so = libssl[self.version][os_type]
             libssl_obj = ctypes.cdll.LoadLibrary(libssl_so)
         except OSError:
-            self.ssl_syslib = None
             self.log.debug("Library '%s' not found in host system", libssl_so)
         else:
             # ctypes.util.find_library only accepts unversioned library names
@@ -115,7 +129,6 @@ class EB_OpenSSL_wrapper(Bundle):
         ssl_include_dirs = [include for include in ssl_include_dirs if os.path.isdir(include)]
 
         # verify that the headers match our OpenSSL version
-        self.ssl_syshead = None
         for include in ssl_include_dirs:
             opensslv = read_file(os.path.join(include, 'opensslv.h'))
             header_majmin_version = re.search("SHLIB_VERSION_NUMBER\s\"([0-9]+\.[0-9]+)", opensslv, re.M)
--- a/easybuild/easyblocks/o/openssl_wrapper.py
+++ b/easybuild/easyblocks/o/openssl_wrapper.py
@@ -39,6 +39,7 @@ from easybuild.tools.run import run_cmd
 from easybuild.tools.systemtools import DARWIN, LINUX, get_os_type, get_shared_lib_ext
 from easybuild.tools.build_log import EasyBuildError, print_warning
 
+
 def locate_solib(libobj):
     """
     Return absolute path to loaded library using dlinfo
@@ -53,7 +54,7 @@ def locate_solib(libobj):
     libdl = ctypes.cdll.LoadLibrary(ctypes.util.find_library('dl'))
 
     dlinfo = libdl.dlinfo
-    dlinfo.argtypes  = ctypes.c_void_p, ctypes.c_int, ctypes.c_void_p
+    dlinfo.argtypes = ctypes.c_void_p, ctypes.c_int, ctypes.c_void_p
     dlinfo.restype = ctypes.c_int
 
     libpointer = ctypes.c_void_p()
@@ -61,6 +62,7 @@ def locate_solib(libobj):
     libpath = ctypes.cast(libpointer, ctypes.POINTER(LINKMAP)).contents.l_name
 
     return libpath
+
 
 class EB_OpenSSL_wrapper(Bundle):
     """
--- a/easybuild/easyblocks/o/openssl_wrapper.py
+++ b/easybuild/easyblocks/o/openssl_wrapper.py
@@ -172,8 +172,14 @@ class EB_OpenSSL_wrapper(Bundle):
             sys_include_dirs.extend(match.groups())
         self.log.debug("Found the following include directories in host system: %s", ', '.join(sys_include_dirs))
 
-        # headers are always in "include/openssl" subdirectories
-        ssl_include_dirs = [os.path.join(include, self.name.lower()) for include in sys_include_dirs]
+        # headers are located in 'include/openssl' by default
+        ssl_include_subdirs = [self.name.lower()]
+        if self.version == '1.1':
+            # but version 1.1 can be installed in 'include/openssl11/openssl' as well
+            # we prefer 'include/openssl' as long as the version of headers matches
+            ssl_include_subdirs.append(os.path.join('openssl11', self.name.lower()))
+
+        ssl_include_dirs = [os.path.join(incd, subd) for incd in sys_include_dirs for subd in ssl_include_subdirs]
         ssl_include_dirs = [include for include in ssl_include_dirs if os.path.isdir(include)]
 
         # verify that the headers match our OpenSSL version
--- a/easybuild/easyblocks/o/openssl_wrapper.py
+++ b/easybuild/easyblocks/o/openssl_wrapper.py
@@ -41,32 +41,8 @@ from easybuild.easyblocks.generic.bundle import Bundle
 from easybuild.framework.easyconfig import CUSTOM
 from easybuild.tools.filetools import change_dir, expand_glob_paths, mkdir, read_file, symlink, which
 from easybuild.tools.run import run_cmd
-from easybuild.tools.systemtools import DARWIN, LINUX, get_os_type, get_shared_lib_ext
+from easybuild.tools.systemtools import DARWIN, LINUX, get_os_type, get_shared_lib_ext, locate_solib
 from easybuild.tools.build_log import EasyBuildError, print_warning
-
-
-def locate_solib(libobj):
-    """
-    Return absolute path to loaded library using dlinfo
-    Based on https://stackoverflow.com/a/35683698
-    """
-    class LINKMAP(ctypes.Structure):
-        _fields_ = [
-            ("l_addr", ctypes.c_void_p),
-            ("l_name", ctypes.c_char_p)
-        ]
-
-    libdl = ctypes.cdll.LoadLibrary(ctypes.util.find_library('dl'))
-
-    dlinfo = libdl.dlinfo
-    dlinfo.argtypes = ctypes.c_void_p, ctypes.c_int, ctypes.c_void_p
-    dlinfo.restype = ctypes.c_int
-
-    libpointer = ctypes.c_void_p()
-    dlinfo(libobj._handle, 2, ctypes.byref(libpointer))
-    libpath = ctypes.cast(libpointer, ctypes.POINTER(LINKMAP)).contents.l_name
-
-    return libpath.decode('utf-8')
 
 
 class EB_OpenSSL_wrapper(Bundle):
--- a/easybuild/easyblocks/o/openssl_wrapper.py
+++ b/easybuild/easyblocks/o/openssl_wrapper.py
@@ -148,37 +148,38 @@ class EB_OpenSSL_wrapper(EasyBlock):
 
         if self.ssl_syslib and self.ssl_syshead:
             # The host system already provides the necessary OpenSSL files
-            ssl_lib_pattern = ['%s*.%s*' % (lib_so, shlib_ext) for lib_so in self.openssl_libs]
-            ssl_lib_pattern = [
-                os.path.join(os.path.dirname(self.ssl_syslib), '%s' % ptrn) for ptrn in ssl_lib_pattern
-            ]
-            ssl_include_path = self.ssl_syshead
-            ssl_bin = which(self.name.lower())
+            lib_pattern = ['%s*.%s*' % (lib_so, shlib_ext) for lib_so in self.openssl_libs]
+            lib_pattern = [os.path.join(os.path.dirname(self.ssl_syslib), '%s' % ptrn) for ptrn in lib_pattern]
+            include_pattern = [os.path.join(self.ssl_syshead, '*')]
+            bin_path = which(self.name.lower())
         else:
             # Use OpenSSL from EasyBuild
             ssl_root = get_software_root(self.name)
-            ssl_lib_pattern = [os.path.join(ssl_root, 'lib', '*.%s*' % shlib_ext)]
-            ssl_include_path = os.path.join(get_software_root(self.name), 'include', self.name.lower())
-            ssl_bin = os.path.join(ssl_root, 'bin', self.name.lower())
+            lib_pattern = [os.path.join(ssl_root, 'lib', '*.%s*' % shlib_ext)]
+            include_pattern = [os.path.join(get_software_root(self.name), 'include', self.name.lower(), '*')]
+            bin_path = os.path.join(ssl_root, 'bin', self.name.lower())
 
         # Link OpenSSL libraries
         lib64_dir = os.path.join(self.installdir, 'lib64')
         mkdir(lib64_dir)
         symlink('lib64', 'lib')
 
-        ssl_lib_files = expand_glob_paths(ssl_lib_pattern)
-        for libso in ssl_lib_files:
+        lib_files = expand_glob_paths(lib_pattern)
+        for libso in lib_files:
             symlink(libso, os.path.join(lib64_dir, os.path.basename(libso)))
 
         # Link OpenSSL headers
-        include_dir = os.path.join(self.installdir, 'include')
-        mkdir(include_dir)
-        symlink(ssl_include_path, os.path.join(include_dir, self.name.lower()))
+        include_dir = os.path.join(self.installdir, 'include', self.name.lower())
+        mkdir(include_dir, parents=True)
+
+        include_files = expand_glob_paths(include_pattern)
+        for header in include_files:
+            symlink(header, os.path.join(include_dir, os.path.basename(header)))
 
         # Link OpenSSL binary
         bin_dir = os.path.join(self.installdir, 'bin')
         mkdir(bin_dir)
-        symlink(ssl_bin, os.path.join(bin_dir, self.name.lower()))
+        symlink(bin_path, os.path.join(bin_dir, os.path.basename(bin_path)))
 
     def make_module_dep(self):
         """Make module file without OpenSSL dependency."""
--- a/easybuild/easyblocks/o/openssl_wrapper.py
+++ b/easybuild/easyblocks/o/openssl_wrapper.py
@@ -27,29 +27,22 @@ EasyBuild support for installing a wrapper module file for OpenSSL
 
 @author: Alex Domingo (Vrije Universiteit Brussel)
 """
-import ctypes
 import os
 import re
-
-try:
-    # only needed on macOS, may not be available on Linux
-    import ctypes.macholib.dyld
-except ImportError:
-    pass
 
 from easybuild.easyblocks.generic.bundle import Bundle
 from easybuild.framework.easyconfig import CUSTOM
 from easybuild.tools.filetools import change_dir, expand_glob_paths, mkdir, read_file, symlink, which
 from easybuild.tools.run import run_cmd
-from easybuild.tools.systemtools import DARWIN, LINUX, get_os_type, get_shared_lib_ext, locate_solib
+from easybuild.tools.systemtools import DARWIN, LINUX, get_os_type, get_shared_lib_ext, find_library_path, locate_solib
 from easybuild.tools.build_log import EasyBuildError, print_warning
 
 
 class EB_OpenSSL_wrapper(Bundle):
     """
-    Locate the installation files of OpenSSL in the host system.
+    Locate the installation files of OpenSSL in the host system
     If available, wrap the system OpenSSL by symlinking all installation files
-    Fall back to the bundled component otherwise.
+    Fall back to the bundled component otherwise
     """
 
     @staticmethod
@@ -67,25 +60,32 @@ class EB_OpenSSL_wrapper(Bundle):
 
         # Libraries packaged in OpenSSL
         openssl_libs = ['libssl', 'libcrypto']
+        # list of relevant library extensions per system and version of OpenSSL
+        # the first item should be the extension of an installation from source,
+        # it will be used in the sanity checks of the component
         openssl_libext = {
             '1.0': {
-                LINUX: 'so.10',
-                DARWIN: '1.0.dylib',
+                LINUX: ('so.1.0.0', 'so.10'),
+                DARWIN: ('1.0.dylib', ),
             },
             '1.1': {
-                LINUX: 'so.1.1',
-                DARWIN: '1.1.dylib',
+                LINUX: ('so.1.1', ),
+                DARWIN: ('1.1.dylib', ),
             },
         }
 
         os_type = get_os_type()
         if self.version in openssl_libext and os_type in openssl_libext[self.version]:
-            # generate names of versioned .so files
-            self.openssl_libs = ['%s.%s' % (lib, openssl_libext[self.version][os_type]) for lib in openssl_libs]
+            # generate matrix of versioned .so filenames
+            self.openssl_libs = [
+                ['%s.%s' % (lib, ext) for ext in openssl_libext[self.version][os_type]]
+                for lib in openssl_libs
+            ]
         else:
             raise EasyBuildError("Don't know name of OpenSSL system library for version %s and OS type %s",
                                  self.version, os_type)
 
+        # folders containing engines libraries
         self.openssl_engines = {
             '1.0': 'engines',
             '1.1': 'engines-1.1',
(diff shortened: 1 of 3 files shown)


easybuild/easyblocks/o/openssl_wrapper.py
```diff
@@ -0,0 +1,239 @@
...
+        self.log.debug("Found the following include directories in host system: %s", ', '.join(sys_include_dirs))
+
+        # headers are always in "include/openssl" subdirectories
+        ssl_include_dirs = [os.path.join(include, self.name.lower()) for include in sys_include_dirs]
+        ssl_include_dirs = [include for include in ssl_include_dirs if os.path.isdir(include)]
+
+        # verify that the headers match our OpenSSL version
+        for include in ssl_include_dirs:
+            opensslv = read_file(os.path.join(include, 'opensslv.h'))
+            header_majmin_version = re.search("SHLIB_VERSION_NUMBER\s\"([0-9]+\.[0-9]+)", opensslv, re.M)
```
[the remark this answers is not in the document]
lexming: this regex seems fine to me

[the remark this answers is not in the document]
boegel: You should use a raw string here:
```suggestion
            header_majmin_version = re.search(r"SHLIB_VERSION_NUMBER\s\"([0-9]+\.[0-9]+)", opensslv, re.M)
```

boegel:
@lexming Fails for me with:
```
$ eb --from-pr 12858 --include-easyblocks-from-pr 2429
...
  File "/data/gent/400/vsc40023/easybuild_easy_installed/lib/python2.7/site-packages/easybuild_framework-4.3.5.dev0-py2.7.egg/easybuild/main.py", line 117, in build_and_install_software
  File "/data/gent/400/vsc40023/easybuild_easy_installed/lib/python2.7/site-packages/easybuild_framework-4.3.5.dev0-py2.7.egg/easybuild/framework/easyblock.py", line 3371, in build_and_install_one
    app = app_class(ecdict['ec'])
  File "/tmp/eb-wb7k98av/included-easyblocks-u3uhdwkv/easybuild/easyblocks/openssl_wrapper.py", line 132, in __init__
    os.path.join(lib_dir, self.openssl_engines[self.version]),
  File "/usr/lib64/python3.6/posixpath.py", line 94, in join
    genericpath._check_arg_types('join', a, *p)
  File "/usr/lib64/python3.6/genericpath.py", line 151, in _check_arg_types
    raise TypeError("Can't mix strings and bytes in path components") from None
TypeError: Can't mix strings and bytes in path components
```

easybuild/easyblocks/o/openssl_wrapper.py
```diff
@@ -0,0 +1,241 @@
...
+# You should have received a copy of the GNU General Public License
+# along with EasyBuild.  If not, see <http://www.gnu.org/licenses/>.
+##
+"""
+EasyBuild support for installing a wrapper module file for OpenSSL
+
+@author: Alex Domingo (Vrije Universiteit Brussel)
+"""
+import ctypes
+import ctypes.macholib.dyld
```
boegel, line 31: This should be in a `try-except`, or under a guard so the import is only done on macOS.

Otherwise I see a hard fail on some OSs, like Ubuntu 18.04:

```
$ cat /etc/lsb-release
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=18.04
DISTRIB_CODENAME=bionic
DISTRIB_DESCRIPTION="Ubuntu 18.04.5 LTS"

$ python3 -V
Python 3.6.9

$ python3 -c 'import ctypes.macholib.dyld'
  File "<string>", line 1, in <module>
ModuleNotFoundError: No module named 'ctypes.macholib'
```

easybuild/easyblocks/o/openssl_wrapper.py
```diff
@@ -0,0 +1,241 @@
...
+
+    libdl = ctypes.cdll.LoadLibrary(ctypes.util.find_library('dl'))
+
+    dlinfo = libdl.dlinfo
+    dlinfo.argtypes = ctypes.c_void_p, ctypes.c_int, ctypes.c_void_p
+    dlinfo.restype = ctypes.c_int
+
+    libpointer = ctypes.c_void_p()
+    dlinfo(libobj._handle, 2, ctypes.byref(libpointer))
+    libpath = ctypes.cast(libpointer, ctypes.POINTER(LINKMAP)).contents.l_name
```
boegel, line 62: You better make sure you have a string value here, not a `bytes` value.
This is the cause for the `TypeError: Can't mix strings and bytes in path components` error I see when using Python 3.

This should suffice (also works in Python 2):

```python
libpath = libpath.decode('utf-8')
```

boegel (changes requested):

boegel:
Let's see if a close/re-open wakes up GitHub Actions...

Micket:
Soo.. do we not want this to find the CentOS7 openssl11 packages? Because they are intentionally not installed into a visible path, since that would break the rest of the system?

boegel:
> Soo.. do we not want this to find the CentOS7 openssl11 packages? Because they are intentionally not installed into a standard path, since that would break the rest of the system?

We could enhance this easyblock to make sure those are found too, but you would need to specify where to find them of course (and we could include the CentOS 7 location as a default in the easyblock)?

Micket:
Right, well, I'm not sure how to really suggest adding those paths, and, it's pretty much a mix here. We've got
1. `/usr/bin/openssl11`
2. `/usr/lib64/openssl11/` (or `/usr/lib64/libssl.so.1.1`)
3. `/usr/include/openssl11/`
So, optional `system_ssl_binary`, `system_ssl_libpath`, `ssl_ssl_includepath`? I think it's preferable for these to be set by default on centos/rhel 7 systems since it would be to easy to miss for a large portion or users.


Also, the automatic fallback method here; there isn't a practical way to update ever update it without having to revert to a live rebuild (ouch)? (as opposed to the more atomic change like with Java and modulerc)

easybuild/easyblocks/o/openssl_wrapper.py
```diff
@@ -0,0 +1,268 @@
...
+                print_warning("Found OpenSSL in host system, but not its engines."
+                              "Falling back to OpenSSL in EasyBuild")
+
+        # Check system include paths for OpenSSL headers
+        cmd = "gcc -E -Wp,-v -xc /dev/null"
+        (out, ec) = run_cmd(cmd, log_all=True, simple=False, trace=False)
+
+        sys_include_dirs = []
+        for match in re.finditer(r'^\s(/[^\0\n]*)+', out, re.MULTILINE):
+            sys_include_dirs.extend(match.groups())
```
Micket, line 172:
```suggestion
            sys_include_dirs.extend(match.groups())
        if self.version = '1.1':
            sys_include_dirs.append('/usr/include/openssl11')
```

easybuild/easyblocks/o/openssl_wrapper.py
```diff
@@ -0,0 +1,268 @@
...
+            mkdir(include_dir, parents=True)
+
+            include_files = expand_glob_paths(include_pattern)
+            for header in include_files:
+                symlink(header, os.path.join(include_dir, os.path.basename(header)))
+
+            # Link OpenSSL binary
+            bin_dir = os.path.join(self.installdir, 'bin')
+            mkdir(bin_dir)
+            symlink(bin_path, os.path.join(bin_dir, os.path.basename(bin_path)))
```
Micket, line 243:
```suggestion
            symlink(bin_path, os.path.join(bin_dir, os.self.name.lower()))
```

easybuild/easyblocks/o/openssl_wrapper.py
```diff
@@ -0,0 +1,268 @@
...
+
+        if self.ssl_syslib and self.ssl_sysheader:
+            # The host system already provides the necessary OpenSSL files
+            lib_pattern = ['%s*.%s*' % (lib_so, shlib_ext) for lib_so in self.openssl_libs]
+            lib_pattern = [os.path.join(os.path.dirname(self.ssl_syslib), '%s' % ptrn) for ptrn in lib_pattern]
+            lib_pattern.append(os.path.join(self.ssl_sysengines, '*'))
+
+            include_pattern = [os.path.join(self.ssl_sysheader, '*')]
+
+            bin_path = which(self.name.lower())
```
Micket, line 215:
```suggestion
            if self.version == '1.1':
                bin_path = which('openssl11', log_error=False)
                if bin_path is None:
                    bin_path = which(self.name.lower())
            else:
                bin_path = which(self.name.lower())
```

Note: the code doesn't check if this is the right version of `openssl` is found. Perhaps it's best to do this in the configure step asl well by introducing a `self.ssl_sysbin`, since we technically need that as well.

Micket (changes requested):
Something like this might suffice for finding system openssl11 on CentOS/RHEL 7

lexming:
@Micket hardcoding paths into the easyblock should not be necessary. If the host system is configured to use OpenSSL 1.1 then the libs and headers will be found. If OpenSSL 1.1 is not part of the environment, we should never link to it by default. I don't want EB to catch any rogue `.so` in my system. In my opinion, all we can do to cover that case is to provide an extra option so that the user can manually add extra paths to be checked by the easyblock.

Regarding the fallback method, a `modulerc` is not an option because it does not preserve the paths to the libraries. It automatically re-loads the _real_ OpenSSL module, so paths to libs change accordingly. In case of an upgrade of OpenSSL, the process is actually simpler with this wrapper as you only have to rebuild the wrapper with the new version, instead of installing the new OpenSSL version + updating the modulerc.

Micket:
I always considered rebuilding to be the worst case scenario, especially for a common dependency, as it means a large section of software is going to break for over a minute during rebuilds (assuming the new build even succeeds).
(A wrapper that symlinks to existing EB installs could work to minimize this, though somewhat messier (and impractical here))

But I really think we should automatically symlink to EPEL7s `openssl11/openssl11-devel` if present by default. In fact, that has been my only goal from all along, and what I think should be the recommended approach for all centos7 users. The only non-standard thing is an alternative include path and binary name (since there are 2 OS versions). The library itself is found in the standard paths (/usr/lib64/libssl.so.1.1)

easybuild/easyblocks/o/openssl_wrapper.py
```diff
@@ -0,0 +1,268 @@
...
+            super(EB_OpenSSL_wrapper, self).extract_step()
+
+    def install_step(self):
+        """Symlink target OpenSSL installation"""
+        shlib_ext = get_shared_lib_ext()
+
+        if self.ssl_syslib and self.ssl_sysheader:
+            # The host system already provides the necessary OpenSSL files
+            lib_pattern = ['%s*.%s*' % (lib_so, shlib_ext) for lib_so in self.openssl_libs]
+            lib_pattern = [os.path.join(os.path.dirname(self.ssl_syslib), '%s' % ptrn) for ptrn in lib_pattern]
```
Micket, line 210: I think this is an issue as it includes to much here, e.g.
```
OpenSSL/1.0/lib64/libcrypto.so -> /lib64/libcrypto.so
OpenSSL/1.0/lib64/libcrypto.so.10 -> /lib64/libcrypto.so.10
OpenSSL/1.0/lib64/libcrypto.so.1.0.2k -> /lib64/libcrypto.so.1.0.2k
OpenSSL/1.0/lib64/libcrypto.so.1.1 -> /lib64/libcrypto.so.1.1
OpenSSL/1.0/lib64/libcrypto.so.1.1.1g -> /lib64/libcrypto.so.1.1.1g
OpenSSL/1.0/lib64/libssl3.so -> /lib64/libssl3.so
OpenSSL/1.0/lib64/libssl.so -> /lib64/libssl.so
OpenSSL/1.0/lib64/libssl.so.10 -> /lib64/libssl.so.10
OpenSSL/1.0/lib64/libssl.so.1.0.2k -> /lib64/libssl.so.1.0.2k
OpenSSL/1.0/lib64/libssl.so.1.1 -> /lib64/libssl.so.1.1
OpenSSL/1.0/lib64/libssl.so.1.1.1g -> /lib64/libssl.so.1.1.1g
```

Also, including down to the specific version 
```
OpenSSL/1.0/lib64/libssl.so.1.0.2k -> /lib64/libssl.so.1.0.2k
```
would be dangerous since this library would disappear during an OS update.

I would say the most desirable outcome is to only have
```
libcrypto.so -> libcrypto.so.X.X  (just local symlink)
libcrypto.so.X.X -> /lib64/libcrypto.so.XX
libssl.so -> libssl.so.X.X  (just local symlink)
libssl.so.X.X -> /lib64/libssl.so.X.X
```

Maybe use something like
```
        libssl = {
            '1.0': {
                LINUX: ('libssl.so.10', 'libcrypto.so.10'),
                DARWIN: ('libssl.1.0.dylib', 'libcrypto.1.0.dylib'),
            },
            '1.1': {
                LINUX: ('libssl.so.1.1', 'libcrypto.so.1.1'),
                DARWIN: ('libssl.1.1.dylib', 'libcrypto.1.1.dylib'),
            },
        }
```
and use the absolute path (from `ctypes.cdll.LoadLibrary`) to these 2 libraries specifically instead of grabbing the directory?

Micket (changes requested):

easybuild/easyblocks/o/openssl_wrapper.py
boegel: `os.self.name` -> `self.name` :)

boegel:
I agree with @Micket here that the `openssl11` stuff in CentOS 7 should be picked up and used automatically if it's installed, since that's the best way forward (since that way you get automatic security updates from the OS without having to reinstall the OpenSSL wrapper).

So perhaps we can do this by default, but only when on CentOS 7, and provide an easy way to opt-out?

easybuild/easyblocks/o/openssl_wrapper.py
```diff
@@ -0,0 +1,268 @@
...
+                    self.ssl_sysengines = engines_path
+                    self.log.debug("Found OpenSSL engines in: %s", self.ssl_sysengines)
+
+            if not self.ssl_sysengines:
+                self.ssl_syslib = None
+                print_warning("Found OpenSSL in host system, but not its engines."
+                              "Falling back to OpenSSL in EasyBuild")
+
+        # Check system include paths for OpenSSL headers
+        cmd = "gcc -E -Wp,-v -xc /dev/null"
```
Flamefire, line 167: You should reset the language here, that has hit me and others in different places already. Maybe:
```suggestion
        cmd = "LC_ALL=C gcc -E -Wp,-v -xc /dev/null"
```

lexming:
@Micket @boegel the situation is not as bad then. If both the libraries and headers are in _known_ paths by the environment, such as `/usr/lib64` and `/usr/include`, all we have to do is add some logic to also consider headers in a `openssl11` subfolder and the `openssl11` binary. That does not need to hardcode any paths in the easyblock.

Micket:
For reference, these are all the relevant paths you get with EPEL7's `openssl11`:
```
/usr/bin/make-dummy-cert
/usr/bin/openssl11
/usr/bin/renew-dummy-cert
```
and `openssl-libs`:
```
/usr/lib64/engines-1.1
/usr/lib64/engines-1.1/afalg.so
/usr/lib64/engines-1.1/capi.so
/usr/lib64/engines-1.1/padlock.so
/usr/lib64/libcrypto.so.1.1
/usr/lib64/libcrypto.so.1.1.1g
/usr/lib64/libssl.so.1.1
/usr/lib64/libssl.so.1.1.1g
```
and `openssl-devel`:
```
```
/usr/include/openssl11
/usr/include/openssl11/openssl
/usr/include/openssl11/openssl/aes.h
...
/usr/lib64/openssl11/libcrypto.so
/usr/lib64/openssl11/libssl.so
/usr/lib64/pkgconfig/libcrypto11.pc
/usr/lib64/pkgconfig/libssl11.pc
/usr/lib64/pkgconfig/openssl11.pc
```
(maybe we should symlink in the pkgconfig files as well?)

File: easybuild/easyblocks/o/openssl_wrapper.py (hunk 8 above)
Comment (lexming) [reply easybuild/easyblocks/o/openssl_wrapper.py]: @Flamefire thanks!

File: easybuild/easyblocks/o/openssl_wrapper.py (hunk 4 above)
Comment (lexming) [reply easybuild/easyblocks/o/openssl_wrapper.py:173]: @Micket fixed in https://github.com/easybuilders/easybuild-easyblocks/pull/2429/commits/4101ba39804eca373611651a26fbf3aa56811e18

File: easybuild/easyblocks/o/openssl_wrapper.py (hunk 5 above)
Comment (lexming) [reply easybuild/easyblocks/o/openssl_wrapper.py]: @Micket `openssl11` binaries fixed in https://github.com/easybuilders/easybuild-easyblocks/pull/2429/commits/25d718c26bc846c5893cb505c0c491820423570d

File: easybuild/easyblocks/o/openssl_wrapper.py (hunk 6 above)
Comment (lexming) [reply easybuild/easyblocks/o/openssl_wrapper.py]: @Micket I also improved the sanity check command to check the actual version of the binary in https://github.com/easybuilders/easybuild-easyblocks/pull/2429/commits/25d718c26bc846c5893cb505c0c491820423570d

**Comment by lexming** [comment]
@Micket thanks for the feedback! I updated the easyblock to be capable of handling a system OpenSSL v1.1 installed in parallel to v1.0. Please check it out.

I already tested it on my side in a CentOS machine with the infamous `openssl11` and it worked well :+1:

**Comment by lexming** [comment]
> (maybe we should symlink in the pkgconfig files as well?)

Those files cannot be symlinked, we would have to re-create them with the appropriate contents.

File: easybuild/easyblocks/o/openssl_wrapper.py (hunk 7 above)
Comment (lexming) [reply easybuild/easyblocks/o/openssl_wrapper.py]: @Micket limited the extend of libraries symlinked in https://github.com/easybuilders/easybuild-easyblocks/pull/2429/commits/bcbe0372bf787e3190e4c00994eef32684dc5835

File: easybuild/easyblocks/o/openssl_wrapper.py [hunk 9]
```diff
@@ -0,0 +1,286 @@
...
+            # Install OpenSSL component
+            super(EB_OpenSSL_wrapper, self).install_step()
+
+    def sanity_check_step(self):
+        """Custom sanity check for OpenSSL wrapper."""
+        shlib_ext = get_shared_lib_ext()
+
+        ssl_files = [os.path.join('bin', self.name.lower())]
+        ssl_files.extend(os.path.join('lib', libso) for libso in self.openssl_libs)
+        ssl_files.extend(os.path.join('lib', '%s.%s' % (libso.split('.')[0], shlib_ext)) for libso in self.openssl_libs)
```
boegel, line 284: This doesn't work for a from-source installation of OpenSSL 1.0 on RHEL8, because there's no `lib/libssl.so.10`:
```
$ ls -l /apps/RHEL8/zen2-ib/software/OpenSSL/1.0/lib/
total 16756
drwxrwxr-x 2 vsc40023 vsc40023    4096 May 17 17:45 engines
-rw-r--r-- 1 vsc40023 vsc40023 4577474 May 17 17:45 libcrypto.a
lrwxrwxrwx 1 vsc40023 vsc40023      18 May 17 17:45 libcrypto.so -> libcrypto.so.1.0.0
-r-xr-xr-x 1 vsc40023 vsc40023 2707224 May 17 17:45 libcrypto.so.1.0.0
-rw-r--r-- 1 vsc40023 vsc40023  763794 May 17 17:45 libssl.a
lrwxrwxrwx 1 vsc40023 vsc40023      15 May 17 17:45 libssl.so -> libssl.so.1.0.0
-r-xr-xr-x 1 vsc40023 vsc40023  513616 May 17 17:45 libssl.so.1.0.0
drwxrwxr-x 2 vsc40023 vsc40023    4096 May 17 17:45 pkgconfig
```

I think we should just check for the `.so` (since that's a symlink to the actual library). If it's a broken symlink, the sanity check will fail.

boegel (changes requested):

easybuild/easyblocks/o/openssl_wrapper.py
```diff
@@ -0,0 +1,286 @@
...
+        ssl_include_dirs = [os.path.join(incd, subd) for incd in sys_include_dirs for subd in ssl_include_subdirs]
+        ssl_include_dirs = [include for include in ssl_include_dirs if os.path.isdir(include)]
+
+        # verify that the headers match our OpenSSL version
+        for include_dir in ssl_include_dirs:
+            opensslv_path = os.path.join(include_dir, 'opensslv.h')
+            if os.path.exists(opensslv_path):
+                opensslv = read_file(opensslv_path)
+                header_majmin_version = re.search(r"SHLIB_VERSION_NUMBER\s\"([0-9]+\.[0-9]+)", opensslv, re.M)
+                if re.match("^{}".format(*header_majmin_version.groups()), self.version):
```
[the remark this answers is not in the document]
boegel: fixed in 2897480

easybuild/easyblocks/o/openssl_wrapper.py
```diff
@@ -0,0 +1,286 @@
...
+                symlink(versioned_lib, unversioned_lib)
+
+            # Link OpenSSL headers in system
+            include_dir = os.path.join(self.installdir, 'include', self.name.lower())
+            mkdir(include_dir, parents=True)
+
+            include_pattern = [os.path.join(self.ssl_sysheader, '*')]
+            include_files = expand_glob_paths(include_pattern)
+            for header in include_files:
+                symlink(header, os.path.join(include_dir, os.path.basename(header)))
```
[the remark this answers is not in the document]
boegel: Actually no, since directory symlinks get resolved more easily.

easybuild/easyblocks/o/openssl_wrapper.py
boegel: fixed in 2897480

easybuild/easyblocks/o/openssl_wrapper.py
```diff
@@ -0,0 +1,286 @@
...
+
+from easybuild.easyblocks.generic.bundle import Bundle
+from easybuild.framework.easyconfig import CUSTOM
+from easybuild.tools.filetools import expand_glob_paths, mkdir, read_file, symlink, which
+from easybuild.tools.run import run_cmd
+from easybuild.tools.systemtools import DARWIN, LINUX, get_os_type, get_shared_lib_ext
+from easybuild.tools.build_log import EasyBuildError, print_warning
+
+
+def locate_solib(libobj):
```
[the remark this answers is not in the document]
boegel: added `locate_solib` to `systemtools` in https://github.com/easybuilders/easybuild-framework/pull/3682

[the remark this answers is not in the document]
boegel: fixed in f07cc40

boegel (dismissed):

easybuild/easyblocks/o/openssl_wrapper.py
```diff
@@ -0,0 +1,286 @@
...
+        """Symlink target OpenSSL installation"""
+
+        if self.ssl_syslib and self.ssl_sysheader:
+            # Link OpenSSL libraries in system
+            lib64_dir = os.path.join(self.installdir, 'lib64')
+            lib64_engines_dir = os.path.join(lib64_dir, os.path.basename(self.ssl_sysengines))
+            mkdir(lib64_engines_dir, parents=True)
+
+            lib_files = [lib_so for lib_so in self.openssl_libs]
+            lib_files = [os.path.join(os.path.dirname(self.ssl_syslib), '%s' % ptrn) for ptrn in lib_files]
```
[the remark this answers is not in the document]
lexming: indeed, leftovers from the multiple back and forths :slightly_smiling_face: thanks for the fix

easybuild/easyblocks/o/openssl_wrapper.py
lexming: Then in this case the easyblock is not wrapping that installation of v1.0. We should add `libssl.so.1.0.0` as a possible file name

lexming: fixed in https://github.com/easybuilders/easybuild-easyblocks/pull/2429/commits/5701d618c3b6e97dbbf6635ede2cd15c43c33324 : now this easyblock can check multiple libraries to be wrapped per system/version and it also can properly sanity check the built component

lexming:
With the addition of multiple library names per system/version, I moved the code loading and finding the library to its own function in systemtools. See https://github.com/easybuilders/easybuild-framework/pull/3683

easybuild/easyblocks/o/openssl_wrapper.py
```diff
@@ -0,0 +1,269 @@
...
+        self.ssl_syslib = None
+        self.ssl_sysheader = None
+        self.ssl_sysengines = None
+        self.ssl_sysbin = None
+
+        if not self.cfg.get('wrap_system_openssl'):
+            return
+
+        # Check the system libraries of OpenSSL
+        for n, libssl in enumerate(self.openssl_libs[0]):
```
boegel, line 104: Don't use single-letter variables outside of list comprehensions please. I'd go with `idx` here.

easybuild/easyblocks/o/openssl_wrapper.py
```diff
@@ -0,0 +1,269 @@
...
+            self.ssl_syslib = find_library_path(libssl)
+            if self.ssl_syslib:
+                # reduce matrix of library names to this one
+                self.openssl_libs = zip(*self.openssl_libs)[n]
```
boegel, line 108: This looks weird, and it doesn't work either (at least not in Python 3):
```
  File "/users/kehoste/easybuild-framework/easybuild/main.py", line 117, in build_and_install_software
  File "/users/kehoste/easybuild-framework/easybuild/framework/easyblock.py", line 3533, in build_and_install_one
    app = app_class(ecdict['ec'])
  File "/tmp/eb-a0bcx10h/included-easyblocks-y6ba563p/easybuild/easyblocks/openssl_wrapper.py", line 108, in __init__
    self.openssl_libs = zip(*self.openssl_libs)[n]
TypeError: 'zip' object is not subscriptable
```

boegel (changes requested):

easybuild/easyblocks/o/openssl_wrapper.py
lexming: fixed in https://github.com/easybuilders/easybuild-easyblocks/pull/2429/commits/724ea6ec2fe2bc2b99e0c74a56820056056edf90

easybuild/easyblocks/o/openssl_wrapper.py
lexming: right, I forgot about python3, `zip` returns an iterator instead of a list. Fixed in https://github.com/easybuilders/easybuild-easyblocks/pull/2429/commits/724ea6ec2fe2bc2b99e0c74a56820056056edf90

Apart from that, I don't know how to make it less weird, it's just an unpack into `zip`. Maybe I could take the element from the list in another line, something like
```
zipped_libs = list(zip(*self.openssl_libs))
self.openssl_libs = zipped_libs[idx]
```

lexming: actually, we can just use a list comprehension. Done in https://github.com/easybuilders/easybuild-easyblocks/pull/2429/commits/9a9f34a64f77d05e2e6a3e6c5d2d8688c164ad25

easybuild/easyblocks/o/openssl_wrapper.py
```diff
@@ -0,0 +1,269 @@
...
+                symlink(header_file, os.path.join(include_dir, os.path.basename(header_file)))
+
+            # link OpenSSL binary in system
+            bin_dir = os.path.join(self.installdir, 'bin')
+            mkdir(bin_dir)
+            symlink(self.ssl_sysbin, os.path.join(bin_dir, self.name.lower()))
+
+        else:
+            # without wrapping, reduce matrix of library names to the first option in each family
+            self.openssl_libs = [lib_name[0] for lib_name in self.openssl_libs]
```
boegel, line 243: There's a problem here, when OpenSSL 1.0 gets installed from source on Ubuntu 18.04, the sanity check fails:

```
== sanity checking...
  >> file 'bin/openssl' found: OK
  >> file 'lib/l' found: FAILED
  >> file 'lib/l' found: FAILED
  >> file 'lib/l.so' found: FAILED
  >> file 'lib/l.so' found: FAILED
  >> (non-empty) directory 'include/openssl' found: OK
  >> (non-empty) directory 'lib/engines' found: OK
  >> running command 'ssl_ver=$(openssl version); [ ${ssl_ver:8:3} == '1.0' ]' ...
  >> result for command 'ssl_ver=$(openssl version); [ ${ssl_ver:8:3} == '1.0' ]': OK
```

The (really) strange thing is that I'm not seeing this on other systems where `OpenSSL/1.0` is being built from source (like on RHEL8 and Ubuntu 20.04).

Could be related to using Python 2 vs Python 3 somehow?!

When I log `self.openssl_libs` right after this statement, I get:
```
== 2021-05-19 07:45:33,411 openssl_wrapper.py:244 INFO self.openssl_libs: ['l', 'l']
```

boegel (changes requested):

easybuild/easyblocks/o/openssl_wrapper.py
lexming: I see what happens, my fault...

boegel: Before this statement, we have:
```
== 2021-05-19 07:50:49,663 openssl_wrapper.py:245 INFO self.openssl_libs: ['libssl.so.1.0.0', 'libcrypto.so.1.0.0']
```

lexming: The problem was that the dimension of `self.openssl_libs` was being reduced twice in the case where the host system had the libraries of OpenSSL but **not** the headers. I guess that your Ubuntu 18.04 lacks `openssl-dev`.

boegel:
@lexming Another crash when `OpenSSL-1.0.eb` is being built from source after your last update:
```
  File "/home/centos/easybuild/easybuild-framework/easybuild/main.py", line 117, in build_and_install_software
  File "/home/centos/easybuild/easybuild-framework/easybuild/framework/easyblock.py", line 3571, in build_and_install_one
  File "/home/centos/easybuild/easybuild-framework/easybuild/framework/easyblock.py", line 3465, in run_all_steps
  File "/home/centos/easybuild/easybuild-framework/easybuild/framework/easyblock.py", line 3320, in run_step
    step_method(self)()
  File "/tmp/eb-t7009csl/included-easyblocks-lj9yuliv/easybuild/easyblocks/openssl_wrapper.py", line 252, in sanity_check_step
    ssl_files.extend(os.path.join('lib', libso) for libso in self.openssl_libs)
  File "/tmp/eb-t7009csl/included-easyblocks-lj9yuliv/easybuild/easyblocks/openssl_wrapper.py", line 252, in <genexpr>
    ssl_files.extend(os.path.join('lib', libso) for libso in self.openssl_libs)
  File "/usr/lib64/python3.6/posixpath.py", line 94, in join
    genericpath._check_arg_types('join', a, *p)
  File "/usr/lib64/python3.6/genericpath.py", line 149, in _check_arg_types
    (funcname, s.__class__.__name__)) from None
TypeError: join() argument must be str or bytes, not 'list'
```

lexming:
@boegel will you stop breaking it? I'll have to ban your username from this easyblock...

In fact, I introduced this bug in my last commit. I'll make a better fix 👍

lexming:
@boegel fixed in https://github.com/easybuilders/easybuild-easyblocks/pull/2429/commits/fd425a2826b63768b576b6d9c693f80764481259. I replaced the changing list of OpenSSL libraries in `self.openssl_libs` into two separate variables, where the target libraries of the installation are always held in a simple list (`self.target_ssl_libs`). This fixes the last two errors.

boegel (approved):

boegel:
Tested extensively with https://github.com/easybuilders/easybuild-easyconfigs/pull/12858 across a range of different systems, operating systems, and CPU architectures.

@lexming: Thanks a lot for your efforts on this!

PR #2435, opened 2021-05-20
Add AOMP EasyBlock support

nordmoen:
This PR adds an AOMP EasyBlock to better pick `CUDA` support.

### DIFF
--- a/easybuild/easyblocks/a/aomp.py
+++ b/easybuild/easyblocks/a/aomp.py
@@ -2,7 +2,7 @@
 Support for building and installing AOMP - AMD OpenMP compiler, implemented as
 an EasyBlock
 
-@author: Jørgen Nordmoen (University Center for Information Technology - UiO)
+@author: Jorgen Nordmoen (University Center for Information Technology - UiO)
 """
 
 from easybuild.easyblocks.generic.binary import Binary
--- a/easybuild/easyblocks/a/aomp.py
+++ b/easybuild/easyblocks/a/aomp.py
@@ -10,7 +10,6 @@ from easybuild.framework.easyconfig import CUSTOM
 from easybuild.tools.build_log import EasyBuildError
 from easybuild.tools.config import build_option
 from easybuild.tools.modules import get_software_root
-from easybuild.tools.systemtools import X86_64, get_cpu_architecture
 
 AOMP_ALL_COMPONENTS = ['roct', 'rocr', 'project', 'libdevice', 'openmp',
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/a/aomp.py
```diff
@@ -0,0 +1,127 @@
...
+class EB_AOMP(Binary):
+    """Support for installing AOMP"""
+
+    @staticmethod
+    def extra_options():
+        extra_vars = Binary.extra_options()
+        extra_vars.update({
+            'components': [None, "AOMP components to build. Possible components: " +
+                           ', '.join(AOMP_ALL_COMPONENTS), CUSTOM],
+            'cuda_compute_capabilities': [[], "List of CUDA compute capabilities to build with", CUSTOM],
```
akesandgren, line 37: This is available in the framework already and is not needed here.

akesandgren (changes requested):

easybuild/easyblocks/a/aomp.py
nordmoen: I have removed it now

nordmoen:
@akesandgren Would you be able to review this again?

easybuild/easyblocks/a/aomp.py
```diff
@@ -71,7 +70,6 @@ def configure_step(self):
             # Use the commandline / easybuild config option if given, else use
             # the value from the EC (as a default)
             cuda_cc = build_option('cuda_compute_capabilities')
-            cuda_cc = cuda_cc or self.cfg['cuda_compute_capabilities']
```
akesandgren: You should still allow for cuda_compute_capabilities to be specified in the easyconfig, i.e. this line should not be removed.
See tensorflow.py line 551 for a example on how to use it.

akesandgren (changes requested):

easybuild/easyblocks/a/aomp.py
nordmoen: I thought you wanted me to remove the EB capability of defining `cuda_compute_capabilities` and that it had to be defined in the `extra_vars` option.

I have added back in the capability to either use the command line parameter or variable defined in EB (copied from TF EB).

akesandgren: No, it was just that  you don't need to specify it in extra_vars since it is made available there from the framework.

akesandgren (approved):
LGTM

akesandgren:
Going in, thanks @nordmoen!

PR #2436, opened 2021-05-20
added fPIC support to g2libc EasyBlock

mboisson:


easybuild/easyblocks/g/g2clib.py
```diff
@@ -60,10 +60,13 @@ def build_step(self):
         # beware: g2clib uses INC, while g2lib uses INCDIR !
         buildopts = ' '.join([
             r'CC="%s"' % os.getenv('CC'),
-            r'CFLAGS="%s \$(INC) \$(DEF) -D__64BIT__"' % os.getenv('CFLAGS'),
             r'FC="%s"' % os.getenv('F90'),
             r'INC="-I%s"' % os.path.join(jasper, 'include'),
         ])
+        if self.cfg["toolchainopts"].get('pic', False):
+            buildopts += r' CFLAGS="%s -fPIC \$(INC) \$(DEF) -D__64BIT__"' % os.getenv('CFLAGS')
```
boegel, line 67: If you use `toolchainopts = {'pic': True}` in the easyconfig file, that should work too?

If you prefer doing this in the easyblock, you should update `$CFLAGS` before the `buildopts = ...`?

boegel (changes requested):

easybuild/easyblocks/g/g2clib.py
mboisson: The EasyConfig toolchainopts parameter is not used without this patch

boegel: If you enable the `pic` toolchain option, then `-fPIC` should get added to `$CFLAGS`?

mboisson: I agree it should. But it does not build with PIC when 
```
toolchainopts = {'pic': True}
``` 
is in the EasyConfig, not if this patch is not there.

mboisson:
nevermind... apparently we did not have `pic` in the past in the recipe, the recipe was modified but not reinstalled... 🙄

PR #2449, opened 2021-05-31
enhance ConfigureMake generic easyblock to add support for building multiple build targets

mboisson:


mboisson:
Here is an example of EasyConfig which would use this feature: 
https://github.com/ComputeCanada/easybuild-easyconfigs/blob/computecanada-main/easybuild/easyconfigs/a/Arioc/Arioc-1.43-gcccorecuda-2020a.eb

easybuild/easyblocks/generic/configuremake.py
```diff
@@ -175,6 +176,7 @@ def extra_options(extra_vars=None):
         extra_vars = EasyBlock.extra_options(extra=extra_vars)
         extra_vars.update({
             'build_cmd': [DEFAULT_BUILD_CMD, "Build command to use", CUSTOM],
+            'build_targets': [DEFAULT_BUILD_TARGETS, "Targets to build", CUSTOM],
```
boegel, line 179: This should say "`Target name (string) or list of target names to build`"?

easybuild/easyblocks/generic/configuremake.py
```diff
@@ -65,6 +65,7 @@
 DEFAULT_CONFIGURE_CMD = './configure'
 DEFAULT_BUILD_CMD = 'make'
 DEFAULT_INSTALL_CMD = 'make install'
+DEFAULT_BUILD_TARGETS = ''
```
boegel, line 68: `DEFAULT_BUILD_TARGET`

easybuild/easyblocks/generic/configuremake.py
```diff
@@ -320,14 +322,20 @@ def build_step(self, verbose=False, path=None):
...
-            self.cfg.get('build_cmd') or DEFAULT_BUILD_CMD,
-            paracmd,
-            self.cfg['buildopts'],
-        ])
+        targets = self.cfg['build_targets'] or DEFAULT_BUILD_TARGETS
+        # ensure it is a list
+        targets = [targets] if isinstance(targets, str) else targets
+
+        for target in targets:
+            cmd = ' '.join([
```
boegel, line 330: let's add a log message here:

```python
self.log.info("Building target '%s'", target)

boegel (changes requested):

easybuild/easyblocks/generic/configuremake.py
mboisson: done

smoors (approved):
tested with the easyconfig linked to (using gcccuda), works nicely

smoors:
Going in, thanks @mboisson!

PR #2454, opened 2021-06-02
enhance sanity check for Clang to verify if CUDA offload library was produced

nordmoen:
When building Clang with the `gcccuda` toolchain (or just `CUDA` loaded) the offloading library might not be built due to a required dependency on `elfutils`, related to easybuilders/easybuild-easyconfigs#13006. This PR adds a check to see if the appropriate CUDA `libomptarget` library has been built.

### DIFF
--- a/easybuild/easyblocks/c/clang.py
+++ b/easybuild/easyblocks/c/clang.py
@@ -481,10 +481,16 @@ class EB_Clang(CMakeMake):
 
         if 'NVPTX' in self.cfg['build_targets']:
             arch = get_cpu_architecture()
-            if arch == POWER:
-                # The 'libomptarget.rtl' on POWER systems is identified using
-                # the following arch
+            # Check architecture explicitly since Clang uses potentially
+            # different names
+            if arch == X86_64:
+                arch = 'x86_64'
+            elif arch == POWER:
                 arch = 'ppc64'
+            elif arch == AARCH64:
+                arch = 'aarch64'
+            else:
+                print_warning("Unknown CPU architecture (%s) for OpenMP offloading!" % arch)
             custom_paths['files'].extend(["lib/libomptarget.%s" % shlib_ext,
                                           "lib/libomptarget-nvptx.a",
                                           "lib/libomptarget.rtl.cuda.%s" % shlib_ext,
--- a/easybuild/easyblocks/c/clang.py
+++ b/easybuild/easyblocks/c/clang.py
@@ -479,6 +479,17 @@ class EB_Clang(CMakeMake):
         if LooseVersion(self.version) >= LooseVersion('3.8'):
             custom_paths['files'].extend(["lib/libomp.%s" % shlib_ext, "lib/clang/%s/include/omp.h" % self.version])
 
+        if 'NVPTX' in self.cfg['build_targets']:
+            arch = get_cpu_architecture()
+            if arch == POWER:
+                # The 'libomptarget.rtl' on POWER systems is identified using
+                # the following arch
+                arch = 'ppc64'
+            custom_paths['files'].extend(["lib/libomptarget.%s" % shlib_ext,
+                                          "lib/libomptarget-nvptx.a",
+                                          "lib/libomptarget.rtl.cuda.%s" % shlib_ext,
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/c/clang.py
```diff
@@ -479,6 +479,17 @@ def sanity_check_step(self):
         if LooseVersion(self.version) >= LooseVersion('3.8'):
             custom_paths['files'].extend(["lib/libomp.%s" % shlib_ext, "lib/clang/%s/include/omp.h" % self.version])
 
+        if 'NVPTX' in self.cfg['build_targets']:
+            arch = get_cpu_architecture()
+            if arch == POWER:
+                # The 'libomptarget.rtl' on POWER systems is identified using
+                # the following arch
+                arch = 'ppc64'
```
boegel, line 487: @nordmoen Let's also cover `aarch64` here, which has `libomptarget.rtl.aarch64.so`:

```python
elif arch == AARCH64:
    arch = 'aarch64'
```

Or equivalenty:

```python
else:
    arch = arch.lower()
```

boegel (changes requested):

easybuild/easyblocks/c/clang.py
nordmoen: I had a feeling it would be wrong on `AArch64`. I have change the code to do explicit comparison against the different architectures so that we are not dependent on the constants.

boegel (changes requested):
With this change, the use of `eb --module-only` is broken for Clang easyconfigs, because the default value for the custom `build_targets` easyconfig parameter is `None` (and it should stay that way):

```
  File "/data/gent/400/vsc40023/easybuild_easy_installed/lib/python2.7/site-packages/easybuild_framework-4.3.5.dev0-py2.7.egg/easybuild/main.py", line 117, in build_and_install_software
  File "/data/gent/400/vsc40023/easybuild_easy_installed/lib/python2.7/site-packages/easybuild_framework-4.3.5.dev0-py2.7.egg/easybuild/framework/easyblock.py", line 3633, in build_and_install_one
  File "/data/gent/400/vsc40023/easybuild_easy_installed/lib/python2.7/site-packages/easybuild_framework-4.3.5.dev0-py2.7.egg/easybuild/framework/easyblock.py", line 3531, in run_all_steps
  File "/data/gent/400/vsc40023/easybuild_easy_installed/lib/python2.7/site-packages/easybuild_framework-4.3.5.dev0-py2.7.egg/easybuild/framework/easyblock.py", line 3386, in run_step
    step_method(self)()
  File "/tmp/eb-561wpn5k/included-easyblocks-eyg4ijqe/easybuild/easyblocks/clang.py", line 481, in sanity_check_step
    if 'NVPTX' in self.cfg['build_targets']:
TypeError: argument of type 'NoneType' is not iterable
```

This is easy to fix though, by moving the logic w.r.t. `build_targets` into the `prepare_step` method, since that step is always run, also when using `--module-only`.

Fixed in https://github.com/nordmoen/easybuild-easyblocks/pull/1

boegel:
Test report by @boegel
 * **SUCCESS** _Clang-11.0.1-GCCcore-10.2.0.eb_ 
 * **SUCCESS** _Clang-11.0.1-gcccuda-2020b.eb_ 
 * **SUCCESS** _Clang-8.0.0-GCCcore-8.2.0-CUDA-10.1.105.eb_ 
 * **SUCCESS** _Clang-8.0.1-GCC-8.2.0-2.31.1-CUDA-10.1.105.eb_ 
 * **SUCCESS** _Clang-8.0.1-GCC-8.3.0-CUDA-10.1.243.eb_ 
 * **SUCCESS** _Clang-9.0.1-GCC-8.3.0-CUDA-10.1.243.eb_ 

node2617.swalot.os - Linux centos linux 7.9.2009, x86_64, Intel(R) Xeon(R) CPU E5-2660 v3 @ 2.60GHz (haswell), Python 3.6.8
See https://gist.github.com/57027dcf464c1257474f6834ba960d9e for a full test report.

boegel (approved):

PR #2459, opened 2021-06-07
Always add add distinct_host_configuration=false to TensorFlow compilation

Flamefire:
This fixes a failure when using optarch=False due to missing CPATH and friends while compiling e.g. protobuf files.

More info:
We pass env vars such as CPATH explicitely because Bazel clears the whole env. If you use `distinct_host_configuration=true`  then bazel will have 2 envs: One for host compilation one for target compilation. We only pass stuff to the target env. Hence the host env is empty and the compiler can't find protobuf due to CPATH not being set

So optarch for TF is broken and likely always has been due to that flag. We can either remove the check which means optarch can now be used, although only for lower archs than the build node, or "fix" it by passing the env vars also to the Bazel host env.
But I think we don't support building for higher archs anyway, as e.g. all tests would fail to execute. Hence IMO we can remove the if and be done

### DIFF
--- a/easybuild/easyblocks/t/tensorflow.py
+++ b/easybuild/easyblocks/t/tensorflow.py
@@ -786,11 +786,9 @@ class EB_TensorFlow(PythonPackage):
         # Ignore user environment for Python
         self.target_opts.append('--action_env=PYTHONNOUSERSITE=1')
 
-        # use same configuration for both host and target programs, which can speed up the build
-        # only done when optarch is enabled, since this implicitely assumes that host and target platform are the same
-        # see https://docs.bazel.build/versions/master/guide.html#configurations
-        if self.toolchain.options.get('optarch'):
-            self.target_opts.append('--distinct_host_configuration=false')
+        # Use the same configuration (i.e. environment) for compiling and using host tools
+        # This means that our action_envs are always passed
+        self.target_opts.append('--distinct_host_configuration=false')
 
         # TF 2 (final) sets this in configure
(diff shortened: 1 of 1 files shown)


boegel:
Without this change the installation of `TensorFlow` 2.5.0 is failing when I add `'optarch': False` to `toolchainopts`, as follows:

```
In file included from ./tensorflow/python/framework/python_op_gen.h:22,
                 from tensorflow/python/framework/python_op_gen_main.cc:16:
bazel-out/host/bin/tensorflow/core/framework/op_def.pb.h:10:10: fatal error: google/protobuf/port_def.inc: No such file or directory
   10 | #include <google/protobuf/port_def.inc>
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
compilation terminated.
```

Reason is that the build environment for the host compilation is not properly set up unless `--distinct_host_configuration=false` is used...

boegel:
Test report by @boegel
 * **SUCCESS** _TensorFlow-2.4.1-foss-2020b.eb_ 

node3535.doduo.os - Linux RHEL 8.2, x86_64, AMD EPYC 7552 48-Core Processor (zen2), Python 3.6.8
See https://gist.github.com/7d7ca644ca495f04e257a8caf5814f73 for a full test report.

boegel:
I also tested this with https://github.com/easybuilders/easybuild-easyconfigs/pull/12906 using `'optarch': False` in `toolchainopts` on a Cascade Lake system, which is enough to make the tests pass there, so this is good to go.

boegel (approved):

PR #2462, opened 2021-06-08
add missing license header to AOMP easyblock + use filetools functions

boegel:
follow-up for @nordmoen's PR #2435


akesandgren (approved):
LGTM

PR #2464, opened 2021-06-09
add sanity_check_step for Aomp easyblock

nordmoen:
Moved sanity checks from easyconfig into easyblock. This change makes it easier to detect configuration changes such as the addition of CUDA.

### DIFF
--- a/easybuild/easyblocks/a/aomp.py
+++ b/easybuild/easyblocks/a/aomp.py
@@ -163,12 +163,14 @@ class EB_AOMP(Binary):
         arch = get_cpu_architecture()
         # Check architecture explicitly since Clang uses potentially
         # different names
-        if arch == X86_64:
-            arch = 'x86_64'
-        elif arch == POWER:
-            arch = 'ppc64'
-        elif arch == AARCH64:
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/a/aomp.py
```diff
@@ -154,3 +156,38 @@ def post_install_step(self):
...
+        shlib_ext = get_shared_lib_ext()
+        arch = get_cpu_architecture()
+        # Check architecture explicitly since Clang uses potentially
+        # different names
+        if arch == X86_64:
+            arch = 'x86_64'
+        elif arch == POWER:
+            arch = 'ppc64'
+        elif arch == AARCH64:
+            arch = 'aarch64'
```
smoors, line 171: this can be simplified a bit:
```suggestion
        arch_map = {
            X86_64: 'x86_64',
            POWER: 'ppc64',
            AARCH64: 'aarch64',
        }

        if arch in arch_map:
            arch = arch_map[arch]
```

smoors (approved):
lgtm

smoors:
Going in, thanks @nordmoen!

PR #2475, opened 2021-06-15
Support arbitrary version strings in OpenSSL wrapper and add minimum_openssl_version option

lexming:
As discussed in Slack, software needing OpenSSL v1.1.1(_eg_ Qt5) will not work with v1.1.0. Therefore we have to make our wrapper aware of the full version of OpenSSL.

Since the soname of `libssl` is limited to a major minor version, the system will only provide a single `libssl.so.1.1` in any case. So the existing code to find the library file still applies. All that is needed is an extra check on the version string contained in the shared object.

Changelog:
* the version of the wrapper easyconfig can have any depth and it determines the minimum version of OpenSSL required in the system
* regex the full version string from OpenSSL library files in the system
* check that the version of the system library fulfils the version of the wrapper
* apply the same regex used for the libraries to the `opensslv.h` header
* check that the version in the header matches the version of the OpenSSL library
* stop the search process in the system as soon as any component is missing

update in https://github.com/easybuilders/easybuild-easyblocks/pull/2475/commits/6d5124e919efb9de7397b823c87c329af8cb32d8
* search the version string in both `libssl` and `libcrypto` as some distros ship `libssl` without any version string

update in https://github.com/easybuilders/easybuild-easyblocks/pull/2475/commits/044ef733482c8a1de3748275c7e2c9a2273e7329, https://github.com/easybuilders/easybuild-easyblocks/pull/2475/commits/70d0186bf5b11bedc57d7ece5f9911cebbebbab4, https://github.com/easybuilders/easybuild-easyblocks/pull/2475/commits/565694963223d9792762d72ccf9b821fc049ea4c
* add extra option `minimum_openssl_version` to explicitly set the minimum version of OpenSSL required in the system
* `minimum_openssl_version` can only increase the depth of the wrapper easyconfig version

**Existing easyconfigs with just `major.minor` versions continue to work.**

Example of the results obtained with different combinations of OpenSSL versions:
| Wrapper Easyconfig Version | minimum_openssl_version | System OpenSSL Version | Result |
| --------------- | --------------- | --------------- | --------------- |
| 1.0 | none | 1.0.2k | wrapped |
| 1.0 | 1.0.2u | 1.0.2k | built in EB |
| 1.0 | 1.1 | 1.0.2k | error |
| 1.0.2 | none | 1.0.2k | wrapped |
| 1.0.2u | none | 1.0.2k | built in EB |
| 1.1 | none | 1.0.2k | built in EB |
| 1.1 | none | 1.1.0k | wrapped |
| 1.1 | none | 1.1.1g | wrapped |
| 1.1 | 1.1.1 | 1.1.0k | built in EB |
| 1.1 | 1.1.1 | 1.1.1g | wrapped |
| 1.1 | 1.1.1k | 1.1.1g | built in EB |
| 1.1 | 1.0 | 1.1.1g | error |
| 1.1 | 1.2 | 1.1.1g | error |
| 1.1.1 | none | 1.1.0k | built in EB |
| 1.1.1 | none | 1.1.1g | wrapped |
| 1.1.1k | none | 1.1.1g | built in EB |

### DIFF
--- a/easybuild/easyblocks/o/openssl_wrapper.py
+++ b/easybuild/easyblocks/o/openssl_wrapper.py
@@ -43,9 +43,18 @@ from easybuild.tools.systemtools import DARWIN, LINUX, get_os_type, get_shared_l
 
 class EB_OpenSSL_wrapper(Bundle):
     """
-    Locate the installation files of OpenSSL in the host system
-    If available, wrap the system OpenSSL by symlinking all installation files
-    Fall back to the bundled component otherwise
+    Find path to installation files of OpenSSL in the host system. Checks in
+    order: library files defined in 'openssl_libs', engines libraries, header
+    files and executables. Any missing component will trigger an installation
+    from source of the fallback component.
+    Libraries are located by soname using the major and minor subversions of
+    the wrapper version. The full version of the wrapper or the option
+    'minimum_openssl_version' determine the minimum required version of OpenSSL
+    in the system. The wrapper checks for version strings in the library files
+    and the opensslv.h header.
+    If OpenSSL in host systems fulfills the version requirements, wrap it by
+    symlinking all installation files. Otherwise fall back to the bundled
+    component.
     """
 
     @staticmethod
--- a/easybuild/easyblocks/o/openssl_wrapper.py
+++ b/easybuild/easyblocks/o/openssl_wrapper.py
@@ -34,10 +34,11 @@ from distutils.version import LooseVersion
 
 from easybuild.easyblocks.generic.bundle import Bundle
 from easybuild.framework.easyconfig import CUSTOM
+from easybuild.tools.build_log import EasyBuildError, print_warning
 from easybuild.tools.filetools import change_dir, expand_glob_paths, mkdir, read_file, symlink, which
+from easybuild.tools.py2vs3 import string_type
 from easybuild.tools.run import run_cmd
 from easybuild.tools.systemtools import DARWIN, LINUX, get_os_type, get_shared_lib_ext, find_library_path
-from easybuild.tools.build_log import EasyBuildError, print_warning
 
 
 class EB_OpenSSL_wrapper(Bundle):
@@ -53,6 +54,7 @@ class EB_OpenSSL_wrapper(Bundle):
         extra_vars = Bundle.extra_options(extra_vars=extra_vars)
         extra_vars.update({
             'wrap_system_openssl': [True, 'Detect and wrap OpenSSL installation in host system', CUSTOM],
+            'minimum_openssl_version': [None, 'Minimum version of OpenSSL required in host system', CUSTOM],
         })
         return extra_vars
 
@@ -61,11 +63,21 @@ class EB_OpenSSL_wrapper(Bundle):
         super(EB_OpenSSL_wrapper, self).__init__(*args, **kwargs)
 
         # Version checks
-        subversions = self.version.split('.')
+        minimum_version = self.cfg.get('minimum_openssl_version')
+
+        if not minimum_version:
+            minimum_version = self.version
+        elif not isinstance(minimum_version, string_type):
+            minimum_version = str(minimum_version)
+
         try:
+            subversions = minimum_version.split('.')
             majmin_version = '%s.%s' % (subversions[0], subversions[1])
-        except IndexError:
-            raise EasyBuildError("OpenSSL version has to contain at least its minor subversion: %s", self.version)
+        except (AttributeError, IndexError):
+            err_msg = "Required OpenSSL version has to contain at least its minor subversion: %s"
+            raise EasyBuildError(err_msg, minimum_version)
+        else:
+            self.log.debug("Requiring minimum OpenSSL version: %s", minimum_version)
 
         openssl_version_regex = re.compile(r'OpenSSL\s+([0-9]+\.[0-9]+(\.[0-9]+.)*)', re.M)
 
@@ -137,14 +149,14 @@ class EB_OpenSSL_wrapper(Bundle):
                         dbg_msg = "System OpenSSL library '%s' does not contain any recognizable version string"
                         self.log.debug(dbg_msg, system_solib)
                     else:
-                        if LooseVersion(openssl_version) >= LooseVersion(self.version):
+                        if LooseVersion(openssl_version) >= LooseVersion(minimum_version):
                             dbg_msg = "System OpenSSL library '%s' version %s fulfills requested version %s"
-                            self.log.debug(dbg_msg, system_solib, openssl_version, self.version)
+                            self.log.debug(dbg_msg, system_solib, openssl_version, minimum_version)
                             self.system_ssl['lib'] = system_solib
                             break
                         else:
                             dbg_msg = "System OpenSSL library '%s' version %s is older than requested version %s"
-                            self.log.debug(dbg_msg, system_solib, openssl_version, self.version)
+                            self.log.debug(dbg_msg, system_solib, openssl_version, minimum_version)
                 else:
                     # one of the OpenSSL libraries is missing, switch to next group of versioned libs
                     break
--- a/easybuild/easyblocks/o/openssl_wrapper.py
+++ b/easybuild/easyblocks/o/openssl_wrapper.py
@@ -89,8 +89,8 @@ class EB_OpenSSL_wrapper(Bundle):
         if majmin_version in openssl_libext and os_type in openssl_libext[majmin_version]:
             # generate matrix of versioned .so filenames
             system_versioned_libs = [
-                ['%s.%s' % (lib, ext) for ext in openssl_libext[majmin_version][os_type]]
-                for lib in openssl_libs
+                ['%s.%s' % (lib, ext) for lib in openssl_libs]
+                for ext in openssl_libext[majmin_version][os_type]
             ]
             self.log.info("Matrix of version library names: %s", system_versioned_libs)
         else:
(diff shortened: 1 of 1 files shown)


Micket:
I have some concerns. I suggested in slack
> We can 
> 1. Consider the name OpenSSL-1.1.eb to refer to the shlib name
> 2. Introduce a minimum actual specific version required for consideration of wrapping (minimum_wrapped_version = 1.1.1g ) and somehow check for that against openssl version  or something. Else, revert to building it from source.

Because I think this approach has some issues
1. To fix what we have already shipped, we would have to replace the existing OpenSSL-1.1.eb easyconfig since it is insufficient, and replace the name in all the easyconfigs that depend on it so far and having everyone rebuild the modules for them. Or make the default easyconfigs unusable for those affected. 
2. The easyconfig would presumably be named something like `OpenSSL-1.1.1g`, but, it might actually be actually be OpenSSL 1.1.1k (built or wrapped)? Or maybe I misunderstand the code (what happens if system is 1.1.1k and eb is 1.1.1g?). If so, then we can't specify anything that works on more than a single OS.
3. Or if you actually do build the version you specify in self.version, you can never do any in-place rebuild-updates without it starting to mismatch

I think having the separate "minimum_wrapped_version" parameter sorts out these issues. 
It could default to `self.minimum_wrapped_version = self.version`, else just verify that `self.minimum_wrapped_version.startswith(self.version)` and proceed to verify against this more specific version. Then we could enrich the existing easyconfigs with this parameter.

lexming:
@Micket with this PR as it is, it will indeed be complicated to push an existing installation using the v1.1 wrapper to v1.1.1. However, I think that it is not necessary to go that way. Existing installations are fine in my opinion. We do not currently have any easyconfig in the tree that needs OpenSSL v1.1.1. So I would not touch the current v1.1 wrapper used in 2021a.

If there is an agreement to set a minimum version with an extra option such as `minimum_wrapped_version` in the wrapper, I'll add it of course. With the current code is quite simple to do. But using an internal version for OpenSSL has also its own issues. The main one is that it is very hard to notice when the wrapper gets updated. For instance, a probable scenario is
* toolchain releases with wrapper version 1.x and internal version `1.x.a`
* months later new software in toolchain needs OpenSSL version `1.x.b` (similar to the situation with Qt5 now)
* that new software gets in the tree and wrapper has the internal version bumped to `1.x.b`
* I, as a user of EB, will only see the easyconfig of the new software landing in the tree (because I blindly pulled the branch or I just don't read changelogs of new releases)
* I'll try to install the new software and it will fail, because I already installed the wrapper months ago and nothing will trigger its re-installation

I would follow the same rules that we have for other software, we set a version per toolchain and we stick to it. In this case, v1.1 for 2021a (which is equivalent to a minimum `1.1.0` version) and we can jump to v1.1.1 for 2021b. The current PR already supports using a letter revision as version of the wrapper, so we could even require minimum revisions of OpenSSL in future toolchains.

Micket:
Just to mention this in the PR for posterity; as far as I know OpenSSL is pretty far from semver. `1.1.1` is an LTS release, while `1.1.0` wasn't. They are even adding functions to the letter-versions, `1.1.1k` has functionality that `1.1.1g` lacks. 
1.1.0 series is already out of support (last update almost 2 years ago) and shouldn't be used by recommendation of OpenSSL developers (and it rarely is, which is probably why we missed it).
I also suspect that there might never be a 1.2, or even a 1.1.2 version, but next jump would be to 3.0.0 so in practice, this is a really specific case we are covering here; making sure those on an outdated 1.1.0 release instead gets an up-to-date version.

We have actually already messed in the 4.4.0 release here since it does already include an implicit assumption on >=1.1.1 with Qt5 5.15.2, so something needs changing, and I don't think we ever want 1.1.0.

I would also find it a bit strange if versions don't actually match exactly if we go with "1.1.1x/1.1.1"  versions compared to the intentionally vague 1.1. 
I considered this to be one where rebuilding isn't totally out of the question, if for no other reason than to just keep up with security fixes.
This does happen every now and then, for Java, R, and whenever we discover some issue we can't really resolve any other way. 

Perhaps some third/fourth/fifth opinions on all of this. @boegel maybe?

lexming:
I saw that `libssl` in Fedora does not contain any version string, but `libcrypto` does. So https://github.com/easybuilders/easybuild-easyblocks/pull/2475/commits/6d5124e919efb9de7397b823c87c329af8cb32d8 changes the search to loop over all known libraries in OpenSSL until it finds one with the version string.

> We have actually already messed in the 4.4.0 release here since it does already include an implicit assumption on >=1.1.1 with Qt5 5.15.2, so something needs changing, and I don't think we ever want 1.1.0.

Oh right, I missed that merge. Then it's already an issue for 2021a. It's not ideal, but in that case the best solution might indeed be to add that internal minimum version to the 1.1 wrapper. I'll look into it.

> This does happen every now and then, for Java, R, and whenever we discover some issue we can't really resolve any other way.

For Java it's different because it is very unlikely that anything breaks if the Java installation is not kept up to date. With OpenSSL, software will start to not build if not updated properly.

lexming:
@Micket added the extra option `minimum_openssl_version` to the wrapper as requested. It sets the minimum version of OpenSSL required in the system, which the easyblock checks in the libraries and headers of OpenSSL. `minimum_openssl_version` has to match the major and minor version of the wrapper easyconfig.

Updated the PR description and table with example cases. Ready for tests.

lexming:
Test report by @lexming
 * **SUCCESS** _OpenSSL-1.0.eb_ 
 * **SUCCESS** _OpenSSL-1.1.eb_ 

node312.hydra.os - Linux centos linux 7.9.2009, x86_64, Intel(R) Xeon(R) Gold 6148 CPU @ 2.40GHz, Python 2.7.5
See https://gist.github.com/e999b48d6526fea3aa2f80543d7f597e for a full test report.

lexming:
Test report by @lexming
 * **SUCCESS** _OpenSSL-1.0.eb_ 
 * **SUCCESS** _OpenSSL-1.1.eb_ 

login3.phoenix.os - Linux centos linux 7.9.2009, x86_64, Intel(R) Xeon(R) Gold 6148 CPU @ 2.40GHz, Python 2.7.5
See https://gist.github.com/4da869a4f6c8daee40669561ec84414e for a full test report.

lexming:
Test report by @lexming
 * **SUCCESS** _OpenSSL-1.0.eb_ 
 * **SUCCESS** _OpenSSL-1.1.eb_ 

login2.cerberus.os - Linux centos linux 7.9.2009, x86_64, Intel(R) Xeon(R) Gold 6126 CPU @ 2.60GHz, Python 2.7.5
See https://gist.github.com/1cdb5971ac7639ce5d2cf1d586efdff7 for a full test report.

lexming:
Test report by @lexming
 * **SUCCESS** _OpenSSL-1.0.eb_ 
 * **SUCCESS** _OpenSSL-1.1.eb_ 

login3.phoenix.os - Linux centos linux 7.9.2009, x86_64, Intel(R) Xeon(R) Gold 6148 CPU @ 2.40GHz, Python 2.7.5
See https://gist.github.com/cc1639e3cad6430c7192586679a62da0 for a full test report.

lexming:
Test report by @lexming
 * **SUCCESS** _OpenSSL-1.0.eb_ 
 * **SUCCESS** _OpenSSL-1.1.eb_ 

login3.phoenix.os - Linux centos linux 7.9.2009, x86_64, Intel(R) Xeon(R) Gold 6148 CPU @ 2.40GHz, Python 2.7.5
See https://gist.github.com/359626d332ef7f4d975a955f955970ab for a full test report.

lexming:
Test report by @lexming
 * **SUCCESS** _OpenSSL-1.0.eb_ 
 * **SUCCESS** _OpenSSL-1.1.eb_ 

angook - Linux Fedora 33, x86_64, Intel(R) Core(TM) i7-8550U CPU @ 1.80GHz, Python 3.9.5
See https://gist.github.com/ff4f8bb2c7b5240e866ca8ecd9bf9e13 for a full test report.

Micket (approved):
lgtm

Micket:
Test report by @Micket
 * **FAIL (unhandled exception: [Errno 2] No such file or directory: '/apps/Test2/software/OpenSSL/1.1')Traceback (most recent call last):
  File "/apps/Common/software/EasyBuild/20210611-dev/lib/python3.6/site-packages/easybuild/main.py", line 117, in build_and_install_software
  File "/apps/Common/software/EasyBuild/20210611-dev/lib/python3.6/site-packages/easybuild/framework/easyblock.py", line 3646, in build_and_install_one
  File "/apps/Common/software/EasyBuild/20210611-dev/lib/python3.6/site-packages/easybuild/framework/easyblock.py", line 3537, in run_all_steps
  File "/apps/Common/software/EasyBuild/20210611-dev/lib/python3.6/site-packages/easybuild/framework/easyblock.py", line 3390, in run_step
    step_method(self)()
  File "/apps/Common/software/EasyBuild/20210611-dev/lib/python3.6/site-packages/easybuild/framework/easyblock.py", line 1872, in check_readiness_step
    self.remove_module_file()
  File "/apps/Common/software/EasyBuild/20210611-dev/lib/python3.6/site-packages/easybuild/framework/easyblock.py", line 1806, in remove_module_file
    elif os.path.exists(old_installdir) and not os.path.samefile(old_installdir, self.installdir):
  File "/usr/lib64/python3.6/genericpath.py", line 97, in samefile
    s2 = os.stat(f2)
FileNotFoundError: [Errno 2] No such file or directory: '/apps/Test2/software/OpenSSL/1.1'
** _OpenSSL-1.1.eb_ 

alvis-c1 - Linux centos linux 7.9.2009, x86_64, Intel Xeon Processor (Skylake), Python 3.6.8
See https://gist.github.com/41adf1157ba68bb9508897a974068198 for a full test report.

Micket:
Test report by @Micket
 * **SUCCESS** _OpenSSL-1.1.eb_ 

alvis-c1 - Linux centos linux 7.9.2009, x86_64, Intel Xeon Processor (Skylake), Python 3.6.8
See https://gist.github.com/a36937b2bc19f6cda57b75410fb82f44 for a full test report.

PR #2487, opened 2021-06-23
add custom easyblock for CRISPR-DAV

deniskristak:
needed to create easyblock due to complex source file changes

### DIFF
--- a/easybuild/easyblocks/c/crisprdav.py
+++ b/easybuild/easyblocks/c/crisprdav.py
@@ -39,7 +39,7 @@ class EB_crisprdav(Binary):
 
     def post_install_step(self):
         """Create config.txt files"""
-        crisprdav_installdir = self.installdir
+        crisprdav_installdir_ex = os.path.join(self.installdir, 'Examples/example1/')
         abra2_dir = os.environ['EBROOTABRA2']
         prinseq_dir = os.environ['EBROOTPRINSEQ']
         flash_dir = os.environ['EBROOTFLASH']
@@ -50,10 +50,10 @@ class EB_crisprdav(Binary):
         # writing to Example directory to run sanity checks
         example_config_file = os.path.join(self.installdir, 'Examples/example1/conf.txt')
 
-        CONFIG_FILE_TEXT_FORMATTED = CONFIG_FILE_TEXT.format(crisprdav_installdir=crisprdav_installdir,
+        CONFIG_FILE_TEXT_FORMATTED = CONFIG_FILE_TEXT.format(crisprdav_installdir_ex=crisprdav_installdir_ex,
                                                              abra2_dir=abra2_dir, prinseq_dir=prinseq_dir,
                                                              flash_dir=flash_dir)
-        FASTQ_LIST_FILE_TEXT_FORMATTED = FASTQ_LIST_FILE_TEXT.format(crisprdav_installdir=crisprdav_installdir)
+        FASTQ_LIST_FILE_TEXT_FORMATTED = FASTQ_LIST_FILE_TEXT.format(crisprdav_installdir_ex=crisprdav_installdir_ex)
         # according to docs, we have to setup conf.txt so that it contains correct paths to dependencies
         # https://github.com/pinetree1/crispr-dav/blob/master/Install-and-Run.md
         # User then has to change conf.txt to include paths to genomes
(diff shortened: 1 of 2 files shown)


easybuild/easyblocks/c/crisprdav.py
```diff
@@ -0,0 +1,100 @@
...
+EasyBuild support for building and installing Metagenome-Atlas, implemented as an easyblock.
+
+@author: Denis Kristak (INUITS)
+"""
+import os
+from easybuild.tools.filetools import write_file
+from easybuild.easyblocks.generic.binary import Binary
+
+
+class EB_crisprdav(Binary):
```
boegel, line 35: please rename to `EB_CRISP_minus_DAV` (see also https://github.com/easybuilders/easybuild-easyconfigs/pull/13139)

easybuild/easyblocks/c/crisprdav.py
```diff
@@ -0,0 +1,100 @@
...
+        write_file(config_file, CONFIG_FILE_TEXT_FORMATTED)
+        write_file(example_config_file, CONFIG_FILE_TEXT_FORMATTED)
+        # also used for sanity checking with Examples/ folder
+        write_file(example_fastq_list_file, FASTQ_LIST_FILE_TEXT_FORMATTED)
+
+
+# obtained from
+# https://github.com/pinetree1/crispr-dav/blob/master/conf.txt
+# see the original file for detailed commentary
+CONFIG_FILE_TEXT = """
```
boegel, line 69: constants should be defined above the `class` definition

easybuild/easyblocks/c/crisprdav.py
```diff
@@ -0,0 +1,100 @@
...
+# change this part to contain paths to your genomes
+[genomex]
+ref_fasta = {crisprdav_installdir_ex}/genome/genomex.fa
+bwa_idx = {crisprdav_installdir_ex}/genome/genomex.fa
+refGene = {crisprdav_installdir_ex}/genome/refgenex.txt
+
+[app]
+abra = {abra2_dir}/abra2-2.23.jar
```
boegel, line 77: Wouldn't it be better to copy the `conf.txt`, and do a couple of search-and-replace operations on that?

Shipping a copy of the `conf.txt` in this easyblock makes it a maintenance problem going forward...

easybuild/easyblocks/c/crisprdav.py
```diff
@@ -0,0 +1,100 @@
...
+realign_flag = Y
+min_mapq = 20
+wing_length = 100
+high_res = 0
+parallel_env = orte
+cores_per_job = 12
+"""
+
+FASTQ_LIST_FILE_TEXT = r"""
+sample1	{crisprdav_installdir_ex}/rawfastq/sample1_R1.fastq.gz	{crisprdav_installdir_ex}/rawfastq/sample1_R2.fastq.gz
```
boegel, line 96: Isn't it better to *generate* this file?

easybuild/easyblocks/c/crisprdav.py
```diff
@@ -0,0 +1,100 @@
...
+
+
+class EB_crisprdav(Binary):
+    """
+    Support for building/installing crispr-dav.
+    """
+
+    def post_install_step(self):
+        """Create config.txt files"""
+        crisprdav_installdir_ex = os.path.join(self.installdir, 'Examples/example1/')
```
boegel, line 42: please avoid using `/` in `os.path.join` (same below):

```python
crisprdav_installdir_ex = os.path.join(self.installdir, 'Examples', 'example1')
```

easybuild/easyblocks/c/crisprdav.py
```diff
@@ -0,0 +1,100 @@
...
+        abra2_dir = os.environ['EBROOTABRA2']
+        prinseq_dir = os.environ['EBROOTPRINSEQ']
+        flash_dir = os.environ['EBROOTFLASH']
```
boegel, line 45: It's better to use the `get_software_root` function here, and you should also *check* that the result is not `None`, which verifies that these are indeed included as dependencies.

easybuild/easyblocks/c/crisprdav.py
```diff
@@ -0,0 +1,100 @@
...
+# EasyBuild is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with EasyBuild.  If not, see <http://www.gnu.org/licenses/>.
+##
+"""
+EasyBuild support for building and installing Metagenome-Atlas, implemented as an easyblock.
```
boegel, line 26: Please fix, this easyblock is for `CRISPR-DAV`, not for `Metagenome-Atlas`

boegel (changes requested):

verdurin:
@deniskristak I tried building the associated easyconfig and saw this error:
```
eb --from-pr=13139 --robot --debug --upload-test-report --include-easyblocks-from-pr 2487
== Temporary log file in case of crash /tmp/eb-jdvm00ah/easybuild-q0xclof4.log
== easyblock crispr_dav.py included from PR #2487
ERROR: Failed to process easyconfig /tmp/eb-jdvm00ah/files_pr13139/c/crispr-dav/crispr-dav-2.3.4-foss-2020b.eb: Failed to obtain class for EB_crisprdav easyblock (not available?): No module named 'easybuild.easyblocks.crisprdav'
```

deniskristak:
@verdurin correct, i forgot to remove an easyconfig with an the old name. Should be working now

boegel:
close/re-open to wake up GitHub Actions...

boegel (approved):

PR #2491, opened 2021-06-26
add custom easyblock for building UCX plugins

Micket:
Custom easyblock for building UCX-plugins.
Easyconfig in https://github.com/easybuilders/easybuild-easyconfigs/pull/13260


boegel:
@Micket Tests are failing because `--module-only --force` is failing, since it trips over the failing `ucx_info -b` command being run in `make_module_extra`.
We may need to inject a mocked `ucx_info` in the test to bypass that, I don't see another easy way out...

Micket:
> @Micket Tests are failing because `--module-only --force` is failing, since it trips over the failing `ucx_info -b` command being run in `make_module_extra`.
> We may need to inject a mocked `ucx_info` in the test to bypass that, I don't see another easy way out...

Yeah this is a general issue for any module generation that isn't entirely static. We actually have this problem in a couple of easyblocks; TensorFlow, TBB, netcdf-python are all broken for `--module-only`, but they hush up the test suite by setting bad default values https://github.com/easybuilders/easybuild-easyblocks/blob/29d2aab1285ecb8e7bec38a3cf520abe07fedd3b/easybuild/easyblocks/generic/pythonpackage.py#L290  so the test suite passes, but --module-only is unusable.

boegel (approved):
lgtm

PR #2493, opened 2021-06-28
Use PYPI_SOURCE as the default for source_urls of extensions of Python easyconfigs

Flamefire:
This is to simplify Python ECs a bit and does basically the same as the PythonPackage itself does.
See https://github.com/easybuilders/easybuild-easyconfigs/pull/13253

Why this didn't work before is because the PythonPackage (likely) gets the source_urls from the parent EC (in this case Python) and hence doesn't use the new default

This fixes this by defaulting the source-urls to PYPI_SOURCE

### DIFF
--- a/easybuild/easyblocks/p/python.py
+++ b/easybuild/easyblocks/p/python.py
@@ -43,6 +43,7 @@ from distutils.version import LooseVersion
 import easybuild.tools.environment as env
 from easybuild.easyblocks.generic.configuremake import ConfigureMake
 from easybuild.framework.easyconfig import CUSTOM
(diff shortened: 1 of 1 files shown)


boegel (approved):
lgtm

boegel:
Test report by @boegel
 * **SUCCESS** _Python-3.9.5-GCCcore-10.3.0.eb_ 

node3615.doduo.os - Linux RHEL 8.2, x86_64, AMD EPYC 7552 48-Core Processor (zen2), Python 3.6.8
See https://gist.github.com/20831b1c8563d85619b2373f3a21840f for a full test report.

PR #2496, opened 2021-06-28
Fix for GCC10

surak:
GCC10 has more things inside the include directory, so the number is clearly bigger than 1 - but for old GCCs, the number is 1. So...

### DIFF
--- a/easybuild/easyblocks/b/bazel.py
+++ b/easybuild/easyblocks/b/bazel.py
@@ -78,7 +78,7 @@ class EB_Bazel(EasyBlock):
         # CROSSTOOL script is no longer there in Bazel 0.24.0
         if LooseVersion(self.version) < LooseVersion('0.24.0'):
             res = glob.glob(os.path.join(gcc_root, 'lib', 'gcc', '*', gcc_ver, 'include'))
-            if res and len(res) == 1:
+            if res and len(res) > 0:
                 gcc_lib_inc = res[0]
             else:
                 raise EasyBuildError("Failed to pinpoint location of GCC include files: %s", res)


Flamefire (changes requested):
As written in Slack already:
This issue can only occur when using an old Bazel (<0.24) with a new GCC (>=9.3 where we added nvptx). I do not think it is worth supporting this combination and hence would rather not accept this.
However if that specific combination(s) are to be supported then the `gcc_lib_inc = res[0]` or (maybe better) the glob needs refinement so it doesn't use the nvptx folder but the actually wanted folder.

surak:
@Flamefire that is not true. I had this failure confirmed with all versions from 3.6.0 to 4.1.0.

https://github.com/bazelbuild/bazel/issues/12056

Flamefire:
The link you posted is unrelated to this patch. Did you mispaste?

Can you post the exact commands used and the effective easyconfig you used?
I keep telling you that this part of the code is unreachable for Bazel >=0.24 ECs due to `if LooseVersion(self.version) < LooseVersion('0.24.0'):` So clearly your issue is somewhere else as this part of the code changed here does NOT need patching for Bazel >= 0.24

surak:
Command? "eb Bazel-3.6.0-GCCcore-9.3.0.eb"

The easyconfig:

```
name = 'Bazel'
version = '3.6.0'

homepage = 'http://bazel.io/'
description = """Bazel is a build tool that builds code quickly and reliably.
It is used to build the majority of Google's software."""

toolchain = {'name': 'GCCcore', 'version': '9.3.0'}

source_urls = ['https://github.com/bazelbuild/bazel/releases/download/%(version)s']
sources = ['%(namelower)s-%(version)s-dist.zip']
patches = ['%(name)s-3.4.1-fix-grpc-protoc.patch']

builddependencies = [
    ('binutils', '2.34'),
    ('Python', '3.8.5'),
    ('Zip', '3.0'),
]

prebuildopts = "export BAZEL_LINKOPTS=-static-libstdc++:-static-libgcc BAZEL_LINKLIBS=-l%:libstdc++.a:-lm && "

dependencies = [('Java', '1.8', '', True)]

moduleclass = 'devel'
```

surak:
You are correct and I am wrong, and I am sorry for that. 

The error is that JSC seems to have an old version of bazel.py: https://github.com/easybuilders/JSC/blob/2020/Custom_EasyBlocks/bazel.py

PR #2506, opened 2021-07-06
Handle failure of running nvidia-smi in TensorFlow tests

Flamefire:
Correctly fall back to skipping GPU tests

### DIFF
--- a/easybuild/easyblocks/t/tensorflow.py
+++ b/easybuild/easyblocks/t/tensorflow.py
@@ -851,7 +851,7 @@ class EB_TensorFlow(PythonPackage):
                 num_gpus_to_use = 0
             else:
                 # determine number of available GPUs via nvidia-smi command, fall back to just 1 GPU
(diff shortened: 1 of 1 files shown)


ocaisa:
@surak Can you check if this works for you?

surak:
> @surak Can you check if this works for you?

Testing now, TF takes a while to install on juwels

surak:
This test passes now

ocaisa (commented):
Code changes are fine but it is takes a bit of digging to figure out why this was the necessary change. My digging said it was to avoid https://github.com/easybuilders/easybuild-framework/blob/develop/easybuild/tools/run.py#L574, I think this warrants a comment and I would still probably explicitly include `log_all=False` even though this is the default since that also features in the relevant line.

ocaisa:
I guess this is a very niche case where `nvidia-smi` is available but there are no GPUs

Flamefire:
@ocaisa Makes sense. Added.

ocaisa (approved):

PR #2541, opened 2021-08-12
preparation for MotionCor2 1.4.4

stefan-wolfsheimer:


stefan-wolfsheimer:
see also:
https://github.com/easybuilders/easybuild-easyconfigs/pull/13716

easybuild/easyblocks/m/motioncor2.py
```diff
@@ -56,6 +56,16 @@ def __init__(self, *args, **kwargs):
         if (LooseVersion(self.version) == LooseVersion("1.3.1")):
             self.motioncor2_verstring = "v%s" % self.version
 
+    @staticmethod
+    def extra_options(extra_vars=None):
+        extra = {
+            'skip_rpath_check': [False, "Skip rpath check", None],
```
boegel, line 62: I would clarify the help string a bit here, and maybe also rename to `skip_rpath_sanity_check`

```python
"Skip RPATH sanity check"
```

@stefan-wolfsheimer Can you clarify why you need this option, in which context?

easybuild/easyblocks/m/motioncor2.py
```diff
@@ -91,12 +101,16 @@ def install_step(self):
...
+            pattern2 = os.path.join(self.builddir,
+                                    "{}_{}".format(self.name, self.version),
+                                    '%s*' % self.motioncor2_bin)
+            matches = glob.glob(pattern1) + glob.glob(pattern2)
             if len(matches) == 1:
                 src_mc2_bin = matches[0]
             else:
                 raise EasyBuildError(
-                    "Found multiple, or no, matching MotionCor2 binary named %s*" % self.motioncor2_bin
+                    "Found multiple, or no, matching MotionCor2 binary named %s" % pattern
```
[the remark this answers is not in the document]
boegel: @stefan-wolfsheimer This is clearly wrong, since there is no variable named `pattern`....
You probably still want to use `self.motioncor2_bin` like it was before (and leave the `*` in place too)?

easybuild/easyblocks/m/motioncor2.py
```diff
@@ -91,12 +101,16 @@ def install_step(self):
         """
 
         if (LooseVersion(self.version) >= LooseVersion("1.4")):
-            matches = glob.glob(os.path.join(self.builddir, '%s*' % self.motioncor2_bin))
+            pattern1 = os.path.join(self.builddir, '%s*' % self.motioncor2_bin)
+            pattern2 = os.path.join(self.builddir,
```
boegel, line 105: Maybe this deserves a comment: it seems like you're expecting to find the binary either directly in the build directory, or in a subdirectory named `<name>_<version>`?

Do you have any idea in which case it's in a subdirectory exactly?

easybuild/easyblocks/m/motioncor2.py
```diff
@@ -140,5 +154,9 @@ def sanity_check_step(self):
...
             'dirs': []
         }
-
-        super(EB_MotionCor2, self).sanity_check_step(custom_paths)
+        for f in custom_paths['files']:
+            fp = os.path.join(self.installdir, f)
+            if not os.path.isfile(fp):
+                raise EasyBuildError("File not installed {}".format(fp))
+        if not self.cfg.get('skip_rpath_check', False):
+            super(EB_MotionCor2, self).sanity_check_step(custom_paths)
```
boegel, line 162: There's an easier way to only skip the RPATH sanity check:

```python
def sanity_check_rpath(self, *args, **kwargs):
    """Custom implementation of RPATH sanity check: allow skipping via skip_rpath_sanity_check."""
    if self.cfg.get('skip_rpath_sanity_check', False):
        self.log.info("Skipping RPATH sanity check, as specified via 'skip_rpath_sanity_check'...")
    else:
        super(EB_MotionCor2, self).sanity_check_rpath(*args, **kwargs)
```

Then you don't have to partially replicate the sanity check for files yourself either...

And maybe we should make `skip_rpath_sanity_check` a general easyconfig parameter in framework instead?

boegel (changes requested):

easybuild/easyblocks/m/motioncor2.py
MaximeMoge: I just tried to build MotionCor2 and the binary is in a subdirectory in the zipball distributed on https://emcore.ucsf.edu/ucsf-software. As far as I can tell this has changed since v1.4.0, for older version the binary was *not* in a subdirectory.

akesandgren: 1.4.2 doesn't have a subdir so this needs more specified version checking.

akesandgren (changes requested):

easybuild/easyblocks/m/motioncor2.py
```diff
@@ -91,7 +91,7 @@ def install_step(self):
         """
 
         if (LooseVersion(self.version) >= LooseVersion("1.4")):
-            matches = glob.glob(os.path.join(self.builddir, '%s*' % self.motioncor2_bin))
+            matches = glob.glob(os.path.join(self.builddir, '%s_%s' % (self.name, self.version), '%s*' % self.motioncor2_bin))
```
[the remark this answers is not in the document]
akesandgren: This is now wrong for 1.4.2

akesandgren (changes requested):

easybuild/easyblocks/m/motioncor2.py
MaximeMoge: hum right... However 1.4.0 does have a subdirectory.
Then I would say @stefan-wolfsheimer was right checking both directly in the build directory and in the subdirectory.

easybuild/easyblocks/m/motioncor2.py
```diff
@@ -91,12 +91,17 @@ def install_step(self):
...
+                                    "{}_{}".format(self.name, self.version),
+                                    '%s*' % self.motioncor2_bin)
+            matches = glob.glob(pattern1) + glob.glob(pattern2)
+
             if len(matches) == 1:
                 src_mc2_bin = matches[0]
             else:
                 raise EasyBuildError(
-                    "Found multiple, or no, matching MotionCor2 binary named %s*" % self.motioncor2_bin
+                    "Found multiple, or no, matching MotionCor2 binary named %s or %s" % (pattern1, patter2)
```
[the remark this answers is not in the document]
akesandgren: And once again, please do not change this.

MaximeMoge:
I hope the latest commit answers all comments :)
- I added a comment to clarify why we look for the binary both directly in the builddir and in a subdirectory.
- I reverted the unnecessary change pointed out by akesandgren

easybuild/easyblocks/m/motioncor2.py
```diff
@@ -142,3 +149,10 @@ def sanity_check_step(self):
         }
 
         super(EB_MotionCor2, self).sanity_check_step(custom_paths)
+
+    def sanity_check_rpath(self, *args, **kwargs):
+        """Custom implementation of RPATH sanity check: allow skipping via skip_rpath_sanity_check."""
```
akesandgren, line 154: Please add comment on why we need to skip the rpath check

akesandgren (approved):

easybuild/easyblocks/m/motioncor2.py
MaximeMoge: Actually looking at this again I think this is a workaround for a broader issue with RPATH for binary installs, and it is not the right approach:
- we use RPATH on our supercomputer at SURF, and we usually fix the issues with binary install by customizing the easyconfig (or adding hooks) to run `patchelf` commands to add the missing rpaths in the binary.
- this is an issue for all binary installs. Ideally, the rpath check in the case of binary install would also automatically fix the missing rpaths using `patchelf`. However `patchelf` cannot be used in the easyblock as it is not necessarily available on the system. So we have to do it at the easyconfig level.
- In the case of the current PR, this custom `sanity_check_rpath` is in my opinion not the right solution. In fact this is a binary install, and the `EB_MotionCor2` class should derive from the `PackedBinary` class instead of directly from the `EasyBlock` class. Since `PackedBinary` skips the rpath check, that would remove the need for a custom `sanity_check_rpath`.

A real fix for binary installs using RPATH should in my opinion be implemented in the `Binary` class rather than in software specific easyblocks, so we should not change the `sanity_check_rpath` at this level.

@akesandgren: my suggestion is to make `EB_MotionCor2` derive from `PackedBinary` instead of `EasyBlock`, and remove the custom `sanity_check_rpath` completely. Does that sound good to you?

akesandgren: Not sure, I need to check what PackedBinary does. Probably won't have time the coming days/week

akesandgren:
Test report by @akesandgren
 * **SUCCESS** _MotionCor2-1.2.6-GCCcore-8.2.0.eb_ 
 * **FAIL (build issue)** _MotionCor2-1.3.2-GCCcore-8.3.0.eb_ (partial log available at https://gist.github.com/95e639edc56341e3bc6d670fb4495c94)
 * **FAIL (build issue)** _MotionCor2-1.3.1-GCCcore-8.3.0.eb_ (partial log available at https://gist.github.com/f73c5f6973749ecd9abd1a9dd07026e2)
 * **SUCCESS** _MotionCor2-1.4.2-GCCcore-10.2.0.eb_ 
 * **FAIL (build issue)** _MotionCor2-1.4.4-GCCcore-10.3.0.eb_ (partial log available at https://gist.github.com/a168a5f5c6001924bd4e5525f069eee1)

Build succeeded for 2 out of 5 (5 easyconfigs in total)
b-an02.hpc2n.umu.se - Linux Ubuntu 20.04, x86_64, Intel(R) Xeon(R) CPU E5-2690 v4 @ 2.60GHz, Python 3.8.10
See https://gist.github.com/ecfc13f664da12b48c7f89e6234a1c6d for a full test report.

akesandgren:
Test report by @akesandgren
 * **SUCCESS** _MotionCor2-1.2.6-GCCcore-8.2.0.eb_ 
 * **SUCCESS** _MotionCor2-1.3.2-GCCcore-8.3.0.eb_ 
 * **SUCCESS** _MotionCor2-1.3.1-GCCcore-8.3.0.eb_ 
 * **SUCCESS** _MotionCor2-1.4.2-GCCcore-10.2.0.eb_ 
 * **FAIL (build issue)** _MotionCor2-1.4.4-GCCcore-10.3.0.eb_ (partial log available at https://gist.github.com/91addd8047b2129cfdb3701fc1ad94d1)

Build succeeded for 4 out of 5 (5 easyconfigs in total)
b-an02.hpc2n.umu.se - Linux Ubuntu 20.04, x86_64, Intel(R) Xeon(R) CPU E5-2690 v4 @ 2.60GHz, Python 3.8.10
See https://gist.github.com/972f0044b13be6288b5ab22fe61e777f for a full test report.

easybuild/easyblocks/m/motioncor2.py
akesandgren: Lets consider that in a separate PR.

casparvl:
@akesandgren which `MotionCor2-1.4.4-GCCcore-10.3.0.eb` did you use? Because the one from https://github.com/easybuilders/easybuild-easyconfigs/pull/13716 still has `CUDAcore`, which should be replaced with `CUDA`

akesandgren:
I used my own which is a copy of the old 1.4.2 but with the required changes for use in 10.3.0 toolchain and deps updated accordingly.

akesandgren:
@casparvl Yes, using PackedBinary is probably a good idea, but as I said before, lets leave that to a separate PR.

akesandgren:
Going in, thanks @stefan-wolfsheimer!

PR #2542, opened 2021-08-16
add custom easyblock for jaxlib

deniskristak:
working jaxlib easyblock


deniskristak:
must be used with https://github.com/easybuilders/easybuild-easyconfigs/pull/13753

boegel:
I'm working on a couple of changes for this, should be done soon...

boegel:
I think it makes more sense to continue with #2545, which is based on this PR, takes a slightly different approach (cleaner), and also tackles the CUDA aspect.

So, I'll close this, thanks for the effort @deniskristak !

PR #2560, opened 2021-09-01
add support for also installing Tosca component in ABAQUS

boegel:


easybuild/easyblocks/a/abaqus.py
```diff
@@ -48,6 +49,13 @@
 class EB_ABAQUS(Binary):
     """Support for installing ABAQUS."""
 
+    @staticmethod
+    def extra_options():
+        extra_vars = {
+            'with_tosca': [False, "Enable installation of Tosca", CUSTOM],
```
boegel, line 55: As far as I know there's no downside to also installing the Tosca component (other than disk usage), so perhaps this should be enabled by default?

@branfosj Thoughts?

branfosj: I'm happy to go with this on by default. My quick test did not hit a problem.

Disk usage increased from 6GB -> 7.3GB, but if someone really want to save space they'll disable the install of the docs (nearly half the disk use).

Micket: Is it always available?

boegel: I'm now auto-enabling the installation of Tosca for sufficiently recent ABAQUS versions (>= 2020) with the changes in 465a95c, which is a good middle ground going forward imho (a.k.a. I don't want to spend time figuring out problems with older versions when also installing Tosca ;-) )

bartoldeman (approved):
LGTM

PR #2570, opened 2021-09-09
Prevent OpenMPI from failing on machines with less than 8 cores

schiotz:
Possible fix for issue https://github.com/easybuilders/easybuild-easyconfigs/issues/13945#issuecomment-916037883

It needs testing, and it might cause problems on machines with two cores and no hyperthreading.


schiotz:
Should we just close this PR, now that @ocaisa has come up with a superior solution (#2611)?

boegel:
Strangely enough I've also run into this on a system with 64 cores (AWS Graviton2 instance `c6gd.16xlarge`)...

I worked around it using `export OMPI_MCA_rmaps_base_oversubscribe=1` followed by `eb --module-only` (but we should go ahead with the fix proposed in #2611, unless we want to set `$OMPI_MCA_rmaps_base_oversubscribe` more broadly in the sanity check via EasyBuild framework when OpenMPI is used as toolchain component, since this also applies to OpenFOAM, see #2615).

boegel:
Closing this in favor of #2611, simply allowing oversubscription is a better/broader fix imho...

PR #2571, opened 2021-09-10
Let GROMACS runtime logs show EasyBuild was used

mabraham:
Knowing that EasyBuild has patched the code and organized the build is potentially valuable information for GROMACS users and developers troubleshooting their runs.

PLUMED does further patches to GROMACS, so that is expressed directly also.

### DIFF
--- a/easybuild/easyblocks/g/gromacs.py
+++ b/easybuild/easyblocks/g/gromacs.py
@@ -215,7 +215,7 @@ class EB_GROMACS(CMakeMake):
             if plumed_root:
                 plumed_version = get_software_version('PLUMED')
                 if plumed_version:
-                    gromacs_version_string_suffix = '%s-%s' % (plumed-version, VERBOSE_VERSION)
+                    gromacs_version_string_suffix = '%s-%s' % (plumed_version, VERBOSE_VERSION)
                 else:
                     gromacs_version_string_suffix = 'PLUMED-%s' % VERBOSE_VERSION
             else:
--- a/easybuild/easyblocks/g/gromacs.py
+++ b/easybuild/easyblocks/g/gromacs.py
@@ -51,7 +51,7 @@ from easybuild.tools.modules import get_software_libdir, get_software_root, get_
 from easybuild.tools.run import run_cmd
 from easybuild.tools.toolchain.compiler import OPTARCH_GENERIC
 from easybuild.tools.systemtools import X86_64, get_cpu_architecture, get_shared_lib_ext, get_cpu_features
-from easybuild.tools.version import VERBOSE_VERSION
+from easybuild.tools.version import VERBOSE_VERSION as EASYBUILD_VERSION
 
 
 class EB_GROMACS(CMakeMake):
@@ -211,11 +211,11 @@ class EB_GROMACS(CMakeMake):
 
         # Ensure that the GROMACS log files report how the code was patched
         # during the build, so that any problems are easier to diagnose.
-        if '-DGMX_VERSION_STRING_OF_FORK=' not in self.cfg['configopts']:
+        # The GMX_VERSION_STRING_OF_FORK feature is available since 2020.
+        if LooseVersion(self.version) >= LooseVersion('2020') and '-DGMX_VERSION_STRING_OF_FORK=' not in self.cfg['configopts']:
+            gromacs_version_string_suffix = 'EasyBuild-%s' % EASYBUILD_VERSION
             if plumed_root:
-                gromacs_version_string_suffix = 'PLUMED-%s-%s' % (get_software_version('PLUMED'), VERBOSE_VERSION)
-            else:
-                gromacs_version_string_suffix = VERBOSE_VERSION
+                gromacs_version_string_suffix += '-PLUMED-%s' % get_software_version('PLUMED')
             self.cfg.update('configopts', '-DGMX_VERSION_STRING_OF_FORK=%s' % gromacs_version_string_suffix)
 
         if LooseVersion(self.version) < LooseVersion('4.6'):
--- a/easybuild/easyblocks/g/gromacs.py
+++ b/easybuild/easyblocks/g/gromacs.py
@@ -213,11 +213,7 @@ class EB_GROMACS(CMakeMake):
         # during the build, so that any problems are easier to diagnose.
         if '-DGMX_VERSION_STRING_OF_FORK=' not in self.cfg['configopts']:
             if plumed_root:
-                plumed_version = get_software_version('PLUMED')
-                if plumed_version:
-                    gromacs_version_string_suffix = '%s-%s' % (plumed_version, VERBOSE_VERSION)
-                else:
-                    gromacs_version_string_suffix = 'PLUMED-%s' % VERBOSE_VERSION
+                gromacs_version_string_suffix = 'PLUMED-%s-%s' % (get_software_version('PLUMED'), VERBOSE_VERSION)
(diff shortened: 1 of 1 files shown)


mabraham:
@boegel Thanks for the guidance!

mabraham:
Only recent GROMACS versions support this, but there's no need add complexity here by limiting to those GROMACS versions, as an unused CMake variable merely issues a warning that can be ignored safely.

easybuild/easyblocks/g/gromacs.py
```diff
@@ -208,6 +209,19 @@ def configure_step(self):
             # version of GROMACS. Just prepare first part of cmd here
             plumed_cmd = "plumed-patch -p -e %s" % engine
 
+        # Ensure that the GROMACS log files report how the code was patched
+        # during the build, so that any problems are easier to diagnose.
+        if '-DGMX_VERSION_STRING_OF_FORK=' not in self.cfg['configopts']:
+            if plumed_root:
+                plumed_version = get_software_version('PLUMED')
+                if plumed_version:
+                    gromacs_version_string_suffix = '%s-%s' % (plumed_version, VERBOSE_VERSION)
```
akesandgren, line 218: @mabraham note that plumed_version does not contain the text "PLUMED" only the version, also if plumed_root is set plumed_version will always return something, so this could be reduced to just
```
plumed_version = get_software_version('PLUMED')
gromacs_version_string_suffix = 'PLUMED-%s-%s' % (plumed_version, VERBOSE_VERSION)
```
skipping the else paert of the "if plumed_version"

mabraham: Thanks, done!

akesandgren (dismissed):
LGTM

akesandgren:
Kicking off some test builds

easybuild/easyblocks/g/gromacs.py
```diff
@@ -208,6 +209,15 @@ def configure_step(self):
             # version of GROMACS. Just prepare first part of cmd here
             plumed_cmd = "plumed-patch -p -e %s" % engine
 
+        # Ensure that the GROMACS log files report how the code was patched
+        # during the build, so that any problems are easier to diagnose.
+        if '-DGMX_VERSION_STRING_OF_FORK=' not in self.cfg['configopts']:
```
boegel, line 214: @mabraham I would prefer to only use `-DGMX_VERSION_STRING_OF_FORK` if the GROMACS version actually supports it. It only avoids a warning, but it prevents questions/confusing for people who see the warning popping up in the build log.

Something like this (with the correct version, of course):

```python
        if '-DGMX_VERSION_STRING_OF_FORK=' not in self.cfg['configopts']:
            # only GROMACS >= 2019 supports specifying a version string
            if LooseVersion(self.version) >= LooseVersion('2019'):
                ...
```

easybuild/easyblocks/g/gromacs.py
```diff
@@ -51,6 +51,7 @@
 from easybuild.tools.run import run_cmd
 from easybuild.tools.toolchain.compiler import OPTARCH_GENERIC
 from easybuild.tools.systemtools import X86_64, get_cpu_architecture, get_shared_lib_ext, get_cpu_features
+from easybuild.tools.version import VERBOSE_VERSION
```
boegel, line 54: Maybe it's better to rename this on import, to avoid confusion between the GROMACS version and the EasyBuild version:

```python
from easybuild.tools.version import VERBOSE_VERSION as EASYBUILD_VERSION

**Review by boegel — CHANGES_REQUESTED** [review]

File: easybuild/easyblocks/g/gromacs.py (hunk 3 above)
Comment (akesandgren) [reply easybuild/easyblocks/g/gromacs.py:54]: +1

File: easybuild/easyblocks/g/gromacs.py (hunk 2 above)
Comment (akesandgren) [reply easybuild/easyblocks/g/gromacs.py:214]: I'd switch those two if-stmts around, first check for gromacs version

Comment (boegel) [reply easybuild/easyblocks/g/gromacs.py:214]: Yeah, both work, doesn't matter much. It's perhaps clearer to do the version check first.

File: easybuild/easyblocks/g/gromacs.py [hunk 4]
```diff
@@ -208,6 +209,15 @@ def configure_step(self):
             # version of GROMACS. Just prepare first part of cmd here
             plumed_cmd = "plumed-patch -p -e %s" % engine
 
+        # Ensure that the GROMACS log files report how the code was patched
+        # during the build, so that any problems are easier to diagnose.
+        if '-DGMX_VERSION_STRING_OF_FORK=' not in self.cfg['configopts']:
+            if plumed_root:
+                gromacs_version_string_suffix = 'PLUMED-%s-%s' % (get_software_version('PLUMED'), VERBOSE_VERSION)
+            else:
+                gromacs_version_string_suffix = VERBOSE_VERSION
```
boegel, line 218: Just noticed this while testing: `VERBOSE_VERSION` doesn't include `EasyBuild` at all (it's verbose because it may include a Git commit ID, if EasyBuild is being run straight from a Git repository).

So you probably want something like this instead:

```python
gromacs_version_string_suffix = 'EasyBuild-%s' % VERBOSE_VERSION
if plumed_root:
    gromacs_version_string_suffix += '-PLUMED-%s' + get_software_version('PLUMED')

boegel (dismissed):

akesandgren:
Test report by @akesandgren
 * **SUCCESS** _GROMACS-2019.3-foss-2019a.eb_ 
 * **SUCCESS** _GROMACS-2021-foss-2020b.eb_ 
 * **SUCCESS** _PLUMED-2.6.0-foss-2020a-Python-3.8.2.eb_ 
 * **SUCCESS** _GROMACS-2019.4-foss-2020a-PLUMED-2.6.0.eb_ 

Build succeeded for 4 out of 4 (3 easyconfigs in total)
b-an02.hpc2n.umu.se - Linux Ubuntu 20.04, x86_64, Intel(R) Xeon(R) CPU E5-2690 v4 @ 2.60GHz, Python 3.8.5
See https://gist.github.com/466e687c088ccd1cb9a81c421084e3fc for a full test report.

easybuild/easyblocks/g/gromacs.py
mabraham: Done

easybuild/easyblocks/g/gromacs.py
mabraham: Done

easybuild/easyblocks/g/gromacs.py
mabraham: Done

mabraham (commented):
Thanks for the feedback, it is better now!

akesandgren:
Test report by @akesandgren
 * **SUCCESS** _GROMACS-2019.3-foss-2019a.eb_ 
 * **SUCCESS** _GROMACS-2019.4-foss-2020a-PLUMED-2.6.0.eb_ 
 * **SUCCESS** _GROMACS-2021-foss-2020b.eb_ 

b-an02.hpc2n.umu.se - Linux Ubuntu 20.04, x86_64, Intel(R) Xeon(R) CPU E5-2690 v4 @ 2.60GHz, Python 3.8.5
See https://gist.github.com/424494504761049858c33745e0464070 for a full test report.

akesandgren (approved):
LGTM

boegel:
Thanks a lot for the contribution @mabraham!

PR #2582, opened 2021-09-24
restore RPATH wrappers for OpenMPI sanity check (WIP)

boegel:
WIP because I need to check whether this fixes a problem reported in Slack, after I reproduce the issue first...

### DIFF
--- a/easybuild/easyblocks/o/openmpi.py
+++ b/easybuild/easyblocks/o/openmpi.py
@@ -198,4 +198,10 @@ class EB_OpenMPI(ConfigureMake):
                     params['nr_ranks'] = 1
                     custom_commands.append(mpi_cmd_tmpl % params)
 
+        # ensure RPATH wrappers are in place, otherwise compiling minimal test programs will fail
+        if build_option('rpath'):
+            if self.toolchain.options.get('rpath', True):
+                self.toolchain.prepare_rpath_wrappers(rpath_filter_dirs=self.rpath_filter_dirs,
+                                                      rpath_include_dirs=self.rpath_include_dirs)
+
         super(EB_OpenMPI, self).sanity_check_step(custom_paths=custom_paths, custom_commands=custom_commands)


stderr-enst:
@boegel thanks for the PR!
Going over OpenMPI in foss-2019b with `--module-only --include-easyblocks-from-pr 2582` seemed to work in my debug build (designated local buildpath, --debug flag and set tmp dirs). To double check I also ran the build process in foss-2021a without `--module-only`, building it in a "production setting" and this is giving me the same error as before:
```
FAILED: Installation ended unsuccessfully (build directory: /tmp/easybuild/rome/OpenMPI/4.1.1/GCC-10.3.0): build failed (first 300 chars): Sanity check failed: sanity check command mpirun -n 8 /tmp/easybuild/rome/OpenMPI/4.1.1/GCC-10.3.0/mpi_test_hello_mpifh exited with code 127 (output: /tmp/easybuild/rome/OpenMPI/4.1.1/GCC-10.3.0/mpi_test_hello_mpifh: error while loading shared libraries: libgfortran.so.5: cannot open shared object f...
```
I'm a bit confused by this and think the only relevant difference should be running with/without `--module-only`.
Could this be affected by possibly left-over files from a previous attempt to build OpenMPI in 2021a?

In any case, I'm trying to apply this again in a clean build with 2021a and get back to you if I'm not successful. Just wanted to give an update on this

stderr-enst:
Building OpenMPI again in 2021a still fails. It also fails in 2019b with `--rebuild`. Running again with `--module-only` with the eb from this PR, seems to make the build happy, but my feeling is that the part resulting in the error is skipped. (although the sanity check step is run?) Not sure why `module-only` makes a difference.

boegel:
Test report by @boegel
 * **SUCCESS** _OpenMPI-4.0.3-GCC-9.3.0.eb_ 

login1 - Linux UNKNOWN UNKNOWN, x86_64, Intel(R) Xeon(R) CPU E5-2690 v3 @ 2.60GHz (haswell), Python 3.9.5
See https://gist.github.com/1eccc898554fcc6badfb7be6fbf95de9 for a full test report.

PR #2615, opened 2021-11-05
skip motorBike tutorial case in OpenFOAM sanity check if available cores < 6

LHurst-JM:
The tutorial is hardcoded to use exactly 6 cores, so the sanity check will always fail if there are fewer cores than this available locally (even if there is no problem with the build).  Problem occurs with both foss and intel toolchains (I've not tested any others).

This change only attempts to run that case as a test if there are 6 or more cores locally.

Closes easybuilders/easybuild-easyconfigs#14291

### DIFF
--- a/easybuild/easyblocks/o/openfoam.py
+++ b/easybuild/easyblocks/o/openfoam.py
@@ -487,37 +487,44 @@ class EB_OpenFOAM(EasyBlock):
         # run motorBike tutorial case to ensure the installation is functional (if it's available);
         # only for recent (>= v6.0) versions of openfoam.org variant
         if self.looseversion >= LooseVersion('6') and self.looseversion < LooseVersion('100'):
-            openfoamdir_path = os.path.join(self.installdir, self.openfoamdir)
-            motorbike_path = os.path.join(openfoamdir_path, 'tutorials', 'incompressible', 'simpleFoam', 'motorBike')
-            if os.path.exists(motorbike_path):
-                test_dir = tempfile.mkdtemp()
+            # motorBike tutorial case is hardcoded to use 6 cores.  mpirun (called by runParallel) will fail with an
+            # 'insufficient slots available' error if there are fewer than 6 cores available, causing the sanity
+            # check to fail.  Only attempt to run it if there are 6 (or more) cores available.
+            if os.cpu_count() >= 6:
+                openfoamdir_path = os.path.join(self.installdir, self.openfoamdir)
+                motorbike_path = os.path.join(openfoamdir_path, 'tutorials', 'incompressible', 'simpleFoam', 'motorBike')
+                if os.path.exists(motorbike_path):
+                    test_dir = tempfile.mkdtemp()
 
-                if self.looseversion >= LooseVersion('9'):
-                    geom_target_dir = 'geometry'
+                    if self.looseversion >= LooseVersion('9'):
+                        geom_target_dir = 'geometry'
+                    else:
+                        geom_target_dir = 'triSurface'
+
+                    cmds = [
+                        "cp -a %s %s" % (motorbike_path, test_dir),
+                        "cd %s" % os.path.join(test_dir, os.path.basename(motorbike_path)),
+                        "source $FOAM_BASH",
+                        ". $WM_PROJECT_DIR/bin/tools/RunFunctions",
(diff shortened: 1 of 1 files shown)


boegel:
An alternative here is to allow oversubscription of MPI ranks on cores, right @branfosj?

LHurst-JM:
@boegel yes, and that would be preferable however the actual mpirun command is burried inside a script within the OpenFOAM install and I didn't find an MPI-agnostic way to do it via enivironment variables (since this is in the easyblock, it needs to work for all the MPIs the easyconfig might be built with).  Basically I couldn't make it work after a few hours struggling with it.

Mormacill:
The _runParallel_ function is defined in the RunFunctions script which is sourced at the beginning by

`. $WM_PROJECT_DIR/bin/tools/RunFunctions`

The function has some predefined options for mpirun but not for the oversubscription flag.

The key section says:
```
[...]
        echo "Running $APP_RUN in parallel on $PWD using $nProcs processes"
        if [ "$LOG_APPEND" = "true" ]; then
            ( mpirun -np $nProcs $APP_RUN -parallel "$@" < /dev/null >> log.$LOG_SUFFIX 2>&1 )
        else
            ( mpirun -np $nProcs $APP_RUN -parallel "$@" < /dev/null > log.$LOG_SUFFIX 2>&1 )
        fi
[...]
```


I think it is not recommended to patch these lines as people might use these functions later in production when oversubscription is not desired.


It might be possible to replace the line 
`runParallel simpleFoam`
in the easyblock to the linked mpirun command, e.g.:
`mpirun -np $(getNumberOfProcessors) $(getApplication) -parallel --oversubscribe < /dev/null >> log.$LOG_SUFFIX 2>&1`

This is not the nice way from a style view but it keeps the sanity check and avoids a skip.


It could also be something possible by changing the entry for `numberOfSubdomains` in the decomposeParDict by the foamDictionary function to change the tutorial only using the available resources. As an advantage the _runParallel_ function can persist because it obtains the number of cores used from the `numberOfSubdomains` keyword in decomposeParDict.
This could look like:
`foamDictionary -entry "numberOfSubdomains" -set $(nproc) system/decomposeParDict`
Problem here just is that motorBike uses _hierarchical_ decomposition, the coefficients need to be adjusted too.
Or to change to _scotch_ with the same command to avoid coefficients at all.

Although I am not sure about this method changing the dicts because it's an intervention to the original tutorial files which might be inappropriate.

LHurst-JM:
@Mormacill yes, there doesn't seem to be a clear easy solution to using this tutorial as a sanitycheck.  I'm not familiar with using OpenFOAM, just building it - I don't suppose you are aware of a bundle tutorial case that would do as an adequate test case but needs 2 or 4 cores instead of 6?  (Doesn't completely solve the problem but 4-cores is far more common at the "low end" of HPC systems, in my experience.)

Mormacill:
@LHurst-JM There are a few cases which are using 4 cores, just one I found with 2 by default (heatTransfer/buoyantSimpleFoam/iglooWithFridges). The advantage of motorBike tutorial sanity check is that it covers a wide range of functions being tested.
It might be a solution to implement a fallback tutorial if host cores are lower than 6 instead of only skipping motorbike.

LHurst-JM:
@Mormacill thanks, I agree that sounds like a better plan.  Not knowing OpenFOAM I find it hard to judge which cases are suitable.

boegel:
@LHurst-JM Does the sanity check also fail when OpenFOAM is built on top of Intel MPI?

I know OpenMPI by default gives up when there are less cores available than "slots" (cores), but I don't remember running into that with Intel MPI?

LHurst-JM:
Hi @boegel, I'm sure I tried intel and had the same problem but testing again now....

LHurst-JM:
@boegel my bad, you are entirely correct - it does seem to build with the Intel toolchain, I must have assumed the issue would be common to all mpirun implementations.  That does explain why I was struggling to find how to make Intel's mpirun oversubscribe - it must do it by default!

boegel:
This change should no longer be needed if https://github.com/easybuilders/easybuild-framework/pull/3909 gets merged...

LHurst-JM:
Looks good to me, I'll close this one.  I think the general solution is much better than playing whack-a-mole with individual easyblocks.

PR #2617, opened 2021-11-12
add custom easyblock for Clang-AOMP

nordmoen:
Required for easybuilders/easybuild-easyconfigs#14327

### DIFF
--- a/easybuild/easyblocks/c/clang_aomp.py
+++ b/easybuild/easyblocks/c/clang_aomp.py
@@ -1,0 +1,238 @@
+##
+# Copyright 2009-2021 Ghent University
+#
+# This file is part of EasyBuild,
+# originally created by the HPC team of Ghent University (http://ugent.be/hpc/en),
+# with support of Ghent University (http://ugent.be/hpc),
+# the Flemish Supercomputer Centre (VSC) (https://www.vscentrum.be),
+# Flemish Research Foundation (FWO) (http://www.fwo.be/en)
+# and the Department of Economy, Science and Innovation (EWI) (http://www.ewi-vlaanderen.be/en).
+#
+# https://github.com/easybuilders/easybuild
+#
+# EasyBuild is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation v2.
+#
+# EasyBuild is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with EasyBuild.  If not, see <http://www.gnu.org/licenses/>.
+##
+"""
+EasyBuild support for building and installing AOMP version of LLVM/Clang
+
+@author: Jørgen Nordmoen (University of Oslo, USIT)
+"""
+import glob
+import os
+
+from easybuild.easyblocks.clang import DEFAULT_TARGETS_MAP as LLVM_ARCH_MAP
+from easybuild.easyblocks.generic.bundle import Bundle
+from easybuild.framework.easyconfig import CUSTOM
+from easybuild.tools.build_log import EasyBuildError, print_msg, print_warning
+from easybuild.tools.config import build_option
+from easybuild.tools.modules import get_software_root
+from easybuild.tools.systemtools import AARCH64, POWER, X86_64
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/c/clang_aomp.py
```diff
@@ -0,0 +1,239 @@
...
+from easybuild.framework.easyconfig import CUSTOM
+from easybuild.tools.build_log import EasyBuildError, print_msg, print_warning
+from easybuild.tools.config import build_option
+from easybuild.tools.modules import get_software_root
+from easybuild.tools.systemtools import AARCH64, POWER, X86_64
+from easybuild.tools.systemtools import get_cpu_architecture, get_shared_lib_ext
+
+
+# Default AMD GPU architectures to build for
+DEFAULT_GFX_ARCHS = ['gfx900', 'gfx902', 'gfx906', 'gfx908', 'gfx90a', 'gfx1030', 'gfx1031']
```
boegel, line 45: @nordmoen Any reference that can be added here that provides more info on what each of these values mean?

easybuild/easyblocks/c/clang_aomp.py
```diff
@@ -0,0 +1,239 @@
...
+            self.amd_gfx_archs = DEFAULT_GFX_ARCHS
+        self.log.info("Building offload support for the following AMD architectures: '%s'",
+                      ' '.join(self.amd_gfx_archs))
+        # Ensure necessary libraries are downloaded and can be found
+        device_lib_dir_pattern = os.path.join(self.builddir, 'ROCm-Device-Libs-*')
+        hits = glob.glob(device_lib_dir_pattern)
+        if len(hits) == 1:
+            self.device_lib_path = hits[0]
+        else:
+            raise EasyBuildError("Could not find 'ROCm-Device-Libs' source directory")
```
boegel, line 119: @nordmoen It's better to move this to `configure_step`, otherwise `--module-only` will always fail hard (since sources are not unpacked then, but `prepare_step` is run).

boegel (changes requested):

easybuild/easyblocks/c/clang_aomp.py
nordmoen: Done =]

easybuild/easyblocks/c/clang_aomp.py
nordmoen: Hopefully this new explanation is a bit more informative?

boegel: Perfect, thanks!

boegel (changes requested):
@nordmoen I've reviewed this in detail, and tweaked it a bit for a variety of reasons, please take a look at https://github.com/nordmoen/easybuild-easyblocks/pull/4

easybuild/easyblocks/c/clang_aomp.py
```diff
@@ -0,0 +1,276 @@
...
+
+        custom_paths['files'].append(os.path.join('lib', 'libomptarget.rtl.%s.%s' % (arch, shlib_ext)))
+
+        # If CUDA offloading support was requested, check that correct omptarget was built
+        if get_software_root('CUDA'):
+            custom_paths['files'].append(os.path.join('lib', 'libomptarget.rtl.cuda.%s' % shlib_ext))
+
+            for arch in self.cuda_archs:
+                sm_arch = 'sm_%s' % arch
+                custom_paths['files'].extend([os.path.join(libdevice, 'lib%s-nvptx-%s' % (x, sm_arch)) for x in libs])
```
boegel, line 197: This looks incorrect, there's no `lib*nvptx* in `lib/libdevice` when building with CUDA support, but there's a `lib/libomptarget-nvptx*.bc` for every target CUDA compute capability.

fixed in https://github.com/nordmoen/easybuild-easyblocks/pull/5

boegel (changes requested):

boegel:
Test report by @boegel
 * **SUCCESS** _Clang-AOMP-4.5.0-GCCcore-11.2.0-CUDA-11.3.1.eb_ 

pg-lab02 - Linux centos linux 8.3.2011, x86_64, AMD EPYC 7542 32-Core Processor (zen2), Python 3.6.8
See https://gist.github.com/f8741c38d5cc293e3cd78b0deed80f37 for a full test report.

boegel (approved):

boegel:
Tested with both https://github.com/easybuilders/easybuild-easyconfigs/pull/14327 and a custom `Clang-AOMP` easyconfig with `CUDA` as additional dependency, works as expected, so time to merge.

Thanks a lot for your efforts on this @nordmoen, it's clear that this wasn't exactly trivial...

PR #2604, opened 2021-10-18
set the correct test directory for CMake Amber and AmberTools

branfosj:
(created using `eb --new-pr`)

Set as draft. I'm unclear if we are currently running the AmberTools tests when building Amber.

PR #2618, opened 2021-11-12
Ambertools support

mboisson:
This PR allows a few tings
1) For my build, CMake failed to find the fortran headers for NetCDF unless I make their location explicit through NetCDF_INCLUDES
2) CMake will not automatically find FlexiBLAS. Defining the libraries 
3) Allows to overwrite the INSTALL_TESTS and COMPILER options. In particular, when compiling with GCC on our stack, if I don't set COMPILER to GNU, disable `usempi` (to get serial compilers for serial code) and enable MPI through the CMake option, it fails to build. 
4) Disable tests when building AmberTools, as not all the binaries are included.

### DIFF
--- a/easybuild/easyblocks/a/amber.py
+++ b/easybuild/easyblocks/a/amber.py
@@ -164,12 +164,18 @@ class EB_Amber(CMakeMake):
             external_libs_list.append('netcdf')
         if get_software_root('netCDF-Fortran'):
             external_libs_list.append('netcdf-fortran')
+            netcdf_fortran_root = get_software_root('netCDF-Fortran')
+            self.cfg.update('configopts', '-DNetCDF_INCLUDES=%s/include' % netcdf_fortran_root)
         if get_software_root('zlib'):
             external_libs_list.append('zlib')
         if get_software_root('Boost'):
             external_libs_list.append('boost')
         if get_software_root('PnetCDF'):
             external_libs_list.append('pnetcdf')
+
+        # Force to use BLAS and LAPACK from environment variables
+        self.cfg.update('configopts', '-DBLAS_LIBRARIES="%s"' % os.getenv('LIBBLAS'))
+        self.cfg.update('configopts', '-DLAPACK_LIBRARIES="%s"' % os.getenv('LIBLAPACK'))
 
         # Force libs for available deps (see cmake/3rdPartyTools.cmake in Amber source for list of 3rd party libs)
         # This provides an extra layer of checking but should already be handled by TRUST_SYSTEM_LIBS=TRUE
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/a/amber.py
```diff
@@ -164,13 +164,19 @@ def configure_step(self):
             external_libs_list.append('netcdf')
         if get_software_root('netCDF-Fortran'):
             external_libs_list.append('netcdf-fortran')
+            netcdf_fortran_root = get_software_root('netCDF-Fortran')
+            self.cfg.update('configopts', '-DNetCDF_INCLUDES=%s/include' % netcdf_fortran_root)
```
smoors, line 168:
```suggestion
        netcdf_fortran_root = get_software_root('netCDF-Fortran')
        if netcdf_fortran_root:
            external_libs_list.append('netcdf-fortran')
            self.cfg.update('configopts', '-DNetCDF_INCLUDES=%s/include' % netcdf_fortran_root)
```

easybuild/easyblocks/a/amber.py
```diff
@@ -315,8 +323,8 @@ def install_step(self):
 
         super(EB_Amber, self).install_step()
 
-        # Run the tests located in the build directory
-        if self.cfg['runtest']:
+        # Run the tests located in the build directory, only for Amber. Tests fail when building only AmberTools
+        if self.cfg['runtest'] and self.name == 'Amber':
```
smoors, line 327: do all tests fail with AmberTools, or only some?
if not all tests fail, I think it's still useful to run them in combination with EB option `--ignore-test-failure`.

mboisson: I did not try to untangle the tests, but not all executables from Amber get created, and the tests assume that they are there.

branfosj: See #2604 (and https://github.com/easybuilders/easybuild-easyconfigs/pull/14028) - it is trying to run the Amber test suite and not the AmberTools test suite. There were still some failures with the AmberTool test suite and I've not had the chance to investigate further.

akesandgren: Yeah, we're doing the runtest part wrong. I think I have the correct stuff in my test-.eb but I got sidetracked and haven't had time to go back and revisit.
It should not matter if we build Amber or AmberTools the test part (CMake and old style) is written so that it only tests what is actually built, if we call it correctly. I checked with the devs on that.

smoors: @akesandgren  did you hear back from the Amber devs yet?

akesandgren: They basically said what I said above, The CMake part should do the right thing when asked to run make test.
But I haven't had time to revisit this on my side...

mboisson: This still fails as of today. Can't build AmberTools without disabling tests.

mboisson:
PR https://github.com/easybuilders/easybuild-easyblocks/pull/2720 addresses some of the same points as this PR, but not all of them. In particular, it does not disable tests for AmberTools, and so the build of Ambertools fails

akesandgren:
PR https://github.com/easybuilders/easybuild-easyblocks/pull/2781 makes sure the tests are run the correct way and should eliminate that problem.

boegel:
@mboisson Can you sync this PR with current `develop`?

mboisson:
Done

akesandgren:
The NetCDF-Fortran problem is solved with AmberTools-20_cmake-locate-netcdf.patch but doing it directly in the easyblock might be beneficial.
FlexiBLAS is solved with AmberTools-21_CMake-FlexiBLAS.patch which I think is the more correct way.

Thoughts on the above two @mboisson ?

(And the testing is finally solved with the latest easyblock+easyconfigs+patches, apart from a few remaining test failures)

PR #2638, opened 2021-12-15
Install and load rpath wrappers in GCCcore module (WIP)

stderr-enst:
WIP PR to install RPATH wrappers together with GCCcore and loading them automatically when loading the module.
This would make GCC in an RPATH stack transparent to a user, if they were to build their own software against the stack.

There are still a couple of things to work out and this is just a first proof of concept.

Left to do:
1. Add configurable option to install rpath wrappers from easyconfig
1. Don't setup wrappers when using GCCcore internally in EasyBuild
1. Refactor `create_rpath_wrappers` into a more generic location, since this is probably used for other compilers in the future
1. Extract wrappers to `ld`, since they are not really 
1. Add environment variable in module file to disable adding wrappers to `PATH`
1. Document this feature in EasyBuild RPATH (and compiler) documentation and point out when to use or avoid this feature. Maybe also in module description?
1. Add similar feature for other compilers?


easybuild/easyblocks/g/gcc.py
```diff
@@ -1043,4 +1048,42 @@ def make_module_req_guess(self):
...
+                # Expect rpath_filter_dirs and rpath_include_dirs as environment variables of
+                # ,-separated lists by default. We might want to allow different separators.
+                # E.g. ":" could be used to get a similar syntax to PATH and LD_LIBRARY_PATH
+                separator = ','
+                rpath_filter_dirs = None
+                if "RPATH_FILTER_DIRS" in os.environ:
+                    rpath_filter_dirs = os.environ["RPATH_FILTER_DIRS"].split(separator)
+                rpath_include_dirs = None
+                if "RPATH_INCLUDE_DIRS" in os.environ:
+                    rpath_include_dirs = os.environ["RPATH_INCLUDE_DIRS"].split(separator)
```
boegel, line 1067: Don't we want these environment variables to be picked up when the wrapper script is being used?

As it is now, you're picking up `$RPATH_FILTER_DIRS` when `GCCcore` is being installed, and baking that into the wrapper.

Allowing for this will require some work in the script itself (in `easybuild-framework`) though...

So for now, I would just call `prepare_rpath_wrappers` without passing anything?

easybuild/easyblocks/g/gcc.py
```diff
@@ -1043,4 +1048,42 @@ def make_module_req_guess(self):
             'LD_LIBRARY_PATH': ['lib', 'lib64'],
             'MANPATH': ['man', 'share/man']
         })
+
+        # Install rpath wrappers, if built with RPATH enabled
+        if build_option('rpath'):
+            def create_rpath_wrappers(targetdir):
+                set_up_configuration(silent=True)
+                tc = Toolchain(name="system", version="0.1")
```
boegel, line 1056: Although it probably doesn't matter much, we should use `GCCcore` as name, and `self.version` as version here to create the toolchain instance:

```python
tc = Toolchain(name=self.name, version=self.name)

File: easybuild/easyblocks/g/gcc.py [hunk 3]
```diff
@@ -1043,4 +1048,42 @@ def make_module_req_guess(self):
...
+
+                # Expect rpath_filter_dirs and rpath_include_dirs as environment variables of
+                # ,-separated lists by default. We might want to allow different separators.
+                # E.g. ":" could be used to get a similar syntax to PATH and LD_LIBRARY_PATH
```
boegel, line 1060: `:` is typically used for `_*PATH` environment variables (which specify a list of paths to search).

I think it makes sense to (only) use `,` here, since `*_DIRS` implies a list of items, and then comma-separated is reasonable.

easybuild/easyblocks/g/gcc.py
```diff
@@ -1043,4 +1048,42 @@ def make_module_req_guess(self):
...
+                    rpath_include_dirs = os.environ["RPATH_INCLUDE_DIRS"].split(separator)
+
+                tc.prepare_rpath_wrappers(
+                        rpath_filter_dirs=rpath_filter_dirs,
+                        rpath_include_dirs=rpath_include_dirs
+                        )
+
+                # Find location of rpath wrappers. Tempfile in prepare_rpath_wrappers seems to be
+                # in a parent directory of our current gettempdir()-location
+                wrapperpath = glob.glob(os.path.join(tempfile.gettempdir(), "..", "tmp*", RPATH_WRAPPERS_SUBDIR))[0]
```
boegel, line 1076: Ideally this just returned by `prepare_rpath_wrappers`, so we don't have to hunting for stuff...

Or allow to specify a location where the wrappers should be created to `prepare_rpath_wrappers`, that's even better.

easybuild/easyblocks/g/gcc.py
```diff
@@ -1043,4 +1048,42 @@ def make_module_req_guess(self):
...
+                move_file(wrapperpath, targetdir)
+
+            rpath_wrapperdir = os.path.join('bin', RPATH_WRAPPERS_SUBDIR)
+            absolute_wrapperdir = os.path.join(self.installdir, rpath_wrapperdir)
+            if not os.path.exists(absolute_wrapperdir):
+                # make_module_req_guess seems to be called multiple times
```
boegel, line 1082: This is a sign that `make_module_req_guess` is not the right place for this, it should probably go into `post_install_step` instead.

Using `make_module_req_guess` makes sense for testing (since you can add the wrappers to an existing installation via `--module-only`), but it doesn't really belong there.

easybuild/easyblocks/g/gcc.py
```diff
@@ -1043,4 +1048,42 @@ def make_module_req_guess(self):
...
+                # create and move rpath wrappers only once!
+                create_rpath_wrappers(absolute_wrapperdir)
+            for wrapper in os.listdir(absolute_wrapperdir):
```
boegel, line 1085: I think we should aim to have all wrapper scripts directly in `bin/rpath_wrappers`, rather than in a subdirectory thereof?

That's what EasyBuild does dynamically, out of an abundance of caution, but it doesn't make much sense when the wrappers are included with `GCCcore` itself I think.

boegel (changes requested):

boegel:
> WIP PR to install RPATH wrappers together with GCCcore and loading them automatically when loading the module. This would make GCC in an RPATH stack transparent to a user, if they were to build their own software against the stack.
> 
> There are still a couple of things to work out and this is just a first proof of concept.
> 
> Left to do:
> 
>     1. Add configurable option to install rpath wrappers from easyconfig

A custom easyconfig parameter (`add_rpath_wrappers = True`, with `False` as default?) can be specified via `extra_options`, see for example https://github.com/easybuilders/easybuild-easyblocks/pull/2639/files

>     2. Don't setup wrappers when using GCCcore internally in EasyBuild

If adding the location of the wrappers to `$PATH` is guarded by an environment variable, EasyBuild framework should just set that as a part of setting up the build environment (so in `EasyBlock.prepare_step` in framework).

Maybe we should also introduce a `$THIS_IS_EASYBUILD` environment variable, so we can easily detect when we're running in an environment set up by EasyBuild.

>     3. Refactor `create_rpath_wrappers` into a more generic location, since this is probably used for other compilers in the future

Makes sense, we could add a function in `easybuild.tools.toolchain.toolchain` for that in framework, and also make the tweaks to `prepare_rpath_wrappers` in the same PR.

>     4. Extract wrappers to `ld`, since they are not really

Not sure what you mean here, can you clarify?

>     5. Add environment variable in module file to disable adding wrappers to `PATH`

I would guard adding the wrapper to `$PATH` to only do it when `$THIS_IS_EASYBUILD` and `$DISABLE_RPATH_WRAPPERS` are *not* defined.

The `conditional_statement` method in `self.module_generator` will be useful here, but we probably also need a method to specify a list of environment variables that should *not* be defined (so you don't need to worry about Lua vs Tcl syntax, that should be handed by `module_generator`.

>     6. Document this feature in EasyBuild RPATH (and compiler) documentation and point out when to use or avoid this feature. Maybe also in module description?

Both make sense, yes.

>     7. Add similar feature for other compilers?

That's outside of the scope of this PR, I would say. Maybe in practice it's even sufficient to just install the RPATH wrappers with `GCCcore`, since it's mainly about the`ld` wrapper I think?

easybuild/easyblocks/g/gcc.py
```diff
@@ -1043,4 +1048,42 @@ def make_module_req_guess(self):
             'LD_LIBRARY_PATH': ['lib', 'lib64'],
             'MANPATH': ['man', 'share/man']
         })
+
+        # Install rpath wrappers, if built with RPATH enabled
+        if build_option('rpath'):
+            def create_rpath_wrappers(targetdir):
+                set_up_configuration(silent=True)
```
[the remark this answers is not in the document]
stderr-enst: done

easybuild/easyblocks/g/gcc.py
```diff
@@ -55,6 +57,9 @@
 from easybuild.tools.systemtools import get_cpu_architecture, get_gcc_version, get_shared_lib_ext
 from easybuild.tools.toolchain.compiler import OPTARCH_GENERIC
 from easybuild.tools.utilities import nub
+from easybuild.tools.options import set_up_configuration
+from easybuild.tools.toolchain.toolchain import Toolchain
+from easybuild.tools.toolchain.toolchain import RPATH_WRAPPERS_SUBDIR
```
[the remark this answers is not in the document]
stderr-enst: done

easybuild/easyblocks/g/gcc.py
stderr-enst: I've removed this for now, since this should go into the wrapper script PR in easybuild-framework

easybuild/easyblocks/g/gcc.py
stderr-enst: Following our discussion, I'm going the `get_toolchain` route now to correctly setup the `self.CC_COMPILER` variables etc.

easybuild/easyblocks/g/gcc.py
stderr-enst: I've removed this for now, since this should go into the wrapper script PR in easybuild-framework. But there we can just go with a `,`-separated list

easybuild/easyblocks/g/gcc.py
stderr-enst: Yes, this is an artifact of testing. It moved to the `post_install_step` now

easybuild/easyblocks/g/gcc.py
stderr-enst: This is another thing that should be addressed by the future easybuild-framework PR.
`prepare_rpath_wrappers` currently creates a single subdir per wrapper script, but it would be better to ship them in a single directory when installing the module.

stderr-enst:
> > WIP PR to install RPATH wrappers together with GCCcore and loading them automatically when loading the module. This would make GCC in an RPATH stack transparent to a user, if they were to build their own software against the stack.
> > There are still a couple of things to work out and this is just a first proof of concept.
> > Left to do:
> > ```
> > 1. Add configurable option to install rpath wrappers from easyconfig
> > ```
> 
> A custom easyconfig parameter (`add_rpath_wrappers = True`, with `False` as default?) can be specified via `extra_options`, see for example https://github.com/easybuilders/easybuild-easyblocks/pull/2639/files

Thank you for the link! I've added the extra option to this easyblock now.

> > ```
> > 2. Don't setup wrappers when using GCCcore internally in EasyBuild
> > ```
> 
> If adding the location of the wrappers to `$PATH` is guarded by an environment variable, EasyBuild framework should just set that as a part of setting up the build environment (so in `EasyBlock.prepare_step` in framework).
> 
> Maybe we should also introduce a `$THIS_IS_EASYBUILD` environment variable, so we can easily detect when we're running in an environment set up by EasyBuild.
> 
> > ```
> > 3. Refactor `create_rpath_wrappers` into a more generic location, since this is probably used for other compilers in the future
> > ```
> 
> Makes sense, we could add a function in `easybuild.tools.toolchain.toolchain` for that in framework, and also make the tweaks to `prepare_rpath_wrappers` in the same PR.
>
So we need to open an accompanying PR in easybuild-framework. Is there a process to declare PR dependencies between projects, or do we just open both, mention both in each others description and make sure that they merge around the same time?

> > ```
> > 4. Extract wrappers to `ld`, since they are not really
> > ```
> 
> Not sure what you mean here, can you clarify?
> 
It looks like I didn't finish my sentence :D
What I meant to point out was about Alan's comment (I think?) in our EESSI meeting. He said that GCCcore might not be the right place to set up a wrapper to `ld`, since `ld` is not shipped with it.
But I'm not really sure how to handle this

> > ```
> > 5. Add environment variable in module file to disable adding wrappers to `PATH`
> > ```
> 
> I would guard adding the wrapper to `$PATH` to only do it when `$THIS_IS_EASYBUILD` and `$DISABLE_RPATH_WRAPPERS` are _not_ defined.
> 
> The `conditional_statement` method in `self.module_generator` will be useful here, but we probably also need a method to specify a list of environment variables that should _not_ be defined (so you don't need to worry about Lua vs Tcl syntax, that should be handed by `module_generator`.
> 
Thanks for mentioning the method! I missed it so far, but I'll look at it later.


> > ```
> > 7. Add similar feature for other compilers?
> > ```
> 
> That's outside of the scope of this PR, I would say. Maybe in practice it's even sufficient to just install the RPATH wrappers with `GCCcore`, since it's mainly about the`ld` wrapper I think?

Okay! :)

boegel:
> > > ```
> > > 3. Refactor `create_rpath_wrappers` into a more generic location, since this is probably used for other compilers in the future
> > > ```
> > 
> > 
> > Makes sense, we could add a function in `easybuild.tools.toolchain.toolchain` for that in framework, and also make the tweaks to `prepare_rpath_wrappers` in the same PR.
> 
> So we need to open an accompanying PR in easybuild-framework. Is there a process to declare PR dependencies between projects, or do we just open both, mention both in each others description and make sure that they merge around the same time?

Just opening the PR and declaring that in this easyblocks PR that it requires framework PRs by mentioning them in the description is enough.
The failing tests in this PR when you start importing stuff from framework that isn't there yet in `develop` will make that clear too. :)

> 
> > > ```
> > > 4. Extract wrappers to `ld`, since they are not really
> > > ```
> > 
> > 
> > Not sure what you mean here, can you clarify?
> 
> It looks like I didn't finish my sentence :D What I meant to point out was about Alan's comment (I think?) in our EESSI meeting. He said that GCCcore might not be the right place to set up a wrapper to `ld`, since `ld` is not shipped with it. But I'm not really sure how to handle this

OK, thanks for clarifying.

We indeed rely on the `binutils` provided by the compat layer in EESSI, which makes this a bit more interesting...

I don't see another way to include the `ld` wrapper in `GCCcore` in that case, since where else would we put it?
I guess parking the RPATH wrapper in the compat layer could be considered, but I'm not sure that a good way forward.

If `binutils` is installed through EasyBuild, the right place would be in the `binutils` installation, of course.

easybuild/easyblocks/g/gcc.py
```diff
@@ -831,6 +837,28 @@ def post_install_step(self, *args, **kwargs):
...
+                        )
+
+                # Find location of rpath wrappers. Tempfile in prepare_rpath_wrappers seems to be
+                # in a parent directory of our current gettempdir()-location
+                wrapperpath = glob.glob(os.path.join(tempfile.gettempdir(), "tmp*", RPATH_WRAPPERS_SUBDIR))[0]
+                os.mkdir(targetdir)
+                for wrapper in os.listdir(wrapperpath):
+                    filename = os.listdir(os.path.join(wrapperpath, wrapper))[0]
+                    filepath = os.path.join(wrapperpath, wrapper, filename)
+                    move_file(filepath, os.path.join(targetdir, filename))
```
boegel, line 857: @stderr-enst We also need to make sure that these wrappers scripts have executable permissions for all users, not just the owner (the account used to do the installation)

boegel (changes requested):

easybuild/easyblocks/g/gcc.py
stderr-enst: This is now happening in the other PR:
https://github.com/easybuilders/easybuild-framework/blob/fc39b6609fab98de2dbbc73cb3d28f72de3fa5fb/easybuild/tools/toolchain/toolchain.py#L1045

PR #2671, opened 2022-02-02
To automatically support some of the CUDA features of ELPA

damianam:


easybuild/easyblocks/e/elpa.py
```diff
@@ -159,6 +161,19 @@ def run_all_steps(self, *args, **kwargs):
...
+        # Add CUDA features
+        if 'CUDA' in [i['name'] for i in self.cfg.dependencies()]:
+            self.cfg.update('configopts', '--enable-nvidia-gpu')
+            cuda_cc_space_sep = self.cfg.template_values['cuda_cc_space_sep'].replace('.', '').split()
+            # Just one is supported, so pick the highest one (but prioritize sm_80)
+            selected_cc = "0"
+            for cc in cuda_cc_space_sep:
+                if int(cc) > int(selected_cc) and int(selected_cc) != 80:
+                    selected_cc = cc
+            self.cfg.update('configopts', f'--with-NVIDIA-GPU-compute-capability=sm_{selected_cc}')
```
SebastianAchilles, line 173: I think support for f strings in Python was added only in newer versions. That is why the CI fails for Python 2.7 and 3.5
```suggestion
            self.cfg.update('configopts', f'--with-NVIDIA-GPU-compute-capability=sm_%s' % selected_cc)
```

damianam: Python 2 still being a thing in the bleeding edge world of EB? 😅  I'll fix it then.

damianam: I just realized that in the mess of migrating to my new laptop, I wrongly pushed this branch in the central repo instead of in my fork. Sorry for that, I'll close the PR, delete the branch and make a new PR from my fork.

PR #2666, opened 2022-01-28
add support to NAMD easyblock to opt out of building with CUDA support even if CUDA is included as dependency

damianam:
Useful when CUDA is part of the dependencies (like when included by MPI
or any other dependency), but we don't want to build Charm++ with CUDA
support

PR #2673, opened 2022-02-03
Add automatic CUDA support for ELPA

damianam:


easybuild/easyblocks/e/elpa.py
```diff
@@ -159,6 +161,24 @@ def run_all_steps(self, *args, **kwargs):
             self.cfg.update('configopts', '--with-mpi=no')
             self.cfg.update('configopts', 'LIBS="$LIBLAPACK"')
 
+        # Add CUDA features
+        cuda_is_dep = 'CUDA' in [i['name'] for i in self.cfg.dependencies()]
```
akesandgren, line 165: All this, including the "cuda" extra_var, can be simplified to
if get_software_root('CUDA'):
 ...

That way you can drop the elif statements and the cuda extra_var.

easybuild/easyblocks/e/elpa.py
```diff
@@ -159,6 +161,24 @@ def run_all_steps(self, *args, **kwargs):
...
+        if cuda_is_dep and (self.cfg['cuda'] is None or self.cfg['cuda']):
+            self.cfg.update('configopts', '--enable-nvidia-gpu')
+            cuda_cc_space_sep = self.cfg.template_values['cuda_cc_space_sep'].replace('.', '').split()
+            # Just one is supported, so pick the highest one (but prioritize sm_80)
```
akesandgren, line 169: Picking the highest one is likely a bad idea since then the code won't run on lower end GPU's in case there sites default settings lists all their models.
In this case it's best to raise an error if there are more than one.

easybuild/easyblocks/e/elpa.py
```diff
@@ -52,6 +53,7 @@ def extra_options():
         """Custom easyconfig parameters for ELPA."""
         extra_vars = {
             'auto_detect_cpu_features': [True, "Auto-detect available CPU features, and configure accordingly", CUSTOM],
+            'cuda': [None, "Enable CUDA build if CUDA is among the dependencies", CUSTOM],
```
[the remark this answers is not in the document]
boegel: This actually make sense: don't make a decision by default, let it depend on whether `CUDA` is included as a dependency or not, but allow explicitely opting in (`True`) or out (`False`).

@damianam is just using the same approach as in #2666 here...

easybuild/easyblocks/e/elpa.py
boegel: That's actually not fully true...

If the `CUDA` module was loaded manually, then `$EBROOTCUDA` is also defined (and that's all that `get_software_root` does: it grabs the corresponding `$EBROOT*` environment variable).

So this is actually a better approach...

We should have proper support for this in framework though, so you can just do `self.cfg.is_dep('CUDA')`

easybuild/easyblocks/e/elpa.py
boegel: Or pick first listed (or minimum rather than max) and print a clear warning?

easybuild/easyblocks/e/elpa.py
ocaisa: For JSC, CUDA is a dep of MPI by default (and that MPI works regardless of whether or not there actually is a GPU through the use of UCX variables). So `get_software_root('CUDA')` is not enough, you need to know if it is actually a dep to decide whether to trigger a GPU build.

easybuild/easyblocks/e/elpa.py
branfosj: The LAMMPS easyblock selects the largest: https://github.com/easybuilders/easybuild-easyblocks/blob/develop/easybuild/easyblocks/l/lammps.py#L307

Having a method, with a clear rational on what is picked, would be good for cases where the software supports building for one, and not all of the, CUDA CC.

easybuild/easyblocks/e/elpa.py
smoors: @ocaisa if CUDA is a dep of MPI regardless of whether or not there actually is a GPU, don’t you want to trigger a GPU build also regardless of whether or not there actually is a GPU?

ocaisa: @smoors No, because then that GPU-enabled build will (typically) not work on a system without a GPU (whereas OpenMPI will work just fine on that system as long as you set the appropriate btl). So at JSC, the GPU supported build is usually explicit

PR #2680, opened 2022-02-14
add fix_shebang to install_step of PythonPackage easyblock so that we can fix shebangs when installing extensions

mboisson:
…nstalling extensions

### DIFF
--- a/easybuild/easyblocks/generic/pythonpackage.py
+++ b/easybuild/easyblocks/generic/pythonpackage.py
@@ -704,6 +704,9 @@ class PythonPackage(ExtensionEasyBlock):
         # (for iterated installations over multiply Python versions)
         self.install_cmd_output += out
 
+        # fix shebangs if specified
+        self.fix_shebang()
+
         # restore env vars if it they were set
         for name in ('PYTHONPATH', 'PATH'):
             value = old_values[name]


mboisson:
The idea is being able to add something like this in an `exts_list` for a `PythonBundle`: 
```
    ('CacheControl', '0.12.10', {
    	'fix_python_shebang_for': ['bin/doesitcache'],
    }),
``` 
or
```
    ('wheel', '0.37.1', {
    	'fix_python_shebang_for': ['bin/wheel'],
    }),
    ('pip', '22.0.3', {
    	'fix_python_shebang_for': ['bin/pip', 'bin/pip3'],
    }),
``` 
in order to be able to install those with `multi_deps` for python.

easybuild/easyblocks/generic/pythonpackage.py
```diff
@@ -704,6 +704,9 @@ def install_step(self):
         # (for iterated installations over multiply Python versions)
         self.install_cmd_output += out
 
+        # fix shebangs if specified
+        self.fix_shebang()
```
akesandgren, line 708: Won't this pick up a shebang list from the non-extension part of the easyconfig?
That may be a bad thing.

mboisson: Possibly. Do you have a suggestion on how to avoid that ?

mboisson: In reality, I always found it weird that extensions pick up *anything* from the non-extension part...  I've been bitten by it more than once using `configopts` and other parameters.

akesandgren: @boegel you should know the answer, does fix_shebang here pick up things from the non-extension part?

boegel: Yes, if `fix_python_shebang_for` (or another `fix_*_shebang_for`) is set for the "parent", then it will be picked up for all extensions as well.

All easyconfig parameters set for the parent are inherited by each extension, which is important in some cases (like `toolchain` for example). There are some exceptions though, see https://github.com/easybuilders/easybuild-framework/blob/95bae20337781a06defe467faf487a122aae8512/easybuild/framework/extension.py#L108

and https://github.com/easybuilders/easybuild-framework/blob/95bae20337781a06defe467faf487a122aae8512/easybuild/framework/extensioneasyblock.py#L81-L86

Changing this now (for EasyBuild v5.0) is worth considering.
There may be some fallout, but that can be handled via `exts_default_options`...

@mboisson If you would like to see this changes, please open a framework issue on this, so we can tag it with the `5.0` milestone.

mboisson: I don't *really* care whether extensions pick up parameters from the parent recipe, but I do know we need to be able to fix shebangs of extensions. This change is in fact already merged in our fork.

boegel: Yeah, I agree.

And even if `fix_*_shebang` is being picked up from the parent once this change is merged, it should still be OK. Worst case, the shebangs get fixed...

boegel:
Test report by @boegel
 * **SUCCESS** _tqdm-4.64.0-GCCcore-11.3.0.eb_ 
 * **SUCCESS** _HTSeq-0.11.3-foss-2021b.eb_ 
 * **SUCCESS** _GTDB-Tk-2.0.0-intel-2021b.eb_ 
 * **SUCCESS** _matplotlib-3.4.2-foss-2021a.eb_ 
 * **SUCCESS** _numexpr-2.8.1-foss-2021a.eb_ 
 * **SUCCESS** _pybind11-2.9.2-GCCcore-11.3.0.eb_ 

node2672.swalot.os - Linux RHEL 8.4, x86_64, Intel(R) Xeon(R) CPU E5-2660 v3 @ 2.60GHz (haswell), Python 3.6.8
See https://gist.github.com/fe59b6de0274bbaacf2e805d514925a6 for a full test report.

boegel (approved):
lgtm

PR #2696, opened 2022-03-10
Add guess for MANPATH for intel-compilers.

### LINKED ISSUE #2695: intel-compilers modules don't set MANPATH
With the newer intel-compilers modules (and thus in the intel toolchain starting from 2021a I guess), the MANPATH variable is not set. The IntelBase block seems to look for a man/ subfolder only:

https://github.com/easybuilders/easybuild-easyblocks/blob/develop/easybuild/easyblocks/generic/intelbase.py#L124

while in the newer installations, the man pages are in a different place, e.g. 

intel-compilers/2021.1.2/compiler/2021.1.2/documentation/en/man/common/

bartoldeman:
Fixes #2695


akesandgren (approved):
LGTM

akesandgren:

PR #2711, opened 2022-04-22
don't use gold linker by default for GCC >= 11.3

### LINKED ISSUE #2657: Revisit default use of `ld.gold` in binutils easyblock
Someone mentioned in Slack that `ld.gold` is no longer being actively developed, so we should probably switch (back) to `ld.bfd` as the default.

bartoldeman:
Gold is suffering from bitrot, it's better to use ld.bfd by default
as most Linux distributions are doing.

Fixes #2657

### DIFF
--- a/easybuild/easyblocks/g/gcc.py
+++ b/easybuild/easyblocks/g/gcc.py
@@ -89,7 +89,8 @@ class EB_GCC(ConfigureMake):
             'pplwatchdog': [False, "Enable PPL watchdog", CUSTOM],
             'prefer_lib_subdir': [False, "Configure GCC to prefer 'lib' subdirs over 'lib64' when linking", CUSTOM],
             'profiled': [False, "Bootstrap GCC with profile-guided optimizations", CUSTOM],
-            'use_gold_linker': [True, "Configure GCC to use GOLD as default linker", CUSTOM],
+            'use_gold_linker': [None, "Configure GCC to use GOLD as default linker "
+                                      "(default: True for GCC < 11.3.0)", CUSTOM],
             'withcloog': [False, "Build GCC with CLooG support", CUSTOM],
             'withisl': [False, "Build GCC with ISL support", CUSTOM],
             'withlibiberty': [False, "Enable installing of libiberty", CUSTOM],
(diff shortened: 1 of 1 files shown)


casparvl:
Just my two cents, since I saw some of @boegel 's questions in #2657  :

I think the approach in this PR is good, since it doesn't change existing behaviour. I.e. if I reinstall something for an older toolchain (GCC < 11.3) I'd like to not change the behaviour if we can help it.

I'm curious though @bartoldeman : did you already do some tests to see if using `ld.bfd` presents any unexpected issues? You could consider compiling a GCCcore-11.2 based stack with `use_gold_linker=False` and just compile some high-level packages with lots of dependencies. I'm hoping it won't lead to substantial issues, but if it does, it's better we know about it :)

bartoldeman:
We used the bfd linker very extensively before with the Nix-based stack (mainly with gcc 5.4.0 and gcc 7.3.0), and I don't remember any issues.
On the contrary using the Gold linker has caused us various issues, Julia and a Lustre filesystem issue come to mind.

It doesn't hurt to test but I think we can just take the plunge during develop and if things really turn out badly we can always revert.

rocm-smi, MariaDB, libunwind are others that need to use `-fuse-ld=bfd` now already.

bartoldeman:
Well I guess I'll need to take it back for now, @boegelbot does not look happy!
I'll try myself to see why binutils didn't compile this way, looks very odd!
https://gist.github.com/boegelbot/43f6fd8d031b1d69d16ce64fe683d0b8#file-binutils-2-38-gcccore-11-3-0_partial-log-L409

bartoldeman:
#2712 fixes the issue with binutils.

casparvl:
Test reports are [here](https://github.com/easybuilders/easybuild-easyconfigs/pull/15311#issuecomment-1107684155), [here](https://github.com/easybuilders/easybuild-easyconfigs/pull/15311#issuecomment-1108554012) and [here](https://github.com/easybuilders/easybuild-easyconfigs/pull/15311#issuecomment-1108582544). All looking good from my point of view.

casparvl (approved):
Switching to `ld.bfd` since `ld.gold` suffers from bitrot seems like a reasonable approach. Current implementation leaves previous behaviour untouched, which makes the most sense to me.

Some issues are to be expected when we add new EasyConfigs (such as the `binutils` issue in #2712), but since other projects are already using `ld.bfd` by default, we can expect most of these to be EasyBuild specific - and thus within our circle of influence to solve. However, since we want to make sure this is the default from the moment GCC-11.3.0 is in EasyBuild, we shouldn't want too long and thus extensive testing is not really an option.

All in all, I say: looks good to me :)

casparvl:
Test report by @casparvl

tcn1.local.snellius.surf.nl - Linux RHEL 8.4, x86_64, AMD EPYC 7H12 64-Core Processor, Python 3.6.8
See https://gist.github.com/dd3a074240f8cba80ab723a4339551ee for a full test report.

bartoldeman:
Thanks, yes, indeed the binutils issue looks like a corner case, something that exposed actually exposed a (mostly harmless) bug (GCCcore binutils linking libiberty from system binutils).

boegel (approved):
lgtm

boegel:
Test report by @boegel
 * **SUCCESS** _GCCcore-11.2.0.eb_ 
 * **SUCCESS** _GCCcore-9.3.0.eb_ 

node3519.doduo.os - Linux RHEL 8.4, x86_64, AMD EPYC 7552 48-Core Processor (zen2), Python 3.6.8
See https://gist.github.com/6b99fb8fb68a7931925f20d5a7d0880d for a full test report.

PR #2713, opened 2022-04-29
Adapt header file check on R for versions >= 4.2.0 (S.h)

i-mtz:
Starting R 4.2.0 the header file `S.h` has been removed  from the sources[C-Level facilities changes](https://cran.r-project.org/bin/windows/base/NEWS.R-devel.html). 

This PR just adds the `S.h` header check for older versions but removes it for `4.2` onwards.


bartoldeman:
Thanks for your first contribution! Approved as obviously correct.

PR #2730, opened 2022-06-02
update libQGLViewer easyblock to take into account changes in the shared library names depending on Qt versions they are compiled with

satishskamath:
For the newer versions of libQGLviewer the qt versions are also included in the library names, therefore sanity check needs to be modified.

### DIFF
--- a/easybuild/easyblocks/l/libqglviewer.py
+++ b/easybuild/easyblocks/l/libqglviewer.py
@@ -67,7 +67,8 @@ class EB_libQGLViewer(ConfigureMake):
                     addition = '-' + addition
             custom_paths = {
                 'files': [('lib/libQGLViewer'+addition+'.prl', 'lib64/libQGLViewer'+addition+'.prl'),
-                          ('lib/libQGLViewer'+addition+'.%s' % shlib_ext, 'lib64/libQGLViewer'+addition+'.%s' % shlib_ext)],
+                          ('lib/libQGLViewer'+addition+'.%s' % shlib_ext, 'lib64/libQGLViewer'+addition+'.%s' \
+                                  % shlib_ext)],
                 'dirs': ['include/QGLViewer'],
             }
 
--- a/easybuild/easyblocks/l/libqglviewer.py
+++ b/easybuild/easyblocks/l/libqglviewer.py
@@ -31,6 +31,7 @@ EasyBuild support for building and installing libQGLViewer, implemented as an ea
 from easybuild.easyblocks.generic.configuremake import ConfigureMake
 from easybuild.tools.run import run_cmd
 from easybuild.tools.systemtools import get_shared_lib_ext
+from easybuild.tools.modules import get_software_root
 from distutils.version import LooseVersion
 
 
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/l/libqglviewer.py
```diff
@@ -49,10 +50,26 @@ def configure_step(self):
...
+                'files': [('lib/libQGLViewer.prl', 'lib64/libQGLViewer.prl'),
+                          ('lib/libQGLViewer.%s' % shlib_ext, 'lib64/libQGLViewer.%s' % shlib_ext)],
+                'dirs': ['include/QGLViewer'],
+            }
+        else:
+            addition = ''
+            for dep in self.cfg.dependencies():
+                if (dep['name'][0] == 'Q' and dep['name'][1] == 't'):
+                    addition = dep['name'].lower()
+                    addition = '-' + addition
```
Micket, line 67: I think this looks a bit cryptic and seems a bit fragile considering things like `QtPy` `Qtconsole`  etc.
I'd say just go with hardcoding Qt5 and 6 for now. We'll worry about generalizing it in 10 years when Qt7 comes.
```suggestion
            for qt_dep in ['Qt5', 'Qt6']:
                if get_software_root('Qt5'):
                    addition = qt_dep.lower()
                    break
            else:
                raise EasyBuildError("Missing Qt5 or Qt6 dependency")
```

Micket (changes requested):

Minidixit (dismissed):

easybuild/easyblocks/l/libqglviewer.py
```diff
@@ -49,10 +52,29 @@ def configure_step(self):
...
+                    break
+            else:
+                raise EasyBuildError("Missing Qt5 or Qt6 dependency")
+            custom_paths = {
+                'files': [('lib/libQGLViewer'+addition+'.prl', 'lib64/libQGLViewer'+addition+'.prl'),
+                          ('lib/libQGLViewer'+addition+'.%s' % shlib_ext,
+                              'lib64/libQGLViewer'+addition+'.%s' % shlib_ext)],
+                'dirs': ['include/QGLViewer'],
+            }
```
Micket, line 78: Sorry for nitpicking here, but i think deduplicating it a bit like this is nicer, and `suffix` seems a bit more appropriate name
```suggestion
        if LooseVersion(self.version) < LooseVersion("2.8.0"):
            suffix = ''
        else:
            for dep in ['Qt5', 'Qt6']:
                if get_software_root(dep):
                    suffix = '-' + dep.lower()
                    break
            else:
                raise EasyBuildError("Missing Qt5 or Qt6 dependency")
        custom_paths = {
            'files': [('lib/libQGLViewer'+addition+'.prl', 'lib64/libQGLViewer'+addition+'.prl'),
                      ('lib/libQGLViewer'+addition+'.%s' % shlib_ext,
                          'lib64/libQGLViewer'+addition+'.%s' % shlib_ext)],
            'dirs': ['include/QGLViewer'],
        }
```

Micket (changes requested):

easybuild/easyblocks/l/libqglviewer.py
```diff
@@ -49,10 +52,25 @@ def configure_step(self):
...
         """Custom sanity check for libQGLViewer."""
         shlib_ext = get_shared_lib_ext()
-
+        """
+        From version 2.8.0 onwards qt version also gets added to the lib file names.
+        """
+        if LooseVersion(self.version) < LooseVersion("2.8.0"):
+            suffix = ''
+        else:
+            suffix = ''
```
Micket, line 61:
```suggestion
```

easybuild/easyblocks/l/libqglviewer.py
```diff
@@ -49,10 +52,25 @@ def configure_step(self):
...
+            for dep in ['Qt5', 'Qt6']:
+                if get_software_root(dep):
+                    suffix = dep.lower()
+                    suffix = '-' + suffix
```
Micket, line 65:
```suggestion
                    suffix = '-' + dep.lower()
```

Micket (changes requested):

Micket (approved):
lgtm

Micket:
I need a test report before i can merge this. Do you have a PR for a newer libQGLViewer?

satishskamath:
> I need a test report before i can merge this. Do you have a PR for a newer libQGLViewer?
https://github.com/easybuilders/easybuild-easyconfigs/pull/15945 , I just created a PR for the easyconfig. But the tests will fail without this PR. :)

Issue #2731, opened 2022-06-03
GROMACS does not play nice as a component


ocaisa:
In https://github.com/easybuilders/easybuild/issues/802 it was observed that the GROMACS easyblock does not play nice as a component of a bundle. It would seem that `self.run_all_steps()` is not being picked from the custom easyblock.

ocaisa:
The issue is that when used as a component, only a limited number of steps are called (see https://github.com/easybuilders/easybuild-easyblocks/blob/develop/easybuild/easyblocks/generic/bundle.py#L297)

Perhaps the best thing to do is error out when we see an easyblock overrides `run_all_steps()` as it is very much a corner case.

ocaisa:
Fixed in #2732

PR #2732, opened 2022-06-03
Don't allow an easyblock that overrides `run_all_steps()` to be used in a bundle

ocaisa:
Can be tested with an extended dry run (`-x`) and the example easyconfig
```
easyblock = 'Bundle'

name = 'GROMACS'
version = '2020.2'

homepage = 'http://www.gromacs.org'

description = 'GROMACS default_easyblock test'

toolchain = {'name': 'intel', 'version': '2020a'}
toolchainopts = {'openmp': True, 'usempi': True}

dependencies = [
    ('CUDA', '11.4.1', '', True)
]

source_urls = ['ftp://ftp.gromacs.org/pub/gromacs/']

default_easyblock = 'CMakeMake'

default_component_specs = {
    'sources':   [SOURCELOWER_TAR_GZ],
    'checksums': [ '7465e4cd616359d84489d919ec9e4b1aaf51f0a4296e693c249e83411b7bd2f3'],
    'separate_build_dir': False,
    'preconfigopts' : '( [ -f CMakeCache.txt ] && /bin/rm CMakeCache.txt ) ; ' +
                      '( [ -f Makefile ] && make clean ) ; ' ,
    'buildopts' : 'all check -j ',
    'start_dir':   '%(namelower)s-%(version)s',
}

components = [
    ('GROMACS', version, {
        'easyblock': 'CMakeMake',
        'configopts': "-DGMX_MPI=OFF -DGMX_THREAD_MPI=ON  -DGMX_DOUBLE=OFF ",
    }),
    ('GROMACS', version, {
        # 'easyblock': 'CMakeMake',
        'configopts': "-DGMX_MPI=ON  -DGMX_THREAD_MPI=OFF -DGMX_DOUBLE=OFF ",
    }),
]

modextravars = {
    'GROMACS_DIR': "%(installdir)s",
    'GMXBIN':      "%(installdir)s/bin",
    'GMXLDLIB':    "%(installdir)s/lib64",
    'GMXMAN':      "%(installdir)s/share/man",
    'GMXDATA':     "%(installdir)s/share/data",
}

moduleclass = 'bio'
```

### DIFF
--- a/easybuild/easyblocks/generic/bundle.py
+++ b/easybuild/easyblocks/generic/bundle.py
@@ -114,7 +114,7 @@ class Bundle(EasyBlock):
             comp_cfg['version'] = comp_version
 
             # determine easyblock to use for this component
-            # - if an easyblock is specified explicitely, that will be used
+            # - if an easyblock is specified explicitly, that will be used
             # - if not, a software-specific easyblock will be considered by get_easyblock_class
             # - if no easyblock was found, default_easyblock is considered
             comp_easyblock = comp_specs.get('easyblock')
@@ -134,6 +134,10 @@ class Bundle(EasyBlock):
 
             if easyblock == 'Bundle':
                 raise EasyBuildError("The Bundle easyblock can not be used to install components in a bundle")
+
+            if easyblock_class.run_all_steps.__code__ is not EasyBlock.run_all_steps.__code__:
+                raise EasyBuildError("Easyblock %s overrides the run_all_steps() method and can not be used to install "
+                                     "a component in a bundle", easyblock)
 
             comp_cfg.easyblock = easyblock_class
 


ocaisa:
Seems to work with both Python 2 and 3

easybuild/easyblocks/generic/bundle.py
```diff
@@ -150,6 +154,11 @@ def __init__(self, *args, **kwargs):
             comp_cfg['sources'] = comp_cfg['source_urls'] = comp_cfg['checksums'] = comp_cfg['patches'] = []
 
             for key in self.cfg['default_component_specs']:
+                if key == 'easyblock':
+                    raise EasyBuildError("You cannot use 'easyblock' within 'default_component_specs', you must use "
+                                         "the 'default_easyblock' easyconfig parameter (which will still prefer "
+                                         "a custom easyblock for a component when one exists) or explicitly specify "
+                                         "the easyblock to be used for the component")
```
jfgrimm, line 161:
```suggestion
                    raise EasyBuildError("You cannot use 'easyblock' within 'default_component_specs'. Instead, use "
                                         "the 'default_easyblock' easyconfig parameter (a custom easyblock for a component  "
                                         "will still be preferred if it exists), or explicitly specify the easyblock to be used for"
                                         "each component")
```

easybuild/easyblocks/generic/bundle.py
```diff
@@ -135,6 +135,10 @@ def __init__(self, *args, **kwargs):
             if easyblock == 'Bundle':
                 raise EasyBuildError("The Bundle easyblock can not be used to install components in a bundle")
 
+            if easyblock_class.run_all_steps.__code__ is not EasyBlock.run_all_steps.__code__:
+                raise EasyBuildError("Easyblock %s overrides the run_all_steps() method and can not be used to install "
```
jfgrimm, line 139:
```suggestion
                raise EasyBuildError("Easyblock %s overrides the run_all_steps() method, and cannot be used to install "
```

boegel:
@ocaisa I wouldn't say that this fixes #2731, but it is helpful as it at least makes sure that the easyblock can be used to install a `Bundle` component...

Is there any way that we could make sure that even if an easyblock redefines `run_all_steps()`, that it could still be used to install a `Bundle` component?

ocaisa:
I'm sure there's a way, but that would be quite involved to implement I suspect and is of little value, there are only a handful of easyblocks that do that and the number of those that would be expected to be used in a bundle must be not far off zero.

PR #2742, opened 2022-06-09
add support for allowing some PyTorch tests to fail + print warning if one or more tests fail

boegel:
Draft because:
* needs more testing
* need to check whether `10` is a good default for `max_failed_tests`
* maybe we should also look into a way of getting the warning pushed "up" so it's included in the comment that is posted for PR test reports (but that will require some changes in framework too)

### DIFF
--- a/easybuild/easyblocks/p/pytorch.py
+++ b/easybuild/easyblocks/p/pytorch.py
@@ -270,11 +270,11 @@ class EB_PyTorch(PythonPackage):
             max_failed_tests = self.cfg['max_failed_tests']
 
             test_or_tests = 'tests' if failed_test_cnt > 1 else 'test'
-            failed_tests_txt = '\n'.join('* %s' % t for t in sorted(failed_tests))
-            msg = "%d %s (out of %d) failed:\n%s"
+            msg = "%d %s (out of %d) failed:\n" % (failed_test_cnt, test_or_tests, test_cnt)
+            msg += '\n'.join('* %s' % t for t in sorted(failed_tests))
 
             if max_failed_tests == 0:
-                raise EasyBuildError(msg, failed_test_cnt, test_or_tests, test_cnt, failed_tests_txt)
+                raise EasyBuildError(msg)
             else:
                 msg += '\n\n' + ' '.join([
                     "The PyTorch test suite is known to include some flaky tests,",
@@ -284,7 +284,7 @@ class EB_PyTorch(PythonPackage):
                     "are known to be flaky, or do not affect your intended usage of PyTorch.",
                     "In case of doubt, reach out to the EasyBuild community (via GitHub, Slack, or mailing list).",
                 ])
-                print_warning(msg, failed_test_cnt, test_or_tests, test_cnt, failed_tests_txt)
+                print_warning(msg)
 
                 if failed_test_cnt > max_failed_tests:
                     raise EasyBuildError("Too many failed tests (%d), maximum allowed is %d",


ocaisa:
In https://github.com/easybuilders/easybuild-framework/pull/4001 there's the log parsing tool that might be useful here. Have you tried that on the log file to see what it would spit out?

boegel:
@casparvl As soon as the PyTorch test command fails (non-zero exit code), a warning message will be printed (to stderr).
When one or more failing tests were found based on the regex, the test names will be listed too.

See this output (from a test report in #15137):
```
WARNING: Test command had non-zero exit code (1)!


WARNING: 5 tests (out of 89230) failed:
* distributed/fsdp/test_fsdp_input
* distributed/fsdp/test_fsdp_memory
* distributed/fsdp/test_fsdp_overlap
* distributed/test_c10d_gloo
* test_autograd
* ```

**Comment by casparvl** [comment]
Oh ok, that looks fine to me :)

Seeing that output, I'm only wondering one thing: should we give the user some _advice_ (below the list of failing tests)? I.e. something along the lines of "The PyTorch test suite often has a small number of tests fail. As long as these are less than `self.cfg['max_failed_tests']`, EasyBuild completes the installation, since it is often these tests that are broken, and not your installation. However, you may want to double check that your installation works as expected." (I'm not actually sure _how_ to phrase it, but _some_ indication of what to _do_ with this warning)

**Comment by boegel** [comment]
Test report by @boegel
 * **SUCCESS** _PyTorch-1.9.0-fosscuda-2020b.eb_ 

node3902.accelgor.os - Linux RHEL 8.4, x86_64, AMD EPYC 7413 24-Core Processor (zen3), 1 x NVIDIA NVIDIA A100-SXM4-80GB, 510.73.08, Python 3.6.8
See https://gist.github.com/a52aeee7a6a9c7ba03bb5f9272d6f1a2 for a full test report.

**Comment by Flamefire** [comment]
I don't really like the approach of allowing ANY test to fail via this new option. Why not disable the tests which we know are failing as we did before?

**Comment by boegel** [comment]
> I don't really like the approach of allowing ANY test to fail via this new option. Why not disable the tests which we know are failing as we did before?

Mainly because that's a very painful process, see how long https://github.com/easybuilders/easybuild-easyconfigs/pull/15137 has been open due to flaky tests (and test reports that take half a day to run).
We're talking 5 tests over 90k that are causing a lot of headaches.

If people want to be very strict and not allow any tests to fail, they can set `max_failed_tests` to `0` via a hook

**Comment by branfosj** [comment]
Ideally we'd have no test failures, but we do see failures and often we see a test fail for only one system / setup. I consider this PR a suitable compromise and with a suitable message going into the standard output allowing people to investigate further.

File: easybuild/easyblocks/p/pytorch.py [hunk 1]
```diff
@@ -49,9 +50,10 @@ class EB_PyTorch(PythonPackage):
     def extra_options():
         extra_vars = PythonPackage.extra_options()
         extra_vars.update({
-            'excluded_tests': [{}, 'Mapping of architecture strings to list of tests to be excluded', CUSTOM],
-            'custom_opts': [[], 'List of options for the build/install command. Can be used to change the defaults ' +
-                                'set by the PyTorch EasyBlock, for example ["USE_MKLDNN=0"].', CUSTOM],
+            'custom_opts': [[], "List of options for the build/install command. Can be used to change the defaults " +
+                                "set by the PyTorch EasyBlock, for example ['USE_MKLDNN=0'].", CUSTOM],
+            'excluded_tests': [{}, "Mapping of architecture strings to list of tests to be excluded", CUSTOM],
+            'max_failed_tests': [10, "Maximum number of failing tests", CUSTOM],
```
Flamefire, line 56: I would set the default to 0 as e.g. existing PyTorch ECs don't have any failed, non-excluded tests (or should at least).
Also having to set this explicitly in the EC makes it clear that there is something "fishy" with that specific PyTorch version.

easybuild/easyblocks/p/pytorch.py
```diff
@@ -253,7 +255,31 @@ def test_step(self):
             'python': self.python_cmd,
             'excluded_tests': ' '.join(excluded_tests)
         })
-        super(EB_PyTorch, self).test_step()
+
+        (out, ec) = super(EB_PyTorch, self).test_step(return_output_ec=True)
+        if ec:
+            print_warning("Test command had non-zero exit code (%s)!" % ec)
```
Flamefire, line 261: I would move that down so it is only printed when the "X tests failed" is NOT printed. And then possibly adjust the message. Currently this is over-verbose:
> Test command had non-zero exit code
> 10 test (out of 20) failed: ...
> Too many failed tests ...

easybuild/easyblocks/p/pytorch.py
```diff
@@ -253,7 +255,31 @@ def test_step(self):
...
+        ran_tests_hits = ran_tests_regex.findall(out)
+        test_cnt = 0
+        for hit in ran_tests_hits:
+            test_cnt += int(hit)
+
+        failed_test_regex = re.compile(r"^(?P<failed_test_name>.*) failed!\s*$", re.M)
+        failed_tests = nub(failed_test_regex.findall(out))
+        failed_test_cnt = len(failed_tests)
+
+        if failed_test_cnt:
```
Flamefire, line 273: If that is zero, we need to check the `ec` and error on that instead of only warn

easybuild/easyblocks/p/pytorch.py
```diff
@@ -253,7 +255,31 @@ def test_step(self):
...
+            test_or_tests = 'tests' if failed_test_cnt > 1 else 'test'
+            failed_tests_txt = '\n'.join('* %s' % t for t in sorted(failed_tests))
+            msg = "%d %s (out of %d) failed:\n%s"
+            print_warning(msg, failed_test_cnt, test_or_tests, test_cnt, failed_tests_txt)
+
+            max_failed_tests = self.cfg['max_failed_tests']
+            if failed_test_cnt > max_failed_tests:
```
Flamefire, line 280: An adjusted error for `max_failed_test==0` would be good.
Suggestion for `max_failed_test==0`: raise the above warning as an error instead

easybuild/easyblocks/p/pytorch.py
```diff
@@ -253,7 +255,31 @@ def test_step(self):
...
             'excluded_tests': ' '.join(excluded_tests)
         })
-        super(EB_PyTorch, self).test_step()
+
+        (out, ec) = super(EB_PyTorch, self).test_step(return_output_ec=True)
+        if ec:
+            print_warning("Test command had non-zero exit code (%s)!" % ec)
+
+        ran_tests_regex = re.compile(r"^Ran (?P<test_cnt>[0-9]+) tests in", re.M)
+        ran_tests_hits = ran_tests_regex.findall(out)
```
Flamefire, line 264: Why the explicit compile?
```suggestion
        ran_tests_hits = re.findall(r"^Ran (?P<test_cnt>[0-9]+) tests in", out, re.M)
```

easybuild/easyblocks/p/pytorch.py
```diff
@@ -253,7 +255,31 @@ def test_step(self):
...
+        test_cnt = 0
+        for hit in ran_tests_hits:
+            test_cnt += int(hit)
+
+        failed_test_regex = re.compile(r"^(?P<failed_test_name>.*) failed!\s*$", re.M)
+        failed_tests = nub(failed_test_regex.findall(out))
```
Flamefire, line 270: See above:
```suggestion
        failed_tests = nub(re.findall(r"^(?P<failed_test_name>.*) failed!\s*$", out, re.M))
```

Flamefire (changes requested):
> Ideally we'd have no test failures, but we do see failures and often we see a test fail for only one system / setup. I consider this PR a suitable compromise and with a suitable message going into the standard output allowing people to investigate further.

I know. I just don't see the reason to run the tests if we allow any test to fail, not only those that are expected to/have failed before and ideally have an upstream issue to refer to.

Anyway see the review. Basically default to no expected failures and some improved handling of the error case. Especially not ignoring `ec` as that could lead to a success when no tests have even started. So only case where `ec` reports and error but the test-step is allowed to succeed is when we have `failed_test_cnt <= max_failed_tests`
Also a nit about the explicit `re.compile` which I wouldn't use

easybuild/easyblocks/p/pytorch.py
boegel: @casparvl Thoughts on this?

ocaisa: I agree with @Flamefire here, we should _know_ what the failing tests are and how many, putting the number in the easyconfig file is a way of ensuring that it at least gets documented.

easybuild/easyblocks/p/pytorch.py
boegel: fixed in 82011bf

easybuild/easyblocks/p/pytorch.py
boegel: fixed in 82011bf

easybuild/easyblocks/p/pytorch.py
boegel: fixed in 82011bf

easybuild/easyblocks/p/pytorch.py
boegel: fixed in 82011bf

easybuild/easyblocks/p/pytorch.py
```diff
@@ -253,7 +255,42 @@ def test_step(self):
...
+
+        failed_tests = nub(re.findall(r"^(?P<failed_test_name>.*) failed!\s*$", tests_out, re.M))
+        failed_test_cnt = len(failed_tests)
+
+        if failed_test_cnt:
+            max_failed_tests = self.cfg['max_failed_tests']
+
+            test_or_tests = 'tests' if failed_test_cnt > 1 else 'test'
+            failed_tests_txt = '\n'.join('* %s' % t for t in sorted(failed_tests))
+            msg = "%d %s (out of %d) failed:\n%s"
```
Flamefire, line 274:
```suggestion
            msg = "%d %s (out of %d) failed:\n" % (failed_test_cnt, test_or_tests, test_cnt)
            msg += '\n'.join('* %s' % t for t in sorted(failed_tests))
```

easybuild/easyblocks/p/pytorch.py
```diff
@@ -253,7 +255,42 @@ def test_step(self):
...
+            else:
+                msg += '\n\n' + ' '.join([
+                    "The PyTorch test suite is known to include some flaky tests,",
+                    "which may fail depending on the specifics of the system or the context in which they are run.",
+                    "For this PyTorch installation, EasyBuild allows up to %d tests to fail." % max_failed_tests,
+                    "We recommend to double check that the failing tests listed above ",
+                    "are known to be flaky, or do not affect your intended usage of PyTorch.",
+                    "In case of doubt, reach out to the EasyBuild community (via GitHub, Slack, or mailing list).",
+                ])
+                print_warning(msg, failed_test_cnt, test_or_tests, test_cnt, failed_tests_txt)
```
Flamefire, line 287: I find this unreadable: The immediately visible text doesn't contain placeholders for those 4(!) arguments.

So as the 4 placeholders in the initial `msg` will always be inserted, I'd strongly suggest to do it above. See the suggestion

```suggestion
                print_warning(msg)
```

Flamefire (changes requested):
One more round as I had trouble understanding what would be placed where in the message.

boegel:
Test report by @boegel
 * **SUCCESS** _PyTorch-1.3.1-foss-2019b-Python-3.7.4.eb_ 

node3622.doduo.os - Linux RHEL 8.4, x86_64, AMD EPYC 7552 48-Core Processor (zen2), Python 3.6.8
See https://gist.github.com/155048595da30306c187d3f31513dddf for a full test report.

PR #2744, opened 2022-06-09
update Rosetta easyblock to take into account that $LD_LIBRARY_PATH, $CPATH, $PATH may not be defined

mboisson:
LD_LIBRARY_PATH, CPATH and PATH are not guaranteed to be defined, calling .split on None results in an error.

### DIFF
--- a/easybuild/easyblocks/r/rosetta.py
+++ b/easybuild/easyblocks/r/rosetta.py
@@ -121,9 +121,12 @@ class EB_Rosetta(EasyBlock):
         self.log.debug("List of extra environment variables to pass down: %s" % str(env_vars))
 
         # create user.settings file
-        paths = os.getenv('PATH').split(':')
-        ld_library_paths = os.getenv('LD_LIBRARY_PATH').split(':')
-        cpaths = os.getenv('CPATH').split(':')
+        paths = os.getenv('PATH')
+        paths = paths.split(':') if paths else []
+        ld_library_paths = os.getenv('LD_LIBRARY_PATH')
+        ld_library_paths = ld_library_paths.split(':') if ld_library_paths else []
+        cpaths = os.getenv('CPATH')
+        cpaths = cpaths.split(':') if cpaths else []
         flags = [str(f).strip('-') for f in self.toolchain.variables['CXXFLAGS'].copy()]
 
         txt = '\n'.join([


mboisson:
In particular, on our system, `LD_LIBRARY_PATH` is empty, so this piece of code fails because it calls `.split()` on `None`

easybuild/easyblocks/r/rosetta.py
```diff
@@ -121,9 +121,12 @@ def configure_step(self):
...
         # create user.settings file
-        paths = os.getenv('PATH').split(':')
-        ld_library_paths = os.getenv('LD_LIBRARY_PATH').split(':')
-        cpaths = os.getenv('CPATH').split(':')
+        paths = os.getenv('PATH')
+        paths = paths.split(':') if paths else []
+        ld_library_paths = os.getenv('LD_LIBRARY_PATH')
+        ld_library_paths = ld_library_paths.split(':') if ld_library_paths else []
+        cpaths = os.getenv('CPATH')
+        cpaths = cpaths.split(':') if cpaths else []
```
branfosj, line 129:
```suggestion
        paths = os.getenv('PATH', '').split(':')
        ld_library_paths = os.getenv('LD_LIBRARY_PATH', '').split(':')
        cpaths = os.getenv('CPATH', '').split(':')
```

branfosj (changes requested):

easybuild/easyblocks/r/rosetta.py
branfosj: Though that gives `['']`, so ignore me.

boegel:
Test report by @boegel
 * **SUCCESS** _Rosetta-2016.46.59086-foss-2016b.eb_ 

node2608.swalot.os - Linux CentOS Linux 7.9.2009, x86_64, Intel(R) Xeon(R) CPU E5-2660 v3 @ 2.60GHz (haswell), Python 2.7.5
See https://gist.github.com/e95fb1435abb4d5adafd2cc74826f1a2 for a full test report.

easybuild/easyblocks/r/rosetta.py
boegel: @branfosj That's not exactly the same:

```
>>> os.getenv('NOTSET', '').split(':')
['']
```

Using a list with a single empty string element vs an empty list is very different...

branfosj: Yes, hence me dismissing my requested change.

I'd forgotten that Python treats `split()` with and without a delimiter differently:
```
>>> os.getenv('BAR', '').split(':')
['']
>>> os.getenv('BAR', '').split()
[]
```

PR #2748, opened 2022-06-15
make sure that CMakeMakeCp uses correct build dir

boegel:
When the `CMakeMakeCp` easyblock is used to install a bundle component while other components also use the `CMakeMake` easyblock (or one that derives from it), blindly using the `easybuild_obj` build directory is wrong, since that was already created for another component.

The `CMakeMake` easyblock is aware of this, and ensure creating a unique build directory via `create_unused_dir`, but this was not taken into account in the `CMakeMakeCp` easyblock which hardcoded `'easybuild_obj'`...


akesandgren:
Going in, thanks @boegel!

PR #2749, opened 2022-06-17
improve PLUMED detection in GROMACS easyblock

ostueker:
This PR adds a custom option to control whether PLUMED patches are applied.

The default remains auto-detection (apply PLUMED if dependency has been loaded), but also offers the option to request PLUMED (and fail with an exception if not found) or force to ignore PLUMED (in case an incompatible version has been pulled in by another dependency).

### DIFF
--- a/easybuild/easyblocks/g/gromacs.py
+++ b/easybuild/easyblocks/g/gromacs.py
@@ -204,16 +204,17 @@ class EB_GROMACS(CMakeMake):
 
         # check whether PLUMED is loaded as a dependency
         plumed_root = get_software_root('PLUMED')
-        if  str(self.cfg['with_plumed']).upper() == 'TRUE' :
+        if str(self.cfg['with_plumed']).upper() == 'AUTO':
+            if plumed_root:
+                self.log.info('PLUMED has been auto-detected.')
+        elif self.cfg['with_plumed']:
             if not plumed_root:
                 msg = "Compilation with PLUMED was expicitly selected, but PLUMED was not found."
                 raise EasyBuildError(msg)
-        elif str(self.cfg['with_plumed']).upper() == 'FALSE' :
+        else:
             if plumed_root:
                 self.log.info('PLUMED was found, but compilation without PLUMED has been requested.')
             plumed_root = None
-        elif str(self.cfg['with_plumed']).upper() == 'AUTO':
-            pass
 
         if plumed_root:
             # Need to check if PLUMED has an engine for this version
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/g/gromacs.py
```diff
@@ -201,9 +203,20 @@ def configure_step(self):
                 # to avoid that GROMACS finds and uses a system-wide CUDA compiler
                 self.cfg.update('configopts', "-DGMX_GPU=OFF")
 
-        # check whether PLUMED is loaded as a dependency
+        # PLUMED detection
+        # enable PLUMED support if PLUMED is listed as a dependency
+        # and PLUMED support is either explicitly enabled (plumed = True) or unspecified ('plumed' not defined)
         plumed_root = get_software_root('PLUMED')
+        if self.cfg['plumed'] and not plumed_root:
+            msg = "The PLUMED module needs to be loaded to build GROMACS with PLUMED support."
```
akesandgren, line 211: msg = "PLUMED is not listed as a dependency but PLUMED support has been requested."
or
msg = "PLUMED support has been requested but PLUMED is not listed as a dependency."

I.e. it's not about the PLUMED module as such since it may not exist when starting the GROMACS build and it should not be loaded manually before running the eb command.

akesandgren (changes requested):

easybuild/easyblocks/g/gromacs.py
ostueker: Thanks. Indeed your suggestions are more clear.  
I went with the second one.

ostueker:
I just want to mention that I have addressed @akesandgren's  suggestions in my commit b852542 several weeks ago.

branfosj:
Test report by @branfosj
 * **SUCCESS** _GROMACS-2021.3-foss-2021a-CUDA-11.3.1-PLUMED-2.7.2.eb_ 
 * **SUCCESS** _GROMACS-2021.5-foss-2021b.eb_ 

bear-pg0105u36b.bear.cluster - Linux RHEL 8.6, x86_64, Intel(R) Xeon(R) Platinum 8360Y CPU @ 2.40GHz (icelake), Python 3.6.8
See https://gist.github.com/383bf0b02b7399e2391ae63cb00124bc for a full test report.

branfosj:
Tests, both are go as expected:
* easyconfig with no `PLUMED` dep and `plumed = True` fails with `PLUMED support has been
requested but PLUMED is not listed as a dependency.`
* easyconfig with `PLUMED` dep and `plumed = False` builds. Log has `INFO PLUMED was found, but compilation without PLUMED has been requested.`

branfosj (approved):
lgtm

branfosj:
Going in, thanks @ostueker!

PR #2750, opened 2022-06-17
GROMACS with CP2K QM/MM

ostueker:
This PR implements building GROMACS with CP2K QM/MM enabled.

As such a version of GROMACS needs to be compiled statically, which doesn't generate any libraries, the sanity_check needs to be adjusted to not expect any libs in such a case.

### DIFF
--- a/easybuild/easyblocks/g/gromacs.py
+++ b/easybuild/easyblocks/g/gromacs.py
@@ -203,6 +203,8 @@ class EB_GROMACS(CMakeMake):
                 self.cfg.update('configopts', "-DGMX_GPU=OFF")
 
         # CP2K detection
+        # enable CP2K support if CP2K is listed as a dependency
+        # and CP2K support is either explicitly enabled (cp2k = True) or unspecified ('cp2k' not defined)
         cp2k_root = get_software_root('CP2K')
         if self.cfg['cp2k'] and not cp2k_root:
             msg = "The CP2K module needs to be loaded to build GROMACS with CP2K support."
@@ -211,13 +213,11 @@ class EB_GROMACS(CMakeMake):
             self.log.info('CP2K was found, but compilation without CP2K has been requested.')
             cp2k_root = None
 
-        # enable CP2K support if CP2K is listed as a dependency
-        # and CP2K support is either explicitly enabled (cp2k = True) or unspecified ('cp2k' not defined)
-        if cp2k_root and (self.cfg['cp2k'] or self.cfg['cp2k'] is None):
+        if cp2k_root:
             if LooseVersion(self.version) < LooseVersion('2022'):
                 msg = 'CP2K support is only available for GROMACS 2022 and newer.'
                 raise EasyBuildError(msg)
-            elif cp2k_root and LooseVersion(get_software_version('CP2K')) < LooseVersion('8.1'):
+            elif LooseVersion(get_software_version('CP2K')) < LooseVersion('8.1'):
                 msg = 'CP2K support in GROMACS requires CP2K version 8.1 or higher.'
                 raise EasyBuildError(msg)
 
--- a/easybuild/easyblocks/g/gromacs.py
+++ b/easybuild/easyblocks/g/gromacs.py
@@ -68,6 +68,7 @@ class EB_GROMACS(CMakeMake):
             'mpiexec_numproc_flag': ['-np', "Flag to introduce the number of MPI tasks when running tests", CUSTOM],
             'mpi_numprocs': [0, "Number of MPI tasks to use when running tests", CUSTOM],
             'ignore_plumed_version_check': [False, "Ignore the version compatibility check for PLUMED", CUSTOM],
+            'cp2k': [None, "Build with CP2K QM/MM. None is auto-detect. True or False forces behaviour.", CUSTOM],
         })
         extra_vars['separate_build_dir'][0] = True
         return extra_vars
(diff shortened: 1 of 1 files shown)


ostueker:
I found an issue when building this for AVX512. Will push some changes to this soon.

ostueker:
Resolved merge-conflict caused by changes in #2749 
This PR is ready for review and this easyblock is already in use at ComputeCanada

easybuild/easyblocks/g/gromacs.py
```diff
@@ -203,6 +205,55 @@ def configure_step(self):
...
+                # These are for OpenMPI (mpifort --showme).
+                "-lmpi_usempif08 -lmpi_usempi_ignore_tkr -lmpi_mpifh",
+                "-L%s/lib/exts/dbcsr" % cp2k_root,
+                # get depenencies for libcp2k.a:
+                "$(pkg-config --libs-only-l libcp2k)"
+            ]
+            if get_software_root('Libint'):
+                # for some reason libint2 is not discovered by pkg-config:
+                cp2k_linker_flags.append('-lint2')
+            self.cfg.update('configopts', '-DCP2K_LINKER_FLAGS="%s"' % " ".join(cp2k_linker_flags))
```
Micket, line 255: This made me a bit curious. It's not like we are doing something strange to Libint here. It should work if we have pkg-config/pkgconf as a dependency.

```
[ohmanm@vera2 ~]$ ml Libint/2.6.0-iimpi-2021a-lmax-6-cp2k pkg-config/0.29.2-GCCcore-10.3.0 
[ohmanm@vera2 ~]$ pkg-config --libs-only-l libint2
-lint2
```

For CP2K/8.2 on the other hand, there doesn't seem to be any pkgconfig file installed, in fact, no libcp2k at all, so i guess this also depends on some updated CP2K easyblock that also builds a library?

Also, since pkg-config is explicitly called here, perhaps the logic should check that pkg-config is added explicitly as a build-dep.

Micket (commented):
In order to merge any easyblock changes, we need easyconfigs to test with first.

easybuild/easyblocks/g/gromacs.py
ostueker: Thanks for the feedback.   Indeed CP2K needs to be build with `library = True`.

On our system we have pkg-config in our Gentoo layer and therefore we don't specify that as a dependency.

I'll prepare a PR with the needed easyconfigs as best as I can over the coming days.
There will be some guesswork with the dependencies as our easybuild installation is a bit different in certain ways.

ostueker:
I've opened easybuild-easyconfigs#16362 with the missing easyconfigs to test this patch.

I'll add the check for `pkg-config` shortly.

ostueker:
Thanks @Micket, 

I've added some checks that CP2K has been compiled with `library = True` and that `pkg-config` was used.
It also checks that `pkg-config`  is has been loaded (e.g. as a build-dep).

Easyconfigs to test this enhancement are in: easybuilders/easybuild-easyconfigs#16362

easybuild/easyblocks/g/gromacs.py
```diff
@@ -203,6 +205,65 @@ def configure_step(self):
...
+                msg = "pkg-config is required as a build-dependency for building GROMACS-CP2K"
+                raise EasyBuildError(msg)
+
+            self.log.info('CP2K support has been enabled.')
+            # Building with CP2K requires static build w/o gmxapi.
+            # https://manual.gromacs.org/documentation/2022/install-guide/index.html#building-with-cp2k-qm-mm-support
+            self.log.info("Building with CP2K QM/MM.")
+            self.cfg['build_shared_libs'] = False
+            self.libext = 'a'
+            cp2k_version = get_software_version('CP2K')
```
akesandgren, line 247: Move this up above the first use of CP2K version

easybuild/easyblocks/g/gromacs.py
```diff
@@ -203,6 +205,65 @@ def configure_step(self):
...
+            raise EasyBuildError(msg)
+        elif cp2k_root and self.cfg['cp2k'] is False:
+            self.log.info('CP2K was found, but compilation without CP2K has been requested.')
+            cp2k_root = None
+
+        if cp2k_root:
+            if LooseVersion(self.version) < LooseVersion('2022'):
+                msg = 'CP2K support is only available for GROMACS 2022 and newer.'
+                raise EasyBuildError(msg)
+            elif LooseVersion(get_software_version('CP2K')) < LooseVersion('8.1'):
```
akesandgren, line 223: This can be a simple "if", then one can put the 
```
cp2k_version = get_software_version('CP2K')
```
before this line.

easybuild/easyblocks/g/gromacs.py
```diff
@@ -203,6 +205,65 @@ def configure_step(self):
...
+            self.cfg.update('configopts', "-DGMX_VERSION_STRING_OF_FORK=CP2K-{:}".format(cp2k_version))
+            self.cfg.update('configopts', "-DCP2K_DIR=%s/lib64" % cp2k_root)
+            cp2k_linker_flags = [
+                # Need MPI linker flags b/c libcp2k.a is compiled with mpifort.
+                # These are for OpenMPI (mpifort --showme).
+                "-lmpi_usempif08 -lmpi_usempi_ignore_tkr -lmpi_mpifh",
+                "-L%s/lib/exts/dbcsr" % cp2k_root,
+                # get depenencies for libcp2k.a:
+                "$(pkg-config --libs-only-l libcp2k)"
+            ]
```
akesandgren, line 261: This is not good. We must not depend on it being built with OpenMPI. To do it this way it MUST first check that OpenMPI is part of the toolchain.
And the pkgconfig for libcp2k should probably include the MPI libs it depends on which "should" then make all this redundant if the CMake for GROMACS does things correctly.

akesandgren (changes requested):

easybuild/easyblocks/g/gromacs.py
ostueker: Unfortunately CP2K's pkgconfig doesn't contain the MPI libs. That's why I had to hack them in manually.

```
$ pkg-config --libs-only-l libcp2k
-lcp2k -ldbcsr -lm -lpthread -lstdc++ -lxcf03 -lxc -lxsmm -lxsmmf -lopenblas -lgfortran -lfftw3_omp -lfftw3 -lpthread -lopenblas -lgfortran -lpthread -lscalapack -lopenblas -lgfortran
```
Should I rework it like this?

```python
cp2k_linker_flags = []
# Need MPI linker flags b/c libcp2k.a is compiled with mpifort.
# These are for OpenMPI (mpifort --showme).
if get_software_root('OpenMPI'):
    cp2k_linker_flags.append("-lmpi_usempif08 -lmpi_usempi_ignore_tkr -lmpi_mpifh")
elif get_software_root('IntelMPI'):
    # append Intel MPI libs for fortran here:
    cp2k_linker_flags.append("...")

cp2k_linker_flags += [
    "-L%s/lib/exts/dbcsr" % cp2k_root,
    # get dependencies for libcp2k.a:
    "$(pkg-config --libs-only-l libcp2k)"
]

```

easybuild/easyblocks/g/gromacs.py
ostueker: done in 6b1f77d88

easybuild/easyblocks/g/gromacs.py
ostueker: done in 6b1f77d88

easybuild/easyblocks/g/gromacs.py
ostueker: In 2bacbb3fc the easyblock is now checking for the cases of Open MPI and Intel MPI to add the fortran-MPI libraries.
Other MPI implementations could be added if needed.

akesandgren (changes requested):

ostueker:
The easyblock has diverged too much from this PR.
I will likely start this over again at some point.

PR #2788, opened 2022-09-19
make the check for `mpirun` optional in OpenMPI easyblock

damianam:


easybuild/easyblocks/o/openmpi.py
```diff
@@ -160,11 +160,12 @@ def load_module(self, *args, **kwargs):
     def sanity_check_step(self):
         """Custom sanity check for OpenMPI."""
 
-        bin_names = ['mpicc', 'mpicxx', 'mpif90', 'mpifort', 'mpirun', 'ompi_info', 'opal_wrapper']
+        bin_names = ['mpicc', 'mpicxx', 'mpif90', 'mpifort', 'ompi_info', 'opal_wrapper']
         if LooseVersion(self.version) >= LooseVersion('5.0.0'):
             bin_names.append('prterun')
```
boegel, line 165: @damianam We're now no longer checking for `mpirun` for OpenMPI v5.0.0, is that intentional?

easybuild/easyblocks/o/openmpi.py
```diff
@@ -160,11 +160,12 @@ def load_module(self, *args, **kwargs):
...
         else:
-            bin_names.append('orterun')
+            if '--with-orte=no' not in self.cfg['configopts'] and '--without-orte' not in self.cfg['configopts']:
```
boegel, line 167: Can you clarify this a bit? Does configuring without `orte` imply not having `mpirun` at all?

damianam: Exactly. `mpirun` is not built. `orterun` is built, but at least IMO not really used and probably should not be built either.

akesandgren: without orterun for older versions you can't start openmpi code. It's always used.

damianam:
This was relevant a bit ago, but according to https://github.com/spack/spack/issues/44902#issuecomment-2194977313 it is no longer desirable for 5.x, so I think we can drop this.

PR #2797, opened 2022-10-06
Make the test output from PythonPackage less verbose

Flamefire:
When the EasyBlock wants to handle the test output itself it doesn't make sense to search it for potential errors using the generic RegExp.

This applies e.g. to PyTorch and TensorFlow which search for test failures themselves.

So set `regexp=False` to avoid making the log much harder to read due to duplicated and, more annoyingly, partial test (fail) output.


bartoldeman (approved):
LGTM

PR #2808, opened 2022-10-18
add custom easyblock for mamba

casparvl:
Mamba is an drop-in replacement for Conda. Even the installation process is similar, it's just that the Anaconda EasyBlock hardcodes a `sanity_check_path` and we need that to be slightly different for `mamba`. Hence, we inherit from `EB_Anaconda`

### DIFF
--- a/easybuild/easyblocks/m/mamba.py
+++ b/easybuild/easyblocks/m/mamba.py
@@ -39,11 +39,11 @@ class EB_Mamba(EB_Anaconda):
 
     def sanity_check_step(self):
         """
-        Custom sanity check for Anaconda and Miniconda
+        Custom sanity check for Mamba
         """
         custom_paths = {
             'files': [os.path.join('bin', x) for x in ['2to3', 'conda', 'pydoc', 'python', 'mamba']],
-            'dirs': ['bin', 'etc', 'lib', 'pkgs'],
+            'dirs': ['etc', 'lib', 'pkgs'],
         }
         # Directly call EB_Anaconda's super, as this sanity_check_step should _overwrite_ Anaconda's (not call it)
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/m/mamba.py
```diff
@@ -0,0 +1,49 @@
...
+class EB_Mamba(EB_Anaconda):
+    """Support for building/installing Mamba."""
+
+    def sanity_check_step(self):
+        """
+        Custom sanity check for Anaconda and Miniconda
+        """
+        custom_paths = {
+            'files': [os.path.join('bin', x) for x in ['2to3', 'conda', 'pydoc', 'python', 'mamba']],
+            'dirs': ['bin', 'etc', 'lib', 'pkgs'],
```
jfgrimm, line 46: already implicitly testing for `bin` in `files`
```suggestion
            'dirs': ['etc', 'lib', 'pkgs'],
```

easybuild/easyblocks/m/mamba.py
```diff
@@ -0,0 +1,49 @@
...
+
+from easybuild.easyblocks.a.anaconda import EB_Anaconda
+
+
+class EB_Mamba(EB_Anaconda):
+    """Support for building/installing Mamba."""
+
+    def sanity_check_step(self):
+        """
+        Custom sanity check for Anaconda and Miniconda
```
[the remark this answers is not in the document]
casparvl: You caught me, it was originally a copy-paste from anaconda.py... ;-) I'll fix it :)

easybuild/easyblocks/m/mamba.py
casparvl: True. Copied from anaconda.py, but indeed silly to keep.

boegel (approved):
lgtm

PR #2816, opened 2022-10-26
add generic JuliaPackage easyblock

lexming:
Easyblock for Julia packages:
* Installs packages from local sources
* Sources can either be a tarball or a git repo (`git_config`)
* Avoids any extra downloads (offline mode)
* Updates `JULIA_DEPOT_PATH` to make Julia aware of the installation
* defines `EXTS_FILTER_JULIA_PACKAGES` to sanity check the installation
* Works for regular easyconfigs and also for extensions
* extra option `bundle_deps` enables Julia to download extra dependencies on-the-fly (opt-in needed to use this easyblock with Julia < 1.5)

Julia needs a git repo to install any package in a regular way because it generates unique hashes for the package from it. Hence, this easyblock detects what kind of sources we have and acts accordingly:
  * git repo: install as regular package with `Pkg.add`
  * non-git tarball: install in development mode with `Pkg.develop`

### DIFF
--- a/easybuild/easyblocks/generic/juliapackage.py
+++ b/easybuild/easyblocks/generic/juliapackage.py
@@ -121,12 +121,13 @@ class JuliaPackage(ExtensionEasyBlock):
                 'Pkg.develop(PackageSpec(path="%s"))' % install_pkg_path,
                 'Pkg.build("%s")' % self.name,
             ])
+
         julia_pkg_cmd = ';'.join(julia_pkg_cmd)
-
-        install_cmd = "julia -e '%s'" % julia_pkg_cmd
-
-        cmd = ' '.join([self.cfg['preinstallopts'], install_cmd, self.cfg['installopts']])
-
+        cmd = ' '.join([
+            self.cfg['preinstallopts'],
+            "julia -e '%s'" % julia_pkg_cmd,
+            self.cfg['installopts'],
+        ])
         (out, _) = run_cmd(cmd, log_all=True, simple=False)
 
         return out
--- a/easybuild/easyblocks/generic/juliapackage.py
+++ b/easybuild/easyblocks/generic/juliapackage.py
@@ -92,8 +92,8 @@ class JuliaPackage(ExtensionEasyBlock):
         """No separate (standard) test procedure for JuliaPackage."""
         pass
 
-    def install_julia_pkg(self):
-        """Install procedure for Julia packages with Pkg"""
+    def install_step(self):
+        """Install Julia package with Pkg"""
 
         # prepend installation directory to Julia DEPOT_PATH
         # see https://docs.julialang.org/en/v1/manual/environment-variables/#JULIA_DEPOT_PATH
@@ -131,20 +131,16 @@ class JuliaPackage(ExtensionEasyBlock):
 
         return out
 
-    def install_step(self):
-        """Install Julia package"""
-        self.install_julia_pkg()
-
     def run(self):
         """Install Julia package as an extension."""
 
         if not self.src:
-            errmsg = "No source found for Julia package %s, require    d for installation. (src: %s)"
+            errmsg = "No source found for Julia package %s, required for installation. (src: %s)"
             raise EasyBuildError(errmsg, self.name, self.src)
         ExtensionEasyBlock.run(self, unpack_src=True)
 
         self.set_pkg_offline()
-        self.install_julia_pkg()
+        self.install_step()
 
     def sanity_check_step(self, *args, **kwargs):
         """Custom sanity check for JuliaPackage"""
--- a/easybuild/easyblocks/generic/juliapackage.py
+++ b/easybuild/easyblocks/generic/juliapackage.py
@@ -56,10 +56,6 @@ class JuliaPackage(ExtensionEasyBlock):
         })
         return extra_vars
 
-    def __init__(self, *args, **kwargs):
-        """JuliaPackage easyblock constructor."""
-        super(JuliaPackage, self).__init__(*args, **kwargs)
-
     def set_pkg_offline(self):
         """Enable offline mode of Julia Pkg"""
         if get_software_root('Julia') is None:
--- a/easybuild/easyblocks/generic/juliapackage.py
+++ b/easybuild/easyblocks/generic/juliapackage.py
@@ -1,0 +1,170 @@
+##
+# Copyright 2022-2022 Vrije Universiteit Brussel
+#
+# This file is part of EasyBuild,
+# originally created by the HPC team of Ghent University (http://ugent.be/hpc/en),
+# with support of Ghent University (http://ugent.be/hpc),
+# the Flemish Supercomputer Centre (VSC) (https://www.vscentrum.be),
+# Flemish Research Foundation (FWO) (http://www.fwo.be/en)
+# and the Department of Economy, Science and Innovation (EWI) (http://www.ewi-vlaanderen.be/en).
+#
+# https://github.com/easybuilders/easybuild
+#
+# EasyBuild is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation v2.
+#
+# EasyBuild is distributed in the hope that it will be useful,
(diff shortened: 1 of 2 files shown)


easybuild/easyblocks/generic/juliapackage.py
```diff
@@ -0,0 +1,170 @@
...
+        extra_vars.update({
+            'bundle_deps': [
+                False, "Let Julia download and bundle all needed dependencies for this installation", CUSTOM
+            ],
+        })
+        return extra_vars
+
+    def __init__(self, *args, **kwargs):
+        """JuliaPackage easyblock constructor."""
+        super(JuliaPackage, self).__init__(*args, **kwargs)
```
boegel, line 61: @lexming This isn't doing anything, so the `__init__` block can just be dropped, won't make a difference

easybuild/easyblocks/generic/juliapackage.py
```diff
@@ -0,0 +1,170 @@
...
+
+class JuliaPackage(ExtensionEasyBlock):
+    """Builds and installs Julia Packages."""
+
+    @staticmethod
+    def extra_options(extra_vars=None):
+        """Extra easyconfig parameters specific to JuliaPackage."""
+        extra_vars = ExtensionEasyBlock.extra_options(extra_vars=extra_vars)
+        extra_vars.update({
+            'bundle_deps': [
```
boegel, line 53: Maybe `download_pkg_deps` is a better name?

easybuild/easyblocks/generic/juliapackage.py
```diff
@@ -0,0 +1,170 @@
...
+
+        cmd = ' '.join([self.cfg['preinstallopts'], install_cmd, self.cfg['installopts']])
+
+        (out, _) = run_cmd(cmd, log_all=True, simple=False)
+
+        return out
+
+    def install_step(self):
+        """Install Julia package"""
+        self.install_julia_pkg()
```
boegel, line 140: Why not move the code that's now in `install_julia_pkg` directly into `install_step`, and then call `install_step` in `run`?

easybuild/easyblocks/generic/juliapackage.py
```diff
@@ -0,0 +1,170 @@
...
+
+    def run(self):
+        """Install Julia package as an extension."""
+
+        if not self.src:
+            errmsg = "No source found for Julia package %s, require    d for installation. (src: %s)"
```
boegel, line 146: `require    d` => `required`

easybuild/easyblocks/generic/juliapackage.py
```diff
@@ -0,0 +1,170 @@
...
+                'Pkg.develop(PackageSpec(path="%s"))' % install_pkg_path,
+                'Pkg.build("%s")' % self.name,
+            ])
+        julia_pkg_cmd = ';'.join(julia_pkg_cmd)
+
+        install_cmd = "julia -e '%s'" % julia_pkg_cmd
+
+        cmd = ' '.join([self.cfg['preinstallopts'], install_cmd, self.cfg['installopts']])
+
+        (out, _) = run_cmd(cmd, log_all=True, simple=False)
```
boegel, line 134: Small code style suggestions:

```python
            ])

        julia_pkg_cmd = ';'.join(julia_pkg_cmd)
        cmd = ' '.join([
            self.cfg['preinstallopts'],
            "julia -e '%s'" % julia_pkg_cmd,
            self.cfg['installopts'],
        ])
        (out, _) = run_cmd(cmd, log_all=True, simple=False)

File: easybuild/easyblocks/generic/juliapackage.py [hunk 6]
```diff
@@ -0,0 +1,170 @@
...
+        # command sequence for Julia.Pkg
+        julia_pkg_cmd = ['using Pkg']
+        if os.path.isdir(os.path.join(self.start_dir, '.git')):
+            # sources from git repos can be installed as any remote package
+            self.log.debug('Installing Julia package in normal mode (Pkg.add)')
+
+            julia_pkg_cmd.extend([
+                # install package from local path preserving existing dependencies
+                'Pkg.add(url="%s"; preserve=Pkg.PRESERVE_ALL)' % self.start_dir,
+            ])
```
boegel, line 115: use `.append`?

```suggestion
            # install package from local path preserving existing dependencies
            julia_pkg_cmd.append('Pkg.add(url="%s"; preserve=Pkg.PRESERVE_ALL)' % self.start_dir)

File: easybuild/easyblocks/generic/juliapackage.py [hunk 7]
```diff
@@ -0,0 +1,170 @@
...
+    def test_step(self):
+        """No separate (standard) test procedure for JuliaPackage."""
+        pass
+
+    def install_julia_pkg(self):
+        """Install procedure for Julia packages with Pkg"""
+
+        # prepend installation directory to Julia DEPOT_PATH
+        # see https://docs.julialang.org/en/v1/manual/environment-variables/#JULIA_DEPOT_PATH
+        env.setvar('JULIA_DEPOT_PATH', ':'.join([self.installdir, os.getenv('JULIA_DEPOT_PATH', '')]))
```
boegel, line 104: Maybe you want to avoid a trailing `:` here:

```python
depot_path = os.getenv('JULIA_DEPOT_PATH')
if depot_path is not None:
    depot_path = ':'.join([self.installdir, depot_path])
env.setvar('JULIA_DEPOT_PATH', depot_path)

boegel (changes requested):

easybuild/easyblocks/generic/juliapackage.py
lexming: Fixed in https://github.com/easybuilders/easybuild-easyblocks/pull/2816/commits/2843f9a298e31d6bd240a635902cc29e96767741

easybuild/easyblocks/generic/juliapackage.py
lexming: sure, done in https://github.com/easybuilders/easybuild-easyblocks/pull/2816/commits/0d4a9936d426db511e7b00db63d02d71031fdb60

easybuild/easyblocks/generic/juliapackage.py
lexming: good point, whatever is done in `install_step` will certainly have to be done in `run` as well. Fixed in https://github.com/easybuilders/easybuild-easyblocks/pull/2816/commits/a8e23e5b8a9dac5de6a603e829229e54841e1c8c

easybuild/easyblocks/generic/juliapackage.py
lexming: Fixed in https://github.com/easybuilders/easybuild-easyblocks/pull/2816/commits/a8e23e5b8a9dac5de6a603e829229e54841e1c8c

easybuild/easyblocks/generic/juliapackage.py
lexming: Done in https://github.com/easybuilders/easybuild-easyblocks/pull/2816/commits/d12181b57c26deffca3385669989f496ce6227d3

easybuild/easyblocks/generic/juliapackage.py
lexming: I prefer to keep the `extend` here. It matches in form with the other branch of the `if` and it is easier to expand in the future if we ever need to execute more command before/after the `Pkg.add`.

easybuild/easyblocks/generic/juliapackage.py
lexming: good point, fixed in https://github.com/easybuilders/easybuild-easyblocks/pull/2816/commits/744a628a8928de76ffbf4bb2e29d5aadd6c2c5e9

lexming:
Requested new tests in the linked easyconfigs.

easybuild/easyblocks/generic/juliapackage.py
boegel: ok, fine :)

boegel (approved):
lgtm

PR #2843, opened 2022-12-08
Rosetta database+compile with serialization

alikerr:
- Fix for database in the version 3.13. 
- Adding the option to compile with mpi,seriallization using a flag in the easyconfig: serialization=True or False

### DIFF
--- a/easybuild/easyblocks/r/rosetta.py
+++ b/easybuild/easyblocks/r/rosetta.py
@@ -1,5 +1,5 @@
 ##
-# Copyright 2009-2022 Ghent University
+# Copyright 2009-2021 Ghent University
 #
 # This file is part of EasyBuild,
 # originally created by the HPC team of Ghent University (http://ugent.be/hpc/en),
@@ -30,6 +30,7 @@ EasyBuild support for building and installing Rosetta, implemented as an easyblo
 @author: Kenneth Hoste (Ghent University)
 @author: Pieter De Baets (Ghent University)
 @author: Jens Timmerman (Ghent University)
+@author: Ali Kerrache (University of Manitoba)
 """
 import fileinput
 import os
@@ -37,6 +38,7 @@ import re
 import shutil
 import sys
 import easybuild.tools.toolchain as toolchain
+
 from easybuild.easyblocks.icc import get_icc_version
 from easybuild.framework.easyblock import EasyBlock
 from easybuild.tools.build_log import EasyBuildError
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/r/rosetta.py
```diff
@@ -249,8 +247,6 @@ def extract_and_copy(dirname_tmpl, optional=False, symlinks=False):
         # (extract and) copy database, docs, demos (incl tutorials) and biotools (if it's there)
         if os.path.exists(os.path.join(self.cfg['start_dir'], 'main', 'database')):
             extract_and_copy(os.path.join('main', 'database') + '%s')
-        else:
-            extract_and_copy('rosetta_database%s')
```
branfosj, line 253: I expect this is needed for older versions. If this is no longer useful for 3.13, then we should change this to a version check, so:
```
        elif LooseVersion(self.version) < LooseVersion("3.13"):
            extract_and_copy('rosetta_database%s')

**Review by branfosj — CHANGES_REQUESTED** [review]

File: easybuild/easyblocks/r/rosetta.py (hunk 1 above)
Comment (alikerr) [reply easybuild/easyblocks/r/rosetta.py]: I think I uploaded a wrong version of the file. I have just updated the right file on my git repo [73dc29d](https://github.com/alikerr/easybuild-easyblocks/commit/73dc29d19e29cb8c94f00cddded0ecc9c44c1d5b)

File: easybuild/easyblocks/r/rosetta.py [hunk 2]
```diff
@@ -121,20 +134,16 @@ def configure_step(self):
...
 
         # create user.settings file
-        paths = os.getenv('PATH')
-        paths = paths.split(':') if paths else []
-        ld_library_paths = os.getenv('LD_LIBRARY_PATH')
-        ld_library_paths = ld_library_paths.split(':') if ld_library_paths else []
-        cpaths = os.getenv('CPATH')
-        cpaths = cpaths.split(':') if cpaths else []
+        paths = os.getenv('PATH').split(':')
+        #ld_library_paths = os.getenv('LD_LIBRARY_PATH').split(':')
```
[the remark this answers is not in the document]
alikerr: This one was fixed in the previous PR #2744

easybuild/easyblocks/r/rosetta.py
```diff
@@ -145,18 +154,7 @@ def configure_step(self):
             "           'defines': %s," % str(defines),
             "       },",
             "       'overrides': {",
-            "           'cc': '%s'," % os.getenv('CC'),
-            "           'cxx': '%s'," % os.getenv('CXX'),
-            "           'ENV': {",
-            "               'INTEL_LICENSE_FILE': '%s'," % os.getenv('INTEL_LICENSE_FILE'),  # Intel license file
-            "               'PATH': %s," % str(paths),
-            "               'LD_LIBRARY_PATH': %s," % str(ld_library_paths),
```
boegel, line 153: @alikerr These look like unintended changes to the exiting easyblock?

boegel (changes requested):

PR #2834, opened 2022-11-25
Fix counting of failures in PyTorch tests

Flamefire:
(created using `eb --new-pr`)

The RegExp handling was wrong, e.g. for an output of:

```
SKIPPED [2] distributions/test_constraints.py:83: `biject_to` not implemented.
FAILED distributions/test_constraints.py::test_constraint[True-constraint_fn5-False-value5]
FAILED distributions/test_constraints.py::test_constraint[True-constraint_fn7-True-value7]
============= 2 failed, 128 passed, 2 skipped, 2 warnings in 8.66s =============
distributions/test_constraints failed!
```

the `summary_matches_pattern2` will yield `"2 failed, 128 passed, 2 skipped, 2 warnings in 8.66s"` which does not match `r"^.*[^0-9](?Pfailures[0-9]+) failed.*$"` because there is nothing in front of the number.

This PR does a bit of refactoring to avoid this: Instead of creating `summary_matches_pattern2` out of a RegEx match and using it in 2 places way below where the exact RegEx might be already forgotten I combined the counting of errors/failures with the summary-message generation and moved the RegEx right on top of the loop handling it and added comments with examples.

Also the anchors and named group is not required:  `r"^.*[^0-9](?Pfailures[0-9]+) failed.*$"` ->  `r"([0-9]+) failed"` which is much cleaner to read.

Finally my [commit](https://github.com/easybuilders/easybuild-easyblocks/commit/30b4e02da937e319e4e6fccf4028760f2ec75b09) to also count tests terminated by a signal was lost so I readded it and reformatted the RegEx assignment to make it easier to read

And as a drive-by fix I added a docstring to `get_count_for_pattern` and made it return `0` instead of `None` for the case where the warning is shown.

PR #2859, opened 2023-01-12
fix finding of failed tests in output of PyTorch test step

boegel:
Test reports in https://github.com/easybuilders/easybuild-easyconfigs/pull/16286 and https://github.com/easybuilders/easybuild-easyconfigs/pull/16385 showed that installation was failing because no details were found on `distributed/test_c10d_gloo`

That's fixed in this PR, and I've also fleshed out the code that extracts the details on the failing test suites/groups into a dedicated function, so we can test it in CI

@Flamefire Please take a look?

### DIFF
--- a/easybuild/easyblocks/p/pytorch.py
+++ b/easybuild/easyblocks/p/pytorch.py
@@ -79,13 +79,12 @@ def extract_failed_tests_info(tests_out):
              r"(?:^(?:(?!failed!).)*$\n)*"
              r"(?P<failed_test_suite_name>.*) failed!(?: Received signal: \w+)?\s*$")
 
-    for m in re.finditer(regex, tests_out, re.M):
+    for match in re.finditer(regex, tests_out, re.M):
         # E.g. 'failures=3, errors=10, skipped=190, expected failures=6'
-        failure_summary = m.group('failure_summary')
-        total, test_suite = m.group('test_cnt', 'failed_test_suite_name')
-        failure_report += "{test_suite} ({total} total tests, {failure_summary})\n".format(
-                test_suite=test_suite, total=total, failure_summary=failure_summary
-            )
+        info = match.groupdict()
+        failure_summary =  info['failure_summary']
+        test_cnt, test_suite = info['test_cnt'], info['failed_test_suite_name']
+        failure_report += test_suite + ' (' + test_cnt + " total tests, " + failure_summary + ')\n'
         failure_cnt += get_count_for_pattern(r"(?<!expected )failures=([0-9]+)", failure_summary)
         error_cnt += get_count_for_pattern(r"errors=([0-9]+)", failure_summary)
         failed_test_suites.append(test_suite)
@@ -94,13 +93,11 @@ def extract_failed_tests_info(tests_out):
     # ===================== 2 failed, 128 passed, 2 skipped, 2 warnings in 3.43s =====================
     regex = r"^=+ (?P<failure_summary>.*) in [0-9]+\.*[0-9]*[a-zA-Z]* =+$\n(?P<failed_test_suite_name>.*) failed!$"
 
-    for m in re.finditer(regex, tests_out, re.M):
+    for match in re.finditer(regex, tests_out, re.M):
         # E.g. '2 failed, 128 passed, 2 skipped, 2 warnings'
-        failure_summary = m.group('failure_summary')
-        test_suite = m.group('failed_test_suite_name')
-        failure_report += "{test_suite} ({failure_summary})\n".format(
-                test_suite=test_suite, failure_summary=failure_summary
-            )
+        info = match.groupdict()
+        failure_summary, test_suite = info['failure_summary'], info['failed_test_suite_name']
+        failure_report += test_suite + ' ' + failure_summary + '\n'
         failure_cnt += get_count_for_pattern(r"([0-9]+) failed", failure_summary)
         error_cnt += get_count_for_pattern(r"([0-9]+) error", failure_summary)
         failed_test_suites.append(test_suite)
@@ -112,11 +109,10 @@ def extract_failed_tests_info(tests_out):
     regex = (r"^AssertionError: (?P<failed_test_cnt>[0-9]+) unit test\(s\) failed:$\n"
              r"(?:^(?:(?!failed!).)*$\n)*"
              r"(?P<failed_test_group_name>.*) failed!$")
-    for m in re.finditer(regex, tests_out, re.M):
-        test_group = m.group('failed_test_group_name')
-        failed_test_cnt = m.group('failed_test_cnt')
-        failure_report += "{test_group} ({failed_test_cnt} failed tests)\n".format(
-                test_group=test_group, failed_test_cnt=failed_test_cnt)
+    for match in re.finditer(regex, tests_out, re.M):
+        info = match.groupdict()
+        test_group, failed_test_cnt = info['failed_test_group_name'], info['failed_test_cnt']
+        failure_report += test_group + ' (' + failed_test_cnt + " failed tests)\n"
         failure_cnt += int(failed_test_cnt)
         failed_test_suites.append(test_group)
 
--- a/easybuild/easyblocks/p/pytorch.py
+++ b/easybuild/easyblocks/p/pytorch.py
@@ -69,7 +69,7 @@ def extract_failed_tests_info(tests_out):
     error_cnt = 0
     failed_test_suites = []
 
-    # Grep for patterns like:
+    # Look for patterns like:
     # Ran 219 tests in 67.325s
     #
     # FAILED (errors=10, skipped=190, expected failures=6)
@@ -89,7 +89,7 @@ def extract_failed_tests_info(tests_out):
         error_cnt += get_count_for_pattern(r"errors=([0-9]+)", failure_summary)
         failed_test_suites.append(test_suite)
 
-    # Grep for patterns like:
+    # Look for patterns like:
     # ===================== 2 failed, 128 passed, 2 skipped, 2 warnings in 3.43s =====================
     regex = r"^=+ (?P<failure_summary>.*) in [0-9]+\.*[0-9]*[a-zA-Z]* =+$\n(?P<failed_test_suite_name>.*) failed!$"
 
@@ -102,7 +102,7 @@ def extract_failed_tests_info(tests_out):
         error_cnt += get_count_for_pattern(r"([0-9]+) error", failure_summary)
         failed_test_suites.append(test_suite)
 
-    # Grep for patterns like:
+    # Look for patterns like:
     # AssertionError: 4 unit test(s) failed:
     # ...
     # distributed/test_c10d_gloo failed!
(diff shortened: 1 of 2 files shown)


easybuild/easyblocks/p/pytorch.py
```diff
@@ -42,6 +42,98 @@
...
+        failed_test_suites.append(test_group)
+
+    # Look for patterns like:
+    # Running test_jit_cuda_fuser ... [2023-01-12 04:04:08.949222]
+    # ...
+    # AttributeError: 'NoneType' object has no attribute 'split'
+    # test_jit_cuda_fuser failed!
+    regex = (r"^Running (?P<test_name>test_[^\s]+) .*\n"
+             r"(?:^(?:(?!failed!).)*$\n)*"
+             r"\1 failed!$")
```
boegel, line 126: Hmm, I need to make this regex a bit stricter, since it partially overlaps with the first regex used in this function.

Need to make sure there's no `Ran [0-9]+ tests` line...

test/easyblocks/easyblock_specific.py
```diff
@@ -265,6 +332,30 @@ def test_det_cmake_version(self):
...
+
+        expected_failure_report = '\n'.join([
+            "distributed/fsdp/test_fsdp_input (2 total tests, failures=2)",
+            "distributions/test_distributions (216 total tests, errors=4)",
+            "test_autograd (464 total tests, failures=1, skipped=52, expected failures=1)",
+            "distributed/test_c10d_gloo (4 failed tests)",
+            "test_jit_cuda_fuser (unknown failed test count)",
+        ])
+        self.assertEqual(res[0].strip(), expected_failure_report)
+        # test failures
```
Flamefire, line 350: When factoring this out and especially for testing I'd strongly suggest to use a dict or namedtuple as the return type. This tuple of "something" requires those comments and might be wrong at some point

easybuild/easyblocks/p/pytorch.py
```diff
@@ -42,6 +42,98 @@
...
+        failed_test_suites.append(test_suite)
+
+    # Look for patterns like:
+    # ===================== 2 failed, 128 passed, 2 skipped, 2 warnings in 3.43s =====================
+    regex = r"^=+ (?P<failure_summary>.*) in [0-9]+\.*[0-9]*[a-zA-Z]* =+$\n(?P<failed_test_suite_name>.*) failed!$"
+
+    for match in re.finditer(regex, tests_out, re.M):
+        # E.g. '2 failed, 128 passed, 2 skipped, 2 warnings'
+        info = match.groupdict()
+        failure_summary, test_suite = info['failure_summary'], info['failed_test_suite_name']
```
Flamefire, line 99: No need to create a dict when you use it only once:
```suggestion
        failure_summary, test_suite = match.group('failure_summary', 'failed_test_suite_name')
```

easybuild/easyblocks/p/pytorch.py
```diff
@@ -42,6 +42,98 @@
...
+    regex = (r"^Ran (?P<test_cnt>[0-9]+) tests.*$\n\n"
+             r"FAILED \((?P<failure_summary>.*)\)$\n"
+             r"(?:^(?:(?!failed!).)*$\n)*"
+             r"(?P<failed_test_suite_name>.*) failed!(?: Received signal: \w+)?\s*$")
+
+    for match in re.finditer(regex, tests_out, re.M):
+        # E.g. 'failures=3, errors=10, skipped=190, expected failures=6'
+        info = match.groupdict()
+        failure_summary = info['failure_summary']
+        test_cnt, test_suite = info['test_cnt'], info['failed_test_suite_name']
```
Flamefire, line 86: What was wrong with the original, shorter:
```suggestion
        failure_summary = m.group('failure_summary')
        total, test_suite = m.group('test_cnt', 'failed_test_suite_name')
```

easybuild/easyblocks/p/pytorch.py
```diff
@@ -42,6 +42,98 @@
...
+
+    # Look for patterns like:
+    # AssertionError: 4 unit test(s) failed:
+    # ...
+    # distributed/test_c10d_gloo failed!
+    regex = (r"^AssertionError: (?P<failed_test_cnt>[0-9]+) unit test\(s\) failed:$\n"
+             r"(?:^(?:(?!failed!).)*$\n)*"
+             r"(?P<failed_test_group_name>.*) failed!$")
+    for match in re.finditer(regex, tests_out, re.M):
+        info = match.groupdict()
```
Flamefire, line 113: See above

easybuild/easyblocks/p/pytorch.py
Flamefire: Do we really want to go into that effort? I mean the current implementation already finds all "test_jit_cuda_fuser failed!" instances in `all_failed_test_suites` and I don't see any additional value in this RegEx which I honestly would need to consult the manual to understand the 2nd line.

I mean this is an instance where we want to make the test step fail as we cannot determine the number of failed tests

test/easyblocks/easyblock_specific.py
```diff
@@ -51,6 +52,72 @@
...
+Running test_jit_cuda_fuser ... [2023-01-12 04:04:08.949222]
+Executing ['/user/gent/400/vsc40023/eb_arcaninescratch/RHEL8/skylake-ib/software/Python/3.9.6-GCCcore-11.2.0/bin/python', 'test_jit_cuda_fuser.py', '-v'] ... [2023-01-12 04:04:08.949319]
+CUDA not available, skipping tests
+monkeytype is not installed. Skipping tests for Profile-Directed Typing
+Traceback (most recent call last):
+  File "/tmp/vsc40023/easybuild_build/PyTorch/1.11.0/foss-2021b/pytorch-v1.11.0/test/test_jit_cuda_fuser.py", line 25, in <module>
+    CUDA_MAJOR, CUDA_MINOR = (int(x) for x in torch.version.cuda.split('.'))
+AttributeError: 'NoneType' object has no attribute 'split'
+test_jit_cuda_fuser failed!
+"""  # noqa
```
Flamefire, line 129: I'd also add regression tests for the example at https://github.com/easybuilders/easybuild-easyblocks/pull/2834#issue-1464477768 and `	test_fx failed! Received signal: SIGSEGV`

easybuild/easyblocks/p/pytorch.py
boegel: fixed in b8770f77f

easybuild/easyblocks/p/pytorch.py
boegel: I mainly started changing this due to the single-letter variable, I don't have a good reason to favor `.groupdict` over using `.group` multiple times

test/easyblocks/easyblock_specific.py
boegel: Do you have a output snipped for a test that fails with `Received signal`?

boegel: Nevermind, found enough in https://gist.github.com/Flamefire/dc1403ccefdebfc3412c6fbb2d5cbabd

boegel:
@Flamefire suggested changes implemented and test enhanced, please take another look before I fire off test builds again with this?

test/easyblocks/easyblock_specific.py
Flamefire: Great. It's just missing a check for this, isn't it?

easybuild/easyblocks/p/pytorch.py
Flamefire: The latter has the advantage that getting multiple values at once is easy :)

Flamefire:
> @Flamefire suggested changes implemented and test enhanced, please take another look before I fire off test builds again with this?

LGTM except for the failing check for the signaled test. Also I'd suggest to use the named tuple members in the easyblock, not just the test.

test/easyblocks/easyblock_specific.py
boegel: fixed in 4952598

test/easyblocks/easyblock_specific.py
boegel: fixed in eaf4056

test/easyblocks/easyblock_specific.py
```diff
@@ -265,6 +357,39 @@ def test_det_cmake_version(self):
...
+        Test extract_failed_tests_info function from PyTorch easyblock.
+        """
+        res = pytorch.extract_failed_tests_info(PYTORCH_TESTS_OUTPUT)
+        self.assertEqual(len(res), 4)
+
+        expected_failure_report = '\n'.join([
+            "distributed/fsdp/test_fsdp_input (2 total tests, failures=2)",
+            "distributions/test_distributions (216 total tests, errors=4)",
+            "test_autograd (464 total tests, failures=1, skipped=52, expected failures=1)",
+            "test_fx (123 total tests, errors=2, skipped=2)",
```
Flamefire, line 371: This or the SIGSEGV test needs to be another one or `failed_test_suites` could pass when the signal was not detected as in the case of a forced termination due to a signla there won't be a test summary, will there?

easybuild/easyblocks/p/pytorch.py
```diff
@@ -268,65 +357,11 @@ def test_step(self):
...
-            failure_summary = m.group('failure_summary')
-            test_suite = m.group('failed_test_suite_name')
-            failure_report += "{test_suite} ({failure_summary})\n".format(
-                    test_suite=test_suite, failure_summary=failure_summary
-                )
-            failure_cnt += get_count_for_pattern(r"([0-9]+) failed", failure_summary)
-            error_cnt += get_count_for_pattern(r"([0-9]+) error", failure_summary)
-            failed_test_suites.append(test_suite)
+        failed_tests_info = extract_failed_tests_info(tests_out)
+        failure_report = failed_tests_info.failure_report
```
Flamefire, line 361: Nit: To avoid long distracting lines like `sorted(set(failed_tests_info.failed_test_suites))` (2 times "failed") maybe either unwrap the named tuple here: `failure_report, failed_test_suites = failed_tests_info.failure_report, failed_tests_info.failed_test_suites` etc. or rename to `test_info` as `test_info.failed_test_suites` is easier to read.
And maybe the `test_cnt` extraction can be done in there too which would fit the name even better (then drop the "failed" from the function name).

test/easyblocks/easyblock_specific.py
Flamefire: Looks like we have a bug here. For the output of
```
Running distributions/rpc/test_tensorpipe_agent ... [2023-01-12 09:06:37.093571]
...
Ran 123 tests in 7.549s
FAILED (errors=2, skipped=2)
...
test_fx failed! Received signal: SIGSEGV
```
it must NOT match the RegExp. The original, correct output is:
```
FAILED (errors=2, skipped=2)
distributed/rpc/test_tensorpipe_agent failed!
```

So the RegExp must be made to only add the counting of "FAILED" if the next line is the failed test, see https://gist.github.com/Flamefire/dc1403ccefdebfc3412c6fbb2d5cbabd#file-pytorch-1-9-0-foss-2020b_partial-log-L480

akesandgren:
@boegel @Flamefire Status on this one?

Flamefire:
This conflicts with https://github.com/easybuilders/easybuild-easyblocks/pull/3085 after that got merged possibly making all changes to the EasyBlock file unnecessary. Not fully sure though if there were any changes to the patterns that are not included yet.

I also have #3255 which further improves the function by pulling the last bits of the parsing out of the easyblock.

PR #2869, opened 2023-01-26
fix JULIA_DEPOT_PATH in installation of multiple JuliaPackage extensions

lexming:
Installation of Julia packages is broken in two specific cases:
* whenever `JULIA_DEPOT_PATH` is unset, as `env.setvar` tries to access a `depot_path` that is `None`
* bundles of extensions prepend multiple times `installdir` to `JULIA_DEPOT_PATH`

### DIFF
--- a/easybuild/easyblocks/generic/juliapackage.py
+++ b/easybuild/easyblocks/generic/juliapackage.py
@@ -96,11 +96,12 @@ class JuliaPackage(ExtensionEasyBlock):
         """Install Julia package with Pkg"""
 
         # prepend installation directory to Julia DEPOT_PATH
+        # extensions in a bundle can share their DEPOT_PATH
         # see https://docs.julialang.org/en/v1/manual/environment-variables/#JULIA_DEPOT_PATH
-        depot_path = os.getenv('JULIA_DEPOT_PATH')
-        if depot_path is not None:
-            depot_path = ':'.join([self.installdir, depot_path])
-        env.setvar('JULIA_DEPOT_PATH', depot_path)
+        depot_path = os.getenv('JULIA_DEPOT_PATH', '')
+        if self.installdir not in depot_path:
+            depot_path = ':'.join([depot for depot in (self.installdir, depot_path) if depot])
+            env.setvar('JULIA_DEPOT_PATH', depot_path)
 
         # command sequence for Julia.Pkg
         julia_pkg_cmd = ['using Pkg']


easybuild/easyblocks/generic/juliapackage.py
```diff
@@ -96,11 +96,12 @@ def install_step(self):
         """Install Julia package with Pkg"""
 
         # prepend installation directory to Julia DEPOT_PATH
+        # extensions in a bundle can share their DEPOT_PATH
         # see https://docs.julialang.org/en/v1/manual/environment-variables/#JULIA_DEPOT_PATH
-        depot_path = os.getenv('JULIA_DEPOT_PATH')
-        if depot_path is not None:
-            depot_path = ':'.join([self.installdir, depot_path])
-        env.setvar('JULIA_DEPOT_PATH', depot_path)
+        depot_path = os.getenv('JULIA_DEPOT_PATH', '')
```
boegel, line 101: We should split on `os.pathsep` here before doing the `in` check, otherwise we may get a partial match:
```python
depot_path = os.getenv('JULIA_DEPOT_PATH', [])
if depot_path:
    depot_path = depot_path.split(os.pathsep)
if self.installdir not in depot_path:
    depot_path = os.pathsep.join([x for x in ([self.installdir] + depot_path) if x])
    env.setvar('JULIA_DEPOT_PATH', depot_path)
```

easybuild/easyblocks/generic/juliapackage.py
```diff
@@ -96,11 +96,12 @@ def install_step(self):
...
+        if self.installdir not in depot_path:
+            depot_path = ':'.join([depot for depot in (self.installdir, depot_path) if depot])
```
boegel, line 103: please use `os.pathsep` instead of `:`

boegel (changes requested):

easybuild/easyblocks/generic/juliapackage.py
lexming: very good point, fixed in https://github.com/easybuilders/easybuild-easyblocks/pull/2869/commits/f2b4131c0821019b47d96568fc604ee0bcf20de0

easybuild/easyblocks/generic/juliapackage.py
lexming: fixed in https://github.com/easybuilders/easybuild-easyblocks/pull/2869/commits/f2b4131c0821019b47d96568fc604ee0bcf20de0

easybuild/easyblocks/generic/juliapackage.py
```diff
@@ -96,11 +96,14 @@ def install_step(self):
...
-        depot_path = os.getenv('JULIA_DEPOT_PATH')
-        if depot_path is not None:
-            depot_path = ':'.join([self.installdir, depot_path])
-        env.setvar('JULIA_DEPOT_PATH', depot_path)
+        depot_path = os.getenv('JULIA_DEPOT_PATH', '')
+        if depot_path:
+            depot_path = depot_path.split(os.pathsep)
+        if self.installdir not in depot_path:
+            depot_path = os.pathsep.join([depot for depot in (self.installdir, depot_path) if depot])
+            env.setvar('JULIA_DEPOT_PATH', depot_path)
```
smoors, line 106: this will fail if depot_path exits, because then it's converted from a string to a list.
```suggestion
        depot_path = os.getenv('JULIA_DEPOT_PATH', [])
        if depot_path:
            depot_path = depot_path.split(os.pathsep)
        if self.installdir not in depot_path:
            depot_path = os.pathsep.join([depot for depot in [self.installdir] + depot_path if depot])
            env.setvar('JULIA_DEPOT_PATH', depot_path)
```

lexming:
Test report by @lexming
 * **SUCCESS** _IJulia-1.23.3-Julia-1.6.7.eb_ 
 * **SUCCESS** _IJulia-1.24.0-Julia-1.8.5.eb_ 

node301.hydra.os - Linux CentOS Linux 7.9.2009, x86_64, Intel(R) Xeon(R) Gold 6148 CPU @ 2.40GHz (skylake_avx512), Python 3.6.8
See https://gist.github.com/94e984c578afe1bc0d3ebde4eb6dec15 for a full test report.

smoors (approved):
lgtm

PR #2875, opened 2023-02-02
ESMF changed the structure and naming of their python bindings in ver…

mboisson:
…sion 8.4

### DIFF
--- a/easybuild/easyblocks/e/esmf.py
+++ b/easybuild/easyblocks/e/esmf.py
@@ -136,22 +136,36 @@ class EB_ESMF(ConfigureMake):
 
         python = get_software_version('Python')
         if python:
-            # then, install the python bindings
-            py_subdir = os.path.join(self.builddir, 'esmf-ESMF_%s' % '_'.join(self.version.split('.')),
-                                     'src', 'addon', 'ESMPy')
+            if LooseVersion(self.version) < LooseVersion('8.4.0'):
+                # then, install the python bindings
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/e/esmf.py
```diff
@@ -136,22 +136,36 @@ def install_step(self):
...
-            cmd = "python setup.py build --ESMFMKFILE=%s/lib/esmf.mk " % self.installdir
-            cmd += " && python setup.py install --prefix=%s" % self.installdir
-            run_cmd(cmd, log_all=True, simple=True, log_ok=True)
+            if LooseVersion(self.version) < LooseVersion('8.4.0'):
+                cmd = "python setup.py build --ESMFMKFILE=%s/lib/esmf.mk " % self.installdir
+                cmd += " && python setup.py install --prefix=%s" % self.installdir
+                run_cmd(cmd, log_all=True, simple=True, log_ok=True)
+            else:
+                cmd = "ESMFMKFILE=%s/lib/esmf.mk " % self.installdir
+                cmd += "python -m pip install . --no-deps --prefix=%s" % self.installdir
```
smoors, line 158: EB also adds pip options `--ignore-installed  --no-index  --no-build-isolation`, shouldn't those be added here too?

mboisson: not tested

boegel: I agree with @smoors that we should add those additional options for `pip install`, they're there for a reason (and it's better to err on the side of caution with installing Python packages).

It would be nicer if we could query `PythonPackage` for which options to pass to `pip install` though...

PR #2877, opened 2023-02-06
fix error when failing PIP version check during PythonPackage sanity check

Flamefire:
The arguments must be passed individually not as a tuple, else you get:
>   File "/easybuild-easyblocks/easybuild/easyblocks/generic/pythonpackage.py", line 921, in sanity_check_step
>     pip_version))
>   File "/easybuild-framework/easybuild/tools/build_log.py", line 81, in __init__
>     msg = msg % args
> TypeError: not enough arguments for format string

@boegel This should go into next release.


boegel (approved):
lgtm

PR #2889, opened 2023-02-14
Use older `ncgen -H` for older netCDF

### LINKED ISSUE #2888: New sanity check commands for netCDF break build of older netCDF
This one took me a while to chase down. It might be a wontfix due to old software, but maybe this issue can at least serve as an FFR.

N.b. the easybuild and netCDF tag numbers are very similar here.

[In easybuild v4.6.2 some `nc-config` and `ncgen` sanity check commands were added to the netCDF easyblock](https://github.com/easybuilders/easybuild-easyblocks/commit/d8aa9420be572ab4df2c5993c5a3cdf370623404).

The command `ncgen -h`, in particular, should show the help text for `ncgen`. However, up until netCDF v4.6.1, `ncgen -h` meant running `ncgen` in 'header-only" mode, while `-H` was used for help. These flags were switched in [this commit](https://github.com/Unidata/netcdf-c/commit/2ea1cf5f1bc2a7352e3f66721f5181e26e556011#diff-6f23c24b125838dbee16fd3fd9edf84acc7a0492bd223c5ed03e9095cd50b15e) which first appeared in netCDF v4.6.2.

The problem is that when `ncgen` is run without any other args, or similarly, in "header-only" mode, it waits for stdin until an EOF. This causes the `ncgen` sanity check in the new easyblock to fail.

I think the fix would be to check `ncgen -H` instead of `ncgen -h` for netCDF older than v4.6.2.

randallpittman:
This seems like a simple enough fix?

Fixes #2888.

### DIFF
--- a/easybuild/easyblocks/n/netcdf.py
+++ b/easybuild/easyblocks/n/netcdf.py
@@ -143,7 +143,7 @@ class EB_netCDF(CMakeMake):
 
         custom_commands = [
             "nc-config --help",
-            "ncgen -h",
+            "ncgen -h" if LooseVersion(self.version) > LooseVersion("4.6.1") else "ncgen -H",
         ]
 
         super(EB_netCDF, self).sanity_check_step(custom_commands=custom_commands, custom_paths=custom_paths)


akesandgren (approved):

akesandgren:
Please submit test reports of netCDF for both >= 4.6.1 and >= 4.6.2 using --include-easyblocks-from-pr to show that is works

PR #2891, opened 2023-02-21
fix error log in PythonPackage for old pip versions

lexming:
There is a syntax error in this error message.


lexming:
I'm too slow, this was recently fixed in #2877

PR #2906, opened 2023-03-16
inform users that gpu package (instead of kokkos) is used for LAMMPS

xwuupb:


ocaisa (approved):
I made some tiny changes to make both the GPU package and Kokkos package branches consistent

xwuupb:
thanks @ocaisa !
I will also PR the LAMMPS easyconfig files for the GPU package later, because currently only the KOKKOS-* variants are available for LAMMPS easyconfig files and my preliminary benchmarks show that the GPU-CUDA build provides much higher performance than the KOKKOS-* variants.

ocaisa:
We did a good few studies when preparing http://www.hpc-carpentry.org/tuning_lammps/ and found the only major differences to be when the GPU package uses single precision (IIRC). I agree the other variants should be added, as which is better I believe depends on your use case (and Kokkos should improve the more recent the LAMMPS version).

Thanks for the PR, and for agreeing to create the easyconfig PRs!

PR #2920, opened 2023-04-18
Add external pio option for ESMF  versions greater than 8.4.0

mvdebolskiy:


mvdebolskiy:
There have been added an option to have an external pio library with ESMF since 8.4.0, this is can be useful for versions of CESM that have pio code that relies on newer ParallelIO versions than the one that is being built within ESMF internally.

Micket:
I'm afraid you must have made some mistakes when opening this PR.
First, they would need to be (based on and) against develop branch, and this now includes a bunch of unrelated things since it's trying to merge some of the stuff from develop into main.

I would recommend just using the EB github integration hook `eb --new-pr esfm.py --pr-commit-msg="Add external pio option for ESMF versions greater than 8.4.0"` which will just do the-right-thing

mvdebolskiy:
@Micket  Will do! Do I close this one?

Micket:
Done!

PR #2921, opened 2023-04-21
update GCC easyblock to ensure that `--sysroot` is passed to linker (but only when it needs to be)

trz42:
This removes an adjustment to `gcc/gcc.c` which avoided that `--sysroot` is passed to the linker. Since a recent change to how Gentoo supports sysroot (see https://github.com/gentoo/gentoo/pull/28851) we need to change the easyblock for gcc.

A bit of testing revealed that only the removed line(s) are not needed any longer.

So far it has been tested for an EESSI-like environment with a recent compatibility layer version (Gentoo commit from April 17) with:
- ~GCC/9.3.0~ not tested, but might be ignore for EESSI at least
- [x] GCC/10.3.0 on `x86_64/{generic,amd/zen2,intel/broadwell,intel/cascadelake,intel/skylake_avx512` and `aarch64/generic`
- [x] GCC/11.3.0 on `x86_64/{generic,amd/zen2,intel/broadwell,intel/cascadelake,intel/skylake_avx512` and `aarch64/generic`
- [x] GCC/12.2.0 on `x86_64/{generic,amd/zen2,intel/broadwell,intel/cascadelake,intel/skylake_avx512` and `aarch64/generic`

### DIFF
--- a/easybuild/easyblocks/g/gcc.py
+++ b/easybuild/easyblocks/g/gcc.py
@@ -357,9 +357,6 @@ class EB_GCC(ConfigureMake):
             # (see https://gcc.gnu.org/install/configure.html)
             self.cfg.update('configopts', '--with-sysroot=%s' % sysroot)
 
-            # avoid that --sysroot is passed to linker by patching value for SYSROOT_SPEC in gcc/gcc.c
-            apply_regex_substitutions(os.path.join('gcc', 'gcc.c'), [('--sysroot=%R', '')])
-
             # prefix dynamic linkers with sysroot
             # this patches lines like:
             # #define GLIBC_DYNAMIC_LINKER64 "/lib64/ld-linux-x86-64.so.2"


bedroge:
In order to not break older installations that still use the old approach, we would have to add some check(s) here. We can perhaps do that by inspecting the file `$EPREFIX/usr/lib64/libc.so`, and find out if the paths in `GROUP( .... )` are prefixed by the value of `sysroot`?

boegel:
Yeah, we should figure out a way to opt-out of patching out the `--sysroot` from `gcc.c` conditionally, since it is required for older EESSI versions.

What was changed exactly that no longer required this, do we have references for this?

cc @bartoldeman, who no doubt can provide some guided info here...

ocaisa:
I think we just add a switch here to turn this off, and a hook in eessi to trigger it

bedroge:
> Yeah, we should figure out a way to opt-out of patching out the `--sysroot` from `gcc.c` conditionally, since it is required for older EESSI versions.
> 
> What was changed exactly that no longer required this, do we have references for this?
> 
> cc @bartoldeman, who no doubt can provide some guided info here...

See: https://github.com/gentoo/gentoo/pull/28851

bedroge:
> In order to not break older installations that still use the old approach, we would have to add some check(s) here. We can perhaps do that by inspecting the file `$EPREFIX/usr/lib64/libc.so`, and find out if the paths in `GROUP( .... )` are prefixed by the value of `sysroot`?

The comment at https://github.com/gentoo/gentoo/pull/28851/files#diff-bbba0e850dfb18cee4bb09d6c00d235e935e24baa2199ce82a628d84dd16cb58R1317 shows the difference between Prefix installations with the old and new approach:

```
	# Before: GROUP ( /foo/lib64/libc.so.6 /foo/usr/lib64/libc_nonshared.a  AS_NEEDED ( /foo/lib64/ld-linux-x86-64.so.2 ) )
	# After: GROUP ( /lib64/libc.so.6 /usr/lib64/libc_nonshared.a  AS_NEEDED ( /lib64/ld-linux-x86-64.so.2 ) )
```

So we may be able to check those paths and determine if they are "prefixed" (with the same prefix as specified with `--sysroot`) or not.

bartoldeman:
@trz42 
See https://github.com/trz42/easybuild-easyblocks/pull/1 for my attempt. I verified it still does the patching in our older sysroot-ed Gentoo installation.

boegel:
This has been tested extensively in EESSI, cfr. https://github.com/EESSI/software-layer/issues/250

Only has impact when EasyBuild is configured with an alternate `sysroot`.

I'll go ahead and merge this to be included last-minute with EasyBuild v4.7.2, test reports are coming up to confirm this doesn't cause trouble.

Thanks for the effort on this @trz42 and @bartoldeman!

boegel (approved):
lgtm

boegel:
Test report by @boegel
 * **SUCCESS** _GCCcore-9.3.0.eb_ 
 * **SUCCESS** _GCCcore-11.3.0.eb_ 
 * **SUCCESS** _GCCcore-12.2.0.eb_ 
 * **SUCCESS** _GCCcore-13.1.0.eb_ 

node5414.dodrio.os - Linux RHEL 8.6, x86_64, AMD EPYC 7763 64-Core Processor (zen3), Python 3.6.8
See https://gist.github.com/boegel/08d674edb228f44ee06bd6cf490becf2 for a full test report.

PR #2922, opened 2023-04-23
use proxy in sanity check of `EB_OpenSSL_wrapper`

trz42:
One sanity check that tries to connect to github.com fails, when `OpenSSL-1.1.eb` is built on a machine without direct access to the internet.

This PR resolves the issue by adding the parameter `-proxy` when `openssl` is run. It only does so when the environment variable `https_proxy` is defined. `urlparse` from `urllib.parse` is used to determine the host and port of the proxy setting which could be a URL such as `https://localproxy.cluster:4321/`.

The change has been tested for environments with and without proxies, ie, direct access to the internet (environment variable `https_proxy` not set) and no direct access to the internet (environment variable `https_proxy` set). See comments https://github.com/NorESSI/software-layer/pull/100#issuecomment-1519069430, https://github.com/NorESSI/software-layer/pull/100#issuecomment-1519069443 (both with direct access to internet) and https://github.com/NorESSI/software-layer/pull/100#issuecomment-1519069432 (no direct access to internet, original build failed, see https://github.com/NorESSI/software-layer/pull/100#issuecomment-1519055078). The changed easyblock was used via `--include-easyblocks=path/to/openssl_wrapper.py`.


easybuild/easyblocks/o/openssl_wrapper.py
```diff
@@ -368,10 +369,20 @@ def sanity_check_step(self):
...
         }
 
+        # use proxy to connect if https_proxy environment variable is defined
+        proxy_arg = ''
+        if os.environ.get('https_proxy'):
+            # only use host & port from https_proxy env var, that is, strip
+            # any protocol prefix and trailing slashes
+            proxy_parsed = urlparse(os.environ.get('https_proxy'))
+            if proxy_parsed.netloc:
+                proxy_arg = f' -proxy {proxy_parsed.netloc}'
```
jfgrimm, line 379: EasyBuild still supports running with Python 2.7 and 3.5, which don't have f-string support

akesandgren: @trz42 ping, please fix

trz42:
Just updated the PR @akesandgren @jfgrimm ... apologies for the long delay.

boegel:
Test report by @boegel
 * **SUCCESS** _OpenSSL-1.0.eb_ 
 * **SUCCESS** _OpenSSL-1.1.eb_ 

node3110.skitty.os - Linux RHEL 8.8, x86_64, Intel(R) Xeon(R) Gold 6140 CPU @ 2.30GHz, Python 3.6.8
See https://gist.github.com/boegel/b32c5a59a4e2d4115dad1995a823acad for a full test report.

boegel (approved):
lgtm, also tested with `OpenSSL/3` wrapper from https://github.com/easybuilders/easybuild-easyconfigs/pull/19623

PR #2930, opened 2023-05-18
Add TURBOMOLE easyblock

sven-hansen:
This PR adds an easyblock for the quantum chemistry software TURBOMOLE. The reasons why I created this easyblock are support for TmoleX, a GUI that is often used alongside TURBOMOLE and support for its integrated tests. We have successfully installed TURBOMOLE and TmoleX on our site using an earlier version of this easyblock and the installations behave as expected.

I will add another PR to the easyconfigs repository that contains ECs for several recent versions and parallelisation flavors.

### DIFF
--- a/easybuild/easyblocks/t/turbomole.py
+++ b/easybuild/easyblocks/t/turbomole.py
@@ -1,0 +1,216 @@
+"""
+EasyBuild support for TURBOMOLE,
+implemented as an easyblock
+
+@author: Sven Hansen (RWTH Aachen University)
+"""
+
+import os
+import re
+import stat
+
+from easybuild.easyblocks.generic.tarball import Tarball
+from easybuild.framework.easyblock import read_file
+from easybuild.framework.easyconfig import CUSTOM
+from easybuild.tools.config import WARN
+from easybuild.tools.build_log import EasyBuildError
+from easybuild.tools.environment import setvar
+from easybuild.tools.filetools import change_dir, adjust_permissions, apply_regex_substitutions
+from easybuild.tools.run import run_cmd
+
+
+class EB_TURBOMOLE(Tarball):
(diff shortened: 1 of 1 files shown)


backelj (approved):
Hi @sven-hansen, I have just installed TURBOMOLE 7.7.1 using the turbomole easyblock you provide here, and it's working just fine, so thanks for that!

Just one remark though: since you include the description "\n\n This module contains the TmoleX GUI.", I would leave out this description in the TURBOMOLE easyconfigs you use in #17922, otherwise the same line occurs twice.

For me, this easyblock is good to go.

easybuild/easyblocks/t/turbomole.py
```diff
@@ -0,0 +1,214 @@
...
+import stat
+
+from easybuild.easyblocks.generic.tarball import Tarball
+from easybuild.framework.easyblock import read_file
+from easybuild.framework.easyconfig import CUSTOM
+from easybuild.tools.config import WARN
+from easybuild.tools.build_log import EasyBuildError
+from easybuild.tools.environment import setvar
+from easybuild.tools.filetools import change_dir, adjust_permissions, apply_regex_substitutions
+from easybuild.tools.run import run_cmd
```
boegel: @sven-hansen Sorry for the radio silence on this one...

Can you look into updating this to be compatible with EasyBuild 5.x (in which we deprecated `run_cmd` in favor of `run_shell_cmd`)?

sven-hansen: @boegel Newest commit uses `run_shell_cmd` and I've replaced the easyconfigs PR with a newer one ( https://github.com/easybuilders/easybuild-easyconfigs/pull/24648). The test report is attached there, should be good to go this time. 🙂

PR #2934, opened 2023-05-24
enhance Cargo constructor to avoid processing list of crates multiple times into sources

boegel:
(created using `eb --new-pr`)


Micket (dismissed):
lgtm

easybuild/easyblocks/generic/cargo.py
```diff
@@ -73,30 +73,36 @@ def __init__(self, *args, **kwargs):
...
-                crate, version, repo, rev = crate_info
-                url, repo_name_git = repo.rsplit('/', maxsplit=1)
-                sources.append({
-                    'git_config': {'url': url, 'repo_name': repo_name_git[:-4], 'commit': rev},
-                    'filename': crate + '-' + version + '.tar.gz',
-                    'source_urls': [CRATESIO_SOURCE],
-                })
+        # Populate sources from "crates" list of tuples (only once)
+        if self.cfg['crates']:
+            # copy list of crates, so we can wipe 'crates' easyconfig paramter,
```
ocaisa, line 78:
```suggestion
            # copy list of crates, so we can wipe 'crates' easyconfig parameter,
```

ocaisa (approved):
Restoring the positive review from @Micket  after fixing typo

PR #2955, opened 2023-06-30
update MXNet easyblock + don't try to install R extension by default for MXNet >= 1.0

deniskristak:
changing mxnet easyblock for https://github.com/vscentrum/vsc-software-stack/issues/89

### DIFF
--- a/easybuild/easyblocks/m/mxnet.py
+++ b/easybuild/easyblocks/m/mxnet.py
@@ -32,6 +32,7 @@ import os
 import shutil
 from distutils.version import LooseVersion
 
+import easybuild.tools.environment as env
 from easybuild.easyblocks.generic.makecp import MakeCp
 from easybuild.easyblocks.generic.pythonpackage import PythonPackage
 from easybuild.easyblocks.generic.rpackage import RPackage
@@ -40,7 +41,6 @@ from easybuild.tools.build_log import EasyBuildError
 from easybuild.tools.filetools import change_dir, mkdir, remove_dir, symlink, write_file
 from easybuild.tools.modules import get_software_root, get_software_version
 from easybuild.tools.run import run_cmd
-import easybuild.tools.environment as env
 from easybuild.tools.systemtools import get_shared_lib_ext
 
 # the namespace file for the R extension
@@ -110,7 +110,7 @@ class EB_MXNet(MakeCp):
         for srcdir in [d for d in os.listdir(self.builddir) if d != os.path.basename(self.mxnet_src_dir)]:
             submodule, _, _ = srcdir.rpartition('-')
 
-            if LooseVersion(self.version) > LooseVersion('1.0'):
+            if LooseVersion(self.version) >= LooseVersion('1.0'):
                 # if newdir starts with 'oneDNN-', we rename it to mkldnn:
                 if submodule == 'oneDNN':
                     submodule = 'mkldnn'
@@ -122,8 +122,8 @@ class EB_MXNet(MakeCp):
                 olddir = os.path.join(self.builddir, srcdir)
                 newdir = os.path.join(self.mxnet_src_dir, '3rdparty', submodule)
             else:
-                newdir = os.path.join(self.mxnet_src_dir, submodule)
                 olddir = os.path.join(self.builddir, srcdir)
+                newdir = os.path.join(self.mxnet_src_dir, submodule)
 
             # first remove empty existing directory
             remove_dir(newdir)
@@ -134,7 +134,6 @@ class EB_MXNet(MakeCp):
                 raise EasyBuildError("Failed to move %s to %s: %s", olddir, newdir, err)
 
             if LooseVersion(self.version) < LooseVersion('1.0'):
-                print('yaayyayayayayayaa')
                 # the nnvm submodules has dmlc-core as a submodule too. Let's put a symlink in place.
                 newdir = os.path.join(self.mxnet_src_dir, "nnvm", "dmlc-core")
                 olddir = os.path.join(self.mxnet_src_dir, "dmlc-core")
@@ -152,7 +151,6 @@ class EB_MXNet(MakeCp):
             self.cfg.update('buildopts', '%s="%s"' % (var, os.getenv(env_var)))
 
         toolchain_blas = self.toolchain.definition().get('BLAS', None)[0]
-        print('toolchain_blas: %s' % toolchain_blas)
         if toolchain_blas == 'imkl':
             blas = "mkl"
             imkl_version = get_software_version('imkl')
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/m/mxnet.py
```diff
@@ -169,7 +187,7 @@ def extensions_step(self):
         # MXNet doesn't provide a list of its R dependencies by default
         write_file("NAMESPACE", R_NAMESPACE)
         change_dir(self.mxnet_src_dir)
-        self.r_ext.prerun()
+        self.r_ext.prerun()  # tried commenting this out
```
Micket, line 190: So, should we comment this out or not? why?

easybuild/easyblocks/m/mxnet.py
```diff
@@ -97,21 +101,26 @@ def extract_step(self):
 
         for srcdir in [d for d in os.listdir(self.builddir) if d != os.path.basename(self.mxnet_src_dir)]:
             submodule, _, _ = srcdir.rpartition('-')
-            newdir = os.path.join(self.mxnet_src_dir, submodule)
+
+            # if newdir starts with 'oneDNN-', we rename it to mkldnn:
```
Micket, line 105: Rather than commenting out what the code literally does, please explain why this is necessary

easybuild/easyblocks/m/mxnet.py
deniskristak: no, sorry just a leftover debug thing

deniskristak: solved

easybuild/easyblocks/m/mxnet.py
deniskristak: solved

easybuild/easyblocks/m/mxnet.py
```diff
@@ -97,21 +101,26 @@ def extract_step(self):
 
         for srcdir in [d for d in os.listdir(self.builddir) if d != os.path.basename(self.mxnet_src_dir)]:
             submodule, _, _ = srcdir.rpartition('-')
-            newdir = os.path.join(self.mxnet_src_dir, submodule)
+
+            # the package mkldnn was renamed to oneDNN since this easyblock was written
```
Micket, line 105: I'm still confused by this. This easyblock makes no assumptions about the name of this package.
If MXNet's build scripts assuming the old name for this submodule? If so

```suggestion
            # MXNet assumes the old for oneDNN which was mkldnn
```

easybuild/easyblocks/m/mxnet.py
```diff
@@ -133,6 +142,8 @@ def configure_step(self):
             blas = "atlas"
         elif toolchain_blas == 'OpenBLAS':
             blas = "openblas"
+        elif toolchain_blas == 'FlexiBLAS':
+            blas = "flexiblas"
```
boegel, line 146: Based on the discussion in https://github.com/easybuilders/easybuild-easyconfigs/pull/18251/files/50b231913a3812526e4247a4a167eb901aae9540#diff-64ce5076894e1d94c36062fe332ab40cef145af9d5ecdd785cd4cdb59c8c0f0d, we also need to add `-lflexiblas` to `$CFLAGS`, which we could do here:

```suggestion
        elif toolchain_blas == 'FlexiBLAS':
            blas = "flexiblas"
            setvar('CFLAGS', "%s -lflexiblas" % os.getenv('CFLAGS'))

File: easybuild/easyblocks/m/mxnet.py [hunk 5]
```diff
@@ -63,11 +63,15 @@ class EB_MXNet(MakeCp):
     """Easyblock to build and install MXNet"""
 
     @staticmethod
-    def extra_options(extra_vars=None):
+    def extra_options():
         """Change default values of options"""
-        extra = MakeCp.extra_options()
+        extra_vars = {
+            'install_r_ext': [False, "Enable installation of R extensions", CUSTOM],
```
boegel, line 69: If we set the default to `None`, we can keep this backwards compatible by only auto-enabling it for old versions of MXNet:

```python
        extra_vars = {
            'install_r_ext': [None, "Enable installation of R extensions", CUSTOM],
}
```

and then in the `__init__` below:

```python
# auto-enable building of R extensions only for old versions of MXNet (< 1.0),
# since for newer versions is broken
if self.cfg['install_r_ext'] is None:
    if LooseVersion(self.version) < LooseVersion('1.0'):
      self.cfg['install_r_ext'] = True
    else:
      self.cfg['install_r_ext'] = False
```

easybuild/easyblocks/m/mxnet.py
```diff
@@ -145,8 +156,7 @@ def configure_step(self):
 
     def install_step(self):
         """Specify list of files to copy"""
-        self.cfg['files_to_copy'] = ['bin', 'include', 'lib',
-                                     (['dmlc-core/include/dmlc', 'nnvm/include/nnvm'], 'include')]
+        self.cfg['files_to_copy'] = ['bin', 'include', 'lib']
```
boegel, line 159: keep backwards-compatible:

```python
if LooseVersion(self.version) < LooseVersion('1.0'):
    self.cfg.update('files_to_copy', [(['dmlc-core/include/dmlc', 'nnvm/include/nnvm'], 'include')])

File: easybuild/easyblocks/m/mxnet.py [hunk 7]
```diff
@@ -97,21 +101,26 @@ def extract_step(self):
...
+            # MXNet assumes the old name for oneDNN which was mkldnn
+            if submodule == 'oneDNN':
+                submodule = 'mkldnn'
+                # rename the file to 'mkldnn':
+                old_srcdir = srcdir
+                srcdir = srcdir.replace('oneDNN', 'mkldnn')
+                os.rename(os.path.join(self.builddir, old_srcdir), os.path.join(self.builddir, srcdir))
+
             olddir = os.path.join(self.builddir, srcdir)
+            newdir = os.path.join(self.mxnet_src_dir, '3rdparty', submodule)
```
boegel, line 114: Does we also need the `1.0` version condition here w.r.t. using `3rdparty` or not?

boegel (changes requested):

easybuild/easyblocks/m/mxnet.py
deniskristak: done

easybuild/easyblocks/m/mxnet.py
deniskristak: solved

boegel:
Tested via https://github.com/easybuilders/easybuild-easyconfigs/pull/18251, good to go!

PR #2947, opened 2023-06-05
improve handling of `optarch` in `Cargo` easyblock

Micket:
(created using `eb --new-pr`)

PR #2969, opened 2023-07-31
Fix type check in cargo EasyBlock

Flamefire:
Instead of `type(foo) == bar` `isinstance(foo, bar)` should be used. [Otherwise flake8 will flag this](https://github.com/easybuilders/easybuild-easyblocks/actions/runs/5711759534/job/15473935720?pr=2968)

Also reduce the indentation by using `if "is-dict", elseif "is-generic", else` style

Followup to #2947 by @Micket

### DIFF
--- a/easybuild/easyblocks/generic/cargo.py
+++ b/easybuild/easyblocks/generic/cargo.py
@@ -74,7 +74,7 @@ class Cargo(ExtensionEasyBlock):
 
         optarch = build_option('optarch')
         if optarch:
-            if type(optarch) == dict:
+            if isinstance(optarch, dict):
                 if 'rustc' in optarch:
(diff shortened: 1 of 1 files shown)


migueldiascosta:
lgtm, just wondering why flake8 is only flagging this now?

Flamefire:
> lgtm, just wondering why flake8 is only flagging this now?

In the initial PR CI used flake8-6.0.0 while now it uses flake8-6.1.0. That also updated pyflakes & pycodestyle.   
So I guess that check is new or more thorough.

migueldiascosta (approved):
lgtm

migueldiascosta:
Going in, thanks @Flamefire!

PR #2970, opened 2023-08-01
update LLVM easyblock for LLVM v16: symlink `third-party` to `third-party-<version>.src`

maxim-masterov:
Required to install LLVM-16.x

### DIFF
--- a/easybuild/easyblocks/l/llvm.py
+++ b/easybuild/easyblocks/l/llvm.py
@@ -112,7 +112,7 @@ class EB_LLVM(CMakeMake):
                 change_dir(cwd)
             else:
                 raise EasyBuildError("Failed to find unpacked CMake modules directory at %s", cmake_modules_path)
-        
+
         if LooseVersion(self.version) >= LooseVersion('16.0'):
             # make sure that third-party modules are available in build directory,
             # and if so make a 'third-party' symlink so LLVM can find them
--- a/easybuild/easyblocks/l/llvm.py
+++ b/easybuild/easyblocks/l/llvm.py
@@ -122,7 +122,7 @@ class EB_LLVM(CMakeMake):
                 symlink('third-party-%s.src' % self.version, 'third-party')
                 change_dir(cwd)
             else:
-                raise EasyBuildError("Failed to find unpacked 'third-party' modules directory at %s", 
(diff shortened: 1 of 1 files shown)


migueldiascosta:
wondering if it wouldn't be clearer to use something like `'extract_cmd': "mkdir third-party && tar xf third-party-%(version)s.src.tar.xz --strip-components=1 -C third-party"` in the easyconfig - if something else changes in the future (e.g. the `third-party-%s.src` folder name format), the symlink approach will require more version checking in the the easyblock, handling this explicitly in the easyconfig may be preferable (?)

@SebastianAchilles @surak @casparvl, since you're testing the LLVM easyconfig, what's your opinion?

boegel (approved):
lgtm

easybuild/easyblocks/l/llvm.py
```diff
@@ -113,4 +113,16 @@ def configure_step(self):
...
                 raise EasyBuildError("Failed to find unpacked CMake modules directory at %s", cmake_modules_path)
 
+        if LooseVersion(self.version) >= LooseVersion('16.0'):
+            # make sure that third-party modules are available in build directory,
+            # and if so make a 'third-party' symlink so LLVM can find them
+            third_party_modules_path = os.path.join(self.builddir, 'third-party-%s.src' % self.version)
+            if os.path.exists(third_party_modules_path):
+                cwd = change_dir(self.builddir)
+                symlink('third-party-%s.src' % self.version, 'third-party')
+                change_dir(cwd)
```
Flamefire, line 123: 2 Questions here: Why the symlink and not a move? The name "third-party-version.src" is only because of the filename of the archive, I don't see how it is required to be kept.

2nd: Why the `change_dir`? `symlink('third-party-%s.src' % self.version, os.path.join(self.builddir, 'third-party'), use_abspath_source=False)` would be better, wouldn't it?   
But I'd just rename it instead.

boegel: @Flamefire If you think it's worth changing this, please open an issue or a PR, following up on stuff in merged PRs is a nightmare I don't want to get into :)

Flamefire: I see: It was C&P from the cmake folder above. Opened PR: https://github.com/easybuilders/easybuild-easyblocks/pull/2994

PR #2992, opened 2023-09-01
Better conda support

xerbalind:


ocaisa:
I think we may need some additional changes in the final module file for a conda package. We should consider:

- Should each conda package should have a `family` attached so that you cannot have multiple loaded at once? In the current approach if you have two loaded, do they interfere with each other (since the second module will overwrite `CONDA_ENV`/`CONDA_PREFIX`/`CONDA_DEFAULT_ENV` , see [here](https://github.com/easybuilders/easybuild-easyblocks/blob/4b7aefb0c025f27e6a118c5194a482eedc594408/easybuild/easyblocks/generic/conda.py#L140-L142))?
- Should we conflict with `GCCcore`? This approach was suggested by @branfosj so that a conda installation is never used as a dependency (since it is _extremely_ likely to interfere with the build)
- Currently the final module sets `PKG_CONFIG_PATH` and I think this might be a mistake as packages may pick up on whatever this contains.

boegel:
I'll look into cleaning this up, and making sure the changes are backwards compatible.

@xerbalind was working in the HPC-UGent team as summer intern, and opened this WIP PR on my request to ensure his efforts were not lost.

ocaisa:
@boegel There are some straightforward improvements to the conda easyblock in #2996 (which also covers the `mamba` use case in a more generic way) that I think I will merge once I've added some test reports. Since you'll take over here I think it'll be ok to do that and then you can integrate that into this branch.

PR #2996, opened 2023-09-07
enhance `Conda` easyblock: add support for using custom `conda` command (like `mamba`), and clean up after installation

rsdmse:
- Set `DEFAULT_CONDA_CMD = 'conda'` and allow for user-defined alternative (e.g. `mamba`)
- Run `conda clean -ya` after env installation to clean up cache files

### DIFF
--- a/easybuild/easyblocks/generic/conda.py
+++ b/easybuild/easyblocks/generic/conda.py
@@ -35,6 +35,7 @@ from easybuild.easyblocks.generic.binary import Binary
 from easybuild.framework.easyconfig import CUSTOM
 from easybuild.tools.run import run_cmd
 
+DEFAULT_CONDA_CMD = 'conda'
 
 class Conda(Binary):
     """Support for installing software using 'conda'."""
@@ -44,6 +45,7 @@ class Conda(Binary):
         """Extra easyconfig parameters specific to Conda easyblock."""
         extra_vars = Binary.extra_options(extra_vars)
         extra_vars.update({
+            'conda_cmd': [DEFAULT_CONDA_CMD, "Conda command to use", CUSTOM],
             'channels': [None, "List of conda channels to pass to 'conda install'", CUSTOM],
             'environment_file': [None, "Conda environment.yml file to use with 'conda env create'", CUSTOM],
             'remote_environment': [None, "Remote conda environment to use with 'conda env create'", CUSTOM],
@@ -58,10 +60,11 @@ class Conda(Binary):
 
     def install_step(self):
         """Install software using 'conda env create' or 'conda create' & 'conda install'."""
+        conda_cmd = self.cfg.get('conda_cmd')
 
         # initialize conda environment
         # setuptools is just a choice, but *something* needs to be there
-        cmd = "conda config --add create_default_packages setuptools"
+        cmd = "%s config --add create_default_packages setuptools" % conda_cmd
         run_cmd(cmd, log_all=True, simple=True)
 
         if self.cfg['environment_file'] or self.cfg['remote_environment']:
@@ -72,7 +75,7 @@ class Conda(Binary):
                 env_spec = self.cfg['remote_environment']
 
             # use --force to ignore existing installation directory
-            cmd = "%s conda env create --force %s -p %s" % (self.cfg['preinstallopts'], env_spec, self.installdir)
+            cmd = "%s %s env create --force %s -p %s" % (self.cfg['preinstallopts'], conda_cmd, env_spec, self.installdir)
             run_cmd(cmd, log_all=True, simple=True)
 
         else:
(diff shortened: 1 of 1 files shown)


ocaisa:
@boegelbot please test @ generoso EB_ARGS="TRUST4-1.0.5.1.eb"

ocaisa:
@boegelbot please test @ generoso EB_ARGS="TRUST4-1.0.5.1.eb"

ocaisa:
@boegelbot please test @ generoso EB_ARGS="TRUST4-1.0.5.1.eb"

easybuild/easyblocks/generic/conda.py
```diff
@@ -44,6 +46,7 @@ def extra_options(extra_vars=None):
         """Extra easyconfig parameters specific to Conda easyblock."""
         extra_vars = Binary.extra_options(extra_vars)
         extra_vars.update({
+            'conda_cmd': [DEFAULT_CONDA_CMD, "Conda command to use", CUSTOM],
```
boegel, line 49: Are `conda` and `mamba` on par w.r.t. supported options?

I'm just wondering if a `use_mamba` easyconfig parameter that can be set to `True` would be a better option here, not sure.

ocaisa: The `mamba` docs say yes: https://mamba.readthedocs.io/en/latest/user_guide/mamba.html#mamba-vs-conda-clis

ocaisa: `micromamba` seems to support everything currently in the easyblock: https://mamba.readthedocs.io/en/latest/user_guide/micromamba.html#quickstarts

ocaisa: If you were being picky you could add a list of supported Conda commands, but I think that is overkill right now

branfosj: Rather than set conda vs mamba via a parameter, should be not detect which is available as a dependency and use it?

ocaisa (changes requested):
@rsdmse I think I agree with @branfosj that rather than set `conda_cmd` on the basis of a config variable we should check the environment. Search for `get_software_root` in this repo to get some examples of usage. For example:
```
...
if get_software_root('mamba'):
    conda_cmd = 'mamba'
elif get_software_root('conda'):
    conda_cmd = 'conda'
else
    raise EasyBuildError("Neither mamba or conda available as a dependency!")
```

rsdmse:
@ocaisa Thanks for the suggestion and code snippet. I've taken into account of the variants that I'm aware of: anaconda, miniconda, mamba, micromamba. If the number of variants keep expanding and go out of hand we may have to revert to allowing a user-defined command.

easybuild/easyblocks/generic/conda.py
```diff
@@ -57,11 +59,19 @@ def extract_step(self):
             super(Conda, self).extract_step()
 
     def install_step(self):
-        """Install software using 'conda env create' or 'conda create' & 'conda install'."""
+        if get_software_root('anaconda') or get_software_root('miniconda'):
```
ocaisa, line 62:
```suggestion
        if get_software_root('anaconda2') or get_software_root('miniconda2') or get_software_root('anaconda3') or get_software_root('miniconda3'):
```
The current batch of easyconfigs that match this dependency are called `Anaconda2/Anaconda3` and `Miniconda2/Miniconda3`.

Note that Mamba also includes a `conda` command so we do actually need to check the dependency to make the right choice.

easybuild/easyblocks/generic/conda.py
```diff
@@ -57,11 +59,19 @@ def extract_step(self):
...
+            conda_cmd = 'conda'
+        elif get_software_root('mamba'):
+            conda_cmd = 'mamba'
+        elif get_software_root('micromamba'):
```
ocaisa, line 66: There is no easyconfig for micromamba, but it does no harm so I leave it here.

ocaisa (changes requested):

easybuild/easyblocks/generic/conda.py
```diff
@@ -57,11 +59,19 @@ def extract_step(self):
             super(Conda, self).extract_step()
 
     def install_step(self):
-        """Install software using 'conda env create' or 'conda create' & 'conda install'."""
+        if (get_software_root('anaconda2') or get_software_root('miniconda2') or
```
ocaisa, line 62:
```suggestion
        """Install software using 'conda env create' or 'conda create' & 'conda install' (or the 'mamba', etc., equivalent)."""
        if (get_software_root('anaconda2') or get_software_root('miniconda2') or
```
Please keep a docstring around

ocaisa:
@boegelbot please test @ generoso EB_ARGS="TRUST4-1.0.5.1.eb"

ocaisa (approved):

ocaisa:
Going in, thanks for the patience @rsdmse !

rsdmse:
@ocaisa Thanks for all the suggestions!

PR #3018, opened 2023-10-11
Use `LooseVersion` from `easybuild.tools` in all easyblocks

appolloford:


easybuild/easyblocks/__init__.py
```diff
@@ -32,9 +32,10 @@
 @author: Jens Timmerman (Ghent University)
 """
 import os
-from distutils.version import LooseVersion
 from pkgutil import extend_path
 
+from easybuild.tools.loose_version import LooseVersion
```
Micket, line 37: I think it's specifically this import that trips up the CI, this is the thing that's used for installed the easyblocks pip package itself.

Really not sure how we fix this though...

easybuild/easyblocks/a/abaqus.py
```diff
@@ -35,14 +35,14 @@
...
 from collections import OrderedDict
-from distutils.version import LooseVersion
 
 from easybuild.easyblocks.generic.binary import Binary
 from easybuild.framework.easyblock import EasyBlock
 from easybuild.framework.easyconfig import CUSTOM
 from easybuild.tools.build_log import EasyBuildError
 from easybuild.tools.environment import setvar
 from easybuild.tools.filetools import change_dir, symlink, write_file
+from easybuild.tools.loose_version import LooseVersion
```
boegel, line 45: Please use `from easybuild.tools import LooseVersion` instead, like we do everywhere in framework?

boegel (changes requested):

easybuild/easyblocks/a/abaqus.py
Micket: (I think this was just a test due to the import not working in the CI)

Micket:
So, I suggested to @appolloford to `isort` these to get the import in the right place, but perhaps this is a painful list to deal with when merging this into develop later since there were so many places where this wasn't followed strictly.

I tried to see in the isort conf if there was some way to limit it to not change so much, like only sorting on type nor alphabetical, but couldn't find anything like that. :(

boegel (approved):
lgtm

boegel:
This is good to go, so I've promoted this from "draft" to "ready to review".

A corresponding change in the framework repo may be needed because both `VERSION` and `VERBOSE_VERSION` are not string values rather than `LooseVersion` instances...

PR #3024, opened 2023-10-24
Fix the blas/lapack name passed to meson when building recent scipy versions (>= 1.9.0) on top of Intel MKL

schiotz:
The meson script expects the ``pkg-config`` name of the MKL package, instead of just 'mkl'.  See https://github.com/easybuilders/easybuild-easyconfigs/pull/18875


schiotz:
Note that this should not break any existing easyconfigs (*), as there is no intel-based SciPy-bundle new enough to build using meson.

(*) Famous last words...

boegel:
EB_ARGS="SciPy-bundle-2022.05-intel-2022a.eb"

boegel:
@boegelbot please test @ jsc-zen2
EB_ARGS="SciPy-bundle-2022.05-intel-2022a.eb"

boegel (approved):
lgtm

PR #3034, opened 2023-11-15
Added easyblock for flook that includes install step

ahkole:
The software flook (https://github.com/ElectronicStructureLibrary/flook) has an unconventional build system. It doesn't have a configure script and requires a manual config file, but it does have an install step. The currently available flook easyconfig uses the `MakeCp` generic easyblock which works fine in most cases but it doesn't automatically install the pkgconf that might be needed by other software to automatically detect that flook has been installed. This PR adds a new easyblock for flook that also performs the install step so that the pkgconf file is generated and installed.

### DIFF
--- a/easybuild/easyblocks/f/flook.py
+++ b/easybuild/easyblocks/f/flook.py
@@ -1,0 +1,43 @@
+##
+# Copyright 2023 Utrecht University
+#
+# This file is part of EasyBuild,
+# originally created by the HPC team of Ghent University (http://ugent.be/hpc/en),
+# with support of Ghent University (http://ugent.be/hpc),
+# the Flemish Supercomputer Centre (VSC) (https://www.vscentrum.be),
+# Flemish Research Foundation (FWO) (http://www.fwo.be/en)
+# and the Department of Economy, Science and Innovation (EWI) (http://www.ewi-vlaanderen.be/en).
+#
+# https://github.com/easybuilders/easybuild
+#
+# EasyBuild is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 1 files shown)


ahkole:
The easyblock is used in https://github.com/easybuilders/easybuild-easyconfigs/pull/19238

ocaisa (approved):
LGTM

ocaisa:
EasyBlock tests are in https://github.com/easybuilders/easybuild-easyconfigs/pull/19238 and works great, thanks a lot @ahkole!

boegel:
@ahkole Would it make sense to also update the existing `flook` easyconfigs to use this easyblock (at least the ones using a supported toolchain, see also https://docs.easybuild.io/policies/toolchains)?

ahkole:
> @ahkole Would it make sense to also update the existing `flook` easyconfigs to use this easyblock (at least the ones using a supported toolchain, see also https://docs.easybuild.io/policies/toolchains)?

@boegel I think this would make sense yes. Looking at the other `flook` easyconfigs they only differ in the toolchain used and the new version properly installs the pkgconfig file which makes it detectable by build systems like cmake. The other easyconfigs could also benefit from this. Should I open a PR for this?

boegel:
> > @ahkole Would it make sense to also update the existing `flook` easyconfigs to use this easyblock (at least the ones using a supported toolchain, see also https://docs.easybuild.io/policies/toolchains)?
> 
> @boegel I think this would make sense yes. Looking at the other `flook` easyconfigs they only differ in the toolchain used and the new version properly installs the pkgconfig file which makes it detectable by build systems like cmake. The other easyconfigs could also benefit from this. Should I open a PR for this?

@ahkole Please do!

PR #3045, opened 2023-11-30
Add requirement for EULA acceptance to CUDA easyblock

ocaisa:


ocaisa:
@boegelbot please test @ generoso EB_ARGS="CUDA-12.1.1.eb"

ocaisa:
Urgh, that shouldn't have worked but did because the sources were already on disk. @boegel Can you remove them so I can ask again (expecting a failure)

ocaisa:
Test report by @ocaisa
 * **SUCCESS** _CUDA-12.1.1.eb_ 

login1 - Linux Rocky Linux 8.5, x86_64, Intel(R) Xeon(R) CPU E5-2667 v3 @ 3.20GHz, Python 3.6.8
See https://gist.github.com/ocaisa/9414518b4479eb80ef7d87a336b9b850 for a full test report.

ocaisa:
And that one is with the updated command line:
```
eb --include-easyblocks-from-pr=3045 CUDA-12.1.1.eb --rebuild --accept-eula-for=CUDA --upload-test-report
```

ocaisa:
I think I'll fold the script to the linking to the host drivers in here as well...working on it now

mboisson:
My point is that the software installer is *not* the end user. Merely a middle-person.

ocaisa:
That's problematic though, as you don't need accept the EULA to run a CUDA application. To be honest, the way we've implemented CUDA support in EESSI this would indeed be possible, we've dropped CUDA to a build dep and rely on RPATH for running the cuda applications. You could require the end user to set a certain environment variable to allow them to load the CUDA module itself. You can do that by configuring an lmod RC file appropriately along the lines of what's done in https://github.com/EESSI/software-layer/pull/381

Unfortunately that's not a global solution though, so I still this as a reasonable approximation to what is happening a lot in practice.

easybuild/easyblocks/c/cuda.py
```diff
@@ -92,6 +92,16 @@ def __init__(self, *args, **kwargs):
         self.cfg.template_values['cudaarch'] = cudaarch
         self.cfg.generate_template_values()
 
+    def fetch_step(self, *args, **kwargs):
+        """Check for EULA acceptance prior to getting sources."""
+        # EULA for CUDA must be accepted via --accept-eula-for EasyBuild configuration option,
+        # or via 'accept_eula = True' in easyconfig file
+        self.check_accepted_eula(
```
boegel, line 99: I'm fine with checking it here, but maybe we should also check again in `install_step` just in case the install file is already available?

I would do that via a custom method `check_accepted_cuda_eula` in this easyblock, which we call twice (in `fetch_step` + `install_step`, or maybe `extract_step` rather than `install_step`)

boegel (changes requested):

boegel:
> Urgh, that shouldn't have worked but did because the sources were already on disk. I'll do the reports manually

It worked because the bot uses `--accept-eula-for=".*"`, looks good as is.

easybuild/easyblocks/c/cuda.py
boegel: Actually, nevermind, `fetch_step` is always run, even when the install files is already available, as @branfosj pointed out to me, so should be good as is.

boegel (approved):

boegel:
Test report by @boegel
 * **SUCCESS** _CUDA-12.1.0.eb_ 

node4006.donphan.os - Linux RHEL 8.8, x86_64, Intel(R) Xeon(R) Gold 6240 CPU @ 2.60GHz (cascadelake), 1 x NVIDIA NVIDIA A2, 535.129.03, Python 3.6.8
See https://gist.github.com/boegel/c601f3b5d88c51d06fd432feadd59cd3 for a full test report.

PR #3050, opened 2023-12-12
new easyblock for DualSPHysics

jfgrimm:
When building with EasyBuild configured to use rpath, the sanity checks for the existing easyconfig fail because some prebuilt binaries and libraries are bundled in the tarball, and these don't have an RPATH section. The new easyblock runs `patchelf` to add a corresponding entry.

corresponding easyconfigs PR: https://github.com/easybuilders/easybuild-easyconfigs/pull/19400

(created using `eb --new-pr`)

### DIFF
--- a/easybuild/easyblocks/d/dualsphysics.py
+++ b/easybuild/easyblocks/d/dualsphysics.py
@@ -75,7 +75,7 @@ class EB_DualSPHysics(CMakeMakeCp):
     def configure_step(self):
         """Custom configure procedure for DualSPHysics."""
         srcdir = os.path.join(self.cfg['start_dir'], 'src/source')
-        CMakeMakeCp.configure_step(self, srcdir=srcdir)
+        super(EB_DUALSPHysics, self).configure_step(srcdir=srcdir)
 
     def install_step(self):
         """Custom install procedure for DualSPHysics."""
--- a/easybuild/easyblocks/d/dualsphysics.py
+++ b/easybuild/easyblocks/d/dualsphysics.py
@@ -59,7 +59,7 @@ class EB_DualSPHysics(CMakeMakeCp):
         super(EB_DualSPHysics, self).__init__(*args, **kwargs)
 
         self.dsph_target = None
-        self.shortver = None
+        self.shortver = '.'.join(self.version.split('.')[0:2])
 
     def prepare_step(self, *args, **kwargs):
         """Determine name of binary that will be installed."""
@@ -69,8 +69,6 @@ class EB_DualSPHysics(CMakeMakeCp):
             self.dsph_target = 'GPU'
         else:
             self.dsph_target = 'CPU'
-
-        self.shortver = '.'.join(self.version.split('.')[0:2])
 
     def configure_step(self):
         """Custom configure procedure for DualSPHysics."""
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/d/dualsphysics.py
```diff
@@ -0,0 +1,151 @@
...
+            self.dsph_target = 'GPU'
+        else:
+            self.dsph_target = 'CPU'
+
+        self.shortver = '.'.join(self.version.split('.')[0:2])
+
+    def configure_step(self):
+        """Custom configure procedure for DualSPHysics."""
+        srcdir = os.path.join(self.cfg['start_dir'], 'src/source')
+        CMakeMakeCp.configure_step(self, srcdir=srcdir)
```
akesandgren, line 78: Why not "super(EB_DUALSPHysics, self)...." as is used in other places (and is what we normally use)

jfgrimm: good point, somehow overlooked that
```suggestion
        super(EB_DUALSPHysics, self).configure_step(srcdir=srcdir)
```

easybuild/easyblocks/d/dualsphysics.py
```diff
@@ -0,0 +1,151 @@
...
+            self.dsph_target = 'GPU'
+        else:
+            self.dsph_target = 'CPU'
+
+        self.shortver = '.'.join(self.version.split('.')[0:2])
+
+    def configure_step(self):
+        """Custom configure procedure for DualSPHysics."""
+        srcdir = os.path.join(self.cfg['start_dir'], 'src/source')
+        super(EB_DUALSPHysics, self).configure_step(srcdir=srcdir)
```
jfgrimm, line 78:
```suggestion
        super(EB_DualSPHysics, self).configure_step(srcdir=srcdir)
```

jfgrimm:
https://github.com/easybuilders/easybuild-easyconfigs/pull/19400#issuecomment-1877335677

easybuild/easyblocks/d/dualsphysics.py
```diff
@@ -0,0 +1,151 @@
...
+    def prepare_step(self, *args, **kwargs):
+        """Determine name of binary that will be installed."""
+        super(EB_DualSPHysics, self).prepare_step(*args, **kwargs)
+
+        if get_software_root('CUDA'):
+            self.dsph_target = 'GPU'
+        else:
+            self.dsph_target = 'CPU'
+
+        self.shortver = '.'.join(self.version.split('.')[0:2])
```
akesandgren, line 73: self.shortver can be set in __init__ so you don't have to repeat it later on.

akesandgren (changes requested):

easybuild/easyblocks/d/dualsphysics.py
jfgrimm: done

akesandgren:
Tested version of DualSPHysics-5.0.175-GCC-11.2.0.eb before using this easyblock (which fails) and the version from https://github.com/easybuilders/easybuild-easyconfigs/pull/19400 (which works)

akesandgren (approved):
LGTM

akesandgren:
Going in, thanks @jfgrimm!

PR #3053, opened 2023-12-14
add ar symlink when building TensorFlow with --rpath

jfgrimm:
seeing issues like:
```
# Configuration: 59a0a0e80bc61398ca9fef36ddd3d195f45bc65f8bfbc1a894cf677de7d5d5e9
# Execution platform: @local_execution_config_platform//:platform
src/main/tools/process-wrapper-legacy.cc:80: "execvp(/tmp/eb-m2n16g8t/tmptk5axscj/rpath_wrappers/ld_wrapper/ar, ...)": No such file or directory
Target //tensorflow/tools/pip_package:build_pip_package failed to build
```

when easybuild is configured with rpath

(created using `eb --new-pr`)


Flamefire:
Added an alternative symlinking **all** binutils to https://github.com/easybuilders/easybuild-easyblocks/pull/3054

Otherwise the build could fail as that path is passed via `-B`

PR #3054, opened 2023-12-14
Add binutils symlinks when building TensorFlow with --rpath

Flamefire:
This fixes a regression by https://github.com/easybuilders/easybuild-framework/pull/2410

Avoids build errors like
> Configuration: 59a0a0e80bc61398ca9fef36ddd3d195f45bc65f8bfbc1a894cf677de7d5d5e9
> Execution platform: @local_execution_config_platform//:platform
> src/main/tools/process-wrapper-legacy.cc:80: "execvp(/tmp/eb-m2n16g8t/tmptk5axscj/rpath_wrappers/ld_wrapper/ar, ...)": No such file or directory
> Target //tensorflow/tools/pip_package:build_pip_package failed to build

### DIFF
--- a/easybuild/easyblocks/t/tensorflow.py
+++ b/easybuild/easyblocks/t/tensorflow.py
@@ -49,6 +49,7 @@ from easybuild.tools.filetools import is_readable, read_file, symlink, which, wr
 from easybuild.tools.modules import get_software_root, get_software_version, get_software_libdir
 from easybuild.tools.run import run_cmd
 from easybuild.tools.systemtools import AARCH64, X86_64, get_cpu_architecture, get_os_name, get_os_version
+from easybuild.tools.toolchain.toolchain import RPATH_WRAPPERS_SUBDIR
 
 
 CPU_DEVICE = 'cpu'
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/t/tensorflow.py
```diff
@@ -461,6 +464,19 @@ def configure_step(self):
...
         self.binutils_bin_path = os.path.dirname(ld_path)
+        if self.toolchain.is_rpath_wrapper(ld_path):
+            if os.path.basename(os.path.dirname(os.path.dirname(ld_path))) == RPATH_WRAPPERS_SUBDIR:
+                # TF expects all binutils binaries in a single path but newer EB puts each in their own subfolder
+                # Add symlinks to each binutils binary into a single folder
+                new_rpath_wrapper_dir = os.path.join(self.wrapper_dir, RPATH_WRAPPERS_SUBDIR)
+                binutils_bin_path = os.path.join(get_software_root('binutils'), 'bin')
+                self.log.info("Found %s to be an rpath wrapper. Adding symlinks for binutils in %s to %s.",
+                              ld_path, binutils_bin_path, new_rpath_wrapper_dir)
+                mkdir(new_rpath_wrapper_dir)
```
jfgrimm, line 475:
```suggestion
                mkdir(new_rpath_wrapper_dir, parents=True)
```

jfgrimm: good point

jfgrimm:
Test report by @jfgrimm
 * **SUCCESS** _TensorFlow-2.11.0-foss-2022a-CUDA-11.7.0.eb_ 

Build succeeded (with --ignore-test-failure) for 1 out of 1 (1 easyconfigs in total)
himem01.viking2.yor.alces.network - Linux Rocky Linux 8.8, x86_64, AMD EPYC 7643 48-Core Processor, Python 3.6.8
See https://gist.github.com/jfgrimm/10c8c324a0221f7ca2fd307fb035cc25 for a full test report.

jfgrimm (approved):
lgtm

easybuild/easyblocks/t/tensorflow.py
```diff
@@ -461,6 +465,19 @@ def configure_step(self):
         # note that this may be an RPATH wrapper script (when EasyBuild is configured with --rpath)
         ld_path = which('ld')
         self.binutils_bin_path = os.path.dirname(ld_path)
+        if self.toolchain.is_rpath_wrapper(ld_path):
+            if os.path.basename(os.path.dirname(os.path.dirname(ld_path))) == RPATH_WRAPPERS_SUBDIR:
+                # TF expects all binutils binaries in a single path but newer EB puts each in their own subfolder
+                # Add symlinks to each binutils binary into a single folder
+                new_rpath_wrapper_dir = os.path.join(self.wrapper_dir, RPATH_WRAPPERS_SUBDIR)
+                binutils_bin_path = os.path.join(get_software_root('binutils'), 'bin')
```
boegel, line 473: The assumption here that `binutils` is provided as a build dependency is not always correct, since it may be filtered out as dependency (which we do in [EESSI](https://eessi.io)).

So this definitely requires a follow-up PR which has some kind of fallback mechanism in case `get_software_root('binutils')` produces `None`.

boegel: see also #2218, the problem being fixed there is essentially being introduced again here

PR #3058, opened 2023-12-21
take into account that binutils might be a filtered dependency for TensorFlow

Flamefire:
E.g. when using `--filter-deps binutils` in combination with `--rpath` the could could fail to determine the files from binutils to collect as the "software root" won't be set.
Fall back to collecting all rpath wrappers and a list of known binutils

Followup to #3054 similar to https://github.com/easybuilders/easybuild-easyblocks/pull/2218

Logic is:
- If binutils root is found use those found in its "bin" folder
- else collect all wrappers as they mostly are binutils and the others don't hurt as we don't put that new folder into PATH anyway. Additionally add a hard-coded list of known binutils which could be extended in case we observe failures due to other missing tools where no rpath wrapper is created. e.g. "ar"

While at it I double-checked the other uses of `which` and used the new list of binutils in `patch_crosstool_files` preserving the existing behavior of raising an error if one of those we currently have isn't found.

Would be great if @jfgrimm could verify this new version still works but I assume it doesn't change behavior for this case as it should basically do exactly the same.

cc @boegel

### DIFF
--- a/easybuild/easyblocks/t/tensorflow.py
+++ b/easybuild/easyblocks/t/tensorflow.py
@@ -489,7 +489,7 @@ class EB_TensorFlow(PythonPackage):
                 mkdir(new_rpath_wrapper_dir)
                 for file in binutils_files:
                     # use `which` to take rpath wrappers where available
-                    # Ignore missing ones if bintuils was filtered (in which case we used a heuristic)
+                    # Ignore missing ones if binutils was filtered (in which case we used a heuristic)
                     path = which(file, on_error=ERROR if binutils_root else WARN)
                     if path:
                         symlink(path, os.path.join(new_rpath_wrapper_dir, file))
--- a/easybuild/easyblocks/t/tensorflow.py
+++ b/easybuild/easyblocks/t/tensorflow.py
@@ -35,6 +35,7 @@ import os
 import re
 import stat
 import tempfile
+from itertools import chain
 
 import easybuild.tools.environment as env
 import easybuild.tools.toolchain as toolchain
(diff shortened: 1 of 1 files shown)


jfgrimm:
I've kicked off a build

jfgrimm:
Test report by @jfgrimm
 * **SUCCESS** _TensorFlow-2.11.0-foss-2022a-CUDA-11.7.0.eb_ 

gpu14.viking2.yor.alces.network - Linux Rocky Linux 8.8, x86_64, AMD EPYC 7643 48-Core Processor, 2 x NVIDIA NVIDIA A40, 535.86.10, Python 3.6.8
See https://gist.github.com/jfgrimm/db8ea3ee916ca672f89a1be1ed943bc8 for a full test report.

boegel (changes requested):

boegel:
Test report by @boegel
 * **FAIL (build issue)** _TensorFlow-2.13.0-foss-2023a.eb_ (partial log available at https://gist.github.com/boegel/c6cb2c168e7ed9886f468b07cec69d7e)

node4112.gallade.os - Linux Debian GNU/Linux 11 (bullseye), x86_64, AMD EPYC 7773X 64-Core Processor, Python 3.11.4
See https://gist.github.com/boegel/57ea446ff397985715780e8db85dc200 for a full test report.

boegel:
Successful test in EESSI build environment, see test report at https://github.com/easybuilders/easybuild-easyconfigs/pull/19268#issuecomment-1866487630

boegel:
Test report by @boegel
 * **SUCCESS** _TensorFlow-2.13.0-foss-2023a.eb_ 

node3109.skitty.os - Linux RHEL 8.8, x86_64, Intel(R) Xeon(R) Gold 6140 CPU @ 2.30GHz (skylake_avx512), Python 3.6.8
See https://gist.github.com/boegel/6a3d23c03c6a44929479642e8ed3348b for a full test report.

boegel:
Tested by @jfgrimm with RPATH enabled, tested in EESSI build environment, tested without RPATH or EESSI, so good to go, thanks @Flamefire!

boegel (approved):

PR #3072, opened 2024-01-10
pick Tcl/Tk lib directory correctly in R easyblock, may be either `lib` or `lib64`

mboisson:
tcl/tk is not always in lib folder, on gentoo it is in lib64, this commit tests it and picks the right one


boegel:
EB_ARGS="--installpath /tmp/pr3072 R-4.3.2-gfbf-2023a.eb"

boegel (approved):
lgtm

PR #3073, opened 2024-01-12
use `(=|:=)` in regex because OpenFOAM rule files use `:=` instead of `=` since v2212

mboisson:


boegel (approved):
lgtm after visual review

boegel:
@mboisson Can you clarify in the PR description how problems manifest if OpenFOAM v2112 was installed without this fix?

boegel:
EB_ARGS="OpenFOAM-v2306-foss-2022b.eb OpenFOAM-v2206-foss-2022a.eb OpenFOAM-11-foss-2023a.eb"
CORE_CNT=16

mboisson:
> @mboisson Can you clarify in the PR description how problems manifest if OpenFOAM v2212 was installed without this fix (which errors, etc.)?

The way it manifested for us is that `redistributePar` and a few other binaries are not compiled with the MPI compiler, which in our case resulted in `libmpi.so.4` not being found (it is coming from another library) because the proper RPATH was not injected in the executable. This is not an issue for sites that use `LD_LIBRARY_PATH` because the shared object is then found through that. For these sites, the issue is probably more subtle and it probably manifests itself by the wrong flags being used. 

A way to detect that would probably to test whether the regex actually applied something. In our case, for example, not fixed (avx2) 
```
[mboisson@build-node ~]$ diff  /cvmfs/soft.computecanada.ca/easybuild/software/2020/avx2/MPI/gcc9/openmpi4/openfoam/v2212/OpenFOAM-v2212/wmake/rules/General/Gcc/c++{,.orig.eb}
``` 
(returns nothing), vs fixed (avx512): 
``` 
[mboisson@build-node ~]$ diff  /cvmfs/soft.computecanada.ca/easybuild/software/2020/avx512/MPI/gcc9/openmpi4/openfoam/v2212/OpenFOAM-v2212/wmake/rules/General/Gcc/c++{,.orig.eb}
9c9
< CC         := OMPI_CXX="g++" mpicxx
---
> CC         := g++$(COMPILER_VERSION) -std=c++11
13c13
< c++OPT     := -O2 -fno-tree-vectorize -march=skylake-avx512 -fno-math-errno -std=c++14 -fuse-ld=bfd
---
> c++OPT     := -O2
```

boegelbot (approved):
lgtm

PR #3077, opened 2024-01-16
Support GROMACS builds with SYCL (AdaptiveCpp implementation)

MaximeVdB:
This expands the `gromacs` easyblock to allow GPU-enabled GROMACS builds using the SYCL backend via the [AdaptiveCpp](https://adaptivecpp.github.io) (formerly known as hipSYCL / Open SYCL) implementation of SYCL. I will soon open another PR for easyconfigs where this easyblock is used in a GROMACS-2023.3 build for AMD GPUs.

### DIFF
--- a/easybuild/easyblocks/g/gromacs.py
+++ b/easybuild/easyblocks/g/gromacs.py
@@ -62,7 +62,7 @@ class EB_GROMACS(CMakeMake):
         extra_vars = CMakeMake.extra_options()
         extra_vars.update({
             'double_precision': [None, "Build with double precision enabled (-DGMX_DOUBLE=ON), " +
-                                 "default is to build double precision unless CUDA is enabled", CUSTOM],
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/g/gromacs.py
```diff
@@ -198,9 +207,15 @@ def configure_step(self):
                     cuda_cc_semicolon_sep = self.cfg.get_cuda_cc_template_value(
                         "cuda_cc_semicolon_sep").replace('.', '')
                     self.cfg.update('configopts', '-DGMX_CUDA_TARGET_SM="%s"' % cuda_cc_semicolon_sep)
+
+            elif acpp:
+                assert LooseVersion(self.version) >= LooseVersion('2023'), \
+                       'SYCL support is not available for this GROMACS version'
```
akesandgren, line 213: @boegel Do we want to use assert in easyblocks like this?

PR #3022, opened 2023-10-23
enable `download_dep_fail`, `use_pip`, `sanity_pip_check` by default in `PythonPackage` easyblock

boegel:
Probably worth doing a couple of test builds for the easyblocks that are touched, to avoid surprises.

Easyconfigs which are currently enabling these easyconfig parameters should be cleaned up.

Non-archived easyconifgs which are *not* enabling these easyconfig parameters yet should be updated to set them to `False` (unless using `pip` and running "`pip check`" works fine for them).

fixes #2127

PR #3079, opened 2024-01-17
disable `use_pip` by default for PyTorch, except for recent versions (>= 2.1)

boegel:
Up until now, we've only enabled `use_pip` for the latest versions of `PyTorch` (>= 2.1).

It's probably not worth the effort to also do that for older PyTorch versions, so we need to overrule the changing default in `PythonPackage` (cfr. https://github.com/easybuilders/easybuild-easyblocks/pull/3022)

With this approach, old `PyTorch` easyconfigs could explicitly set `use_pip = False`, and future PyTorch easyconfigs could opt-in to not using `pip` by using `use_pip = False`, should the need arise (since PyTorch currently still uses the legacy install procedure that involves `setup.py build`).

### DIFF
--- a/easybuild/easyblocks/p/pytorch.py
+++ b/easybuild/easyblocks/p/pytorch.py
@@ -54,8 +54,14 @@ class EB_PyTorch(PythonPackage):
             'excluded_tests': [{}, "Mapping of architecture strings to list of tests to be excluded", CUSTOM],
             'max_failed_tests': [0, "Maximum number of failing tests", CUSTOM],
         })
+
+        # disable use of pip to install PyTorch by default, overwriting the default set in PythonPackage;
+        # see also https://github.com/easybuilders/easybuild-easyblocks/pull/3022
+        extra_vars['use_pip'][0] = None
+
         extra_vars['download_dep_fail'][0] = True
         extra_vars['sanity_pip_check'][0] = True
+
         # Make pip show output of build process as that may often contain errors or important warnings
         extra_vars['pip_verbose'][0] = True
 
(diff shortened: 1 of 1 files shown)


Flamefire:
> (since PyTorch currently still uses the legacy install procedure that involves `setup.py build`).

Just to clarify: `pip install .` for PyTorch executes the equivalent (at least) of `setup.py build` due to the use of the legacy `build-backend` in `pyproject.toml`: https://github.com/pytorch/pytorch/blob/40a6710ad34ae4c6f4987f0e47b5c94df3fc8ec7/pyproject.toml#L14

Nothing wrong with that. AFAIK this means our use of `setup.py install` and `pip install .` are equivalent (for now, until they change the pyproject.toml)

I would even use 2.x as the cutoff version to simplify this to "PyTorch 2 uses pip by default", no need to go into minor versions here.

boegel:
EB_BRANCH=5.0.x
EB_ARGS="PyTorch-2.0.1-foss-2022a.eb PyTorch-2.0.1-foss-2022b.eb PyTorch-2.1.2-foss-2022a.eb PyTorch-2.1.2-foss-2022b.eb PyTorch-2.1.2-foss-2023a.eb"
CORE_CNT=16

boegel:
We haven't had test reports for `PyTorch` on `jsc-zen3` yet (only on `jsc-zen2` which is being discontinued).

For `PyTorch-2.0.1-foss-2022a.eb`:
```
== 2024-01-19 20:45:24,371 pytorch.py:480 WARNING 0 test failures, 3 test errors (out of 180622):
distributed/_tensor/test_device_mesh (24 total tests, errors=3)
```

For `PyTorch-2.0.1-foss-2022b.eb`:
```
== 2024-01-20 03:39:30,018 pytorch.py:480 WARNING 2 test failures, 6 test errors (out of 180621):
test_quantization (997 total tests, failures=2, errors=1, skipped=74)
distributed/_tensor/test_device_mesh (24 total tests, errors=4)
distributed/_tensor/test_dtensor (15 total tests, errors=1)
```

I'm pretty sure that they are not caused by enabling `use_pip`, and hence shouldn't block this PR.

@Flamefire Thoughts?

Flamefire:
- test_quantization is kind of expected -> See comment in EC (test may randomly generate inputs where it fails)
- distributed/_tensor/test_device_mesh -> Only seen this for CUDA builds (when NCCL is included)
- distributed/_tensor/test_dtensor -> New to me

Generally those are test errors, not failures. So very likely the test run into something it didn't expect from the environment. E.g for the CUDA build I had it because PyTorch was built with NCCL but run without GPUs. But can't tell from this alone

boegel:
@Flamefire There's indeed no GPU available in this test setup.

In any case, you don't see any reason to block this PR, right?

Flamefire:
> @Flamefire There's indeed no GPU available in this test setup.
> 
> In any case, you don't see any reason to block this PR, right?

That was just an example that applies to the in-progress CUDA ECs. But yes, I agree that this shouldn't block this PR. Might be worth investigating the failures later though.

branfosj (approved):

PR #3082, opened 2024-01-18
4 Conda Fixes to allow Conda.py  to work with Micromamba

dirkpetersen:
These 4 changes will allow EB to use Micromamba as conda /  mamba replacement which is much faster

- Check for existence of `micromamba` first because `mamba` is sometimes used as alias for micromamba but typically not vice versa 
- remove the --force option from command line because Micromamba does not support it. I have requested that they implement it . 
- do not bootstrap config --add create_default_packages setuptools with Micromamba, there is no Python and it does not like setuptools 
- delete the empty installdir os.rmdir(self.installdir) as Micromamba likes to create that itself and complains if it is already there.

### DIFF
--- a/easybuild/easyblocks/generic/conda.py
+++ b/easybuild/easyblocks/generic/conda.py
@@ -64,17 +64,25 @@ class Conda(Binary):
         if (get_software_root('anaconda2') or get_software_root('miniconda2') or
                 get_software_root('anaconda3') or get_software_root('miniconda3')):
             conda_cmd = 'conda'
-        elif get_software_root('mamba'):
-            conda_cmd = 'mamba'
         elif get_software_root('micromamba'):
             conda_cmd = 'micromamba'
+        elif get_software_root('mamba'):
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/generic/conda.py
```diff
@@ -64,17 +64,25 @@ def install_step(self):
...
-        # initialize conda environment
+        # initialize conda environment, except for micromamba
         # setuptools is just a choice, but *something* needs to be there
-        cmd = "%s config --add create_default_packages setuptools" % conda_cmd
-        run_cmd(cmd, log_all=True, simple=True)
+        force = '--force'
+        if conda_cmd == 'micromamba':
+            force = ''
+            try:
+                os.rmdir(self.installdir)
```
smoors, line 80: better use `easybuild.tools.filetools.remove_dir`, which already takes care of the try-except for you.

boegel:
No follow up since March 2024, so closing...

PR #3086, opened 2024-01-22
remove trailing slash in `PETSC_DIR` set by custom easyblock for PETSc

furstj:
Compilation of PETSc 3.20.3 fails due to wrong format of PETSC_DIR. The PETSC_DIR variable should not contain the trailing slash (see configuration example in https://petsc.org/release/install/install_tutorial/#configuration). 

Older versions of PETSc are probably less sensitive to this error, so I was able to compile PETSc 3.19 both with original easyblock (i.e. with trailing slash) and with the new one (without the slash).

### DIFF
--- a/easybuild/easyblocks/p/petsc.py
+++ b/easybuild/easyblocks/p/petsc.py
@@ -275,8 +275,9 @@ class EB_PETSc(ConfigureMake):
                 self.cfg.update('configopts', ' '.join([withdep + spec for spec in ['=1', inc_spec, lib_spec]]))
 
             # set PETSC_DIR for configure (env) and build_step
-            env.setvar('PETSC_DIR', self.cfg['start_dir'])
-            self.cfg.update('buildopts', 'PETSC_DIR=%s' % self.cfg['start_dir'])
+            petsc_dir = os.path.abspath(self.cfg['start_dir'])
+            env.setvar('PETSC_DIR', petsc_dir)
+            self.cfg.update('buildopts', 'PETSC_DIR=%s' % petsc_dir)
 
             if self.cfg['sourceinstall']:
                 if self.petsc_arch:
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/p/petsc.py
```diff
@@ -275,8 +279,9 @@ def configure_step(self):
                 self.cfg.update('configopts', ' '.join([withdep + spec for spec in ['=1', inc_spec, lib_spec]]))
 
             # set PETSC_DIR for configure (env) and build_step
-            env.setvar('PETSC_DIR', self.cfg['start_dir'])
-            self.cfg.update('buildopts', 'PETSC_DIR=%s' % self.cfg['start_dir'])
+            petsc_dir = os.path.abspath(self.cfg['start_dir'])
```
boegel, line 282: If this is really only about the trailing slash, maybe we should use `.rstrip(os.path.sep)` instead?

```python
>>> import os
>>> os.path.sep
'/'
>>> path_with_slash = '/tmp/example/'
>>> path_with_slash.rstrip(os.path.sep)
'/tmp/example'
```

Using `abspath` will resolve potential symbolic links , which we should avoid I think...

boegel (changes requested):

easybuild/easyblocks/p/petsc.py
furstj: Yes, the problem is only the trailing slash. So .rstrip is better.

boegel (approved):
lgtm

boegel:
Test report by @boegel
 * **SUCCESS** _PETSc-3.17.4-foss-2022a.eb_ 
 * **SUCCESS** _PETSc-3.19.2-foss-2022b.eb_ 

node3147.skitty.os - Linux RHEL 8.8, x86_64, Intel(R) Xeon(R) Gold 6140 CPU @ 2.30GHz, Python 3.6.8
See https://gist.github.com/boegel/e59442a3c031d5312dd61b2aad401b28 for a full test report.

PR #3147, opened 2024-02-03
replace `run_cmd` with `run_shell_cmd` in custom easyblock for Rosetta (`rosetta.py`)

branfosj:
(created using `eb --new-pr`)

PR #3148, opened 2024-02-03
remove unused custom easyblock for Rosetta (`rosetta.py`)

branfosj:
(created using `eb --new-pr`)


boegel:
@branfosj #2744 suggests that The Alliance is still using this easyblock, so maybe we should make the effort on porting this to `run_shell_cmd`, and ask @mboisson to test?

branfosj:
> @branfosj #2744 suggests that The Alliance is still using this easyblock, so maybe we should make the effort on porting this to `run_shell_cmd`, and ask @mboisson to test?

Reverting to #3147 - I just noticed that BEAR make use of this.

PR #3151, opened 2024-02-03
remove unused custom easyblock for CFDEMcoupling (`cfdemcoupling.py`)

branfosj:
(created using `eb --new-pr`)


jfgrimm (approved):
lgtm, only used in easyconfigs from the 2018a and 2019b toolchains

PR #3211, opened 2024-02-10
remove unused custom easyblock for Mono (`mono.py`)

branfosj:
(created using `eb --new-pr`)

* only easyconfig (`Mono-6.12.0.122-GCCcore-11.2.0.eb`) does not use the custom easyblock (it uses `ConfigureMake` directly)


bartoldeman (approved):
LGTM

PR #3233, opened 2024-02-27
enhance `CMakeMake` easyblock to check whether correct Python installation was picked up by `cmake`

MartinsNadia:
Sanity check to verify if Cmake apply the new policy CMP0094

### DIFF
--- a/easybuild/easyblocks/generic/cmakemake.py
+++ b/easybuild/easyblocks/generic/cmakemake.py
@@ -258,7 +258,7 @@ class CMakeMake(ConfigureMake):
                 'CXXFLAGS': 'CMAKE_CXX_FLAGS',
                 'FFLAGS': 'CMAKE_Fortran_FLAGS',
             })
-
+#teste
         for env_name, option in env_to_options.items():
             value = os.getenv(env_name)
             if value is not None:
--- a/easybuild/easyblocks/generic/cmakemake.py
+++ b/easybuild/easyblocks/generic/cmakemake.py
@@ -331,18 +331,16 @@ class CMakeMake(ConfigureMake):
                 pythonIncludePath = line.split('=')[1].strip()
             elif line.startswith('Python3_LIBRARY:FILEPATH='):
                 pythonLibraryPath = line.split('=')[1].strip()
-        self.log.info("Python executable path: %s",pythonExecPath)
-        self.log.info("Python include path: %s",pythonIncludePath)
-        self.log.info("Python library path: %s",pythonLibraryPath)
- 
+        self.log.info("Python executable path: %s", pythonExecPath)
+        self.log.info("Python include path: %s", pythonIncludePath)
+        self.log.info("Python library path: %s", pythonLibraryPath)
         # Check if paths include EBROOTPYTHON
         ebrootpython_path = os.getenv('EBROOTPYTHON')
-        if all(path and ebrootpython_path in path for path in [pythonExecPath,pythonIncludePath,pythonLibraryPath]):
+        if all(path and ebrootpython_path in path for path in [pythonExecPath, pythonIncludePath, pythonLibraryPath]):
             self.log.info("Python related paths configured correctly.")
             return out
         else:
             raise EasyBuildError("Python related paths do not include EBROOTPYTHON, check log")
-            
     def test_step(self):
         """CMake specific test setup"""
         # When using ctest for tests (default) then show verbose output if a test fails
--- a/easybuild/easyblocks/generic/cmakemake.py
+++ b/easybuild/easyblocks/generic/cmakemake.py
@@ -357,8 +357,31 @@ class CMakeMake(ConfigureMake):
 
         (out, _) = run_cmd(command, log_all=True, simple=False)
 
-        return out
+def check_python_paths(file_path):
+    with open(file_path, 'r') as file:
+        lines = file.readlines()
 
+    # Search for Python paths in each line
+    for line in lines:
+        if line.startswith('Python3_EXECUTABLE:FILEPATH='):
+            exec_path = line.split('=')[1].strip()
+        elif line.startswith('Python3_INCLUDE_DIR:FILEPATH='):
+            include_path = line.split('=')[1].strip()
+        elif line.startswith('Python3_LIBRARY:FILEPATH='):
+            library_path = line.split('=')[1].strip()
+
+    # Check if paths include EBROOTPYTHON
+    ebrootpython_path = "/cvmfs/software.eessi.io/versions/2023.06/software/linux/x86_64/intel/skylake_avx512/software/Python/3.11.3-GCCcore-12.3.0"
+    if all(path and ebrootpython_path in path for path in [exec_path, include_path, library_path]):
+        print("Paths include EBROOTPYTHON.")
+    else:
+        print("Paths do not include EBROOTPYTHON.")
+
+    cmake_cache_path = 'CMakeCache.txt'
+    check_python_paths(cmake_cache_path)
+
+    return out
+ 
     def test_step(self):
         """CMake specific test setup"""
         # When using ctest for tests (default) then show verbose output if a test fails
--- a/easybuild/easyblocks/generic/cmakemake.py
+++ b/easybuild/easyblocks/generic/cmakemake.py
@@ -49,7 +49,6 @@ from easybuild.tools.run import run_cmd
 from easybuild.tools.systemtools import get_shared_lib_ext
 from easybuild.tools.utilities import nub
 
-print ("Using Nadia EasyBlock") 
 DEFAULT_CONFIGURE_CMD = 'cmake'
 
 def det_cmake_version():
@@ -337,7 +336,7 @@ class CMakeMake(ConfigureMake):
             "include": [],
             "library": [],
         }
-        errors = [] # List to store all encountered errors
+        errors = []
 
         for line in lines:
             match = re.match(r"(?i)_Python(\d+)_((?:EXECUTABLE|INCLUDE_DIR|LIBRARY)):.*?=(.*)", line)
--- a/easybuild/easyblocks/generic/cmakemake.py
+++ b/easybuild/easyblocks/generic/cmakemake.py
@@ -322,67 +322,67 @@ class CMakeMake(ConfigureMake):
     def sanitycheck_for_configuration(self, out):
         self.log.info("Checking Python paths")
     
-        if LooseVersion(self.cmake_version) >= '3.16'
-        try:
-         with open('CMakeCache.txt', 'r') as file:
-            lines = file.readlines()
-        # The case where CMakeCache.txt is not found 
-        except FileNotFoundError:
-            self.log.warning("CMakeCache.txt not founf. Python paths checks skipped.")
-            return 
-        # EBROOTPYTHON check 
-        if not os.getenv('EBROOTPYTHON'):
-            self.log.warning("EBROOTPYTHON is not set.")
-            return 
-        else: 
-            self.log.info ("Skipping Python path checks as CMake version might be lower than 3.16.")
-            return
+        if LooseVersion(self.cmake_version) >= '3.16':
+            try:
+                with open('CMakeCache.txt', 'r') as file:
+                    lines = file.readlines()
+            except FileNotFoundError:
+                self.log.warning("CMakeCache.txt not found. Python paths checks skipped.")
+                return 
+     
+            if not os.getenv('EBROOTPYTHON'):
+                self.log.warning("EBROOTPYTHON is not set.")
+                return 
+            else: 
+                self.log.info("Skipping Python path checks as CMake version might be lower than 3.16.")
+                return
 
-    # Initialize empty lists to store paths 
-    python_exec_path = []
-    python_include_path = []
-    python_library_path = []
+        python_exec_path = []
+        python_include_path = []
+        python_library_path = []
 
-    # Define a regular expression to match all relevant prefixes
-    python_prefixes = r"(Python|Python2|Python3)_"
+        # Define a regular expression to match all relevant prefixes
+        python_prefixes = r"(Python|Python2|Python3)_"
 
-    for line in lines:
-        if line.startswith(python_prefixes + r"EXECUTABLE:FILEPATH="):
-          python_exec_paths.append(line.split('=')[1].strip())
-          self.log.info("Python executable path: %s", python_exec_path[-1])
-        elif line.startswith(python_prefixes + r"INCLUDE_DIR:FILEPATH="):
-          python_include_paths.append(line.split('=')[1].strip())
-          self.log.info("Python include path: %s", python_include_path[-1])
-        elif line.startswith(python_prefixes + r"LIBRARY:FILEPATH="):
-          python_library_paths.append(line.split('=')[1].strip())
-          self.log.info("Python library path: %s", python_library_path[-1])
+        for line in lines:
+            if line.startswith(python_prefixes + r"EXECUTABLE:FILEPATH="):
+                python_exec_path.append(line.split('=')[1].strip())
+                self.log.info("Python executable path: %s", python_exec_path[-1])
+            elif line.startswith(python_prefixes + r"INCLUDE_DIR:FILEPATH="):
+                python_include_path.append(line.split('=')[1].strip())
+                self.log.info("Python include path: %s", python_include_path[-1])
+            elif line.startswith(python_prefixes + r"LIBRARY:FILEPATH="):
+                python_library_path.append(line.split('=')[1].strip())
+                self.log.info("Python library path: %s", python_library_path[-1])
     
-    # Check if paths are found and validate paths for symlinks
-    for path_list in [python_exec_path, python_include_path, python_library_path]: 
-        for path in path_list:
-            if path and not os.path.exists(path): 
-                raise EasyBuildError(f"Path does not exist: {path}")
+            # Check if paths are found and validate paths for symlinks
+            for path_list in [python_exec_path, python_include_path, python_library_path]: 
+                for path in path_list:
+                    if path and not os.path.exists(path): 
+                        raise EasyBuildError(f"Path does not exist: {path}")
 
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/generic/cmakemake.py
```diff
@@ -318,8 +318,29 @@ def configure_step(self, srcdir=None, builddir=None):
...
+            elif line.startswith('Python3_LIBRARY:FILEPATH='):
+                pythonLibraryPath = line.split('=')[1].strip()
+        self.log.info("Python executable path: %s", pythonExecPath)
+        self.log.info("Python include path: %s", pythonIncludePath)
+        self.log.info("Python library path: %s", pythonLibraryPath)
+        # Check if paths include EBROOTPYTHON
+        ebrootpython_path = os.getenv('EBROOTPYTHON')
+        if all(path and ebrootpython_path in path for path in [pythonExecPath, pythonIncludePath, pythonLibraryPath]):
+            self.log.info("Python related paths configured correctly.")
+            return out
```
ocaisa, line 341: Move the return statement out of the `if/else` block, it is hidden here.

easybuild/easyblocks/generic/cmakemake.py
```diff
@@ -318,8 +318,29 @@ def configure_step(self, srcdir=None, builddir=None):
...
+        for line in lines:
+            if line.startswith('Python3_EXECUTABLE:FILEPATH='):
+                pythonExecPath = line.split('=')[1].strip()
+            elif line.startswith('Python3_INCLUDE_DIR:FILEPATH='):
+                pythonIncludePath = line.split('=')[1].strip()
+            elif line.startswith('Python3_LIBRARY:FILEPATH='):
+                pythonLibraryPath = line.split('=')[1].strip()
+        self.log.info("Python executable path: %s", pythonExecPath)
+        self.log.info("Python include path: %s", pythonIncludePath)
+        self.log.info("Python library path: %s", pythonLibraryPath)
```
ocaisa, line 336: I would only print these if they have a value

easybuild/easyblocks/generic/cmakemake.py
```diff
@@ -318,8 +318,29 @@ def configure_step(self, srcdir=None, builddir=None):
...
-        return out
-
+        # sanitycheck for the configuration step
+        self.log.info("Checking python paths")
+        with open('CMakeCache.txt', 'r') as file:
+            lines = file.readlines()
+
+        # Search for Python paths in each line
+        for line in lines:
+            if line.startswith('Python3_EXECUTABLE:FILEPATH='):
```
ocaisa, line 328: It looks like these are only guaranteed for CMake 3.16+ so should probably make this check conditional on that version

ocaisa: They also don't cover the FindPython case: https://cmake.org/cmake/help/latest/module/FindPython.html#artifacts-specification

easybuild/easyblocks/generic/cmakemake.py
```diff
@@ -318,8 +318,29 @@ def configure_step(self, srcdir=None, builddir=None):
...
+                pythonExecPath = line.split('=')[1].strip()
+            elif line.startswith('Python3_INCLUDE_DIR:FILEPATH='):
+                pythonIncludePath = line.split('=')[1].strip()
+            elif line.startswith('Python3_LIBRARY:FILEPATH='):
+                pythonLibraryPath = line.split('=')[1].strip()
+        self.log.info("Python executable path: %s", pythonExecPath)
+        self.log.info("Python include path: %s", pythonIncludePath)
+        self.log.info("Python library path: %s", pythonLibraryPath)
+        # Check if paths include EBROOTPYTHON
+        ebrootpython_path = os.getenv('EBROOTPYTHON')
```
ocaisa, line 338:
```suggestion
        ebrootpython_path = get_software_root("Python")
```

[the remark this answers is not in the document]
ocaisa: I imagine that what you want to do is complain if `pythonExecPath` (and friends) are set but no Python dependency is found

easybuild/easyblocks/generic/cmakemake.py
```diff
@@ -318,8 +318,29 @@ def configure_step(self, srcdir=None, builddir=None):
...
-
+        # sanitycheck for the configuration step
+        self.log.info("Checking python paths")
+        with open('CMakeCache.txt', 'r') as file:
+            lines = file.readlines()
+
+        # Search for Python paths in each line
+        for line in lines:
+            if line.startswith('Python3_EXECUTABLE:FILEPATH='):
+                pythonExecPath = line.split('=')[1].strip()
```
boegel, line 329: Small nitpicking suggestion: let's not start using CamelCase for variable names (we only use that for class names), so please use `python_exec_path`, `python_include_path`, `python_library_path`

boegel (changes requested):

easybuild/easyblocks/generic/cmakemake.py
```diff
@@ -318,8 +318,29 @@ def configure_step(self, srcdir=None, builddir=None):
 
         (out, _) = run_cmd(command, log_all=True, simple=False)
 
-        return out
-
+        # sanitycheck for the configuration step
+        self.log.info("Checking python paths")
+        with open('CMakeCache.txt', 'r') as file:
```
Flamefire, line 323: This should handle the case that this file may not exist. E.g. when `configure_cmd != DEFAULT_CONFIGURE_CMD` anything can happen.

easybuild/easyblocks/generic/cmakemake.py
```diff
@@ -318,8 +318,29 @@ def configure_step(self, srcdir=None, builddir=None):
...
+            lines = file.readlines()
```
Flamefire, line 324: Maybe use `readfile(...).split('\n')`?

easybuild/easyblocks/generic/cmakemake.py
Flamefire: Yes maybe use a regex? `Python[2-3]?_EXECUTABLE`?

easybuild/easyblocks/generic/cmakemake.py
Flamefire: Not only that: You could have them multiple times when you consider `Python_EXECUTABLE` and `Python3_EXECUTABLE` or the variables may be completely unset if the CMake file didn't use Python at all

Maybe create thos variables as empty lists?

easybuild/easyblocks/generic/cmakemake.py
```diff
@@ -318,8 +318,29 @@ def configure_step(self, srcdir=None, builddir=None):
...
+            elif line.startswith('Python3_INCLUDE_DIR:FILEPATH='):
+                pythonIncludePath = line.split('=')[1].strip()
+            elif line.startswith('Python3_LIBRARY:FILEPATH='):
+                pythonLibraryPath = line.split('=')[1].strip()
+        self.log.info("Python executable path: %s", pythonExecPath)
+        self.log.info("Python include path: %s", pythonIncludePath)
+        self.log.info("Python library path: %s", pythonLibraryPath)
+        # Check if paths include EBROOTPYTHON
+        ebrootpython_path = os.getenv('EBROOTPYTHON')
+        if all(path and ebrootpython_path in path for path in [pythonExecPath, pythonIncludePath, pythonLibraryPath]):
```
Flamefire, line 339: See above: `pythonExecPath` & co might not have been set at all. And maybe `path.startswith(ebrootpython_path)`? We possibly need to take symlinks into account too: either of the 2 might be a resolved symlink or not.

Flamefire (changes requested):
This might be a bit more tricky than it seems...

easybuild/easyblocks/generic/cmakemake.py
```diff
@@ -318,7 +319,70 @@ def configure_step(self, srcdir=None, builddir=None):
...
+    
+        if LooseVersion(self.cmake_version) >= '3.16'
+        try:
+         with open('CMakeCache.txt', 'r') as file:
+            lines = file.readlines()
+        # The case where CMakeCache.txt is not found 
+        except FileNotFoundError:
+            self.log.warning("CMakeCache.txt not founf. Python paths checks skipped.")
+            return 
+        # EBROOTPYTHON check
```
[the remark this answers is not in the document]
Flamefire: And please remove this comment. It doesn't add any information to the code below

easybuild/easyblocks/generic/cmakemake.py
```diff
@@ -318,6 +319,73 @@ def configure_step(self, srcdir=None, builddir=None):
...
+
+            # Check if paths are found and handle EBROOTPYTHON cases
+            if any(path for path in [python_exec_path, python_include_path, python_library_path]):
+                if not os.getenv('EBROOTPYTHON'):
+                    self.log.error(
+                        "EBROOTPYTHON is not set and Python related paths are found."
+                    )
+                elif not all(
+                    (path and os.getenv('EBROOTPYTHON') in path)
+                    for path in [python_exec_path, python_include_path, python_library_path]
```
Flamefire, line 371: You are checking only for `$EBROOTPYTHON` being in a list of paths, which likely is always false. Maybe rename the variables to be able to spot that: `python_exec_path` -> `python_exec_paths` and so on.

It might even make sense to show which path is "wrong". I.e. which of the 3 lists and which exact path is outside EB.

Flamefire (changes requested):
Please recheck also the use(s) of `self.log.error`. When raising an `EasyBuildError` you don't need to log the same because that will already be done.   
I also thing that you don't need `self.log.error` but only `EasyBuildError`, i.e. fail if there is an error.

Another enhancement would be to gather all issues (e.g. each wrong path) in a list and then fail if that list is non-empty. This way ALL issues would be reported, not only the first found which might help in fixing this.

easybuild/easyblocks/generic/cmakemake.py
```diff
@@ -317,9 +317,60 @@ def configure_step(self, srcdir=None, builddir=None):
                 self.cfg['configopts']])
 
         (out, _) = run_cmd(command, log_all=True, simple=False)
+        self.check_python_paths()
 
         return out
 
+    def check_python_paths(self):
+        """Check that there are no detected Python paths outside the EB installed PythonF"""
```
boegel, line 325:
```suggestion
        """Check that there are no detected Python paths outside the Python dependency provided by EasyBuild"""
```

easybuild/easyblocks/generic/cmakemake.py
```diff
@@ -317,9 +317,60 @@ def configure_step(self, srcdir=None, builddir=None):
...
+        if not os.path.exists('CMakeCache.txt'):
+            self.log.warning("CMakeCache.txt not found. Python paths checks skipped.")
+            return
+        cmake_cache = read_file('CMakeCache.txt')
+        if not cmake_cache:
```
boegel, line 330: Hmm, I don't see how this can happen, since at this point you know that the file exists?

easybuild/easyblocks/generic/cmakemake.py
```diff
@@ -317,9 +317,60 @@ def configure_step(self, srcdir=None, builddir=None):
...
+            return
+
+        self.log.info("Checking Python paths")
+
+        python_paths = {
+            "executable": [],
+            "include_dir": [],
+            "library": [],
+        }
+        PYTHON_RE = re.compile(r"_?(?:Python|PYTHON)\d_(EXECUTABLE|INCLUDE_DIR|LIBRARY)[^=]*=(.*)")
```
boegel, line 341: Don't use "constants" inline:
```suggestion
        python_regex = re.compile(r"_?(?:Python|PYTHON)\d_(EXECUTABLE|INCLUDE_DIR|LIBRARY)[^=]*=(.*)")
```

easybuild/easyblocks/generic/cmakemake.py
```diff
@@ -317,9 +317,60 @@ def configure_step(self, srcdir=None, builddir=None):
...
+        for line in cmake_cache.splitlines():
+            match = PYTHON_RE.match(line)
```
boegel, line 343:
```suggestion
            match = python_regex.match(line)

File: easybuild/easyblocks/generic/cmakemake.py [hunk 15]
```diff
@@ -317,9 +317,60 @@ def configure_step(self, srcdir=None, builddir=None):
...
+            if match:
+                path_type = match[1].lower()
+                path = match[2].strip()
+                self.log.info(f"Python {path_type} path: {path}")
+                python_paths[path_type].append(path)
+
+        ebrootpython_path = get_software_root("Python")
+        if not ebrootpython_path:
+            if any(python_paths.values()) and not self.toolchain.comp_family() == toolchain.SYSTEM:
+                self.log.warning("Found Python paths in CMake cache but Python is not a dependency")
```
boegel, line 353: I wouldn't make this a warning, just info message?
```suggestion
                self.log.info("Found Python paths in CMake cache but Python is not a dependency")

File: easybuild/easyblocks/generic/cmakemake.py [hunk 16]
```diff
@@ -317,9 +317,60 @@ def configure_step(self, srcdir=None, builddir=None):
...
+            return
+        ebrootpython_path = os.path.realpath(ebrootpython_path)
+
+        errors = []
+        for path_type, paths in python_paths.items():
+            for path in paths:
+                if path.endswith('-NOTFOUND'):
+                    continue
+                if not os.path.exists(path):
+                    errors.append(f"Python {path_type} does not exist: {path}")
```
boegel, line 364: Since this is a PR to `develop`, we need to be careful with f-strings, since EasyBuild 4.x is still supposed to support Python 2.7:
```suggestion
                    errors.append("Python %s does not exist: %s" % (path_type, path))
```

easybuild/easyblocks/generic/cmakemake.py
```diff
@@ -317,9 +317,60 @@ def configure_step(self, srcdir=None, builddir=None):
...
+            "include_dir": [],
+            "library": [],
+        }
+        PYTHON_RE = re.compile(r"_?(?:Python|PYTHON)\d_(EXECUTABLE|INCLUDE_DIR|LIBRARY)[^=]*=(.*)")
+        for line in cmake_cache.splitlines():
+            match = PYTHON_RE.match(line)
+            if match:
+                path_type = match[1].lower()
+                path = match[2].strip()
+                self.log.info(f"Python {path_type} path: {path}")
```
boegel, line 347: Since this is a PR to develop, we need to be careful with f-strings, since EasyBuild 4.x is still supposed to support Python 2.7:
```suggestion
                self.log.info("Python %s path: %s" % (path_type, path))
```

easybuild/easyblocks/generic/cmakemake.py
```diff
@@ -317,9 +317,60 @@ def configure_step(self, srcdir=None, builddir=None):
...
+
+        errors = []
+        for path_type, paths in python_paths.items():
+            for path in paths:
+                if path.endswith('-NOTFOUND'):
+                    continue
+                if not os.path.exists(path):
+                    errors.append(f"Python {path_type} does not exist: {path}")
+                elif not os.path.realpath(path).startswith(ebrootpython_path):
+                    errors.append(f"Python {path_type} path '{path}' is outside EBROOTPYTHON ({ebrootpython_path})")
```
boegel, line 366: Since this is a PR to develop, we need to be careful with f-strings, since EasyBuild 4.x is still supposed to support Python 2.7:
```suggestion
                    errors.append("Python %s path '%s' is outside EBROOTPYTHON (%s)" % (path_type, path, ebrootpython_path))

File: easybuild/easyblocks/generic/cmakemake.py [hunk 19]
```diff
@@ -317,9 +317,60 @@ def configure_step(self, srcdir=None, builddir=None):
...
+
+        if errors:
+            # Combine all errors into a single message
+            error_message = "\n".join(errors)
+            raise EasyBuildError(f"Python path errors:\n{error_message}")
```
boegel, line 371: Since this is a PR to develop, we need to be careful with f-strings, since EasyBuild 4.x is still supposed to support Python 2.7:
```suggestion
            raise EasyBuildError("Python path errors:\n" + error_message)

**Review by boegel — CHANGES_REQUESTED** [review]

**Comment by boegel** [comment]
@boegelbot please test @ generoso
EB_ARGS="--installpath /tmp/pr3233 FlexiBLAS-3.3.1-GCC-12.3.0.eb SPAdes-3.15.4-GCC-12.3.0.eb Doxygen-1.9.7-GCCcore-12.3.0.eb poppler-23.09.0-GCC-12.3.0.eb"

File: easybuild/easyblocks/generic/cmakemake.py (hunk 12 above)
Comment (Flamefire) [reply easybuild/easyblocks/generic/cmakemake.py:330]: Yes, but it could be empty for some reason

File: easybuild/easyblocks/generic/cmakemake.py (hunk 15 above)
Comment (Flamefire) [reply easybuild/easyblocks/generic/cmakemake.py:353]: I think not, it is likely an error. https://github.com/easybuilders/easybuild-easyblocks/pull/3233#discussion_r1504293421 requested to "complain" ;-)

File: easybuild/easyblocks/generic/cmakemake.py (hunk 17 above)
Comment (Flamefire) [reply easybuild/easyblocks/generic/cmakemake.py:347]: No need to format before logging:
```suggestion
                self.log.info("Python %s path: %s", path_type, path)
```

Flamefire:
I added https://github.com/MartinsNadia/easybuild-easyblocks/pull/3 to get this done fast. It addresses the review and handles the "FindPython" (w/o version number) case.

boegel:
@MartinsNadia ping on reviewing/merging https://github.com/MartinsNadia/easybuild-easyblocks/pull/3 to update this PR?

Flamefire:
As @bedroge retargeted this to 5.x which creates difficult conflicts due to the considerable changes during review I created #3399 combining those 15 commits into 2

boegel:
These changes (and some enhancements on top) got merged via #3399, so closing this (it's not being auto-closed because commits were squashed in #3399)

PR #3256, opened 2024-03-15
set `$LIBRARY_PATH` and disable building of NVPTX offload support on RISC-V systems

bedroge:
Not entirely sure why this is needed, but on my RISC-V system the build will fail (in stage 2 and at the end in the sanity check) with errors like:
```
nvme/easybuild/software/binutils/2.40/bin/ld: cannot find -lgcc_s: No such file or directory
```
Even though `libgcc_s.so` is just in the `lib` directory of the stage 1 / final installation directory. For some reason, it doesn't seem to be in GCC's internal search path. By setting `$LIBRARY_PATH`, it does pick it up.


boegel (approved):
lgtm

boegel:
extensively tested via https://github.com/easybuilders/easybuild-easyconfigs/pull/20035

PR #3261, opened 2024-03-28
sync with develop (20240328)

boegel:


easybuild/easyblocks/q/quantumespresso.py
```diff
@@ -27,232 +27,416 @@
...
+        self._add_elpa()
+        self._add_fftw(comp_fam)
+        self._add_ace()
+        self._add_beef()
+        self._add_fox()
+        self._add_epw()
+        self._add_gipaw()
+        self._add_wannier90()
+
+        run_cmd("module list", log_all=True, log_ok=True, simple=False, regexp=False)
```
Micket, line 389: Should be using run_shell_cmd?

Micket (changes requested):

Micket (approved):
lgtm

PR #3265, opened 2024-04-02
remove unused easyblock for `ncurses` (`ncurses.py`)

branfosj:
All current easyconfigs are `ConfigureMake` and I'm not bothering to work out how far back we made that change.


boegel:
switching to `ConfigureMake` for `ncurses` was done in 2015, see https://github.com/easybuilders/easybuild-easyconfigs/pull/1337

boegel (approved):

PR #3306, opened 2024-04-14
Archive autotools-based QuantumESPRESSO easyblock and switch default to CMakeMake-based easyblock

ocaisa:
This will require updating all of the Quantum ESPRESSO easyconfigs, so is a draft until those are also ready to merge (both should be done at the same time).


ocaisa:
@boegel I've got the test suite passing for this now

easybuild/easyblocks/q/quantumespresso.py
```diff
@@ -25,816 +25,384 @@
...
-        if LooseVersion(self.version) >= LooseVersion("7.3.1"):
-            raise EasyBuildError(
-                "QuantumESPRESSO 7.3.1 and later are not supported with the this easyblock (ConfigureMake), " +
-                "use the EB_QuantumESPRESSOcmake (CMakeMake) easyblock instead."
-                )
+    def _add_gipaw(self):
+        """Enable GIPAW for Quantum ESPRESSO."""
+        if LooseVersion(self.version) == LooseVersion('7.3.1'):
+            # See issue: https://github.com/dceresoli/qe-gipaw/issues/19
+            raise EasyBuildError('GIPAW will fail to compile in QE 7.3.1')
```
ocaisa, line 193:
```suggestion
        if LooseVersion(self.version) == LooseVersion('7.3.1'):
            # See issue: https://github.com/dceresoli/qe-gipaw/issues/19
            raise EasyBuildError('GIPAW will fail to compile in QE 7.3.1')
```
No longer necessary due to the change in https://github.com/easybuilders/easybuild-easyconfigs/commit/a563666c80665dcb70cbf748bf5f487788db3dc0

ocaisa:
Closing this in favour of #3338

PR #3331, opened 2024-05-15
Fix for plumed genereting patch file inside module location

Crivella:
The command

`plumed patch --new mw2` was actually generating an empty `mw2.diff` file under the `$EBROOTPLUMED/lib/plumed/patches` directory which will cause trouble in system where this location is read-only (e.g. EESSI)

Switched to manually generating a `mw2.diff` empty file inside the build folder and point to that using the `-d` option for plumed

Also added a compatibility check as for `f90wrap <= 0.2.13` for `MetalWalls <= 21.06.1`


ocaisa (approved):
LGTM

PR #3339, opened 2024-05-24
Add RISCV64 support and update config options of new versions of Extrae

julianmorillo:

PR #3344, opened 2024-05-31
use `--with-xml` to specify `libxml2` root dir in Extrae easyblock

bedroge:
Without this option Extrae will fail to find an EB-installed `libxml2`, and XML support will then be disabled.


bedroge:
Closing this in favor of #3339.

PR #3354, opened 2024-06-09
Update wrf.py for 4.5.2

thanghn90:
WRF 4.5.2 actually use "WRF-4.5.2" subdirectory naming scheme. I'm about to make another pull request to add my WRF-4.5.2-intel-2023a.eb easyconfig file, along with an associated patch file, to easyconfigs repo, and I need this change to go through before that.

### DIFF
--- a/easybuild/easyblocks/w/wrf.py
+++ b/easybuild/easyblocks/w/wrf.py
@@ -56,7 +56,8 @@ def det_wrf_subdir(wrf_version):
 
     if LooseVersion(wrf_version) < LooseVersion('4.0'):
         wrf_subdir = 'WRFV%s' % wrf_version.split('.')[0]
-    elif LooseVersion(wrf_version) == LooseVersion('4.5.1'): # WRF 4.5.2 actually use "WRF-4.5.2" subdirectory naming scheme
+    elif LooseVersion(wrf_version) == LooseVersion('4.5.1'):
+        # WRF 4.5.2 actually use "WRF-4.5.2" subdirectory naming scheme
         wrf_subdir = 'WRFV%s' % wrf_version
     else:
         wrf_subdir = 'WRF-%s' % wrf_version
--- a/easybuild/easyblocks/w/wrf.py
+++ b/easybuild/easyblocks/w/wrf.py
@@ -56,7 +56,7 @@ def det_wrf_subdir(wrf_version):
 
     if LooseVersion(wrf_version) < LooseVersion('4.0'):
         wrf_subdir = 'WRFV%s' % wrf_version.split('.')[0]
-    elif LooseVersion(wrf_version) >= LooseVersion('4.5.1'):
+    elif LooseVersion(wrf_version) == LooseVersion('4.5.1'): # WRF 4.5.2 actually use "WRF-4.5.2" subdirectory naming scheme
         wrf_subdir = 'WRFV%s' % wrf_version
     else:
         wrf_subdir = 'WRF-%s' % wrf_version


bedroge:
Thanks for your contribution, @thanghn90!

I just had a look at the release tarballs of WRF 4.5.1 and 4.5.2, but didn't really see a difference in the directory name:
```
$ tar tzf v4.5.1.tar.gz | head -n 3
WRFV4.5.1/
WRFV4.5.1/README
WRFV4.5.1/main/

$ tar tzf v4.5.2.tar.gz | head -n 3
WRFV4.5.2/
WRFV4.5.2/README
WRFV4.5.2/main/
```

But looking at your easyconfig PR https://github.com/easybuilders/easybuild-easyconfigs/pull/20779, I see you're using `git_config` to obtain the sources, and clone them into `WRF-%(version)s`. Isn't that causing the difference? Is there a particular reason why you're using this approach, instead of just using the release tarball? The latter is done in the PR for version 4.5.1: https://github.com/easybuilders/easybuild-easyconfigs/pull/18741.

thanghn90:
Hi Bob!
Thanks for the headsup.
My bad, i was completely unaware of an WRF 4.5.1 eb config pull request.
Hence, i was only following WRF 4.4.1 eb config file's syntax.
Will try the new syntax in the WRF 4.5.1 PR and update my branch soon.
Thang

bedroge:
I hadn't seen that the 4.4 easyconfigs were using the git clone approach. I'm not sure if there's a good reason for it, but if there is, it's of course fine to stick to that. I also see that the PR for 4.5.1 still has failing tests, so perhaps it's even related to that?

ocaisa:
I looked into this a bit in slack and the following was required
```
source_urls = ['https://github.com/wrf-model/WRF/archive/']
# sources = ['v%(version)s.tar.gz']
sources = [{
    'filename': 'v%(version)s.tar.gz',
    'git_config': {
        'url': 'https://github.com/wrf-model',
        'repo_name': 'WRF',
        'tag': 'v%(version)s',
        'clone_into': 'WRF-%(version)s',
        'recursive': True,
        'keep_git_dir': True,
    },
}]
```
They've also changed how they detect netCDF, and this was also required:
```
preconfigopts = "export NETCDF=$EBROOTNETCDFMINFORTRAN && export NETCDF_C=$EBROOTNETCDF &&"
```

bedroge:
> You must stick to the `git clone ...` approach as WRF does not ship all the required submodules (see https://www2.mmm.ucar.edu/wrf/users/download/get_sources_new.php)
> 
> > **Beginning with V4.4, submodules are obtained during compile; therefore this command is required to obtain the WRF code.**
> > git clone --recurse-submodules https://github.com/wrf-model/WRF

Things are quite confusing with their tarballs, as the one for the tag uses `WRF-<version>`, while the one for the release uses `WRFV<version>`. But both don't include the git submodule for noahmp, so we should indeed try to stick to the `git_config` approach then.
@thanghn90 That means you can probably just set `clone_into` to `WRFV%(version)s`, and then the change in the easyblock shouldn't be required anymore.

bedroge:
The release page says this though:
> NOTE!!!! : If downloading the files from the section below, please choose either the v4.5.1.tar.gz file, or the v4.5.1.zip file. DO NOT choose those named "Source Code." They do not include the mandatory NoahMP submodule - needed for compiling WRF.
So apparently it should include it?

bedroge:
Ah, I guess I checked the wrong file, it is indeed there:
```
$ wget https://github.com/wrf-model/WRF/releases/download/v4.5.2/v4.5.2.tar.gz
$ tar tzf v4.5.2.tar.gz  | grep noahmp
WRFV4.5.2/phys/noahmp/
WRFV4.5.2/phys/noahmp/parameters/
WRFV4.5.2/phys/noahmp/parameters/GENPARM.TBL
WRFV4.5.2/phys/noahmp/parameters/MPTABLE.TBL
WRFV4.5.2/phys/noahmp/parameters/SOILPARM.TBL
WRFV4.5.2/phys/noahmp/drivers/
WRFV4.5.2/phys/noahmp/drivers/wrf/
WRFV4.5.2/phys/noahmp/drivers/wrf/module_sf_noahmpdrv.F
WRFV4.5.2/phys/noahmp/README.md
WRFV4.5.2/phys/noahmp/.git
WRFV4.5.2/phys/noahmp/src/
WRFV4.5.2/phys/noahmp/src/module_sf_noahmp_groundwater.F
WRFV4.5.2/phys/noahmp/src/module_sf_noahmp_glacier.F
WRFV4.5.2/phys/noahmp/src/module_sf_noahmplsm.F
WRFV4.5.2/Registry/registry.noahmp

```

But the one of the git tag doesn't include it:
```
$ wget https://github.com/wrf-model/WRF/archive/refs/tags/v4.5.2.tar.gz
$ tar tzf v4.5.2.tar.gz | grep noahmp
WRF-4.5.2/Registry/registry.noahmp
WRF-4.5.2/phys/noahmp/
```

So we just have to be careful and download the right file in the easyconfig,

thanghn90:
Hmmm...
Anyways, I guess I'll try the approach of
easybuilders/easybuild-easyconfigs#18741
<https://github.com/easybuilders/easybuild-easyconfigs/pull/18741> to see
if  it works on my end. As I mentioned before, I was completely unaware of
this. AND I'll attempt the preconfigopts mentioned by ocaisa without
applying the netcdf-fortran-separate-path patch, though I have a very low
confidence this would work, given how conservative WRF tends to be.
Will keep you posted.
Thang

thanghn90:
I was able to compile WRF 4.5.2 using the github release tar file (the one that creates WRFV4.5.2 subdirectory). This change should now be ignored.

PR #3365, opened 2024-06-17
use `flake8` 7.0.0 for linting easyblocks

branfosj:
Only for `develop` to avoid having to fix `geant4.py` (see https://github.com/easybuilders/easybuild-easyblocks/actions/runs/9552967707/job/26330689071?pr=3364).

Not needed in `5.0.x` as we've removed that section of the easyblock.

### DIFF
--- a/.github/workflows/linting.yml
+++ b/.github/workflows/linting.yml
@@ -22,7 +22,8 @@ jobs:
     - name: install Python packages
       run: |
         pip install --upgrade pip
-        pip install --upgrade flake8
+        # fix to this version for develop branch (to avoid needing to fix geant4.py)
+        pip install --upgrade flake8<7.1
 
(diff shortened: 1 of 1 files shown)


.github/workflows/linting.yml
```diff
@@ -22,7 +22,8 @@ jobs:
     - name: install Python packages
       run: |
         pip install --upgrade pip
-        pip install --upgrade flake8
+        # fix to this version for develop branch (to avoid needing to fix geant4.py)
+        pip install --upgrade "flake8<7.1"
```
boegel, line 26: We'll sync `develop` to `5.0.x` eventually, so this will also land in there (unless we kick it out when doing sync), but OK for now...

boegel (approved):

PR #3366, opened 2024-06-17
make sure that OpenFOAM's wmake can find MPFR and GMP if CGAL >= 5.0 (header-only) is used

bedroge:
With EESSI we ran into several undefined references to GMP and MPFR when building OpenFOAM v2312, see https://github.com/EESSI/software-layer/pull/563#issuecomment-2154947665

This seems related to CGAL, which is header-only since version 5.0. Apparently, here's still some kind of indirect dependency on GMP and MPFR (the Ubuntu installation instructions also have these packages listed: https://gitlab.com/openfoam/openfoam/-/blob/master/doc/Requirements.md#ubuntu-eg-2204), which are lacking in the EESSI build container.

Even though OpenFOAM's wmake scripts seems to contain functionality for handling these situations (there is a `cgal-no-mpfr` rule, for instance: https://gitlab.com/openfoam/openfoam/-/blob/master/wmake/rules/General/cgal-no-mpfr?ref_type=heads), there are lots of issues about the same topic, e.g. this one related to Spack: https://develop.openfoam.com/Development/openfoam/-/issues/3101.
They even stick to CGAL 4.x: https://github.com/spack/spack/blob/develop/var/spack/repos/builtin/packages/openfoam/package.py#L372.

This PR fixes this by overriding some variables in the `have_cgal` script, which finds CGAL and MPFR, in case CGAL >= 5.0 is used and MPFR is listed as dependency in the easyconfig. For the latter I'll open an easyconfigs PR.

### DIFF
--- a/easybuild/easyblocks/o/openfoam.py
+++ b/easybuild/easyblocks/o/openfoam.py
@@ -41,6 +41,7 @@ import re
 import shutil
 import stat
 import tempfile
+import textwrap
 from easybuild.tools import LooseVersion
 
 import easybuild.tools.environment as env
@@ -48,7 +49,7 @@ import easybuild.tools.toolchain as toolchain
 from easybuild.easyblocks.generic.cmakemake import setup_cmake_env
 from easybuild.framework.easyblock import EasyBlock
 from easybuild.tools.build_log import EasyBuildError
-from easybuild.tools.filetools import adjust_permissions, apply_regex_substitutions, mkdir
+from easybuild.tools.filetools import adjust_permissions, apply_regex_substitutions, mkdir, write_file
 from easybuild.tools.modules import get_software_root, get_software_version
 from easybuild.tools.run import run_cmd, run_cmd_qa
 from easybuild.tools.systemtools import get_shared_lib_ext, get_cpu_architecture, AARCH64, POWER
@@ -292,6 +293,23 @@ class EB_OpenFOAM(EasyBlock):
                         env.setvar("BOOST_ROOT", get_software_root('Boost'))
                     else:
                         env.setvar("%s_ROOT" % depend.upper(), dependloc)
+
+            if get_software_root('CGAL') and LooseVersion(get_software_version('CGAL')) >= LooseVersion('5.0'):
+                # CGAL >= 5.x is header-only, but when using it OpenFOAM still needs MPFR.
+                # It may fail to find it, so inject the right settings and paths into the "have_cgal" script.
+                have_cgal_script = os.path.join(self.builddir, self.openfoamdir, 'wmake', 'scripts', 'have_cgal')
+                if get_software_root('MPFR') and os.path.exists(have_cgal_script):
+                    eb_cgal_config = textwrap.dedent('''
+                    # Injected by EasyBuild
+                    HAVE_CGAL=true
+                    HAVE_MPFR=true
+                    CGAL_FLAVOUR=header
+                    CGAL_INC_DIR=${EBROOTCGAL}/include
+                    CGAL_LIB_DIR=${EBROOTCGAL}/lib
+                    MPFR_INC_DIR=${EBROOTMPFR}/include
+                    MPFR_LIB_DIR=${EBROOTMPFR}/lib
+                    ''')
+                    write_file(have_cgal_script, eb_cgal_config, append=True)
 
     def build_step(self):
         """Build OpenFOAM using make after sourcing script to set environment."""
--- a/easybuild/easyblocks/o/openfoam.py
+++ b/easybuild/easyblocks/o/openfoam.py
@@ -486,6 +486,11 @@ class EB_OpenFOAM(EasyBlock):
             test_foammonitor = "! foamMonitor -h 2>&1 | grep 'not installed'"
             custom_commands.append(' && '.join([load_openfoam_env, test_foammonitor]))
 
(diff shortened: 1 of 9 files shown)


easybuild/easyblocks/o/openfoam.py
```diff
@@ -292,6 +292,25 @@ def configure_step(self):
                         env.setvar("BOOST_ROOT", get_software_root('Boost'))
                     else:
                         env.setvar("%s_ROOT" % depend.upper(), dependloc)
+            for dep in self.cfg.dependencies():
+                # CGAL >= 5.x is header-only, but when using it OpenFOAM still needs MPFR and GMP.
+                # It may fail to find them, so inject the right settings and paths into the "have_cgal" script.
+                if dep['name'] == 'CGAL' and LooseVersion(dep['version']) >= LooseVersion('5.0'):
```
boegel, line 298: Why not use `get_software_root('CGAL')` to check whether CGAL is loaded as dependency, and `get_software_version('CGAL')` to do the version check?

easybuild/easyblocks/o/openfoam.py
```diff
@@ -292,6 +292,25 @@ def configure_step(self):
                         env.setvar("BOOST_ROOT", get_software_root('Boost'))
                     else:
                         env.setvar("%s_ROOT" % depend.upper(), dependloc)
+            for dep in self.cfg.dependencies():
+                # CGAL >= 5.x is header-only, but when using it OpenFOAM still needs MPFR and GMP.
```
boegel, line 296: comment mentions GMP, but nothing is being done for GMP below, only MPFR?

boegel (changes requested):
https://github.com/bedroge/easybuild-easyblocks/pull/1

easybuild/easyblocks/o/openfoam.py
bedroge: Didn't know about `get_software_version`, that's way easier indeed!

easybuild/easyblocks/o/openfoam.py
bedroge: Yeah, good question. It does look like it needs GMP, but there's no `GMP_LIB_DIR`... In for instance https://develop.openfoam.com/Development/openfoam/-/blob/OpenFOAM-v2312/wmake/rules/General/cgal-header-only?ref_type=tags they do add `-lgmp`. So I'm not completely sure what to do here...

bedroge: With the current approach it does build fine, but we did have undefined references to GMP before. So I guess it does somehow find GMP.

easybuild/easyblocks/o/openfoam.py
```diff
@@ -292,6 +292,25 @@ def configure_step(self):
...
+                        # Injected by EasyBuild
+                        HAVE_CGAL=true
+                        HAVE_MPFR=true
+                        CGAL_FLAVOUR=header
+                        CGAL_INC_DIR=${EBROOTCGAL}/include
+                        CGAL_LIB_DIR=${EBROOTCGAL}/lib
+                        MPFR_INC_DIR=${EBROOTMPFR}/include
+                        MPFR_LIB_DIR=${EBROOTMPFR}/lib
+                        '''
+                        apply_regex_substitutions(have_cgal_script, [(r"^(esac)$", r"\1\n" + eb_cgal_config)])
```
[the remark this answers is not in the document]
bedroge: Ah, didn't know that was possible, and even looked in filetools to find something that could do this, but didn't find this. Thanks!

easybuild/easyblocks/o/openfoam.py
```diff
@@ -292,6 +292,25 @@ def configure_step(self):
                         env.setvar("BOOST_ROOT", get_software_root('Boost'))
                     else:
                         env.setvar("%s_ROOT" % depend.upper(), dependloc)
+            for dep in self.cfg.dependencies():
+                # CGAL >= 5.x is header-only, but when using it OpenFOAM still needs MPFR and GMP.
+                # It may fail to find them, so inject the right settings and paths into the "have_cgal" script.
+                if dep['name'] == 'CGAL' and LooseVersion(dep['version']) >= LooseVersion('5.0'):
+                    if get_software_root('MPFR'):
+                        have_cgal_script = os.path.join(
+                            self.builddir, self.openfoamdir, 'wmake', 'scripts', 'have_cgal'
```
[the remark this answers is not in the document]
bedroge: I thought I had put this in an else branch that only applies to the dotcom version, but I just noticed that's not true. And better to be careful with this anyway.

Maybe we can even take this new stuff out of this if-else statement?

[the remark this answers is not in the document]
boegel: Keeping it there is fine

boegel:
@bedroge CI problem not related to changes here is fixed via #3365, this PR needs a sync with `develop`

bedroge:
> @bedroge CI problem not related to changes here is fixed via #3365, this PR needs a sync with `develop`

@boegel done!

boegel:
CORE_CNT=16
EB_ARGS="OpenFOAM-v2306-foss-2022b.eb OpenFOAM-v2312-foss-2023a.eb OpenFOAM-11-foss-2023a.eb"

boegel (approved):

boegel (approved):

PR #3373, opened 2024-06-25
enhance `LLVM` easyblock for compilation of clang/flang + other llvm-projects

Crivella:
This EB is a modified version of the current `Clang` one (whether to keep them separate, having a simpler EB for Clang-only, or rename Clang to LLVMcore or viceversa is up for debate).

Requires testing of:

- [ ] https://github.com/easybuilders/easybuild-easyconfigs/pull/20902 (Changes to old ECs for compatibility)
- [ ] https://github.com/easybuilders/easybuild-easyconfigs/pull/21832 (LLVM 18.1.8 + GCCcore/13.3.0)
- [ ] https://github.com/easybuilders/easybuild-easyconfigs/pull/21833 (LLVM 18.1.8 + system)
- [ ] https://github.com/easybuilders/easybuild-easyconfigs/pull/21611 (LLVM 19.1.1)

It enables building most of the main [LLVM projects](https://github.com/llvm/llvm-project). What is still missing is/has not been tested:

- `libc`
- `libclc`
- `llvm-libgcc`
- `pstl`

The current EB improves on the Clang one by:

- Implementing new version of flags for version of `LLVM >= 18`
- Implement running the test-suite for all LLVM projects (with an EC specified allowed number of failures)
- Implement both a bootstrapped and non-bootstrapped version of the toolchain
- Implement a *LLVM only* build (bootstrapped only) where all dependencies from GCC are removed at the final stage

Currently/WIP:

- The `full_llvm` build finishes successfully with 74 out of 119128 tests failing
  - 51 failures in `bolt`
  - 1 failure in `clang tools`
  - 18 failures in `mlir`
  - 3 failures in `lldb`
  - 1 failure in `ompd`
  - Same set of failures without the MLIR ones and only 1 `lldb` with a static build (55 out of 119090)
- the `bootstrapped` non `full_llvm` build finishes successfully with 46 out of 119897 tests failing
  - 2 failures in `bolt`
  - 1 failure in `clang tools`
  - 18 failures in `mlir`
  - 21 failures in `MemProfiler-x86_64-linux`
  - 3 failures in `lldb`
  - 1 failure in `ompd`
- the non `bootstrapped` version builds successfully but not build the test-suite (running `make -j 1 check-all` inside the builddir
  worked leading to 52 failures + 2 timeout out of 119900) (*NOTE*: the number of MemProfiler and timeout failure is not consistent)

  - Note: arguable benefit of the non-bootstrapped build as it took almost the same time as a 3 stage bootstrapped one
    The last stage with all the required projects is always the longest and it seems that `clang` is faster at building (at-least within the scope of the LLVM project)
  - Test if problem is with `make -j XXX (!=1)`
     - **UPDATE**: Added a step to try compiling the tests with `-j 1` if the parallel build fails and the installation runs successfully
  - Investigate further failures once the build runs successfully with EB only
  - 2 failures in `bolt`
  - 1 failure in `clang tools`
  - 18 failures in `mlir`
  - 25 failures in `MemProfiler-x86_64-linux` (this number is not consistent across builds)
  - 5 failures in `lldb`
  - 1 failure in `ompd`
- [X] Improving sanity check depending on enabled options

Notes:

- Building a simple C++ openmp program using `iostream` seems to work both with `libc++` and `libstdc++`
  (tested with the `full_llvm` build and the `bootstrapped` non full one)
- For both builds an hello world openmp program compiles with both `libomp` and `libgomp` but the one compiled with
  `libgomp` does not seem to recognize `OMP_NUM_THREADS` and runs with 1 thread.
  Tested with C (`clang`), C++ (clang++) and fortran (`flang-new`)

Possible TODOs:

- [X] Test GPU offloading by building and running on a system with GPU
  - For now works only with `clang` (Tested on VEGA front-end with an A100)
  - See [this discourse post](https://discourse.llvm.org/t/offloading-on-nvptx64-target-with-flang-new-leads-to-undefined-reference-s/80237) for `flang-new`
- [X] Test RPATH wrappers


easybuild/easyblocks/l/llvmcore.py
```diff
@@ -0,0 +1,748 @@
...
+# -DOPENMP_ENABLE_OMPT_TOOLS=OFF \
+# -DOPENMP_ENABLE_TESTING=OFF \
+# -DCOMPILER_RT_CXX_LIBRARY=libcxx \
+# -DSANITIZER_CXX_ABI=libc++ \
+# -DSANITIZER_CXX_ABI_INTREE=ON \
+# -DSANITIZER_USE_STATIC_CXX_ABI=ON \
+# -DSANITIZER_TEST_CXX=libc++ \
+# -DSANITIZER_TEST_CXX_INTREE=ON \
+# -DSANITIZER_USE_STATIC_TEST_CXX=ON \
+# -DSANITIZER_LIT_USE_LIBCXX=ON \
```
Thyre, line 142: Is there any reason for the comments being present? It looks like some are based on LLVM issues or other discussions.

easybuild/easyblocks/l/llvmcore.py
```diff
@@ -0,0 +1,748 @@
...
+    'CLANG_DEFAULT_LINKER': 'lld',
+    'CLANG_DEFAULT_UNWINDLIB': 'libunwind',
+
+    'LIBCXX_HAS_GCC_S_LIB': 'Off',
+    'LIBCXXABI_HAS_GCC_S_LIB': 'Off',
+    'LIBUNWIND_HAS_GCC_S_LIB': 'Off',
+    # 'COMPILER_RT_HAS_GCC_S_LIB': 'Off',
+    # 'CLANG_HAS_GCC_S_LIB': 'Off',
+
+    # 'OPENMP_ENABLE_OMPT_TOOLS': 'Off',
```
Thyre, line 181: For the EasyBlock, I would give users the choice to disable the OpenMP Tools Interface, if they wish to do so. There is little reason to do so, and I would leave it on by default, but maybe users are interested in disabling it.
That option should exist even when not doing a full independent LLVM build though.

easybuild/easyblocks/l/llvmcore.py
```diff
@@ -0,0 +1,748 @@
...
+# -DCOMPILER_RT_CXX_LIBRARY=libcxx \
+# -DSANITIZER_CXX_ABI=libc++ \
+# -DSANITIZER_CXX_ABI_INTREE=ON \
+# -DSANITIZER_USE_STATIC_CXX_ABI=ON \
+# -DSANITIZER_TEST_CXX=libc++ \
+# -DSANITIZER_TEST_CXX_INTREE=ON \
+# -DSANITIZER_USE_STATIC_TEST_CXX=ON \
+# -DSANITIZER_LIT_USE_LIBCXX=ON \
+
+remove_gcc_opts = {
```
Thyre, line 144:
```suggestion
# Additional CMake flags required in an independent LLVM build to remove dependencies to GCCcore toolchain.
remove_gcc_dependency_opts = {
```

These look like options only used when opting to do a `full_llvm` build, right? `remove_gcc_opts` sounds a bit misleading in my opinion.

easybuild/easyblocks/l/llvmcore.py
Crivella: No, they were leftovers to keep track of some of the combinations i had tried.
Since now also the non bootstrapped build is working I've removed them in https://github.com/easybuilders/easybuild-easyblocks/pull/3373/commits/541e65a73973fc4438dbb0378a2e3e2a359225bd.

Also thanks for the fast review and sorry for the late reply i went on holiday right after opening the PR.

easybuild/easyblocks/l/llvmcore.py
Crivella: Yes, I agree. I've done the change in https://github.com/easybuilders/easybuild-easyblocks/pull/3373/commits/e18fedd093bfc34cec85130ce58a65aa90f8b666

easybuild/easyblocks/l/llvmcore.py
Crivella: I agree `remove_gcc_dependency_opts` would be cleared i changed the name in https://github.com/easybuilders/easybuild-easyblocks/pull/3373/commits/ce1227c5dd754ce2a51763583140c6897dd034d0

boegel:
I haven't looked at this in detail, but I'm not a big fan of "forking" an existing easyblock, and then updating it. There seems to be a lot of overlap still with the custom easyblock for Clang.

I do understand that the other option (updating the existing custom easyblock for Clang to introduce updates/enhancements) is painful too though, partially since there's a lot of history there.

@Crivella How difficult would it be to integrate the functionality you need in the existing easyblock for Clang (or LLVM, I'm not even sure which one is most relevant here, but I guess it's the Clang one).

Crivella:
@boegel I will try to work on it as soon as I finish all the current tests (there is still something to iron out on non shared builds, and understanding if the test failures should be addressed or can be ignored).
I went this route originally as I was working both on a standalone toolchain for building the single components separately and this one (and also for ease of testing features out). I will have to check how compatible the way the bootstrap was being done in the previous build with respect to this one, but it should be doable.

I do not know if it would still make sense to keep the Clang naming scheme as this would end up including also Flang and other optimization tools usable also with other frontends.

ocaisa:
@boegel Isn't it a lot to ask to merge these easyblocks? The reality is that we already have two other forks for LLVM and AOMP, so the precedent already exists for taking a new path.

Crivella:
@boegel 
I've been looking back at the 2 EB side by side and while not impossible to merge them, i think the most feasible approach would be the QE route of calling a different EB based on the version of the software.
While a lot of pieces have been taken from the `Clang` EB, here I've taken a fundamentally different approach for a few options and for the bootstrap build which would require deprecating options from the old EB if a new version is used.
Concerning the bootstrap build, in this EB, not all tools are being built at all stages of the build (only the minimally required one) and different options are being passed at different stages which would require rewriting half of the `Clang` EB to merge.

There is also the fact that for the `Clang` EB, `flang` is just an optional part of the build to be manually specified through the `llvm_projects` extra option.
The idea behind this EB is not to build just Clang but all the tooling required to have an LLVM toolchain in place which would require both `clang` and `flang` to always be present

ocaisa:
One thing I don't actually like is the name, `clang.py`, because it is not _just_ `clang`, it is LLVM with front end compilers. That is part of the complication, because in reality it should be `llvm.py`, and `clang.py` should stop being used. QE-style forking within the relevant easyblocks seems like the only available option.

Crivella:
As an update, i've also tested now with all tools enabled and `build_targets = ['all']`.
The compilation + test_suite was successful with only a slight increase in errors (with respect to the previous bootstrap+full_llvm build with default build target):

- 77 errors out of 119238 tests
  - 50 due to `bolt`
  - 6 due to `clang` + PowerPC architecture
  - 1 due to `clang-tools`
  - 18 due to `mlir`
  - 1 due to `lldb`
  - 1 due to `ompd`

It should be noted that before 34341 (28.84%) were flagged as unsupported while now only 4588 (3.85%) which should be due to the `experimental` archs

Crivella:
**UPDATE**

Testing with openmp offloading on VEGA frontend node:
- AMD EPYC 7452 (Zen2 Rome)
- NVIDIA A100

With:

```
build_targets = ['X86', 'NVPTX']
cuda_compute_capabilities = ['8.0']
```

I encountered extra errors related to  `libomptarget`:
- 270 due to x86 (out of 275)
- 11 due to nvptx (out of 302)

As a sanitycheck I also recompiled on my WS by allowing libomptarget to be compiled for x86 (before i had it enabled only  for nvptx and amd) and i get a similar number of test errors (268)

Beside that i tried to compile a simple openmp application and the GPU offloading seems to be working correctly (verified the GPU is being used from nvtop)

Crivella:
**UPDATE**

Testing with openmp offloading on VEGA frontend node:
- AMD EPYC 7452 (Zen2 Rome)
- NVIDIA A100

With:

```
build_targets = ['X86', 'NVPTX']
cuda_compute_capabilities = ['8.0']
```

I encountered extra errors related to  `libomptarget`:
- 270 due to x86 (out of 275)
- 11 due to nvptx (out of 302)

As a sanitycheck I also recompiled on my WS by allowing libomptarget to be compiled for x86 (before i had it enabled only  for nvptx and amd) and i get a similar number of test errors (268)

Beside that i tried to compile a simple openmp application and the GPU offloading seems to be working correctly (verified the GPU is being used from nvtop)

Thyre:
I've tried to build LLVM using your easyblock and easyconfig PRs on a fresh `devel` installation. I had some issues with Python, where the tests reported that Python itself was not found even though it was installed. After disabling the Python parts I had no trouble building both LLVMcore variants.

I haven't looked to deeply into which and how many tests failed though. I built everything on my local workstation (Intel Core i7-12700, AMD Radeon RX7700XT).

Crivella:
@Thyre

Thanks for the report, will look into it
The error was related to the `python_bindings = True` setting in the EC files?

Thyre:
> @Thyre
> 
> Thanks for the report, will look into it The error was related to the `python_bindings = True` setting in the EC files?

Yes, exactly. When setting it to `False`, everything worked fine. 
The steps to reproduce were basically:

1. Clone the EB repos (with your PRs) (instructions taken from [here](http://tutorial.easybuild.io/2023-eb-eessi-uk-workshop/easybuild-installation/#method-3-development-setup))
2. Run the following command: `eb --robot LLVMcore-18.1.7.eb LLVMcore-18.1.7-GCCcore-13.3.0.eb --parallel 8`

easybuild/easyblocks/l/llvm.py
```diff
@@ -94,31 +305,611 @@ def configure_step(self):
...
+            # list of CUDA compute capabilities to use can be specifed in two ways (where (2) overrules (1)):
+            # (1) in the easyconfig file, via the custom cuda_compute_capabilities;
+            # (2) in the EasyBuild configuration, via --cuda-compute-capabilities configuration option;
+            ec_cuda_cc = self.cfg['cuda_compute_capabilities']
+            cfg_cuda_cc = build_option('cuda_compute_capabilities')
+            cuda_cc = cfg_cuda_cc or ec_cuda_cc or []
+            if not cuda_cc and self.nvptx_cond:
+                raise EasyBuildError(
+                    "Can't build Clang with CUDA support without specifying 'cuda-compute-capabilities'"
+                    )
```
Thyre, line 333: Please correct me if I'm wrong:

The current approach would just take the host architecture and not NVPTX / AMDGPU if the user doesn't explicitly set anything in their EasyConfig, right? Would it make sense to use the approach from the [Clang EasyBlock](https://github.com/easybuilders/easybuild-easyblocks/blob/21ad3f57a7c40078a5402bf0cdb4be710ae49a15/easybuild/easyblocks/c/clang.py#L180) or detect enabling them in another way (e.g. via `cuda_compute_capabilities`)?

```py
if build_targets is None:
   deps = [dep['name'].lower() for dep in self.cfg.dependencies()]
   arch = get_cpu_architecture()
   try:
       default_targets = DEFAULT_TARGETS_MAP[arch][:]
       # If CUDA is included as a dep, add NVPTX as a target
       # There are (old) toolchains with CUDA as part of the toolchain
       cuda_toolchain = hasattr(self.toolchain, 'COMPILER_CUDA_FAMILY')
       if 'cuda' in deps or cuda_toolchain:
           default_targets += ['NVPTX']
       # For AMDGPU support we need ROCR-Runtime and
       # ROCT-Thunk-Interface, however, since ROCT is a dependency of
       # ROCR we only check for the ROCR-Runtime here
       # https://openmp.llvm.org/SupportAndFAQ.html#q-how-to-build-an-openmp-amdgpu-offload-capable-compiler
       if 'rocr-runtime' in deps:
           default_targets += ['AMDGPU']
       self.cfg['build_targets'] = build_targets = default_targets
       self.log.debug("Using %s as default build targets for CPU/GPU architecture %s.", default_targets, arch)
   except KeyError:
       raise EasyBuildError("No default build targets defined for CPU architecture %s.", arch)
```

Crivella: Yes currently only the host architecture is used by default.

What the `Clang` EB does is just check if `cuda` or `rocr-runtime` is in the dependencies.
I think atleast for CUDA i had removed it as i was able to get offload going without adding CUDA to the deps (will have to double check as i did it a couple months ago and don't have access to the frontend where i tested this).

For using `cuda_compute_capabilities` i followed the same approach as the `Clang` EB of checking that it is defined when `NVPTX` is enabled as a target and set the proper variables (or fail if it is not defined).

I guess it would make sense to automatically enable NVPTX/AMDGPU if the respective deps are specified in the EC or if `cuda_compute_capabilities`/`amd_gfx_list` are defined

I am thinking the advantage of the automatic detection would be to be able to have NVPTX/AMDGPU enabled without having to also manually specify the CPU target.
I can try adding the original check and see how it goes

Thyre: > I guess it would make sense to automatically enable NVPTX/AMDGPU if the respective deps are specified in the EC or if cuda_compute_capabilities/amd_gfx_list are defined

For our systems at JSC, I added this small snippet to our "custom" LLVM EasyBlock and used your second idea to check if NVPTX should be built:

```python
cuda_cc_list = build_option('cuda_compute_capabilities') or self.cfg['cuda_compute_capabilities']
if cuda_cc_list:
    build_targets += ["NVPTX"]
    self.log.debug("Add NVPTX as additional build target architecture")
```

The code is not perfect, but is likely sufficient for our use case. LLVM does not require having CUDA / ROCm to build the respective NVPTX / AMDGPU backends as far as I am aware of, only when actually using them. 

> I am thinking the advantage of the automatic detection would be to be able to have NVPTX/AMDGPU enabled without having to also manually specify the CPU target.

Agreed, this was my thought as well. Especially if one has a "global" repository for multiple systems with different architectures, needing to specify the CPU target each time might get messy. Users are also still able to specify the targets manually, as the automatic detection is only done when `build_targets` is not defined.

Crivella: I've implemented the automatic detection based on having `cuda_compute_capabilities` or `amd_gfx_list` defined ([cac8fe9](https://github.com/easybuilders/easybuild-easyblocks/pull/3373/commits/cac8fe94162462a7b3860b5eaf2c38ed356f9995)).

For now this is being done only if no `build_targets` are being manually specified.
It is also true that if someone is specifying either of the 2 variables they might want to have the target enabled, but i feel giving priority should still be given to  `build_targets` (Maybe could add a check for printing a warning or outright raising an error if cuda_cc/amd_gfx are defined and the corresponding targets are not).

Thyre: I think your current approach is a good idea, only enabling the backends if the user has not defined which ones to build. Adding a warning (or at least a log entry) if he might've missed one might be a good idea.

Crivella:
Following a discussion in slack about the deprecation warning about `GCC_INSTALL_PREFIX` switched to an error for LLVM 19 (see also https://github.com/llvm/llvm-project/pull/85891):

- COMMITS:
  - https://github.com/easybuilders/easybuild-easyblocks/pull/3373/commits/bf310765b3affb958e71cbdaea58b933751f703f
  - https://github.com/easybuilders/easybuild-easyblocks/pull/3373/commits/13a7a9751653bd0e8a5db3e825817a831ea40e9d

- Added version check to use the following for LLVM >= 19 instead of `GCC_INSTALL_PREFIX`
  - For the build: 
        ```-DRUNTIMES_CMAKE_ARGS="-DCMAKE_C_FLAGS=--gcc-install-dir=$GCC_ROOT;-DCMAKE_CXX_FLAGS=--gcc-install-dir=$GCC_ROOT"```
  where `$GCC_ROOT` should point to `$EBROOT.../lib/gcc/<triple>/<gcc_version>` (see https://github.com/llvm/llvm-project/pull/85891#issuecomment-2021370667)
  -  Added generation of config files in the `post_install_step` with the same name as the compilers next to them to automatically set `--gcc-install-dir` when the compiler is used (used to get the right location for the `crt` files like `crtbeginS.o`) (see https://discourse.llvm.org/t/add-gcc-install-dir-deprecate-gcc-toolchain-and-remove-gcc-install-prefix/65091).
  This behavior can be overwritten by manually setting the flag at compiler invocation or using another config files on top of the generated one
      - NOTE: The `flang-new` compiler looks for a `flang.cfg` instead of `flang-new.cfg` configuration files
  - Added a sanity check function to invoke `compiler_name -v` and check if the correct GCC installation is being detected
- LLVM < 19 will still used GCC_INSTALL_PREFIX (technically deprecated since 16 but still working)
  - NOTE: Tried using the same fix also for LLVM 18, but the options `--gcc-toolchain` and `--gcc-install-dir` have been added to flang only recently and will be included in LLVM 19 (see https://github.com/llvm/llvm-project/pull/87360)

Crivella:
Changed `runtimes_cmake_args` into a list to allow modifications from multiple parts of the code (will be checked and added before each `configure` call.

Ensure correct python is detected also during the RUNTIMES CMake call

This python discussion overlaps with the discussion in:

- #3463

Crivella:
As an update everything should work now with LLVM 19.1.1 (manually tested with EC from https://github.com/easybuilders/easybuild-easyconfigs/pull/21611)

The problem related to https://github.com/llvm/llvm-project/issues/111667 should be fixed by [7151ec2](https://github.com/easybuilders/easybuild-easyblocks/pull/3373/commits/7151ec25aceaa8bb00598e017b0b62f9387202dc) adding `--unwindlib=none` to `CMAKE_EXE_LINKER_FLAGS` during the runtimes compilation

WIP:

- Enabling offloading increases the number of errors. Upon further inspection this is due to errors of the kind
  <details>
    <summary>Test report</summary>
  
    ```
    ******************** TEST 'libomptarget :: x86_64-pc-linux-gnu-LTO :: offloading/multiple_reductions_simple.c' FAILED ********************
    Exit Code: 1

    Command Output (stdout):
    --
    # RUN: at line 1
    /home/crivella/.local/easybuild/build/LLVM/19.1.1/GCCcore-13.3.0/llvm.obj.3/./bin/clang    -I /home/crivella/.local/easybuild/build/LLVM/19.1.1/GCCcore-13.3.0/llvm-project-19.1.1.src/offload/test -I  -L /home/crivella/.local/easybuild/build/LLVM/19.1.1/GCCcore-13.3.0/llvm.obj.3/runtimes/runtimes-bins/offload -L /home/crivella/.local/easybuild/build/LLVM/19.1.1/GCCcore-13.3.0/llvm.obj.3/./lib  -nogpulib -Wl,-rpath,/home/crivella/.local/easybuild/build/LLVM/19.1.1/GCCcore-13.3.0/llvm.obj.3/runtimes/runtimes-bins/offload -Wl,-rpath, -Wl,-rpath,/home/crivella/.local/easybuild/build/LLVM/19.1.1/GCCcore-13.3.0/llvm.obj.3/./lib -foffload-lto  -fopenmp-targets=x86_64-pc-linux-gnu /home/crivella/.local/easybuild/build/LLVM/19.1.1/GCCcore-13.3.0/llvm-project-19.1.1.src/offload/test/offloading/multiple_reductions_simple.c -o /home/crivella/.local/easybuild/build/LLVM/19.1.1/GCCcore-13.3.0/llvm.obj.3/runtimes/runtimes-bins/offload/test/x86_64-pc-linux-gnu-LTO/offloading/Output/multiple_reductions_simple.c.tmp /home/crivella/.local/easybuild/build/LLVM/19.1.1/GCCcore-13.3.0/llvm.obj.3/./lib/libomptarget.devicertl.a && /home/crivella/.local/easybuild/build/LLVM/19.1.1/GCCcore-13.3.0/llvm.obj.3/runtimes/runtimes-bins/offload/test/x86_64-pc-linux-gnu-LTO/offloading/Output/multiple_reductions_simple.c.tmp | /home/crivella/.local/easybuild/build/LLVM/19.1.1/GCCcore-13.3.0/llvm.obj.3/./bin/FileCheck /home/crivella/.local/easybuild/build/LLVM/19.1.1/GCCcore-13.3.0/llvm-project-19.1.1.src/offload/test/offloading/multiple_reductions_simple.c
    # executed command: /home/crivella/.local/easybuild/build/LLVM/19.1.1/GCCcore-13.3.0/llvm.obj.3/./bin/clang -I /home/crivella/.local/easybuild/build/LLVM/19.1.1/GCCcore-13.3.0/llvm-project-19.1.1.src/offload/test -I -L /home/crivella/.local/easybuild/build/LLVM/19.1.1/GCCcore-13.3.0/llvm.obj.3/runtimes/runtimes-bins/offload -L /home/crivella/.local/easybuild/build/LLVM/19.1.1/GCCcore-13.3.0/llvm.obj.3/./lib -nogpulib -Wl,-rpath,/home/crivella/.local/easybuild/build/LLVM/19.1.1/GCCcore-13.3.0/llvm.obj.3/runtimes/runtimes-bins/offload -Wl,-rpath, -Wl,-rpath,/home/crivella/.local/easybuild/build/LLVM/19.1.1/GCCcore-13.3.0/llvm.obj.3/./lib -foffload-lto -fopenmp-targets=x86_64-pc-linux-gnu /home/crivella/.local/easybuild/build/LLVM/19.1.1/GCCcore-13.3.0/llvm-project-19.1.1.src/offload/test/offloading/multiple_reductions_simple.c -o /home/crivella/.local/easybuild/build/LLVM/19.1.1/GCCcore-13.3.0/llvm.obj.3/runtimes/runtimes-bins/offload/test/x86_64-pc-linux-gnu-LTO/offloading/Output/multiple_reductions_simple.c.tmp /home/crivella/.local/easybuild/build/LLVM/19.1.1/GCCcore-13.3.0/llvm.obj.3/./lib/libomptarget.devicertl.a
    # .---command stderr------------
    # | clang: error: '-fopenmp-targets' must be used in conjunction with a '-fopenmp' option compatible with offloading; e.g., '-fopenmp=libomp' or '-fopenmp=libiomp5'
    # `-----------------------------
    # error: command failed with exit status: 1
    ```

  </details>

  Will investigate if there is some option to fix this or if it is a bug and the CMake files need a patch
- Once the aforementioned is fixed the number of allowed failures could be pinned more precisely in an EasyConfig files

Micket:
Related? https://github.com/llvm/llvm-project/issues/90333

Crivella:
> Related? [llvm/llvm-project#90333](https://github.com/llvm/llvm-project/issues/90333)

Seems to be related, the offload CMake does not seem to be properly configuring `OPENMP_TEST_OPENMP_FLAGS`. Tried to manually patch the `-fopenmp` but then this results in a `omp.h` not found, so I guess I will also have to patch the location of the header during the test

Crivella:
@Thyre 
In reference to:

- https://github.com/easybuilders/easybuild-easyblocks/pull/3480#issuecomment-2410589255

Modularized functionalities related to detecting/setting/sanity_checking `gcc_prefix` in tre functions (https://github.com/easybuilders/easybuild-easyblocks/pull/3373/commits/997a689f041ace9b9e79a1ee6b0ac34aa237a185):

- get_gcc_prefix
- create_compiler_config_file
- sanity_check_gcc_prefix

Tested with EC from:

- https://github.com/easybuilders/easybuild-easyconfigs/pull/21611

Crivella:
@Micket The solution was slightly more involved but atleast on x86_64 was able to make a patch file (needs testing on other archs and gpus).
Opened an issue on llvm, hope it will get picked up and maybe fix the problem at the source in a more robust way

- https://github.com/llvm/llvm-project/issues/112210

Thyre:
> @Thyre In reference to:
> 
>     * [Enhance AOCC EasyBlock to correctly pass GCC toolchain and compiler driver #3480 (comment)](https://github.com/easybuilders/easybuild-easyblocks/pull/3480#issuecomment-2410589255)
> 
> 
> Modularized functionalities related to detecting/setting/sanity_checking `gcc_prefix` in tre functions ([997a689](https://github.com/easybuilders/easybuild-easyblocks/commit/997a689f041ace9b9e79a1ee6b0ac34aa237a185)):
> 
>     * get_gcc_prefix
> 
>     * create_compiler_config_file
> 
>     * sanity_check_gcc_prefix
> 
> 
> Tested with EC from:
> 
>     * [{compiler}[SYSTEM,GCCcore/13.3.0] LLVM 19.1.1 easybuild-easyconfigs#21611](https://github.com/easybuilders/easybuild-easyconfigs/pull/21611)

Great, thanks a lot! Once this PR gets merged, we should probably cleanup the AOCC EasyBlock and maybe include these checks and/or settings in the other LLVM-based compilers.

Thyre:
As a (very early) heads-up: LLVM 20 will include the rename of `flang-new` to `flang` (https://github.com/llvm/llvm-project/pull/110023). A symlink to `flang-new` still exists.

Crivella:
Rebased onto 5.0.x

easybuild/easyblocks/l/llvm.py
```diff
@@ -60,65 +271,938 @@ def __init__(self, *args, **kwargs):
...
+        else:
+            self.log.info("Building LLVM")
+            print_msg("Building stage 1/1")
+        # change_dir(self.llvm_obj_dir_stage1)
+        # super(EB_LLVM, self).build_step(verbose, path)
+        import shutil
+        change_dir(self.builddir)
+        print_msg("TESTING!!!: Copying from previosu build (REMOVE ME)")
+        shutil.rmtree('llvm.obj.1', ignore_errors=True)
+        shutil.copytree(os.path.join('..', 'llvm.obj.1'), 'llvm.obj.1')
```
boegel, line 831: @Crivella This looks like some testing stuff that needs to be cleaned up?

boegel (changes requested):

easybuild/easyblocks/l/llvm.py
Crivella: Yeah sorry ended up committing it by mistake.
Removed the comments and kept it as a stash in case i will still need them.
It was used to copy previous stage 1 and 2 builds when i was doing changes only to stage 3 to speed up the builds

easybuild/easyblocks/l/llvm.py
```diff
@@ -60,65 +225,972 @@ def __init__(self, *args, **kwargs):
...
+
+        change_dir(basedir)
+        lib_path = ''
+        if self.cfg['build_runtimes']:
+            lib_dir_runtime = self.get_runtime_lib_path(basedir, fail_ok=False)
+            lib_path = os.path.join(basedir, lib_dir_runtime)
+        old_cflags = os.getenv('CFLAGS', '')
+        old_cxxflags = os.getenv('CXXFLAGS', '')
+        if build_option('rpath'):
+            setvar('CFLAGS', "%s %s" % (old_cflags, '-Wno-unused-command-line-argument'))
```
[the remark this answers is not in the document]
Crivella: Done in https://github.com/easybuilders/easybuild-easyblocks/pull/3373/commits/489b026e882a4bb96b09414adfea22cf64257118

boegel (approved):
Extensively tested in https://github.com/easybuilders/easybuild-easyconfigs/pull/20902, it's high time to get this merged.

@Crivella Thanks a lot for all your efforts on this, and sorry for the holdup...

PR #3374, opened 2024-06-27
update `PythonPackage` easyblock to allow installation of Python packages with `$PIP_REQUIRE_VIRTUALENV` set + move temporary pip folder into build dir

Flamefire:
Using the tempdir usually works but the contents of that folder are usually build artifacts.
As they might become rather large use the builddir for $XDG_CACHE_HOME instead.
As an alternative `os.path.abspath(build_path())` could be used but that might clash with parallel installations.

The user using EasyBuild might have $PIP_REQUIRE_VIRTUALENV set which will fail the installation of PythonPackages with
> ERROR: Could not find an activated virtualenv (required).

So disable this.
pip 20.2.3+ seems to require `false` as the value instead of any other value (including the empty string)


boegel (approved):
lgtm

boegel:
Test report by @boegel
 * **SUCCESS** _matplotlib-3.8.2-gfbf-2023b.eb_ 

node3158.skitty.os - Linux RHEL 8.8, x86_64, Intel(R) Xeon(R) Gold 6140 CPU @ 2.30GHz, Python 3.6.8
See https://gist.github.com/boegel/a8414a50350aa1562e768e8edb044e0d for a full test report.

PR #3380, opened 2024-07-04
add custom easyblock for Bandicoot

tanmoy1989:
Adding this `bandicoot.py` easyblock to install [Bandicoot](https://coot.sourceforge.io/) library using [Bandicoot-1.14.0-foss-2023b.eb (easyconfigs PR #20962)](https://github.com/easybuilders/easybuild-easyconfigs/pull/20962).

### DIFF
--- a/easybuild/easyblocks/b/bandicoot.py
+++ b/easybuild/easyblocks/b/bandicoot.py
@@ -1,0 +1,78 @@
+##
+# Copyright 2009-2024 Ghent University
+#
+# This file is part of EasyBuild,
+# originally created by the HPC team of Ghent University (http://ugent.be/hpc/en),
+# with support of Ghent University (http://ugent.be/hpc),
+# the Flemish Supercomputer Centre (VSC) (https://www.vscentrum.be),
+# Flemish Research Foundation (FWO) (http://www.fwo.be/en)
+# and the Department of Economy, Science and Innovation (EWI) (http://www.ewi-vlaanderen.be/en).
+#
+# https://github.com/easybuilders/easybuild
+#
+# EasyBuild is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation v2.
+#
+# EasyBuild is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with EasyBuild.  If not, see <http://www.gnu.org/licenses/>.
+##
+"""
+EasyBuild support for Bandicoot, implemented as an easyblock
+
+@author: Kenneth Hoste (Ghent University)
+@author: Tanmoy Chakraborty (University of Warwick)
+"""
+import os
+from easybuild.tools import LooseVersion
+from easybuild.easyblocks.generic.cmakemake import CMakeMake
+from easybuild.tools.build_log import EasyBuildError
+from easybuild.tools.modules import get_software_root
+from easybuild.tools.systemtools import get_shared_lib_ext
+
+
+class EB_Bandicoot(CMakeMake):
+    """Support for building Bandicoot."""
+
+    @staticmethod
+    def extra_options():
+        """Extra easyconfig parameters for Bandicoot."""
+        extra_vars = CMakeMake.extra_options()
+        extra_vars['separate_build_dir'][0] = True
+        return extra_vars
+
+    def configure_step(self):
+        """Set some extra environment variables before configuring."""
+
+        boost = get_software_root('Boost')
+        if not boost:
+            raise EasyBuildError("Dependency module Boost not loaded?")
+
+        self.cfg.update('configopts', "-DBoost_DIR=%s" % boost)
+        self.cfg.update('configopts', "-DBOOST_INCLUDEDIR=%s/include" % boost)
+        self.cfg.update('configopts', "-DBoost_DEBUG=ON -DBOOST_ROOT=%s" % boost)
+
+        self.cfg.update('configopts', '-DBLAS_LIBRARY:PATH="%s"' % os.getenv('LIBBLAS'))
+        self.cfg.update('configopts', '-DLAPACK_LIBRARY:PATH="%s"' % os.getenv('LIBLAPACK'))
+        self.cfg.update('configopts', '-DFIND_OPENCL=OFF')
+
+        super(EB_Bandicoot, self).configure_step()
+
+    def sanity_check_step(self):
+        """Custom sanity check for Bandicoot."""
+        if LooseVersion(self.version) < LooseVersion('7.950.1'):
+            libdir = 'lib'
+        else:
+            libdir = 'lib64'
+
+        custom_paths = {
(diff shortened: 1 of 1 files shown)


tanmoy1989:
Any updates on this, please?

easybuild/easyblocks/b/bandicoot.py
```diff
@@ -0,0 +1,77 @@
...
+
+
+class EB_Bandicoot(CMakeMake):
+    """Support for building Bandicoot."""
+
+    @staticmethod
+    def extra_options():
+        """Extra easyconfig parameters for Bandicoot."""
+        extra_vars = CMakeMake.extra_options()
+        extra_vars['separate_build_dir'][0] = True
```
boegel, line 46: @tanmoy1989 This is already the default for `CMakeMake`, so there's no need for customizing `extra_options` at all?

easybuild/easyblocks/b/bandicoot.py
```diff
@@ -0,0 +1,77 @@
...
+        boost = get_software_root('Boost')
+        if not boost:
+            raise EasyBuildError("Dependency module Boost not loaded?")
+
+        self.cfg.update('configopts', "-DBoost_DIR=%s" % boost)
+        self.cfg.update('configopts', "-DBOOST_INCLUDEDIR=%s/include" % boost)
+        self.cfg.update('configopts', "-DBoost_DEBUG=ON -DBOOST_ROOT=%s" % boost)
+
+        self.cfg.update('configopts', '-DBLAS_LIBRARY:PATH="%s"' % os.getenv('LIBBLAS'))
+        self.cfg.update('configopts', '-DLAPACK_LIBRARY:PATH="%s"' % os.getenv('LIBLAPACK'))
```
boegel, line 61: `PATH=` seems wrong here, since `$LIBBLAS` is a list of options to link to libraries (like `-lfoo`)?

easybuild/easyblocks/b/bandicoot.py
```diff
@@ -0,0 +1,77 @@
...
+        self.cfg.update('configopts', '-DFIND_OPENCL=OFF')
+
+        super(EB_Bandicoot, self).configure_step()
+
+    def sanity_check_step(self):
+        """Custom sanity check for Bandicoot."""
+        if LooseVersion(self.version) < LooseVersion('7.950.1'):
+            libdir = 'lib'
+        else:
+            libdir = 'lib64'
```
boegel, line 71: EasyBuild automatically symlinks `lib` to `lib64` and vice versa, so there's no need for this version check at all, just always use `lib`

easybuild/easyblocks/b/bandicoot.py
```diff
@@ -0,0 +1,77 @@
...
+
+    def configure_step(self):
+        """Set some extra environment variables before configuring."""
+
+        boost = get_software_root('Boost')
+        if not boost:
+            raise EasyBuildError("Dependency module Boost not loaded?")
+
+        self.cfg.update('configopts', "-DBoost_DIR=%s" % boost)
+        self.cfg.update('configopts', "-DBOOST_INCLUDEDIR=%s/include" % boost)
```
boegel, line 57:
```suggestion
        self.cfg.update('configopts', "-DBOOST_INCLUDEDIR=%s" % os.path.join(boost, 'include'))

**Review by boegel — CHANGES_REQUESTED** [review]

File: easybuild/easyblocks/b/bandicoot.py (hunk 1 above)
Comment (tanmoy1989) [reply easybuild/easyblocks/b/bandicoot.py]: Done

File: easybuild/easyblocks/b/bandicoot.py (hunk 3 above)
Comment (tanmoy1989) [reply easybuild/easyblocks/b/bandicoot.py]: Done

File: easybuild/easyblocks/b/bandicoot.py (hunk 2 above)
Comment (boegel) [reply easybuild/easyblocks/b/bandicoot.py]: Sorry for not getting back to you on this...

I think the `:PATH` part is a hint for CMake what type of value is provided.

The contents of the `$LIBLAPACK` environment variable is not a path (or list of paths) though, it's a set of compiler options like `-lflexiblas` to specify how to link to LAPACK.

Strictly speaking this is a bug in the Armadillo easyblock, but since it's just a hint for CMake, I guess it's actually harmless. Strictly speaking, it's wrong though, so I would get rid of it:

```suggestion
        self.cfg.update('configopts', '-DLAPACK_LIBRARY="%s"' % os.getenv('LIBLAPACK'))

File: easybuild/easyblocks/b/bandicoot.py [hunk 5]
```diff
@@ -0,0 +1,65 @@
...
+
+        boost = get_software_root('Boost')
+        if not boost:
+            raise EasyBuildError("Dependency module Boost not loaded?")
+
+        self.cfg.update('configopts', "-DBoost_DIR=%s" % boost)
+        self.cfg.update('configopts', "-DBOOST_INCLUDEDIR=%s" % os.path.join(boost, 'include'))
+        self.cfg.update('configopts', "-DBoost_DEBUG=ON -DBOOST_ROOT=%s" % boost)
+
+        self.cfg.update('configopts', '-DBLAS_LIBRARY:PATH="%s"' % os.getenv('LIBBLAS'))
```
boegel:
```suggestion
        self.cfg.update('configopts', '-DBLAS_LIBRARY="%s"' % os.getenv('LIBBLAS'))

File: easybuild/easyblocks/b/bandicoot.py [hunk 6]
```diff
@@ -0,0 +1,65 @@
+##
+# Copyright 2009-2024 Ghent University
```
boegel:
```suggestion
# Copyright 2009-2026 Ghent University
```

easybuild/easyblocks/b/bandicoot.py
```diff
@@ -0,0 +1,65 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with EasyBuild.  If not, see <http://www.gnu.org/licenses/>.
+##
+"""
+EasyBuild support for Bandicoot, implemented as an easyblock
+
+@author: Kenneth Hoste (Ghent University)
```
boegel: No need to list me as an author here, I'm not really
```suggestion
```

boegel (changes requested):

easybuild/easyblocks/b/bandicoot.py
```diff
@@ -0,0 +1,63 @@
...
+
+        boost = get_software_root('Boost')
+        if not boost:
+            raise EasyBuildError("Dependency module Boost not loaded?")
+
+        self.cfg.update('configopts', "-DBoost_DIR=%s" % boost)
+        self.cfg.update('configopts', "-DBOOST_INCLUDEDIR=%s" % os.path.join(boost, 'include'))
+        self.cfg.update('configopts', "-DBoost_DEBUG=ON -DBOOST_ROOT=%s" % boost)
+
+        self.cfg.update('configopts', '-DLAPACK_LIBRARY="%s"' % os.getenv('LIBLAPACK'))
```
boegel, line 52: we probably want to retain the `BLAS` equivalent here, so let's add that back:
```suggestion
        self.cfg.update('configopts', '-DBLAS_LIBRARY="%s"' % os.getenv('LIBBLAS'))
        self.cfg.update('configopts', '-DLAPACK_LIBRARY="%s"' % os.getenv('LIBLAPACK'))

boegel (changes requested):

boegel (approved):
lgtm

PR #3369, opened 2024-06-19
support >= 2024a `for MCR`

branfosj:
Fixes https://github.com/easybuilders/easybuild-easyconfigs/issues/20797

Successful build of the latest MCR with this PR is in https://github.com/easybuilders/easybuild-easyconfigs/pull/20860

PR #3389, opened 2024-07-09
modification on mcr.py 

aroudgar:
Please modify  line 87 of mcr.py file as follow:

    "agreeToLicense=Yes",

to

    "agreeToLicense=yes",

Apparently he new matlab compiler does not accept "Yes". It accept "yes" instead.

### DIFF
--- a/.github/workflows/linting.yml
+++ b/.github/workflows/linting.yml
@@ -1,6 +1,9 @@
 name: Static Analysis
 on: [push, pull_request]
 
+permissions:
+  contents: read # to fetch code (actions/checkout)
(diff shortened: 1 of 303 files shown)


aroudgar:
@branfosj yes, that is for computecanada. Sorry about that, however, I believe anyone who is intended  to install new MCR will hit this bug.

branfosj:
> @branfosj yes, that is for computecanada. Sorry about that, however, I believe anyone who is intended to install new MCR will hit this bug.

We found the same issue and fixed it in #3369

PR #3392, opened 2024-07-16
Disable dynamic memory tracing, which causes MPI tests fail in some systems

julianmorillo:
In some systems (see https://github.com/easybuilders/easybuild-easyconfigs/pull/20690) MPI tests cause a Segmentation Fault.
The problem comes from a bad interaction between the dynamic library loader of the target system and the tracing of dynamic memory (malloc, calloc, free, etc.). Extrae developers are looking into it, but in the meantime, just disable dynamic memory tracing.

### DIFF
--- a/easybuild/easyblocks/e/extrae.py
+++ b/easybuild/easyblocks/e/extrae.py
@@ -79,6 +79,9 @@ class EB_Extrae(ConfigureMake):
         # TODO: make this optional dependencies
         self.cfg.update('configopts', "--without-dyninst")
 
+        # Workaround to make MPI tests pass in some platforms
+        self.cfg.update('configopts', "--disable-instrument-dynamic-memory")
+
         # Needed to build in RISC-V architectures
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/e/extrae.py
```diff
@@ -79,6 +79,9 @@ def configure_step(self):
         # TODO: make this optional dependencies
         self.cfg.update('configopts', "--without-dyninst")
 
+        # Workaround to make MPI tests pass in some platforms
+        self.cfg.update('configopts', "--disable-instrument-dynamic-memory")
```
ocaisa, line 83: This is (hopefully) something version specific, so we might be better off putting this in the easyconfig for that version rather than making it permanent here (or at least add a version guard here)

julianmorillo: You are right. I put it in the easyconfig.

PR #3393, opened 2024-07-17
Extrae: force --without-unwind for ARM and add --with-libz option for all architectures

julianmorillo:
Extrae does not work with libunwind on ARM.


ocaisa:
Successful builds in https://github.com/easybuilders/easybuild-easyconfigs/pull/21017 , waiting on https://github.com/EESSI/software-layer/pull/640#issuecomment-2236622179 to confirm the issue is fixed on ARM

ocaisa (approved):

PR #3394, opened 2024-07-26
Remove redundant backslashes in GEANT easyblock

ocaisa:
This is causing CI to fail suddenly, see https://github.com/ocaisa/easybuild-easyblocks/actions/runs/10111159687/job/27962541425

There's no real way of testing this as it only applies to old (archived) versions of Geant4 (see https://github.com/easybuilders/easybuild-easyblocks/pull/3394/files#diff-709e5f8d7395f32564d55701217becbde05841477a60bbe696992a2488b19b8eR85)

### DIFF
--- a/.github/workflows/linting.yml
+++ b/.github/workflows/linting.yml
@@ -23,7 +23,7 @@ jobs:
       run: |
         pip install --upgrade pip
         # fix to this version for develop branch (to avoid needing to fix geant4.py)
-        pip install --upgrade "flake8<7.1"
+        pip install --upgrade "flake8"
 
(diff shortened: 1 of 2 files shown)


ocaisa:
CORE_CNT=16
EB_ARGS="Geant4-11.1.2-GCC-11.3.0.eb"

ocaisa:
Ah, I didn't notice this was only in my branch, closing

ocaisa:
This is danced around via https://github.com/easybuilders/easybuild-easyblocks/pull/3365 but it's not a hard thing to fix

branfosj (approved):

PR #3396, opened 2024-07-26
enhance custom easyblock for GCC to use `with-arch` option for nvptx with 13.1+

Thyre:
## Motivation

GCC, like other compilers, allows users to use offloading via OpenMP & OpenACC for example to utilize accelerators in their written programs. While some compilers require the presence of CUDA for this e.g. Clang, GCC has no requirement for it to simply build and run an executable containing offloading code.

By default, GCC targets a _very low_ architecture for NVIDIA GPUs though. In GCC 12.3.0, this was `sm_30`. In GCC 13.3.0, the default version is still the same, but [recent `nvptx-tools` can bump this to `sm_50`](https://github.com/SourceryTools/nvptx-tools/commit/13216705844e5d7c9d20b4e34f89a39017e191d4) when CUDA is detected. With this, GCC can work around the removal of `sm_3x` in more recent CUDA versions, avoiding the following error message:

**GCC 12.3.0**
```console
$ gcc -fopenmp -foffload=nvptx-none test.c
ptxas fatal   : Value 'sm_35' is not defined for option 'gpu-name'
nvptx-as: ptxas returned 255 exit status
mkoffload: fatal error: x86_64-pc-linux-gnu-accel-nvptx-none-gcc returned 1 exit status
compilation terminated.
lto-wrapper: fatal error: /p/software/fs/jurecadc/stages/2024/software/GCCcore/12.3.0/bin/../libexec/gcc/x86_64-pc-linux-gnu/12.3.0//accel/nvptx-none/mkoffload returned 1 exit status
compilation terminated.
/p/software/jurecadc/stages/2024/software/binutils/2.40-GCCcore-12.3.0/bin/ld: error: lto-wrapper failed
collect2: error: ld returned 1 exit status
```

**GCC 13.3.0**
```console
$ gcc --verbose -fopenmp -foffload=nvptx-none test.c
[...]
/p/software/fs/jurecadc/stages/2025/software/GCCcore/13.3.0/bin/../libexec/gcc/x86_64-pc-linux-gnu/13.3.0/accel/nvptx-none/lto1 -quiet -dumpbase ./a.xnvptx-none.mkoffload -m64 -mgomp -misa=sm_30 -version -fno-openacc -fno-pie -fcf-protection=none -foffload-abi=lp64 -fopenmp @/tmp/ccZrKG87 -o /tmp/ccbOTp8K.s
[...]
Verifying sm_30 code with sm_50 code generation.
 ptxas -c -o /dev/null /tmp/cc7PheNR.o --gpu-name sm_50 -O0
[...]
``` 

However, this may break once again as soon as NVIDIA decides to remove the already deprecated support for `sm_50` (in CUDA 11.0). Fortunately, GCC has added a configure option to overwrite the default nvptx architecture. Beginning with GCC 13.1.0, one can pass `--with-arch=sm_[x]` to set the default option, as long as GCC can understand it.

In addition, choosing a newer architecture by default might bring performance improvements and access to additional features.

## Scope of this PR

This pull request adds the new option `--with-arch=sm_[x]` to GCC builds starting with GCC 13.1.0 if offloading support via nvptx is enabled. To choose which architecture is being passed, a new function named `map_nvptx_capability` is implemented. This function retrieves `cuda_compute_capabilities` and matches them against the official GCC mappings (which can be found in [`${GCC_SRC}/gcc/config/nvptx/nvptx.opt`](https://github.com/gcc-mirror/gcc/blob/master/gcc/config/nvptx/nvptx.opt))  being used for the `-march-map=` argument. 

Since GCC only allows to set a single default architecture, I decided to use the lowest one available. For example, JURECA-DC sets both 7.5 and 8.0 for EasyBuild. Therefore, 7.5 would be chosen. 
If parsing the architecture mappings fails, for example because the file layout changed or the file was moved, a warning is returned. In this case, we stick to the default of GCC. This is also the case if the architectures in `cuda_compute_capabilities` cannot be mapped at all. This makes the additions more resilient to upstream changes. 

Generally, this helps users as they are not required to pass architectures manually every single time as it is the case with CUDA 12 + GCC 12.3.0 right now. Here, one would need to pass `-foffload-options=-misa=sm_80`.

### DIFF
--- a/easybuild/easyblocks/g/gcc.py
+++ b/easybuild/easyblocks/g/gcc.py
@@ -366,6 +366,70 @@ class EB_GCC(ConfigureMake):
             'versions': versions
         }
 
+    def map_nvptx_capability(self):
+        """
+        Convert PTX ISA architecture passed via EasyBuild configs to a version which is understood by GCC.
+        Valid architecture strings include 'sm_30', 'sm_35', 'sm_53', 'sm_70', 'sm_75', 'sm_80'.
+        (as of GCC 14.1.0). Some additional architectures may be mapped.
+        See: https://github.com/gcc-mirror/gcc/commit/de0ef04419e90eacf0d1ddb265552a1b08c18d4b
+
+        As this list is updated regularly, try to parse the GCC source file (gcc/config/nvptx/nvptx.opt)
+        and extract the supported architectures and their mapping. Based on the result, determine the lowest
+        architecture required to support all 'cuda_compute_capabilities' and return this value.
+        """
+        cuda_cc_list = build_option('cuda_compute_capabilities') or self.cfg['cuda_compute_capabilities']
+        architecture_mappings_file = os.path.join(self.cfg['start_dir'], 'gcc', 'config', 'nvptx', 'nvptx.opt')
+        architecture_mappings_flag = "march-map="
+        architecture_mappings_replacement = "misa=,"
+
+        # Determine which compute capabilities are configured. If there are none, return immediately.
+        if cuda_cc_list is None:
+            return None
+        cuda_sm_list = [f"sm_{cc.replace('.', '')}" for cc in cuda_cc_list]
+
+        if not os.path.exists(architecture_mappings_file):
+            warn_msg = f"Tried to parse nvptx.opt but file {architecture_mappings_file} was not found. " \
+                       "Please check the path and update the EasyBlock if necessary!"
+            self.log.warning(warn_msg)
+            return None
+
+        # We want to read the mappings found in the GCC sources and create a map for this.
+        # We're searching for the following pattern:
+        # march-map=sm_32
+        # Target RejectNegative Alias(misa=,sm_30)
+        gcc_architecture_mappings = {}
+        file_content = read_file(architecture_mappings_file).splitlines()
+        for line_idx, line in enumerate(file_content):
+            line = line.strip()
+            if line.startswith(architecture_mappings_flag):
+                key = line.split('=')[1]
+                # Mapped architecture can be found in the following line
+                line = file_content[line_idx + 1]
+                if architecture_mappings_replacement not in line:
+                    warn_msg = "Tried to parse nvptx.opt but failed to extract mapped architectures! " \
+                                f"Expected to find substring '{architecture_mappings_replacement}' in " \
+                                f"line {line_idx + 1} but found '{line}'. Choosing default of GCC."
+                    self.log.warning(warn_msg)
+                    # Bail out, since results of mapping cannot be trusted
+                    return None
+                value = line.split(architecture_mappings_replacement)[1].rstrip(')')
+                gcc_architecture_mappings[key] = value
+        self.log.info(f"Available architecture mappings in GCC {self.version}:\n {str(gcc_architecture_mappings)}")
+
+        # Map compute capabilities to GCC ones
+        # If no compute capability can be mapped, stick to default of GCC and return None
+        gcc_cc = [gcc_architecture_mappings[cc] if cc in gcc_architecture_mappings else None for cc in cuda_sm_list]
+        self.log.info(f"Mapped architectures: {str(cuda_sm_list)} -> {str(gcc_cc)}")
+        if any(cc is None for cc in gcc_cc):
+            self.log.info("At least one architecture could not be mapped. Choosing default of GCC.")
+            return None
+
+        # Get the lowest architecture mapping and return it
+        sorted_gcc_cc = sorted(gcc_cc)
+        self.log.info("Choosing first architecture in sorted list as default nvptx "
+                      f"architecture: {str(sorted_gcc_cc)}")
+        return sorted_gcc_cc[0]
+
     def run_configure_cmd(self, cmd):
         """
         Run a configure command, with some extra checking (e.g. for unrecognized options).
@@ -544,6 +608,10 @@ class EB_GCC(ConfigureMake):
                         self.create_dir("build-nvptx-gcc")
                         target = 'nvptx-none'
                         self.cfg.update('configopts', "--enable-newlib-io-long-long")
+                        if LooseVersion(self.version) >= LooseVersion('13.1.0'):
+                            cuda_cc = self.map_nvptx_capability()
+                            if cuda_cc:
+                                self.cfg.update('configopts', f'--with-arch={cuda_cc}')
                     else:
                         # compile AMD GCN target compiler
                         self.create_dir("build-amdgcn-gcc")


SebastianAchilles:
Test report by @sebastianachilles
 * **SUCCESS** _GCCcore-12.3.0.eb_ 
 * **SUCCESS** _GCCcore-13.1.0.eb_ 
 * **SUCCESS** _GCCcore-13.2.0.eb_ 
 * **SUCCESS** _GCCcore-13.3.0.eb_ 
 * **SUCCESS** _GCCcore-14.1.0.eb_ 

jscclxc1.int.jsc-clx.fz-juelich.de - Linux Rocky Linux 9.4, x86_64, Intel Xeon Processor (Cascadelake) (cascadelake), Python 3.9.18
See https://gist.github.com/SebastianAchilles/6890d9cc1f0024fe7541e064ba5009f8 for a full test report.

easybuild/easyblocks/g/gcc.py
```diff
@@ -366,6 +366,63 @@ def prep_extra_src_dirs(self, stage, target_prefix=None):
...
+        """
+        # Determine which compute capabilities are configured. If there are none, return immediately.
+        cuda_cc_list = build_option('cuda_compute_capabilities') or self.cfg['cuda_compute_capabilities']
+        if cuda_cc_list is None:
+            return None
+        cuda_sm_list = [f"sm_{cc.replace('.', '')}" for cc in cuda_cc_list]
+
+        architecture_mappings_file = os.path.join(self.cfg['start_dir'], 'gcc', 'config', 'nvptx', 'nvptx.opt')
+        if not os.path.exists(architecture_mappings_file):
+            warn_msg = "Tried to parse nvptx.opt but file was not found. " \
```
boegel, line 388: Please mention path that was considered in the warning message

easybuild/easyblocks/g/gcc.py
```diff
@@ -366,6 +366,63 @@ def prep_extra_src_dirs(self, stage, target_prefix=None):
...
+            self.log.warning(warn_msg)
+            return None
+
+        # We want to read the mappings found in the GCC sources and create a map for this.
+        # We're searching for the following pattern:
+        # march-map=sm_32
+        # Target RejectNegative Alias(misa=,sm_30)
+        gcc_architecture_mappings = {}
+        with open(architecture_mappings_file) as file:
+            file_content = file.read().splitlines()
```
boegel, line 399: just use the `read_file` function instead (better error reporting):

```python
file_content = read_file(architecture_mappings_file).splitlines()

File: easybuild/easyblocks/g/gcc.py [hunk 3]
```diff
@@ -366,6 +366,63 @@ def prep_extra_src_dirs(self, stage, target_prefix=None):
...
+            for line_idx in range(len(file_content)):
```
boegel, line 400:
```python
for line_idx, line in enumerate(file_content):

File: easybuild/easyblocks/g/gcc.py [hunk 4]
```diff
@@ -366,6 +366,63 @@ def prep_extra_src_dirs(self, stage, target_prefix=None):
...
+                line = file_content[line_idx].strip()
+                if line.startswith("march-map="):
+                    key = line.split('=')[1]
+                    # Mapped architecture can be found in the following line
+                    line = file_content[line_idx + 1]
+                    if "misa=," not in line:
+                        warn_msg = "Tried to parse nvptx.opt but failed to extract mapped architectures! " \
```
boegel, line 407: Maybe try to provide more context on when you gave up, like mention the line number and some clue as to why the format isn't what you expect?

easybuild/easyblocks/g/gcc.py
```diff
@@ -366,6 +366,63 @@ def prep_extra_src_dirs(self, stage, target_prefix=None):
...
+                                   "Choosing default of GCC."
+                        self.log.warning(warn_msg)
+                        # Bail out, since results of mapping cannot be trusted
+                        return None
+                    value = line.split('misa=,')[1][:-1]
```
boegel, line 412: Maybe better to avoid surprises, only strip off trailing `)` rather than cutting off last character?

```python
value = line.split('misa=,')[1].rstrip(')')

File: easybuild/easyblocks/g/gcc.py [hunk 6]
```diff
@@ -366,6 +366,63 @@ def prep_extra_src_dirs(self, stage, target_prefix=None):
...
+                    gcc_architecture_mappings[key] = value
+        self.log.info(f"Available architecture mappings: {str(gcc_architecture_mappings)}")
+
+        # Map compute capabilities to GCC ones
+        # If no compute capability can be mapped, stick to default of GCC and return None
+        gcc_cc = [gcc_architecture_mappings[cc] if cc in gcc_architecture_mappings else None for cc in cuda_sm_list]
+        gcc_cc = [cc for cc in gcc_cc if cc is not None]
+        if not gcc_cc:
```
boegel, line 420:
```python
if any(cc is None for cc in gcc_cc):
    return None

File: easybuild/easyblocks/g/gcc.py [hunk 7]
```diff
@@ -366,6 +366,63 @@ def prep_extra_src_dirs(self, stage, target_prefix=None):
...
+            return None
+
+        # Get the lowest architecture mapping and return it
+        return sorted(gcc_cc)[0]
```
boegel, line 424: Maybe log sorted entries before returning, in case the sorting isn't what we expect somehow?

boegel (changes requested):

Thyre:
Thanks a lot for the review. I agree with your comments and am working on adding them to the PR.

easybuild/easyblocks/g/gcc.py
Thyre: Warning will now show:
```
== 2024-08-01 11:30:38,548 gcc.py:390 WARNING Tried to parse nvptx.opt but file /dev/shm/reuter1/jedi/GCCcore/14.1.0/system-system/gcc-14.1.0/gcc/config/nvptxa/nvptx.opt was not found. Please check the path and update the EasyBlock if necessary!
```

easybuild/easyblocks/g/gcc.py
Thyre: Changed to:
```python
        # Get the lowest architecture mapping and return it
        sorted_gcc_cc = sorted(gcc_cc)
        self.log.info("Choosing first architecture in sorted list as default nvptx " \
                      f"architecture: {str(sorted_gcc_cc)}")
        return sorted_gcc_cc[0]
```

easybuild/easyblocks/g/gcc.py
Thyre: Changed to:
```python
                line = file_content[line_idx + 1]
                if architecture_mappings_replacement not in line:
                    warn_msg = "Tried to parse nvptx.opt but failed to extract mapped architectures! " \
                                f"Expected to find substring '{architecture_mappings_replacement}' in " \
                                f"line {line_idx + 1} but found '{line}'. Choosing default of GCC."
                    self.log.warning(warn_msg)
                    # Bail out, since results of mapping cannot be trusted
                    return None
```

Thyre: Warning will now show (changed searched pattern to trigger warning):
```
== 2024-08-01 12:06:18,969 gcc.py:409 WARNING Tried to parse nvptx.opt but failed to extract mapped architectures! Expected to find substring 'misa=,' in line 63 but found 'Target RejectNegative Alias(misa=,sm_30)'. Choosing default of GCC.
```

easybuild/easyblocks/g/gcc.py
Thyre: This info will now be shown:
```
== 2024-08-01 12:42:09,984 gcc.py:428 INFO Choosing first architecture in sorted list as default nvptx architecture: ['sm_80']
```

SebastianAchilles:
Test report by @sebastianachilles
 * **SUCCESS** _GCCcore-12.3.0.eb_ 
 * **SUCCESS** _GCCcore-13.1.0.eb_ 
 * **SUCCESS** _GCCcore-13.2.0.eb_ 
 * **SUCCESS** _GCCcore-13.3.0.eb_ 
 * **SUCCESS** _GCCcore-14.1.0.eb_ 

jscclxc1.int.jsc-clx.fz-juelich.de - Linux Rocky Linux 9.4, x86_64, Intel Xeon Processor (Cascadelake) (cascadelake), Python 3.9.18
See https://gist.github.com/SebastianAchilles/2b79825a144baa828025421933035a68 for a full test report.

easybuild/easyblocks/g/gcc.py
```diff
@@ -366,6 +366,69 @@ def prep_extra_src_dirs(self, stage, target_prefix=None):
...
+                    self.log.warning(warn_msg)
+                    # Bail out, since results of mapping cannot be trusted
+                    return None
+                value = line.split(architecture_mappings_replacement)[1].rstrip(')')
+                gcc_architecture_mappings[key] = value
+        self.log.info(f"Available architecture mappings in GCC {self.version}:\n {str(gcc_architecture_mappings)}")
+
+        # Map compute capabilities to GCC ones
+        # If no compute capability can be mapped, stick to default of GCC and return None
+        gcc_cc = [gcc_architecture_mappings[cc] if cc in gcc_architecture_mappings else None for cc in cuda_sm_list]
```
boegel, line 421: We should also log the result of this, along with the contents of `cuda_sm_list` (above), otherwise we don't really know which mapping failed based on log info?

boegel (changes requested):

boegel:
EB_ARGS="GCCcore-10.2.0.eb GCCcore-12.3.0.eb GCCcore-14.2.0.eb --installpath /tmp/$USER/pr-3396"

boegel (approved):

boegel:
Test report by @boegel
 * **SUCCESS** _GCCcore-12.3.0.eb_ 
 * **SUCCESS** _GCCcore-13.2.0.eb_ 
 * **SUCCESS** _zlib-1.3.1.eb_ 
 * **SUCCESS** _binutils-2.42.eb_ 
 * **SUCCESS** _GCCcore-14.2.0.eb_ 

Build succeeded for 5 out of 5 (3 easyconfigs in total)
node3900.accelgor.os - Linux RHEL 8.8, x86_64, AMD EPYC 7413 24-Core Processor, 1 x NVIDIA NVIDIA A100-SXM4-80GB, 545.23.08, Python 3.6.8
See https://gist.github.com/boegel/1dcd0f4c7656396fcdacc42dfa4f04f7 for a full test report.

PR #3400, opened 2024-07-31
change custom easyblock for Python to set `pythonpath` class variable to a fixed location relative to build path

gkaf89:
The `pythonpath` variable in the `python.py` EasyBlock is used to create temporary directories to store build artifacts. However, there are 2 problems with the current definition of the `pythonpath` instance field:

- it uses the log directory to store build artifacts, and
- the functions that use `pythonpath` do not accept a full path.

The `build_path()` function cannot be used instead of `log_path()` as the function provides a full path as well. Thus we opted to hard-code a relative path with respect to the build directory.

Issue: https://github.com/easybuilders/easybuild-easyconfigs/issues/21078

### DIFF
--- a/easybuild/easyblocks/p/python.py
+++ b/easybuild/easyblocks/p/python.py
@@ -45,7 +45,7 @@ from easybuild.easyblocks.generic.configuremake import ConfigureMake
 from easybuild.framework.easyconfig import CUSTOM
 from easybuild.framework.easyconfig.templates import PYPI_SOURCE
 from easybuild.tools.build_log import EasyBuildError, print_warning
-from easybuild.tools.config import build_option, ERROR, log_path, PYTHONPATH, EBPYTHONPREFIXES
+from easybuild.tools.config import build_option, ERROR, PYTHONPATH, EBPYTHONPREFIXES
 from easybuild.tools.modules import get_software_libdir, get_software_root, get_software_version
 from easybuild.tools.filetools import apply_regex_substitutions, change_dir, mkdir
 from easybuild.tools.filetools import read_file, remove_dir, symlink, write_file
@@ -142,8 +142,8 @@ class EB_Python(ConfigureMake):
 
         self.pyshortver = '.'.join(self.version.split('.')[:2])
 
-        # Used for EBPYTHONPREFIXES handler script
-        self.pythonpath = os.path.join(log_path(), 'python')
+        # Path of the directory used to store the EBPYTHONPREFIXES handler script, relative to the install directory
+        self.python_script_dir_relative_path = os.path.join('etc', 'python')
 
         ext_defaults = {
             # Use PYPI_SOURCE as the default for source_urls of extensions.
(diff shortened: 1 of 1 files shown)


boegel:
@gkaf89 Thanks a lot for your in-depth analysis in https://github.com/easybuilders/easybuild-easyconfigs/issues/21078, that's very helpful.

I'm not sure the proposed fix is correct though...
I think you're seeing a secondary effect of another problem that is likely to cause trouble somewhere else too.

I'll need some time to deep dive into this though, and fully understand what's going on.

gkaf89:
With the proposed modification the following file directory structure is created in the Python installation directory:
```
easybuild/python/sitecustomize.py
```

The `easybuild` normally contains the log files. If `sitecustomize.py` script is part of the log files, then the commit indeed has the potential to break any system that depends using `sitecustomize.py` from the log files.

I should modify this commit to copy the Python path in the log file directory.

Flamefire:
@boegel Independent of #3493 this looks correct to me as it is merely a semantic change without any actual change as far as I can tell. I.e. the usage of `log_path` at that point isn't correct from a semantic POV

Flamefire:
BTW: Can you change the title and description to make it fit better:

- The path is relative to the install directory not the build path
- The variable is **not** "used to create temporary directories to store build artifacts". There is nothing temporary about `pythonpath`
- "the functions that use pythonpath do not accept a full path." -> "... expect a relative path"
- The `build_path()` function -> Again nothing about the build path here, it is always used (more or less) relative to the install path

This would help better understanding the change some time later

easybuild/easyblocks/p/python.py
```diff
@@ -143,7 +143,8 @@ def __init__(self, *args, **kwargs):
         self.pyshortver = '.'.join(self.version.split('.')[:2])
 
         # Used for EBPYTHONPREFIXES handler script
-        self.pythonpath = os.path.join(log_path(), 'python')
+        script_directory = 'etc'
+        self.pythonpath = os.path.join(script_directory, 'python')
```
Flamefire, line 147: The variable is only used once and I don't understand the name: We put a script inside `pythonpath`, so why is there another `script_directory`?

```suggestion
        self.pythonpath = os.path.join('etc', 'python')
```

easybuild/easyblocks/p/python.py
```diff
@@ -143,7 +143,8 @@ def __init__(self, *args, **kwargs):
         self.pyshortver = '.'.join(self.version.split('.')[:2])
 
         # Used for EBPYTHONPREFIXES handler script
```
Flamefire, line 145:
```suggestion
        # Path of the directory used to store the EBPYTHONPREFIXES handler script, relative to the install directory
```

easybuild/easyblocks/p/python.py
gkaf89: I removed the extra variable as suggested.

easybuild/easyblocks/p/python.py
gkaf89: Added the comment.

gkaf89:
The commit is reorganized according to the suggestions. All modification are made in a single commit so that they are easier to track. The underlying issue is that `sitecustomize.py` is an environment configuration script, and as such it should be stored under an `etc` directory in the installation directory of Python.

I was a bit confused about the role of `sitecustomize.py` as it was stored in the directory containing the EasyBuild installation logs.

Flamefire:
This now got superseded by getting rid of the path: https://github.com/easybuilders/easybuild-easyblocks/pull/3514

This PR would break `--module-only` builds as the path changed. So I'd go with the other that takes care of that too

boegel:
Indeed, let's go forward with the approach proposed in #3514, so closing...

PR #3405, opened 2024-08-06
support revisions in crates extraction procedure in `Cargo` easyblock

Flamefire:
There might be entries like
> source = "git+https://github.com/charliermarsh/rs-async-zip?rev=011b24604fa7bc223daaad7712c0694bac8f0a87#011b24604fa7bc223daaad7712c0694bac8f0a87"

We do handle "?branch=sha#sha" but not "?rev=sha#sha", so extend this to "rev". To be safe match the revision.


boegel (approved):
lgtm

PR #3406, opened 2024-08-07
fix double initialization of `Cargo` by `CargoPythonPackage` by removing incorrect custom `__init__` implementation + fix use of `super()` in PALM easyblock (since that doesn't work with Python 2.7)

Flamefire:
The issue intended to be fixed with #2934 is actually (likely) caused by Pythons MRO:

CargoPythonPackage explicitely calls `__init__` of **both** `Cargo` and `PythonPackage`
1. `super().__init__` in Cargo calls `ExtensionEasyBlock->EasyBlock` stop (Could then call `Extension` but `ExtensionEasyBlock does not use `super`)
2. Then `PythonPackage.__init__` is called by `CargoPythonPackage`
3. `super().__init__` in PythonPackage calls `Cargo.__init__` again(!)
4. Then all other methods are called again as per 1.

The MRO order is
`[CargoPythonPackage, PythonPackage, Cargo, ExtensionEasyBlock, EasyBlock, Extension]` which explains that.

Fix is to consistently use `super()` in the CargoPythonPackage inheritance chain.

### DIFF
--- a/easybuild/easyblocks/generic/cargopythonpackage.py
+++ b/easybuild/easyblocks/generic/cargopythonpackage.py
@@ -47,10 +47,6 @@ class CargoPythonPackage(PythonPackage, Cargo):  # PythonPackage must come first
 
         return extra_vars
 
-    def __init__(self, *args, **kwargs):
-        """Constructor for CargoPythonPackage easyblock."""
-        super().__init__(*args, **kwargs)
-
     def extract_step(self):
         """Specifically use the overloaded variant from Cargo as is populates vendored sources with checksums."""
         return Cargo.extract_step(self)
--- a/easybuild/easyblocks/generic/cargopythonpackage.py
+++ b/easybuild/easyblocks/generic/cargopythonpackage.py
@@ -49,8 +49,7 @@ class CargoPythonPackage(PythonPackage, Cargo):  # PythonPackage must come first
 
     def __init__(self, *args, **kwargs):
         """Constructor for CargoPythonPackage easyblock."""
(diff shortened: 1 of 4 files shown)


Flamefire:
Please have a look at this @boegel @Micket (involved in #2934)

Related easyconfig PR: https://github.com/easybuilders/easybuild-easyconfigs/pull/21143

easybuild/easyblocks/generic/cargopythonpackage.py
```diff
@@ -49,8 +49,7 @@ def extra_options(extra_vars=None):
 
     def __init__(self, *args, **kwargs):
         """Constructor for CargoPythonPackage easyblock."""
-        Cargo.__init__(self, *args, **kwargs)
-        PythonPackage.__init__(self, *args, **kwargs)
+        super().__init__(*args, **kwargs)
```
boegel, line 52: This looks very sensible, but I seem to remember there was a reason why we didn't do it like this initially...

@Micket Do you recall why we implemented it like this?

Flamefire:
As discussed in the confcall I made that compatible with Python 2 which doesn't support argument-less `super()`

I found another instance of that in Palm and an `__init__` which did nothing and could be removed, see https://pylint.readthedocs.io/en/latest/user_guide/messages/warning/useless-parent-delegation.html

boegel:
EB_ARGS="maturin-1.1.0-GCCcore-12.3.0.eb"

boegel (approved):

PR #3415, opened 2024-08-19
Replace `run_cmd` with `run_shell_cmd` in custom easyblock for MetalWalls

Crivella:
Replaced `run_cmd` with `run_shell_cmd` for 5.0 release

### DIFF
--- a/easybuild/easyblocks/m/metalwalls.py
+++ b/easybuild/easyblocks/m/metalwalls.py
@@ -34,7 +34,7 @@ import easybuild.tools.toolchain as toolchain
 from easybuild.tools import LooseVersion
 from easybuild.tools.build_log import EasyBuildError
 from easybuild.tools.modules import get_software_root, get_software_version
-from easybuild.tools.run import run_cmd
+from easybuild.tools.run import run_shell_cmd
 
 from easybuild.easyblocks.generic.makecp import MakeCp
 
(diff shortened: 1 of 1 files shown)


boegel:
EB_ARGS="MetalWalls-21.06.1-foss-2023a.eb"
EB_BRANCH=5.0.x

Crivella:
I tested also on my system with `--rpath` and i get the same error, but `libplumed.so` and `libplumedKernel.so` are in the same directory.
Also I don't think we had this failure when building MW for EESSI, which should have RPATH enabled

Having a closer look at the build process, it seems that the plumed commands is used only to generate links to the `Plumed.inc` which than is used by the Makefile to link plumed as the full path to the library.
The secondary dependency of `libplumed.so` -> `libplumedKernel.so` is not being picked up as NEEDED in the elf and consequently by RPATH.
Is there a workaround that is implemented in the EESSI build for this?

boegel:
Hmm, from the top of my head I don't recall us implementing a workaround in EESSI for this, so I'm also a bit puzzled here...

The installation in EESSI doesn't show the problems being caught here then?

Crivella:
Yeah it went through without any problem, also ran the ReFrame test on my WS from the EESSI installation

Crivella:
As suggested by @boegel in a discussion it seems one of the dependencies (in this case PLUMED) was not build with RPATH enabled on the build cluster.
Tested on my WS and rebuildig PLUMED with `--rpath` leads to a successful build of MW without any easyblock modifications

boegel:
It looks like our set up `jsc-zen3` is not exactly ideal, as build done with EasyBuild 4.x and 5.x are mixed...

So indeed, `PLUMED` was installed with EasyBuild 4.x (without RPATH).

Should be easy to work around though, I'll just trigger a rebuild of PLUMED too...

boegel:
EB_ARGS="PLUMED-2.9.0-foss-2023a.eb MetalWalls-21.06.1-foss-2023a.eb"
EB_BRANCH=5.0.x

PR #3424, opened 2024-08-26
remove unused easyblock for BamTools

branfosj:
* [x] https://github.com/easybuilders/easybuild-easyconfigs/pull/21263

We can switch the easyconfigs to using the generic `CMakeMake` easyblock and then remove this easyblock.


boegel (approved):
lgtm (but https://github.com/easybuilders/easybuild-easyconfigs/pull/21263 should get merged first

PR #3425, opened 2024-08-28
use non-system toolchain for testing `--module-only` for ELPA and FFTW easyblocks, since they rely on toolchain options like 'pic' being set

boegel:
fix for failing test suite due to changes in:
* https://github.com/easybuilders/easybuild-framework/pull/4585


Micket (approved):
lgtm

PR #3433, opened 2024-09-04
This is the cp2k.py(Easyblock) that i have written being based on PR(2918)

beeebiii:
Hi, as discussed in Slack, and being based in in these three PRs:
https://github.com/easybuilders/easybuild-easyconfigs/pull/17718
https://github.com/easybuilders/easybuild-easyblocks/pull/2349
https://github.com/easybuilders/easybuild-easyblocks/pull/2918
I have written easyblock for CP2K, during the compilation of **CP2K-2024.2-foss-2023a-CUDA-12.2.0.eb **. Almost all the tests pass except for around 50 tests of total 4000 tests fail.


easybuild/easyblocks/c/cp2k.py
```diff
@@ -219,22 +230,39 @@ def configure_step(self):
...
         cuda = get_software_root('CUDA')
         if cuda:
-            options['DFLAGS'] += ' -D__ACC -D__DBCSR_ACC'
-            options['LIBS'] += ' -lcudart -lcublas -lcufft -lrt'
-            options['NVCC'] = ' nvcc'
-
+            if cp2k_version >= LooseVersion('2024'):
+                options['DFLAGS'] += ' -D__OFFLOAD_CUDA -D__DBCSR_ACC '
+                options['LIBS'] += ' -lcufft -lcudart -lnvrtc -lcuda -lcublas'
+                options['OFFLOAD_CC'] = 'nvcc'
```
gkaf89, line 254: The `OFFLOAD_CC` variable is used in the makefile of CP2K to build files with OpenACC offloading, even versions for older that 2024. For instance 2023.1 fails to build with CUDA support with the old easyblock.

Maybe enable the modification for some versions earlier than 2024 if they are still relevant?

gkaf89 (commented):
Modification relevant, fix CUDA builds. Tested with CP2K 2024.3 with and without CUDA (12.1.1 with gpuver=V100).

easybuild/easyblocks/c/cp2k.py
beeebiii: Yeah it can surely be modified, but i have not tested them with old version,s when i test them will definitely modify if required. Thanks!!

beeebiii:
Tested it on two of the clusters with (gpuver=A100 and gpuver=H100) worked without any issues.

pavelToman:
@boegelbot please test @ jsc-zen3 EB_ARGS="CP2K-2023.1-foss-2023a.eb"

pavelToman:
This PR is replaced by: 
- https://github.com/easybuilders/easybuild-easyblocks/pull/3888

PR #3435, opened 2024-09-05
make sure user packages are not used in sanity check of PythonBundle

Flamefire:
Some PythonBundle ECs (e.g. IPython) use custom sanity checks using python. That might pick up python packages installed to the users `$HOME` which can yield wrong results.
We already take care of this for the extensions sanity check in PythonPackage and for installing extensions in PythonBundle.

However for the sanity check of PythonBundle itself the variable is not defined (even if it was defined during the build) as the environment gets reset when loading the module used for the sanity check.

Hence use the same approach used in PythonPackage and set `$PYTHONNOUSERSITE` after loading the module.

### DIFF
--- a/easybuild/easyblocks/generic/pythonbundle.py
+++ b/easybuild/easyblocks/generic/pythonbundle.py
@@ -155,6 +155,18 @@ class PythonBundle(Bundle):
 
         return txt
 
+    def load_module(self, *args, **kwargs):
+        """
+        Make sure that $PYTHONNOUSERSITE is defined after loading module file for this software."""
+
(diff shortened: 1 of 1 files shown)


Flamefire:
Test report by @Flamefire
 * **SUCCESS** _IPython-7.18.1-GCCcore-10.2.0.eb_ 

n1552 - Linux RHEL 8.9 (Ootpa), x86_64, Intel(R) Xeon(R) Platinum 8470 (icelake), Python 3.8.17
See https://gist.github.com/Flamefire/289e5201213f6908e3670003a1c3b0a2 for a full test report.

boegel (approved):
lgtm

Issue #3429, opened 2024-09-03
clangversion string style fails sanity check in latest AOCC


matthewabrown:
https://github.com/easybuilders/easybuild-easyblocks/blob/cc5b95dab1aabae2a64360c492a4f068f8eab267/easybuild/easyblocks/a/aocc.py#L150

I'm creating a new easyconfig AOCC-4.2.0-GCCcore-13.2.0.eb. I use the latest 4.0.0 easyconfig but 
 - provided a newer GCCcore: 13.2.0 
 - the latest version of AOCC: 4.2.0
 - updated clangversion string in the easyconfig to match the version provided by GCCcore

The user supplied dotted decimal (16.0.3) is reused during sanity check but the AOCC installer uses major version (16) only.

```
# EasyConfigs have a clear pattern of using dotted decimals
$ grep clang *
AOCC-3.1.0-GCCcore-10.2.0.eb:clangversion = '12.0.0'
AOCC-3.1.0-GCCcore-10.3.0.eb:clangversion = '12.0.0'
AOCC-3.1.0-GCCcore-11.2.0.eb:clangversion = '12.0.0'
AOCC-3.2.0-GCCcore-11.2.0.eb:clangversion = '13.0.0'
AOCC-3.2.0-GCCcore-11.3.0.eb:clangversion = '13.0.0'
AOCC-4.0.0-GCCcore-11.3.0.eb:clangversion = '14.0.6'
AOCC-4.0.0-GCCcore-12.2.0.eb:clangversion = '14.0.6'
AOCC-4.0.0-GCCcore-12.3.0.eb:clangversion = '14.0.6'
AOCC-4.2.0-GCCcore-13.2.0.eb:clangversion = '16.0.3'
```

Installation failed during sanity checking:
```
== postprocessing...
== sanity checking...
== ... (took 1 secs)
== FAILED: Installation ended unsuccessfully (build directory: /dev/shm/brownm12/build/AOCC/4.2.0/GCCcore-13.2.0): build failed (first 300 chars): Sanity check failed: no file found at 'lib/clang/16.0.3/include/omp.h' in /apps/easybuild/software/owl-genoa/AOCC/4.2.0-GCCcore-13.2.0
no file found at 'lib/clang/16.0.3/include/stddef.h' in /apps/easybuild/software/owl-genoa/AOCC/4.2.0-GCCcore-13.2.0
no (non-empty) directory found at 'lib/clang/16. (took 52 secs)
```

[...]

PR #3437, opened 2024-09-06
Update AOCC easyblock to work with LLVM 16 and above

maxim-masterov:
See Issue #3429

### DIFF
--- a/easybuild/easyblocks/a/aocc.py
+++ b/easybuild/easyblocks/a/aocc.py
@@ -1,5 +1,5 @@
 ##
-# Copyright 2020-2023 Forschungszentrum Juelich GmbH
+# Copyright 2020-2024 Forschungszentrum Juelich GmbH
 #
 # This file is triple-licensed under GPLv2 (see below), MIT, and
 # BSD three-clause licenses.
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/a/aocc.py
```diff
@@ -67,6 +67,10 @@ def __init__(self, *args, **kwargs):
 
         self.clangversion = self.cfg['clangversion']
 
+        # Starting from LLVM v16, the folder naming is based on the major version of clang.
+        if LooseVersion(self.clangversion) >= LooseVersion('16.0.0'):
+            self.clangversion = '16'
+
```
Thyre, line 73: This will break once AOCC updates to a newer Clang version. In addition, updating the `clangversion` to the major version only feels not correct. This is only needed for the sanity check. Therefore, such conversion should only be done where necessary. 

In the Clang EasyBlock, this is done. I've posted a draft for AOCC in the issue for this EasyBlock, but didn't have the time to create a PR and test it.

easybuild/easyblocks/a/aocc.py
```diff
@@ -67,6 +67,10 @@ def __init__(self, *args, **kwargs):
...
     def _aocc_guess_clang_version(self):
```
Thyre, line 74: We should update the list with known AOCC versions, since this simplifies writing the respective EasyConfigs

Thyre (changes requested):

boegel:
Is this superseded by #3458?

maxim-masterov:
@boegel I would say - yes. The solution from @Thyre is more elegant.

maxim-masterov:
Closing in favor of #3458

PR #3440, opened 2024-09-11
ignore user python packages by using `-s` option when running Python code via `python -c`

Flamefire:
To avoid the user environment (Python packages installed to $HOME/lib/python*) pass `-s` to all invocations of `python -c`.


boegel:
@Flamefire Also here, we should add a check to the test suite to avoid that use of `python -c` without `-s` is introduced again going forward...

Flamefire:
@boegel Can we get this in for the release?

PR #3452, opened 2024-09-20
Make `CMakeMake` respect the toolchainopts when selecting build type

Micket:
See title

PR #3453, opened 2024-09-20
 Make `MesonNinja` respect the toolchainopts when selecting build type #3452 

### LINKED ISSUE #3280: set a default buildtype in MesonNinja easyblock
for e.g. CMake we add `-DCMAKE_BUILD_TYPE=Release` by default. 
We don't currently do anything like that for meson: https://github.com/easybuilders/easybuild-easyblocks/blob/develop/easybuild/easyblocks/generic/mesonninja.py#L95

this means we actually default to `debug` without optimisations ([ref](https://mesonbuild.com/Builtin-options.html#core-options)):
|Option|Default value|Description|Is per machine|Is per subproject|
|---|---|---|---|---|
|buildtype {plain, debug, debugoptimized, release, minsize, custom}|debug|Build type to use|no|no|

these options correspond to ([ref](https://mesonbuild.com/Builtin-options.html#details-for-buildtype)):
|buildtype|debug|optimization|
|---|---|---|
|plain|false|plain|
|debug|true|0
|debugoptimized|true|2|
|release|false|3|
|minsize|true|s|

Micket:
Fixes #3280

Also have another variant #3454 (which was suppoesd to be an alternative approach, but i guess it ended up being more of a superset of this PR int he end)

### DIFF
--- a/easybuild/easyblocks/generic/mesonninja.py
+++ b/easybuild/easyblocks/generic/mesonninja.py
@@ -53,11 +53,27 @@ class MesonNinja(EasyBlock):
         extra_vars.update({
             'build_dir': [None, "build_dir to pass to meson", CUSTOM],
             'build_cmd': [DEFAULT_BUILD_CMD, "Build command to use", CUSTOM],
+            'build_type': [None, "Build type for meson, e.g. release."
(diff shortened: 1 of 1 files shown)


boegel:
@Micket It seems like this should be closed in favor of #3454?

PR #3454, opened 2024-09-20
Make `MesonNinja` respect the toolchainopts with buildtype as well as `--debug` and `--optimization` flags

### LINKED ISSUE #3280: set a default buildtype in MesonNinja easyblock

Micket:
Fixes #3280 

Should also merge https://github.com/easybuilders/easybuild-easyconfigs/pull/21619 once this is merged.

In this version, i double up on both buildtype as well as --optimizationand --debug flags to respect the toolchainopts. I think this will be the better approach, but i haven't tested anything yet.

### DIFF
--- a/easybuild/easyblocks/generic/mesonninja.py
+++ b/easybuild/easyblocks/generic/mesonninja.py
@@ -128,7 +128,7 @@ class MesonNinja(EasyBlock):
             'preconfigopts': self.cfg['preconfigopts'],
             'source_dir': build_dir,
             'buildtype': self.build_type(),
-            'optimization': self.optimizatoin(),
+            'optimization': self.optimization(),
             'debug': self.toolchain.options.get('debug', False),
         }
         res = run_shell_cmd(cmd)
--- a/easybuild/easyblocks/generic/mesonninja.py
+++ b/easybuild/easyblocks/generic/mesonninja.py
@@ -127,8 +127,8 @@ class MesonNinja(EasyBlock):
             'installdir': self.installdir,
             'preconfigopts': self.cfg['preconfigopts'],
             'source_dir': build_dir,
-            'buildtype': self.build_type(),
-            'optimization': self.optimization(),
+            'buildtype': self.build_type,
+            'optimization': self.optimization,
             'debug': self.toolchain.options.get('debug', False),
         }
         res = run_shell_cmd(cmd)
--- a/easybuild/easyblocks/generic/mesonninja.py
+++ b/easybuild/easyblocks/generic/mesonninja.py
@@ -56,6 +56,8 @@ class MesonNinja(EasyBlock):
             'build_type': [None, "Build type for meson, e.g. release."
                                  "Defaults to 'release', 'debugoptimized' or 'debug' depending on "
                                  "toolchainopts[debug,noopt]", CUSTOM],
+            'ndebug': [True, "Sets -Db_ndebug which in turn defines NDEBUG for C/C++ builds."
+                             "This disabled costly asserts in code, typical for production.", CUSTOM],
             'configure_cmd': [DEFAULT_CONFIGURE_CMD, "Configure command to use", CUSTOM],
             'install_cmd': [DEFAULT_INSTALL_CMD, "Install command to use", CUSTOM],
             'separate_build_dir': [True, "Perform build in a separate directory", CUSTOM],
@@ -69,9 +71,9 @@ class MesonNinja(EasyBlock):
         # While we do set optimization and debug flag separately, build scripts may be adding additional
         # defines and flags based on the build_type as well so we pick the closest match.
         if build_type is None:
-            if self.toolchain.options.get('noopt', None):  # also implies debug but is the closest match
+            if self.toolchain.options.get('noopt', False):  # also implies debug but is the closest match
                 build_type = 'debug'
-            elif self.toolchain.options.get('debug', None):
+            elif self.toolchain.options.get('debug', False):
                 build_type = 'debugoptimized'
             else:
                 build_type = 'release'
@@ -121,15 +123,16 @@ class MesonNinja(EasyBlock):
         build_dir = self.cfg.get('build_dir') or self.start_dir
 
         cmd = ("%(preconfigopts)s %(configure_cmd)s --prefix %(installdir)s --buildtype %(buildtype)s %(configopts)s "
-               "--optimization %(optimization)s %(debug)s %(source_dir)s") % {
+               "--optimization %(optimization)s %(debug)s -Db_ndebug=%(ndebug)s %(source_dir)s") % {
             'configopts': self.cfg['configopts'],
             'configure_cmd': configure_cmd,
             'installdir': self.installdir,
             'preconfigopts': self.cfg['preconfigopts'],
-            'source_dir': build_dir,
             'buildtype': self.build_type,
             'optimization': self.optimization,
             'debug': '--debug' if self.toolchain.options.get('debug', False) else '',
+            'ndebug': str(self.cfg.get('ndebug')).lower(),
+            'source_dir': build_dir,
         }
         res = run_shell_cmd(cmd)
         return res.output
@@ -144,7 +147,7 @@ class MesonNinja(EasyBlock):
         if self.cfg['parallel']:
             parallel = "-j %s" % self.cfg['parallel']
 
-        cmd = "%(prebuildopts)s %(build_cmd)s %(parallel)s %(buildopts)s" % {
+        cmd = "%(prebuildopts)s %(build_cmd)s -v %(parallel)s %(buildopts)s" % {
             'buildopts': self.cfg['buildopts'],
             'build_cmd': build_cmd,
             'parallel': parallel,
--- a/easybuild/easyblocks/generic/mesonninja.py
+++ b/easybuild/easyblocks/generic/mesonninja.py
@@ -53,11 +53,39 @@ class MesonNinja(EasyBlock):
         extra_vars.update({
             'build_dir': [None, "build_dir to pass to meson", CUSTOM],
             'build_cmd': [DEFAULT_BUILD_CMD, "Build command to use", CUSTOM],
+            'build_type': [None, "Build type for meson, e.g. release."
+                                 "Defaults to 'release', 'debugoptimized' or 'debug' depending on "
+                                 "toolchainopts[debug,noopt]", CUSTOM],
             'configure_cmd': [DEFAULT_CONFIGURE_CMD, "Configure command to use", CUSTOM],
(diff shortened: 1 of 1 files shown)


boegel (dismissed):

Micket:
> WARNING: Recommend using either -Dbuildtype or -Doptimization + -Ddebug. Using both is redundant since they override each other. See: https://mesonbuild.com/Builtin-options.html#build-type-options

ok, so apparently meson doesn't want you to do this. I find this strange, as just using the optimization + debug options wouldn't cover things like -DNODEBUG defines. Maybe noone using meson uses such defines in their C/C++ codes? :/

Micket:
https://github.com/mesonbuild/meson-python/pull/325 No this seems to indicate otherwise. We certainly always want the NDEBUG set for our builds...

Micket:
Alright, so to summarize

1. I add the -v flag so we can see what we are compiling
2. I set the build type, in case they have some "if-release" logic that adds extra flags and such.
3. I set optimization and debug on top of that, which produces a warning that I disagree with due to the point mentioned above.
4. I set the `-bn_debug=true` flag for all builds that aren't noopt, so that we get -DNDEBUG when building debugoptimized as well as release. I think by default, the meson-python actually adds the flag `-bn_debug=if-release` but if we are enabling debug symbols, you don't want to squash this. They mention in the issue tracking n_debug flag that they don't add it automatically to release because they expect packagers to set all their flags themselves anyway.. and i guess they are right (now).

All in all this brings meson builds in line with the rest of the stuff we build, and I don't think any of it is controversial. 
Wouldn't surprise me if we see a large speedup in some packages after these changes, but it's hard to test things like "cairo" or "Wayland" for performance.

Micket:
Test report by @Micket
 * **SUCCESS** _cairo-1.18.0-GCCcore-13.3.0.eb_ 
 * **SUCCESS** _GLib-2.78.1-GCCcore-13.2.0.eb_ 
 * **SUCCESS** _libglvnd-1.7.0-GCCcore-13.3.0.eb_ 
 * **SUCCESS** _Wayland-1.23.0-GCCcore-13.3.0.eb_ 

vera-skylake-build - Linux Rocky Linux 8.9, x86_64, Intel Xeon Processor (Skylake, IBRS, no TSX), Python 3.6.8
See https://gist.github.com/Micket/1c22578d1e11a2d81ddd7f38a3c0030e for a full test report.

Micket:
Test report by @Micket
 * **SUCCESS** _cairo-1.18.0-GCCcore-13.3.0.eb_ 
 * **SUCCESS** _GLib-2.78.1-GCCcore-13.2.0.eb_ 
 * **SUCCESS** _libglvnd-1.7.0-GCCcore-13.3.0.eb_ 
 * **FAIL (build issue)** _Wayland-1.23.0-GCCcore-13.3.0.eb_ (partial log available at https://gist.github.com/Micket/b25cd2d50bfff9f3de75fde00bbb16a9)

Build succeeded for 3 out of 4 (4 easyconfigs in total)
vera-skylake-build - Linux Rocky Linux 8.9, x86_64, Intel Xeon Processor (Skylake, IBRS, no TSX), Python 3.6.8
See https://gist.github.com/Micket/d182683aa7be4718a6ad9167888a51d9 for a full test report.

Micket:
So, Wayland apparently needs to be built with asserts, else it fails the test suite. Which means, one can't actually test  a true install... Not sure how much it impacts performance, or if we care about tools like wayland for that sort of performance in HPC settings, so I'm making n_debug into an option.

Micket:
OK, so meson is really strange when specifying both buildtype and optmization/debug flags. 
They claim that they mutually exclusive, but I disagree; the value of "buildtype" itself has value given things like
```
if get_option('buildtype') == 'release':
   # custom logic here
```

In particular `-Db_ndebug=if-release`  is something a lot of things set (all of meson-python projects do this by default), so, it's not some little thing either. Though, if-release also leaves a lot to be desired; what about debugoptimized? is that supposed to be a release-like thing? Again `debug` is being used vaguely here.

I checked the meson.build from all the stuff that uses MesonNinja to see what, if any, such custom logic anyone might be doing, and apart from some harmless flags and b_ndebug=if-release, I couldn't find anything that worried me, so I'm going to go with the option of making buildtype and optimization/debug mutually exclusive for our easyblock as well, and just rely on optimization, debug, and b_ndebug flags.

Micket:
Test report by @Micket
 * **SUCCESS** _ATK-2.38.0-GCCcore-13.2.0.eb_ 
 * **SUCCESS** _at-spi2-atk-2.38.0-GCCcore-13.2.0.eb_ 
 * **SUCCESS** _at-spi2-core-2.50.0-GCCcore-13.2.0.eb_ 
 * **SUCCESS** _cairo-1.18.0-GCCcore-13.3.0.eb_ 
 * **FAIL (build issue)** _cairomm-1.16.2-GCC-12.3.0.eb_ (partial log available at https://gist.github.com/Micket/21e0eee0053815efb7ebedf48fbbfef8)
 * **SUCCESS** _desktop-file-utils-0.27-GCCcore-12.3.0.eb_ 
 * **SUCCESS** _DFT-D4-3.6.0-intel-2022a.eb_ 
 * **SUCCESS** _evince-45.0-GCC-12.3.0.eb_ 
 * **SUCCESS** _freebayes-1.3.7-gfbf-2023a-R-4.3.2.eb_ 
 * **SUCCESS** _FUSE-3.16.2-GCCcore-12.3.0.eb_ 
 * **SUCCESS** _Gdk-Pixbuf-2.42.10-GCCcore-13.2.0.eb_ 
 * **SUCCESS** _GLib-2.80.4-GCCcore-13.3.0.eb_ 
 * **SUCCESS** _GLibmm-2.78.1-GCCcore-13.2.0.eb_ 
 * **SUCCESS** _glib-networking-2.72.1-GCCcore-11.3.0.eb_ 
 * **SUCCESS** _GObject-Introspection-1.78.1-GCCcore-13.2.0.eb_ 
 * **SUCCESS** _Graphene-1.10.8-GCCcore-13.2.0.eb_ 
 * **SUCCESS** _GST-plugins-bad-1.22.5-GCC-12.3.0.eb_ 
 * **SUCCESS** _GST-plugins-base-1.22.5-GCC-12.3.0.eb_ 
 * **SUCCESS** _GStreamer-1.22.5-GCC-12.3.0.eb_ 
 * **SUCCESS** _GTK3-3.24.39-GCCcore-13.2.0.eb_ 
 * **SUCCESS** _GTK4-4.13.1-GCC-12.3.0.eb_ 
 * **SUCCESS** _inih-57-GCCcore-12.3.0.eb_ 
 * **SUCCESS** _libdrm-2.4.122-GCCcore-13.3.0.eb_ 
 * **SUCCESS** _libepoxy-1.5.10-GCCcore-13.2.0.eb_ 
 * **SUCCESS** _libGLU-9.0.3-GCCcore-13.2.0.eb_ 
 * **SUCCESS** _libglvnd-1.7.0-GCCcore-13.3.0.eb_ 
 * **SUCCESS** _libgxps-0.3.2-GCCcore-12.3.0.eb_ 
 * **SUCCESS** _libhandy-1.8.2-GCCcore-12.3.0.eb_ 
 * **SUCCESS** _libpciaccess-0.18.1-GCCcore-13.3.0.eb_ 
 * **SUCCESS** _LibSoup-3.0.8-GCC-11.3.0.eb_ 
 * **SUCCESS** _mm-common-1.0.6-GCCcore-13.2.0.eb_ 
 * **SUCCESS** _MODFLOW-6.4.4-foss-2023a.eb_ 
 * **SUCCESS** _Pango-1.51.0-GCCcore-13.2.0.eb_ 
 * **SUCCESS** _pixman-0.43.4-GCCcore-13.3.0.eb_ 
 * **SUCCESS** _PretextMap-0.1.9-GCCcore-12.3.0.eb_ 
 * **SUCCESS** _PyCairo-1.25.1-GCCcore-13.2.0.eb_ 
 * **SUCCESS** _Wayland-1.23.0-GCCcore-13.3.0.eb_ 
 * **SUCCESS** _wpebackend-fdo-1.14.1-GCCcore-11.3.0.eb_ 
 * **SUCCESS** _X11-20240607-GCCcore-13.3.0.eb_ 
 * **SUCCESS** _xtb-6.7.0-gfbf-2023a.eb_ 
 * **SUCCESS** _Xvfb-21.1.9-GCCcore-13.2.0.eb_ 

Build succeeded for 40 out of 41 (41 easyconfigs in total)
vera-skylake-build - Linux Rocky Linux 8.9, x86_64, Intel Xeon Processor (Skylake, IBRS, no TSX), Python 3.6.8
See https://gist.github.com/Micket/49102596d81d8d925377b46e82b7475f for a full test report.

Micket:
The error on cairomm is something unrelated, 2 tests crashes for me no matter what easyblock i use, even in 4.9.2, so it shouldn't let this block the merge.

jfgrimm (approved):
lgtm

easybuild/easyblocks/generic/mesonninja.py
```diff
@@ -53,12 +53,26 @@ def extra_options(extra_vars=None):
...
+    @property
+    def optimization(self):
+        """Optimization level"""
+        if self.toolchain.options.get('noopt', False):
+            return 0
+        elif self.toolchain.options.get('lowopt', False):
+            return 1
+        else:
+            return 2
+
```
jfgrimm, line 75: wait, what if we have `'opt': True` (as in, use `-O3`)?
surely this needs another elif?

jfgrimm (changes requested):

easybuild/easyblocks/generic/mesonninja.py
jfgrimm:
```suggestion
    @property
    def optimization(self):
        """Optimization level"""
        if self.toolchain.options.get('noopt', False):
            return 0
        elif self.toolchain.options.get('lowopt', False):
            return 1
        elif self.toolchain.options.get('opt', False):
            return 3
        else:
            return 2

```

jfgrimm (approved):

PR #3460, opened 2024-09-24
Avoid failure in Python package installation and sanity check when `$PIP_REQUIRE_VIRTUALENV` is set

Flamefire:
We already have it for the installation if set by users but if a module sets it (e.g Python) the sanity check will fail

### DIFF
--- a/easybuild/easyblocks/generic/pythonbundle.py
+++ b/easybuild/easyblocks/generic/pythonbundle.py
@@ -159,6 +159,8 @@ class PythonBundle(Bundle):
         # Required here to ensure that it is defined for sanity check commands of the bundle
         # because the environment is reset to the initial environment right before loading the module
         env.setvar('PYTHONNOUSERSITE', '1', verbose=False)
+        # Set (again) in case the module changes it
+        env.setvar('PIP_REQUIRE_VIRTUALENV', 'false', verbose=False)
 
     def sanity_check_step(self, *args, **kwargs):
         """Custom sanity check for bundle of Python package."""
--- a/easybuild/easyblocks/generic/pythonbundle.py
+++ b/easybuild/easyblocks/generic/pythonbundle.py
@@ -30,12 +30,11 @@ EasyBuild support for installing a bundle of Python packages, implemented as a g
 import os
 
 from easybuild.easyblocks.generic.bundle import Bundle
-from easybuild.easyblocks.generic.pythonpackage import EXTS_FILTER_PYTHON_PACKAGES, run_pip_check
+from easybuild.easyblocks.generic.pythonpackage import EXTS_FILTER_PYTHON_PACKAGES, run_pip_check, set_py_env_vars
 from easybuild.easyblocks.generic.pythonpackage import PythonPackage, get_pylibdirs, find_python_cmd_from_ec
 from easybuild.tools.build_log import EasyBuildError
 from easybuild.tools.config import build_option, PYTHONPATH, EBPYTHONPREFIXES
 from easybuild.tools.modules import get_software_root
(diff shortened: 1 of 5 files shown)


easybuild/easyblocks/generic/pythonbundle.py
```diff
@@ -181,6 +181,8 @@ def load_module(self, *args, **kwargs):
         # Required here to ensure that it is defined for sanity check commands of the bundle
         # because the environment is reset to the initial environment right before loading the module
         env.setvar('PYTHONNOUSERSITE', '1', verbose=False)
+        # Set (again) in case the module changes it
+        env.setvar('PIP_REQUIRE_VIRTUALENV', 'false', verbose=False)
```
boegel, line 185: @Flamefire Shouldn't we introduce a function that can be called from both `pythonbundle` and `pythonpackage`?

boegel (changes requested):

easybuild/easyblocks/generic/pythonbundle.py
Flamefire: You mean one that sets `PYTHONNOUSERSITE`, `PIP_REQUIRE_VIRTUALENV` and `PIP_DISABLE_PIP_VERSION_CHECK`?

Given that we already need it in quite some places it makes sense, yes. Maybe named `setup_pip_env` with a `verbose` param? I'd make that default to `False` and only set it to `True` in `PythonPackage.__init__` to reduce log spam. Even there it gets likely duplicated a lot for bundles but completely omitting it feels wrong.

boegel: Sounds good

Flamefire: Ok, implemented it in the Python EasyBlock file to avoid cyclic includes of that and PythonPackage. Those variables are also required/useful for TensorFlow & friends, hence I made them a constant.

To avoid the log spam the actual set is only executed if any of the variables need changing.

boegel:
EB_ARGS="tensorflow-compression-2.11.0-foss-2022a-CUDA-11.7.0.eb TensorFlow-2.15.1-foss-2023a.eb --installpath /tmp/$USER/pr3460"

Flamefire:
@boegel ping on that. Just ran into this failure while testing EB 5 ECs.

Flamefire:
Rebased for 5.0x

Flamefire:
Can this be included in 5.0.1? I had to patch our 5.0.0 installation again after the update. And test reports looked fine too so I don't expect any downsides.

boegel:
> Can this be included in 5.0.1? I had to patch our 5.0.0 installation again after the update. And test reports looked fine too so I don't expect any downsides.

I'll see what I can do to get this merged ASAP

boegel:
Test report by @boegel
 * **SUCCESS** _Python-2.7.18-GCCcore-12.3.0.eb_ 
 * **SUCCESS** _Python-3.12.3-GCCcore-13.3.0.eb_ 
 * **SUCCESS** _matplotlib-3.9.2-gfbf-2024a.eb_ 
 * **SUCCESS** _numexpr-2.9.0-foss-2023a.eb_ 
 * **SUCCESS** _TensorFlow-2.13.0-foss-2023a.eb_ 

node3502.doduo.os - Linux RHEL 9.4, x86_64, AMD EPYC 7552 48-Core Processor (zen2), Python 3.9.18
See https://gist.github.com/boegel/63fade637b46537fb613f90197c33773 for a full test report.

Flamefire:
Rebased without changes to resolve merge conflict

boegel (approved):
lgtm

PR #3283, opened 2024-04-05
Ignore Python from virtualenvs in GROMACS configure via `-DPython3_FIND_VIRTUALENV=STANDARD`

Flamefire:
(created using `eb --new-pr`)

As reported in [Slack](https://easybuild.slack.com/archives/C34UA1HT7/p1712147065575909) GROMACS may pick up Python from an active virtualenv (e.g. when EasyBuild is installed in a virtualenv) instead of from the Python module

This is caused by GROMACS explicitly preferring that:
```
if(NOT Python3_FIND_VIRTUALENV)
    # We advocate using Python venvs to manage package availability, so by default
    # we want to preferentially discover user-space software.
    set(Python3_FIND_VIRTUALENV FIRST)
endif()
```

This happens when `find_package(Python3)` is used which is the case for version 2021 and up. The current `self.cfg.update('configopts', "-DPYTHON_EXECUTABLE=%s" % bin_python)` only applies to usages of `find_package(PythonInterp ...)` which are still present (e.g. in the gtest submodule) but not used exclusively as they are deprecated by CMake 3.12+ in favor of `FindPython/2/3`

This change avoids it picking up the virtualenv by setting the CMake variable which avoids GROMACS changing it.

Especially together with #3282 the intended Python should be picked up.

I also did a quick Search&Replace to avoid constructing `LooseVersion(self.version)` many times in this function.

PR #3282, opened 2024-04-05
Set CMake hints for path to Python installation if it's direct or indirect dependency (when using CMake >= 3.12)

Flamefire:
(created using `eb --new-pr`)

PR #3463, opened 2024-09-26
let `CMakeMake` easyblock also set `Python_EXECUTABLE` option, as well as `Python3_EXECUTABLE` and `Python2_EXECUTABLE` derivatives (when appropriate)

Crivella:
## EDIT
Following the thread discussion this has changed from setting `Python3_FIND_VIRTUALENV` to setting `Python_EXECUTABLE` and all its variants in order to force the CMakeMake easyblock to pick up the correct python when loaded as a module.

This would replace:

- #3282

## OLD

This is related to:

- https://github.com/easybuilders/easybuild-easyblocks/pull/3283
- https://github.com/easybuilders/easybuild-easyblocks/pull/3373
- Possibly any CMake build that is invoking `FindPython3` via `find_package(Python3)`

`Python3_FIND_VIRTUALENV` controls how CMake finds the python3 interpreter in case a `virtualenv` or `conda` environment is activated, by default prioritizing the virtual environment.
Setting it to `STANDARD` switch the behavior to looking for python3 in `PATH` which allows the correct python to be found when invoking `eb` from within a virtual environment

No version check has been added as the flag is supported since CMake 3.15 which now is tied to an archived toolchain (GCC 8.3.0 https://docs.easybuild.io/policies/toolchains/)

I think this should always be on by default.
Even if a build process tries to work with virtual environments (eg creating a venv, activating it and tha running cmake trying to use the python from the venv) i think that activating the virtualenv inside the build process would prepend to the PATH and still allow finding the correct python

### DIFF
--- a/easybuild/easyblocks/generic/cmakemake.py
+++ b/easybuild/easyblocks/generic/cmakemake.py
@@ -105,8 +105,8 @@ def setup_cmake_env_python_hints(cmake_version):
             setvar('Python2_ROOT_DIR', python_root)
 
 
-def _get_cmake_python_config():
-    """Get the CMake configuration options for Python hints."""
+def get_cmake_python_config_dict():
+    """Get a dictionary with the CMake configuration options for Python hints."""
     options = {}
     python_root = get_software_root('Python')
     if python_root:
@@ -123,14 +123,9 @@ def _get_cmake_python_config():
     return options
 
 
-def get_cmake_python_config_dict():
-    """Get a dictionary with the CMake configuration options for Python hints."""
-    return _get_cmake_python_config()
-
-
 def get_cmake_python_config_str():
     """Get a string with the CMake configuration options for Python hints."""
-    options = _get_cmake_python_config()
+    options = get_cmake_python_config_dict()
     return ' '.join(['-D%s=%s' % (key, value) for key, value in options.items()])
 
 
@@ -327,7 +322,7 @@ class CMakeMake(ConfigureMake):
         options['CMAKE_FIND_USE_PACKAGE_REGISTRY'] = 'OFF'
 
         # ensure CMake uses EB python, not system or virtualenv python
-        options.update(self._get_cmake_python_config())
+        options.update(get_cmake_python_config_dict())
 
         if not self.cfg.get('allow_system_boost', False):
             boost_root = get_software_root('Boost')
--- a/easybuild/easyblocks/generic/cmakemake.py
+++ b/easybuild/easyblocks/generic/cmakemake.py
@@ -86,6 +86,7 @@ def setup_cmake_env(tc):
     setvar("CMAKE_INCLUDE_PATH", include_paths)
     setvar("CMAKE_LIBRARY_PATH", library_paths)
 
+
 def setup_cmake_env_python_hints(cmake_version):
     """Convenience function to set CMake hints for FindPython[_2/3] as environment variables.
     Needed to avoid wrong Python being picked up by CMake when not called directly by EasyBuild but as step in a
@@ -102,6 +103,7 @@ def setup_cmake_env_python_hints(cmake_version):
             setvar('Python3_ROOT_DIR', python_root)
         else:
             setvar('Python2_ROOT_DIR', python_root)
+
 
 def _get_cmake_python_config():
     """Get the CMake configuration options for Python hints."""
@@ -120,9 +122,11 @@ def _get_cmake_python_config():
             options['Python2_EXECUTABLE'] = python_exe
     return options
 
+
 def get_cmake_python_config_dict():
     """Get a dictionary with the CMake configuration options for Python hints."""
     return _get_cmake_python_config()
+
 
 def get_cmake_python_config_str():
     """Get a string with the CMake configuration options for Python hints."""
--- a/easybuild/easyblocks/generic/cmakemake.py
+++ b/easybuild/easyblocks/generic/cmakemake.py
@@ -284,6 +284,10 @@ class CMakeMake(ConfigureMake):
         # disable CMake user package repository
         options['CMAKE_FIND_USE_PACKAGE_REGISTRY'] = 'OFF'
 
+        # ensure CMake uses PATH to determine python without prioritizing a virtualenv
+        # necessary to pick up on the correct python version when `eb` is run from a virtualenv
+        options['Python3_FIND_VIRTUALENV'] = 'STANDARD'
+
         if not self.cfg.get('allow_system_boost', False):
             boost_root = get_software_root('Boost')
             if boost_root:
--- a/easybuild/easyblocks/generic/cmakemake.py
+++ b/easybuild/easyblocks/generic/cmakemake.py
@@ -92,7 +92,8 @@ def setup_cmake_env_python_hints(cmake_version):
     Needed to avoid wrong Python being picked up by CMake when not called directly by EasyBuild but as step in a
     build and no option is provided to set custom CMake variables.
     """
-    cmake_version = det_cmake_version()
+    if cmake_version is None:
+        cmake_version = det_cmake_version()
     if LooseVersion(cmake_version) < '3.12':
         raise EasyBuildError("Setting Python hints for CMake requires CMake version 3.12 or newer")
     python_root = get_software_root('Python')
--- a/easybuild/easyblocks/generic/cmakemake.py
+++ b/easybuild/easyblocks/generic/cmakemake.py
@@ -88,9 +88,10 @@ def setup_cmake_env(tc):
 
 
 def setup_cmake_env_python_hints(cmake_version=None):
-    """Convenience function to set CMake hints for FindPython[_2/3] as environment variables.
-    Needed to avoid wrong Python being picked up by CMake when not called directly by EasyBuild but as step in a
-    build and no option is provided to set custom CMake variables.
+    """Set environment variables as hints for CMake to prefer the Python module, if loaded.
+    Useful when there is no way to specify arguments for CMake directly,
+    e.g. when CMake is called from within another build system.
+    Otherwise get_cmake_python_config_[str/dict] should be used instead.
     """
     if cmake_version is None:
         cmake_version = det_cmake_version()
@@ -107,7 +108,7 @@ def setup_cmake_env_python_hints(cmake_version=None):
 
 
 def get_cmake_python_config_dict():
-    """Get a dictionary with the CMake configuration options for Python hints."""
+    """Get a dictionary with CMake configuration options for finding Python if loaded as a module."""
     options = {}
     python_root = get_software_root('Python')
     if python_root:
@@ -125,9 +126,11 @@ def get_cmake_python_config_dict():
 
 
 def get_cmake_python_config_str():
-    """Get a string with the CMake configuration options for Python hints."""
+    """Get CMake configuration arguments for finding Python if loaded as a module.
+    This string is intended to be passed to the invocation of `cmake`.
+    """
     options = get_cmake_python_config_dict()
-    return ' '.join(['-D%s=%s' % (key, value) for key, value in options.items()])
+    return ' '.join('-D%s=%s' % (key, value) for key, value in options.items())
 
 
 class CMakeMake(ConfigureMake):
--- a/easybuild/easyblocks/generic/cmakemake.py
+++ b/easybuild/easyblocks/generic/cmakemake.py
@@ -86,6 +86,49 @@ def setup_cmake_env(tc):
     setvar("CMAKE_INCLUDE_PATH", include_paths)
     setvar("CMAKE_LIBRARY_PATH", library_paths)
 
+def setup_cmake_env_python_hints(cmake_version):
+    """Convenience function to set CMake hints for FindPython[_2/3] as environment variables.
+    Needed to avoid wrong Python being picked up by CMake when not called directly by EasyBuild but as step in a
+    build and no option is provided to set custom CMake variables.
+    """
+    cmake_version = det_cmake_version()
+    if LooseVersion(cmake_version) < '3.12':
+        raise EasyBuildError("Setting Python hints for CMake requires CMake version 3.12 or newer")
+    python_root = get_software_root('Python')
+    if python_root:
+        python_version = LooseVersion(get_software_version('Python'))
+        setvar('Python_ROOT_DIR', python_root)
+        if python_version >= "3":
+            setvar('Python3_ROOT_DIR', python_root)
+        else:
+            setvar('Python2_ROOT_DIR', python_root)
+
+def _get_cmake_python_config():
+    """Get the CMake configuration options for Python hints."""
+    options = {}
+    python_root = get_software_root('Python')
+    if python_root:
+        python_version = LooseVersion(get_software_version('Python'))
+        python_exe = os.path.join(python_root, 'bin', 'python')
+        # This is required for (deprecated) `find_package(PythonInterp ...)`
+        options['PYTHON_EXECUTABLE'] = python_exe
+        # Ensure that both `find_package(Python) and find_package(Python2/3)` work as intended
+        options['Python_EXECUTABLE'] = python_exe
+        if python_version >= "3":
+            options['Python3_EXECUTABLE'] = python_exe
+        else:
+            options['Python2_EXECUTABLE'] = python_exe
+    return options
+
+def get_cmake_python_config_dict():
+    """Get a dictionary with the CMake configuration options for Python hints."""
+    return _get_cmake_python_config()
+
+def get_cmake_python_config_str():
+    """Get a string with the CMake configuration options for Python hints."""
+    options = _get_cmake_python_config()
+    return ' '.join(['-D%s=%s' % (key, value) for key, value in options.items()])
+
 
 class CMakeMake(ConfigureMake):
     """Support for configuring build with CMake instead of traditional configure script"""
@@ -273,11 +316,6 @@ class CMakeMake(ConfigureMake):
             # see https://github.com/Kitware/CMake/commit/3ec9226779776811240bde88a3f173c29aa935b5
             options['CMAKE_SKIP_RPATH'] = 'ON'
 
-        # make sure that newer CMAKE picks python based on location, not just the newest python
-        # Avoids issues like e.g. https://github.com/EESSI/software-layer/pull/370#issuecomment-1785594932
-        if LooseVersion(self.cmake_version) >= '3.15':
-            options['CMAKE_POLICY_DEFAULT_CMP0094'] = 'NEW'
-
         # show what CMake is doing by default
         options['CMAKE_VERBOSE_MAKEFILE'] = 'ON'
 
@@ -324,48 +362,6 @@ class CMakeMake(ConfigureMake):
         (out, _) = run_cmd(command, log_all=True, simple=False)
 
         return out
-
-    def _get_cmake_python_config(self):
-        """Get the CMake configuration options for Python hints."""
-        options = {}
-        python_root = get_software_root('Python')
-        if python_root:
-            python_version = LooseVersion(get_software_version('Python'))
-            python_exe = os.path.join(python_root, 'bin', 'python')
-            # This is required for (deprecated) `find_package(PythonInterp ...)`
-            options['PYTHON_EXECUTABLE'] = python_exe
-            # Ensure that both `find_package(Python) and find_package(Python2/3)` work as intended
-            options['Python_EXECUTABLE'] = python_exe
-            if python_version >= "3":
-                options['Python3_EXECUTABLE'] = python_exe
-            else:
-                options['Python2_EXECUTABLE'] = python_exe
-        return options
-
-    def get_cmake_python_config_dict(self):
-        """Get a dictionary with the CMake configuration options for Python hints."""
-        return self._get_cmake_python_config()
-
-    def get_cmake_python_config_str(self):
-        """Get a string with the CMake configuration options for Python hints."""
-        options = self._get_cmake_python_config()
-        return ' '.join(['-D%s=%s' % (key, value) for key, value in options.items()])
-
-    def set_cmake_python_env_hints(self):
-        """Convenience function to set CMake hints for FindPython[_2/3] as environment variables.
-        Needed to avoid wrong Python being picked up by CMake when not called directly by EasyBuild but as step in a
-        build and no option is provided to set custom CMake variables.
-        """
-        if LooseVersion(self.cmake_version) < '3.12':
-            raise EasyBuildError("Setting Python hints for CMake requires CMake version 3.12 or newer")
-        python_root = get_software_root('Python')
-        if python_root:
-            python_version = LooseVersion(get_software_version('Python'))
-            setvar('Python_ROOT_DIR', python_root)
-            if python_version >= "3":
-                setvar('Python3_ROOT_DIR', python_root)
-            else:
-                setvar('Python2_ROOT_DIR', python_root)
 
     def test_step(self):
         """CMake specific test setup"""
--- a/easybuild/easyblocks/generic/cmakemake.py
+++ b/easybuild/easyblocks/generic/cmakemake.py
@@ -285,18 +285,7 @@ class CMakeMake(ConfigureMake):
         options['CMAKE_FIND_USE_PACKAGE_REGISTRY'] = 'OFF'
 
         # ensure CMake uses EB python, not system or virtualenv python
-        python_root = get_software_root('Python')
-        if python_root:
-            python_version = LooseVersion(get_software_version('Python'))
-            python_exe = os.path.join(python_root, 'bin', 'python')
-            # This is required for (deprecated) `find_package(PythonInterp ...)`
-            options['PYTHON_EXECUTABLE'] = python_exe
-            # Ensure that both `find_package(Python) and find_package(Python2/3)` work as intended
-            options['Python_EXECUTABLE'] = python_exe
-            if python_version >= "3":
-                options['Python3_EXECUTABLE'] = python_exe
-            else:
-                options['Python2_EXECUTABLE'] = python_exe
+        options.update(self._get_cmake_python_config())
 
         if not self.cfg.get('allow_system_boost', False):
             boost_root = get_software_root('Boost')
@@ -336,7 +325,33 @@ class CMakeMake(ConfigureMake):
 
         return out
 
-    def set_cmake_env_vars_python(self):
+    def _get_cmake_python_config(self):
+        """Get the CMake configuration options for Python hints."""
+        options = {}
+        python_root = get_software_root('Python')
+        if python_root:
+            python_version = LooseVersion(get_software_version('Python'))
+            python_exe = os.path.join(python_root, 'bin', 'python')
+            # This is required for (deprecated) `find_package(PythonInterp ...)`
+            options['PYTHON_EXECUTABLE'] = python_exe
+            # Ensure that both `find_package(Python) and find_package(Python2/3)` work as intended
+            options['Python_EXECUTABLE'] = python_exe
+            if python_version >= "3":
+                options['Python3_EXECUTABLE'] = python_exe
+            else:
+                options['Python2_EXECUTABLE'] = python_exe
+        return options
+
+    def get_cmake_python_config_dict(self):
+        """Get a dictionary with the CMake configuration options for Python hints."""
+        return self._get_cmake_python_config()
+
+    def get_cmake_python_config_str(self):
+        """Get a string with the CMake configuration options for Python hints."""
+        options = self._get_cmake_python_config()
+        return ' '.join(['-D%s=%s' % (key, value) for key, value in options.items()])
+
+    def set_cmake_python_env_hints(self):
         """Convenience function to set CMake hints for FindPython[_2/3] as environment variables.
         Needed to avoid wrong Python being picked up by CMake when not called directly by EasyBuild but as step in a
         build and no option is provided to set custom CMake variables.
--- a/easybuild/easyblocks/generic/cmakemake.py
+++ b/easybuild/easyblocks/generic/cmakemake.py
@@ -289,10 +289,10 @@ class CMakeMake(ConfigureMake):
         if python_root:
             python_version = LooseVersion(get_software_version('Python'))
             python_exe = os.path.join(python_root, 'bin', 'python')
-            options['Python_EXECUTABLE'] = python_exe
-            # This is needed if someone is still using `find_package(PythonInterp ...)` in their CMakeLists.txt
+            # This is required for (deprecated) `find_package(PythonInterp ...)`
             options['PYTHON_EXECUTABLE'] = python_exe
             # Ensure that both `find_package(Python) and find_package(Python2/3)` work as intended
+            options['Python_EXECUTABLE'] = python_exe
             if python_version >= "3":
                 options['Python3_EXECUTABLE'] = python_exe
             else:
--- a/easybuild/easyblocks/generic/cmakemake.py
+++ b/easybuild/easyblocks/generic/cmakemake.py
@@ -87,7 +87,7 @@ def setup_cmake_env(tc):
     setvar("CMAKE_LIBRARY_PATH", library_paths)
 
 
-def setup_cmake_env_python_hints(cmake_version):
+def setup_cmake_env_python_hints(cmake_version=None):
     """Convenience function to set CMake hints for FindPython[_2/3] as environment variables.
     Needed to avoid wrong Python being picked up by CMake when not called directly by EasyBuild but as step in a
     build and no option is provided to set custom CMake variables.
--- a/easybuild/easyblocks/generic/cmakemake.py
+++ b/easybuild/easyblocks/generic/cmakemake.py
@@ -336,6 +336,22 @@ class CMakeMake(ConfigureMake):
 
         return out
 
+    def set_cmake_env_vars_python(self):
+        """Convenience function to set CMake hints for FindPython[_2/3] as environment variables.
+        Needed to avoid wrong Python being picked up by CMake when not called directly by EasyBuild but as step in a
+        build and no option is provided to set custom CMake variables.
+        """
+        if LooseVersion(self.cmake_version) < '3.12':
+            raise EasyBuildError("Setting Python hints for CMake requires CMake version 3.12 or newer")
+        python_root = get_software_root('Python')
+        if python_root:
+            python_version = LooseVersion(get_software_version('Python'))
+            setvar('Python_ROOT_DIR', python_root)
+            if python_version >= "3":
+                setvar('Python3_ROOT_DIR', python_root)
+            else:
+                setvar('Python2_ROOT_DIR', python_root)
+
     def test_step(self):
         """CMake specific test setup"""
         # When using ctest for tests (default) then show verbose output if a test fails
--- a/easybuild/easyblocks/generic/cmakemake.py
+++ b/easybuild/easyblocks/generic/cmakemake.py
@@ -284,9 +284,19 @@ class CMakeMake(ConfigureMake):
         # disable CMake user package repository
         options['CMAKE_FIND_USE_PACKAGE_REGISTRY'] = 'OFF'
 
-        # ensure CMake uses PATH to determine python without prioritizing a virtualenv
-        # necessary to pick up on the correct python version when `eb` is run from a virtualenv
-        options['Python3_FIND_VIRTUALENV'] = 'STANDARD'
+        # ensure CMake uses EB python, not system or virtualenv python
+        python_root = get_software_root('Python')
+        if python_root:
+            python_version = LooseVersion(get_software_version('Python'))
+            python_exe = os.path.join(python_root, 'bin', 'python')
+            options['Python_EXECUTABLE'] = python_exe
+            # This is needed if someone is still using `find_package(PythonInterp ...)` in their CMakeLists.txt
+            options['PYTHON_EXECUTABLE'] = python_exe
+            # Ensure that both `find_package(Python) and find_package(Python2/3)` work as intended
+            if python_version >= "3":
+                options['Python3_EXECUTABLE'] = python_exe
+            else:
+                options['Python2_EXECUTABLE'] = python_exe
 
         if not self.cfg.get('allow_system_boost', False):
             boost_root = get_software_root('Boost')
--- a/easybuild/easyblocks/generic/cmakemake.py
+++ b/easybuild/easyblocks/generic/cmakemake.py
@@ -316,6 +316,8 @@ class CMakeMake(ConfigureMake):
                 options['BOOST_ROOT'] = boost_root
                 options['Boost_NO_SYSTEM_PATHS'] = 'ON'
 
+        self.cmake_options = options
+
         if self.cfg.get('configure_cmd') == DEFAULT_CONFIGURE_CMD:
             self.prepend_config_opts(options)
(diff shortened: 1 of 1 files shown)


ocaisa:
@Flamefire You could probably give the most informed review of this. I would be inclined to accept it.

Flamefire:
I don't see any problem with this in general besides increasing the number of (potentially superflous) parameters we pass to configure.

However:
The [GROMACS change](https://github.com/easybuilders/easybuild-easyblocks/pull/3283) was very focused on that specific software and restricted to "faulty" versions. Here we'd not only need to pass `Python3_FIND_VIRTUALENV` but also `Python_FIND_VIRTUALENV` and `Python_FIND_VIRTUALENV` to cover all cases.

I'd prefer to pass the Python binary path explicitly as done in [this PR](https://github.com/easybuilders/easybuild-easyblocks/pull/3282) This not only avoids it finding a virtualenv Python ("fixed" here) but also covers many other cases we might not found yet.   
So that is more general without unnecessarily increasing verbosity for cases where we don't need it.

@Crivella What do you think about that? Can you try #3282 with your https://github.com/easybuilders/easybuild-easyblocks/pull/3373 not setting the `Python3_FIND_VIRTUALENV`  to see if that works?

Crivella:
@Flamefire  Started the test it will take a couple hours to finish, but since stage 1 passed i don't think there would be a problem.
My concern was the order with which the hints where evaluated since #3283 was necessary, but `Python_ROOT_DIR` seems to take precedence above all

Just ran a minimal test with

```
    cmake_minimum_required(VERSION 3.15.0)

    project(MyProject)

    find_package(Python3)
```

and

```
#!/usr/bin/env bash

module purge
module load CMake/3.29.3-GCCcore-13.3.0
module load Python/3.12.3-GCCcore-13.3.0
module list

source $HOME/.virtualenvs/test/bin/activate
which python

rm -rf build
cmake -B build -S . \
    -DPython3_ROOT_DIR=/usr/bin \
    -DPython3_FIND_VIRTUALENV=ONLY
```

which results in 

```
...
-- Found Python3: /usr/bin/python3.12 (found version "3.12.5") found components: Interpreter
...
```

Was there some weirder reason that did not make GROMACS work with #3282 ?

Flamefire:
I'll try to compile my findings on the order of paths considered.

The documentation states:
```
    `Python3_FIND_VIRTUALENV`  variable can be set to one of the following:

        FIRST: The virtual environment is used before any other standard paths to look-up for the interpreter. This is the default.

        ONLY: Only the virtual environment is used to look-up for the interpreter.

        STANDARD: The virtual environment is not used to look-up for the interpreter but environment variable PATH is always considered. In this case, variable Python3_FIND_REGISTRY (Windows) or CMAKE_FIND_FRAMEWORK (macOS) can be set with value LAST or NEVER to select preferably the interpreter from the virtual environment.
```

The relevant file is https://gitlab.kitware.com/cmake/cmake/-/blob/master/Modules/FindPython/Support.cmake which gets "called" by "FindPython*" with `${_PYTHON_PREFIX}` set to e.g. "Python3"

We can ignore the "STANDARD" for now because when `*_FIND_VIRTUALENV` is not set to FIRST or ONLY then `$CMAKE_PREFIX_PATH` is used which we set in the module.

Which [leads to](https://gitlab.kitware.com/cmake/cmake/-/blob/master/Modules/FindPython/Support.cmake#L1932-1942):

And [`find_program` considers](https://cmake.org/cmake/help/latest/command/find_program.html):

1. `<PackageName>_ROOT` CMake variable, where `<PackageName>` is the case-preserved package name.
2. `<PACKAGENAME>_ROOT` CMake variable, where `<PACKAGENAME>` is the upper-cased package name.
3. HINTS: ENV variable `${_PYTHON_PREFIX}_ROOT_DIR`
4. PATHS: ENV variables `VIRTUAL_ENV` & `CONDA_PREFIX`

I excluded all of the ones mentioned in the docs that are excluded by the explicit `NO_*` options. So we should be safe. However there is "[The first 2 steps] can be skipped by setting the `CMAKE_FIND_USE_PACKAGE_ROOT_PATH` variable to `FALSE`.

So maybe we need to also set the `*_ROOT_DIR` env variables to be safe to not get a virtualenv.

Crivella:
By looking at the CMake source you linked (never realized it was so messy XD) and experimenting a bit more, I am wondering if we should just cut to the source and set `Python_EXECUTABLE` and/or `Python[2/3]_EXECUTABLE` to avoid going to the `HINTS` part of the code.

It seems to also find the correct components when specifying `COMPONENTS Development`

test.sh
```
#!/usr/bin/env bash

module purge
module load CMake/3.29.3-GCCcore-12.3.0
module load Python/3.11.3-GCCcore-12.3.0
module load SciPy-bundle/2023.07-gfbf-2023a
module list

source $HOME/.virtualenvs/test/bin/activate
which python

_PYTHON="/home/crivella/.local/easybuild/software/Python/3.11.3-GCCcore-12.3.0/bin/python"

rm -rf build
cmake -B build -S . \
    -DPython_FIND_VIRTUALENV=ONLY \
    -DPython_ROOT=/usr/bin \
    -DPython_EXECUTABLE=${_PYTHON} \
```

CMakeLists.txt
```
cmake_minimum_required(VERSION 3.15.0)

project(MyProject)

find_package(
    Python 3.11.0...3.11.10
    COMPONENTS Interpreter Development NumPy
    )

```

CMakeCache.txt
```
...
//Compiler reason failure
_Python_Compiler_REASON_FAILURE:INTERNAL=
_Python_DEVELOPMENT_EMBED_SIGNATURE:INTERNAL=6e4a7d11dfc3083ba9e17dd9c0c9ffd5
_Python_DEVELOPMENT_MODULE_SIGNATURE:INTERNAL=7f9cffd5f241febb4fdd6ed780db4712
//Development reason failure
_Python_Development_REASON_FAILURE:INTERNAL=
_Python_EXECUTABLE:INTERNAL=/home/crivella/.local/easybuild/software/Python/3.11.3-GCCcore-12.3.0/bin/python
//Path to a file.
_Python_INCLUDE_DIR:INTERNAL=/home/crivella/.local/easybuild/software/Python/3.11.3-GCCcore-12.3.0/include/python3.11
//Python Properties
_Python_INTERPRETER_PROPERTIES:INTERNAL=Python;3;11;3;64;;cpython-311-x86_64-linux-gnu;abi3;/home/crivella/.local/easybuild/software/Python/3.11.3-GCCcore-12.3.0/lib/python3.11;/home/crivella/.local/easybuild/software/Python/3.11.3-GCCcore-12.3.0/lib/python3.11;/home/crivella/.local/easybuild/software/Python/3.11.3-GCCcore-12.3.0/lib/python3.11/site-packages;/home/crivella/.local/easybuild/software/Python/3.11.3-GCCcore-12.3.0/lib/python3.11/site-packages
_Python_INTERPRETER_SIGNATURE:INTERNAL=a6927044a8e489828e7111e599980346
//Path to a library.
_Python_LIBRARY_RELEASE:INTERNAL=/home/crivella/.local/easybuild/software/Python/3.11.3-GCCcore-12.3.0/lib/libpython3.11.so
_Python_NUMPY_SIGNATURE:INTERNAL=64759e84174ed96a40e217a0380274d0
//Path to a file.
_Python_NumPy_INCLUDE_DIR:INTERNAL=/home/crivella/.local/easybuild/software/SciPy-bundle/2023.07-gfbf-2023a/lib/python3.11/site-packages/numpy/core/include
...
```

The fact that  `_Python_Development_REASON_FAILURE` is still there is due to it never getting unset in the code, but `_Python_LIBRARY_RELEASE:` pointing to the correct .so makes me thing stuff is working correctly

Flamefire:
> I am wondering if we should just cut to the source and set `Python_EXECUTABLE` and/or `Python[2/3]_EXECUTABLE` 

Yes I guess that would also work although I don't know if there is some (old?) CMake user code out there that relies on `Python_ROOT`. We'd need to check at least the history of the module in the CMake repo.

I also guess there is no trouble in using `$EBROOT_PYTHON/bin/python` (resolved by the easyblock) as the param value

Crivella:
I've been browsing through the CMake source code previous versions.
I've also tested up to CMake v3.12 where `FindPython` was introduced and `Python_EXECUTABLE` works for all of them.

I think the logic was properly implemented in 3.16, using `_${PYTHON_PREFIX}_EXECUTABLE` as the store variable for `find_program` but even before that, since `find_program` is being called directly on `${PYTHON_PREFIX}_EXECUTABLE` due to the behavior of `find_program`

> If the program is found the result is stored in the variable and the search will not be repeated unless the variable is cleared. If nothing is found, the result will be <VAR>-NOTFOUND.

This is still working since we are manually setting the result to what we want and `${PYTHON_PREFIX}_EXECUTABLE` is not being unset by the module.

I also tested that there is no problem in case the python version does not matches the requirements since the version is later validated even if we are forcing the `find_program` result.

I've made a new commit with the proposed changes (https://github.com/easybuilders/easybuild-easyblocks/pull/3463/commits/47f21b974fda5c4b4404e09f3a677de0e0667b49).

I would argue that unless we want to analyze the CMakeLists.txt we have to set both `Python_EXECUTABLE` and `PYTHON_EXECUTABLE` in case some code is still using the deprecated `find_package{PythonInterp ...)` (unfortunately they are case sensitive)

Crivella:
I guess there is also an argument to be made to actually set `Python_EXECUTABLE` to a dump location so that if people are making an EC for a code that requires python they are forced to use one through EB modules instead of randomly picking up a system one for reproducibility?

easybuild/easyblocks/generic/cmakemake.py
```diff
@@ -284,6 +284,20 @@ def configure_step(self, srcdir=None, builddir=None):
...
+        # ensure CMake uses EB python, not system or virtualenv python
+        python_root = get_software_root('Python')
+        if python_root:
+            python_version = LooseVersion(get_software_version('Python'))
+            python_exe = os.path.join(python_root, 'bin', 'python')
+            options['Python_EXECUTABLE'] = python_exe
+            # This is needed if someone is still using `find_package(PythonInterp ...)` in their CMakeLists.txt
+            options['PYTHON_EXECUTABLE'] = python_exe
+            # Ensure that both `find_package(Python) and find_package(Python2/3)` work as intended
+            if python_version >= "3":
```
Flamefire, line 296: Just some IMO better wording as the first option should be below the comment referring to it:
```suggestion
            # This is required for (deprecated) `find_package(PythonInterp ...)`
            options['PYTHON_EXECUTABLE'] = python_exe
            # Ensure that both `find_package(Python) and find_package(Python2/3)` work as intended
            options['Python_EXECUTABLE'] = python_exe
            if python_version >= "3":
```

Flamefire (commented):
I think this works. https://gitlab.kitware.com/cmake/cmake/-/blob/d0ad8fd49c2d211081d32cdc2a7f96990947c340/Modules/FindPython/Support.cmake?page=2#L1877-1878 checks for an absolute path in that variable and always keeps it so the block starting at https://gitlab.kitware.com/cmake/cmake/-/blob/d0ad8fd49c2d211081d32cdc2a7f96990947c340/Modules/FindPython/Support.cmake?page=2#L1909 gets skipped.

However https://github.com/easybuilders/easybuild-easyblocks/pull/3282/files should work for CMake >= 3.12 while this only works for >= 3.16 which includes https://gitlab.kitware.com/cmake/cmake/-/commit/cea2010b5c7 that honors this variable

>  I've also tested up to CMake v3.12 where FindPython was introduced and Python_EXECUTABLE works for all of them.

What exactly have you tested? Did you test 3.12 and below or 3.12 and up? Because how I understand the change in 3.16 the `*_EXECUTABLE` isn't honored before.

However checking the code for 3.15 it eventuall runs 
```
        find_program (${_PYTHON_PREFIX}_EXECUTABLE
                      NAMES ${_${_PYTHON_PREFIX}_NAMES}
                            ${_${_PYTHON_PREFIX}_IRON_PYTHON_NAMES}
                      NAMES_PER_DIR
                      PATHS ${_${_PYTHON_PREFIX}_REGISTRY_PATHS}
                      PATH_SUFFIXES bin ${_${_PYTHON_PREFIX}_IRON_PYTHON_PATH_SUFFIXES}
                      NO_DEFAULT_PATH)
        _python_validate_interpreter (${${_PYTHON_PREFIX}_FIND_VERSION})
```

So unless the python cmd set is invalid it should be kept because `find_program` does nothing if the variable is already set.

Hence I think this change is OK

Crivella:
> What exactly have you tested? Did you test 3.12 and below or 3.12 and up? Because how I understand the change in 3.16 the *_EXECUTABLE isn't honored before.

Checked 3.12 and above since before there is only `FindPythonInterp` which has not been changed

> So unless the python cmd set is invalid it should be kept because find_program does nothing if the variable is already set.

Yes this is what i meant in my previous comment, even if it is not explicitly the behavior of `FindPython` due to the behavior of `find_program` as long as we set `${PYTHON_PREFIX}_EXECUTABLE` to a correct path there should be no problem (also concerning version checks since they happens after and will give the correct error if we are passing a non supported version of python)

Flamefire:
GROMACS 2023.x fails during the installation of the Python extension which uses CMake through `pip` and then runs into the same problem. We do that extension installation (in the current way) only since 2023.x and 2024 uses scikit-build which initializes the CMakeCache with some Python variables.

I'm not sure how to solve this. The extension in this EasyConfig is a "regular" `PythonPackage` but it supports an env variable `CMAKE_ARGS` to pass arguments to CMake.

We could likely also set `Python3_ROOT_DIR` as env variables which should also avoid it finding the virtualenv.

However: Any ideas how to pass either the arguments or the env variables to the extension? I guess the gromacs easyblock could set the env variables before installing the extensions but then it would be good to have some way to get the values from the cmake easyblock to avoid duplication

Crivella:
Have not worked much with extensions yet, but by looking at the code I think the easiest  solution would be to modify the `CMAKE_ARGS` to define either `Python3_ROOT_DIR` or `Python3_EXECUTABLE`.
That could be done either overriding [install_extensions_sequential](https://github.com/easybuilders/easybuild-framework/blob/8ff6ba0d426b79a9b3da1248a7359d922b50596d/easybuild/framework/easyblock.py#L1902) setting the variable before calling the original method eg

```
    def install_extensions_sequential(self, *args, **kwargs):
        """Install the gmxapi extension"""
        python_bin = os.path.join(python_root, 'bin', 'python')
        env.setvar(
            'CMAKE_ARGS', 
            "-Dgmxapi_ROOT=%s -DPython3_EXECUTABLE=%s -C %s/share/cmake/gromacs_mpi/gromacs-hints_mpi.cmake" % 
            (self.installdir, python_bin, self.installdir)
            )
        super(EB_GROMACS, self).install_extensions_sequential(*args, **kwargs)
```

or possibly defining an hook for the extension (https://github.com/easybuilders/easybuild-framework/blob/8ff6ba0d426b79a9b3da1248a7359d922b50596d/easybuild/framework/easyblock.py#L1916)

I am not sure if there is a way to modify the extension steps in some other way

> However: Any ideas how to pass either the arguments or the env variables to the extension? I guess the gromacs easyblock could set the env variables before installing the extensions but then it would be good to have some way to get the values from the cmake easyblock to avoid duplication

I guess this would be a design choice. I can think of two way:

- Split the creation of the `options` dictionary from the configure step to a separate function. This would allow a child easyblock to just call that function and then pick and use the options it needs (or atleast save the options after configure is called to allow for later reuse)
- Add a convenience function eg `get_cmakeopts_python` to get all cmake options related to python

Also I would say the 2 are not mutually exclusive and both could benefit also from implementing a *scoped* version of the `options` dictionary where options are split into groups that make sense to recall together  eg ('policies', 'generic/base', 'software_[python/...]', ...)

Flamefire:
> Have not worked much with extensions yet, but by looking at the code I think the easiest solution would be to modify the `CMAKE_ARGS` to define either `Python3_ROOT_DIR` or `Python3_EXECUTABLE`. That could be done either overriding [install_extensions_sequential](https://github.com/easybuilders/easybuild-framework/blob/8ff6ba0d426b79a9b3da1248a7359d922b50596d/easybuild/framework/easyblock.py#L1902) setting the variable before calling the original method eg

That specific variable is only for GROMACS: https://github.com/easybuilders/easybuild-easyconfigs/blob/ef56e11c4066828fa4af98b4709ba40728fc4cc6/easybuild/easyconfigs/g/GROMACS/GROMACS-2024.1-foss-2023b.eb#L78

Hence the idea for that specific easyblock to get required env vars from cmakemake (e.g. `Python3_ROOT_DIR`) to set them before installing an extension. For that it might be easier, if we use those env vars in cmakemake too, instead of *_EXECUTABLE.
>     * Add a convenience function eg `get_cmakeopts_python` to get all cmake options related to python

Yes that is what I had in mind: `get_cmake_env_vars_python` or so. Or even just `get_cmake_env_vars` and it would detect which ones to set dependent on loaded modules, e.g. add the python ones if Python is somewhere in the dependencies

Crivella:
I tested the snippet in the previous comment (added to the gromacs easyblock) and it works also by passing `Python3_EXECUTABLE` to `CMAKE_ARGS`. I think it is just about making the `find_package` [here](https://github.com/gromacs/gromacs/blob/9e01c152b5cc17537ac0fc9f8950c5416781e4d8/python_packaging/gmxapi/CMakeLists.txt#L76) happy.

> And GROMACS was the only example I know of where the virtual env caused problems before

I've also this problem in the LLVM easyblock, since I have to do a multistage build while changing CMake variables across the stages. I ended up just reimplementing part of the CMakeMake easyblock and working with a new dictionary, but also there it might help to be able to inherit all/selectively the options from the CMakeMake.configure

Flamefire:
> I tested the snippet in the previous comment (added to the gromacs easyblock) and it works also by passing `Python3_EXECUTABLE` to `CMAKE_ARGS`. I think it is just about making the `find_package` [here](https://github.com/gromacs/gromacs/blob/9e01c152b5cc17537ac0fc9f8950c5416781e4d8/python_packaging/gmxapi/CMakeLists.txt#L76) happy.

I'm not sure how that works from within the easyblock as the variable is set in the easyconfig. So as far as I see even if you set it in the easyblock the easyconfig will overwrite this. Or what exactly did you change and which EC did you test? `GROMACS-2023.1-foss-2022a.eb` from within a minimal virtualenv is one that fails for me. One of my tests uses `export Python3_ROOT_DIR=$EBROOTPYTHON && export CMAKE_ARGS=...`. This is running over the (here extended) weekend

> I've also this problem in the LLVM easyblock, since I have to do a multistage build while changing CMake variables across the stages. I ended up just reimplementing part of the CMakeMake easyblock and working with a new dictionary, but also there it might help to be able to inherit all/selectively the options from the CMakeMake.configure

That might be a valuable datapoint: If it works easily for GROMACS base, the extension, and LLVM we have quite a few cases covered. But I guess for LLVM setting either the env variable or the `*_EXECUTABLE` argument would work as you call the cmakemake configure function

I'm just wondering if using the env variables is more flexible as we won't need to figure out how to pass flags to all the cmake invocation, such as in the `pip install` case with some custom env variable that might/will be different for each software if there even is any.

Crivella:
> I'm not sure how that works from within the easyblock as the variable is set in the easyconfig.

Sorry forgot to mention i also removed the `preinstallopts` from the easyconfig

> I'm just wondering if using the env variables is more flexible as we won't need to figure out how to pass flags to all the cmake invocation, such as in the pip install case with some custom env variable that might/will be different for each software if there even is any.

I am not sure, I think the env is not preserved (or reinitialized) across steps so it is possible the environment variable will have to be set in the needed step on a case by case basis?
Not entirely sure, but before i tried adding that snippet before [the call to easyblock.extensions_step](https://github.com/easybuilders/easybuild-easyblocks/blob/bd7fe7e016211705b5e5c5c5c847119f1c5d119c/easybuild/easyblocks/g/gromacs.py#L545), and it was carried to the installation of the extension

Flamefire:
> I am not sure, I think the env is not preserved (or reinitialized) across steps so it is possible the environment variable will have to be set in the needed step on a case by case basis? Not entirely sure, but before i tried adding that snippet before [the call to easyblock.extensions_step](https://github.com/easybuilders/easybuild-easyblocks/blob/bd7fe7e016211705b5e5c5c5c847119f1c5d119c/easybuild/easyblocks/g/gromacs.py#L545), and it was carried to the installation of the extension

IIRC the environment is (partially?) reset when [loading the fake module](https://github.com/easybuilders/easybuild-framework/blob/8ff6ba0d426b79a9b3da1248a7359d922b50596d/easybuild/framework/easyblock.py#L2887) for the extension installation. But I think we can use either `prepare_for_extensions` or `install_extensions` to set the env variables.   
And yes I guess that would be easyblock specific. But I imagine something like this for easyblocks like GROMACS:
```
def prepare_for_extensions(self):
    super().prepare_for_extensions()
    set_cmake_env_vars_python()
```

That would be slightly more generic than only getting the `-D` options. Although I do like to be able to directly set the python executables. Especially as that is more "focused/controlled" than an env variable. But not sure if it is always possible

Crivella:
I've added both the storing of the generated `options` inside the class and the suggested convenience function.
Indeed when calling CMake from within a build process there is no defined way to pass it arguments and it will dependent on the implementation.

Concerning the usage of the environment variables one thing to be careful of: i could also see the edge case where the buldsystem itself reset/starts from a new environment before making the call to CMake, but at that point I do not think there is much that can be done beside abiding to the buildsystem way of passing args to CMake or patching.

Flamefire (changes requested):
> I've added both the storing of the generated `options` inside the class and the suggested convenience function. Indeed when calling CMake from within a build process there is no defined way to pass it arguments and it will dependent on the implementation.

Yes storing the options is a good idea.

> Concerning the usage of the environment variables one thing to be careful of: i could also see the edge case where the buldsystem itself reset/starts from a new environment before making the call to CMake

Yes, Bazel is one buildsystem that does reset the environment. However that doesn't call CMake AFAIK. I'm not aware of anything that calls CMake but resets the environment.

>I do not think there is much that can be done beside abiding to the buildsystem way of passing args to CMake or patching.

Fully agreed.

I'm not sure if we should use 2 different approaches here (args and env variables). We might end up using almost always one approach and never notice when the rarely used one fails or becomes outdated. Using just the environment variables, not the `*_EXECUTABLES` args might be enough:

Before CMake 3.12 there was only `FindPythonInterp` which uses a [plain `find_program`](https://gitlab.kitware.com/cmake/cmake/-/blob/v3.11.0/Modules/FindPythonInterp.cmake?ref_type=tags#L67) Hence that will always pick up the one we have first in PATH which is ours.

After 3.12 it should pick up the one from the env variable according to my [earlier analysis](https://github.com/easybuilders/easybuild-easyblocks/pull/3463#issuecomment-2378748936). However only when `Python*_FIND_STRATEGY` is set to `LOCATION` or [the policy is set to new](https://github.com/easybuilders/easybuild-easyblocks/blob/bd7fe7e016211705b5e5c5c5c847119f1c5d119c/easybuild/easyblocks/generic/cmakemake.py#L279). If that is not the case (e.g. in the GROMACS python extension case) I think it searches for the `python3.13` binary first [in this loop](https://gitlab.kitware.com/cmake/cmake/-/blob/d734f8dc6dec8834b94a77f06836e3c457ede3fc/Modules/FindPython/Support.cmake?page=4#L2044) and if it finds it in the virtualenv after not finding it in our module (e.g. Python 3.10 is the dependency) it will still use the virtualenv python.

So after writing all this I see that we can't reliably control the chosen Python binary via the environment. Hence for direct cmake invocations we should use `*_EXECUTABLE` and remove the [policy workaround](https://github.com/easybuilders/easybuild-easyblocks/blob/bd7fe7e016211705b5e5c5c5c847119f1c5d119c/easybuild/easyblocks/generic/cmakemake.py#L279). The env variables might still be useful in some cases. E.g. IIRC GROMACS changes the search strategy to LOCATION so it will pick up the env variables first.

easybuild/easyblocks/generic/cmakemake.py
```diff
@@ -320,6 +336,22 @@ def configure_step(self, srcdir=None, builddir=None):
 
         return out
 
+    def set_cmake_env_vars_python(self):
```
boegel, line 339: I'm a bit confused by this one, what's the plan with this?

As is, it can only be used in easyblocks that derive from `CMakeMake`?

Flamefire: Yes I would make 2 free functions out of that:
1) Return a configure string with `-DPython*_EXECUTABLE=...`, maybe optionally only as a dict
2) Return these env variables

The first function is the safe way for every situation we can pass CMake variables. We can use that to enhance the GROMACS easyblock to move the CMake config options out from the EasyConfig to the easyblock and add those options.   

The 2nd function can be used whenever we cannot use the first. It *might* still pick up a `python` from a virtualenv when `Python*_FIND_STRATEGY` isn't set to `LOCATION` ("older" codes default this to LOCATION if unset) but at least we did a best effort to avoid that which might work for most cases.

Crivella: Added the 2 separate functions to return the options as a dict or as a string:

- `get_cmake_python_config_dict`
- `get_cmake_python_config_str`

 and changed the name of the function to set the env vars to `set_cmake_python_env_hints`  (https://github.com/easybuilders/easybuild-easyblocks/pull/3463/commits/9a31f79bf8c3c4d9bb5e18ecf6d6f6f21e57d486)

There might be an argument of making that function a context manager to be used with `with`

easybuild/easyblocks/generic/cmakemake.py
```diff
@@ -320,6 +325,48 @@ def configure_step(self, srcdir=None, builddir=None):
...
+            options['Python_EXECUTABLE'] = python_exe
+            if python_version >= "3":
+                options['Python3_EXECUTABLE'] = python_exe
+            else:
+                options['Python2_EXECUTABLE'] = python_exe
+        return options
+
+    def get_cmake_python_config_dict(self):
+        """Get a dictionary with the CMake configuration options for Python hints."""
+        return self._get_cmake_python_config()
```
Flamefire, line 347: I don't see a need for this function if it does nothing but forwarding, so they are the same function.

Flamefire (changes requested):
I'd do 2 more changes:

- Make the 2 functions freestanding. They don't use anything from the class instance so other easyblocks not deriving from CMakeMake directly should be able to use them too
- Remove the policy code at https://github.com/easybuilders/easybuild-easyblocks/blob/bd7fe7e016211705b5e5c5c5c847119f1c5d119c/easybuild/easyblocks/generic/cmakemake.py#L276-L280 which is now better handled by the new arguments: Where we'd pass the policy argument we'll also pass the EXECUTABLEs making the policy irrelevant, doesn't it?

easybuild/easyblocks/generic/cmakemake.py
Crivella: Implemented with [fa30b9c](https://github.com/easybuilders/easybuild-easyblocks/pull/3463/commits/fa30b9cd6c40ec1a02db44b802c6cb39525ec338)

Flamefire: But there are still 2 functions where `get_cmake_python_config_dict  == _get_cmake_python_config`. I'd just use one.

Crivella: It was to have them more modularized in case in the future there is need for more separation between the dict and str functions, but i guess for now they can just be one

easybuild/easyblocks/generic/cmakemake.py
```diff
@@ -87,6 +87,48 @@ def setup_cmake_env(tc):
     setvar("CMAKE_LIBRARY_PATH", library_paths)
 
 
+def setup_cmake_env_python_hints(cmake_version):
+    """Convenience function to set CMake hints for FindPython[_2/3] as environment variables.
+    Needed to avoid wrong Python being picked up by CMake when not called directly by EasyBuild but as step in a
+    build and no option is provided to set custom CMake variables.
```
Flamefire, line 93:
```suggestion
    """Set environment variables as hints for CMake to prefer the Python module, if loaded.
    Useful when there is no way to specify arguments for CMake directly,
    e.g. when CMake is called from within another build system.
    Otherwise get_cmake_python_config_[str/dict] should be used instead.
```

easybuild/easyblocks/generic/cmakemake.py
```diff
@@ -87,6 +87,48 @@ def setup_cmake_env(tc):
...
+        python_version = LooseVersion(get_software_version('Python'))
+        setvar('Python_ROOT_DIR', python_root)
+        if python_version >= "3":
+            setvar('Python3_ROOT_DIR', python_root)
+        else:
+            setvar('Python2_ROOT_DIR', python_root)
+
+
+def get_cmake_python_config_dict():
+    """Get a dictionary with the CMake configuration options for Python hints."""
```
Flamefire, line 109:
```suggestion
    """Get a dictionary with CMake configuration options for finding Python if loaded as a module."""
```

easybuild/easyblocks/generic/cmakemake.py
```diff
@@ -87,6 +87,48 @@ def setup_cmake_env(tc):
...
+        options['Python_EXECUTABLE'] = python_exe
+        if python_version >= "3":
+            options['Python3_EXECUTABLE'] = python_exe
+        else:
+            options['Python2_EXECUTABLE'] = python_exe
+    return options
+
+
+def get_cmake_python_config_str():
+    """Get a string with the CMake configuration options for Python hints."""
```
Flamefire, line 127:
```suggestion
    """Get CMake configuration arguments for finding Python if loaded as a module.
    This string is intended to be passed to the invocation of `cmake`.
    """
```

easybuild/easyblocks/generic/cmakemake.py
```diff
@@ -87,6 +87,48 @@ def setup_cmake_env(tc):
...
+    options = get_cmake_python_config_dict()
+    return ' '.join(['-D%s=%s' % (key, value) for key, value in options.items()])
```
Flamefire, line 129: No need to create a list first:
```suggestion
    return ' '.join('-D%s=%s' % (key, value) for key, value in options.items())
```

easybuild/easyblocks/generic/cmakemake.py
```diff
@@ -87,6 +87,48 @@ def setup_cmake_env(tc):
     setvar("CMAKE_LIBRARY_PATH", library_paths)
 
 
+def setup_cmake_env_python_hints(cmake_version):
```
Flamefire, line 90: The argument looks unused? If you want to use it then I'd suggest this and below:
```suggestion
def setup_cmake_env_python_hints(cmake_version=None):
```

easybuild/easyblocks/generic/cmakemake.py
```diff
@@ -87,6 +87,48 @@ def setup_cmake_env(tc):
...
+    """Convenience function to set CMake hints for FindPython[_2/3] as environment variables.
+    Needed to avoid wrong Python being picked up by CMake when not called directly by EasyBuild but as step in a
+    build and no option is provided to set custom CMake variables.
+    """
+    cmake_version = det_cmake_version()
```
Flamefire, line 95:
```suggestion
    if cmake_version is None:
        cmake_version = det_cmake_version()
```

Flamefire (changes requested):
Ok almost done, thanks for bearing with me so far!

If you want to keep the `cmake_version` in the function I'd suggest to include it also in the others: For CMake < 3.12 we only need `PYTHON_EXECUTABLE`

And some suggestions to improve the wording of the docstrings as I'd like to convey the following:
- The options/argument-str should be preferred
- That is not a hint but (pretty much) forces CMake to use that.
- All functions only return some non-empty value when a Python module is loaded
- Avoid suggesting that setting the env variables fully avoids CMake picking up the wrong Python. Due to the policies it can still happen although it is less likely

Besides that I wouldn't mention "FindPython" which likely isn't clear to anyone not familiar with CMake internals. If at all we could mention `find_package(Python*)` as that is what is visible from reading the CML. Also "Convenience function" sounds a bit to verbose for describing what the function does.

Check if you like the suggestions or if there's a better wording.

easybuild/easyblocks/generic/cmakemake.py
Crivella: I had forgotten to remove it but i think it might not hurt to allow it to be passed [7aff4c0](https://github.com/easybuilders/easybuild-easyblocks/pull/3463/commits/7aff4c0026a6c5e4b2b0a605cef4e847b698886b)

Flamefire (approved):
LGTM

ocaisa (approved):
Until @Flamefire goes the last mile and becomes a maintainer, I'll need to merge this!

ocaisa:
EB_ARGS="GROMACS-2023.3-foss-2023a.eb GROMACS-2021.3-foss-2021a.eb HOOMD-blue-4.0.1-foss-2022a.eb vcflib-1.0.9-gfbf-2023a-R-4.3.2.eb --installpath /tmp/$USER/pr3463"
CORE_CNT=16

boegel:
Test report by @boegel
 * **SUCCESS** _ALAMODE-1.4.2-foss-2022b.eb_ 
 * **SUCCESS** _ANTs-2.5.0-foss-2022b.eb_ 
 * **SUCCESS** _AOMP-13.0-2-GCCcore-10.2.0.eb_ 
 * **SUCCESS** _Arcade-Learning-Environment-0.8.1-foss-2023a.eb_ 
 * **SUCCESS** _Arrow-16.1.0-gfbf-2023b.eb_ 
 * **SUCCESS** _awscli-2.11.21-GCCcore-11.3.0.eb_ 
 * **SUCCESS** _btllib-1.7.0-GCC-12.3.0.eb_ 
 * **SUCCESS** _CPPE-0.3.1-GCC-12.2.0.eb_ 
 * **SUCCESS** _CVXPY-1.4.2-foss-2023a.eb_ 
 * **SUCCESS** _CheckM2-1.0.2-foss-2022b.eb_ 
 * **SUCCESS** _cppyy-3.1.2-GCCcore-13.2.0.eb_ 
 * **SUCCESS** _DIRAC-23.0-intel-2023a.eb_ 
 * **SUCCESS** _Doxygen-1.9.1-GCCcore-10.3.0.eb_ 
 * **SUCCESS** _EGTtools-0.1.11-foss-2022a.eb_ 
 * **SUCCESS** _FLINT-3.1.1-gfbf-2023b.eb_ 
 * **SUCCESS** _FlexiBLAS-3.3.1-GCC-13.2.0.eb_ 
 * **SUCCESS** _flatbuffers-23.1.4-GCCcore-12.2.0.eb_ 
 * **SUCCESS** _fumi_tools-0.18.2-GCC-11.2.0.eb_ 
 * **SUCCESS** _HepMC3-3.2.6-GCC-12.3.0.eb_ 
 * **SUCCESS** _HiGHS-1.7.0-gfbf-2023b.eb_ 
 * **SUCCESS** _kineto-0.4.0-GCC-12.3.0.eb_ 
 * **SUCCESS** _libcint-6.1.2-gfbf-2024a.eb_ 
 * **SUCCESS** _MDI-1.4.29-gompi-2023b.eb_ 
 * **SUCCESS** _MEGAHIT-1.2.9-GCCcore-9.3.0.eb_ 
 * **SUCCESS** _MetaBAT-2.15-gompi-2021b.eb_ 
 * **SUCCESS** _ncbi-vdb-3.1.1-gompi-2023b.eb_ 
 * **SUCCESS** _ONNX-1.15.0-gfbf-2023a.eb_ 
 * **SUCCESS** _OpenAI-Gym-0.26.2-foss-2022a.eb_ 
 * **SUCCESS** _OpenCV-4.8.1-foss-2023a-contrib.eb_ 
 * **SUCCESS** _OpenMEEG-2.5.7-foss-2023a.eb_ 
 * **SUCCESS** _OpenMM-8.0.0-foss-2022a.eb_ 
 * **SUCCESS** _PICI-LIGGGHTS-3.8.1-foss-2022a.eb_ 
 * **SUCCESS** _PySCF-2.4.0-foss-2022b.eb_ 
 * **FAIL (build issue)** _PyTorch-bundle-2.1.2-foss-2023a.eb_ (partial log available at https://gist.github.com/boegel/e25d9c0bc294b8b23b3f5524822cc485)
 * **SUCCESS** _pFUnit-4.7.3-gompi-2022a.eb_ 
 * **SUCCESS** _poppler-24.04.0-GCC-13.2.0.eb_ 
 * **SUCCESS** _pybind11-2.9.2-GCCcore-11.3.0.eb_ 
 * **SUCCESS** _pypmt-1.2.0-gfbf-2023a.eb_ 
 * **SUCCESS** _python-mujoco-3.1.4-foss-2023a.eb_ 
 * **FAIL (build issue)** _RDKit-2022.03.5-foss-2021b.eb_ (partial log available at https://gist.github.com/boegel/e67aeb26c548d6bae579b3f9a1b33475)
 * **SUCCESS** _Rust-1.70.0-GCCcore-12.3.0.eb_ 
 * **SUCCESS** _SPAdes-3.15.2-GCC-10.2.0-Python-2.7.18.eb_ 
 * **SUCCESS** _SRA-Toolkit-3.0.5-gompi-2022b.eb_ 
 * **SUCCESS** _SentencePiece-0.2.0-GCC-13.2.0.eb_ 
 * **SUCCESS** _SeqAn-2.4.0-GCCcore-9.3.0.eb_ 
 * **SUCCESS** _scikit-build-core-0.5.0-GCCcore-12.3.0.eb_ 
 * **SUCCESS** _tmap-20220502-GCC-11.2.0.eb_ 
 * **SUCCESS** _UShER-0.5.0-gompi-2021b.eb_ 
 * **SUCCESS** _vcflib-1.0.9-gfbf-2023a-R-4.3.2.eb_ 
 * **SUCCESS** _xtensor-0.24.7-foss-2023a.eb_ 
 * **SUCCESS** _Z3-4.13.0-GCCcore-13.3.0.eb_ 
 * **SUCCESS** _zlib-ng-2.2.1-GCCcore-13.3.0.eb_ 
 * **SUCCESS** _manta-1.6.0-GCC-10.2.0-Python-2.7.18.eb_ 
 * **SUCCESS** _pybind11-2.7.1-GCCcore-11.2.0-Python-2.7.18.eb_ 
 * **SUCCESS** _MetaBAT-2.14-gompi-2019a.eb_ 
 * **SUCCESS** _Gymnasium-0.29.1-foss-2023a.eb_ 
 * **SUCCESS** _ISCE2-2.6.3-foss-2023a.eb_ 

Build succeeded for 56 out of 58 (58 easyconfigs in total)
node4002.donphan.os - Linux RHEL 8.8, x86_64, Intel(R) Xeon(R) Gold 6240 CPU @ 2.60GHz, 1 x NVIDIA NVIDIA A2, 545.23.08, Python 3.6.8
See https://gist.github.com/boegel/01bd2c5d78f5e0776a55c63ceeb283f0 for a full test report.

boegel:
* `PyTorch-bundle-2.1.2-foss-2023a.eb` failed due with error `-9`, so probably a segfault (could be due to limited available memory in the test environment, only ~27GB
* `RDKit-2022.03.5-foss-2021b.eb` failed with error below, has nothing to do with the changes being made here
   ```
   Downloading https://fonts.google.com/download?family=Comic%20Neue...
   CMake Error at Code/cmake/Modules/RDKitUtils.cmake:257 (MESSAGE):
     The md5 checksum for
     /tmp/easybuild_build/RDKit/2022.03.5/foss-2021b/rdkit-Release_2022_03_5/Code/GraphMol/MolDraw2D/Comic_Neue.zip
     is incorrect; expected: 850b0df852f1cda4970887b540f8f333, found:
     4cd32046e7d03f3905d0d6c433bfae43
   ```

PR #3468, opened 2024-10-03
Bugfix for MetalWalls easyblock 

Crivella:
Fix exception caused by calling `LoosVersion` on `f90wrap_version` even when it is None


boegel:
EB_ARGS="MetalWalls-21.06.1-foss-2023a.eb"

boegel (approved):
lgtm

PR #3477, opened 2024-10-11
fix `enhance_sanity_check` for `PythonPackage`

Flamefire:
When trying to add, not replace paths/commands for the sanity check of Python packages the default `lib/python/%(pyshortver)s/site-packages` gets replaced.
This leads to duplicating this in many easyconfigs. Use the `custom_paths` argument for that but still honor if that is given from a child class.

- [x] Requires https://github.com/easybuilders/easybuild-framework/pull/4679

### DIFF
--- a/easybuild/easyblocks/generic/pythonpackage.py
+++ b/easybuild/easyblocks/generic/pythonpackage.py
@@ -1012,7 +1012,7 @@ class PythonPackage(ExtensionEasyBlock):
         else:
             self.log.debug("Detection of downloaded dependencies not enabled")
 
-        # Use directory usually containg the python packages by default
+        # Use directory usually containing the python packages by default
         # but only for stand-alone installations, not for extensions;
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/generic/pythonpackage.py
```diff
@@ -1029,15 +1029,15 @@ def sanity_check_step(self, *args, **kwargs):
         else:
             self.log.debug("Detection of downloaded dependencies not enabled")
 
-        # inject directory path that uses %(pyshortver)s template into default value for sanity_check_paths,
+        # Use directory usually containg the python packages by default
```
boegel, line 1032:
```suggestion
        # Use directory usually containing the python packages by default
```

boegel (changes requested):

boegel (approved):
lgtm

PR #3481, opened 2024-10-14
avoid name clash in cargo crates downloaded from git repositories

Flamefire:
Crates referenced by their git revision have a version but might have different revisions.
The source filename ignored the revision such that sources from different revisions were considered the same file which either fails the checksum verification or the build.
Append the revision if specified to disambiguate them.


Micket (approved):
lgtm

Micket:
Going with the combined PR to save myself multiple steps of testing

PR #3482, opened 2024-10-14
Handle cargo workspaces in git repositories

Flamefire:
Crates downloaded from crates.io can just be put into the vendored-sources directory.
However a referenced git repository might contain multiple crates in a so-called "workspace".
If we put such a workspace into the vendor folder, cargo fails with
> found a virtual manifest at [...]Cargo.toml instead of a package manifest

This solution uses a separate folder for sources downloaded from git repositories and checks them for workspaces.
If they are "regular", i.e. don't have a workspace, they are put to the vendored sources from crates.io.
Otherwise they are left in the git-vendor directory.

Additionally refine the config.toml entry for sources vendored from git. This allows to override a specific git URL and revision with the path to the extracted repository.
Otherwise we would need entries for each crate in workspaces but then cargo tries to verify the revision and fails with
> Unable to update https://github.com/[...]?rev=[...]
> Caused by: can't checkout from 'https://github.com/[...]': you are in the offline mode (--offline)

I also have a Stack Overflow question on that in case there is a better solution: https://stackoverflow.com/questions/79065892/how-to-patch-a-cargo-git-dependency-that-contains-a-workspace

### DIFF
--- a/easybuild/easyblocks/generic/cargo.py
+++ b/easybuild/easyblocks/generic/cargo.py
@@ -27,6 +27,7 @@ EasyBuild support for installing Cargo packages (Rust lang package system)
 
 @author: Mikael Oehman (Chalmers University of Technology)
 @author: Alex Domingo (Vrije Universiteit Brussel)
+@author: Alexander Grund (TU Dresden)
 """
 
 import os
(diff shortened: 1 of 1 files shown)


Micket:
Going with the combined PR

PR #3510, opened 2024-11-18
Add support for STAR-CCM+ installs with .aol installer

WilleBell:


easybuild/easyblocks/s/star_ccm.py
```diff
@@ -69,15 +94,33 @@ def install_step(self):
...
+                'HOME=%s' % self.builddir,
+                install_script,
+                "-i silent",
+                # for some reason the installation directory's name cannot be the version
+                # So using builddir and then moving to installdir
+                "-DINSTALLDIR=%s" % self.builddir,
+                "-DINSTALL_LICENSING=false",
+                "-DADDSYSTEMPATH=false",
+                self.cfg['installopts'],
+                "&& mv %s/%s %s" % (self.builddir, self.version, self.installdir),
```
smoors, line 112: for the `-r8` variant, suffix `-R8` is added to this directory. simplest solution is to add a wildcard:
```suggestion
                "&& mv %s/%s* %s" % (self.builddir, self.version, self.installdir),
```

easybuild/easyblocks/s/star_ccm.py
```diff
@@ -46,6 +47,26 @@ def __init__(self, *args, **kwargs):
         self.starccm_subdir = None
         self.starview_subdir = None
 
+    # adding an extract_step to support Siemens distributions with extension .aol
+    def extract_step(self):
+        # Siemens distributions are tarballs or executables with .aol extension
+        if self.src[0]['name'].endswith('.aol'):
+            self.aol_install = True
```
appolloford, line 54: The source file we got from users are still tarball file. Maybe this variable can be set by version instead of extension name?

WilleBell: This should still work for tarballs. It doesn't for you? I'm not sure how Siemens is planning to distribute this (I'm not even sure how they do it now), the idea is that this supports both tarballs and .aol

appolloford: I tested with my STAR-CCM easyconfig: (https://github.com/easybuilders/easybuild-easyconfigs/pull/21911) and it looks trying to look for .sh file
```
```

MaximeVdB: Just adding that this didn't work for me either (STAR-CCM+/19.06.009 from a source tarball containing a .aol installer).

WilleBell: @MaximeVdB The source you have is a tarball containing (only) a .aol file? (I didn't come across that type of source file)
@appolloford Is that the case for your tarball too?
The changes in this PR should only have effect if the source is .aol, so they must have changed their tarballs
Unfortunatelly, I only have a tarball for 18.06.007 to test and it works for that one.

appolloford: Here are the content of my tarball
```
$ ls starccm+_19.06.008/
STAR-CCM+2410_008_linux-x86_64-2.28_gnu11.4.aol  doc  fnp_LicAdmin.pdf
```

WilleBell: So if you extract the tarball manually the installation works?

appolloford: I tried to make another tarball having only `.aol` installer in it as following:
```
$ ls starccm+_19.06.008/
STAR-CCM+2410_008_linux-x86_64-2.28_gnu11.4.aol
```
and it also tried to look for `.sh` file and failed for the same reason. The only way for me to use the easyblock is untar the tarball and have `.aol` in both easyconfig and source directory.

I recent got newer starccm tarball and the structure looks the same as what I got last time:
```
$ ls starccm+_20.02.007/
STAR-CCM+2502_007_linux-x86_64-2.28_clang17.0.aol  doc  fnp_LicAdmin.pdf
```

appolloford: No, but what I asked from the beginning was that it should support for both tarball with/without .aol file. Even if they change to use .aol installer, users still downloads and provides tarball. As the install_step also search the `install_script_pattern`, I don't think it is required to check the file extension in the beginning. I think using version as the criteria should work

boegel: For STAR-CCM+ 2410, I also got a tarball with a `.aol` file in it:
```
$ ls /tmp/build/STARCCM/2410_008/system-system-r8
starccm+_19.06.008
$ ls /tmp/build/STARCCM/2410_008/system-system-r8/starccm+_19.06.008/
doc  fnp_LicAdmin.pdf  STAR-CCM+2410_008_linux-x86_64-2.28_gnu11.4-r8.aol
```

So ideally EasyBuild still unpacks first, then checks for `.aol` file to determine what to do...

boegel:
I had to do some additional hacks to get STAR-CCM+ 2410_008 installed, even after manually extracting the `.aol` file...

see https://github.com/WilleBell/easybuild-easyblocks/pull/1

easybuild/easyblocks/s/star_ccm.py
WilleBell: STAR-CCM_distros
Siemens seem to distribute the .aol in tarballs but also as an unarchived .aol file. I added support for the first case too in commit [3d35657](https://github.com/easybuilders/easybuild-easyblocks/pull/3510/commits/3d356578cc20e246a5e262a289eb057f0d587e46)

smoors:
@WilleBell you have merge conflicts that should be resolved first

WilleBell:
Test report with a tar.gz source for `STAR-CCM+` v20.02.007 w/ r8: 
https://github.com/easybuilders/easybuild-easyconfigs/pull/22704#issuecomment-2976485512

easybuild/easyblocks/s/star_ccm.py
```diff
@@ -69,23 +106,54 @@ def install_step(self):
...
+                'HOME=%s' % self.builddir,
+                install_script,
+                "-i silent",
+                # for some reason the installation directory's name cannot be the version
+                # So using builddir and then moving to installdir
+                "-DINSTALLDIR=%s" % self.builddir,
+                "-DINSTALL_LICENSING=false",
+                "-DADDSYSTEMPATH=false",
+                self.cfg['installopts'],
+                "&& mv %s/%s* %s" % (self.builddir, self.version, self.installdir),
```
lexming: the easyblock already does the move in line 148, isn't it?
```
run_shell_cmd("mv " + os.path.join(self.builddir, entry) + ' ' + self.installdir)
```

I like that`run_shell_cmd` more than appending a `mv` to this installation command, so if this one is indeed redundant please remove.

WilleBell: Done in commit [5c4e318](https://github.com/easybuilders/easybuild-easyblocks/pull/3510/commits/5c4e318f20ff0de8ef7263a205f1f30ad11a0f82)

WilleBell:
Test report by @WilleBell
 * **SUCCESS** _STAR-CCM+-20.02.007-r8.eb_ 

Build succeeded for 1 out of 1 (total: 7 mins 30 secs) (1 easyconfigs in total)
node702.hydra.os - Linux Rocky Linux 9.6, x86_64, AMD EPYC 9535 64-Core Processor, Python 3.9.21
See https://gist.github.com/WilleBell/a32eb1483fd3eb4a48d221662dce5e5d for a full test report.

lexming (approved):
LGTM

lexming:
Merging, thanks @WilleBell !

PR #3516, opened 2024-11-24
update CUDA easyblock to add CUPTI and nvvm library directories to `$LIBRARY_PATH`

trz42:
When using `RPATH` linking (e.g., in EESSI) we need to ensure that all directories containing shared libraries must be listed in `$LIBRARY_PATH`. Otherwise the RPATH linker wrapper would not add these directories to the RPATH header, and thus the libraries cannot be resolved by the dynamic linker later.

Here, we add two directories to `LIBRARY_PATH` such that corresponding `prepend_path` commands are added to the module file for CUDA. These directories are known to the easyblock, but currently only added to `$LD_LIBRARY_PATH` which is filtered in EESSI.

To restrict the change to certain environments, it is only used when `LD_LIBRARY_PATH` is a filtered environment variable and `LIBRARY_PATH` is not.

### DIFF
--- a/easybuild/easyblocks/c/cuda.py
+++ b/easybuild/easyblocks/c/cuda.py
@@ -43,7 +43,7 @@ from easybuild.tools import LooseVersion
 from easybuild.easyblocks.generic.binary import Binary
 from easybuild.framework.easyconfig import CUSTOM
 from easybuild.tools.build_log import EasyBuildError
-from easybuild.tools.config import IGNORE
+from easybuild.tools.config import build_option, IGNORE
 from easybuild.tools.filetools import adjust_permissions, change_dir, copy_dir, expand_glob_paths
 from easybuild.tools.filetools import patch_perl_script_autoflush, remove_file, symlink, which, write_file
 from easybuild.tools.run import run_cmd, run_cmd_qa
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/c/cuda.py
```diff
@@ -356,10 +356,18 @@ def make_module_req_guess(self):
             lib_path.append(os.path.join('nvvm', 'lib64'))
             inc_path.append(os.path.join('nvvm', 'include'))
 
+        library_path = ['lib64', os.path.join('stubs', 'lib64')]
```
ocaisa, line 359: You don't want the `stubs` in the final RPATH, these need to come from the OS at runtime

ocaisa: Ah, sorry, this is explicitly filtered in framework (I am pretty sure...)

ocaisa: I think this can be much simpler than this, there's no need for the checking, you can just directly set
```
'LIBRARY_PATH': lib_path + [os.path.join('stubs', 'lib64')],
```
below. The stubs location is filtered out by the rpath wrappers at link time (which is correct).

ocaisa: The only difference between the link libraries and runtime libraries is the stubs location

trz42: If we can always add `extras/CUPTI/lib64` and `nvvm/lib64` to `LIBRARY_PATH`, then we can implement it as you suggested.

BTW, while it may not relevant anymore, `CUPTI` and `nvvm` lib dirs should only be added for CUDA 7 or newer. That's currently not the case in the PR, but would be following your suggestion.

ocaisa: The directories being added are just guesses, if they don't exist they won't be included in the final module file (I think...)

trz42: Implemented suggestions in https://github.com/easybuilders/easybuild-easyblocks/pull/3516/commits/5077104baf2e0c80d035ba65e5bca3e4d7ce30d1

ocaisa (approved):

ocaisa:
This looks like it does what it should when testing via an extended dry run:
```
[make_module_step method]
List of paths that would be searched and added to module file:

 $ACLOCAL_PATH: share/aclocal
 $CLASSPATH: *.jar
 $CMAKE_LIBRARY_PATH: lib64
 $CMAKE_PREFIX_PATH:
 $CPATH: include, extras/CUPTI/include, nvvm/include
 $GI_TYPELIB_PATH: lib/girepository-*, lib32/girepository-*, lib64/girepository-*
 $LD_LIBRARY_PATH: lib64, extras/CUPTI/lib64, nvvm/lib64
 $LIBRARY_PATH: lib64, extras/CUPTI/lib64, nvvm/lib64, stubs/lib64
 $MANPATH: man, share/man
 $PATH: bin, nvvm/bin
 $PKG_CONFIG_PATH: pkgconfig
 $XDG_DATA_DIRS: share
 ```

PR #3519, opened 2024-11-26
force use of `bash` for `Allwmake` scripts in OpenFOAM easyblock

### LINKED ISSUE #3518: OpenFOAM v2406 fails to build the plugins on systems with `/bin/sh -> dash`
@lhcamilo reported on Slack that his OpenFOAM v2406 build failed in the sanity check, because of some missing binaries that should be part of the plugins:
```
  >> file 'OpenFOAM-v2406/platforms/linuxARM64GccDPInt32Opt/bin/cartesianMesh' found: FAILED
```
I reproduced the issue in the EESSI environment, see https://github.com/EESSI/software-layer/pull/826. Both have in common that they use Apptainer and a Debian-based distro for the builds.

The build step basically does `Allwmake && Allwmake-plugins`, but in our case the second step somehow seems to rerun the first one, as the easybuild log also shows:
```
$ grep "Finished compile of" easybuild-jix8_9wn.log

  Finished compile of OpenQBMM with OpenFOAM-v2406
  Finished compile of adiosFoam with OpenFOAM-v2406
  Finished compile of external-solver (petsc) with OpenFOAM-v2406
Finished compile of visualization with OpenFOAM-v2406
  Finished compile of OpenQBMM with OpenFOAM-v2406
  Finished compile of adiosFoam with OpenFOAM-v2406
  Finished compile of external-solver (petsc) with OpenFOAM-v2406
Finished compile of visualization with OpenFOAM-v2406
```

After adding some debugging statement to `Allwmake-plugins`, I found that [this line](https://develop.openfoam.com/Development/openfoam/-/blob/OpenFOAM-v2406/Allwmake-plugins?ref_type=tags#L10) seems to be the culprit, and it never reaches the bottom part of this script that actually builds the plugins. Running the command from that line interactively did work fine, though. In the end I figured that it may perhaps be related to the `/bin/sh` shebang in this script (it's also used in all the other `Allwmake` scripts); Debian-based distros usually have `/bin/sh -> dash`. I'm still not entirely sure why, but changing the shebang to `/bin/bash` does indeed solve the issue, and so does changing the symlink itself. It also solved the issue for @lhcamilo.

We could fix this in the OpenFOAM easyblock by changing the shebangs or calling the scripts with something like `bash Allwmake`.

bedroge:
Fixes #3518 by running all `Allwmake` scripts with `bash`. I'm currently running a test build, see https://github.com/EESSI/software-layer/pull/826#issuecomment-2501313733.

### DIFF
--- a/easybuild/easyblocks/o/openfoam.py
+++ b/easybuild/easyblocks/o/openfoam.py
@@ -372,7 +372,7 @@ class EB_OpenFOAM(EasyBlock):
                 if self.looseversion >= LooseVersion('2406'):
                     # Also build the plugins
                     cmd += ' && %s bash %s -log' % (self.cfg['prebuildopts'],
-                                               os.path.join(self.builddir, self.openfoamdir, 'Allwmake-plugins'))
+                                                    os.path.join(self.builddir, self.openfoamdir, 'Allwmake-plugins'))
 
             run_cmd(cmd_tmpl % cmd, log_all=True, simple=True, log_output=True)
 
--- a/easybuild/easyblocks/o/openfoam.py
+++ b/easybuild/easyblocks/o/openfoam.py
@@ -340,7 +340,7 @@ class EB_OpenFOAM(EasyBlock):
             cleancmd = "wcleanAll"
 
         # make directly in install directory
-        cmd_tmpl = "%(precmd)s && %(cleancmd)s && %(prebuildopts)s %(makecmd)s" % {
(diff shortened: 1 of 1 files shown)


ocaisa (approved):
I like this solution as it also makes sure to use the `bash` in the current path

bedroge:
> I like this solution as it also makes sure to use the `bash` in the current path

True, and I initially even suspected that this was the issue for EESSI. But other people reported the same issue on Slack, so apparently it's really related to dash being used.

ocaisa (approved):
LGTM

PR #3521, opened 2024-11-28
Print a warning when mixing `$EBPYTHONPREFIXES` and `$PYTHONPATH` modules

Flamefire:
$PYTHONPATH takes precedence which might lead to unexpected or wrong packages being used especially when a dependency includes a newer package than another dependency.
As it is very hard to find this as the cause print a warning when this situation is found.

I ran into this while building `scipy` of the `SciPy-bundle` which includes patches that require a newer `Cython` as the build dependency then what is available in the `Python-bundle-PyPI` module. As only Python-bundle used `$PYTHONPATH` the package from the `Cython` module was ignored even though `which cython` points to the right binary.

The error wasn't helpful:
```
Error compiling Cython file:
------------------------------------------------------------
...
    cdef __Pyx_memviewslice memslice
    cdef Py_ssize_t itemsize
    cdef bint dtype_signed
    cdef char kind
    itemsize = -1
    cdef bint ___pyx_const int8_is_signed
                          ^
------------------------------------------------------------

(tree fragment):16:27: Syntax error in C variable declaration
```

The least we can do is to detect such issues and warn the user.

The check is triggered for direct dependencies when both `$PYTHONPATH` and `$EBPYTHONPREFIXES` are set and the Python module is loaded.
I only check for modules that have a `lib/python*/site-packages` folder and specifically exclude the Python module which sets `$PYTHONPATH` for the `site-customize.py`   
This should catch most false positives while still catching enough real issues

### DIFF
--- a/easybuild/easyblocks/generic/pythonpackage.py
+++ b/easybuild/easyblocks/generic/pythonpackage.py
@@ -47,7 +47,7 @@ from easybuild.framework.easyconfig import CUSTOM
 from easybuild.framework.easyconfig.default import DEFAULT_CONFIG
 from easybuild.framework.easyconfig.templates import PYPI_SOURCE
 from easybuild.framework.extensioneasyblock import ExtensionEasyBlock
-from easybuild.tools.build_log import EasyBuildError, print_msg
+from easybuild.tools.build_log import EasyBuildError, print_msg, print_warning
 from easybuild.tools.config import build_option, PYTHONPATH, EBPYTHONPREFIXES
 from easybuild.tools.filetools import change_dir, mkdir, read_file, remove_dir, symlink, which, write_file
 from easybuild.tools.modules import ModEnvVarType, get_software_root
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/generic/pythonpackage.py
```diff
@@ -508,6 +508,41 @@ def set_pylibdirs(self):
...
+                root = os.path.realpath(get_software_root(dependency['name']))
+            except OSError:
+                return False
+            # Search for the path or a subdirectory of it
+            return any(path == root or path.startswith(root + os.sep) for path in pythonpaths)
+
+        conflicts = [dep['name'] for dep in self.cfg.dependencies() if is_in_pythonpaths(dep)]
+        if conflicts:
+            print_warning('The following modules set $PYTHONPATH but $%s is also used: %s\n'
+                          'Mixing such modules might cause unexpected results during the build of Python packages. '
```
boegel, line 542: Should we use `problems` instead of `results` here?

Flamefire: I wanted to be more generic because the result may not be a problem but still not what is expected or even intended

Micket (dismissed):
looks fine to me, i'll upload a test report

Micket:
Hm, i wanted to test this in conjunction with --prefer-python-search-path, it would help me if this targeted 5.0.x instead for testing.

Flamefire:
Rebased and retargeted to 5.0x

boegel:
@Flamefire I changed to target branch in this PR from `5.0.x` to `develop`, you should synchronize your PR branch with current `develop` branch (which has received a massive update after the release of EasyBuild v5.0.0, see https://github.com/easybuilders/easybuild-easyblocks/pull/3670)

PR #3522, opened 2024-11-29
WRF 4.6.1 WPS 4.6.0: Pass NETCDF-Fortran path

stefan-wolfsheimer:
Pass the NETCDF-Fortran path to configure scripts of WRF and WPS if the module is loaded.

### DIFF
--- a/easybuild/easyblocks/w/wps.py
+++ b/easybuild/easyblocks/w/wps.py
@@ -106,6 +106,9 @@ class EB_WPS(EasyBlock):
             wrfdir = os.path.join(wrf, det_wrf_subdir(get_software_version('WRF')))
         else:
             raise EasyBuildError("WRF module not loaded?")
+        netcdf_fortran = get_software_root('NETCDFMINFORTRAN')
+        if netcdf_fortran:
+            env.setvar('NETCDFF_DIR', netcdf_fortran)
 
         self.compile_script = 'compile'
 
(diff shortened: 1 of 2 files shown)


easybuild/easyblocks/w/wps.py
```diff
@@ -106,6 +106,9 @@ def configure_step(self):
             wrfdir = os.path.join(wrf, det_wrf_subdir(get_software_version('WRF')))
         else:
             raise EasyBuildError("WRF module not loaded?")
+        netcdf_fortran = get_software_root('NETCDFMINFORTRAN')
```
ocaisa, line 109: From what I have seen recently, the correct flags for this are
```
preconfigopts = "export NETCDF=$EBROOTNETCDFMINFORTRAN && export NETCDF_C=$EBROOTNETCDF &&"
```

ocaisa: You would also need to version check these settings (they are only valid for 4.6.something onwards)

ocaisa: Oops, WPS, this may be only valid for WRF

stefan-wolfsheimer: I removed the change for WRF, but WPS still requires it.

boegel:
EB_ARGS="WPS-4.4-foss-2022a-dmpar.eb"

boegel (approved):
lgtm

PR #3536, opened 2024-12-18
Add support for detecting external MAGMA library for ELSI

ahkole:
Adds an if branch to the ELSI easyblock to automatically detect the external MAGMA solver if it is installed on the system, similarly as is done for the other solvers that ELSI can link to.

For the flags that have to be set for MAGMA see https://gitlab.com/elsi_project/elsi_interface/-/blob/master/INSTALL.md?ref_type=heads .

A similar thing can probably be added for other solvers that are still missing from the easyblock like eigenexa, but I haven't tried/tested yet if that would work.


boegel (approved):
lgtm

boegel:
EB_ARGS="--installpath /tmp/$USER/pr3536 ELSI-2.11.0-foss-2023a-PEXSI.eb"

PR #3538, opened 2024-12-19
add custom easyconfig parameter `with_cxx_backtrace` to GCC easyblock to enable the build of the libstdcxx-backtrace library

maxim-masterov:
Add an option to build a static library for the stack trace support in C++23.

### DIFF
--- a/easybuild/easyblocks/g/gcc.py
+++ b/easybuild/easyblocks/g/gcc.py
@@ -156,7 +156,7 @@ class EB_GCC(ConfigureMake):
             'withppl': [False, "Build GCC with PPL support", CUSTOM],
             'withnvptx': [False, "Build GCC with NVPTX offload support", CUSTOM],
             'withamdgcn': [False, "Build GCC with AMD GCN offload support", CUSTOM],
-            'with_cxx_backtrace': [False, "Build GCC with enabled stack trace for C++23", CUSTOM],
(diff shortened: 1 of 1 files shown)


Thyre:
EB_ARGS="GCCcore-13.3.0.eb GCCcore-14.2.0.eb GCCcore-14.3.0.eb GCCcore-15.2.0.eb --installpath=/tmp/$USER/ebpr-3538 --try-amend=with_cxx_backtrace=True"

easybuild/easyblocks/g/gcc.py
```diff
@@ -643,6 +644,10 @@ def configure_step(self):
         else:
             self.configopts += " --disable-lto"
 
+        # enable builind of libstdc++_libbacktrace
+        if self.cfg['with_cxx_backtrace']:
+            self.configopts += " --enable-libstdcxx-backtrace"
+
```
Thyre: With the current approach, having `with_cxx_backtrace=False` by default, we should probably also disable support via a flag for GCC versions supporting backtrace.

We could go the route and define `with_cxx_backtrace` as `None` by default, passing no flag and just use what GCC has by default. Users could then pass `True` to explicitly enable support, and `False` to explicitly disable support.

boegel: +1 on this

@maxim-masterov Are you up for updating the PR accordingly?

maxim-masterov: done

PR #3548, opened 2025-01-10
Enhance Score-P EasyBlock for future releases and better oneAPI support.

Thyre:
The following changes are implemented:
- Properly handle Qt5 & Qt6 dependencies for CubeGUI, not only Qt4
- Remove `--with-cube` option, as this is not supported by Score-P 6.0+ & recent Scalasca versions
- Add dependency handling for GOTCHA, as this will likely be added with Score-P 9.0+
- Switch to `--with-nocross-compiler-suite=oneapi` if Intel toolchain uses oneAPI compilers (checked via 'oneapi' toolchain option). This allows building Score-P with oneAPI 2024.0.0 and newer.

Targeting `5.0.x`, as `develop` still contains the EasyConfigs [Cube-4.3.4-foss-2016a.eb](https://github.com/easybuilders/easybuild-easyconfigs/blob/254236fd3ea7c9d513b9bdaa2adf83087255c089/easybuild/easyconfigs/c/Cube/Cube-4.3.4-foss-2016a.eb#L1) and [Scalasca-2.3-foss-2016a.eb](https://github.com/easybuilders/easybuild-easyconfigs/blob/develop/easybuild/easyconfigs/s/Scalasca/Scalasca-2.3-foss-2016a.eb), still using `--with-cube=` as the option.

### DIFF
--- a/easybuild/easyblocks/s/score_p.py
+++ b/easybuild/easyblocks/s/score_p.py
@@ -24,13 +24,14 @@
 ##
 """
 EasyBuild support for software using the Score-P configuration style (e.g., Cube, OTF2, Scalasca, and Score-P),
-implemented as an easyblock.
+implemented as an EasyBlock.
 
 @author: Kenneth Hoste (Ghent University)
 @author: Bernd Mohr (Juelich Supercomputing Centre)
 @author: Markus Geimer (Juelich Supercomputing Centre)
 @author: Alexander Grund (TU Dresden)
 @author: Christian Feld (Juelich Supercomputing Centre)
+@author: Jan Andre Reuter (Juelich Supercomputing Centre)
 """
 import os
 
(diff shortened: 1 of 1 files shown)


Thyre:
I can split this PR into two, removing the `--with-cube` option in a separate one. With this, the remaining changes could also be used in EasyBuild 4.9.x.
However, I'm not sure if this is worth the effort, given that `5.0` is likely the next release.

Thyre:
See https://github.com/easybuilders/easybuild-easyconfigs/pull/22143 for a local test with Intel oneAPI EasyConfigs.

Thyre:
EB_BRANCH=5.0.x
EB_ARGS="Score-P-8.4-gompi-2024a-CUDA-12.6.0.eb Score-P-8.4-gompi-2024a.eb CubeGUI-4.8-GCCcore-11.3.0.eb Scalasca-2.6.1-gompi-2022a.eb --installpath /tmp/$USER"

Thyre:
Some comments that came up with offline discussion with @geimer:

- At least since Score-P v8.0 ee generally want `libbfd.so` instead of `libbfd.a`. The `.a` should still work I think, but may have side-effects we ran into while testing on some systems.
- We may want to rework how we pass dependencies to the tools using these EasyBlocks, since we may pass too many (unneeded) `configure` arguments. Instead of going through the list of defined dependencies of the EasyBlock, we should go through the `dependencies` and `builddependencies` passed by the EasyConfigs.

Both could be done in a separate PR to clean things up.

SebastianAchilles:
EB_BRANCH=5.0.x
EB_ARGS="Score-P-8.4-gompi-2024a-CUDA-12.6.0.eb Score-P-8.4-gompi-2024a.eb CubeGUI-4.8-GCCcore-11.3.0.eb Scalasca-2.6.1-gompi-2022a.eb --installpath /tmp/$USER"

SebastianAchilles (approved):
lgtm

SebastianAchilles:
Going in, thanks @Thyre!

PR #3549, opened 2025-01-13
Make `CmdCp` easier to use

Flamefire:
The name suggest running a command and copy files. However the commands need to be specified as a regex-map list which might be executed for every source instead of once. Often a workflow might be just `./custom_configure && ./custom_build.sh` which is not easy to map to this.
Instead allow non-tuples which are considered as commands to be run. When mixing tuples and non-tuples the plain commands are run last.

This was always a concern of mine that a workflow consisting of running only a command that involves some custom install procedure was less obvious than it might be. It came up [recently](https://github.com/easybuilders/easybuild-framework/pull/4580#issuecomment-2585204615) where an EC uses `MakeCp` requiring `parallel = False` as it replaces the `make` cmd by a single command. In that case it is `./configure --options` and `./coconut` which isn't easily applicable using `CmdCp`

With this change it can be:
```
cmds_map = [
    "./configure --option1 --option2",
    "./coconut --flag 1 --flag 2",
]
```

I'd say this is easier to read and understand than:
```
cmds_map = [
   (".*", "./configure --option1 --option2 && ./coconut --flag 1 --flag 2"),
]
```

Also error reporting would be more focused, i.e. either "./configure failed" or "./coconut failed"

I can port a couple easyconfigs to the new syntax in a PR to the easyconfig repo for test reports and usage examples.

### DIFF
--- a/easybuild/easyblocks/generic/cmdcp.py
+++ b/easybuild/easyblocks/generic/cmdcp.py
@@ -48,7 +48,11 @@ class CmdCp(MakeCp):
         extra_vars = MakeCp.extra_options(extra_vars=extra_vars)
         extra_vars['cmds_map'] = [
             [('.*', "$CC $CFLAGS %(source)s -o %(target)s")],
-            "List of regex/template command (with 'source'/'target' fields) tuples",
+            "List of commands to run. ",
+            "Entries are tuples of a regex to a command (may use 'source'&'target' templates) "
+            "to be executed on sources matching the regex.\n",
+            "Any element that is not a tuple, i.e. a string, "
+            "is a command to be executed after all tuple entries are handled.",
             CUSTOM,
         ]
         return extra_vars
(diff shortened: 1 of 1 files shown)


bartoldeman (dismissed):
LGTM

I've always found the syntax for CmdCp a little odd as well.
Another confusing bit is that `%(source)s` refers to the original unpatched source (i.e. it takes files from the source directory, not the unpacked archive), if you apply `patches` they are ignored. I found that out the hard way with
https://github.com/ComputeCanada/easybuild-easyconfigs/blob/computecanada-main/easybuild/easyconfigs/s/slurm-completion/slurm-completion-23.02.7.eb

bartoldeman:
A very long time ago, CmdCp had a `cmd` parameter:
https://github.com/easybuilders/easybuild-easyblocks/commit/ffe30c3468afeed01b4069f3dc8d2e4b464c10b6
I think perhaps it would be better to reintroduce `cmd` for simple string commands but use `cmds_map` for the more complex case? @boegel may know as author of the above commit?

Flamefire:
The problem is the default for `cmds_map`.
We could add a `cmds` parameter to work in the same way as here that, when set, removes the default for `cmds_map` but that makes the default disappear from the docstring. Can add it to the help though.

> Another confusing bit is that `%(source)s` refers to the original unpatched source 

Currently it is the file stem (name w/o extension) of the source, so not sure how it could refer to the "unpatched source". Maybe that has changed? Or I misunderstand the issue.

boegel:
@Flamefire I changed to target branch in this PR from `5.0.x` to `develop`, you should synchronize your PR branch with current `develop` branch (which has received a massive update after the release of EasyBuild v5.0.0, see https://github.com/easybuilders/easybuild-easyblocks/pull/3670)

Issue #3527, opened 2024-12-11
replace `make_module_req_guess` with `module_load_environment` in all EasyBlocks


lexming:
Tracking issue for the update of all easyblocks relying on `make_module_req_guess` to work with the changes in framework: https://github.com/easybuilders/easybuild-framework/pull/4653 and https://github.com/easybuilders/easybuild-framework/pull/4655

* [x] `a/abaqus.py`: #3578
* [x] `a/advisor.py`: #3551
* [x] `a/anaconda.py`:
  * [x] #3577
  * [x] #3585
* [x] `a/ansys.py`: #3579
* [x] `a/aocc.py`: #3594
* [x] `c/clang.py`: #3586
* [x] `c/cudacompat.py`: #3598 
* [x] `c/cuda.py`: #3587
* [x] `e/eigen.py`: #3599 
* [x] `f/flexiblas.py`:
  * [x] #3529 
  * [x] #3584
* [x] `f/fluent.py`: #3600
* [x] `f/freesurfer.py`: #3601
* [x] `f/freetype.py`: #3604
* [x] `f/fsl.py`: #3606 
* [x] `generic/bundle.py`: #3513
* [x] `generic/conda.py`:
  * [x] #3562 
  * [x] #3585
* [x] `generic/intelbase.py`:
  * [x] #3551
  * [x] #3584
* [x] `generic/perlmodule.py`: #3529
* [x] `generic/pythonpackage.py`: #3530
* [x] `generic/rpm.py`: #3561
* [x] `generic/systemcompiler.py`: #3559 
* [x] `generic/systemmpi.py`: #3560 
* [x] `g/gcc.py`:
  * [x] #3556
  * [x] #3584
* [x] `g/gromacs.py`: #3607 
* [x] `h/hdf5.py`: #3608 
* [x] `i/iccifort.py`:
  * [x] #3550
  * [x] #3584
* [x] `i/icc.py`:
  * [x] #3550
  * [x] #3584
* [x] `i/ifort.py`:
  * [x] #3550
  * [x] #3584
* [x] `i/imkl_fftw.py`: #3612
* [x] `i/imkl.py`:
  * [x] #3555
  * [x] #3584
* [x] `i/impi.py`:
  * [x] #3553
  * [x] #3584
* [x] `i/inspector.py`: #3551
* [x] `i/intel_compilers.py`:
  * [x] #3552
  * [x] #3584
* [x] `i/itac.py`: #3609 
[...]

PR #3556, opened 2025-01-14
adopt `module_load_environment`: GCC

lexming:
Update of GCC easyblock for https://github.com/easybuilders/easybuild-easyblocks/issues/3527

Note: `PATH` and `LD_LIBRARY_PATH` are default and could be removed

### DIFF
--- a/easybuild/easyblocks/g/gcc.py
+++ b/easybuild/easyblocks/g/gcc.py
@@ -203,6 +203,13 @@ class EB_GCC(ConfigureMake):
             self.log.warning('Setting withnvptx to False, since we are building on a RISC-V system')
             self.cfg['withnvptx'] = False
 
+        # define list of subdirectories in search paths of module load environment
+        self.module_load_environment.PATH = ['bin']
+        self.module_load_environment.LD_LIBRARY_PATH = ['lib', 'lib64']
+        # GCC can find its own headers and libraries but the .so's need to be in LD_LIBRARY_PATH
+        self.module_load_environment.CPATH = []
+        self.module_load_environment.LIBRARY_PATH = ['lib', 'lib64'] if get_cpu_family() == RISCV else []
+
     def create_dir(self, dirname):
         """
         Create a dir to build in.
@@ -1194,17 +1201,3 @@ class EB_GCC(ConfigureMake):
 
         super(EB_GCC, self).sanity_check_step(custom_paths=custom_paths, custom_commands=custom_commands,
                                               extra_modules=extra_modules)
-
-    def make_module_req_guess(self):
-        """
-        GCC can find its own headers and libraries but the .so's need to be in LD_LIBRARY_PATH
-        """
-        guesses = super(EB_GCC, self).make_module_req_guess()
-        guesses.update({
-            'PATH': ['bin'],
-            'CPATH': [],
-            'LIBRARY_PATH': ['lib', 'lib64'] if get_cpu_family() == RISCV else [],
-            'LD_LIBRARY_PATH': ['lib', 'lib64'],
-            'MANPATH': ['man', 'share/man']
-        })
-        return guesses


boegel (approved):
lgtm

boegel:
EB_ARGS="--installpath /tmp/$USER/pr3556 GCCcore-10.2.0.eb GCCcore-11.3.0.eb GCCcore-12.3.0.eb GCCcore-13.3.0.eb GCCcore-14.2.0.eb"
CORE_CNT=16
EB_BRANCH=5.0.x

easybuild/easyblocks/g/gcc.py
```diff
@@ -203,6 +203,13 @@ def __init__(self, *args, **kwargs):
             self.log.warning('Setting withnvptx to False, since we are building on a RISC-V system')
             self.cfg['withnvptx'] = False
 
+        # define list of subdirectories in search paths of module load environment
+        self.module_load_environment.PATH = ['bin']
+        self.module_load_environment.LD_LIBRARY_PATH = ['lib', 'lib64']
```
boegel, line 208: These two lines don't need to be here since framework takes care of this exactly like this

boegel (changes requested):

easybuild/easyblocks/g/gcc.py
```diff
@@ -203,6 +203,13 @@ def __init__(self, *args, **kwargs):
...
+        # GCC can find its own headers and libraries but the .so's need to be in LD_LIBRARY_PATH
+        self.module_load_environment.CPATH = []
+        self.module_load_environment.LIBRARY_PATH = ['lib', 'lib64'] if get_cpu_family() == RISCV else []
```
boegel, line 211: The RISC-V exception was added here in #3256, we should check the impact of not including `lib`/`lib64` on a RISC-V system.

@bedroge Was this really necessary, or is it really only required during the build of GCC itself?

easybuild/easyblocks/g/gcc.py
```diff
@@ -203,6 +203,13 @@ def __init__(self, *args, **kwargs):
             self.log.warning('Setting withnvptx to False, since we are building on a RISC-V system')
             self.cfg['withnvptx'] = False
 
+        # define list of subdirectories in search paths of module load environment
+        self.module_load_environment.PATH = ['bin']
+        self.module_load_environment.LD_LIBRARY_PATH = ['lib', 'lib64']
+        # GCC can find its own headers and libraries but the .so's need to be in LD_LIBRARY_PATH
+        self.module_load_environment.CPATH = []
```
boegel, line 210: Rather than setting `CPATH` explicitly, this should be changed to using the utility functions being added in:
* https://github.com/easybuilders/easybuild-framework/pull/4655

boegel (changes requested):

boegel:
Test report by @boegel
 * **SUCCESS** _GCCcore-11.2.0.eb_ 
 * **SUCCESS** _GCC-12.2.0.eb_ 
 * **SUCCESS** _GCCcore-13.2.0.eb_ 

node3561.doduo.os - Linux RHEL 8.8, x86_64, AMD EPYC 7552 48-Core Processor (zen2), Python 3.6.8
See https://gist.github.com/boegel/41c067f3b160ea2644ec3e906b94f5e6 for a full test report.

boegel (approved):
lgtm

boegel:
I'll leave this as-is, since it gets rid of using the deprecated `make_module_req_guess`.

We'll need another pass over this after https://github.com/easybuilders/easybuild-framework/pull/4655 got merged since we're hard setting `$CPATH`...

easybuild/easyblocks/g/gcc.py
bedroge: Not 100% sure, but I think the final installation wouldn't work without this either.

easybuild/easyblocks/g/gcc.py
lexming: fixed in https://github.com/easybuilders/easybuild-easyblocks/pull/3584/files/04dc2b383f7116d218e7fe984ae47cd7ab063eed..9926576e2256d421d0a0cc768c2d6ca85d6b7119

easybuild/easyblocks/g/gcc.py
lexming: IMO leaving those paths in `LIBRARY_PATH` should not cause any harm beyond having two extra dirs in that environment variable. So I updated this easyblock to remove `CPATH` but leave `LIBRARY_PATH` with default paths: https://github.com/easybuilders/easybuild-easyblocks/pull/3584/files/04dc2b383f7116d218e7fe984ae47cd7ab063eed..9926576e2256d421d0a0cc768c2d6ca85d6b7119

easybuild/easyblocks/g/gcc.py
lexming: fixed in https://github.com/easybuilders/easybuild-easyblocks/pull/3584/files/04dc2b383f7116d218e7fe984ae47cd7ab063eed..9926576e2256d421d0a0cc768c2d6ca85d6b7119

PR #3559, opened 2025-01-17
split `SystemCompilerGCC` out of `SystemCompiler` and adopt `module_load_environment`

lexming:
This PR fixes three issues with `SystemCompiler`:

1. fixes #2815

   It is not possible to have a single class that handles GCC and Intel compilers, because inheritance in python is linear, so as soon as multiple classes are added to `SystemCompiler` all of them come up in the chain of calls generated by `super()`.
   Solution is to make `SystemCompiler` agnostic of any compiler family and create child classes per compiler family. This PR adds `SystemCompilerGCC` for the GCC family. I'm not adding any for `icc/ifort` as those easyconfigs have been already archived. In the future we could add one for intel-compilers if there is interest.
   This approach is quite simple as the main logic of handling the multiple families remains inside `SystemCompiler`, avoiding any duplication of code. Child classes only need to set `self.compiler_class` appropriately.

   This PR is an alternative solution to #3402. The present solution avoids changes in other easyblocks like `EB_GCC` and is more future-proof as sub-classes handling different compiler families are completely isolated.

2. fixes #3558

   RPATH is forced disable in `SystemCompilers` as it has no role for this type of installation.

3. updates `SystemCompiler` easyblock for https://github.com/easybuilders/easybuild-easyblocks/issues/3527

PR #3560, opened 2025-01-17
 adopt module_load_environment: SystemMPI

lexming:
updates `SystemMPI` easyblock for https://github.com/easybuilders/easybuild-easyblocks/issues/3527


boegel:
@lexming I'm not sure how useful this is, or even if it's actively used by anyone, but I would like to keep this one around (unless it's an absolute nightmare to port it over to `module_load_environment`...

lexming:
@boegel updated using the same approach as for `SystemCompiler` #3559

boegel (approved):
lgtm

PR #3569, opened 2025-01-28
add support for Sapphire Rapids in LAMMPS easyblock

bedroge:
I used the same approach as for Icelake, support for Sapphire Rapids was also added in version 2Aug2023, see https://github.com/lammps/lammps/blob/stable_2Aug2023/lib/kokkos/cmake/kokkos_arch.cmake#L62.

Do note that the 2Aug2023 has a dependency on a too old archspec version which detects Sapphire Rapids as Icelake, so we may want to have another fix for that (e.g. modify the archspec version in the easyconfig?).

### DIFF
--- a/easybuild/easyblocks/l/lammps.py
+++ b/easybuild/easyblocks/l/lammps.py
@@ -126,6 +126,7 @@ KOKKOS_CPU_MAPPING = {
     'skylake_avx512': 'SKX',
     'cascadelake': 'SKX',
     'icelake': 'SKX',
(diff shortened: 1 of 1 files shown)


ocaisa (approved):
Need an actual build on the arch to make sure this works so relying on the EESSI build as a test report

ocaisa:
Confirmation in https://github.com/EESSI/software-layer/pull/901#issuecomment-2621537481 that this works,

PR #3571, opened 2025-01-29
Restructure RPATH and elf interpreter setting for Java, and add the LIBRARY_PATH to the additional RPATH

casparvl:
This is the first step in fixing the issue observed [here](https://gitlab.com/eessi/support/-/issues/99) with Aladin-desktop that certain libraries aren't properly RPATH-ed, and as a result, the relevant X11 library cannot be found. The reason is that 

1. Java is a binary installation, so the compiler wrappers don't do any RPATH-ing. A previous PR already fixed that (OS) libraries from a custom sysroot would be added to the RPATH. However, that did not add anything from the regular dependencies to the RPATH. That omission is fixed in this PR.
2. Java is currently at the system toolchain level, so the X11 libraries are not even among the dependencies (even though the Java libraries depend on them). This will be addressed in a follow-up PR to `easybuild-easyconfigs`.

PR #3583, opened 2025-02-07
Restructure RPATH and elf interpreter setting for Java, and add the LIBRARY_PATH to the additional RPATH and separate java awt libs

casparvl:
Supersedes https://github.com/easybuilders/easybuild-easyblocks/pull/3571


lexming:
@casparvl all development is on 5.0 now, can you please move the target of this PR to the branch `5.0.x`?

boegel:
> @casparvl all development is on 5.0 now, can you please move the target of this PR to the branch `5.0.x`?

If this can wait a bit, there's no need to, soon the `5.0.x` branch will be collapsed into `develop`...

easybuild/easyblocks/generic/binary.py
```diff
@@ -65,6 +68,22 @@ def extra_options(extra_vars=None):
             'prepend_to_path': [PREPEND_TO_PATH_DEFAULT, "Prepend the given directories (relative to install-dir) to "
                                                          "the environment variable PATH in the module file. Default "
                                                          "is the install-dir itself.", CUSTOM],
+            # We should start moving away from skipping the RPATH sanity check and towards patching RPATHS
+            # using patchelf, see e.g. https://github.com/easybuilders/easybuild-easyblocks/pull/3571
+            # The option run_rpath_sanity_check supports a gradual transition where binary installs that properly
+            # patch the RPATH can start running the sanity check
+            'run_rpath_sanity_check': [False, "Whether or not to run the RPATH sanity check", CUSTOM],
```
casparvl, line 75: This used to be needed not to break any existing `rpath` sanity checking that's going on. But, that may no longer be the case: the _default_ RPATH sanity check has been relaxed https://docs.easybuild.io/easybuild-v5/changes/#rpath . So, unless people actively filter `LD_LIBRARY_PATH` in their Easybuild config, the sanity check wouldn't fail anyway.

easybuild/easyblocks/j/java.py
```diff
@@ -28,24 +28,50 @@
...
+                'libsplashscreen.so', 'libjsound.so']
+
+    @staticmethod
+    def extra_options(extra_vars=None):
+        """Extra easyconfig parameters specific to Java easyblock."""
+        extra_vars = PackedBinary.extra_options(extra_vars)
+        extra_vars.update({
+            # Overwrite the default value for run_rpath_sanity_check to True, since we have RPATH patching with
+            # patchelf for Java
+            'run_rpath_sanity_check': [True, "Whether or not to run the RPATH sanity check", CUSTOM],
```
casparvl, line 58: can (probably) be removed, see above.

laraPPr:
@casparvl do you have the time to sync this pr with develop. Or should we close it and start anew?

easybuild/easyblocks/generic/binary.py
```diff
@@ -131,8 +150,105 @@ def install_step(self):
...
+        # Then, add paths from sysroot to the extra RPATH
+        if add_sysroot_libdirs_to_rpath:
+            sysroot = build_option('sysroot')
+            sysroot_lib_paths = glob.glob(os.path.join(sysroot, 'lib*'))
+            sysroot_lib_paths += glob.glob(os.path.join(sysroot, 'usr', 'lib*'))
+            sysroot_lib_paths += glob.glob(os.path.join(sysroot, 'usr', 'lib*', 'gcc', '*', '*'))
+            if sysroot_lib_paths:
+                self.log.info("List of library paths in sysroot %s to add to RPATH section: %s", sysroot,
+                              sysroot_lib_paths)
+                extra_rpaths += sysroot_lib_paths
```
ocaisa, line 239: If you are changing the dynamic linker, this should not be necessary since these are locations that are searched by default.

ocaisa:
@casparvl Maybe we can spar on this? There are a couple of things that I don't think are being taken into account (and I was looking into them last week):

- `LIBRARY_PATH` is a good starting place, but there are things that need to be excluded (like `stubs` locations which are meant to be resolved at runtime)
- We explicitly want to include some locations in the RPATH (EasyBuild has a setting for this, and we use it to allow for our MPI injection approach)
  - This requires additional options to protect these when using `--shrink-rpath` (and we also need to think a bit about using this option to account for temporary locations that resolve libraries but do not exist after installation)
- `patchelf` can be unstable, you only want to call it once so you need to collect all the options

casparvl:
Yes, let's meet on this. Just a heads-up: this is a bit less pressing to me right now, so I don't think I'll be able to put a lot of time in _implementing_ a solution. But I know that it may be relevant for you from an EFP point of view. We can definitely brainstorm on this, and then maybe you can do the (re-)implementation of this idea - and I could review/test? Anyway, let's get in touch on chat to set a meeting,

PR #3605, opened 2025-02-17
Fix bug in FlexiBLAS easyblock to allow AOCL-BLAS to be default

lcniel:
In the current version, if one sets AOCL-BLAS as the default, it will be (silently) ignored and FlexiBLAS will use NETLIB instead, since the eventual name of the AOCL-BLAS backend will be the CMake name, AOCL_MT. This PR fixes this by substituting the name in `configopts['FLEXIBLAS_DEFAULT']`, just like how it's handled in the backends variable.


boegel (approved):

boegel:
EB_BRANCH=5.0.x
EB_ARGS="FlexiBLAS-3.4.4-GCC-13.3.0.eb --installpath /tmp/$USER/pr3605"

PR #3606, opened 2025-02-17
adopt `module_load_environment` in FSL easyblock

lexming:
Update of FSL easyblock for https://github.com/easybuilders/easybuild-easyblocks/issues/3527

### DIFF
--- a/easybuild/easyblocks/f/fsl.py
+++ b/easybuild/easyblocks/f/fsl.py
@@ -53,8 +53,8 @@ class EB_FSL(EasyBlock):
 
         self.fsldir = None
 
-        self.module_load_environment.PATH = ["fsl/bin"]
-        self.module_load_environment.LD_LIBRARY_PATH = ["fsl/lib"]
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/f/fsl.py
```diff
@@ -53,6 +53,9 @@ def __init__(self, *args, **kwargs):
 
         self.fsldir = None
 
+        self.module_load_environment.PATH = ["fsl/bin"]
+        self.module_load_environment.LD_LIBRARY_PATH = ["fsl/lib"]
```
boegel, line 57:
```suggestion
        self.module_load_environment.PATH = [os.path.join('fsl', 'bin')]
        self.module_load_environment.LD_LIBRARY_PATH = [os.path.join('fsl', 'lib')]

boegel (changes requested):

boegel (approved):
lgtm

PR #3623, opened 2025-02-19
Adapt cp2k regtest argument

stevenvdb:
This builds upon https://github.com/easybuilders/easybuild-easyblocks/pull/3433. For CP2K >= 2025.1, the `do_regtest.py` script expects a full path instead of just `self.typearch` as first argument (see https://github.com/cp2k/cp2k/pull/3640)


pavelToman:
There are also some chages to this block: https://github.com/easybuilders/easybuild-easyblocks/pull/3862

pavelToman:
@beeebiii could you look at the failing [REPORT](https://github.com/easybuilders/easybuild-easyblocks/pull/3623#issuecomment-3183607802)?
There is some problem with your changes to the cp2k.py block - failing for older version such CP2K-2023.1-foss-2023a.eb

PR #3651, opened 2025-03-04
update workflows to use Ubuntu 22.04

boegel:
Ubuntu 20.04 image for GitHub Actions is deprecated since 2025-02-01, see also https://github.com/actions/runner-images/issues/11101

### DIFF
--- a/.github/workflows/unit_tests.yml
+++ b/.github/workflows/unit_tests.yml
@@ -10,10 +10,10 @@ concurrency:
 
 jobs:
   build:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
     strategy:
       matrix:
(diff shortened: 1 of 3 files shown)


.github/workflows/unit_tests.yml
```diff
@@ -10,10 +10,10 @@ concurrency:
 
 jobs:
   build:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
     strategy:
       matrix:
-        python: [3.6, 3.7, 3.8, 3.9, '3.10', '3.11', '3.12']
```
boegel, line 16: Removing testing with Python 3.6 for now, since that's not supported in Ubuntu 22.04

Once the problems with the custom workflow being added in https://github.com/easybuilders/easybuild-framework/pull/4790 are figured out, we can restore testing with Python 3.6...

lexming (approved):
LGTM

PR #3653, opened 2025-03-05
Allow test case filtering and other unittest arguments

Flamefire:
The test suite always runs all tests without a way to filter. Use the unittest loader to handle all unittest arguments, especially `-k testpattern`


Flamefire:
Can this be merged @boegel ? Working on a test issue in a specific test but can only run all tests without this

Flamefire:
@Crivella Could you take a look at this? This would be really useful for developing and debugging specific tests

test/easyblocks/suite.py
```diff
@@ -46,35 +46,48 @@
...
-res = unittest.TextTestRunner().run(SUITE)
-
-fancylogger.logToFile(log_fn, enable=False)
-shutil.rmtree(os.environ['EASYBUILD_TMP_LOGDIR'])
-del os.environ['EASYBUILD_TMP_LOGDIR']
-
-if not res.wasSuccessful():
-    sys.stderr.write("ERROR: Not all tests were successful.\n")
-    print("Log available at %s" % log_fn)
-    sys.exit(2)
```
Crivella, line 76: Is there a reason that the behavior of `sys.exit(2)` is not kept in the changes?

I think without this the test command will return 1 on failure, not sure if this matters for anything

test/easyblocks/suite.py
```diff
@@ -46,35 +46,48 @@
...
+
+        try:
+            tmpdir = set_tmpdir(raise_error=True)
+        except EasyBuildError as err:
+            sys.stderr.write("No execution rights on temporary files, specify another location via $TMPDIR: %s\n" % err)
+            sys.exit(1)
+
+        os.environ['EASYBUILD_TMP_LOGDIR'] = tempfile.mkdtemp(prefix='easyblocks_test_')
+
+        res = super().run(*args, **kwargs)
```
Crivella, line 71: Would there be a difference in behavior between
`<TestSuite oject>.run()`
and
`<TextTestRunner object>.run(<TestSuite object>`
?

Flamefire: Not really. `unittest.main` uses `TextTestRunner` as the default test runner and calls `testrunner.run(suite)`. So it ends up being exactly the same

test/easyblocks/suite.py
Flamefire: It is so that unittest can to the reporting and handling based on the returned test result. But yes it might change the exit code from 2 to:

> By default main calls [sys.exit()](https://docs.python.org/3/library/sys.html#sys.exit) with an exit code indicating success (0) or failure (1) of the tests run. An exit code of 5 indicates that no tests were run or skipped.

I don't think it matters though

PR #3657, opened 2025-03-07
Explicitly disable libxml2 support in LLVM easyblock if libxml2 is not available as dependency

Crivella:
Attempt to fix https://github.com/easybuilders/easybuild-easyconfigs/issues/22491

Avoid using system `libxml2` which could cause inconsistent test result failures (or possibly install failures?)

### DIFF
--- a/easybuild/easyblocks/l/llvm.py
+++ b/easybuild/easyblocks/l/llvm.py
@@ -110,7 +110,8 @@ remove_gcc_dependency_opts = {
     'LIBUNWIND_USE_COMPILER_RT': 'On',
 
     # Libxml2 from system gets automatically detected and linked in bringing dependencies from stdc++, gcc_s, icuuc, etc
-    'LLVM_ENABLE_LIBXML2': 'Off',
+    # Moved to a check at the configure step. See https://github.com/easybuilders/easybuild-easyconfigs/issues/22491
+    # 'LLVM_ENABLE_LIBXML2': 'Off',
 
     'SANITIZER_USE_STATIC_LLVM_UNWINDER': 'On',
 }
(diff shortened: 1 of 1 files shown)


boegel:
EB_ARGS="LLVM-14.0.3-GCCcore-11.3.0.eb LLVM-14.0.6-GCCcore-12.3.0-llvmlite.eb LLVM-14.0.6-GCCcore-13.2.0-llvmlite.eb LLVM-15.0.5-GCCcore-12.2.0.eb LLVM-16.0.6-GCCcore-12.3.0.eb LLVM-16.0.6-GCCcore-13.2.0.eb LLVM-18.1.8-GCCcore-13.3.0-minimal.eb --installpath /tmp/$USER/pr3657"
CORE_CNT=16
EB_BRANCH=5.0.x

Thyre:
Hm, looks like this isn't enough to fix the tests:
See https://github.com/easybuilders/easybuild-easyconfigs/pull/22489#issuecomment-2708238005

boegel:
@Thyre What if you add `libxml2` as a direct dependency to LLVM, does that change anything?

Thyre:
> @Thyre What if you add `libxml2` as a direct dependency to LLVM, does that change anything?

I added a direct dependency, see https://github.com/easybuilders/easybuild-easyconfigs/pull/22489/commits/840f2b0015d0e1ae42977bfb9fee2dea9687bbe3. My notebook, running Zorin OS 17, still failed to pass the tests. Another system, with Arch Linux, was able to.

I'll try to debug this further...

Thyre:
`llvm-mt` was using the `libLLVM-14.so` from an incomplete installation in the `software` folder which failed before, not the `build` folder. This installation was missing enabling `libxml2` for the build.
I restarted the build process with this installation removed.

boegel:
@Thyre So, what's the conclusion here?

We need both this + add `libxml2` as direct dependency?

Thyre:
Oh sorry, forgot to add a comment with the results. We do not need to add `libxml2` from what I've gathered (see https://github.com/easybuilders/easybuild-easyconfigs/pull/22489#issuecomment-2708806218), though I haven't tested it with rpathing disabled. I'll start another test build with this real quick.

Thyre:
Same result with rpathing disabled, we do not need `libxml2` as `dependency` :smile:

Crivella:
> Edit: Yeah, the incomplete installation was causing these issues.

@Thyre Yeah also had similar problem when running with `--rebuild` as the RPATH to the install dir is being hardcoded during the build if the install dir is not deleted before it it will pick up symbols from there instead of the build directory

Not sure if this has been a problem also for other software packages build with rpath @boegel .
One thing we might try is to append instead of prepending the install dir rpath, but that would need some extensive testing to make sure we don't end up picking up wrong libraries in some dependecies combination?

boegel:
EB_ARGS="LLVM-14.0.6-GCCcore-12.3.0-llvmlite.eb LLVM-16.0.6-GCCcore-13.2.0.eb --installpath /tmp/$USER/pr3657"
CORE_CNT=16
EB_BRANCH=5.0.x

boegel (approved):

PR #3658, opened 2025-03-10
Fix llvm18 rpath

Crivella:
This is a Fix for the errors observer in 

- https://github.com/easybuilders/easybuild-easyconfigs/pull/21832#issuecomment-2701853276
- https://github.com/easybuilders/easybuild-easyconfigs/pull/21833#issuecomment-2701828795

Fixes https://github.com/easybuilders/easybuild-easyconfigs/issues/22509

Previously a fix was tested by setting `CFLAGS` and `CXXFLAGS` to add `-Wno-unused-command-line-argument` before running `make check-all` but this did not fix the problem as this was not being used (the CMake during the build step is responsible for this and probably in 18.1.X not all flags are properly set for the tests of the runtimes).

Manually fixing the Compiler path to the non RPATH-wrapped one in the `*.cfg.in` files for the runtime tests seems to fix this.

This leads to a few other errors in the offloading test suite which i believe can be ignored as offloading in 18.1.8 is still not mature (and this do not seem to appear in 19.X)

### DIFF
--- a/easybuild/easyblocks/l/llvm.py
+++ b/easybuild/easyblocks/l/llvm.py
@@ -909,12 +909,15 @@ class EB_LLVM(CMakeMake):
             # needs more digging into the CMake logic
             if build_option('rpath') and LooseVersion(self.version) < LooseVersion('19') and self.cfg['build_runtimes']:
                 self.log.warning("Removing rpath wrapping of compiler from test suite as it will result in errors")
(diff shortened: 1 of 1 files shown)


Crivella:
Closed in favor of adding patch also used form LLVM 19.1.7 for running the test suite

PR #3675, opened 2025-03-23
Fix build failures in updated LLVM EasyBlock

Thyre:
Fixes the following issues in the new LLVM EasyBlock:
- Fix build failure in multi-stage build introduced in https://github.com/easybuilders/easybuild-easyblocks/pull/3674 due to incorrect glob pattern
- Add string quotations for `LIBOMPTARGET_PLUGINS_TO_BUILD`, as the separator `;` for the plugins would be incorrectly picked up as ending the command. This causes CMake to use the incorrect source folder (and not pick up all passed flags), failing the build.

PR #3674, opened 2025-03-21
Fix LLVM build on non-x86

Flamefire:
(created using `eb --new-pr`)

The glob pattern introduced in https://github.com/easybuilders/easybuild-easyblocks/commit/7a9905360d64c3454d114a4dd3b18d5d4443f0eb only works for X86 where our arch constant is lowercase `x86`. 

However on ARM the folder is e.g. `lib/gcc/aarch64-linux-gnu/13` and on POWER it is `powerpc64le-*`

PR #3676, opened 2025-03-26
Fix LLVM build on RISCV

julianmorillo:


easybuild/easyblocks/l/llvm.py
```diff
@@ -999,7 +1001,8 @@ def post_processing_step(self):
     def get_runtime_lib_path(self, base_dir, fail_ok=True):
         """Return the path to the runtime libraries."""
         arch = get_arch_prefix()
-        glob_pattern = os.path.join(base_dir, 'lib', f'%s-{arch}')
+        # glob_pattern = os.path.join(base_dir, 'lib', f'%s-{arch}')
+        glob_pattern = os.path.join(base_dir, 'lib', f'{arch}-*')
```
ocaisa, line 1005: This doesn't look right, should this be arch-specific?

julianmorillo: It worked for me in RISC-V. 

I'm not sure what the original line does, but with the original code, `glob_pattern` was equal to `/tmp/eb/easy build/build/LLVM/20.1.1/GCCcore-13.2.0/llvm.obj.1/lib/%s-riscv64`. And with that, matches was null after `matches = glob.glob(glob_pattern)`, so the ERROR "Could not find runtime library directory" was raised.

On the other hand, with my modification, matches[0] is `/tmp/eb/easybuild/build/LLVM/20.1.1/GCCcore-13.2.0/llvm.obj.1/lib/riscv64-unknown-linux-gnu` and the function finally returns `lib/riscv64-unknown-linux-gnu`

How does it work in other architectures?

ocaisa: @Crivella would know better than me...

Crivella: There was an error introduced in https://github.com/easybuilders/easybuild-easyblocks/pull/3674 when replacing the old python2 formatting with f-strings.
This is also currently being fixed in https://github.com/easybuilders/easybuild-easyblocks/pull/3675, but it might take a bit before merging as there is also a lot of device testing going on there.

Regarding the location, during compilation LLVM puts the runtimes (specific for the host target) in $ROOT/lib/<target_triple>.
I am not aware of it ever compiling more than one set of runtimes (maybe if you are cross-compiling LLVM itself?), but that was the easiest way i had to find the runtime library directory

Crivella: @ocaisa 

On second thought, should the `get_arch_prefix` introduced in #3674 be a framework function? (maybe renamed to `get_compiler_arch_prefix`)

Crivella:
EB_ARGS="LLVM-18.1.8-GCCcore-13.3.0-minimal.eb --installpath /tmp/$USER/ebpr-3676"

ocaisa:
EB_ARGS="LLVM-18.1.8-GCCcore-13.3.0-minimal.eb --installpath /tmp/$USER/ebpr-3676"

Crivella:
As a note on the test, i do not think we have a test runner to properly test the arch string, but I do not see any particular exceptions looking at

https://github.com/llvm/llvm-project/blob/120e940356561035bf37bc4fcb6fab771e7a9b7d/llvm/cmake/config.guess#L1006-L1029

Maybe we can also add the string for 32bit systems, but this might be as discussed better moved to a framework implementation.

@julianmorillo Just in case you have the [easybuild github integration](https://docs.easybuild.io/integration-with-github/) installed would you mind running a minimal build uploading the report? EG `eb LLVM-18.1.8-GCCcore-13.3.0-minimal.eb --include-easyblocks-from-pr 3676 --upload-test-report`

Thyre:
We need to test a non-minimal build for this, as the arch prefix is only relevant if runtimes are built (or tests in a multi-stage build). 

The new LLVM-EasyConfig PRs would be ideal for this.

Crivella:
Agree the test on RISCV should be on a full build uploaded the message and also retyping here:

For the RISCV test `eb --from-pr 22517 --include-easyblocks-from-pr 3676 --upload-test-report`

I tested the minimal as `self._set_gcc_prefix()` is still called during the configure step, but it would be ideal to test runtime builds on RISCV

Crivella:
Now also considering the changes in #3706 I am starting to wonder if it would be better to have `get_arch_prefix` at the framework level in general, or for every toolchain, or possibly at the easyblock level ...  🤔

jfgrimm:
@Crivella perhaps throw it in framework's `systemtools`? 
https://github.com/easybuilders/easybuild-framework/blob/develop/easybuild/tools/systemtools.py

boegel:
This looks really trivial now...

@julianmorillo Is there really all that's necessary to fix the installation of LLVM on RISC-V?

julianmorillo:
`eb --optarch=GENERIC -r LLVM-20.1.5-GCCcore-13.2.0.eb --from-commit a0e254bc8a04d20dcd97ded5e339214628074ff3 --include-easyblocks-from-pr=3676`

Gives these results:

```
Testing Time: 11664.78s

Total Discovered Tests: 139851
  Skipped          :     32 (0.02%)
  Unsupported      :   5751 (4.11%)
  Passed           : 133367 (95.36%)
  Expectedly Failed:    269 (0.19%)
  Timed Out        :      1 (0.00%)
  Failed           :    431 (0.31%)
```
The build process works fine, but since only 10 failing tests are allowed in the easyconfig, it does not get finally installed. This is the complete log:
[easybuild-LLVM-20.1.5-20250612.091751.cYUNa.log.zip](https://github.com/user-attachments/files/20722095/easybuild-LLVM-20.1.5-20250612.091751.cYUNa.log.zip)

Crivella:
Many of the failing tests seems to be due to errors of the kind
```
# .---command stderr------------
# | error: OpenMP target is invalid: 'riscv64-unknown-linux-gnu'
# `-----------------------------
```

So i would guess just changing that is not enough of a fix.
It is possible is just a problem with the tests not recognizing differences in targets between stuff like `-pc-linux-` and `-unknown-linux`

We might have to add a different patch similar to https://github.com/easybuilders/easybuild-easyconfigs/blob/develop/easybuild/easyconfigs/l/LLVM/LLVM-19.1.7_libomptarget_tests.patch or modify it to catch this case

easybuild/easyblocks/l/llvm.py
```diff
@@ -667,6 +668,113 @@ def _update_test_ignore_patterns(self):
...
+            new_ignore_patterns.append('lldb-shell :: SymbolFile/DWARF/anon_class_w_and_wo_export_symbols.ll')
+            new_ignore_patterns.append(
+                'lldb-shell :: SymbolFile/DWARF/clang-ast-from-dwarf-unamed-and-anon-structs.cpp'
+            )
+            new_ignore_patterns.append('lldb-shell :: SymbolFile/DWARF/clang-gmodules-type-lookup.c')
+
+            # binutils 2.40 is too old and doesn't recognize the 'zaamo' ISA extension used in the test.
+            # With binutils >= 2.41, this test would work
+            binutils_ver = get_software_version('binutils')
+            if LooseVersion(binutils_ver) < LooseVersion('2.41'):
```
Crivella:
```suggestion
            if binutils_ver is None or LooseVersion(binutils_ver) < LooseVersion('2.41'):
```

Crivella: In this way, if `binutils` is not present (resulting `get_software_version` returning `None`), the test would always be ignored

Crivella:
Going in, thanks @julianmorillo!

PR #3678, opened 2025-03-28
Add option to impi easyblock to rebuild Fortran 2008 bindings

Thyre:
The Fortran 2008 bindings shiped with Intel MPI can cause
compilation failures when using `CLASS(*)` arguments.
This bug is known, and a solution for this given in the following
thread:

https://community.intel.com/t5/Intel-MPI-Library/MPI-f08-with-polymorphic-argument-CLASS/m-p/1590421

However, this bug is not yet included in Intel MPI releases to
preserve backwards compatibility. Instead, one needs to
rebuild the bindings themselves and replace them.
Add an option for this, and leave it disabled by default.
EasyConfigs can then decide if they want to rebuild these modules,
or preserve the backwards compatibility.


easybuild/easyblocks/i/impi.py
```diff
@@ -149,6 +154,36 @@ def post_processing_step(self):
...
+                    f"Requested to rebuild Fortran 2008 bindings, but the bindings tarball in {bindings_tarball} "
+                    f"does not exist.")
+            self.log.info(f"Found bindings tarball at {bindings_tarball}. Rebuilding Fortran 2008 bindings.")
+            # Extract the tarball
+            build_dir = tempfile.mkdtemp()
+            run_shell_cmd(f"tar -xzf {bindings_tarball} -C {build_dir}")
+            # Build the bindings
+            change_dir(os.path.join(build_dir, 'f08'))
+            run_shell_cmd(f"make MPI_INST={self.installdir}/mpi/latest F90=ifx NAME=ifx")
+            change_dir(os.path.join(get_cwd(), 'include', 'ifx'))
```
Thyre, line 175: Similar to the sanity check, we probably either want to add the option to rebuild with the toolchain compiler, or only allow that option if `intel-compilers` is the toolchain.
I need to figure out which compilers are supported by the `Makefile` withtin `intel-mpi-binding-kit.tar.gz`

easybuild/easyblocks/i/impi.py
```diff
@@ -42,7 +44,8 @@
 from easybuild.framework.easyconfig import CUSTOM
 from easybuild.tools.build_log import EasyBuildError
 from easybuild.tools.config import build_option
-from easybuild.tools.filetools import apply_regex_substitutions, change_dir, extract_file
+from easybuild.tools.filetools import apply_regex_substitutions, change_dir, extract_file, mkdir, \
+    copy_files, remove, get_cwd
```
boegel, line 48:
```suggestion
from easybuild.tools.filetools import apply_regex_substitutions, change_dir, copy_files, extract_file
from easybuild.tools.filetools import mkdir, remove, get_cwd

**Review by boegel — CHANGES_REQUESTED** [review]

File: easybuild/easyblocks/i/impi.py [hunk 3]
```diff
@@ -149,6 +154,41 @@ def post_processing_step(self):
...
+            build_dir = tempfile.mkdtemp()
+            run_shell_cmd(f"tar -xzf {bindings_tarball} -C {build_dir}")
+            # Build the bindings
+            change_dir(os.path.join(build_dir, 'f08'))
+            run_shell_cmd(f"make MPI_INST={self.installdir}/mpi/latest F90={f90_compiler} NAME={f90_compiler}")
+            change_dir(os.path.join(get_cwd(), 'include', f'{f90_compiler}'))
+            initial_module_files = glob.glob(f"{self.installdir}/mpi/latest/include/mpi/*.mod")
+            written_module_files = glob.glob("*.mod")
+            # Preserve the initial module files for people to use
+            mkdir(f"{self.installdir}/mpi/latest/include/mpi/back", parents=True)
```
boegel, line 184: I guess the `back` is short for backup?

Makes sense to introduce a local variable for this path, since it must be identical to what's used below.

I would also introduce a dedicated method for this whole code block.

Done in https://github.com/Thyre/easybuild-easyblocks/pull/1

boegel (changes requested):

boegel:
EB_ARGS="--installpath /tmp/$USER impi-2021.2.0-intel-compilers-2021.2.0.eb impi-2021.13.0-intel-compilers-2024.2.0.eb"

boegel (approved):
lgtm

PR #3070, opened 2024-01-09
update PyTorch easyblock to avoid configure warnings by disabling some options

Flamefire:
Explicitely disable some options to avoid warnings during the configure.

PR #3683, opened 2025-04-02
Use MAGMA option for PyTorch 2+

Flamefire:
(created using `eb --new-pr`)

I found that warning while investigating PyTorch 2.6.
This was likely a merge-conflict or oversight while creating https://github.com/easybuilders/easybuild-easyblocks/pull/3070 which introduced this:

- Magma should not be used in CPU builds
- In GPU builds it is optional

Reflect this correctly in the code. See their CMakeLists for reference: https://github.com/pytorch/pytorch/blob/40a6710ad34ae4c6f4987f0e47b5c94df3fc8ec7/cmake/Dependencies.cmake#L1692-L1706

Given that this is PyTorch I suggest to just inspect the code and don't do a full test run or at least with `--skip-test-step` as the change is pretty trivial


easybuild/easyblocks/p/pytorch.py
```diff
@@ -490,14 +490,20 @@ def add_enable_option(name, enabled):
...
         if pytorch_version >= '2.0':
-            add_enable_option('USE_ROCM', get_software_root('ROCm'))
-        elif pytorch_version >= 'v1.10.0':
-            add_enable_option('USE_MAGMA', get_software_root('magma'))
+            has_rocm = get_software_root('ROCm')
+            add_enable_option('USE_ROCM', has_rocm)
+            if has_rocm:
+                with_gpu_support = True
+
+        if pytorch_version >= 'v1.10.0':
```
jfgrimm, line 505: is the `v` in the version intentional? I would have thought that might trip up the version comparison, even with `LooseVersion`?

PR #3684, opened 2025-04-03
New easyblock for Basix

moravveji:


easybuild/easyblocks/b/basix.py
```diff
@@ -0,0 +1,60 @@
...
+class EB_Basix(CMakePythonPackage):
+    """Custom easyblock for Basix"""
+    def build_step(self):
+        """Build step need not be overridden"""
+        super(EB_Basix, self).build_step()
+
+        # Building 'lib64/libbasix.so'
+        self.cfg['build_type'] = 'Release'
+        self.cfg['srcdir'] = 'cpp'
+        self.cfg['build_cmd'] = 'cmake --build %s/easybuild_obj' % self.builddir
```
Flamefire, line 17: Why is this used? Especially "Build step need not be overridden" sounds like you don't need this whole method, do you?

easybuild/easyblocks/b/basix.py
```diff
@@ -0,0 +1,60 @@
...
+          1. Suppress the unwanted '--no-dep --ignore-cache' options of
+             the 'cmake --install <path>' command
+             Install the libbasix.so
+          2. Install the Python interface
+          3. Copy out lib the folder
+        """
+        self.cfg['installopts'] = ''
+        self.log.info('Resetting "installopts" in basix.py')
+
+        # step 1: Installing the C++ shared lib
```
Flamefire, line 31: You are not installing anything here, are you? I'd expect a `super().install_step()` call at this point.

Why exactly do you need to clear `installopts`?

easybuild/easyblocks/b/basix.py
```diff
@@ -0,0 +1,60 @@
...
+        self.install_cmd = 'cmake --install %s/easybuild_obj' % self.builddir
+        self.log.info('Installing lib64/libbasix.so')
+
+        # step 2: Installing the Python interface using PDM
+        _namelower = self.name.lower()
+        _loc_basix_dir = os.path.join(self.builddir, '%s-%s' % (_namelower, self.version))
```
Flamefire, line 37: IIRC this is just `self.start_dir`

easybuild/easyblocks/b/basix.py
```diff
@@ -0,0 +1,60 @@
...
+        out, err = run_cmd(_pdm_cmd, simple=False, log_ok=True, log_all=True, trace=True)
+        if err:
+            raise EasyBuildError('Error running "%s". message: %s' % (_pdm_cmd, out))
+        else:
+            self.log.info('"%s" succeeded' % _pdm_cmd)
+
+        # step 3: Copy out the new lib folder to the installation folder
+        _loc_venv_dir = os.path.join(_loc_basix_dir, '.venv')
+        _loc_venv_lib_dir = os.path.join(_loc_venv_dir, 'lib')
+        _cp_cmd = 'cp -r %s %s' % (_loc_venv_lib_dir, self.installdir)
```
Flamefire, line 48: No need for a copy command. Better use `copy_*` functions from `easybuild.tools.fileutils` (IIRC the name)

Flamefire (changes requested):
Is this required for some newer Basix version? There is another [Basix PR](https://github.com/easybuilders/easybuild-easyconfigs/pull/22422/files#diff-a7301e8dcd1fda8318b66978b0042239f0cacd98f3a1f3d267e66fa0775f705c) that seems to work fine with the regular `PythonPackage` easyblock

moravveji:
> Is this required for some newer Basix version?

I have followed the official installation steps, in order to make sure I build the C++ shared library `libbasix.so` alongside the Python interface. That is the major difference between the approach here, and the one in the other PR (which is limited only to the Python interface).

Flamefire:
> I have followed the official installation steps, in order to make sure I build the C++ shared library `libbasix.so` alongside the Python interface. That is the major difference between the approach here, and the one in the other PR (which is limited only to the Python interface).

I think I'm missing something here. The [documentation](https://github.com/FEniCS/basix/blob/main/INSTALL.md) states:
 > Basix can be installed using `pip install .`
> ### Advanced:
> It is also possible to install the C++ and Python interfaces separately (see below). This is useful if you only need the C++ interface

This to me sounds like the pip build installs both the Python and the C++ interface and additionally you can also build the C++ library separately first with CMake which would then be picked up by the pip installation.
However in that documentation there is no mention of `pdm install` which makes me wonder whether there is a different documentation that achieves more than this basic one.

moravveji:
> This to me sounds like the pip build installs both the Python and the C++ interface

This is correct. I wonder how I did not clearly test for that from the beginning. This means that most probably, I do not require this easyblock at all. I shall try to build `Basix` as a `PythonBundle` trying to omit the need for this easyblock. I will keep you posted here.

Flamefire:
> I shall try to build `Basix` as a `PythonBundle` trying to omit the need for this easyblock. I will keep you posted here.

Great. Looks like you could even just use a basic PythonPackage.

PR #3686, opened 2025-04-04
Fix access to unassigned variable when installing Python < 3.9.1 with using sysroot 

Flamefire:
```
if LooseVersion(self.version) >= "3.9.1":
    orig_ld_config_call = "with subprocess.Popen(['/sbin/ldconfig', '-p'],"
if orig_ld_config_call:
```

This code will crash when the Python version to be installed is less than 3.9.1 as the variable is unbound/unset in that case.
Fix is trivial: Remove the line of the 2nd if as this is the only place where the variable is set

I included 2 other changes suggested by PyLint:
- Use `dict.items` instead of iterating over keys and then requesting values
- Combine nested (single) ifs to remove one level of indentation at the cost of a longer condition

I can remove one or both again but IMO they make sense. At least the former which doesn't has any drawbacks I can see

### DIFF
--- a/easybuild/easyblocks/p/python.py
+++ b/easybuild/easyblocks/p/python.py
@@ -290,10 +290,9 @@ class EB_Python(ConfigureMake):
         self.log.debug("exts_default_options: %s", self.cfg['exts_default_options'])
 
         self.install_pip = self.cfg['install_pip']
-        if self.install_pip:
-            if not self._has_ensure_pip():
-                raise EasyBuildError("The ensurepip module required to install pip (requested by install_pip=True) "
-                                     "is not available in Python %s", self.version)
+        if self.install_pip and not self._has_ensure_pip():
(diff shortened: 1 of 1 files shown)


Flamefire:
@boegel is this a candidate for 5.0.1? Excluding the spaces the change is trivial

boegel (approved):
lgtm

boegel:
Test report by @boegel
 * **SUCCESS** _Python-2.7.18-GCCcore-12.3.0.eb_ 
 * **SUCCESS** _Python-3.13.1-GCCcore-14.2.0.eb_ 

node3502.doduo.os - Linux RHEL 9.4, x86_64, AMD EPYC 7552 48-Core Processor (zen2), Python 3.9.18
See https://gist.github.com/boegel/3c86b3d73436a37a282e5285312a14ac for a full test report.

Issue #2881, opened 2023-02-08
Regression: Broken templating in EasyBuild 4.7.0


mboisson:
I am not sure exactly when it happened, but recent EasyBuild versions broke template replacement in `configopts` when using `multi_deps`, i.e. : 
```python
multi_deps = {'Python': ['3.10', '3.9', '3.8'] }
configopts = "-DPYQT5_SIP_DIR=$EBROOTQT5/share/python%(pyshortver)s/site-packages/sip/PyQt5 -DPYQT5_MOD_DIR=$EBROOTQT5/lib/python%(pyshortver)s/site-packages/PyQt5 "
``` 
for a `CMakeMake` EasyBlock (maybe others, I have not tested) no longer yield builds with different `configopts` argument based on `%(pyshortver)s`. 

In a previous build log, `%(pyshortver)s` gets replaced for every iteration: 
```
grep run.py /cvmfs/soft.computecanada.ca/easybuild/software/2020/avx2/Compiler/gcc9/qgis/3.16.10/easybuild/*.log | grep cmake | grep running
[...]

PR #3693, opened 2025-04-09
Don't modify `self.cfg['configopts']` in CMakeMake easyblock

bartoldeman:
We use `multi_deps` a lot, e.g. in
https://github.com/ComputeCanada/easybuild-easyconfigs/blob/computecanada-main/easybuild/easyconfigs/r/RDKit/RDKit-2024.03.5-GCC-12.3.0.eb
The problem was now that `CMakeMake` automatically adds `-DPython_EXECUTABLE` and `-DPython3_EXECUTABLE` pointing to e.g. first python 3.10 but in the second iteration the logic keeps it that way in `configopts`, and it's wrong since there it should be pointing to 3.11 instead.

We can avoid this by *not* modifying `configopts`, so it's reconstructed from scratch in every iteration. Another way would be to use `$EBROOTPYTHON` and let the shell expand the path, instead of `get_software_root()`.

### DIFF
--- a/easybuild/easyblocks/generic/cmakemake.py
+++ b/easybuild/easyblocks/generic/cmakemake.py
@@ -213,7 +213,7 @@ class CMakeMake(ConfigureMake):
         # All options are of the form '-D<key>=<value>'
         new_opts = ' '.join('-D%s=%s' % (key, value) for key, value in config_opts.items()
                             if '-D%s=' % key not in cfg_configopts)
-        self.cfg['configopts'] = ' '.join([new_opts, cfg_configopts])
+        return ' '.join([new_opts, cfg_configopts])
 
     def configure_step(self, srcdir=None, builddir=None):
         """Configure build using cmake"""
@@ -383,12 +383,11 @@ class CMakeMake(ConfigureMake):
         self.cmake_options = options
 
         if self.cfg.get('configure_cmd') == DEFAULT_CONFIGURE_CMD:
-            self.prepend_config_opts(options)
             command = ' '.join([
                 self.cfg['preconfigopts'],
                 DEFAULT_CONFIGURE_CMD,
                 generator,
-                self.cfg['configopts'],
+                self.prepend_config_opts(options),
                 srcdir])
         else:
             command = ' '.join([


bartoldeman:
Tagging @flamefire and @crivella as authors of the code I just modified here.

Crivella:
I've checked that no other easyblock uses `CMakeMake.prepend_config_opts`, but we should also be careful as some easyblock could potentially do stuff with `cfg['configopts']` after calling configure.
EG i do something similar in the LLVM easyblock but taking configopts before (not after) calling configure and than reusing it for the intermediate build stages.

Flamefire:
I think this is rather a bug in framework that should restore easyconfig parameters on iterations, shouldn't it? But maybe not for multideps?
There is also this comment in this function:
>         # need to disable template resolution or it will remain the same for all runs

Not sure what "all runs" means though.

Reason why I'd be careful with changing it here is that other easyblock might react on options set here and e.g. add/change defaults. But I could find anything in the official ones by some grep-magic.

So if the no-restoring behavior for multideps IS intended then I'd say this is fine.

bartoldeman:
> > ```
> >     # need to disable template resolution or it will remain the same for all runs
> > ```
> 
> Not sure what "all runs" means though.

that was the same idea, but with e.g. `%(pyshortver)s`, that would otherwise be expanded too early.
see #2881

Indeed addressing this in framework may be best as it would avoid these kind of headaches once and for all.

boegel:
After checking the discussion, I agree it would be better to properly "reset" things before starting a new iteration in framework.
That's a bug fix for sure, but there may be fallout there too: maybe some easyblocks currently (unintentionally) rely on that bug? I'd say that's unlikely though...

@bartoldeman How do you think we should proceed here? Should we merge this (after thorough testing) so it's included in the upcoming EasyBuild v5.0.1 release, or can we wait until this is properly fixed in framework (through your PR https://github.com/easybuilders/easybuild-framework/pull/4848)

bartoldeman:
I'm hopeful a framework change could be simple enough to include in 5.0.1.. I'll get back to you about this tomorrow.

boegel:
@bartoldeman This can be closed since https://github.com/easybuilders/easybuild-framework/pull/4848 got merged?

PR #3695, opened 2025-04-09
Fix incorrect glob pattern & use `report_test_failure` in test step in LLVM easyblock

Thyre:
Cherry-picked from #3675 for EasyBuild 5.0.1


boegel (approved):
lgtm

PR #3697, opened 2025-04-11
Workaround for NWChem: prepend "-" to "DMKL_ILP64" flag

thanghn90:
The easybuild framework intelmkl.py is missing a "-" before "DMKL_ILP64" flag, causing NWChem to fail to build. This workaround will search for " DMKL_ILP64" (note the space at the beginning) in CFLAGS and replace it with " -DMKL_ILP64". This will avoid double-replacing the correct flag should easybuild framework devs decide to fix their intelmkl.py.


Micket:
I'll just go ahead and merge the framework fix Anyone wanting a workaround can use 
```
eb --include-easyblock-from-pr 3697
```
(which will work even when i close this PR)

PR #3699, opened 2025-04-15
Update tensorflow.py for v2.17.1

pavelToman:


easybuild/easyblocks/t/tensorflow.py
```diff
@@ -954,23 +954,26 @@ def build_step(self):
...
             + ['bazel']
             + self.bazel_opts
             + ['build']
             + self.target_opts
             + [self.cfg['buildopts']]
+            # add PATH to bazel build cmd
+            + [f"--action_env=PATH={binpath}:$PATH --host_action_env=PATH={binpath}:$PATH "]
             # specify target of the build command as last argument
-            + ['//tensorflow/tools/pip_package:build_pip_package']
+            + ['//tensorflow/tools/pip_package:wheel --repo_env=WHEEL_NAME=tensorflow']
```
Flamefire, line 968: This will break earlier tensorflow versions.

easybuild/easyblocks/t/tensorflow.py
```diff
@@ -954,23 +954,26 @@ def build_step(self):
...
         )
 
         with self.set_tmp_dir():
             run_shell_cmd(' '.join(cmd))
 
-            # run generated 'build_pip_package' script to build the .whl
-            cmd = "bazel-bin/tensorflow/tools/pip_package/build_pip_package %s" % self.builddir
-            run_shell_cmd(cmd)
+            # run generated 'build_pip_package' script to build the .whl -> not used ?
+            # cmd = "bazel-bin/tensorflow/tools/pip_package/build_pip_package %s" % self.builddir
```
Flamefire, line 975: Same here. If this is required/not required for TF 2.17 guard it by a version check

easybuild/easyblocks/t/tensorflow.py
pavelToman: yes, but I will update this as soon as I made 2.17.1 works

easybuild/easyblocks/t/tensorflow.py
pavelToman: I hope this is not used in 2.17.1 anymore. I will update this properly with versions check as soon as I made 2.17.1 works

easybuild/easyblocks/t/tensorflow.py
```diff
@@ -954,23 +954,26 @@ def build_step(self):
...
+        binpath = os.getenv('PATH', '')
         cmd = (
             [self.cfg['prebuildopts']]
             + ['bazel']
             + self.bazel_opts
             + ['build']
             + self.target_opts
             + [self.cfg['buildopts']]
+            # add PATH to bazel build cmd
+            + [f"--action_env=PATH={binpath}:$PATH --host_action_env=PATH={binpath}:$PATH "]
```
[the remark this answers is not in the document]
Flamefire: Basically yes.
But there is a dictionary above called `action_env` which should do. I also don't really understand this approach: You basically do `PATH=$PATH:$PATH` with the `binpath` above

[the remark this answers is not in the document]
pavelToman: I will add PATH to `action_env ` if it is really needed - I add it for now to fix patchelf problem - seems it is not working.

The `--action_env=PATH={binpath}:$PATH` should add to bazel build subcommands the PATH envs: 
```PATH=/tmp/vsc47063/eb-avwk66j7/tmp_dp09jye/rpath_wrappers/ld.bfd_wrapper:/tmp/vsc47063/eb-avwk66j7/tmp_dp09jye/rpath_wrappers/ld.gold_wrapper:/tmp/vsc47063/eb-avwk66j7/tmp_dp09jye/rpath_wrappers/ld_wrapper:/tmp/vsc47063/eb-avwk66j7/tmp_dp09jye/rpath_wrappers/gfortran_wrapper:/tmp/vsc47063/eb-avwk66j7/tmp_dp09jye/rpath_wrappers/gxx_wrapper:/tmp/vsc47063/eb-avwk66j7/tmp_dp09jye/rpath_wrappers/gcc_wrapper:/scratch/gent/vo/001/gvo00117/easybuild/RHEL9/cascadelake-ampere-ib/software/TensorFlow/2.17.1-foss-2024a/sbin: ... :$PATH```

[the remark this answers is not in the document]
Comment (Flamefire) [reply easybuild/easyblocks/t/tensorflow.py:966]: What I meant: You read `$PATH`, save it in `binpath` and then append `$PATH` to that again which looks redundant.
Using the `action_env` dict is more flexible and e.g. takes care of `host_action_env`

But issue is likely the missing `use_default_shell_env` in the source and this change here has no effect on that. There is a similar patch already that fixes (likely) the same issue in other places

File: easybuild/easyblocks/t/tensorflow.py [hunk 4]
```diff
@@ -961,16 +961,20 @@ def build_step(self):
             + ['build']
             + self.target_opts
             + [self.cfg['buildopts']]
-            # specify target of the build command as last argument
-            + ['//tensorflow/tools/pip_package:build_pip_package']
         )
+        if LooseVersion(self.version) < LooseVersion('2.16'):
```
Flamefire, line 965: Just FYI: `LooseVersion(self.version) < '2.16':` is equivalent, if you prefer that.

easybuild/easyblocks/t/tensorflow.py
```diff
@@ -961,16 +961,20 @@ def build_step(self):
...
+        else:
+            cmd += ['//tensorflow/tools/pip_package:wheel --repo_env=WHEEL_NAME=tensorflow']
 
         with self.set_tmp_dir():
             run_shell_cmd(' '.join(cmd))
-
-            # run generated 'build_pip_package' script to build the .whl
-            cmd = "bazel-bin/tensorflow/tools/pip_package/build_pip_package %s" % self.builddir
-            run_shell_cmd(cmd)
+            if LooseVersion(self.version) < LooseVersion('2.16'):
```
boegel, line 974: Can you clarify the comment below to mention why calling `build_pip_package` is no longer required for TensorFlow >= 2.16?

boegel (changes requested):

easybuild/easyblocks/t/tensorflow.py
Flamefire: We discussed this in Slack and didn't copy it here:
TF removed this target in favor of a wheel build: https://github.com/tensorflow/tensorflow/issues/63298#issuecomment-1986352608

Basically: Instead of creating a "tool" that will then build a pip-package/wheel they now have a target to do that directly.

pavelToman: https://easybuild.slack.com/archives/C34UA1HT7/p1744709459947309

pavelToman:
EB_ARGS="TensorFlow-2.15.1-foss-2023a.eb TensorFlow-2.13.0-foss-2023a.eb"

easybuild/easyblocks/t/tensorflow.py
```diff
@@ -595,11 +595,12 @@ def configure_step(self):
             print_warning('Jemalloc is not supported in TensorFlow %s, the EC option with_jemalloc has no effect',
                           self.version)
         # Disable support of some features via config switch introduced in 1.12.1
-        if LooseVersion(self.version) >= LooseVersion('1.12.1'):
-            self.target_opts += ['--config=noaws', '--config=nogcp', '--config=nohdfs']
-            # Removed in 2.1
-            if LooseVersion(self.version) < LooseVersion('2.1'):
-                self.target_opts.append('--config=nokafka')
+        if LooseVersion(self.version) >= LooseVersion("1.12.1"):
```
Flamefire, line 598: nit: The changes to double-quotes causes the diff to look bigger than it should. Was there a reason? For consistency in this file I'd rather use single quotes.

easybuild/easyblocks/t/tensorflow.py
```diff
@@ -595,11 +595,12 @@ def configure_step(self):
...
+            self.target_opts += ["--config=nogcp"]
+            if LooseVersion(self.version) < LooseVersion("2.18"):
+                self.target_opts += ["--config=noaws", "--config=nohdfs"]  # works in 2.17.1 not in 2.18.1
```
Flamefire, line 601: The comment is a bit odd given the checked versions. Better:
```suggestion
                self.target_opts += ['--config=noaws', '--config=nohdfs']  # Removed in 2.18
```

I found https://github.com/tensorflow/tensorflow/pull/72475 as a related PR upstream although this isn't the one where the support for the option was removed.

easybuild/easyblocks/t/tensorflow.py
```diff
@@ -595,11 +595,12 @@ def configure_step(self):
             print_warning('Jemalloc is not supported in TensorFlow %s, the EC option with_jemalloc has no effect',
                           self.version)
         # Disable support of some features via config switch introduced in 1.12.1
-        if LooseVersion(self.version) >= LooseVersion('1.12.1'):
-            self.target_opts += ['--config=noaws', '--config=nogcp', '--config=nohdfs']
-            # Removed in 2.1
-            if LooseVersion(self.version) < LooseVersion('2.1'):
-                self.target_opts.append('--config=nokafka')
+        if LooseVersion(self.version) >= LooseVersion("1.12.1"):
+            self.target_opts += ["--config=nogcp"]
```
Flamefire, line 599: If you touch this part of the code: Maybe we should either consistently use `+= [...]` or `.append` at least for single elements

easybuild/easyblocks/t/tensorflow.py
```diff
@@ -1136,10 +1141,16 @@ def install_step(self):
...
-            whl_paths = glob.glob(os.path.join(self.builddir, 'tensorflow-*.whl'))
+        if LooseVersion(self.version) < LooseVersion('2.16'):
+            whl_paths = glob.glob(os.path.join(self.builddir, f"tensorflow-{whl_version}-*.whl"))
+            if not whl_paths:
+                whl_paths = glob.glob(os.path.join(self.builddir, 'tensorflow-*.whl'))
+        else:
+            whl_paths = glob.glob(os.path.join(
+                self.builddir,
+                f'TensorFlow/tensorflow-{self.version}/bazel-bin/tensorflow/tools/pip_package/wheel_house',
+                f"tensorflow-{whl_version}-*.whl"
```
Flamefire, line 1152: The other code path has a fallback. Not sure how relevant this still is but maybe we should keep that.
Idea: set `whl_dir = self.builddir` if <2.16 else `that-long-thing` and then use the existing code with the fallback with whl_dir instead of builddir

Flamefire (commented):
A bit polishing and the last comment looks possibly important.

The PR looks good to me and those changes aren't very important.

easybuild/easyblocks/t/tensorflow.py
pavelToman: the black formater is the reason :)

Flamefire: I see: https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#strings
> Black prefers double quotes (" and """) over single quotes (' and '''). It will replace the latter with the former as long as it does not result in more backslash escapes than before.

easybuild/easyblocks/t/tensorflow.py
```diff
@@ -1136,10 +1141,12 @@ def install_step(self):
             whl_version = self.version.replace("-rc", "rc")
         else:
             whl_version = self.version
-
-        whl_paths = glob.glob(os.path.join(self.builddir, 'tensorflow-%s-*.whl' % whl_version))
+        whl_dir = self.builddir if LooseVersion(self.version) < '2.16' else (
+            f"{self.builddir}/TensorFlow/tensorflow-{self.version}/bazel-bin/tensorflow/tools/pip_package/wheel_house"
+        )
```
Flamefire, line 1146: uff. That looks more ugly than I anticipated... ;-)
I think this is a bit more readable and easier to extend in case they change their path again:
```suggestion
        if LooseVersion(self.version) < '2.16':
            whl_dir = self.builddir
        else:
            whl_dir = os.path.join(self.start_dir, 'bazel-bin/tensorflow/tools/pip_package/wheel_house')
```

We usually don't hard-code path separators, but I guess it's fine in this case as the alternative would be very verbose.
But at least the parent folder is not hidden this way. Using `start_dir` makes it much shorter

Flamefire:
TF 2.19 seems to require `--config=cuda_wheel` independent whether CUDA being used. Otherwise [it fails ](https://github.com/tensorflow/tensorflow/blob/a3736ccf6be0321e5f4773d8921d05b171dc2957/tensorflow/tools/pip_package/utils/tf_wheel.bzl#L82)with
> Please provide `--config=cuda_wheel` for bazel build command. If you absolutely need to add CUDA dependencies, provide `--@local_config_cuda//cuda:override_include_cuda_libs=true`.

This flag is available [since 2.18](https://github.com/tensorflow/tensorflow/commit/d9071b91c4e550e6c984357158c3460346616db5)

easybuild/easyblocks/t/tensorflow.py
```diff
@@ -902,9 +903,11 @@ def build_step(self):
         action_env['PYTHONNOUSERSITE'] = '1'
 
         # TF 2 (final) sets this in configure
-        if LooseVersion(self.version) < LooseVersion('2.0'):
-            if self._with_cuda:
-                self.target_opts.append('--config=cuda')
+        if (LooseVersion(self.version) < LooseVersion('2.0')) and self._with_cuda:
+            self.target_opts.append('--config=cuda')
+        # TF 2.18 with CUDA needs to set cuda_wheel to config
+        if (LooseVersion(self.version) >= LooseVersion('2.18')) and self._with_cuda:
```
Flamefire, line 909: From how I read the error this is required for non-CUDA builds too. Or did you test this or got other information?

pavelToman:
EB_ARGS="TensorFlow-2.15.1-foss-2023a-CUDA-12.1.1.eb"

pavelToman:
EB_ARGS="TensorFlow-2.15.1-foss-2023a-CUDA-12.1.1.eb"

boegel (approved):
lgtm

boegel:
Test report by @boegel

Build succeeded for 0 out of 0 (1 easyconfigs in total)
node4013.donphan.os - Linux RHEL 9.4, x86_64, Intel(R) Xeon(R) Gold 6240 CPU @ 2.60GHz (cascadelake), 1 x NVIDIA NVIDIA A2, 570.86.15, Python 3.9.18
See https://gist.github.com/boegel/871282338c11b56373a1e0dbcf82b354 for a full test report.

boegel:
Test report by @boegel
 * **SUCCESS** _Bazel-6.1.0-GCCcore-12.3.0.eb_ 
 * **SUCCESS** _TensorFlow-2.15.1-foss-2023a-CUDA-12.1.1.eb_ 

Build succeeded for 2 out of 2 (1 easyconfigs in total)
node3905.accelgor.os - Linux RHEL 9.4, x86_64, AMD EPYC 7413 24-Core Processor (zen3), 1 x NVIDIA NVIDIA A100-SXM4-80GB, 570.133.20, Python 3.9.18
See https://gist.github.com/boegel/b535d1f67711010bbcec8be180a74e13 for a full test report.

boegel:
Test report by @boegel
 * **SUCCESS** _TensorFlow-2.15.1-foss-2023a-CUDA-12.1.1.eb_ 

node4013.donphan.os - Linux RHEL 9.4, x86_64, Intel(R) Xeon(R) Gold 6240 CPU @ 2.60GHz (cascadelake), 1 x NVIDIA NVIDIA A2, 570.86.15, Python 3.9.18
See https://gist.github.com/boegel/6fd3e6aee60833d1b98ec59b63a4dd5a for a full test report.

PR #3701, opened 2025-04-23
Remove superflous assignment of sanity check module data in custom easyblock for torchvision

Flamefire:
`sanity_check_load_module` returns `self.fake_mod_data` so there is no need to assign it to itself. Remove it to avoid confusion.

Similar `EasyBlock` is guaranteed to have the attribute so remove the check.

See https://github.com/easybuilders/easybuild-framework/pull/4851

### DIFF
--- a/easybuild/easyblocks/t/torchvision.py
+++ b/easybuild/easyblocks/t/torchvision.py
@@ -82,8 +82,8 @@ class EB_torchvision(PythonPackage):
 
         # load module early ourselves rather than letting parent sanity_check_step method do so,
         # so the correct 'python' command is used to by det_pylibdir() below;
-        if hasattr(self, 'sanity_check_module_loaded') and not self.sanity_check_module_loaded:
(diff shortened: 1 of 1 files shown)


boegel:
EB_ARGS="--installpath /tmp/$USER/pr3701 torchvision-0.17.0-foss-2023b.eb"

boegel:
> > test_sparse_csr 1/1 failed! Received signal: SIGKILL
> 
> Memory issue or something like that? Any further info in the log?

Those logs aren't kept, unfortunately.

boegel:
EB_ARGS="--installpath /tmp/$USER/pr3701 torchvision-0.17.0-foss-2023b.eb"
CORE_CNT=16

boegel (approved):
Also tested via https://github.com/easybuilders/easybuild-easyconfigs/pull/22937, good to go, thanks @Flamefire!

PR #3703, opened 2025-04-25
binutils: Add option to build without ld.gold

Thyre:
Required for binutils 2.44 and newer versions going forward, as ld.gold with be phased out. Not every release will contain ld.gold, causing the sanity checks to fail.

The alternative would be to check if the sources support `ld.gold`, or choose `binutils-with-gold-[...].tar` for now.

See https://lists.gnu.org/archive/html/info-gnu/2025-02/msg00001.html for more information.


jfgrimm:
if their plan is to drop ld.gold altogether, we might as well just make it a version check instead of introducing an easyconfig parameter?
```
if LooseVersion(self.version) >= LooseVersion('2.44') or get_cpu_family() == RISCV:
    # ld.gold linker is not supported on RISC-V, and is being phased out starting from v2.44
    self.use_gold = False
else:
     self.use_gold = True
```

Thyre:
That's not entirely determined. In the release notes, they mention:
> This is because the gold linker is now deprecated and will eventually be removed unless volunteers step forward and offer to continue development and maintenance.

At the moment, I would expect that `ld.gold` will be gone at some point.
I'll change to your suggestion though, this is easier to maintain, and keeps the EasyConfigs clean.

jfgrimm:
with this change, I suppose we might want to update the `gcc.py` easyblockas well: https://github.com/easybuilders/easybuild-easyblocks/blob/develop/easybuild/easyblocks/g/gcc.py#L685

Thyre:
Right, probably a good idea to either abort when `use_gold_linker` is used with a `binutils` version which doesn't support the gold linker anymore, or at least raise a warning and switch to the else case.

I'll take a look.

Thyre:
EB_ARGS="GCCcore-10.2.0.eb GCCcore-12.3.0.eb GCCcore-14.2.0.eb --installpath /tmp/$USER/pr3703"

Thyre:
Test report by @Thyre
 * **SUCCESS** _GCCcore-14.2.0.eb_ 

Linux - Linux Arch Linux UNKNOWN, x86_64, AMD Ryzen 7 7800X3D 8-Core Processor, 1 x AMD Navi 48 [RX 9070/9070 XT] (device id: 0x7550, gfx: gfx1201, driver: 6.14.3-arch1-1), 1 x AMD Raphael (device id: 0x164e, gfx: gfx1036, driver: 6.14.3-arch1-1), Python 3.13.3
See https://gist.github.com/Thyre/f0bfba24a08032c7bc4a076017fd8026 for a full test report.

Thyre:
EB_ARGS="GCCcore-10.2.0.eb GCCcore-12.3.0.eb GCCcore-14.2.0.eb --installpath /tmp/$USER/pr3703"

jfgrimm (approved):
lgtm

jfgrimm:
Going in, thanks @Thyre!

PR #3584, opened 2025-02-09
honor `--module-search-path-headers` in easyblocks that hardcode CPATH

lexming:
Incorporate changes in https://github.com/easybuilders/easybuild-framework/pull/4655 to easyblocks that hardcode `CPATH`:
* flexiblas: follow-up to #3529
* gcc : follow-up to #3556
* icc, iccifort, ifort: follow-up to #3550
* intelbase: follow-up to #3551
* intel-compilers: follow-up to #3552
* imkl: follow-up to ##3555
* impi: follow-up to #3553

PR #3713, opened 2025-05-07
Add an option to GCC easyblock to export `LIBRARY_PATH`

ocaisa:
It's come up in EESSI that setting this can be useful when using the GCC compiler directly, see https://www.eessi.io/docs/using_eessi/building_on_eessi/#manually-building-software-on-top-of-eessi-without-easybuild

Basically with this change, instead of having to do
```
export LD_RUN_PATH=$LIBRARY_PATH:$EBROOTGCCCORE/lib64
```
to get a working executable, you would just need
```
export LD_RUN_PATH=$LIBRARY_PATH
```
which requires less explaining.


ocaisa:
This is not a correct fix and has actually been already addressed in https://github.com/easybuilders/easybuild-easyblocks/pull/3584 (see https://github.com/easybuilders/easybuild-easyblocks/blob/develop/easybuild/easyblocks/g/gcc.py#L206-L212)

PR #3728, opened 2025-05-23
add `-std=c++11` to `$CXXFLAGS` when building binutils 2.42+ with system GCC 4.8.1 to 5.x

lijiebio:
https://github.com/easybuilders/easybuild-easyconfigs/issues/22651

### DIFF
--- a/easybuild/easyblocks/b/binutils.py
+++ b/easybuild/easyblocks/b/binutils.py
@@ -209,7 +209,8 @@ class EB_binutils(ConfigureMake):
                     LooseVersion(gcc_version) >= LooseVersion('4.8.1')
                     and LooseVersion(gcc_version) < LooseVersion('6.1.0')
                 ):
-                    self.cfg.update('buildopts', 'CXXFLAGS="-std=c++11"')
+                    # append "-std=c++11" to $CXXFLAGS, not overriding
+                    self.cfg.update('buildopts', 'CXXFLAGS="$CXXFLAGS -std=c++11"')
 
     def install_step(self):
         """Install using 'make install', also install libiberty if desired."""
--- a/easybuild/easyblocks/b/binutils.py
+++ b/easybuild/easyblocks/b/binutils.py
@@ -204,11 +204,10 @@ class EB_binutils(ConfigureMake):
                 LooseVersion(self.version) >= LooseVersion('2.42')
                 and self.toolchain.comp_family() == toolchain.SYSTEM
             ):
-                res = run_shell_cmd('g++ --version')
-                gxx_version = res.output.strip().split(' ')[2]
+                gcc_version = get_gcc_version()
                 if (
-                    LooseVersion(gxx_version) >= LooseVersion('4.8.1')
-                    and LooseVersion(gxx_version) < LooseVersion('6.1.0')
+                    LooseVersion(gcc_version) >= LooseVersion('4.8.1')
+                    and LooseVersion(gcc_version) < LooseVersion('6.1.0')
                 ):
                     self.cfg.update('buildopts', 'CXXFLAGS="-std=c++11"')
 
--- a/easybuild/easyblocks/b/binutils.py
+++ b/easybuild/easyblocks/b/binutils.py
@@ -101,6 +101,8 @@ class EB_binutils(ConfigureMake):
     def configure_step(self):
         """Custom configuration procedure for binutils: statically link to zlib, configure options."""
 
+        version = LooseVersion(self.version)
+
         if self.toolchain.is_system_toolchain():
             # determine list of 'lib' directories to use rpath for;
             # this should 'harden' the resulting binutils to bootstrap GCC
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/b/binutils.py
```diff
@@ -199,6 +200,18 @@ def configure_step(self):
                 # since not specifying any optimization level implies -O0...
                 self.cfg.update('buildopts', 'CFLAGS="-g -O2 -fPIC"')
 
+            if (
+                LooseVersion(self.version) >= LooseVersion('2.42')
+                and self.toolchain.comp_family() == toolchain.SYSTEM
+            ):
+                res = run_shell_cmd('g++ --version')
+                gxx_version = res.output.strip().split(' ')[2]
```
akesandgren, line 208: This is not strictly correct, it should be 
```
gxx_version = res.output.strip().split(' ')[-1]
```
Since it is the last part of the output that contains the version.

On Ubuntu [2] would return something similar to "13.3.0-6ubuntu2~24.04)" while [-1] returns "13.3.0"

Also, I don't really know if we want to support a SYSTEM GCC compiler that old...
What is it from?

akesandgren: And this can be easily solved by adding a hook that adds that to configopts (where it should really go) for a site with such an old system gcc.
And it should add -std=c++11 to CXXFLAGS, not override it in the style of the lines above.

boegel: We have a function for this in framework, better use this?

Either `get_gcc_version`, or `get_tool_version('g++') (both from `easybuild.tools.systemtools`)

boegel (changes requested):

easybuild/easyblocks/b/binutils.py
lijiebio: My command to get g++ version is indeed not strictly correct. I will commit after.

```
gxx_version = res.output.strip().split(' ')[-1]
```
This is not strictly correct for RHEL-based Linux distributions as well. Maybe `g++ -v` not `g++ --version` is better. The last line of `g++ -v` output string looks easier to parse. I will check it later.

The reason I created this PR is I'm working on CentOS 7 to compile universal binaries targeting glibc 2.17. `eb foss-2024a.eb -r` will first compile binutils-2.42 with SYSTEM toolchain including gcc-4.8.5. However, gcc-4.8.5 uses the dafault C++ standard (C++ 98) during compilation even if gcc-4.8.5 had already support C++ 11 standard. The build option `--std=c++11` must be explicitly specified. This will cause compilation failure, as described in the issue https://github.com/easybuilders/easybuild-easyconfigs/issues/22651.

Next, what is the range of affected versions of gcc? See https://gcc.gnu.org/projects/cxx-status.html.
> GCC 4.8.1 was the first feature-complete implementation of the 2011 C++ standard, previously known as C++0x.

So, the start of range is 4.8.1 (include 4.8.1).
> This mode is the default in GCC 6.1 up until GCC 10 (inclusive); it can be explicitly selected with the -std=c++14 command-line flag, or -std=gnu++14 to enable GNU extensions as well.

Up util gcc-6.1.0, the default standard changed to C++ 14. So, the end of range is 6.1.0(not include 6.1.0). That is, gcc 4.8.1 to 5.x. The command `g++ -dM -E -x c++ /dev/null | grep __cplusplus` can be secondary verification.

lijiebio: > We have a function for this in framework, better use this?
> 
> Either `get_gcc_version`, or `get_tool_version('g++') (both from `easybuild.tools.systemtools`)

I will try this. Thank you.

boegel:
EB_ARGS="binutils-2.32.eb binutils-2.38.eb binutils-2.42.eb binutils-2.44.eb binutils-2.37-GCCcore-11.2.0.eb binutils-2.42-GCCcore-13.3.0.eb --installpath /tmp/$USER/pr3728"
CORE_CNT=16

easybuild/easyblocks/b/binutils.py
boegel: One concrete example that you'll hit with the soon to be released EasyBuild v5.1.0: https://github.com/easybuilders/easybuild-easyconfigs/pull/22798

boegel (approved):

boegel:
Test report by @boegel
 * **SUCCESS** _binutils-2.32.eb_ 
 * **SUCCESS** _binutils-2.38.eb_ 
 * **SUCCESS** _binutils-2.42.eb_ 
 * **SUCCESS** _binutils-2.44.eb_ 
 * **SUCCESS** _binutils-2.37-GCCcore-11.2.0.eb_ 
 * **SUCCESS** _binutils-2.42-GCCcore-13.3.0.eb_ 

node4221.shinx.os - Linux RHEL 9.4, x86_64, AMD EPYC 9654 96-Core Processor (zen4), Python 3.9.18
See https://gist.github.com/boegel/faf99dd956810307c7ebf2983f6f2193 for a full test report.

boegel:
Test report by @boegel
 * **SUCCESS** _binutils-2.32.eb_ 
 * **SUCCESS** _binutils-2.38.eb_ 
 * **SUCCESS** _binutils-2.42.eb_ 
 * **SUCCESS** _binutils-2.44.eb_ 

node3502.doduo.os - Linux CentOS Linux 7.9.2009, x86_64, AMD EPYC 7552 48-Core Processor (zen2), Python 3.6.8
See https://gist.github.com/boegel/fa413e7084636d79bc08748e6dca9ad5 for a full test report.

PR #3732, opened 2025-05-23
produce clean trace message for '`pip check`'

boegel:
Currently:

```
== sanity checking...
  >> running shell command:
        python -m pip check
        [started at: 2025-05-23 10:26:11]
        [working dir: /tmp/vsc40023/easybuild_build/Python/3.13.1/GCCcore-14.2.0]
        [output and state saved to /tmp/eb-d_ec_z2f/run-shell-cmd-output/python-54b60zib]
```

with this change:

```
== sanity checking...
  >> Check on requirements for installed Python packages with 'pip check': OK
```

### DIFF
--- a/easybuild/easyblocks/p/python.py
+++ b/easybuild/easyblocks/p/python.py
@@ -53,6 +53,7 @@ from easybuild.tools.filetools import apply_regex_substitutions, change_dir, mkd
 from easybuild.tools.filetools import read_file, remove_dir, symlink, write_file
 from easybuild.tools.run import run_shell_cmd
 from easybuild.tools.systemtools import get_shared_lib_ext
+from easybuild.tools.utilities import trace_msg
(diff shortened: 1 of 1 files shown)


lexming (commented):
Similar comment to #4892:
* what is the output on a failed command? I still want to know workdir and output in that case.

boegel:
I think we need to do a better job on error reporting for failing sanity check commands, you now never get the (direct) path to the relevant `cmd.sh` (but you can guess it from the trace output, sure).

I'd argue that for `pip check` you only want to see the output if it fails (the working directory is not relevant, since it's a global check which doesn't pick up on files in the working directory).

This isn't really changed here since if `pip check` fails, an `EasyBuildError` will be raised that includes the output of `pip check` in full.

lexming (approved):
LGTM

lexming:
Merging, thanks @boegel !

PR #3743, opened 2025-05-28
Make sure rpath wrappers shipped via buildenv are world executable

ocaisa:
Rpath wrappers are only executable by the user (by default)


boegel (approved):
lgtm

PR #3746, opened 2025-05-28
Deprecate the Clang Easyblock in favor of the LLVM one for newer (>=18.1.6) versions of LLVM

Crivella:
The LLVM Easyblock should replace the Clang one as it supports building Clang + other subproject and has been extensively reworked to ensure the proper testing, interplay between projects, and newer subproject/project structure introduced in LLVM >= 19, in particular related to offloading

edit: need to make sure that Clang 18.1.8 easyconfigs no longer use the Clang easyblock, so requires merge of:
* https://github.com/easybuilders/easybuild-easyconfigs/pull/23055

### DIFF
--- a/easybuild/easyblocks/c/clang.py
+++ b/easybuild/easyblocks/c/clang.py
@@ -124,7 +124,7 @@ class EB_Clang(CMakeMake):
         if LooseVersion(self.version) >= LooseVersion('19'):
             raise EasyBuildError(
                 "The Clang EasyBlock has been deprecated and does not support LLVM versions >= 19. "
-                "Please use the 'LLVM' EasyBlock instead, which supports building Clang as well"
+                "Please use the 'LLVM' EasyBlock instead, which supports building Clang as well "
                 "as other LLVM projects."
             )
 
--- a/easybuild/easyblocks/c/clang.py
+++ b/easybuild/easyblocks/c/clang.py
@@ -121,9 +121,9 @@ class EB_Clang(CMakeMake):
 
         super().__init__(*args, **kwargs)
 
-        if LooseVersion(self.version) >= LooseVersion('19'):
+        if LooseVersion(self.version) >= LooseVersion('18.1.6'):
             raise EasyBuildError(
-                "The Clang EasyBlock has been deprecated and does not support LLVM versions >= 19. "
+                "The Clang EasyBlock has been deprecated and does not support LLVM versions >= 18.1.6. "
                 "Please use the 'LLVM' EasyBlock instead, which supports building Clang as well "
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/c/clang.py
```diff
@@ -120,6 +120,14 @@ def __init__(self, *args, **kwargs):
         """Initialize custom class variables for Clang."""
 
         super().__init__(*args, **kwargs)
+
+        if LooseVersion(self.version) >= LooseVersion('19'):
+            raise EasyBuildError(
+                "The Clang EasyBlock has been deprecated and does not support LLVM versions >= 19. "
+                "Please use the 'LLVM' EasyBlock instead, which supports building Clang as well"
+                "as other LLVM projects."
+            )
```
Thyre:
```suggestion
            raise EasyBuildError(
                "The Clang EasyBlock has been deprecated and does not support LLVM versions >= 19. "
                "Please use the 'LLVM' EasyBlock instead, which supports building Clang as well "
                "as other LLVM projects."
            )
```

Crivella:
Main thing I am not sure is if we want to merge this before or after we start migrating testing all software that uses Clang as a dependency with LLVM instead

Crivella:
This https://github.com/easybuilders/easybuild-easyblocks/compare/3faa418fd4784e8b1ffbc4ce0729dd8efcad9e39..aac5947b5809c99e02276d387951baf3451e924c got lost in a rebase gonna reapply the change sorry

Flamefire:
I would go as far as using 18.1.6 as the cutoff which is the first version that can be used with the LLVM easyblock. I'm not even sure why we have this seemingly arbitrary version there. As far as I can tell it would work fine with LLVM 18 already and in fact actually fixes a bug: The Clang easyblock for LLVM 18 installs all GPU architectures instead of the selected ones. See https://github.com/easybuilders/easybuild-easyconfigs/pull/23055 which can be considered a companion PR for this one

Crivella:
> Sorry that was a bit misleading in this context. I was referring to the arbitrary looking restriction to a specific patch version in the LLVM easyblock. As far as I can tell it would work fine with LLVM 18 already

Mostly because i think `18.1.6` was the latest version when i started working on this and that was what i was heavily testing at the beginning.

boegel:
Based on test reports in https://github.com/easybuilders/easybuild-easyconfigs/pull/23055, it seems like this will require more work, so this won't make it in for EasyBuild v5.1.1

Crivella:
@boegel I think following the merge of

- https://github.com/easybuilders/easybuild-easyconfigs/pull/23055

This is ready for review/merge

PR #3717, opened 2025-05-13
Fix Python shebangs for Python and Conda-like modules

Flamefire:
The core Python modules have scripts too that use `python`. However their shebangs use absolute paths to the installed python binary.
This causes serious problems with virtualenvs as e.g `pip install` would still use the core Python binary instead of the virtualenv binary leading to failures if `$PIP_REQUIRE_VIRTUALENV` is set.

Copy over the default shebang-fix from PythonPackage.

PR #3760, opened 2025-06-05
revert Python shebangs change for conda

branfosj:
Revert the shebang change in #3717 as it breaks installs.

### DIFF
--- a/easybuild/easyblocks/a/anaconda.py
+++ b/easybuild/easyblocks/a/anaconda.py
@@ -33,7 +33,6 @@ import os
 import stat
 
 from easybuild.easyblocks.generic.binary import Binary
-from easybuild.framework.easyconfig import CUSTOM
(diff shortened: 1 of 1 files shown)


jfgrimm:
@branfosj any more detail on exactly how it breaks installs?

branfosj:
Yes ^ was the issue we were seeing.

Specifically we saw it with
```
module load Miniforge3/24.11.3-0
eval "$(${EBROOTMINIFORGE3}/bin/conda shell.bash hook)"
source "${EBROOTMINIFORGE3}/etc/profile.d/mamba.sh"
conda create --yes --prefix /tmp/condaenv python=3.9
conda activate /tmp/condaenv
```

Then
```
$ conda --help
  File "/home/simon/work/eb/installpath/software/Miniforge3/24.11.3-0/bin/conda", line 12, in <module>
    from conda.cli import main
ModuleNotFoundError: No module named 'conda'
```

PR #3769, opened 2025-06-12
Fix start_dir for components of Bundle easyblock

Flamefire:
(created using `eb --new-pr`)

I tried using a patch for a bunde component but it failed because the start_dir is wrong. It will always use `build_dir` (or a path relative to it if set in `start_dir` EC parameter) because at the point of `guess_startdir` there are **no** sources in `self.src` hence it can't be relative to the unpacked source.

The comment and code don't add up too:
```
# location of first unpacked source is used to determine where to apply patch(es)
comp.src[-1]['finalpath'] = comp.cfg['start_dir']
```

Why does it set the **last** sources final path to the start_dir?

So I'd say the only correct way would be to set `comp.src` correctly and call `guess_start_dir` afterwards. Touching a sources `finalpath` can never be right ever, can it?

Edit: Further investigation shows that `finalpath` isn't set correctly for multiple sources. What happens for bundles of multiple components with a single source:

- Start with an empty builddir
- Extract source of 1st component -> Only folder -> `finalpath` set to that folder
- Extract source of 2nd component -> 2 folders -> Parent folder (builddir) is used as `finalpath`
- When setting `component2.src = src2` and then calling `guess_start_dir` it will find the source and use its  `finalpath` as the start_dir, hence wrongly using the builddir
- If `start_dir` is manually set in the easyconfig it will still find the correct start_dir
- However `apply_patches` uses the `finalpath` of the source at an index given in the patch spec, defaulting to 0
- Hence it will try to apply the patch in the build dir instead of the component dir

So this requires either https://github.com/easybuilders/easybuild-framework/pull/4922 or `comp.src[0]['finalpath'] = comp.start_dir` after the `comp.guess_start_dir` so that patches work correctly.

Issue #3680, opened 2025-03-31
LLVM EasyBlock fails when used as a Bundle component


Crivella:
Currently the EasyBlock accesses the `installdir` and `builddir` variable at the `__init__` causing this values to not be properly set when used as a `Bundle` compoent

Thanks to @bedroge  for the report and investigation
https://github.com/easybuilders/easybuild-easyblocks/blob/develop/easybuild/easyblocks/generic/bundle.py#L261

Fix will require moving all configurations pertaining to this variable to be moved in the `configure_step`

bedroge:
I'm probably running into another issue now due to using this in a Bundle: while testing things with ROCm-LLVM 6.4.0, which is based on LLVM 19 (older versions were based LLVM 18), I'm suddenly getting the following error which I didn't get for older versions:
```
== FAILED: Installation ended unsuccessfully: GCC installation path '{check_prefix}' does not match expected path '{gcc_prefix}' (took 44 mins 2 secs)
```

It looks like the `<compiler>.cfg` files are not being created for the final installation, and I think it's because the `post_processing_step` is not being called for bundle components. Not sure if this can be fixed easily, also because there's some more functionality in that same function, but otherwise I can probably work around it quite easily.

Crivella:
If i recall correctly that used to be a `post_install_step` that got deprecated in 5.x.
Since i believe that step is part of the installation, it might be worth trying moving the logic from `post_processing_step` to `install_step` after calling `super(EB_LLVM, self).install_step()`

Flamefire:
This might be caused by a wrongly set `start_dir` when used in a bundle with multiple components. See #3769

PR #3770, opened 2025-06-12
Use `self.start_dir` in LLVM easyblock

Flamefire:
`self.cfg['start_dir']` gets initialized in `guess_start_dir` and hence is always set to an absolute path, so non-empty.
Note that `path.join(anything, abs-path) == abs-path`

So `llvm_src_dir` is always equal to `self.start_dir` and we can remove this property.

@Crivella am I missing anything?

I'm also wondering about the src extension in `llvm-project-%s.src`. Haven't seen that in my tests anywhere and it likely didn't cause issues due to the above which means that code path wasn't used.

### DIFF
--- a/easybuild/easyblocks/l/llvm.py
+++ b/easybuild/easyblocks/l/llvm.py
@@ -462,6 +462,12 @@ class EB_LLVM(CMakeMake):
         self._cmakeopts = {}
         self._cfgopts = list(filter(None, self.cfg.get('configopts', '').split()))
 
+    @property
+    def llvm_src_dir(self):
+        """Return root source directory of LLVM (containing all components)"""
+        # LLVM is the first source so we already have this in start_dir. Might be changed later
+        return self.start_dir
+
     def prepare_step(self, *args, **kwargs):
         """Prepare step, modified to ensure install dir is deleted before building"""
         super(EB_LLVM, self).prepare_step(*args, **kwargs)


Crivella:
Umh if `start_dir` is always set i guess the else part would not be used

> I'm also wondering about the src extension in llvm-project-%s.src

All the sources are of the type `llvm-project-XXX.src.tar.xz` and gets uncompressed to the `llvm-project-XXX.src` in the build directory (Not sure if you are seeing a different behavior)

> Note that path.join(anything, abs-path) == abs-path

We could add a check to be more transparent about this behavior and not do the join in case of an abspath.
I think we still need the join as the starting directory could be different from the base one in case of other builds like the ROCM-llvm
@bedroge can probably comment on this

Flamefire:
> Umh if `start_dir` is always set i guess the else part would not be used

It is and even checked for existence: https://github.com/easybuilders/easybuild-framework/blob/87b3d9d959b5f3ec739f3db4a350982b531c58fb/easybuild/framework/easyblock.py#L2299-L2338

> All the sources are of the type `llvm-project-XXX.src.tar.xz` and gets uncompressed to the `llvm-project-XXX.src` in the build directory (Not sure if you are seeing a different behavior)

Ah, the release sources yes. I was using a specific commit (required for Triton) where that extension is missing.

> 
> > Note that path.join(anything, abs-path) == abs-path
> 
> We could add a check to be more transparent about this behavior and not do the join in case of an abspath. I think we still need the join as the starting directory could be different from the base one in case of other builds like the ROCM-llvm @bedroge can probably comment on this

`start_dir` is always absolute, at least now. Do you know a situation where either `llvm*.tar.gz` is NOT the first source of the easyconfig/component or where you want a different folder?   
I found a bug which might be related, or the actual cause, of https://github.com/easybuilders/easybuild-easyblocks/issues/3680 for which you designed the fix: For Bundle-components `start_dir` wasn't set correctly and always set to `builddir`. So I guess that might have been why this piece of code was required.

I'm fixing that in https://github.com/easybuilders/easybuild-easyblocks/pull/3769

Flamefire:
After some more work on the Bundle easyblock I don't see a way to fix it without breaking existing easyconfigs/easyblocks.

But it still stands: `self.cfg['start_dir']` is always an absolute path at this point, so this PR doesn't introduce an effective change, and just cleans up code that might lead to misunderstandings

Flamefire:
Test report by @Flamefire
'Failed to process easyconfig /home/s3248973/.local/EasyBuildDev/easybuild-easyconfigs/easybuild/easyconfigs/l/LLVM/LLVM-14.0.3-GCCcore-11.3.0.eb: You may have some typos in your easyconfig file: use_polly -> usepolly'
Failed during parsing of the easyconfigs, so no ecs were built (5 easyconfigs in total)
i7007 - Linux Rocky Linux 8.9 (Green Obsidian), x86_64, AMD EPYC 7702 64-Core Processor (zen2), Python 3.9.18
See https://gist.github.com/Flamefire/107c2e420c8bf07b3acb4d023188d6a7 for a full test report.

Thyre:
> BTW: I haven't found any easyconfigs using LLVM in components. Do you know any?

ROCm-LLVM will, see:

https://github.com/Thyre/easybuild-custom/blob/support-passing-amdgcn/easybuild/easyconfigs/r/ROCm-LLVM/ROCm-LLVM-6.4.1-GCCcore-14.2.0.eb

This order is required, due to the weird handling of dependencies by AMD. 

CC @bedroge

Flamefire:
I'm convinced that would still work: You use the (usual) workaround of manually setting `start_dir` (because the Bundle easyblock doesn't):
https://github.com/Thyre/easybuild-custom/blob/8d141968038f4fa48026f09f284e80969b1b2ca7/easybuild/easyconfigs/r/ROCm-LLVM/ROCm-LLVM-6.4.1-GCCcore-14.2.0.eb#L66

So it will be set by the point it gets to the `if` here and the `else` is also not used nor required

Flamefire:
Test report by @Flamefire
 * **SUCCESS** _LLVM-14.0.3-GCCcore-11.3.0.eb_ 
 * **SUCCESS** _LLVM-16.0.6-GCCcore-12.3.0.eb_ 
 * **SUCCESS** _LLVM-16.0.6-GCCcore-13.2.0.eb_ 
 * **FAIL (build issue)** _LLVM-20.1.4-GCCcore-13.3.0.eb_ (partial log available at https://gist.github.com/Flamefire/7e9bfec627a06e30bd66c7c9ace5dcd7)

Build succeeded for 4 out of 5 (5 easyconfigs in total)
i7016 - Linux Rocky Linux 8.9 (Green Obsidian), x86_64, AMD EPYC 7702 64-Core Processor (zen2), Python 3.9.18
See https://gist.github.com/Flamefire/38d79f8eaee632170a77e4f8042db58f for a full test report.

Flamefire:
Test report by @Flamefire
 * **SUCCESS** _LLVM-20.1.4-GCCcore-13.3.0.eb_ 

i7008 - Linux Rocky Linux 8.9 (Green Obsidian), x86_64, AMD EPYC 7702 64-Core Processor (zen2), Python 3.9.18
See https://gist.github.com/Flamefire/48d270a6b47f4881b9d0b7d9607e5485 for a full test report.

Crivella:
For me I am fine with the `if/else` being removed but i would leave the `llvm_src_dir` attribute as is, as it is already being used [in other derived easyblocks](https://github.com/bedroge/eb-rocm/blob/main/6.4.0/rocm_llvm.py) (even if not upstream), and in general i think it is good to keep it separate from `start_dir` as it makes clearer that attribute points to the source directory of the llvm project (especially if in the future we might want to add some third party project/sources to the build).

Crivella:
> I'm not sure if the first commit (https://github.com/easybuilders/easybuild-easyblocks/commit/f710d05c4085a22d3865b702d42e4e546de50497) can have any side-effects. I remember @Crivella adding them explicitly in https://github.com/easybuilders/easybuild-easyblocks/commit/94b960be643a2fda60067a73d0fbad2c5ca98d25, but I don't remember why.

That can be removed, at some point i also added it to being called inside `_create_compiler_config_file` for safety but did not remove the calls outside 

https://github.com/easybuilders/easybuild-easyblocks/blob/f710d05c4085a22d3865b702d42e4e546de50497/easybuild/easyblocks/l/llvm.py#L838-L840

Crivella:
> That was what I thought. I'd even go further and use a property which initializes itself on access so we never end up reading None when we forget a call to _set_gcc_prefix

I think that would be a good idea, but i would leave that for a separate PR

Crivella:
EB_ARGS="--installpath /tmp/$USER/ebpr-3770 LLVM-20.1.5-GCCcore-13.3.0.eb"
CORE_CNT=16

Crivella:
EB_ARGS="--installpath /tmp/$USER/ebpr-3770 LVM-16.0.6-GCCcore-12.3.0.eb"

PR #3771, opened 2025-06-12
Fix test step for CMakeNinja easyblock by customizing `test_cmd` and `runtest` easyconfig parameters

jngrad:
Use `ninja` instead of `make` as the default command to run the test target in a project built by the `CMakeNinja` class.

### DIFF
--- a/easybuild/easyblocks/generic/cmakeninja.py
+++ b/easybuild/easyblocks/generic/cmakeninja.py
@@ -32,6 +32,8 @@ CMake configure step and Ninja build install.
 from easybuild.easyblocks.generic.cmakemake import CMakeMake
 from easybuild.easyblocks.generic.mesonninja import MesonNinja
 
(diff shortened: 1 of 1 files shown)


jngrad:
This issue was revealed by https://github.com/multixscale/dev.eessi.io-espresso/pull/1. Without this patch, one has to add `test_cmd = 'ninja'` to the `.eb` file to avoid an error message like `make: *** No rule to make target 'test'` when using `CMakeNinja`.

Micket (approved):
lgtm, got any test reports coming in?

boegel:
@jngrad See https://docs.easybuild.io/integration-with-github/#github_upload_test_report

That takes a bit of setting up though...

Test reports are definitely welcome, but we can also take care of that to make sure these changes don't cause trouble

boegel (approved):

PR #3748, opened 2025-05-30
Also run test step in a bundle component

ocaisa:

PR #3775, opened 2025-06-14
Reset `runtest` for components in a bundle

ocaisa:
Follow up to #3748 for the case where it is have been shown to be problematic (components in a Pythonbundle)

### DIFF
--- a/easybuild/easyblocks/generic/bundle.py
+++ b/easybuild/easyblocks/generic/bundle.py
@@ -39,6 +39,7 @@ import os
 import easybuild.tools.environment as env
 from easybuild.framework.easyblock import EasyBlock
 from easybuild.framework.easyconfig import CUSTOM
+from easybuild.framework.easyconfig.default import DEFAULT_CONFIG
 from easybuild.framework.easyconfig.easyconfig import get_easyblock_class
 from easybuild.tools.build_log import EasyBuildError, print_msg
 from easybuild.tools.config import build_option
@@ -130,7 +131,13 @@ class Bundle(EasyBlock):
                 # which is not a valid value for many easyblocks.
                 # Reset runtest to the original default, if people want the test step
                 # they can set it explicitly.
-                comp_cfg['runtest'] = None
+                if comp_cfg._config['runtest'] != DEFAULT_CONFIG["runtest"]:
+                    self.log.warning(
+                        "Resetting runtest to default value for component easyblock "
+                        f"(from {comp_cfg._config['runtest']})."
+                        )
+                    comp_cfg["runtest"] = DEFAULT_CONFIG["runtest"][0]
+                    comp_cfg._config['runtest'] = DEFAULT_CONFIG["runtest"]
 
                 # determine easyblock to use for this component
                 # - if an easyblock is specified explicitly, that will be used
--- a/easybuild/easyblocks/generic/bundle.py
+++ b/easybuild/easyblocks/generic/bundle.py
@@ -130,13 +130,12 @@ class Bundle(EasyBlock):
                 #  runtest = True
                 # which is not a valid value for many easyblocks.
                 # Reset runtest to the original default, if people want the test step
-                # they can set it explicitly (or it can be set by the component easyblock).
+                # they can set it explicitly, in default_component_specs or by the component easyblock
                 if comp_cfg._config['runtest'] != DEFAULT_CONFIG["runtest"]:
                     self.log.warning(
                         "Resetting runtest to default value for component easyblock "
                         f"(from {comp_cfg._config['runtest']})."
                         )
-                    comp_cfg["runtest"] = DEFAULT_CONFIG["runtest"][0]
                     comp_cfg._config['runtest'] = DEFAULT_CONFIG["runtest"]
 
                 # determine easyblock to use for this component
--- a/easybuild/easyblocks/generic/bundle.py
+++ b/easybuild/easyblocks/generic/bundle.py
@@ -130,7 +130,7 @@ class Bundle(EasyBlock):
                 #  runtest = True
                 # which is not a valid value for many easyblocks.
                 # Reset runtest to the original default, if people want the test step
-                # they can set it explicitly.
+                # they can set it explicitly (or it can be set by the component easyblock).
                 if comp_cfg._config['runtest'] != DEFAULT_CONFIG["runtest"]:
                     self.log.warning(
                         "Resetting runtest to default value for component easyblock "
--- a/easybuild/easyblocks/generic/bundle.py
+++ b/easybuild/easyblocks/generic/bundle.py
@@ -124,8 +124,16 @@ class Bundle(EasyBlock):
                 comp_cfg['name'] = comp_name
                 comp_cfg['version'] = comp_version
 
+                # The copy above may include unexpected settings for common values.
+                # In particular a Pythonbundle for we have seen a component inheriting
+                #  runtest = True
+                # which is not a valid value for many easyblocks.
+                # Reset runtest to the original default, if people want the test step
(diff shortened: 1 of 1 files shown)


ocaisa:
EB_ARGS="purge_dups-1.2.6-foss-2023b.eb"

Flamefire:
I don't think this is the right solution. [PythonPackage has a default](https://github.com/easybuilders/easybuild-easyblocks/blob/2238d46cbdb1c3515a6800f6c0ed8f57d4a1712e/easybuild/easyblocks/generic/pythonpackage.py#L417) of `runtest = True` which is valid.

So I think if you set this to `None` it would change the default, specifically set (see `FooBundle` below).

The problem IMO is (only?) `PythonBundle` [changing defaults ](https://github.com/easybuilders/easybuild-easyblocks/blob/2238d46cbdb1c3515a6800f6c0ed8f57d4a1712e/easybuild/easyblocks/generic/pythonbundle.py#L54) and they are (IMO correctly) propagated to components. With #3748 the `runtest` setting is now honored for components causing issues in Easyconfigs not prepared for that.

I think it might be better to **remove** the entry if present so components use their easyblock specific defaults.   

In general it makes me wonder if this is the correct approach: Why would we reset/change a single property but not any other? What makes `runtest` special except that it is now honored?   
IMO it is rather the easyconfigs that need changing/updating.

Consider for example a `class FooBundle(Bundle)` that sets `runtest=True` as the default to have it set for all components but with this change this is no longer possibly, is it?

Maybe we should rather only copy specific entries from the parent config and let `default_component_specs` handle the rest. Of course this will likely break other easyconfigs not prepared for that. I guess `PythonBundle` would work fine as it copies it's own settings to `exts_default_options` so components are left alon

ocaisa:
I don't think what you're saying is exactly correct, this doesn't change the default values of the parent easyblock, it changes what is inherited by the components. The default value for runtest is None, as set in framework. All this does is restore that default, the easyblock used by the components can still override that, as can also be done by `default_component_specs`.

ocaisa:
I could check it's value and explicitly log that I'm doing the reset, with advice about how to work around it?

ocaisa:
> > I don't think what you're saying is exactly correct, this doesn't change the default values of the parent easyblock, it changes what is inherited by the components. The default value for runtest is None, as set in framework. All this does is restore that default, the easyblock used by the components can still override that, as can also be done by `default_component_specs`.
> 
> IIRC on construction of an easyblock instance from an easyconfig it will only take the defaults if the easyconfig doesn't specify a value. But now you do specify a value: `None`. So if you have a `PythonPackage` component `runtest` will no longer default to `True` but `None` as set here. So component easyblocks can't overwrite the default via `extra_options` anymore. Or am I missing anything?
> 

Oh, you might be right about that, I should actually check. If removing it instead has the desired effect I'm fine with that. I am just afraid that that may also have unintended consequences.

Flamefire:
> Oh, you might be right about that, I should actually check.

Should be easy to check. And I guess there aren't many easyblocks changing the default so the (remaining) impact (of the now run test step) should be low.

> I could check it's value and explicitly log that I'm doing the reset, with advice about how to work around it?

Might still be a good idea. How many Bundle-inherited easyblocks do we have changing the default of `runtest`? And how would the log detect the difference of intended vs unintended? I.e. how can we change `PythonBundle` to not trigger that message?   
I have no idea so I'd add a comment here that this is a workaround for existing easyconfigs after a newly introduced change and in EB 6 only copy a couple options, if any at all.

boegel (approved):
lgtm

boegel:
Test report by @boegel
 * **SUCCESS** _FlexiBLAS-3.4.5-GCC-14.2.0.eb_ 
 * **SUCCESS** _matplotlib-3.8.2-gfbf-2023b.eb_ 

node3611.doduo.os - Linux RHEL 9.4, x86_64, AMD EPYC 7552 48-Core Processor (zen2), Python 3.9.18
See https://gist.github.com/boegel/a74cf3da82b5655dcd7f32278593a5bc for a full test report.

ocaisa:
@Flamefire I've updated this a bit and tested it with some print statements when setting the `default_component_easyblock` to `PythonPackage`:
```
== processing EasyBuild easyconfig /home/ocaisa/EasyBuild/easybuild-easyblocks/easybuild/easyblocks/generic/purge_dups-1.2.6-foss-2023b.eb
== Running parse hook...
== Running parse hook...
INITIAL_VALUE._config = [True, 'Run unit tests.', (1, 'easyblock-specific')]
INITIAL_VALUE = True
RESET_VALUE._config = [None, 'Indicates if a test should be run after make; should specify argument after make (for e.g.,"test" for make test)', (3, 'build')]
RESET_VALUE = None
FINAL_VALUE._config = [True, 'Run unit tests.', (1, 'easyblock-specific')]
FINAL_VALUE = True
ERROR: Failed to get application instance for purge_dups (easyblock: PythonBundle): Use of unknown easyconfig parameter 'files_to_copy' when setting parameter value to '['bin', 'scripts']'
```

This works as much as I think it should. It still allows the component easyblock to make default changes to the value of `runtest`.

easybuild/easyblocks/generic/bundle.py
```diff
@@ -124,8 +125,22 @@ def __init__(self, *args, **kwargs):
                 comp_cfg['name'] = comp_name
                 comp_cfg['version'] = comp_version
 
+                # The copy above may include unexpected settings for common values.
+                # In particular for a Pythonbundle we have seen a component inheriting
+                #  runtest = True
+                # which is not a valid value for many easyblocks.
+                # Reset runtest to the original default, if people want the test step
+                # they can set it explicitly.
```
ocaisa:
```suggestion
                # they can set it explicitly (or it can be set by the component easyblock).
```

easybuild/easyblocks/generic/bundle.py
```diff
@@ -124,8 +125,22 @@ def __init__(self, *args, **kwargs):
...
+                # which is not a valid value for many easyblocks.
+                # Reset runtest to the original default, if people want the test step
+                # they can set it explicitly (or it can be set by the component easyblock).
+                if comp_cfg._config['runtest'] != DEFAULT_CONFIG["runtest"]:
+                    self.log.warning(
+                        "Resetting runtest to default value for component easyblock "
+                        f"(from {comp_cfg._config['runtest']})."
+                        )
+                    comp_cfg["runtest"] = DEFAULT_CONFIG["runtest"][0]
+                    comp_cfg._config['runtest'] = DEFAULT_CONFIG["runtest"]
```
Flamefire: The 2nd overwrites the first so it can be removed

```suggestion
                    comp_cfg._config['runtest'] = DEFAULT_CONFIG["runtest"]
```

See:
```
    def __setitem__(self, key, value):
    ...
            self._config[key][0] = value
```

easybuild/easyblocks/generic/bundle.py
```diff
@@ -124,8 +125,22 @@ def __init__(self, *args, **kwargs):
                 comp_cfg['name'] = comp_name
                 comp_cfg['version'] = comp_version
 
+                # The copy above may include unexpected settings for common values.
+                # In particular for a Pythonbundle we have seen a component inheriting
+                #  runtest = True
+                # which is not a valid value for many easyblocks.
+                # Reset runtest to the original default, if people want the test step
+                # they can set it explicitly (or it can be set by the component easyblock).
```
Flamefire: Shall we mention default_component_specs for replacing `runtest = True` at the toplevel?

```suggestion
                # they can set it explicitly, in default_component_specs or by the component easyblock
```

boegel (approved):

PR #3352, opened 2024-06-07
Patch `ctypes` for Python installations when filtering `LD_LIBRARY_PATH`

ocaisa:
`ctypes` relies on `LD_LIBRARY_PATH` and doesn't respect rpath linking, this is a workaround for the EasyBuild context for this setup.

PR #3776, opened 2025-06-16
lammps sanity checks: remove LD_LIBRARY_PATH setting

lorisercole:
With the `foss-2024a` toolchain (CUDA 12.6+), prepending `LD_LIBRARY_PATH` with `LIBRARY_PATH` causes LAMMPS to crash, likely because the dynamic linker gets pointed to the CUDA stubs libs instead of the system libs. Strangely, this is not a problem for older toolchains. At the same time, it appears that setting `LD_LIBRARY_PATH` has no effect on the functionality of the code (it seems that Python is able to retrieve `liblammps.so` without it), so I propose we just get rid of it, unless there are actual scenarios where it causes problems.

### DIFF
--- a/easybuild/easyblocks/l/lammps.py
+++ b/easybuild/easyblocks/l/lammps.py
@@ -572,13 +572,7 @@ class EB_LAMMPS(CMakeMake):
         if self.toolchain.options.get('usempi', None):
             custom_commands = [self.toolchain.mpi_cmd_for(cmd, 1) for cmd in custom_commands]
 
-        # Requires liblammps.so to be findable by the runtime linker (which it might not be if using
-        # rpath and filtering out LD_LIBRARY_PATH)
-        set_ld_library_path = ''
-        if self.installdir not in os.getenv('LD_LIBRARY_PATH', default=''):
-            # Use LIBRARY_PATH to set it
-            set_ld_library_path = "LD_LIBRARY_PATH=$LIBRARY_PATH:$LD_LIBRARY_PATH "
-        custom_commands = ["cd %s && " % execution_dir + set_ld_library_path + cmd for cmd in custom_commands]
+        custom_commands = ["cd %s && " % execution_dir + cmd for cmd in custom_commands]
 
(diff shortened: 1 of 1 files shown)


ocaisa:
I'd agree with this, this is no longer needed due to https://github.com/easybuilders/easybuild-easyblocks/pull/3352

casparvl:
I'm with Alan on this. The fact you can't reproduce the original issue is likely because indeed it was resolved in the meantime, likely by https://github.com/easybuilders/easybuild-easyblocks/pull/3352 . And even if it isn't, I'd say it's a `ctypes` issue and should be fixed there - not in the EasyBlock like this. Especially not since it's causing other issues.

In the `ctypes` fix, we can consider if it's needed to filter e.g. any prefix containing `stubs` as a subdir... It'd again be a big hammer, but not all that different from what we do in our RPATH wrappers (I think).

Crivella:
EB_ARGS="--installpath=/tmp/$USER/ebpr-3776   LAMMPS-29Aug2024_update2-foss-2023a-kokkos.eb LAMMPS-2Aug2023_update2-foss-2023a-kokkos-CUDA-12.1.1.eb LAMMPS-23Jun2022-foss-2022a-kokkos.eb LAMMPS-29Aug2024-foss-2023b-kokkos.eb"
CORE_CNT=16

Crivella:
I would say in conjunction with https://github.com/easybuilders/easybuild-easyconfigs/pull/23117#issuecomment-2980782865 this has been extensively tested, not sure what happened with the CI, some problem with the GH runners?

lorisercole:
> I would say in conjunction with [easybuilders/easybuild-easyconfigs#23117 (comment)](https://github.com/easybuilders/easybuild-easyconfigs/pull/23117#issuecomment-2980782865) this has been extensively tested, not sure what happened with the CI, some problem with the GH runners?

Thanks. Not sure but it seems to be a temporary problem with the CI. Perhaps you can rerun them manually?

Crivella (approved):
LGTM

PR #3778, opened 2025-06-16
Make matching component sources in Bundle easyblock more reliable

Flamefire:
Store where sources have been added to get the corresponding filled structure without relying on partially resolved templates.

Also do not fetch component patches twice

I've also move some stuff around to make the structure clearer:

- `fetch_step` now fetches patches instead of doing that (again) in the install step
- Set properties like dirs and parallel in prepare step as done in `EasyBlock`
- Set the finalpath of the first source as stated by the comment. Most (all?) components only have a single source so this shouldn't cause any issues.

Note that this manual setting of `finalpath` isn't logically correct as `start_dir` of a component could be the builddir or the extracted source of some component

`guess_start_dir` will always get the builddir (due to no sources present), potentially with a suffix as specified in the easyconfigs which many do (as `foo-1.2.3`) because the extracted source isn't available and the `finalpath` for the 2nd component and all following is wrong until https://github.com/easybuilders/easybuild-framework/pull/4922

But I don't see how to fix this without breaking existing easyconfigs and easyblocks.

### DIFF
--- a/easybuild/easyblocks/generic/bundle.py
+++ b/easybuild/easyblocks/generic/bundle.py
@@ -105,6 +105,10 @@ class Bundle(EasyBlock):
         # (like adding component sources to top-level sources easyconfig parameter)
         self.cfg = self.cfg.copy()
 
+        # Keep track of sources for each component to restore them:
+        # (Component instance, start idx [in self.src], end idx)
+        self.comp_to_src_idxs = []
+
         # disable templating to avoid premature resolving of template values
         # Note that self.cfg.update also resolves templates!
         with self.cfg.disable_templating():
@@ -197,6 +201,10 @@ class Bundle(EasyBlock):
 
                 comp_cfg.generate_template_values()
 
+                # Combine all component sources into the top-level sources parameter
+                # This allows reusing top-level source_urls and unpacking them all in the extract_step
+
+                old_num_srcs = len(self.cfg.get('sources', resolve=False))
                 # Don't require that all template values can be resolved at this point but still resolve them.
                 # This is important to ensure that template values like %(name)s and %(version)s
                 # are correctly resolved with the component name/version before values are copied over to self.cfg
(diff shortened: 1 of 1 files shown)


Thyre:
Regarding

> Set the finalpath of the first source as stated by the comment. Most (all?) components only have a single source so this shouldn't cause any issues

https://github.com/easybuilders/easybuild-easyconfigs/pull/21582 sets two sources for the Intel compilers. However, if I understand correctly, `final_path` is mostly used for patches, which this PR doesn't have. I'll test this PR with your changes later.

Flamefire:
> [easybuilders/easybuild-easyconfigs#21582](https://github.com/easybuilders/easybuild-easyconfigs/pull/21582) sets two sources for the Intel compilers. However, if I understand correctly, `finalpath` is mostly used for patches, which this PR doesn't have. I'll test this PR with your changes later.

Yes it is used for patches and for `guess_start_dir`. The Bundle easyblock (wrongly) doesn't use it for the latter. Patches can specify a `source index` to apply in a different sources finalpath. But I think this is broken for everything until https://github.com/easybuilders/easybuild-framework/pull/4922 as the 2nd sources finalpath will be the builddir. So I doubt it's used

I feel like we need a new Bundle easyblock which fixes all the quirks of the current one:
- `finalpath`s not determined correctly
- `start_dir` of components relative to builddir instead of extracted source
- skipping of multiple, possibly important, steps like `prepare_step`

It looks like you can just put an easyconfig into a component and it would work as-if standalone but it doesn't in subtile ways due to this. And I don't see a way to fix it which would be backwards compatible.

So this PR just reduces chances for one surprise

Thyre:
I think we're slowly but surely finding out which parts of the Bundle EasyBlock are broken. I agree, that maybe starting a new Bundle EasyBlock rewriting everything from the ground up might be needed to fix all of these lingering issues. This will take quite some time, and until then, trying to fix or work around the existing issues is a good idea...

Flamefire:
Yes that was mispasted in that PR.

PR #3787, opened 2025-06-18
handle PATH environment variable with `module_load_environment` in init of Binary easyblock

lexming:
This went through the cracks with the introduction of `module_load_environment` in EB5


ocaisa (approved):
LGTM

PR #3788, opened 2025-06-18
refactor NVHPC easyblock into NvidiaBase, nvidia-compilers and NVHPC

lexming:


lexming:
Test report by @lexming
 * **SUCCESS** _NVHPC-22.11-CUDA-11.7.0.eb_ 
 * **SUCCESS** _NVHPC-23.7-CUDA-12.1.1.eb_ 
 * **SUCCESS** _NVHPC-24.7-CUDA-12.6.0.eb_ 

node250.hydra.os - Linux Rocky Linux 9.5 (Blue Onyx), x86_64, Intel(R) Xeon(R) CPU E5-2650 v4 @ 2.20GHz, 1 x NVIDIA Tesla P100-PCIE-16GB, 570.158.01, Python 3.9.21
See https://gist.github.com/lexming/0f04145f88e30abf70ff640ae6e6f103 for a full test report.

easybuild/easyblocks/generic/nvidiabase.py
```diff
@@ -0,0 +1,577 @@
...
+                os.path.join(abs_install_subdir, 'comm_libs',  'hpcx*'),
+                os.path.join(abs_install_subdir, 'comm_libs', self.CUDA_VERSION_GLOB, 'hpcx*'),
+            ]
+            remove([dir_path for dir_glob in mpi_dir_globs for dir_path in glob.glob(dir_glob)])
+            remove(os.path.join(abs_install_subdir, 'comm_libs',  'mpi'))
+        if not self.cfg['module_add_nccl']:
+            nccl_dir_glob = os.path.join(abs_install_subdir, 'comm_libs', self.CUDA_VERSION_GLOB, 'nccl')
+            remove(glob.glob(nccl_dir_glob))
+            if LooseVersion(self.version) >= LooseVersion('25.0'):
+                remove(os.path.join(abs_install_subdir, 'comm_libs',  'nccl'))
```
Thyre: We tried to use the EasyBlock on our JSC systems to install `nvidia-compilers`. While this EasyBlock worked on _almost all_ machines, this particular `remove` failed on JUWELS. I don't know why to be honest, but for some reason this folder doesn't exist. The one above (`nccl_dir_glob`) does exist and is removed.

Running the install command again in a shell results in the following directory structure:
```console
eb-shell> ls /p/project1/cjsc/reuter1/EasyBuild/Next/easybuild/juwels/software/nvidia-compilers/25.9-CUDA-12/Linux_x86_64/25.9/comm_libs
12.9  13.0  hpcx  mpi
eb-shell> ls /p/project1/cjsc/reuter1/EasyBuild/Next/easybuild/juwels/software/nvidia-compilers/25.9-CUDA-12/Linux_x86_64/25.9/comm_libs/12.9
hpcx  nccl  nccl-2.26  nccl-2.27  nvshmem
eb-shell> ls /p/project1/cjsc/reuter1/EasyBuild/Next/easybuild/juwels/software/nvidia-compilers/25.9-CUDA-12/Linux_x86_64/25.9/comm_libs/13.0/
hpcx  nccl  nvshmem
```

I would suggest to not fail if those paths (i.e. the generic `nccl`, `mpi` and `nvshmem` ones) do not exist to begin with...

Thyre: Removing the math libraries could also pose problems, since not everything is provided by CUDA, NCCL and NVSHMEM.
With NVHPC 25.9, one could provide most things with the new environment variables. The math libraries could be a bit more complicated though, as e.g. `cuFFTmp` is a separate package not included in CUDA. However, we only have one environment variable to pass the path :see_no_evil: 

(ignoring that `cuFFTMp` doesn't support CUDA 13 yet, which would also cause issues if we e.g. want to include that in the CUDA installation).

easybuild/easyblocks/n/nvidia_compilers.py
```diff
@@ -0,0 +1,57 @@
...
+        # Unsupported NVHPC options in nvidia-compilers are forced disabled
+        disabled_nvhpc_options = [
+            'module_byo_compilers',
+            'module_nvhpc_own_mpi',
+            'module_add_nccl',
+            'module_add_math_libs',
+        ]
+        for nvhpc_opt in disabled_nvhpc_options:
+            if self.cfg[nvhpc_opt]:
+                self.log.warning(f"Option '{nvhpc_opt}' forced to disabled in {self.name}-{self.version}")
```
Thyre: I'd prefer that this immediately aborts the build instead of "silently" changing the option.
This only shows up in the logs, so people might not be aware of the changes (like I just did).

I'm also confused why we remove `NCCL` and the math libraries, but keep `NVSHMEM`?
I'd prefer to allow all three of them, but disable them by default.

easybuild/easyblocks/generic/nvidiabase.py
```diff
@@ -0,0 +1,577 @@
...
+# split this value at colons, separate by -L, prepend 1st one by -L
+variable library_path is
+default($if($LIBRARY_PATH,-L$replace($LIBRARY_PATH,":", -L)));
+
+# add the -L arguments to the link line
+append LDLIBARGS=$library_path;
+
+# also include the location where libm & co live on Debian-based systems
+# cfr. https://github.com/easybuilders/easybuild-easyblocks/pull/919
+append LDLIBARGS=-L/usr/lib/x86_64-linux-gnu;
```
Thyre, line 72: Should we be careful with this, in the case we ever want to include NVHPC (or at least modules built on top of it) in EESSI in some way?

easybuild/easyblocks/generic/nvidiabase.py
```diff
@@ -0,0 +1,577 @@
...
+                # we cannot error out here to avoid breaking existing easyconfigs
+                print_warning(
+                    f"CUDA version '{external_cuda_version}' from external CUDA might not be supported by "
+                    f"{self.name}-{self.version}: {supported_cuda_commasep}"
+                )
+            self.log.info(
+                f"Using version '{external_cuda_version}' of loaded CUDA dependency "
+                f"as 'default_cuda_version' of {self.name}"
+            )
+            return external_cuda_version
```
Thyre: We probably should adopt https://github.com/easybuilders/easybuild-easyblocks/pull/3924 here, once that is ready...

Just relying on the defined software version of the module might not be sufficient for a proper result, since sites may omit the minor and patch version, or even might use something like `default`.

easybuild/easyblocks/generic/nvidiabase.py
```diff
@@ -0,0 +1,577 @@
...
+
+        # default CUDA version set in configuration
+        default_cuda_version = self.cfg['default_cuda_version']
+        if default_cuda_version is not None:
+            if supported_cuda_versions and default_cuda_version not in supported_cuda_versions:
+                raise EasyBuildError(
+                    f"Selecte default CUDA version '{default_cuda_version}' not supported by "
```
Thyre:
```suggestion
                    f"Selected default CUDA version '{default_cuda_version}' not supported by "
```

easybuild/easyblocks/generic/nvidiabase.py
```diff
@@ -0,0 +1,577 @@
...
+                    f"{self.name}-{self.version}: {supported_cuda_commasep}"
+                )
+            return default_cuda_version
+
+        # default CUDA version undefined, pick one if obvious
+        if len(supported_cuda_versions) == 1:
+            active_cuda_version = supported_cuda_versions[0]
+            self.log.info(
+                f"Missing 'default_cuda_version' or CUDA dependency. Using CUDA version '{active_cuda_version}' "
```
Thyre:
```suggestion
                f"Missing 'default_cuda_version' or CUDA dependency. Using CUDA version '{active_cuda_version}', "
```

easybuild/easyblocks/generic/nvidiabase.py
```diff
@@ -0,0 +1,577 @@
...
+
+        run_shell_cmd(cmd)
+
+        # If an OS libnuma is NOT found, makelocalrc creates symbolic links to libpgnuma.so
+        # If we use the EB libnuma, delete those symbolic links to ensure they are not used
+        if get_software_root("numactl"):
+            for filename in ["libnuma.so", "libnuma.so.1"]:
+                path = os.path.join(compilers_subdir, "lib", filename)
+                if os.path.islink(path):
+                    os.remove(path)
```
Thyre: Is there a reason why we use `os.remove` here, and `remove` below?

easybuild/easyblocks/n/nvidia_compilers.py
lexming: Good point, changed to an error in https://github.com/easybuilders/easybuild-easyblocks/pull/3788/commits/a78668e94633461b989ecf81e40ac582936c2d2a
Regarding `NVSHMEM` that is just an oversight, `nvidia-compilers` does not provide anything else than the compilers. If any users needs extra stuff, `NVHPC` should be used.

easybuild/easyblocks/generic/nvidiabase.py
lexming: > I would suggest to not fail if those paths (i.e. the generic nccl, mpi and nvshmem ones) do not exist to begin with...

That makes sense, several removes already were under a glob expansion to ensure existence of those paths. I have extended that to all removes so that none can fail. See https://github.com/easybuilders/easybuild-easyblocks/pull/3788/commits/195e18a52e89ff31a67caefc82e4a0203d08f901

easybuild/easyblocks/generic/nvidiabase.py
lexming: this is actually very old code and one of the few things I have not changed in this PR, so I suggest to leave this potential issue for another PR

easybuild/easyblocks/generic/nvidiabase.py
lexming: Code from https://github.com/easybuilders/easybuild-easyblocks/pull/3924 merged here

easybuild/easyblocks/generic/nvidiabase.py
lexming: fixed :+1:

easybuild/easyblocks/generic/nvidiabase.py
lexming: Because this is old code from current `nvhpc.py` and at the time our `remove()` didn't support symlinks. Fixed in https://github.com/easybuilders/easybuild-easyblocks/pull/3788/commits/d9a3adb0939c5c7e23cf05ddac3e71d190305e1d

lexming:
Test report by @lexming
Using easyblocks from PR(s) https://github.com/easybuilders/easybuild-easyblocks/pull/3788
#### Overview of tested easyconfigs from https://github.com/easybuilders/easybuild-easyconfigs/pull/23125 (in order)
* SUCCESS nvidia-compilers-25.1-CUDA-12.6.0.eb
* SUCCESS nvidia-compilers-25.1.eb
* SUCCESS nvidia-compilers-25.3.eb
* SUCCESS NVHPC-25.1-CUDA-12.6.0.eb
* SUCCESS NVHPC-25.1.eb
* SUCCESS NVHPC-25.3.eb
* SUCCESS nvidia-compilers-25.3-CUDA-12.8.0.eb
* SUCCESS NVHPC-25.3-CUDA-12.8.0.eb

Build succeeded for 8 out of 8 (total: 32 mins 55 secs) (8 easyconfigs in total)
node800.hydra.os - Linux Rocky Linux 9.6 (Blue Onyx), x86_64, AMD EPYC 9275F 24-Core Processor, 1 x NVIDIA NVIDIA H200 NVL, 580.95.05, Python 3.9.21
See https://gist.github.com/lexming/5f38d13b06fd79c6a99eed7bc31de87e for a full test report.

Thyre:
To correctly set the default CUDA version, we need to change the command in `install_components/install`. This is especially important with CUDA 13 (NVHPC 25.9+).

By default, the installer runs:

```
    CC=$INSTALL_DIR/$arch/$release/compilers/bin/nvc
    DESIREDCUDA=$($CC -printcudaversion 2>&1 | grep -i "selected cuda version" | cut -d'=' -f2)
```

which determines e.g. the paths for the symlinks created by NVHPC.

Without a GPU present, this will likely result in the maximum CUDA version. We should replace `$CC -printcudaversion` by e.g. `nvc -printcudaversion -acc -gpu=sm_70` where `sm_70` should be the lowest GPU architecture one has specified. I need to look into this to fix our installation on e.g. JUSUF...

lexming:
We can do this in a follow-up PR. CUDA 13 is not covered here.

easybuild/easyblocks/n/nvidia_compilers.py
Thyre: > If any users needs extra stuff, `NVHPC` should be used.

Well, this is not compatible with the approach chosen in framework, where toolchains now build on top of `nvidia-compilers`. So one would need to maintain a custom set of toolchains to keep the options one has right now.

Until we're certain that we are able to provide _all_ the libraries NVHPC ships internally, I'm reluctant to simply completely remove all the libraries with giving users only the option to use _none_ (`nvidia-compilers`) or _all_ (`NVHPC`) of the libraries (including HPCX), with nothing sensible in between.

Just to give an overview what is missing in CUDA (just the math libraries alone):

- [cuBLASMp](https://docs.nvidia.com/cuda/cublasmp/) -- https://github.com/easybuilders/easybuild-easyconfigs/pull/24772
- [cuFFTMp](https://docs.nvidia.com/cuda/cufftmp/index.html) -- [available as EC, but not for `nvidia-compilers`](https://github.com/easybuilders/easybuild-easyconfigs/blob/develop/easybuild/easyconfigs/c/cuFFTMp/cuFFTMp-11.4.0-gompi-2025b-CUDA-12.9.1.eb)
- [cuSOLVERMp](https://docs.nvidia.com/cuda/cusolvermp/) -- https://github.com/easybuilders/easybuild-easyconfigs/pull/24773
- [cuTENSOR](https://docs.nvidia.com/cuda/cutensor/latest/index.html) & [cuTENSORMg](https://docs.nvidia.com/cuda/cutensor/latest/user_guide_cutensorMg.html) -- [available as EC](https://github.com/easybuilders/easybuild-easyconfigs/blob/develop/easybuild/easyconfigs/c/cuTENSOR/cuTENSOR-2.3.1.0-CUDA-12.9.1.eb)

There are applications which rely on these for certain functionality. Take e.g. GROMACS, using cuFFTMp as an optional feature.

Thyre: Keeping any of the `math_libs` also requires keeping `nccl` and `nvshmem`, since those are directly linked. Alternatively, one would need to provide the libraries via dependencies, which will be complicated if we want to build NVSHMEM (which needs the compiler we're just installing).

easybuild/easyblocks/f/fftw.py
```diff
@@ -236,17 +236,18 @@ def test_step(self):
...
-        if self.toolchain.mpi_family() == toolchain.OPENMPI and not self.toolchain.comp_family() == TC_CONSTANT_FUJITSU:
+        if self.toolchain.mpi_family() == toolchain.OPENMPI and self.toolchain.comp_family() != TC_CONSTANT_FUJITSU:
 
             # allow oversubscription of number of processes over number of available cores with OpenMPI 3.0 & newer,
             # to avoid that some tests fail if only a handful of cores are available
             ompi_ver = get_software_version('OpenMPI')
-            if LooseVersion(ompi_ver) >= LooseVersion('3.0') and LooseVersion(ompi_ver) < LooseVersion('5.0'):
-                if 'OMPI_MCA_rmaps_base_oversubscribe' not in self.cfg['pretestopts']:
-                    self.cfg.update('pretestopts', "export OMPI_MCA_rmaps_base_oversubscribe=true && ")
             if LooseVersion(ompi_ver) >= LooseVersion('5.0'):
```
boegel: `ompi_ver` may be `None` here, which happens when `NVHPC` toolchain is used, so we need to be a little bit more careful here:

```suggestion
            if ompi_ver and LooseVersion(ompi_ver) >= LooseVersion('5.0'):

File: easybuild/easyblocks/generic/nvidiabase.py (hunk 3 above)
Comment (boegel) [reply easybuild/easyblocks/generic/nvidiabase.py:72]: Indeed, let's deal with this in a follow-up PR.

It should boils down to making this easyblock aware of the `sysroot` EasyBuild configuration setting.

May be of interest to @adammccartney

File: easybuild/easyblocks/generic/nvidiabase.py [hunk 9]
```diff
@@ -0,0 +1,593 @@
...
+        # In the end, set LIBRARY_PATH equal to LD_LIBRARY_PATH
+        self.module_load_environment.LIBRARY_PATH = self.module_load_environment.LD_LIBRARY_PATH
+
+    def __init__(self, *args, **kwargs):
+        """Easyblock constructor, define custom class variables specific to NVHPC."""
+        super().__init__(*args, **kwargs)
+
+        # EULA for NVHPC must be accepted via --accept-eula-for EasyBuild configuration option,
+        # or via 'accept_eula = True' in easyconfig file
+        self.check_accepted_eula(more_info='https://docs.nvidia.com/hpc-sdk/eula/index.html', name="NVHPC")
```
boegel: This should only be done in `install_step`.

We don't want to require the EULA just to parse an `NVHPC` or `nvidia-compilers` easyconfig, but only to actually *install* it.

easybuild/easyblocks/generic/nvidiabase.py
```diff
@@ -0,0 +1,593 @@
...
+        if len(supported_cuda_versions) == 1:
+            active_cuda_version = supported_cuda_versions[0]
+            self.log.info(
+                f"Missing 'default_cuda_version' or CUDA dependency. Using CUDA version '{active_cuda_version}', "
+                f"as it is the only version supported by {self.name}-{self.version}."
+            )
+            return filter_major_minor(active_cuda_version)
+
+        error_msg = f"Missing 'default_cuda_version' or CUDA dependency for {self.name}. "
+        error_msg += "Either add CUDA as dependency or manually define 'default_cuda_version'."
```
boegel: should also mention `nvidia-compilers` here, not only `CUDA`...

boegel (changes requested):

easybuild/easyblocks/generic/nvidiabase.py
boegel: fixed in da1034166f96714124a0a3f7bd04466a316f5d2d

easybuild/easyblocks/generic/nvidiabase.py
boegel: fixed in c8fa1283e9f3fc2e3d39f537bf6e95646df76b0a

boegel:
Test report by @boegel
 * **SUCCESS** _FFTW.MPI-3.3.10-gompi-2023b.eb_ 
  
 * **SUCCESS** _FFTW.MPI-3.3.10-lompi-2025b.eb_ 
  

Build succeeded for 2 out of 2 (total: 56 mins 14 secs) (2 easyconfigs in total)
node4245.shinx.os - Linux RHEL 9.6, x86_64, AMD EPYC 9654 96-Core Processor (zen4), Python 3.9.21
See https://gist.github.com/boegel/c2d2b767764aeb742f1b5345324970d1 for a full test report.

PR #3789, opened 2025-06-18
add option to run unittests of LAMMPS

laraPPr:


easybuild/easyblocks/l/lammps.py
```diff
@@ -370,6 +371,9 @@ def configure_step(self, **kwargs):
             for package in self.cfg['user_packages']:
                 self.cfg.update('configopts', '-D%s%s=on' % (self.pkg_user_prefix, package))
 
+        if self.cfg['runtest']:
+            self.cfg.update('configopts', '-DENABLE_TESTING=on')
```
ocaisa: This requires that PyYAML is available as a build dependency. It probably is as an in-direct dependency, but better to check for it

easybuild/easyblocks/l/lammps.py
```diff
@@ -243,6 +243,7 @@ def extra_options(**kwargs):
             'kokkos_arch': [None, "Set kokkos processor arch manually, if auto-detection doesn't work.", CUSTOM],
             'user_packages': [None, "List user packages (without prefix PKG_ or USER-PKG_).", CUSTOM],
             'sanity_check_test_inputs': [None, "List of tests for sanity-check.", CUSTOM],
+            'runtest': [None, "Enable running the lammps unittests", CUSTOM],
```
ocaisa: I don't think you need this override do you? It's available by default from `CMakeMake`. The only thing you might want to check is that you only accept a Boolean.

ocaisa: And maybe if it is set to `None`, that you auto-enable the tests (and watch all the problems pop up!)

ocaisa: If we did auto-enable tests, I would only do it from a certain LAMMPS version onwards

laraPPr: when it is set to `None` it does not run the `ctests`

ocaisa: Yes, but that would mean that the easyconfig has not changed the default value. From that you can decide at an easyblock level whether to therefore run the tests by default. This would still allow a user to add
```
runtest = False
```
in an easyconfig and have it respected.

laraPPr: I understand. So you want it to run by default from a certain lammps version (Or maybe only the stable versions?). And than it can be overwritten in the easyconfig with `runtest = False`?

easybuild/easyblocks/l/lammps.py
laraPPr: It started compiling it themselves. Can it be added as build-dependency?

laraPPr: And it also compiled ´google framework´

laraPPr: It also seems to need googletest. But I'm hitting an error when it is one of the build-dependencies

laraPPr: If googletest is in the builddependencies it still downloads it but an older version `1.12.1` so gonna leave that depency for now.

easybuild/easyblocks/l/lammps.py
```diff
@@ -370,6 +370,15 @@ def configure_step(self, **kwargs):
             for package in self.cfg['user_packages']:
                 self.cfg.update('configopts', '-D%s%s=on' % (self.pkg_user_prefix, package))
 
+        if LooseVersion(self.cur_version) >= LooseVersion('29Aug2024'):
+            if self.cfg['runtest'] is None or self.cfg['runtest']:
+                build_dep_names = [d['name'] for d in self.cfg.builddependencies()]
+                for tool in ['PyYAML']:
```
laraPPr, line 380: check for PyYAML which is not indirect dependeny of lammps. When PyYAML is not a builddependency it gets downloaded and compiled in the background.

laraPPr: There is also `googletests` version `1.12.1` which now is dowloaded and compiled in the background even when I added [`googletest`, `1.14.0`](https://github.com/easybuilders/easybuild-easyconfigs/blob/develop/easybuild/easyconfigs/g/googletest/googletest-1.14.0-GCCcore-13.2.0.eb) as a builddependency.

easybuild/easyblocks/l/lammps.py
```diff
@@ -370,6 +370,15 @@ def configure_step(self, **kwargs):
             for package in self.cfg['user_packages']:
                 self.cfg.update('configopts', '-D%s%s=on' % (self.pkg_user_prefix, package))
 
+        if LooseVersion(self.cur_version) >= LooseVersion('29Aug2024'):
+            if self.cfg['runtest'] is None or self.cfg['runtest']:
```
ocaisa, line 377: If it is None set it to True, otherwise the test step will not be run

easybuild/easyblocks/l/lammps.py
```diff
@@ -370,6 +370,16 @@ def configure_step(self, **kwargs):
             for package in self.cfg['user_packages']:
                 self.cfg.update('configopts', '-D%s%s=on' % (self.pkg_user_prefix, package))
 
+        if LooseVersion(self.cur_version) >= LooseVersion('29Aug2024'):
```
lorisercole:
```suggestion
        if LooseVersion(self.cur_version) >= LooseVersion(translate_lammps_version('29Aug2024')):
```

lorisercole:
> When testing with a CUDA build I hit the `-Xcudafe --diag_suppress=unrecognized_pragma` error again. Now testing again with the patch that I have. But maybe someone else can also try to see if they also hit it?

Yes I also hit this error. It's related to this: https://github.com/lammps/lammps/issues/4052
Citing Axel's comment:

> The unittest tree does not support KOKKOS with GPUs anyway so if you want a production binary, you don't need to enable testing and if you want to do testing, you don't need to include KOKKOS with CUDA.
> There are currently no plans to add this kind of support to the unittest tree.

PR #3793, opened 2025-06-23
use property for gcc_prefix/gcc_root in LLVM easyblock

Flamefire:
As [discussed](https://github.com/easybuilders/easybuild-easyblocks/pull/3770#issuecomment-2996119299) this uses properties that auto-initialize on (first) use to avoid them being used before setting in `_set_gcc_prefix` which acted as a configure-method AND setting those properties

**BUGFIX** (Separated out to https://github.com/easybuilders/easybuild-easyblocks/pull/3817 but would cause a conflict)
For `bootstrap = False` the function `_create_compiler_config_file` was called without checking the version which results in adding a CLI argument for clang compilers that was introduced in LLVM 19 which causes failures.

### DIFF
--- a/easybuild/easyblocks/l/llvm.py
+++ b/easybuild/easyblocks/l/llvm.py
@@ -391,8 +391,9 @@ class EB_LLVM(CMakeMake):
 
     def _set_gcc_prefix_probs(self):
         """Set properties of currently loaded GCC installation"""
-        self._gcc_root, self._gcc_prefix = self._get_gcc_prefix()
-        self.log.debug("Using %s as the gcc install location", self._gcc_prefix)
+        if self._gcc_root is None:
+            self._gcc_root, self._gcc_prefix = self._get_gcc_prefix()
+            self.log.debug("Using %s as the gcc install location", self._gcc_prefix)
 
     @property
     def llvm_src_dir(self):
--- a/easybuild/easyblocks/l/llvm.py
+++ b/easybuild/easyblocks/l/llvm.py
@@ -260,7 +260,8 @@ class EB_LLVM(CMakeMake):
             # Bypass the .mod file check for GCCcore installs
             self.cfg['skip_mod_files_sanity_check'] = True
         self.final_runtimes = []
-        self.gcc_prefix = None
+        self._gcc_prefix = None
+        self._gcc_root = None
         self.runtimes_cmake_args = {
             'CMAKE_C_COMPILER': [],
             'CMAKE_C_FLAGS': [],
@@ -375,6 +376,23 @@ class EB_LLVM(CMakeMake):
 
         self._cmakeopts = {}
         self._cfgopts = list(filter(None, self.cfg.get('configopts', '').split()))
+
+    @property
+    def gcc_prefix(self):
+        """Return the GCC prefix (versioned folder in <gcc_root>/lib)."""
+        self._set_gcc_prefix_probs()
+        return self._gcc_prefix
+
+    @property
+    def gcc_root(self):
+        """Return the GCC root folder from dependencies/toolchain."""
+        self._set_gcc_prefix_probs()
+        return self._gcc_root
+
+    def _set_gcc_prefix_probs(self):
+        """Set properties of currently loaded GCC installation"""
+        self._gcc_root, self._gcc_prefix = self._get_gcc_prefix()
+        self.log.debug("Using %s as the gcc install location", self._gcc_prefix)
 
     @property
     def llvm_src_dir(self):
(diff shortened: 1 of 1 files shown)


Flamefire:
Test report by @Flamefire
 * **SUCCESS** _LLVM-19.1.7-GCCcore-13.3.0.eb_ 
 * **FAIL (build issue)** _LLVM-18.1.8-GCCcore-13.3.0.eb_ (partial log available at https://gist.github.com/Flamefire/0b549d9a4d49725cb3648efcc70b82bd)

Build succeeded for 1 out of 2 (3 easyconfigs in total)
c132 - Linux AlmaLinux 9.4, x86_64, AMD EPYC 9334 32-Core Processor (zen4), 4 x NVIDIA NVIDIA H100, 560.35.03, Python 3.9.18
See https://gist.github.com/Flamefire/a646d411b1c6ab9fd49d567f6d49e854 for a full test report.

Flamefire:
Test report by @Flamefire
 * **SUCCESS** _LLVM-19.1.7-GCCcore-13.3.0.eb_ 
 * **SUCCESS** _LLVM-20.1.5-GCCcore-13.3.0.eb_ 
 * **FAIL (build issue)** _LLVM-18.1.8-GCCcore-13.3.0.eb_ (partial log available at https://gist.github.com/Flamefire/a5d75f36955ef6221ce48ef3eff1d13c)

Build succeeded for 2 out of 3 (3 easyconfigs in total)
n1623 - Linux RHEL 8.9 (Ootpa), x86_64, Intel(R) Xeon(R) Platinum 8470 (sapphirerapids), Python 3.9.18
See https://gist.github.com/Flamefire/89e8448451c51ac24bfe2d7bd27b90cc for a full test report.

boegel:
@Flamefire Merge conflict to fix now that https://github.com/easybuilders/easybuild-easyblocks/pull/3817 is merged?

Flamefire:
Test report by @Flamefire
"Can't find path /home/h3/s3248973/LLVM/18.1.8-GCCcore-13.3.0"
Failed during parsing of the easyconfigs, so no ecs were built (3 easyconfigs in total)
n1322 - Linux RHEL 8.9 (Ootpa), x86_64, Intel(R) Xeon(R) Platinum 8470 (sapphirerapids), Python 3.9.18
See https://gist.github.com/Flamefire/e84cd867bfaac91ec2c9836fc9826bbf for a full test report.

Flamefire:
Rebased

Flamefire:
Test report by @Flamefire
 * **SUCCESS** _LLVM-19.1.7-GCCcore-13.3.0.eb_ 
 * **SUCCESS** _LLVM-20.1.5-GCCcore-13.3.0.eb_ 

c83 - Linux AlmaLinux 9.4, x86_64, AMD EPYC 9334 32-Core Processor (zen4), 4 x NVIDIA NVIDIA H100, 560.35.03, Python 3.9.18
See https://gist.github.com/Flamefire/a602c8f8c30061b8362f4bed6c7a001c for a full test report.

Flamefire:
Test report by @Flamefire
 * **SUCCESS** _LLVM-19.1.7-GCCcore-13.3.0.eb_ 
 * **SUCCESS** _LLVM-20.1.5-GCCcore-13.3.0.eb_ 

i7005 - Linux Rocky Linux 8.9 (Green Obsidian), x86_64, AMD EPYC 7702 64-Core Processor (zen2), Python 3.9.18
See https://gist.github.com/Flamefire/eeccc718e59e802a66519b9a2911c76f for a full test report.

easybuild/easyblocks/l/llvm.py
```diff
@@ -376,6 +377,23 @@ def __init__(self, *args, **kwargs):
...
+
+    @property
+    def gcc_root(self):
+        """Return the GCC root folder from dependencies/toolchain."""
+        self._set_gcc_prefix_probs()
+        return self._gcc_root
+
+    def _set_gcc_prefix_probs(self):
+        """Set properties of currently loaded GCC installation"""
+        self._gcc_root, self._gcc_prefix = self._get_gcc_prefix()
```
Crivella: I would put this inside an if as it was done originally.
I do not foresee any situation where these variables would change during the build process

easybuild/easyblocks/l/llvm.py
```diff
@@ -792,7 +790,17 @@ def configure_step(self):
...
+        # But the --gcc-toolchain and --gcc-install-dir for flang are not supported before LLVM 19
+        # https://github.com/llvm/llvm-project/pull/87360
+        if LooseVersion(self.version) < '19':
+            self.log.debug("Using GCC_INSTALL_PREFIX")
+            self.general_opts['GCC_INSTALL_PREFIX'] = self.gcc_root
+        else:
+            # See https://github.com/llvm/llvm-project/pull/85891#issuecomment-2021370667
+            self.log.debug("Using '--gcc-install-dir' in CMAKE_C_FLAGS and CMAKE_CXX_FLAGS")
+            self.runtimes_cmake_args['CMAKE_C_FLAGS'] += [f'--gcc-install-dir={self.gcc_prefix}']
+            self.runtimes_cmake_args['CMAKE_CXX_FLAGS'] += [f'--gcc-install-dir={self.gcc_prefix}']
```
Crivella, line 804: I would still make sure this is called every time we used to call `_set_gcc_prefix`.
Right now we never reset `runtimes_cmake_args` across stages, but in case we ever start doing it, that would prevent shooting ourselves in the foot

easybuild/easyblocks/l/llvm.py
Flamefire: This is a method only used on access of the properties to set the "backing storage". So the if is outside this function

I agree they won't change, so there is no harm if this method is (wrongly) called again

Crivella: There is no `if` either inside `gcc_prefix` nor `gcc_root` so the method `self._get_gcc_prefix()` will be called every time one of these 2 properties are requested

easybuild/easyblocks/l/llvm.py
Flamefire: We **append** to the flags here so calling it again would be a mistake, wouldn't it? If we reset it I'd say it would be unexpected to suddenly add it. That should be done on purpose / with clear intent, shouldn't it?

As for previous usages:
- `_create_compiler_config_file` shouldn't set CMake flags, should it? I see that in the iterations of staged builds this is called to the flags are indirectly set but also only once: By the `self._set_gcc_prefix()` here in configure step after which `self._set_gcc_prefix()` is a no-op
- When running `sanity_check_step` this won't be used anymore

So IMO the least surprising place to set them is here in `configure_step` which should be exactly the same behavior as before unless we end up skipping this step

Crivella: True, also before it was not checking for `runtimes_cmake_args` and only if `gcc_prefix` is set so doing it like this would be fine, we can think on re-initializing `runtimes_cmake_args` if we ever get there

easybuild/easyblocks/l/llvm.py
Flamefire: You are right, thanks! Seems I couldn't decide where to put the `if` and forgot it in the end.
Added

Crivella:
EB_ARGS="LLVM-16.0.6-GCCcore-12.3.0.eb LLVM-18.1.8-GCCcore-13.3.0.eb LLVM-19.1.7-GCCcore-13.3.0.eb LLVM-20.1.5-GCCcore-13.3.0.eb"
CORE_CNT=16

Issue #3508, opened 2024-11-14
Update numpy easyblock for numpy>=2.0


jpecar:
Current numpy easyblock is a child of fortranpythonpackage and this one wants to run "python setup.py build" to build it. Newer numpy comes without setup.py and seems to be using meson build system. Therefore numpy easyblock needs to be adapted.

Not sure how to even start here. Would adapted numpy easyblock need to become a child of mesonninja easyblock? Can import lines at the top be also selected by "if LooseVersion..." conditions? How would that then affect the rest of the code?

Thyre:
If we want to keep a single Python file, we could do something like with QuantumESPRESSO, where two separate EasyBlocks exist, which get selected from an EasyBlock wrapping both, see https://github.com/easybuilders/easybuild-easyblocks/blob/57c0eaed8dc29e223fe68a75f7bf195cca0c2d04/easybuild/easyblocks/q/quantumespresso.py

```python
class EB_QuantumESPRESSO(EasyBlock):
    @staticmethod
    def extra_options():
        """Custom easyconfig parameters for Quantum ESPRESSO."""
        extra_opts = EB_QuantumESPRESSOconfig.extra_options()
        extra_opts.update(EB_QuantumESPRESSOcmake.extra_options())

        return extra_opts

    def __getattribute__(self, name):
        try:
            ebclass = object.__getattribute__(self, 'ebclass')
        except AttributeError:
            ebclass = None
        if name == 'ebclass':
            return ebclass
        if ebclass is None:
            return object.__getattribute__(self, name)
        return ebclass.__getattribute__(name)

    def __init__(self, ec, *args, **kwargs):
        """Select the correct EB depending on version."""
        super(EB_QuantumESPRESSO, self).__init__(ec, *args, **kwargs)

        if LooseVersion(self.version) < LooseVersion('7.3.1'):
            self.log.info('Using legacy easyblock for Quantum ESPRESSO')
            eb = EB_QuantumESPRESSOconfig
        else:
            self.log.info('Using CMake easyblock for Quantum ESPRESSO')
            eb = EB_QuantumESPRESSOcmake

        # Required to avoid CMakeMake default extra_opts to override the ConfigMake ones
        new_ec = EasyConfig(ec.path, extra_options=eb.extra_options())
        self.ebclass = eb(new_ec, *args, **kwargs)
```

boegel:
[...]

PR #3797, opened 2025-06-24
enhance` PythonPackage` to allow specifying location to temporary installation used in test step + also update `$PATH` to pick up on binaries/scripts from it

boegel:
required for:
- #3508


akesandgren (approved):
LGTM

PR #3798, opened 2025-06-26
WIP: update Python easyblock to use custom ctypes for shared libraries

dagonzalezfo:
This was first exposed as a patch on https://github.com/easybuilders/easybuild-easyconfigs/pull/23042 
After reviewing it, @casparvl suggested that this should be integrated into the easyblock.

### DIFF
--- a/easybuild/easyblocks/p/python.py
+++ b/easybuild/easyblocks/p/python.py
@@ -365,7 +365,6 @@ class EB_Python(ConfigureMake):
         if 'LD_LIBRARY_PATH' in filtered_env_vars and 'LIBRARY_PATH' not in filtered_env_vars:
             ctypes_util_py = os.path.join("Lib", "ctypes", "util.py")
             orig_gcc_so_name = None
-            
             # Let's do this incrementally since we are going back in time
             # if LooseVersion(self.version) >= "3.9.1":
             #    # From 3.9.1 to at least v3.12.4 there is only one match for this line
--- a/easybuild/easyblocks/p/python.py
+++ b/easybuild/easyblocks/p/python.py
@@ -50,7 +50,7 @@ from easybuild.tools.build_log import EasyBuildError, print_warning
 from easybuild.tools.config import build_option, ERROR, EBPYTHONPREFIXES
 from easybuild.tools.modules import get_software_libdir, get_software_root, get_software_version
 from easybuild.tools.filetools import apply_regex_substitutions, change_dir, mkdir
-from easybuild.tools.filetools import read_file, remove_dir, symlink, write_file, open_file
+from easybuild.tools.filetools import read_file, remove_dir, symlink, write_file
 from easybuild.tools.run import run_shell_cmd
 from easybuild.tools.systemtools import get_shared_lib_ext
 from easybuild.tools.utilities import trace_msg
--- a/easybuild/easyblocks/p/python.py
+++ b/easybuild/easyblocks/p/python.py
@@ -365,11 +365,12 @@ class EB_Python(ConfigureMake):
         if 'LD_LIBRARY_PATH' in filtered_env_vars and 'LIBRARY_PATH' not in filtered_env_vars:
             ctypes_util_py = os.path.join("Lib", "ctypes", "util.py")
             orig_gcc_so_name = None
+            
             # Let's do this incrementally since we are going back in time
-            #if LooseVersion(self.version) >= "3.9.1":
+            # if LooseVersion(self.version) >= "3.9.1":
             #    # From 3.9.1 to at least v3.12.4 there is only one match for this line
             #    orig_gcc_so_name = "_get_soname(_findLib_gcc(name)) or _get_soname(_findLib_ld(name))"
-            #if orig_gcc_so_name:
+            # if orig_gcc_so_name:
             #    orig_gcc_so_name_regex = r'(\s*)' + re.escape(orig_gcc_so_name) + r'(\s*)'
             #    # _get_soname() takes the full path as an argument and uses objdump to get the SONAME field from
             #    # the shared object file. The presence or absence of the SONAME field in the ELF header of a shared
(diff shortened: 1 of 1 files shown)


boegel:
@casparvl Can you help shepherd this contribution?

casparvl:
@dagonzalezfo so the key problem here is that using a regular expression to replace one regular expression by another is a pain? Who would have thought ;-)

I know that we conventionally use `apply_regex_substitutions()` as a way to patch things at the EasyBlock level, but... we don't _have_ to do it that way, right? I'm wondering about the alternatives... I guess we can call `sed`? Or: define the context for a patch file in the python code, write that to a temporary file, then call `patch` (or some easybuild equivalent: framework has a patch step, so should already have this functionality somewhere, I guess).

I'm also trying to understand what fundamentally goes wrong here 

```
OSError: /cvmfs/software.eessi.io/versions/2023.06/compat/linux/x86_64/usr/lib/../lib64/libc.so: invalid ELF header
```

What were you trying to do/run? I see that this file is actually _not_ a shared library:

```
$ file /cvmfs/software.eessi.io/versions/2023.06/compat/linux/x86_64/usr/lib64/libc.so
/cvmfs/software.eessi.io/versions/2023.06/compat/linux/x86_64/usr/lib64/libc.so: ASCII text
```
But the same is true for a regular `libc.so`, so it's no surprise:
```
file /usr/lib64/libc.so
/usr/lib64/libc.so: ASCII text
```

But something must be calling it with the assumption that it is an ELF file?

easybuild/easyblocks/p/python.py
```diff
@@ -367,13 +368,45 @@ def patch_step(self, *args, **kwargs):
...
+                        [(orig_import, textwrap.dedent(updated_import))],
+                        on_missing_match=ERROR
+                        )
+                # add posix specific support
+                # this code snip use util_findLib_ld function to recover the 
+                # complete librarypath
+                orig_support_marker=r'if _sys.platform.startswith\(\"aix\"\):'
+                updated_support_marker="""
+                if _os.name == "posix":
+                    if name and name.endswith(".so"):
```
casparvl: This may be too strict. E.g. someone might try to load a versioned library, e.g. `libSDL2-2.0.so.0`. That still fails, because we never enter this 'if'.

What case are you really trying to exclude here?

I did notice that I couldn't really find a `find_library` call that would work for `libSDL2-2.0` either (e.g. `python -c "import ctypes; print(ctypes.util.find_library('SDL2-2.0'))"` returns `None`), maybe that's the case you were trying to exclude?

dagonzalezfo: The code should mimic this behaviour:
LD_LIBRARY_PATH=$LIBRARY_PATH python -c "import ctypes; print(ctypes.CDLL('libSDL2-2.0.so.0'))"

casparvl:
Writing down some more things to help me understand what's going on, after some of it became clearer in a call today with @dagonzalezfo .

We need to make sure that libraries from both the `software-layer` and `compatibility-layer` are found correctly, for calls to `ctypes.util.find_library()`, `ctypes.CDLL()` and `ctypes.LoadLibrary()` (which essentally calls `ctypes.CDLL()`). Breaking this down one by one:

- `ctypes.util.find_library()` first calls `_findSoname_ldconfig`. This is how it will find libraries on the standard search paths, but for this to work, it needs to use the `ldconfig` from the `sysroot` (if a sysroot was configured). That was patched [here](https://github.com/easybuilders/easybuild-easyblocks/blob/b1bdebb6d66c86487802c1c57bde526acd8b5336/easybuild/easyblocks/p/python.py#L440) in the EasyBlock. Essentially, it then grabs the library _name_ (just the name, not full path) from the output of `<sysroot>/sbin/ldconfig -p`.

- `ctypes.util.find_library()` then calls `findLib_gcc` which calls e.g.
```
$ gcc -Wl,-t -o /dev/null -lSDL2
...
/cvmfs/software.eessi.io/versions/2023.06/software/linux/x86_64/amd/zen4/software/SDL2/2.28.5-GCCcore-13.2.0/lib/../lib64/libSDL2.so
...
``` 
 `gcc -Wl,-t -o /dev/null -lSDL2` to find the `SDL2` library. This prints something like `/cvmfs/software.eessi.io/versions/2023.06/software/linux/x86_64/amd/zen4/software/SDL2/2.28.5-GCCcore-13.2.0/lib/../lib64/libSDL2.so` on the output. But: standard `ctypes.util.find_library()` wraps this in a `_get_soname()` call, to strip the path and only return the so-name (on Linux). A [previous patch](https://github.com/easybuilders/easybuild-easyblocks/pull/3352/) by @ocaisa changed this behaviour to return the full path, as the result of `find_library()` is often passed to a call that then loads the library (like `ctypes.CDLL`, but other methods may be used, like `cffi`).

- `ctypes.util.find_library()` then calls `findLib_ld`, which loops through entries in the `LD_LIBRARY_PATH`, each time calling `ld -t -L <dir_from_ld_library_path> -o /dev/null -lSDL2`. I.e. it would look something like:

```
$ ld -t -L /cvmfs/software.eessi.io/versions/2023.06/software/linux/x86_64/amd/zen4/software/SDL2/2.28.5-GCCcore-13.2.0/lib -o /dev/null -lSDL2
/cvmfs/software.eessi.io/versions/2023.06/software/linux/x86_64/amd/zen4/software/SDL2/2.28.5-GCCcore-13.2.0/lib/libSDL2.so
```
Again, this is wrapped in a `_get_soname()` by default, to strip the path and return only the so-name (on Linux). Again, this behavior was changed to return full path by @ocaisa 's previous patch.

This, however, was not enough, because if someone calls `ctypes.CDLL` directly (by library name, not full path), it would still fail. That's where this PR comes in.

<This comment is to be continued... need to leave now...>

dagonzalezfo:
Changes implemented https://github.com/easybuilders/easybuild-easyconfigs/pull/23499 
Please use it with the modified easyblock

PR #3747, opened 2025-05-29
LLVM: Use dlopen for libomptarget / offload

Thyre:
Avoid linking in a system CUDA / ROCm by mistake by forcing LLVM to use `dlopen` the libraries instead. This also removes the need to pass in these libraries as dependencies when building LLVM altogether, except if testing of the built libraries is wanted.

This should only really matter for systems where a system ROCm or CUDA exists.

PR #3799, opened 2025-06-26
move openmp to a runtime in newer versions of LLVM

### LINKED ISSUE #3795: LLVM is missing OpenMP Tools Interface for offload
Something in the way we build LLVM breaks the detection of the OpenMP Tools Interface for the `offload` project of LLVM.
When checking for OpenMP Tools Interface support, the logs report:

```
easybuild/easybuild-LLVM-20.1.5-20250624.112522.log:38438:-- OMPT target disabled
easybuild/easybuild-LLVM-20.1.5-20250624.112522.log:186357:-- OMPT target disabled
easybuild/easybuild-LLVM-20.1.5-20250624.112522.log:323358:-- OMPT target disabled
```

This happens for single-stage and multi-stage builds, and also affects ROCm-LLVM.
Without this, performance tools might not be able to report all information related to OpenMP offload.

Looking at the `CMakeLists.txt` of LLVM, I don't immediately see what is broken. 
One would need to dig deeper to find out which things are set incorrectly...

Crivella:
With newer versions of LLVM, openmp has been moved from a project to a runtime.
Currently both version can be used but `projects` build are being deprecated https://github.com/llvm/llvm-project/pull/136314

This change is needed also in order to fix
- Fixes #3795
as building `openmp` as a project makes it so the proper CMake variables `LIBOMP_OMPT_SUPPORT` and `LIBOMPTARGET_OMPT_SUPPORT` are not set during the runtime builds, causing OMPT to not be enabled when building `offload`

## TODO

- [x] Version check the changes to make sure it works with older versions of LLVM
  It seems also in `18.1.x` `openmp` can be built as a runtime, will defer to the result of the test builds...
  **WORKED** also with `18.1.8`
- [x] Fix failing `libomptarget` tests
  With this fix `libomptarget` tests are failing again. In this case `libomp.so` is not being found as it is not anymore insilde `.lib/libomp.so` but `./runtimes/runtimes-bins/openmp/runtime/src/libomp.so`
  - possible fix: add the specified folder to `LIBRARY_PATH`/`LD_LIBRARY_PATH` during testing
  - **USED FIX**: Create symlink from the new location of `libomp.so` to the expected one in the build directory

### DIFF
--- a/easybuild/easyblocks/l/llvm.py
+++ b/easybuild/easyblocks/l/llvm.py
@@ -1124,6 +1124,16 @@ class EB_LLVM(CMakeMake):
             gcc_lib = self._get_gcc_libpath(strict=True)
             lib_path = ':'.join(filter(None, [gcc_lib, lib_path]))
 
+        # After switching `openmp` to a runtime build, libomp.so is not produced inside `<builddir>/lib` anymore but
+        # in `<builddir>/runtimes/runtimes-bins/openmp/runtime/src/libomp.so`
+        # This causes the `libomptarget` tests to fail because the compiler cannot find `libomp.so`.
+        check_libomp = os.path.join(basedir, 'runtimes', 'runtimes-bins', 'openmp', 'runtime', 'src', 'libomp.so')
+        needed_libomp = os.path.join(basedir, 'lib', 'libomp.so')
+        if os.path.exists(check_libomp) and not os.path.exists(needed_libomp):
+            # Create a symlink to the libomp.so in the runtimes directory
+            mkdir(os.path.dirname(needed_libomp), parents=True)
+            symlink(check_libomp, needed_libomp)
+
         with _wrap_env(os.path.join(basedir, 'bin'), lib_path):
             cmd = f"make -j {parallel} check-all"
             res = run_shell_cmd(cmd, fail_on_error=False)


Crivella:
Test report by @Crivella
 * **FAIL (build issue)** _123.eb_ (partial log available at https://gist.github.com/Crivella/e1d4d14fca9f992a8f469de198179242)

crivella-desktop - Linux Ubuntu 22.04.5 LTS (Jammy Jellyfish), x86_64, 13th Gen Intel(R) Core(TM) i9-13900K (skylake), Python 3.11.13
See https://gist.github.com/Crivella/781d2143ccf244851dd8ad056f46c34d for a full test report.

Crivella:
Test report by @Crivella
 * **SUCCESS** _123.eb_ 

crivella-desktop - Linux Ubuntu 22.04.5 LTS (Jammy Jellyfish), x86_64, 13th Gen Intel(R) Core(TM) i9-13900K (skylake), Python 3.11.13
See https://gist.github.com/Crivella/03c606b35bb4841344db7ab59014a9d2 for a full test report.

Crivella:
EB_ARGS="--installpath /tmp/$USER/ebpr-3799 LLVM-18.1.8-GCCcore-13.3.0.eb"
CORE_CNT=16

Crivella:
Test report by @Crivella
 * **SUCCESS** _LLVM-19.1.7-GCCcore-13.3.0.eb_ 

crivella-desktop - Linux Ubuntu 22.04.5 LTS (Jammy Jellyfish), x86_64, 13th Gen Intel(R) Core(TM) i9-13900K (skylake), Python 3.11.13
See https://gist.github.com/Crivella/0b752eeb53913bd6c7a861785c297c58 for a full test report.

Crivella:
EB_ARGS="--installpath /tmp/$USER/ebpr-3799 LLVM-20.1.5-GCCcore-13.3.0.eb"
CORE_CNT=16

Thyre:
Test report by @Thyre
"Failed to process easyconfig /home/jreuter/Projects/EasyBuild/easybuild-custom/easybuild/easyconfigs/l/LLVM/LLVM-20.1.7-GCCcore-14.2.0.eb: Failed to find easyconfig file 'lit-18.1.8-GCCcore-14.2.0.eb' when determining module name for: {'full_mod_name': None, 'short_mod_name': None, 'name': 'lit', 'version': '18.1.8', 'versionsuffix': '', 'toolchain': {'name': 'GCCcore', 'version': '14.2.0'}, 'toolchain_inherited': True, 'system': False, 'hidden': False, 'build_only': True, 'external_module': False, 'external_module_metadata': {}}"
Failed during parsing of the easyconfigs, so no ecs were built (1 easyconfigs in total)
Linux - Linux Arch Linux UNKNOWN, x86_64, AMD Ryzen 7 7800X3D 8-Core Processor, 1 x AMD Navi 48 [RX 9070/9070 XT] (device id: 0x7550, gfx: gfx1201, driver: Linuxversion6.15.3-arch1-1(linux@archlinux)(gcc(GCC)15.1.120250425,GNUld(GNUBinutils)2.44.0)#1SMPPREEMPT_DYNAMICThu,19Jun202514:41:19+0000), 1 x AMD Raphael (device id: 0x164e, gfx: gfx1036, driver: Linuxversion6.15.3-arch1-1(linux@archlinux)(gcc(GCC)15.1.120250425,GNUld(GNUBinutils)2.44.0)#1SMPPREEMPT_DYNAMICThu,19Jun202514:41:19+0000), Python 3.13.5
See https://gist.github.com/Thyre/b63c67d03595065b55fbb301c0219f6a for a full test report.

Thyre:
Forgot `--upload-test-report`.

Here's a manual upload: https://gist.github.com/Thyre/1407f397f09e8ee168b162e9f2a8be97

Thyre:
Test run with the LLVM:

```
 jreuter@Linux  ~  which clang
/data/EasyBuild-develop/software/LLVM/20.1.7-GCCcore-14.2.0/bin/clang
 jreuter@Linux  ~  clang -fopenmp --offload-arch=gfx1201 test.c
 jreuter@Linux  ~  OMP_TARGET_OFFLOAD=mandatory ./a.out
[callback_device_initialize] device_num = 0 | type = gfx1201 | device = 0x55ed0788b560 | lookup = 0x7f4edb870ad0 | documentation = (null)
 jreuter@Linux  ~  grep -rni "OMPT target" $EBROOTLLVM/easybuild/*.log
127958:-- OMPT target enabled
187478:-- OMPT target enabled
324243:-- OMPT target enabled
```

Crivella:
@boegel I think this is ready for review

Thyre:
Test report by @Thyre
 * **SUCCESS** _ROCm-LLVM-6.4.1-GCCcore-14.2.0.eb_ 

jrc0850.jureca - Linux Rocky Linux 9.5 (Blue Onyx), x86_64, AMD EPYC 7443 24-Core Processor (zen3), Python 3.9.21
See https://gist.github.com/Thyre/7d0bc093428840b80534d5d1d87b5330 for a full test report.

Crivella:
# Manual test report `19.1.7` (SUCCESS)

On a RHEL9 VM with only lmod + EESSI + EESSI-extend

https://gist.github.com/Crivella/40eb67c82e37433f7b2772f44ea906c8

Thyre:
With ROCm-LLVM (on Linux, i.e. my Arch Linux machine):

```
[ 98%] Linking C shared library libomp.so
cd /data/EasyBuild-develop/build/ROCmLLVM/6.4.1/GCCcore-14.2.0/llvm.obj.1/runtimes/runtimes-bins/openmp/runtime/src && /data/EasyBuild-develop/software/CMake/3.31.3-GCCcore-14.2.0/bin/cmake -E cmake_link_script CMakeFiles/omp.dir/link.txt --verbose=1
/tmp/eb-o98_nop5/tmpq9xaqhte/rpath_wrappers/clang_wrapper/clang --target=x86_64-unknown-linux-gnu -fPIC --gcc-install-dir=/data/EasyBuild-develop/software/GCCcore/14.2.0/lib/gcc/x86_64-pc-linux-gnu/14.2.0 -Werror=date-time -Werror=unguarded-availability-new -Wall -Wextra -Wno-unused-parameter -Wwrite-strings -Wcast-qual -Wmissing-field-initializers -Wimplicit-fallthrough -Wcovered-switch-default -Wno-comment -Wstring-conversion -ffunction-sections -fdata-sections -Wall -fcolor-diagnostics -Wcast-qual -Wformat-pedantic -Wimplicit-fallthrough -Wsign-compare -Wno-extra -Wno-pedantic -fno-semantic-interposition -fdata-sections -O3 -DNDEBUG -Wl,--as-needed -Wl,--version-script=/data/EasyBuild-develop/build/ROCmLLVM/6.4.1/GCCcore-14.2.0/llvm-project-19.0.0-rocm-6.4.1.src/openmp/runtime/src/exports_so.txt -Wl,--undefined-version -static-libgcc -Wl,-z,noexecstack -Xlinker --dependency-file=CMakeFiles/omp.dir/link.d -L/data/EasyBuild-develop/software/numactl/2.0.19-GCCcore-14.2.0/lib64 -L/data/EasyBuild-develop/software/numactl/2.0.19-GCCcore-14.2.0/lib -L/data/EasyBuild-develop/software/libdrm/2.4.125-GCCcore-14.2.0/lib64 -L/data/EasyBuild-develop/software/libdrm/2.4.125-GCCcore-14.2.0/lib -L/data/EasyBuild-develop/software/libffi/3.4.5-GCCcore-14.2.0/lib64 -L/data/EasyBuild-develop/software/libffi/3.4.5-GCCcore-14.2.0/lib -L/data/EasyBuild-develop/software/ncurses/6.5-GCCcore-14.2.0/lib64 -L/data/EasyBuild-develop/software/ncurses/6.5-GCCcore-14.2.0/lib -L/data/EasyBuild-develop/software/libxml2/2.13.4-GCCcore-14.2.0/lib64 -L/data/EasyBuild-develop/software/libxml2/2.13.4-GCCcore-14.2.0/lib -L/data/EasyBuild-develop/software/GMP/6.3.0-GCCcore-14.2.0/lib64 -L/data/EasyBuild-develop/software/GMP/6.3.0-GCCcore-14.2.0/lib -L/data/EasyBuild-develop/software/hwloc/2.11.2-GCCcore-14.2.0/lib64 -L/data/EasyBuild-develop/software/hwloc/2.11.2-GCCcore-14.2.0/lib -L/data/EasyBuild-develop/software/zlib/1.3.1-GCCcore-14.2.0/lib64 -L/data/EasyBuild-develop/software/zlib/1.3.1-GCCcore-14.2.0/lib -L/data/EasyBuild-develop/software/elfutils/0.193-GCCcore-14.2.0/lib64 -L/data/EasyBuild-develop/software/elfutils/0.193-GCCcore-14.2.0/lib -L/data/EasyBuild-develop/software/pkgconf/2.3.0-GCCcore-14.2.0/lib64 -L/data/EasyBuild-develop/software/pkgconf/2.3.0-GCCcore-14.2.0/lib -L/data/EasyBuild-develop/software/Python/3.13.1-GCCcore-14.2.0/lib64 -L/data/EasyBuild-develop/software/Python/3.13.1-GCCcore-14.2.0/lib -L/data/EasyBuild-develop/software/Perl/5.40.0-GCCcore-14.2.0/lib64 -L/data/EasyBuild-develop/software/Perl/5.40.0-GCCcore-14.2.0/lib -L/data/EasyBuild-develop/software/GCCcore/14.2.0/lib64 -L/data/EasyBuild-develop/software/GCCcore/14.2.0/lib -Wl,-z,defs -Wl,-z,nodelete -shared -Wl,-soname,libomp.so -o libomp.so CMakeFiles/omp.dir/kmp_alloc.cpp.o CMakeFiles/omp.dir/kmp_atomic.cpp.o CMakeFiles/omp.dir/kmp_csupport.cpp.o CMakeFiles/omp.dir/kmp_debug.cpp.o CMakeFiles/omp.dir/kmp_itt.cpp.o CMakeFiles/omp.dir/kmp_environment.cpp.o CMakeFiles/omp.dir/kmp_error.cpp.o CMakeFiles/omp.dir/kmp_global.cpp.o CMakeFiles/omp.dir/kmp_i18n.cpp.o CMakeFiles/omp.dir/kmp_io.cpp.o CMakeFiles/omp.dir/kmp_runtime.cpp.o CMakeFiles/omp.dir/kmp_settings.cpp.o CMakeFiles/omp.dir/kmp_str.cpp.o CMakeFiles/omp.dir/kmp_tasking.cpp.o CMakeFiles/omp.dir/kmp_threadprivate.cpp.o CMakeFiles/omp.dir/kmp_utility.cpp.o CMakeFiles/omp.dir/kmp_barrier.cpp.o CMakeFiles/omp.dir/kmp_wait_release.cpp.o CMakeFiles/omp.dir/kmp_affinity.cpp.o CMakeFiles/omp.dir/kmp_dispatch.cpp.o CMakeFiles/omp.dir/kmp_lock.cpp.o CMakeFiles/omp.dir/kmp_sched.cpp.o CMakeFiles/omp.dir/kmp_collapse.cpp.o CMakeFiles/omp.dir/z_Linux_util.cpp.o CMakeFiles/omp.dir/kmp_gsupport.cpp.o CMakeFiles/omp.dir/thirdparty/ittnotify/ittnotify_static.cpp.o CMakeFiles/omp.dir/kmp_taskdeps.cpp.o CMakeFiles/omp.dir/kmp_cancel.cpp.o CMakeFiles/omp.dir/kmp_ftn_cdecl.cpp.o CMakeFiles/omp.dir/kmp_ftn_extra.cpp.o CMakeFiles/omp.dir/kmp_version.cpp.o "CMakeFiles/omp.dir/ompt-general.cpp.o" "CMakeFiles/omp.dir/ompd-specific.cpp.o" CMakeFiles/omp.dir/z_Linux_asm.S.o  /data/EasyBuild-develop/software/hwloc/2.11.2-GCCcore-14.2.0/lib/libhwloc.so -lm -ldl
cd /data/EasyBuild-develop/build/ROCmLLVM/6.4.1/GCCcore-14.2.0/llvm.obj.1/runtimes/runtimes-bins/openmp/runtime/src && /data/EasyBuild-develop/software/CMake/3.31.3-GCCcore-14.2.0/bin/cmake -E create_symlink libomp.so libgomp.so
cd /data/EasyBuild-develop/build/ROCmLLVM/6.4.1/GCCcore-14.2.0/llvm.obj.1/runtimes/runtimes-bins/openmp/runtime/src && /data/EasyBuild-develop/software/CMake/3.31.3-GCCcore-14.2.0/bin/cmake -E create_symlink libomp.so libiomp5.so
cd /data/EasyBuild-develop/build/ROCmLLVM/6.4.1/GCCcore-14.2.0/llvm.obj.1/runtimes/runtimes-bins/openmp/runtime/src && /data/EasyBuild-develop/software/CMake/3.31.3-GCCcore-14.2.0/bin/cmake -E make_directory /data/EasyBuild-develop/build/ROCmLLVM/6.4.1/GCCcore-14.2.0/llvm-project-19.0.0-rocm-6.4.1.src/openmp/runtime/exports/common.ompt.optional/include
cd /data/EasyBuild-develop/build/ROCmLLVM/6.4.1/GCCcore-14.2.0/llvm.obj.1/runtimes/runtimes-bins/openmp/runtime/src && /data/EasyBuild-develop/software/CMake/3.31.3-GCCcore-14.2.0/bin/cmake -E copy omp.h /data/EasyBuild-develop/build/ROCmLLVM/6.4.1/GCCcore-14.2.0/llvm-project-19.0.0-rocm-6.4.1.src/openmp/runtime/exports/common.ompt.optional/include
Error copying file "omp.h" to "/data/EasyBuild-develop/build/ROCmLLVM/6.4.1/GCCcore-14.2.0/llvm-project-19.0.0-rocm-6.4.1.src/openmp/runtime/exports/common.ompt.optional/include".
make[5]: *** [openmp/runtime/src/CMakeFiles/omp.dir/build.make:636: openmp/runtime/src/libomp.so] Error 1
make[5]: *** Deleting file 'openmp/runtime/src/libomp.so'
```

Crivella:
Maybe in the same spirit as the symlink i am doing for `libomp.so` we could do a copy in the expected path before running the install, but since the normal build are working I am not sure if this should be down at the level of the child easyblock?

Thyre:
> Maybe in the same spirit as the symlink i am doing for `libomp.so` we could do a copy in the expected path before running the install, but since the normal build are working I am not sure if this should be down at the level of the child easyblock?

Yeah... I'm trying to figure out where LLVM 19 & ROCm 6.4.1 diverge. I feel like this is just one or two commits that differ and cause these issues.

Crivella:
Rebased after merging
- https://github.com/easybuilders/easybuild-easyblocks/pull/3747

easybuild/easyblocks/l/llvm.py
```diff
@@ -313,7 +313,7 @@ def __init__(self, *args, **kwargs):
             self.final_runtimes += ['compiler-rt', 'libunwind', 'libcxx', 'libcxxabi']
 
         if self.cfg['build_openmp']:
-            self.final_projects.append('openmp')
+            self.final_runtimes.append('openmp')
```
boegel, line 326: @Crivella This isn't only for newer LLVM versions as the PR title suggests, but for *all* LLVM versions?

Thyre:
> That PR isn't in any released version yet. So I think we should not (unconditionally) do this for all LLVM 18+ builds (where this easyblock is used)

It already breaks significant functionality for LLVM 17 and newer, so that's a good reason to move it to runtimes.

easybuild/easyblocks/l/llvm.py
Crivella: Basically all we support. I had wrote newer originally also because i was not sure this would work with 18.x but it ended up working there as well

Thyre: We may want to limit that, yeah (at least if we allow building `libomp.so` without building a full LLVM).
LLVM supports `openmp` for `LLVM_ENABLE_RUNTIMES` since LLVM 12.0.0-rc1. Prior versions only accept this as a project.

Choosing between `LLVM_ENABLE_RUNTIMES` and `LLVM_ENABLE_PROJECTS`  matters the most for builds involving OpenMP offload. If we only focus on the OpenMP Tools Interface, which this PR aims to fix, only LLVM 17+ has support for the offload part. Any version before that could still use `LLVM_ENABLE_PROJECTS`.

So basically... all versions we build a full LLVM with can work with `openmp` being a `runtime` argument.

Crivella: > We may want to limit that, yeah (at least if we allow building libomp.so without building a full LLVM).
LLVM supports openmp for LLVM_ENABLE_RUNTIMES since LLVM 12.0.0-rc1. Prior versions only accept this as a project.

Anything before 18.1.6 will build with minimal only, which includes only base llvm without any of the projects.
So i think this will be safe on that aspect

Flamefire:
> Considering [llvm/llvm-project#136314](https://github.com/llvm/llvm-project/pull/136314) and the fact that on the docs they specify to always build something as a runtime if it is available as both a runtime and a project i think we should definitely do this

Seems odd though, but I understand

> From my understanding while going through the CMake `LIBOMP_OMPT_SUPPORT` is not an option but an internal variable being set after the checks

I see `set(LIBOMP_OMPT_SUPPORT ${OMPT_DEFAULT} CACHE BOOL  "OMPT-support?")` (at least LLVM 20.1.5). It being a `CACHE` variable is a clear sign to me that it is intended as a user setting especially with the `OMPT_DEFAULT` variable being used

> The breaking is in the test suite, and there are many we are already patching out. The problem is that during the offload tests use `-fopenmp` but do not specify where to find `libomp` (which is not in the same place as it used to for the runtime build)

Wouldn't that be something to be fixed upstream? Or are they not seeing this? I'd really like to get some feedback from the LLVM maintainers on what the intended solution is.
And are we **really** sure this same problem won't happen in the installed version if we only patch/symlink the intermediate/build dir?   
This part in particular makes me nervous that's why I'm asking.

Thyre:
> I see set(LIBOMP_OMPT_SUPPORT ${OMPT_DEFAULT} CACHE BOOL  "OMPT-support?") (at least LLVM 20.1.5). It being a CACHE variable is a clear sign to me that it is intended as a user setting especially with the OMPT_DEFAULT variable being used

`LIBOMP_OMPT_SUPPORT` is a cache variable in `openmp`, but not in `offload`. In `offload`, it is only set for `OPENMP_STANDALONE_BUILD`, else it tries to use the variable passed through otherwise.
This is where we observed the issue.

> Wouldn't that be something to be fixed upstream? Or are they not seeing this? I'd really like to get some feedback from the LLVM maintainers on what the intended solution is.

Since their CI runners seemingly don't show the issue, they either do things differently, or do not build both `openmp` and `offload` at the same time? I agree that we should open an issue for this. 

> And are we really sure this same problem won't happen in the installed version if we only patch/symlink the intermediate/build dir?

Since the installed LLVM has the expected infrastructure again, the issue won't occur. Clang should know where to find its libraries. Only the intermediate testing is the issue, because Clang doesn't expect this. This is at least my understanding.

Flamefire:
> > However I see an issue with the if: We can set LIBOMPTARGET_OMPT_SUPPORT but it also depends on the default, which looks wrong. So it boils down to LIBOMP_HAVE_OMPT_SUPPORT which is set by openmp/runtime/cmake/config-ix.cmake based on compile checks and also determines the default for LIBOMP_OMPD_SUPPORT & LIBOMP_OMPT_SUPPORT
> 
> I think the logic behind that section is the following: They want to turn on `LIBOMPTARGET_OMPT_SUPPORT` by default if it is possible for it to be there, but also give you a way to disable it. They still check for `OMPT_TARGET_DEFAULT` because that will be true if omp supports OMPT. They do not want you trying to force `LIBOMPTARGET_OMPT_SUPPORT` if omp does not support it

I guess yes. But that means they ignore the users choice and should rather error as they do for the host OMPT. I opened a PR with them: https://github.com/llvm/llvm-project/pull/146865

Flamefire:
I haven't submitted my last message:

> But LIBOMP_OMPT_SUPPORT should be ON already (default), I'll check what's going on...

Ok, found it: LLVM_PROJECTS are added via `add_subdirectory`, so they share the same variables and cache. LLVM_RUNTIMES are added as a (single) subproject on which the main build depends but is semi-independent.
So all runtimes share a CMakeCache which is another one as the main cache.

So if openmp and offload are both runtimes they can see each others cache and the OMPT variable(s) is correctly set such that support is enabled.


 TLDR: You are right with

> . The CMake cache of the runtimes build is not shared with the one of the projects to my understanding, which was the entire reason the checks for OMP + OMPT where off causing LIBOMPTARGET to not have OMPT support.


> I really do not see why we should not follow the preferred upstream build procedures with a minor fix to get 200 tests pass because libomp is not in the expected location for clang rather than change the entire build system

That wasn't my intention. We should definitely go with this approach of moving openmp to runtimes. And *in addition* set `LIBOMP_OMPT_SUPPORT` such that the build will error out if it would otherwise default to OFF

Flamefire:
> @Flamefire I'm looking at that as a nice-to-have in case you want to add a follow-up PR (or open an issue)

Done https://github.com/easybuilders/easybuild-easyblocks/pull/3853

PR #3803, opened 2025-06-26
Fix issues with PyTorch test-results (XML files) parsing and add tests

Flamefire:
Select some test reports from real runs, trim them down a bit and parse them in the test.

Provide a script for automatic cleanup of test reports:

- Shorten unused values and filenames
- Trim or remove stdout/stderr
- Remove whitespace
- Format XML

The script is fully deterministic so rerunning it won't change the files. The idea is to have realistic looking files for debugging, provide enough reference to original files (e.g. just trim, not replace filenames) and get the file size down

The test files are mostly randomly selected with a few inclusions and modifications to cover (almost) all of the parsing code and reproduce past failures (tested with initial version of the parsing code).   
I also created a few artificial, files to run into most error conditions checked by the parser.

In a recent run I discovered a too strict condition (tests can be skipped before/after rerunning) which I fixed and included a test-testcase

@boegel what do you think about this? I could try reducing the number of test files but I think the current 30 files aren't too much given they are all text only.

### DIFF
--- a/easybuild/easyblocks/g/gnupg_bundle.py
+++ b/easybuild/easyblocks/g/gnupg_bundle.py
@@ -1,0 +1,87 @@
+##
+# Copyright 2021-2025  Ghent University
+#
+# This file is part of EasyBuild,
+# originally created by the HPC team of Ghent University (http://ugent.be/hpc/en),
+# with support of Ghent University (http://ugent.be/hpc),
+# the Flemish Supercomputer Centre (VSC) (https://www.vscentrum.be),
+# Flemish Research Foundation (FWO) (http://www.fwo.be/en)
+# and the Department of Economy, Science and Innovation (EWI) (http://www.ewi-vlaanderen.be/en).
+#
+# https://github.com/easybuilders/easybuild
+#
+# EasyBuild is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation v2.
+#
+# EasyBuild is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with EasyBuild.  If not, see <http://www.gnu.org/licenses/>.
+##
+"""
+Easyconfig for installing a gnupg bundle
+
+The gnupg-bundle is using a UNIX socket path in its tests, specifically for the gpgme component. However, UNIX socket
+paths are limited to 108 characters in length. This easyblock ensures that the build directory is located in a path
+sufficiently small so that the UNIX sockets for the tests can be created.
+
+Source: https://gitlab.archlinux.org/archlinux/archlinux-wsl/-/issues/10
+
+@author: Georgios Kafanas (University of Luxembourg)
+"""
+
+import random
+import time
+from string import ascii_letters
+
+from easybuild.tools.build_log import EasyBuildError, print_msg
+from easybuild.easyblocks.generic.bundle import Bundle
+from easybuild.framework.easyconfig import CUSTOM
+
+
+class EB_gnupg_minus_bundle(Bundle):
+    MAX_UNIX_SOCKET_SAFE_BUILD_PATH_LENGTH = 60
+
+    @staticmethod
+    def extra_options(extra_vars=None):
+        """Easyconfig parameters specific to gnupg-bundle"""
+        extra_vars = Bundle.extra_options(extra_vars=extra_vars)
+        extra_vars.update({
+            'unix_socket_compliant_buildpath': [
+                None,
+                "A build path that ensures that test UNIX socket paths are less than 108 characters long",
+                CUSTOM
+            ]
+        })
+        return extra_vars
+
+    @staticmethod
+    def _get_random_build_path(base_path, prefix):
+        timestamp = time.strftime('%Y%m%d-%H%M%S')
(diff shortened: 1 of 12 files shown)


boegel:
EB_ARGS="PyTorch-2.6.0-foss-2024a.eb --installpath /tmp/$USER/pr3803"
CORE_CNT=16

boegel:
EB_ARGS="PyTorch-2.1.2-foss-2023a-CUDA-12.1.1.eb --installpath /tmp/$USER/pr3803"
CORE_CNT=16

boegel:
@Flamefire Are we now just detecting more failing tests?
```
== 2025-07-31 05:00:52,030 build_log.py:434 WARNING 3 test failures, 0 test errors (out of 209700):
	dynamo/test_functions (1 failed, 165 passed, 2 skipped, 2 rerun)
	dynamo/test_dynamic_shapes (2 failed, 2028 passed, 52 skipped, 32 xfailed, 4 rerun)
```

Flamefire:
Not by this PR at least, maybe a prior one. Does the EC include `PyTorch-2.1.0_skip-dynamo-test_predistpatch.patch`? That should fix the 2nd failure according to my notes.

And/or we can increase the max-failures to like 5...

boegel:
> Not by this PR at least, maybe a prior one. Does the EC include `PyTorch-2.1.0_skip-dynamo-test_predistpatch.patch`? That should fix the 2nd failure according to my notes.

Yes, since the easyconfigs used by the bot are the ones in `develop`.

> 
> And/or we can increase the max-failures to like 5...

@Flamefire I'm actually in favor of bumping the default value of `max_failed_tests` to 10.

That seems like a reasonable number to me, since a warning will always be printed as soon as there's a single failed test.

It's pretty trivial to be more strict locally by introducing a hook that lower `max_failed_tests` to 5, 2, or even 0...

If that makes sense, please open a separate trivial PR for that (just so it stands out in changelog, and so we can add some motivation to the PR description for that change).

Flamefire:
> It's pretty trivial to be more strict locally by introducing a hook that lower `max_failed_tests` to 5, 2, or even 0...

I'd argue the other way round: It is pretty trivial to increase it to 10 ;-)

> If that makes sense, please open a separate trivial PR for that 

I'd just increased the number of this single EC. But we could as well change the default in the easyblock from 0 to 10 and cleanup the easyconfigs.   
Current PyTorch 2 ECs:

> p/PyTorch/PyTorch-2.0.1-foss-2022a.eb:max_failed_tests = 2
> p/PyTorch/PyTorch-2.0.1-foss-2022b.eb:max_failed_tests = 3
> p/PyTorch/PyTorch-2.1.2-foss-2022a.eb:max_failed_tests = 2
> p/PyTorch/PyTorch-2.1.2-foss-2022b.eb:max_failed_tests = 2
> p/PyTorch/PyTorch-2.1.2-foss-2023a-CUDA-12.1.1.eb:max_failed_tests = 2
> p/PyTorch/PyTorch-2.1.2-foss-2023a.eb:max_failed_tests = 2
> p/PyTorch/PyTorch-2.1.2-foss-2023b.eb:max_failed_tests = 2
> p/PyTorch/PyTorch-2.3.0-foss-2023b.eb:max_failed_tests = 6
> p/PyTorch/PyTorch-2.6.0-foss-2024a.eb:max_failed_tests = 16

We seem to be fine with "3" (including the current failure here) for the majority and I'd keep the number as low as possible even though I've been more generous with the new 2.6 EC.

> > Not by this PR at least, maybe a prior one. Does the EC include `PyTorch-2.1.0_skip-dynamo-test_predistpatch.patch`? That should fix the 2nd failure according to my notes.
> 
> Yes, since the easyconfigs used by the bot are the ones in `develop`.

Then we'd need to look at the logs to see why it failed. But 2/2030 seems to be OK to ignore.

boegel:
```
12 test failures, 0 test errors (out of 254957):
Failed tests (suites/files):
	dynamo/test_dynamic_shapes (3 failed, 1735 passed, 76 skipped, 0 errors)
	dynamo/test_inline_inbuilt_nn_modules (3 failed, 1214 passed, 30 skipped, 0 errors)
	dynamo/test_misc (3 failed, 522 passed, 22 skipped, 0 errors)
	inductor/test_cpu_select_algorithm (2 failed, 124 passed, 1211 skipped, 0 errors)
	inductor/test_torchinductor_opinfo (1 failed, 3016 passed, 590 skipped, 0 errors)
Counted failures of tests from the following test suites/files that are not contained in the summary output of PyTorch:
inductor/test_torchinductor_opinfo (at easybuild/easyblocks/pytorch.py:747 in test_step)
```

@Flamefire More than before?

Flamefire:
The failing install is
> Failing because there were unexpected failures detected: inductor/test_torchinductor_opinfo

Likely due to
> The following tests failed and then succeeded when run in a new process['test/inductor/test_torchinductor_opinfo.py::TestInductorOpInfoCPU::test_comprehensive_nn_functional_max_pool2d_cpu_float16']

For that the success isn't detected after detecting the failure.  (A test where reports have success and failure should be merged as a success result by this parsing logic)   
 I'd need the XML folder and/or need to double check why that was missed.  
So looks like a fluke only.

boegel (approved):
This has been tested extensively with both existing PyTorch easyconfigs like `PyTorch-2.1.2-foss-2023b-CUDA-12.1.1.eb` and `PyTorch-2.6.0-foss-2024a.eb`, as well as easyconfigs for newer PyTorch versions (cfr. https://github.com/easybuilders/easybuild-easyconfigs/pull/23923 for PyTorch 2.7.1 and https://github.com/easybuilders/easybuild-easyconfigs/pull/24365 for PyTorch 2.9.1).

Another small step towards taming the PyTorch beast...

PR #3816, opened 2025-07-01
clean installation directory instead of hard removing it in various easyblocks

smoors:
depends on:
- https://github.com/easybuilders/easybuild-framework/pull/4932


boegel:
I won't merge this as is at this point, since we're too close to the EasyBuild v5.1.2, and this hasn't seen any testing yet.

@smoors Maybe we should pull this apart into smaller PRs which are easier to test & get merged?

PR #3823, opened 2025-07-03
new easyblock for rocm_llvm

Thyre:
(created using `eb --new-pr`)


easybuild/easyblocks/r/rocm_llvm.py
```diff
@@ -0,0 +1,122 @@
...
+    """
+    Support for building the ROCm-LLVM compilers with some modifications on top of the LLVM easyblock.
+    """
+
+    def _configure_general_build(self):
+        super(EB_ROCm_minus_LLVM, self)._configure_general_build()
+        self._cmakeopts.update({
+            'LLVM_EXTERNAL_PROJECTS': '"device-libs"',
+            'LLVM_EXTERNAL_DEVICE_LIBS_SOURCE_DIR': os.path.join(self.llvm_src_dir, 'amd', 'device-libs'),
+            'LLVM_ENABLE_PER_TARGET_RUNTIME_DIR': 'ON',
```
Zeldhoron, line 56:
```suggestion
            'LLVM_ENABLE_PER_TARGET_RUNTIME_DIR': 'ON',
            'LLVM_TARGETS_TO_BUILD': '"X86;AMDGPU"',
            'CLANG_ENABLE_AMDCLANG': 'ON',
```

This fixes an issue with the ROCm-LLVM easyconfig I had, where the build failed because the AMDGPU target was not available during a build step.

easybuild/easyblocks/r/rocm_llvm.py
```diff
@@ -0,0 +1,122 @@
...
+                                      [(r'set\(CMAKE_CXX_COMPILER ${CMAKE_BINARY_DIR}/bin/clang\+\+\)',
+                                        'set(CMAKE_CXX_COMPILER %s)' % mock_clangxx)])
+
+        super(EB_ROCm_minus_LLVM, self).build_with_prev_stage(prev_dir, stage_dir)
+
+    def _configure_final_build(self):
+        super(EB_ROCm_minus_LLVM, self)._configure_final_build()
+        self._cmakeopts.update({
+            'LIBOMP_OMPD_SUPPORT': 'ON',
+            'CLANG_ENABLE_AMDCLANG': 'ON',
```
Zeldhoron:
```suggestion
```

Becomes unnecessary if an earlier suggestion of mine is included.

easybuild/easyblocks/r/rocm_llvm.py
```diff
@@ -0,0 +1,122 @@
...
+            'LLVM_EXTERNAL_PROJECTS': '"device-libs"',
+            'LLVM_EXTERNAL_DEVICE_LIBS_SOURCE_DIR': os.path.join(self.llvm_src_dir, 'amd', 'device-libs'),
+            'LLVM_ENABLE_PER_TARGET_RUNTIME_DIR': 'ON',
+            'CLANG_DEFAULT_RTLIB': 'compiler-rt',
+            'CLANG_DEFAULT_UNWINDLIB': 'libgcc',
+            'DEFAULT_ROCM_PATH': self.installdir,
+            'LIBOMP_COPY_EXPORTS': 'OFF',
+        })
+
+        amd_gfx_list = build_option('amdgcn_capabilities') or self.cfg['amdgcn_capabilities'] or []
```
Zeldhoron:
```suggestion
        amd_gfx_list = build_option('amdgcn_capabilities', default=[])
        if not amd_gfx_list and 'amdgcn_capabilities' in self.cfg:
            amd_gfx_list = self.cfg['amdgcn_capabilities']
        if not amd_gfx_list and 'AMDGCN_CAPABILITIES' in os.environ:
            amd_gfx_list = os.environ.get('AMDGCN_CAPABILITIES').split(',')
```

Allows us to test/use this easyblock with the current version of EasyBuild (v5.1.1) by using an environment variable instead of a build option.

Zeldhoron (dismissed):
During testing this, when I was trying to build ROCm-LLVM, I added some small changes that fixed some issues for me.

easybuild/easyblocks/r/rocm_llvm.py
Thyre: Hm, we might want to set this as a default when no GPU capabilities are passed, which would enable `AMDGPU` automatically with https://github.com/easybuilders/easybuild-easyblocks/pull/3824.

However, hard-coding these values would block building ROCm-LLVM with NVPTX support, which should theoretically work, and can be done with the AOMP scripts (though I don't know if this is actually well tested).

easybuild/easyblocks/r/rocm_llvm.py
Thyre: Agreed. Will move it, together with handling your comment above.

easybuild/easyblocks/r/rocm_llvm.py
```diff
@@ -0,0 +1,126 @@
...
+        })
+
+        amd_gfx_list = build_option('amdgcn_capabilities', default=[])
+        if not amd_gfx_list and 'amdgcn_capabilities' in self.cfg:
+            amd_gfx_list = self.cfg['amdgcn_capabilities']
+        if not amd_gfx_list and 'AMDGCN_CAPABILITIES' in os.environ:
+            amd_gfx_list = os.environ.get('AMDGCN_CAPABILITIES').split(',')
+        if not amd_gfx_list:
+            raise EasyBuildError("Expected amdgcn_capabilities to be set to build this EasyConfig. "
+                                 "Please specify either --amdgcn_capabilities, or set amdgcn_capabilities "
```
Zeldhoron:
```suggestion
                                 "Please specify either --amdgcn-capabilities, or set amdgcn_capabilities "
```

A small typo

Thyre: Thanks 😄

easybuild/easyblocks/r/rocm_llvm.py
Thyre: Changed so that `AMDGPU` is added if not present, but existing values are not changed. With this, one _could_ try to pass NVPTX as a target to build.

Thyre:
First test report:
https://github.com/easybuilders/easybuild-easyconfigs/pull/23304#issuecomment-3126676436

easybuild/easyblocks/r/rocm_llvm.py
```diff
@@ -0,0 +1,133 @@
...
+    def build_with_prev_stage(self, prev_dir, stage_dir):
+        # Similar handling to case above, just for multi-stage build.
+        # Here, we need to create mock wrappers ourselves, as call to LLVM build will start the build process.
+        if build_option('rpath'):
+            self._prepare_runtimes_rpath_wrappers(stage_dir)
+            mock_clangxx = which('clang++')
+            amdllvm_cmakelists = os.path.join(self.llvm_src_dir, 'clang-tools-extra', 'amdllvm', 'CMakeLists.txt')
+            apply_regex_substitutions(amdllvm_cmakelists,
+                                      [(r'set\(CMAKE_CXX_COMPILER ${CMAKE_BINARY_DIR}/bin/clang\+\+\)',
+                                        'set(CMAKE_CXX_COMPILER %s)' % mock_clangxx)])
```
Crivella, line 132: This seems to be substituting on top of the sources,  after the 1st stage i would expect here something like
`set(CMAKE_CXX_COMPILER /tmp/XXX/YYY/clang)`

are you sure this regex is replacing this again and not always pointing to the mock wrappers of the 1st stage?

Thyre: Right, the first one (in `configure_step`) should only be replaced if we don't have a bootstrap build.

Thyre: Hm... wait, we pass `device-libs` for all stages. So for proper rpathing, we should also replace it in the first stage.

Let me check what we to in the LLVM EasyBlock between stages...

Thyre: Okay, we extract the sources only once.
Therefore, we need to change the replacement here to change from the replacement above to the current stage dir.

Thyre: The latest commit should try to resolve this, by storing the original `CMakeLists.txt`, and restoring it before trying to do a replacement again.

Needs testing though...

bedroge:
@Thyre I was giving this another try with the easyconfig from https://github.com/easybuilders/easybuild-easyconfigs/pull/23304, but I'm getting an error for `openmp`:
```
 == installing bundle component llvm-project-openmp v19.0.0-rocm-6.4.1 (4/5)...
==    patching component llvm-project-openmp...
  >> applying patch ROCm-LLVM-6.4.1_llvm-project-19.0.0_fix-offload-build.patch
==    ... (took < 1 sec)
==    configuring component llvm-project-openmp...
==    ... (took 6 secs)
== ... (took 46 mins 19 secs)
== FAILED: Installation ended unsuccessfully: Nothing found to replace 'cmake_policy\(SET CMP0114 OLD\)' in cmake/Modules/CMakePolicy.cmake (took 46 mins 35 secs)
```
Looks like that's because of the changes from https://github.com/easybuilders/easybuild-easyblocks/pull/3875 (which uses `on_missing_match=ERROR` for the regex substitution), and because we're reusing the llvm source dir for openmp. The file has already been patched then, so there's nothing to replace anymore, and it fails.

I guess we either need to use a clean source dir for openmp, or disable the `on_missing_match=ERROR` (or make that optional somehow)?

Thyre:
> @Thyre I was giving this another try with the easyconfig from https://github.com/easybuilders/easybuild-easyconfigs/pull/23304, but I'm getting an error for `openmp`:
> 
> ```
> 
>  == installing bundle component llvm-project-openmp v19.0.0-rocm-6.4.1 (4/5)...
> ==    patching component llvm-project-openmp...
>   >> applying patch ROCm-LLVM-6.4.1_llvm-project-19.0.0_fix-offload-build.patch
> ==    ... (took < 1 sec)
> ==    configuring component llvm-project-openmp...
> ==    ... (took 6 secs)
> == ... (took 46 mins 19 secs)
> == FAILED: Installation ended unsuccessfully: Nothing found to replace 'cmake_policy\(SET CMP0114 OLD\)' in cmake/Modules/CMakePolicy.cmake (took 46 mins 35 secs)
> 
> ```
> [...]
> 
> I guess we either need to use a clean source dir for openmp, or disable the `on_missing_match=ERROR` (or make that optional somehow)?

I ran into this with ROCm 7.0.0 in the first LLVM build and created #3937 to work around the issue for that. I didn't keep in mind that this also affects the second build, since the sources are patched at that point already. 

I think starting from a clean source is the best solution, since there might be other unexpected side-effects otherwise. I'll update the ROCm-LLVM EasyConfig PRs accordingly.

Thyre:
@bedroge, I updated the ROCm-LLVM 6.4.1 PR to extract the sources again during `openmp` and triggered a build on `jsc-zen3`. Lets see if that is sufficient...

bedroge:
> @bedroge, I updated the ROCm-LLVM 6.4.1 PR to extract the sources again during `openmp` and triggered a build on `jsc-zen3`. Lets see if that is sufficient...

Thanks, I'll do another build as well.

PR #3849, opened 2025-07-16
Add explicit enabling/disabling of `libffi` with custom paths

### LINKED ISSUE #3848: LLVM: Failures in building `offload` with `ffi.h` in non-standard location
If `libffi` is detected but the header files are in a non standard location, it is possible the build of the host plugin for `offload` will fail

Reported from slack @bartoldeman 

```
/tmp/oldeman/avx2/LLVM/20.1.5/GCCcore-13.3-gentoo/llvm-project-20.1.5.src/offload/plugins-nextgen/host/src/rtl.cpp:15:10: fatal error: 'ffi.h' file not found
   15 | #include <ffi.h>
      |          ^~~~~~~
1 error generated.
make[5]: *** [offload/plugins-nextgen/host/CMakeFiles/omptarget.rtl.host.dir/build.make:79: offload/plugins-nextgen/host/CMakeFiles/omptarget.rtl.host.dir/src/rtl.cpp.o] Error 1
```

Crivella:
Check if `libffi` is a dependency, if not disable using it or if yes explicitly add the location for `FFI_INCLUDE_DIR` and `FFI_LIBRARY_DIR`

- Fixes #3848

### DIFF
--- a/easybuild/easyblocks/l/llvm.py
+++ b/easybuild/easyblocks/l/llvm.py
@@ -748,6 +748,12 @@ class EB_LLVM(CMakeMake):
         else:
             self.general_opts['LLVM_ENABLE_LIBXML2'] = 'OFF'
 
+        libffi_root = get_software_root('libffi')
+        self.general_opts['LLVM_ENABLE_LIBFFI'] = 'ON' if libffi_root else 'OFF'
+        if libffi_root:
+            self.general_opts['FFI_INCLUDE_DIR'] = os.path.join(libffi_root, 'include')
+            self.general_opts['FFI_LIBRARY_DIR'] = os.path.join(libffi_root, 'lib64')
+
         # If 'ON', risk finding a system zlib or zstd leading to including /usr/include as -isystem that can lead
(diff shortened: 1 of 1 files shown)


ocaisa (approved):
Testing done in https://github.com/easybuilders/easybuild-easyconfigs/pull/23429

While there may be some lingering potential issues, they _may_ only be realised in a scenario we don't have easyconfigs for.

easybuild/easyblocks/l/llvm.py
```diff
@@ -748,6 +748,12 @@ def configure_step(self):
         else:
             self.general_opts['LLVM_ENABLE_LIBXML2'] = 'OFF'
 
+        libffi_root = get_software_root('libffi')
```
Thyre, line 751: We should've probably checked if `libffi` is a dependency, right (via `self.cfg.dependencies()`)?
I think the failure in https://github.com/easybuilders/easybuild-easyconfigs/pull/23603#issuecomment-3168186067 may well be caused by this change (though I don't know why not on `jsc-zen3`)

PR #3824, opened 2025-07-03
make LLVM easyblock aware of `amdgcn_capabilities` build option

Thyre:
**Requires:**
- [x] https://github.com/easybuilders/easybuild-framework/pull/4860

This allows building LLVM with AMD GPU support without requiring to adapt the LLVM EasyConfig beforehand. Deprecate `amd_gfx_list`, but still accept the value if the new build option is not set.

PR #3851, opened 2025-07-22
LLVM: Fix missing `LIBOMPTARGET_DEVICE_ARCHITECTURES` in third stage of bootstrap build

Thyre:
By using `_cmakeopts`, the argument was only passed for the first stage of the build, and ignored for subsequent stages. We therefore build all available architectures for AMDGPU / NVPTX when not necessary. Fix this by moving the argument to `general_opts`.

### DIFF
--- a/easybuild/easyblocks/l/llvm.py
+++ b/easybuild/easyblocks/l/llvm.py
@@ -385,7 +385,7 @@ class EB_LLVM(CMakeMake):
         self.log.info("Final projects to build: %s", ', '.join(self.final_projects))
         self.log.info("Final runtimes to build: %s", ', '.join(self.final_runtimes))
 
-        # CMake options passed to each building stage.
+        # CMake options passed to each build stage.
         # Will be cleared between stages. If arguments are needed in multiple stages,
         # consider adding them to general_opts instead.
         self._cmakeopts = {}
--- a/easybuild/easyblocks/l/llvm.py
+++ b/easybuild/easyblocks/l/llvm.py
@@ -821,7 +821,7 @@ class EB_LLVM(CMakeMake):
             gpu_archs = self.cfg.get_cuda_cc_template_value("cuda_sm_space_sep", required=False).split()
             gpu_archs += self.amd_gfx
             if gpu_archs:
-                self._cmakeopts['LIBOMPTARGET_DEVICE_ARCHITECTURES'] = self.list_to_cmake_arg(gpu_archs)
+                self.general_opts['LIBOMPTARGET_DEVICE_ARCHITECTURES'] = self.list_to_cmake_arg(gpu_archs)
 
         self._configure_general_build()
         self.add_cmake_opts()
--- a/easybuild/easyblocks/l/llvm.py
+++ b/easybuild/easyblocks/l/llvm.py
@@ -291,6 +291,10 @@ class EB_LLVM(CMakeMake):
         if self.cfg['use_pic']:
             on_opts.append('CMAKE_POSITION_INDEPENDENT_CODE')
 
+        # General options being passed to every build stage.
+        # Here, options that will be required in all build stages should be added.
+        # Update _cmakeopts in _configure_{general,intermediate,final}_build if
+        # build option is only relevant for a single build step.
         self.general_opts = GENERAL_OPTS.copy()
 
         for opt in on_opts:
@@ -381,6 +385,9 @@ class EB_LLVM(CMakeMake):
         self.log.info("Final projects to build: %s", ', '.join(self.final_projects))
         self.log.info("Final runtimes to build: %s", ', '.join(self.final_runtimes))
 
+        # CMake options passed to each building stage.
+        # Will be cleared between stages. If arguments are needed in multiple stages,
+        # consider adding them to general_opts instead.
         self._cmakeopts = {}
         self._cfgopts = list(filter(None, self.cfg.get('configopts', '').split()))
 


Thyre:
EB_ARGS="LLVM-19.1.7-GCCcore-13.3.0.eb LLVM-20.1.5-GCCcore-13.3.0.eb"
CORE_CNT=16

Flamefire (approved):
Good catch. Maybe add documentation to those 2 members because it is hard to tell what the difference is which led to this issue

Thyre:
> Good catch. Maybe add documentation to those 2 members because it is hard to tell what the difference is which led to this issue

Good idea. I'll try to find a place where we can document this.
Just noticed that while testing #3824, because I saw all the `amdgpu` options but not `gfx1101`.

Thyre:
It fortunately isn't a really noticeable regression. We just build a few architectures too much.
It's really easy to miss some things when dealing with such a huge EasyBlock.

I wonder though if it would be worth adding tests to avoid this from happening again.
Might be complicated though, as we've seen e.g. the NVPTX bitcode library being built even without NVPTX support.

easybuild/easyblocks/l/llvm.py
```diff
@@ -381,6 +385,9 @@ def __init__(self, *args, **kwargs):
         self.log.info("Final projects to build: %s", ', '.join(self.final_projects))
         self.log.info("Final runtimes to build: %s", ', '.join(self.final_runtimes))
 
+        # CMake options passed to each building stage.
```
Flamefire: For consistency with above:
```suggestion
        # CMake options passed to each build stage.
```

Flamefire (approved):

Crivella (approved):
LGTM, tested both here and in https://github.com/easybuilders/easybuild-easyconfigs/pull/23459

easybuild/easyblocks/l/llvm.py
```diff
@@ -381,6 +385,9 @@ def __init__(self, *args, **kwargs):
         self.log.info("Final projects to build: %s", ', '.join(self.final_projects))
         self.log.info("Final runtimes to build: %s", ', '.join(self.final_runtimes))
 
+        # CMake options passed to each build stage.
+        # Will be cleared between stages. If arguments are needed in multiple stages,
+        # consider adding them to general_opts instead.
         self._cmakeopts = {}
```
Flamefire, line 391: Too late for this PR now, but I thought we could set this to `None` at this point such that adding dict entries to this too early would fail loudly.

What do you think?

Thyre: Probably a good idea. We can either add this in a separate PR, or just add it to one of the remaining LLVM ones.

Flamefire: I think https://github.com/easybuilders/easybuild-easyblocks/pull/3799 would be a good candidate as IMO that is ready.

@Crivella ?

Crivella: I think we are already setting it at the edge where we want to access it. If you try to use it before in the init it will be non-existent and fail regardless

Flamefire: I think it must only be used after the `_configure_general_build` call in `configure_step`: https://github.com/Thyre/easybuild-easyblocks/blob/9771da4f8969ba373c19b99a6988b3b65113d225/easybuild/easyblocks/l/llvm.py#L833

There cmakeopts should be initialized to a copy of `generalopts`. Or is there any earlier use intended? Maybe derived easyblocks adding options to that before calling the super-method?

Crivella: Umh not necessarily as `_configure_general_build` does not reset them but runs an update
````
    def _configure_general_build(self):
        """General configuration step for LLVM."""
        self._cmakeopts.update(self.general_opts)
        self._add_cmake_runtime_args()
````

I agree that in general `_cmakeopts` should not be touched manually as much as possible, i would say that having it as an ~hidden attribute with the `_`  should be enough.

If we wanted to improve clarity we should add another `stage_opts` dictionary that is reset at the end of every stage and added at the start of one, but not sure it is warranted the effort and testing

Comment (Flamefire) [reply easybuild/easyblocks/l/llvm.py:391]: > Umh not necessarily as `_configure_general_build` does not reset them but runs an update

But that is not what it should be semantically, is it? Isn't the behavior we want that this function initializes the current options (`_cmakeopts`) and afterwards additional options can be added as relevant for the other stages?

And actually it isn't even an update: Prior to that function we usually clear the dict so this line is partially misleading.

> i would say that having it as an ~hidden attribute with the _ should be enough.

I'm just thinking about how to avoid reintroducing issues like the one fixed here where the attribute is touched by code in this class which is "valid" for hidden attributes.

Comment (Crivella) [reply easybuild/easyblocks/l/llvm.py:391]: Not in the way it used right now, having `_configure_general_build` reset `_cmakeopts` would nullify the changes done by
```
        if self.cfg['bootstrap']:
            self._configure_intermediate_build()
        else:
            self._configure_final_build()
```
in the configure step

Right now the `_configure_...` methods are used as utilities to add to `_cmake_opts`

The idea of the EB right now is to use `configure_step` as if it was `configure_step1` vs the `configure_step2/3` counterparts.

The only thing i could see is resetting/first-initializing `_cmakeopts` at the start of that as we do for the others (agree might not make a lot of sense to set them in the steps before), but i think it should be a valid variable throughout all of the configure step

> I'm just thinking about how to avoid reintroducing issues like the one fixed here where the attribute is touched by code in this class which is "valid" for hidden attributes.

In general i would say that having `general_opts` as options used for every stage while `_cmakeopts` is valid only for the current stage is fine, i think there is a greater chance of reintroducing errors if we refactor again, vs reviewing changes in the future.

Flamefire: > Right now the `_configure_...` methods are used as utilities to add to `_cmake_opts`

Hm, I would expect to have `general` always before specific `configure...` calls which is always the case except for that one instance. So I guess it would make sense to move it further below to https://github.com/Thyre/easybuild-easyblocks/blob/9771da4f8969ba373c19b99a6988b3b65113d225/easybuild/easyblocks/l/llvm.py#L833 to be in line with other usages

> The idea of the EB right now is to use `configure_step` as if it was `configure_step1` vs the `configure_step2/3` counterparts.

But that is violated by that switched order, isn't it? https://github.com/Thyre/easybuild-easyblocks/blob/9771da4f8969ba373c19b99a6988b3b65113d225/easybuild/easyblocks/l/llvm.py#L881-L883

> but i think it should be a valid variable throughout all of the configure step

Then we could unexpectedly overwrite an intentional options by the `_configure_general_build` 

> In general i would say that having `general_opts` as options used for every stage while `_cmakeopts` is valid only for the current stage is fine, i think there is a greater chance of reintroducing errors if we refactor again, vs reviewing changes in the future.

IMO there are still more surprises compared to "`_cmakeopts` must only be used after a call to `_configure_general_build` which sets it to (only) the defaults."   
I'd like to reduce them as it was missed in a change and the review before. So to me those additional safeguards do make sense

I'd do that change in a PR that needs testing against a couple ECs anyway so the chance of catching bugs by this change (even though I don't expect any) is high enough.

Crivella: > I'd like to reduce them as it was missed in a change and the review before. So to me those additional safeguards do make sense
TBH that merge was quite rushed i feel. Also that error would not have been caught by this suggested change

In general i think that if there is a clash from what is coming from `general_opts` and `_cmakeotps` that would be a problem in itself as it means one of the 2 is not being used as intended, for that we might add a check to print a warning in `_configure_general_build` if there is any overlap.

> But that is violated by that switched order, isn't it?
Agree that could probably safely also go below the call to `_configure_general_build`

Flamefire: >  TBH that merge was quite rushed i feel. Also that error would not have been caught by this suggested change

if `self._cmakeopts` was set to `None` in `__init__` and to a copy of `general_opts` this particular instance actually would have been caught. And with the full, overwriting assignment of `cmakeopts` in `_configure_general_build` it could have been more clear how the 2 options are meant to be used, e.g. when encountering the error (of subscripting `None`)

Crivella: Only if `_cmakeopts` was not an attribute there which i still think it should.

EG what if we want a package based configuration like all the ones that happen in the `configure_step` to not modify `general_opts` but only do something for the first step?

Best think i could come up with if we wanted to improve clarity is to never allow `_cmakeopts` to be touched and add a method to explicitly add to it like `add_stage_cmake_configuration_param` (or something like it)

Flamefire: > Only if `_cmakeopts` was not an attribute there which i still think it should.

If you set `self._cmakeopts = None` in `__init__` it will still be an attribute, won't it?

> EG what if we want a package based configuration like all the ones that happen in the `configure_step` to not modify `general_opts` but only do something for the first step?

Then we add to `self._cmakeopts` after the call to `_configure_general_build` as done in all other configure-steps

> Best think i could come up with if we wanted to improve clarity is to never allow `_cmakeopts` to be touched and add a method to explicitly add to it like `add_stage_cmake_configuration_param` (or something like it)

A bit better, but we can't enforce it not being touched. But: We should rename `_cmakeopts` to `_stage_cmake_opts` or similar. That likely achieves the same clarity as that additional method

Crivella: > If you set self._cmakeopts = None in __init__ it will still be an attribute, won't it?

That really only change the type of exception you would get from an `AttributeError` if it is not set to a `TypeError` if it is set to `None` and you try to assign a key to it

Flamefire: Then I dont understand your remark:

> Only if `_cmakeopts` was not an attribute there which i still think it should.

Getting an error is intended to detect using it too early. And IMO the explicit `None` shows that it does not exist yet and provides means to add a comment what it is and where it will be set.

Crivella: What i meant is that i think `_cmakeopts` should be already usable in the configure step.

I think unless we decide do alter the logic quite a bit this will not really save us from PRs using the wrong attributes/methods if we do not catch it in review...
What i think is more worth the effort is increase the sanity checks to make sure all the feature we want on are on.

Flamefire: But should it be usable before `_configure_general_build` in the configure step?

I'd just make it as hard as possible to use the wrong one if it doesn't impact what we really need and my suggestion is a step towards it while not impacting current usage as far as I can tell.

Better sanity checks would be good either way, yes.

Then maybe the minimal common ground would be renaming the generically named `cmakeopts` to something more specific like "_stage_cmake_opts"? That would make it easier to review without any real change

Crivella: > But should it be usable before _configure_general_build in the configure step?

I would say so, otherwise we would need a section for updating `general_opts` based on present packages (what we have right now before `_configure_general_build` in the `configure_step` and one to update the single stage after.
I feel that having 2 section there could also lead to confusion and people misplacing stuff (eg modifying `general_opts` after the call to `_configure_general_build`

> Then maybe the minimal common ground would be renaming the generically named cmakeopts to something more specific like "_stage_cmake_opts"? That would make it easier to review without any real change
Not against renaming variables for clarity, might even make it `_current_stage_cmake_opts` for clarity

I think the only way to really make sure stuff is not misused is to introduce 4 methods:
- add_stage1_cmake_opts
- add_stage2_cmake_opts
- add_stage3_cmake_opts
- add_all_stages_cmake_opts

And than we could ensure they are never called after a particular stage has already been ran, and that the internal methods are never touched manually.
This would require quite a lot of refactoring, and high possibility to introduce new errors which i would avoid for now

PS would also require rewriting the logic in how the `configopts` coming from the EC file now are treated as `general_opts`  + would need to streamline and clarify how stages are treated in the case of `bootstrap` vs not

Flamefire: > I think the only way to really make sure stuff is not misused is to introduce 4 methods:

Would it then need to be a property at all? The current member methods that additionally modify it could just a return a dict that is then used to update a local cmakeopts

PR #3853, opened 2025-07-25
Explicitely set OMPT options for LLVM

Flamefire:
As discussed [here](https://github.com/easybuilders/easybuild-easyblocks/pull/3799#issuecomment-3035168311) it makes sense to set `LIBOMP_OMPT_SUPPORT` as that will cause a configure error instead of not building the tools interface with only a small configure message

Additionally also set the related `LIBOMPTARGET_OMPT_SUPPORT` option

Note that our definition of `build_openmp_tools` is not that of `OPENMP_ENABLE_OMPT_TOOLS` but rather matches `LIBOMP_OMPT_SUPPORT`, see https://github.com/llvm/llvm-project/blob/adb2421202e4014b4986a9f1eb40833dc7ec25ad/openmp/README.rst#optional-features

> LIBOMP_OMPT_SUPPORT = ON|OFF
    Include support for the OpenMP Tools Interface (OMPT). 
>
> option(OPENMP_ENABLE_OMPT_TOOLS "Enable building ompt based tools for OpenMP."

So we enable more than just the interface, but I don't think we should change that.

Question: Should we also explicitely enable the optional OMPT features?
> LIBOMP_OMPT_OPTIONAL = ON|OFF
    Include support for optional OMPT functionality.

It currently defaults to ON, so we can ignore it.

### DIFF
--- a/easybuild/easyblocks/l/llvm.py
+++ b/easybuild/easyblocks/l/llvm.py
@@ -805,11 +805,6 @@ class EB_LLVM(CMakeMake):
         self.general_opts.update(python_opts)
         self.runtimes_cmake_args.update(python_opts)
 
-        if self.cfg['bootstrap']:
-            self._configure_intermediate_build()
-        else:
-            self._configure_final_build()
-
         if self.cfg['skip_sanitizer_tests'] and build_option('strict') != ERROR:
             self.log.info("Disabling the sanitizer tests")
             self.disable_sanitizer_tests()
@@ -855,6 +850,11 @@ class EB_LLVM(CMakeMake):
                 self.general_opts['LIBOMPTARGET_DEVICE_ARCHITECTURES'] = self.list_to_cmake_arg(gpu_archs)
 
         self._configure_general_build()
+        if self.cfg['bootstrap']:
+            self._configure_intermediate_build()
+        else:
+            self._configure_final_build()
+
         self.add_cmake_opts()
 
         src_dir = os.path.join(self.start_dir, 'llvm')
--- a/easybuild/easyblocks/l/llvm.py
+++ b/easybuild/easyblocks/l/llvm.py
@@ -874,7 +874,7 @@ class EB_LLVM(CMakeMake):
                 self.log.warning("`LLVM_HOST_TRIPLE` not found in the output of the configure step")
 
         if not self.cfg['bootstrap']:
-            if build_option('rpath') and self._cmakeopts['LLVM_ENABLE_RUNTIMES'] != '""':
+            if build_option('rpath') and self._cmakeopts['LLVM_ENABLE_RUNTIMES'] not in ('""', "''", ''):
                 # Ensure RPATH wrappers are used for the runtimes also at the first stage
                 # Call configure again now that the host triple is known from the previous configure call
                 remove_dir(self.llvm_obj_dir_stage1)
--- a/easybuild/easyblocks/l/llvm.py
+++ b/easybuild/easyblocks/l/llvm.py
@@ -1371,7 +1371,7 @@ class EB_LLVM(CMakeMake):
             for suffix in ('.c', '.o', '.x'):
                 remove_file(f'{test_fn}{suffix}')
 
-    def sanity_check_step(self, *args, custom_paths=None, custom_commands=None, **kwargs):
+    def sanity_check_step(self, custom_paths=None, custom_commands=None, *args, **kwargs):
         """Perform sanity checks on the installed LLVM."""
         lib_dir_runtime = None
         if self.cfg['build_runtimes']:
--- a/easybuild/easyblocks/l/llvm.py
+++ b/easybuild/easyblocks/l/llvm.py
@@ -575,8 +575,13 @@ class EB_LLVM(CMakeMake):
                         self._cmakeopts['LIBOMPTARGET_FORCE_DLOPEN_LIBCUDA'] = 'ON'
             self._cmakeopts['OPENMP_ENABLE_LIBOMPTARGET'] = 'ON'
             self._cmakeopts['LIBOMP_INSTALL_ALIASES'] = 'ON' if self.cfg['build_openmp_library_aliases'] else 'OFF'
-            if not self.cfg['build_openmp_tools']:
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/l/llvm.py
```diff
@@ -1366,7 +1371,7 @@ def _sanity_check_dynamic_linker(self):
             for suffix in ('.c', '.o', '.x'):
                 remove_file(f'{test_fn}{suffix}')
 
-    def sanity_check_step(self, custom_paths=None, custom_commands=None, *args, **kwargs):
+    def sanity_check_step(self, *args, custom_paths=None, custom_commands=None, **kwargs):
```
Crivella: Since those are not forced to be keyword-only arguments in `framework` i do not think we should force them here

easybuild/easyblocks/l/llvm.py
```diff
@@ -869,7 +874,7 @@ def configure_step(self):
                 self.log.warning("`LLVM_HOST_TRIPLE` not found in the output of the configure step")
 
         if not self.cfg['bootstrap']:
-            if build_option('rpath') and self._cmakeopts['LLVM_ENABLE_RUNTIMES'] != '""':
+            if build_option('rpath') and self._cmakeopts['LLVM_ENABLE_RUNTIMES'] not in ('""', "''", ''):
```
Crivella: Since the expected value is tied to the result of `list_to_cmake_arg` we might want to do

```suggestion
            if build_option('rpath') and self._cmakeopts['LLVM_ENABLE_RUNTIMES'] != self.list_to_cmake_arg([]):
```

Alternatively we could just check `self.final_runtimes` here

```suggestion
            if build_option('rpath') and self.final_runtimes:
```

easybuild/easyblocks/l/llvm.py
Flamefire: Then `*args` should be removed or funny things might happen according to [PyLint](https://pylint.readthedocs.io/en/latest/user_guide/messages/warning/keyword-arg-before-vararg.html):

> When defining a keyword argument before variable positional arguments,
> one can end up in having multiple values passed for the aforementioned parameter
> in case the method is called with keyword arguments.

TBH: I don't think we need either at all. Other easyblocks don't have any params or just args&kwargs to pass through.

easybuild/easyblocks/l/llvm.py
Flamefire: The 2nd approach wouldn't work: The reason cmakeopts is checked is because it might be final or intermediate runtimes at this point depending on bootstrapping

Crivella: If `bootstrap` is False it is always gonna be the final one, still no easy safeguard if the logics get changed in the future

Flamefire: Oh right, this is also guarded. Then I prefer the 2nd approach.

> still no easy safeguard if the logics get changed in the future

I don't fully get this. So keep it or use `self.final_runtimes` which IMO is clearer?

easybuild/easyblocks/l/llvm.py
Crivella: What pylint is complaining about is the fact that you can passe the same argument twice leading to a `TypeError`

```
>>> def a(var=0, *args, **kwargs):
...   print(var)
... 
>>> a(2, var=2)
TypeError: a() got multiple values for argument 'var'
```

Problem being that this would preclude any parent class using the same API as defined in framework.

On the other hand we overwrite those parameters, so we could remove them from the function definition and make sure they are also not present in `kwargs` or pass them to super through `kwargs`.
If we were to remove from the funciton definition without this other change we would risk the other side of the coin of `custom_paths` being in `kwargs` and us passing it twice

```
>>> def b(var=0, **kwargs):
...   a(var2=0, **kwargs)
... 
>>> b(var2=2)
  File "<stdin>", line 2, in b
TypeError: __main__.a() got multiple values for keyword argument 'var2'
>>> 
```

Flamefire: We should just be using the same signature as the parent class or only `*args, **kwargs` and pass them through.   
With the current (updated) approach we will run into that TypeError when those are passed as positional args. So that needs to be changed again.


So I'd go with `def sanity_check_step(self, custom_paths=None, custom_commands=None)`.
There is even a pylint check because if an easyblock changed the signature of this then the framework might misbehave. E.g. `def sanity_check_step(self, new_param=0, custom_paths=None, custom_commands=None)` would cause assigning the wrong params.

easybuild/easyblocks/l/llvm.py
Crivella: Checking `_cmakeopts` would guard against behavior changes on `self.final_runtimes` but i really do not think we should ever touch that as it is already fairly clear.

So yes i think the 2nd option is cleaner

easybuild/easyblocks/l/llvm.py
Crivella: The current syntax works both as positional and keywork

```
>>> def a(var=0, *args, **kwargs):
...   print(var)
... 
>>> a(1)
1
>>> a(var=2)
2
```

I do not see an easy way to fix it. The ideal would be to keep the signature `(self, *args, **kwargs)` and than run checks on both args and kwargs. Right now this problem is there also in framework

```
    def _dispatch_sanity_check_step(self, *args, **kwargs):
        """Decide whether to run the dry-run or the real version of the sanity-check step"""
        if self.dry_run:
            self._sanity_check_step_dry_run(*args, **kwargs)
        else:
            self._sanity_check_step(*args, **kwargs)

...
    def _sanity_check_step(self, custom_paths=None, custom_commands=None, extension=False, extra_modules=None):
```

You can still hit the double pass error with this

Crivella: I would probably fix this in a dedicated PR ...  accompanied to a framework one if needed

Flamefire: Ok, just reverted this for now.

Crivella (approved):
LGTM

Crivella:
Going in, thanks @Flamefire!

PR #3856, opened 2025-07-29
remove use of verbose in build_step 

laraPPr:
We give a hard error on our systems when a warning is reported so we are already hitting this deprecation warning for EasyBuidld 6.0.

```
ERROR: Installation of BLAT-3.7-GCC-12.3.0.eb failed: "DEPRECATED (since v6.0) functionality used: The 'verbose' parameter to build_step is deprecated and unneeded.; see https://docs.easybuild.io/deprecated-functionality/ for more information"
```

### DIFF
--- a/easybuild/easyblocks/b/blat.py
+++ b/easybuild/easyblocks/b/blat.py
@@ -42,7 +42,7 @@ class EB_BLAT(MakeCp):
         """Configure build: just create a 'bin' directory."""
         mkdir("bin")
 
-    def build_step(self, verbose=False):
+    def build_step(self):
         """Build BLAT using make and the appropriate options (e.g. BINDIR=)."""
         self.cfg.update('prebuildopts', "MACHTYPE=x86_64")
         bindir = os.path.join(os.getcwd(), "bin")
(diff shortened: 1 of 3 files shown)


easybuild/easyblocks/generic/waf.py
```diff
@@ -52,7 +52,7 @@ def configure_step(self, cmd_prefix=''):
 
         return res.output
 
-    def build_step(self, verbose=False, path=None):
+    def build_step(self, path=None, *args, **kwargs):
```
boegel: since `path` is not used:
```suggestion
    def build_step(self, *args, **kwargs):

File: easybuild/easyblocks/generic/mesonninja.py [hunk 2]
```diff
@@ -130,7 +130,7 @@ def configure_step(self, cmd_prefix=''):
         res = run_shell_cmd(cmd)
         return res.output
 
-    def build_step(self, verbose=False, path=None):
+    def build_step(self, path=None, *args, **kwargs):
```
boegel: since `path` is not used:
```suggestion
    def build_step(self, *args, **kwargs):

PR #3860, opened 2025-07-30
Use patch when filter ld library path

casparvl:
This change introduces extra easyblock-specific configuration items. They allow us to define patches that should be applied conditionally if EasyBuild is configured to filter `LD_LIBRARY_PATH`. Since these patches (see Should be used together with: https://github.com/easybuilders/easybuild-easyconfigs/pull/23499) are quite fundamental (change `ctypes` functions) we prefer them to only be applied in the scenario where they are really needed.

Replaces: https://github.com/easybuilders/easybuild-easyblocks/pull/3798
Should be used together with: https://github.com/easybuilders/easybuild-easyconfigs/pull/23499

### DIFF
--- a/easybuild/easyblocks/p/python.py
+++ b/easybuild/easyblocks/p/python.py
@@ -349,12 +349,13 @@ class EB_Python(ConfigureMake):
                 return ext[1]
         return None
 
-    def patch_step(self, *args, **kwargs):
-        """
-        Custom patch step for Python:
-        * patch setup.py when --sysroot EasyBuild configuration setting is used
-        """
 
+    def fetch_step(self, *args, **kwargs):
+        """
+        Custom fetch step for Python: add patches from patches_filter_ld_library_path to 'patches' if
+        EasyBuild is configured to filter LD_LIBRARY_PATH (and is configured not to filter LIBRARY_PATH).
+        This needs to be done in (or before) the fetch step to ensure that those patches are also fetched.
+        """
         # If EasyBuild is configured to filter LD_LIBRARY_PATH (but not LIBRARY_PATH)
         # add any patches listed in `patches_filter_ld_library_path` to the list of patches
         # Also, add the checksums_filter_ld_library_path to the checksums list in that case.
@@ -382,6 +383,15 @@ class EB_Python(ConfigureMake):
                 msg = f"The length of patches_filter_ld_library_path (%s) "
                 msg += f"is not equal to the length of checksums_filter_ld_library_path(%s)."
                 raise EasyBuildError(msg, len(additional_patches), len(additional_checksums))
+
+        super().fetch_step(*args, **kwargs)
+
+
+    def patch_step(self, *args, **kwargs):
+        """
+        Custom patch step for Python:
+        * patch setup.py when --sysroot EasyBuild configuration setting is used
+        """
 
         super().patch_step(*args, **kwargs)
 
--- a/easybuild/easyblocks/p/python.py
+++ b/easybuild/easyblocks/p/python.py
@@ -365,7 +365,6 @@ class EB_Python(ConfigureMake):
         if 'LD_LIBRARY_PATH' in filtered_env_vars and 'LIBRARY_PATH' not in filtered_env_vars:
             ctypes_util_py = os.path.join("Lib", "ctypes", "util.py")
             orig_gcc_so_name = None
-            
             # Let's do this incrementally since we are going back in time
             # if LooseVersion(self.version) >= "3.9.1":
             #    # From 3.9.1 to at least v3.12.4 there is only one match for this line
--- a/easybuild/easyblocks/p/python.py
+++ b/easybuild/easyblocks/p/python.py
@@ -50,7 +50,7 @@ from easybuild.tools.build_log import EasyBuildError, print_warning
 from easybuild.tools.config import build_option, ERROR, EBPYTHONPREFIXES
 from easybuild.tools.modules import get_software_libdir, get_software_root, get_software_version
 from easybuild.tools.filetools import apply_regex_substitutions, change_dir, mkdir
-from easybuild.tools.filetools import read_file, remove_dir, symlink, write_file, open_file
+from easybuild.tools.filetools import read_file, remove_dir, symlink, write_file
 from easybuild.tools.run import run_shell_cmd
 from easybuild.tools.systemtools import get_shared_lib_ext
 from easybuild.tools.utilities import trace_msg
--- a/easybuild/easyblocks/p/python.py
+++ b/easybuild/easyblocks/p/python.py
@@ -365,11 +365,12 @@ class EB_Python(ConfigureMake):
         if 'LD_LIBRARY_PATH' in filtered_env_vars and 'LIBRARY_PATH' not in filtered_env_vars:
             ctypes_util_py = os.path.join("Lib", "ctypes", "util.py")
             orig_gcc_so_name = None
+            
             # Let's do this incrementally since we are going back in time
-            #if LooseVersion(self.version) >= "3.9.1":
+            # if LooseVersion(self.version) >= "3.9.1":
             #    # From 3.9.1 to at least v3.12.4 there is only one match for this line
             #    orig_gcc_so_name = "_get_soname(_findLib_gcc(name)) or _get_soname(_findLib_ld(name))"
-            #if orig_gcc_so_name:
+            # if orig_gcc_so_name:
             #    orig_gcc_so_name_regex = r'(\s*)' + re.escape(orig_gcc_so_name) + r'(\s*)'
             #    # _get_soname() takes the full path as an argument and uses objdump to get the SONAME field from
             #    # the shared object file. The presence or absence of the SONAME field in the ELF header of a shared
--- a/easybuild/easyblocks/p/python.py
+++ b/easybuild/easyblocks/p/python.py
@@ -302,6 +302,10 @@ class EB_Python(ConfigureMake):
             'ulimit_unlimited': [False, "Ensure stack size limit is set to '%s' during build" % UNLIMITED, CUSTOM],
             'use_lto': [None, "Build with Link Time Optimization (>= v3.7.0, potentially unstable on some toolchains). "
                         "If None: auto-detect based on toolchain compiler (version)", CUSTOM],
+            'patches_filter_ld_library_path': [[], "Patches to apply when EasyBuild is configured to filter "
+                                               "LD_LIBRARY_PATH.", CUSTOM]
+            'checksums_filter_ld_library_path': [[], "Checksums to use for validating the patches listed in "
+                                                 "patches_filter_ld_library_path.", CUSTOM]
         }
         return ConfigureMake.extra_options(extra_vars)
 
@@ -351,6 +355,31 @@ class EB_Python(ConfigureMake):
         * patch setup.py when --sysroot EasyBuild configuration setting is used
         """
 
+        # If EasyBuild is configured to filter LD_LIBRARY_PATH (but not LIBRARY_PATH)
+        # add any patches listed in `patches_filter_ld_library_path` to the list of patches
+        # Also, add the checksums_filter_ld_library_path to the checksums list in that case.
+        # This mechanism e.g. makes sure we can patch ctypes, which normally strongly relies on LD_LIBRARY_PATH to find
+        # libraries. But, we want to do the patching conditional on EasyBuild configuration (i.e. which env vars
+        # are filtered)
+        filtered_env_vars = build_option('filter_env_vars') or []
+        additional_patches = self.cfg['patches_filter_ld_library_path']
+        additional_checksums = self.cfg['checksums_filter_ld_library_path']
+        if ('LD_LIBRARY_PATH' in filtered_env_vars and 'LIBRARY_PATH' not in filtered_env_vars and
+            len(additional_patches) > 0):
+            # Some sanity checking so we can raise an early and clear error if needed
+            if len(additional_patches) == len(additional_checksums):
+                msg = "EasyBuild was configured to filter LD_LIBRARY_PATH (and not to filter LIBRARY_PATH). "
+                msg += "The ctypes module relies heavily on LD_LIBRARY_PATH for locating its libraries. "
+                msg += "The following patches will be applied to make sure ctypes.CDLL, ctypes.cdll.LoadLibrary "
+                msg += f"and ctypes.util.find_library will still work correctly: {additional_patches}."
+                self.log.info(msg)
+                self.cfg['patches'].extend(additional_patches)
+                self.cfg['checksums'].extend(additional_checksums)
+            else:
+                msg = f"The length of patches_filter_ld_library_path (%s) "
+                msg += f"is not equal to the length of checksums_filter_ld_library_path(%s)."
+                raise EasyBuildError(msg, len(additional_patches), len(additional_checksums))
+
         super().patch_step(*args, **kwargs)
 
         if self.install_pip:
@@ -361,10 +390,10 @@ class EB_Python(ConfigureMake):
         # libraries installed with EasyBuild (see https://github.com/EESSI/software-layer/issues/192).
         # ctypes is using GCC (and therefore LIBRARY_PATH) to figure out the full location but then only returns the
         # soname, instead let's return the full path in this particular scenario
-        filtered_env_vars = build_option('filter_env_vars') or []
-        if 'LD_LIBRARY_PATH' in filtered_env_vars and 'LIBRARY_PATH' not in filtered_env_vars:
-            ctypes_util_py = os.path.join("Lib", "ctypes", "util.py")
-            orig_gcc_so_name = None
+        # filtered_env_vars = build_option('filter_env_vars') or []
+        # if 'LD_LIBRARY_PATH' in filtered_env_vars and 'LIBRARY_PATH' not in filtered_env_vars:
+            # ctypes_util_py = os.path.join("Lib", "ctypes", "util.py")
+            # orig_gcc_so_name = None
             # Let's do this incrementally since we are going back in time
             # if LooseVersion(self.version) >= "3.9.1":
             #    # From 3.9.1 to at least v3.12.4 there is only one match for this line
--- a/easybuild/easyblocks/p/python.py
+++ b/easybuild/easyblocks/p/python.py
@@ -50,7 +50,7 @@ from easybuild.tools.build_log import EasyBuildError, print_warning
 from easybuild.tools.config import build_option, ERROR, EBPYTHONPREFIXES
 from easybuild.tools.modules import get_software_libdir, get_software_root, get_software_version
 from easybuild.tools.filetools import apply_regex_substitutions, change_dir, mkdir
-from easybuild.tools.filetools import read_file, remove_dir, symlink, write_file
+from easybuild.tools.filetools import read_file, remove_dir, symlink, write_file, open_file
 from easybuild.tools.run import run_shell_cmd
 from easybuild.tools.systemtools import get_shared_lib_ext
 from easybuild.tools.utilities import trace_msg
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/p/python.py
```diff
@@ -345,6 +352,43 @@ def _get_pip_ext_version(self):
...
+        EasyBuild is configured to filter LD_LIBRARY_PATH (and is configured not to filter LIBRARY_PATH).
+        This needs to be done in (or before) the fetch step to ensure that those patches are also fetched.
+        """
+        # If we filter out LD_LIBRARY_PATH (not unusual when using rpath), ctypes is not able to dynamically load
+        # libraries installed with EasyBuild (see https://github.com/EESSI/software-layer/issues/192).
+        # If EasyBuild is configured to filter LD_LIBRARY_PATHm any patches listed in `patches_filter_ld_library_path`
+        # to the list of patches. Also, add the checksums_filter_ld_library_path to the checksums list in that case.
+        # This mechanism e.g. makes sure we can patch ctypes, which normally strongly relies on LD_LIBRARY_PATH to find
+        # libraries. But, we want to do the patching conditional on EasyBuild configuration (i.e. which env vars
+        # are filtered), hence this setup based on the custom config option 'patches_filter_ld_library_path'
```
ocaisa:
```suggestion
        # If we filter out LD_LIBRARY_PATH (not unusual when using rpath), ctypes is not able to dynamically load
        # libraries installed with EasyBuild (see https://github.com/EESSI/software-layer/issues/192).
        # If EasyBuild is configured to filter LD_LIBRARY_PATH any patches listed in `patches_filter_ld_library_path` are
        # added to the list of patches. Also, we add the checksums_filter_ld_library_path to the checksums list in that case.
        # This mechanism e.g. makes sure we can patch ctypes, which normally strongly relies on LD_LIBRARY_PATH to find
        # libraries. But, we want to do the patching conditionally on EasyBuild configuration (i.e. which env vars
        # are filtered), hence this setup based on the custom config option 'patches_filter_ld_library_path'
```

easybuild/easyblocks/p/python.py
```diff
@@ -302,6 +302,13 @@ def extra_options():
             'ulimit_unlimited': [False, "Ensure stack size limit is set to '%s' during build" % UNLIMITED, CUSTOM],
             'use_lto': [None, "Build with Link Time Optimization (>= v3.7.0, potentially unstable on some toolchains). "
                         "If None: auto-detect based on toolchain compiler (version)", CUSTOM],
+            'patches_filter_ld_library_path': [[], "The ctypes module strongly relies on LD_LIBRARY_PATH to find "
```
boegel: This can be simplified, I would only add 1 custom easyconfig parameter `patch_custom_ctypes` (and add the checksum for that patch file directly to `checksums`, should work fine, no need for custom `checksums_filter_ld_library_path` easyconfig parameter

see also https://github.com/easybuilders/easybuild-easyconfigs/pull/23499#pullrequestreview-3158948975

boegel (changes requested):

casparvl:
EB_ARGS="Python-3.11.5-GCCcore-13.2.0.eb Python-3.12.3-GCCcore-13.3.0.eb Python-3.13.1-GCCcore-14.2.0.eb"

easybuild/easyblocks/p/python.py
```diff
@@ -345,6 +351,45 @@ def _get_pip_ext_version(self):
...
+        # This mechanism e.g. makes sure we can patch ctypes, which normally strongly relies on LD_LIBRARY_PATH to find
+        # libraries. But, we want to do the patching conditionally on EasyBuild configuration (i.e. which env vars
+        # are filtered), hence this setup based on the custom config option 'patches_custom_ctypes'
+        filtered_env_vars = build_option('filter_env_vars') or []
+        additional_patches = []
+        if self.cfg['patch_custom_ctypes'] is not None:
+            additional_patches = [self.cfg['patch_custom_ctypes']]
+        checksums = self.cfg['checksums']
+        sources = self.cfg['sources']
+        if ('LD_LIBRARY_PATH' in filtered_env_vars and len(additional_patches) > 0):
```
boegel: We should also check that `LIBRARY_PATH` is *not* filtered, as we were doing before

boegel:
```suggestion
if ('LD_LIBRARY_PATH' in filtered_env_vars and 'LIBRARY_PATH' not in filtered_env_vars and patch_custom_ctypes):
```

easybuild/easyblocks/p/python.py
```diff
@@ -345,6 +351,45 @@ def _get_pip_ext_version(self):
...
+        # If EasyBuild is configured to filter LD_LIBRARY_PATH any patches listed in `patches_custom_ctypes`
+        # are added to the list of patches. Also, we add the checksums_filter_ld_library_path to the checksums list in
+        # that case.
+        # This mechanism e.g. makes sure we can patch ctypes, which normally strongly relies on LD_LIBRARY_PATH to find
+        # libraries. But, we want to do the patching conditionally on EasyBuild configuration (i.e. which env vars
+        # are filtered), hence this setup based on the custom config option 'patches_custom_ctypes'
+        filtered_env_vars = build_option('filter_env_vars') or []
+        additional_patches = []
+        if self.cfg['patch_custom_ctypes'] is not None:
+            additional_patches = [self.cfg['patch_custom_ctypes']]
```
boegel: No need for the `if`, just allow getting `None` (as long as you don't use `len` below)
```python
       patch_custom_ctypes = self.cfg.get('patch_custom_ctypes')
```

and align the local variable name with what you're grabbing here (also below)

easybuild/easyblocks/p/python.py
```diff
@@ -302,6 +302,12 @@ def extra_options():
             'ulimit_unlimited': [False, "Ensure stack size limit is set to '%s' during build" % UNLIMITED, CUSTOM],
             'use_lto': [None, "Build with Link Time Optimization (>= v3.7.0, potentially unstable on some toolchains). "
                         "If None: auto-detect based on toolchain compiler (version)", CUSTOM],
+            'patch_custom_ctypes': [None, "The ctypes module strongly relies on LD_LIBRARY_PATH to find "
```
boegel: Maybe better to have a more direct name like `patch_ctypes_ld_library_path` ?

boegel (changes requested):

easybuild/easyblocks/p/python.py
```diff
@@ -749,6 +805,18 @@ def sanity_check_step(self):
...
 
+        # If the conditions for applying the patch specified through patch_ctypes_ld_library_path are met, check that
+        # the patch applied correctly (and fixed the issue). Note that the condition should be identical to the one
+        # used to determine if the patch_ctypes_ld_library_path patch should be applied
+        filtered_env_vars = build_option('filter_env_vars') or []
+        patch_ctypes_ld_library_path = self.cfg.get('patch_ctypes_ld_library_path')
+        if (
+            'LD_LIBRARY_PATH' in filtered_env_vars and
+            'LIBRARY_PATH' not in filtered_env_vars and
+            patch_ctypes_ld_library_path
```
casparvl: Should we remove the condition that a patch is applied? In other words: should we allow a user to build with filtered `LD_LIBRARY_PATH` but _without_ the patch? (currently, that is allowed, the user will just receive a warning from the `elif` on line 397

casparvl:
EB_ARGS="Python-3.11.5-GCCcore-13.2.0.eb Python-3.12.3-GCCcore-13.3.0.eb Python-3.13.1-GCCcore-14.2.0.eb"

casparvl:
Strangely enough, I got a failure from the sanity check for 3.10.8:

```
== FAILED: Installation ended unsuccessfully: Opening of libpython3.so using ctypes.CDLL('libpython3.so') failed. Ctypes requires a patch
when EasyBuild is configured to filter LD_LIBRARY_PATH. Please check if you specified a patch through patch_ctypes_ld_library_path and check
the logs to see if it applied correctly. (took 10 mins 25 secs)
```

But inspecting the files that should have been patched, it seems the patch applied correctly I therefore installed with `--skip-sanity-check`, and tried to run the sanity check commands manually: they produce the correct output (different output from what was produced as part of the sanity check!). I don't understand why, maybe some different things in the environment of the sanity check are present that do not match the environment if I try after the install is fully completed? 

Anyway, to help debug, I'll add a `find_library` call to the sanity check, and make sure that gets run first. It's good to have as part of the sanity check anyway. And it'll help us pinpoint where in the process things go wrong (e.g. does find_library return something incorrect?).

easybuild/easyblocks/p/python.py
```diff
@@ -345,6 +352,62 @@ def _get_pip_ext_version(self):
...
 
+    def fetch_step(self, *args, **kwargs):
+        """
+        Custom fetch step for Python: add patches from patches_custom_ctypes to 'patches' if
+        EasyBuild is configured to filter LD_LIBRARY_PATH (and is configured not to filter LIBRARY_PATH).
+        This needs to be done in (or before) the fetch step to ensure that those patches are also fetched.
+        """
+        # If we filter out LD_LIBRARY_PATH (not unusual when using rpath), ctypes is not able to dynamically load
+        # libraries installed with EasyBuild (see https://github.com/EESSI/software-layer/issues/192).
+        # If EasyBuild is configured to filter LD_LIBRARY_PATH any patches listed in `patches_custom_ctypes`
```
boegel:
```suggestion
        # If EasyBuild is configured to filter $LD_LIBRARY_PATH the patch specified in 'patch_ctypes_ld_library_path'

File: easybuild/easyblocks/p/python.py [hunk 8]
```diff
@@ -345,6 +352,62 @@ def _get_pip_ext_version(self):
                 return ext[1]
         return None
 
+    def fetch_step(self, *args, **kwargs):
+        """
+        Custom fetch step for Python: add patches from patches_custom_ctypes to 'patches' if
+        EasyBuild is configured to filter LD_LIBRARY_PATH (and is configured not to filter LIBRARY_PATH).
+        This needs to be done in (or before) the fetch step to ensure that those patches are also fetched.
+        """
+        # If we filter out LD_LIBRARY_PATH (not unusual when using rpath), ctypes is not able to dynamically load
```
boegel:
```suggestion
        # If we filter out $LD_LIBRARY_PATH (not unusual when using rpath), ctypes is not able to dynamically load

File: easybuild/easyblocks/p/python.py [hunk 9]
```diff
@@ -345,6 +352,62 @@ def _get_pip_ext_version(self):
...
+        # libraries installed with EasyBuild (see https://github.com/EESSI/software-layer/issues/192).
+        # If EasyBuild is configured to filter LD_LIBRARY_PATH any patches listed in `patches_custom_ctypes`
+        # are added to the list of patches. Also, we add the checksums_filter_ld_library_path to the checksums list in
+        # that case.
+        # This mechanism e.g. makes sure we can patch ctypes, which normally strongly relies on LD_LIBRARY_PATH to find
```
boegel:
```suggestion
        # This mechanism e.g. makes sure we can patch ctypes, which normally strongly relies on $LD_LIBRARY_PATH to find

File: easybuild/easyblocks/p/python.py [hunk 10]
```diff
@@ -345,6 +352,62 @@ def _get_pip_ext_version(self):
...
+        # libraries. But, we want to do the patching conditionally on EasyBuild configuration (i.e. which env vars
+        # are filtered), hence this setup based on the custom config option 'patches_custom_ctypes'
```
boegel:
```suggestion
        # are filtered), hence this setup based on the custom easyconfig parameter 'patch_ctypes_ld_library_path'

File: easybuild/easyblocks/p/python.py [hunk 11]
```diff
@@ -345,6 +352,62 @@ def _get_pip_ext_version(self):
...
+            # Some sanity checking so we can raise an early and clear error if needed
+            # We expect a (one) checksum for the patch_ctypes_ld_library_path
+            checksums = self.cfg['checksums']
+            sources = self.cfg['sources']
+            patches = self.cfg.get('patches')
+            len_patches = len(patches) if patches else 0
+            if len_patches + len(sources) + 1 == len(checksums):
+                msg = "EasyBuild was configured to filter LD_LIBRARY_PATH (and not to filter LIBRARY_PATH). "
+                msg += "The ctypes module relies heavily on LD_LIBRARY_PATH for locating its libraries. "
+                msg += "The following patches will be applied to make sure ctypes.CDLL, ctypes.cdll.LoadLibrary "
```
boegel:
```suggestion
                msg += "The following patch will be applied to make sure ctypes.CDLL, ctypes.cdll.LoadLibrary "

File: easybuild/easyblocks/p/python.py [hunk 12]
```diff
@@ -345,6 +352,62 @@ def _get_pip_ext_version(self):
...
+                msg += f"and ctypes.util.find_library will still work correctly: {patch_ctypes_ld_library_path}."
+                self.log.info(msg)
+                self.log.info(f"Original list of patches: {self.cfg['patches']}")
+                self.log.info(f"Patch to be added: {patch_ctypes_ld_library_path}")
+                self.cfg.update('patches', [patch_ctypes_ld_library_path])
+                self.log.info(f"Updated list of patches: {self.cfg['patches']}")
+            else:
+                msg = "The length of 'checksums' (%s) was not equal to the total amount of sources (%s) + patches (%s)"
+                msg += ". Did you forget to add a checksum for patch_ctypes_ld_library_path?."
```
boegel:
```suggestion
                msg = "The length of 'checksums' (%s) is not equal to the total amount of sources (%s) + patches (%s). "
                msg += "Did you forget to add a checksum for patch_ctypes_ld_library_path?"

File: easybuild/easyblocks/p/python.py [hunk 13]
```diff
@@ -345,6 +352,62 @@ def _get_pip_ext_version(self):
...
+                raise EasyBuildError(msg, len(checksums), len(sources), len(len_patches + 1))
+        # If LD_LIBRARY_PATH is filtered, but no patch is specified, warn the user that his may not work
+        elif (
+            'LD_LIBRARY_PATH' in filtered_env_vars and
+            'LIBRARY_PATH' not in filtered_env_vars and
+            not patch_ctypes_ld_library_path
+        ):
+            msg = "EasyBuild was configured to filter LD_LIBRARY_PATH (and not to filter LIBRARY_PATH). "
```
boegel:
```suggestion
            msg = "EasyBuild was configured to filter $LD_LIBRARY_PATH (and not to filter $LIBRARY_PATH). "

File: easybuild/easyblocks/p/python.py [hunk 14]
```diff
@@ -345,6 +352,62 @@ def _get_pip_ext_version(self):
...
+            msg += "However, no patch for ctypes was specified through 'patch_ctypes_ld_library_path' in the "
+            msg += "EasyConfig. Note that ctypes.util.find_library, ctypes.CDLL and ctypes.cdll.LoadLibrary heavily "
+            msg += "rely on LD_LIBRARY_PATH. Without the patch, a setup without LD_LIBRARY_PATH will likely not work "
```
boegel:
```suggestion
            msg += "easyconfig. Note that ctypes.util.find_library, ctypes.CDLL and ctypes.cdll.LoadLibrary heavily "
            msg += "rely on $LD_LIBRARY_PATH. Without the patch, a setup without $LD_LIBRARY_PATH will likely not work "

File: easybuild/easyblocks/p/python.py [hunk 15]
```diff
@@ -685,6 +721,45 @@ def install_step(self):
                 symlink(target_lib_dynload, lib_dynload)
                 change_dir(cwd)
 
+    def _sanity_check_ctypes_ld_library_path_patch(self):
+        """Check that the patch for ctypes that should be applied when filtering LD_LIBRARY_PATH works"""
```
boegel:
```suggestion
        """
        Check that ctypes.util.find_library and ctypes.CDLL work as expected.
        When $LD_LIBRARY_PATH is filtered, a patch is required for this to work correctly
        (see patch_ctypes_ld_library_path).
        """

File: easybuild/easyblocks/p/python.py [hunk 16]
```diff
@@ -685,6 +721,45 @@ def install_step(self):
...
+        res = run_shell_cmd(cmd)
+        out = res.output.strip()
+        escaped_python_root = re.escape(self.installdir)
+        pattern = rf"^{escaped_python_root}.*libpython3\.so$"
+        match = re.match(pattern, out)
+        self.log.debug(f"Matching regular expression pattern {pattern} to string {out}")
+        if match:
+            msg = "Call to ctypes.util.find_library('libpython3.so') succesfully found libpython3.so under the prefix "
+            msg += "of the current python installation. indicating that the patch that fixes ctypes when EasyBuild is "
+            msg += "configured to filter LD_LIBRARY_PATH was applied succesfully."
```
boegel:
```suggestion
            msg = "Call to ctypes.util.find_library('libpython3.so') successfully found libpython3.so under "
            msg += f"the installation prefix of the current Python installation ({self.installdir}). "
            if self.cfg.get('patch_ctypes_ld_library_path'):
                msg += "This indicates that the patch that fixes ctypes when EasyBuild is "
                msg += "configured to filter $LD_LIBRARY_PATH was applied succesfully."

File: easybuild/easyblocks/p/python.py [hunk 17]
```diff
@@ -685,6 +721,45 @@ def install_step(self):
...
+            self.log.info(msg)
+        else:
+            msg = "Finding the library libpython3.so using ctypes.util.find_library('libpython3.so') failed. "
+            msg += "Ctypes requires a patch when EasyBuild is configured to filter LD_LIBRARY_PATH. "
```
boegel:
```suggestion
            msg += "The ctypes Python module requires a patch when EasyBuild is configured to filter $LD_LIBRARY_PATH. "

```

easybuild/easyblocks/p/python.py
```diff
@@ -685,6 +721,45 @@ def install_step(self):
...
+        # Now that we know find_library was patched correctly, check if ctypes.CDLL is also patched correctly
+        cmd = "python -c 'import ctypes; print(ctypes.CDLL(\"libpython3.so\"))'"
+        res = run_shell_cmd(cmd)
+        out = res.output.strip()
+        pattern = rf"^<CDLL '{escaped_python_root}.*libpython3\.so', handle [a-f0-9]+ at 0x[a-f0-9]+>$"
+        match = re.match(pattern, out)
+        self.log.debug(f"Matching regular expression pattern {pattern} to string {out}")
+        if match:
+            msg = "Call to ctypes.CDLL('libpython3.so') succesfully opened libpython3.so, indicating that the patch "
+            msg += "that fixes ctypes when EasyBuild is configured to filter LD_LIBRARY_PATH was applied succesfully."
```
boegel:
```suggestion
            msg = "Call to ctypes.CDLL('libpython3.so') succesfully opened libpython3.so. "
            if self.cfg.get('patch_ctypes_ld_library_path'):
                msg += "This indicates that the patch that fixes ctypes when $LD_LIBRARY_PATH is not set "
                msg += "was applied successfully."

File: easybuild/easyblocks/p/python.py [hunk 19]
```diff
@@ -685,6 +721,45 @@ def install_step(self):
...
+            self.log.info(msg)
+        else:
+            msg = "Opening of libpython3.so using ctypes.CDLL('libpython3.so') failed. "
+            msg += "Ctypes requires a patch when EasyBuild is configured to filter LD_LIBRARY_PATH. "
```
boegel:
```suggestion
            msg += "The ctypes Python module requires a patch when EasyBuild is configured to filter $LD_LIBRARY_PATH. "

**Comment by boegel** [comment]
Test report by @boegel
#### Overview of tested easyconfigs (in order)
 * **SUCCESS** _Python-3.12.3-GCCcore-13.3.0.eb_ 
 * **FAIL (build issue)** _Python-3.13.1-GCCcore-14.2.0.eb_ (partial log available at https://gist.github.com/boegel/02fcd3ef8029c1def4ae39e9373d820d)

Build succeeded for 1 out of 2 (2 easyconfigs in total)
node3541.doduo.os - Linux RHEL 9.6 (Plow), x86_64, AMD EPYC 7552 48-Core Processor, Python 3.13.4
See https://gist.github.com/boegel/e7e852989d768b926bccda5c01503e4a for a full test report.

edit: Failure is due to the patch from https://github.com/easybuilders/easybuild-easyconfigs/pull/24189 not being used, I'll look into that (probably outdated easyconfigs installation on my end)

File: easybuild/easyblocks/p/python.py (hunk 6 above)
Comment (boegel) [reply easybuild/easyblocks/p/python.py]: I'm inclined to simply always run this additional sanity check, regardless of whether or not `$LD_LIBRARY_PATH` is being filtered.

But that won't work, since the unpatched `find_library` won't find a hit for `libpython3.so` (only for `python3`).

With that in mind, I remove `patch_ctypes_ld_library_path` from this condition: we want to make sure that if `$LD_LIBRARY_PATH` is filtered, that we still have a working `find_library`...

**Comment by boegel** [comment]
@boegelbot please test @ jsc-zen3
CORE_CNT=16
EB_ARGS="Python-2.7.18-GCCcore-12.3.0.eb Python-3.10.8-GCCcore-12.2.0.eb Python-3.12.3-GCCcore-13.3.0.eb Python-3.13.5-GCCcore-14.3.0.eb --installpath /tmp/$USER/pr3860"

File: easybuild/easyblocks/p/python.py [hunk 20]
```diff
@@ -345,6 +352,64 @@ def _get_pip_ext_version(self):
                 return ext[1]
         return None
 
+    def fetch_step(self, *args, **kwargs):
+        """
+        Custom fetch step for Python.
+
+        Add patch specified in patch_ctypes_ld_library_path to list of patches if
```
boegel, line 359: It's better to do this in the constructor, see:
* https://github.com/easybuilders/easybuild-easyblocks/pull/3965

PR #3862, opened 2025-07-31
Update test step of CP2K easyblock for v2024 + add custom easyconfig parameters to control CP2K test suite parameters

pavelToman:
resolves https://github.com/vscentrum/vsc-software-stack/issues/297

### DIFF
--- a/easybuild/easyblocks/c/cp2k.py
+++ b/easybuild/easyblocks/c/cp2k.py
@@ -732,7 +732,11 @@ class EB_CP2K(EasyBlock):
             if LooseVersion(self.version) > LooseVersion('2024.0'):
                 regtest_script = os.path.join(self.cfg['start_dir'], 'tests', 'do_regtest.py')
                 regtest_cmd = [
-                    'python', regtest_script, '--maxtasks 1 --mpiranks 1 --ompthreads 1', self.typearch, self.cfg['type']
+                    'python',
+                    regtest_script,
+                    '--maxtasks 1 --mpiranks 1 --ompthreads 1',
+                    self.typearch,
+                    self.cfg['type'],
                 ]
             else:
                 regtest_script = os.path.join(self.cfg['start_dir'], 'tools', 'regtesting', 'do_regtest')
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/c/cp2k.py
```diff
@@ -729,8 +729,19 @@ def test_step(self):
...
             cfg_fn = 'cp2k_regtest.cfg'
 
-            regtest_script = os.path.join(self.cfg['start_dir'], 'tools', 'regtesting', 'do_regtest')
-            regtest_cmd = [regtest_script, '-nobuild', '-config', cfg_fn]
+            if LooseVersion(self.version) > LooseVersion('2024.0'):
+                regtest_script = os.path.join(self.cfg['start_dir'], 'tests', 'do_regtest.py')
+                regtest_cmd = [
+                    'python',
+                    regtest_script,
+                    '--maxtasks 1 --mpiranks 1 --ompthreads 1',
```
boegel: Running tests single-core as default makes sense, partially because we know these tests can be quite memory hungry, but it would be nice to provide custom easyconfig parameters so people can tweak these values if they have enough memory in the easyconfig file (or through a hook), like:
```python
tests_maxtasks = 2
tests_mpiranks = 4
tests_ompthreads = 2
```

and same goes for the `--maxerrors` and `--timeout` options.

I would use a very high value for default for `maxerrors`, to always get a full result of the test suite, like 10k (`10000`). And likewise for `timeout` (`1000` makes sense there I think)

boegel: see https://docs.easybuild.io/implementing-easyblocks/#implementing_easyblocks_custom_parameters for docs on supporting custom easyconfig parameters

easybuild/easyblocks/c/cp2k.py
```diff
@@ -729,8 +729,19 @@ def test_step(self):
...
+                    self.typearch,
+                    self.cfg['type'],
+                ]
+            else:
```
boegel, line 749: Stuff below should also go under this else, since it assumes the options are for the old `do_regtest`

easybuild/easyblocks/c/cp2k.py
pavelToman: done!

pavelToman:
@boegelbot please test @ jsc-zen3 EB_ARGS="CP2K-2023.1-foss-2023a.eb"

pavelToman:
This PR is replaced by: 
- https://github.com/easybuilders/easybuild-easyblocks/pull/3888

Issue #3878, opened 2025-08-09
LLVM: Dynamic linker check blocks building old minimal LLVM versions with sysroot


Thyre:
In https://github.com/easybuilders/easybuild-easyblocks/pull/3741, we introduced a hard check for LLVM 19+ to be able to build LLVM with sysroot.

https://github.com/easybuilders/easybuild-easyblocks/blob/77e25387b64f139f55931bfef9eda7e4d8e18fce/easybuild/easyblocks/l/llvm.py#L842-L849

This unfortunately breaks installations of old LLVM versions where we never require:

https://github.com/easybuilders/easybuild-easyblocks/blob/77e25387b64f139f55931bfef9eda7e4d8e18fce/easybuild/easyblocks/l/llvm.py#L1038-L1050

from `_create_compiler_config_file`, as we only build a minimal LLVM without Clang and such.
This will also prevent e.g. `numba` from being built for a new toolchain in EESSI, as `llvmlite` still uses LLVM 15 (in a minimal variant).

We should probably limit the hard check to only apply if we actually build a compiler.

PR #3880, opened 2025-08-09
Allow sysroot for minimal builds for LLVM versions < 19

Thyre:
Resolves https://github.com/easybuilders/easybuild-easyblocks/issues/3878


boegel (approved):
lgtm

PR #3884, opened 2025-08-12
Allow oversubscription for FFTW tests, also for OpenMPI-5.X onwards

casparvl:
`OMPI_MCA_rmaps_base_oversubscribe=true` was converted to `PRTE_MCA_rmaps_default_mapping_policy=:oversubscribe`, as per OpenMPI 5.X. See https://docs.open-mpi.org/en/main/mca.html#converting-mapping-parameters 

I've hit failing test steps for `FFTW.MPI` in an interactive allocation created with `salloc -n 1 -c 16 ...`. This fixes it.

PR #3885, opened 2025-08-13
Allow superlu tests to oversubscribe

casparvl:
Essentially does the same as https://github.com/easybuilders/easybuild-easyblocks/pull/3884 does, but for `SuperLU_DIST`

### DIFF
--- a/easybuild/easyblocks/s/superlu.py
+++ b/easybuild/easyblocks/s/superlu.py
@@ -143,12 +143,12 @@ class EB_SuperLU(CMakeMake):
         # allow oversubscription of number of processes over number of available cores with OpenMPI 3.0 & newer,
         # to avoid that some tests fail if only a handful of cores are available
         ompi_ver = get_software_version('OpenMPI')
-        if LooseVersion(ompi_ver) >= LooseVersion('3.0') and LooseVersion(ompi_ver) < LooseVersion('5.0'):
-            if 'OMPI_MCA_rmaps_base_oversubscribe' not in self.cfg['pretestopts']:
-                self.cfg.update('pretestopts', "export OMPI_MCA_rmaps_base_oversubscribe=true && ")
         if LooseVersion(ompi_ver) >= LooseVersion('5.0'):
             if 'PRTE_MCA_rmaps_default_mapping_policy' not in self.cfg['pretestopts']:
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/s/superlu.py
```diff
@@ -139,6 +139,17 @@ def test_step(self):
...
+
+        # allow oversubscription of number of processes over number of available cores with OpenMPI 3.0 & newer,
+        # to avoid that some tests fail if only a handful of cores are available
+        ompi_ver = get_software_version('OpenMPI')
+        if LooseVersion(ompi_ver) >= LooseVersion('3.0') and LooseVersion(ompi_ver) < LooseVersion('5.0'):
+            if 'OMPI_MCA_rmaps_base_oversubscribe' not in self.cfg['pretestopts']:
+                self.cfg.update('pretestopts', "export OMPI_MCA_rmaps_base_oversubscribe=true && ")
+        if LooseVersion(ompi_ver) >= LooseVersion('5.0'):
+            if 'PRTE_MCA_rmaps_default_mapping_policy' not in self.cfg['pretestopts']:
+                self.cfg.update('pretestopts', "export PRTE_MCA_rmaps_default_mapping_policy=:oversubscribe && ")
```
branfosj:
```suggestion
        if LooseVersion(ompi_ver) >= LooseVersion('5.0'):
            if 'PRTE_MCA_rmaps_default_mapping_policy' not in self.cfg['pretestopts']:
                self.cfg.update('pretestopts', "export PRTE_MCA_rmaps_default_mapping_policy=:oversubscribe && ")
        elif LooseVersion(ompi_ver) >= LooseVersion('3.0'):
            if 'OMPI_MCA_rmaps_base_oversubscribe' not in self.cfg['pretestopts']:
                self.cfg.update('pretestopts', "export OMPI_MCA_rmaps_base_oversubscribe=true && ")
```

An improvement?

casparvl: Yeah, makes sense I guess. I copied this from https://github.com/easybuilders/easybuild-easyblocks/pull/3884 and there the `OMPI_MCA_` thing was first, because that _already existed_. But now you get the most relevant thing first, which is nice.

branfosj (approved):

PR #3888, opened 2025-08-14
Update CP2K easyblock for v2024 and v2025

pavelToman:


pavelToman:
Try alt version to debug v2024 failing tests:
- https://github.com/easybuilders/easybuild-easyblocks/pull/3892

PR #3894, opened 2025-08-20
Update LAMMPS easyblock: v2025 fix + extra tests

### LINKED ISSUE #3785: LAMMPS sanity checks should test accelerator packages (Kokkos, Intel, ...)
Currently, the [sanity check step](https://github.com/easybuilders/easybuild-easyblocks/blob/bacc8b3ffdb3943538ecef1d3cffe802ba93b299/easybuild/easyblocks/l/lammps.py#L537) runs a few example inputs using LAMMPS Python interface.
By default, these just run the standard CPU implementation of the packages.
Many of LAMMPS potentials have [accelerated variants](https://docs.lammps.org/Speed_packages.html) that use packages such as Kokkos, Intel, OpenMP, etc which can greatly speed up the execution.

Many applications leverage Kokkos to run large-scale simulations on GPUs.
Some easyconfigs to build LAMMPS with Kokkos (and CUDA) are available, but not properly tested.
To test this, one would just need to change the command used to launch the simulations, e.g. by specifying to use the `kokkos` suffix of the pair styles, as well as the number of GPUs (and of threads, using OpenMP threading at the same time):
https://docs.lammps.org/Speed_kokkos.html#running-lammps-with-the-kokkos-package

lorisercole:
A few improvements to the LAMMPS Easyblock.

### Sanity checks & tests
- fix to execute sanity-check Python run on LAMMPS v2025+.
  - [x] Requires `mpi4py` to be installed:
    https://github.com/easybuilders/easybuild-easyconfigs/pull/23719
- Add test runs for accelerator packages (KOKKOS, GPU, INTEL, OPENMP)
- Run LAMMPS test runs on multiple cores (<=4) if possible

- Enable LAMMPS unit tests (for version >= `29Aug2024_version2`)
  - [x] Requires some easyconfig patches for `29Aug2024_version2`
    https://github.com/easybuilders/easybuild-easyconfigs/pull/23719

### other fixes
- Enable Kokkos serial backend build: Kokkos serial backend can be built next to the OpenMP/CUDA backend. This avoids performance-related warnings when using Kokkos with a single thread.
- Code cleanup and minor bug fixes

Closes #3785 
Supersedes #3789

### DIFF
--- a/easybuild/easyblocks/g/gnupg_bundle.py
+++ b/easybuild/easyblocks/g/gnupg_bundle.py
@@ -1,0 +1,87 @@
+##
+# Copyright 2021-2025  Ghent University
+#
+# This file is part of EasyBuild,
+# originally created by the HPC team of Ghent University (http://ugent.be/hpc/en),
+# with support of Ghent University (http://ugent.be/hpc),
+# the Flemish Supercomputer Centre (VSC) (https://www.vscentrum.be),
+# Flemish Research Foundation (FWO) (http://www.fwo.be/en)
+# and the Department of Economy, Science and Innovation (EWI) (http://www.ewi-vlaanderen.be/en).
+#
+# https://github.com/easybuilders/easybuild
+#
+# EasyBuild is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation v2.
+#
+# EasyBuild is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with EasyBuild.  If not, see <http://www.gnu.org/licenses/>.
+##
+"""
+Easyconfig for installing a gnupg bundle
+
+The gnupg-bundle is using a UNIX socket path in its tests, specifically for the gpgme component. However, UNIX socket
+paths are limited to 108 characters in length. This easyblock ensures that the build directory is located in a path
+sufficiently small so that the UNIX sockets for the tests can be created.
+
+Source: https://gitlab.archlinux.org/archlinux/archlinux-wsl/-/issues/10
+
+@author: Georgios Kafanas (University of Luxembourg)
+"""
+
+import random
+import time
+from string import ascii_letters
+
+from easybuild.tools.build_log import EasyBuildError, print_msg
+from easybuild.easyblocks.generic.bundle import Bundle
+from easybuild.framework.easyconfig import CUSTOM
+
+
+class EB_gnupg_minus_bundle(Bundle):
+    MAX_UNIX_SOCKET_SAFE_BUILD_PATH_LENGTH = 60
+
+    @staticmethod
+    def extra_options(extra_vars=None):
+        """Easyconfig parameters specific to gnupg-bundle"""
+        extra_vars = Bundle.extra_options(extra_vars=extra_vars)
+        extra_vars.update({
+            'unix_socket_compliant_buildpath': [
+                None,
+                "A build path that ensures that test UNIX socket paths are less than 108 characters long",
+                CUSTOM
+            ]
+        })
+        return extra_vars
+
+    @staticmethod
+    def _get_random_build_path(base_path, prefix):
+        timestamp = time.strftime('%Y%m%d-%H%M%S')
+        salt = ''.join(random.choices(ascii_letters, k=5))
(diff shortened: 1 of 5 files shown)


easybuild/easyblocks/l/lammps.py
```diff
@@ -517,11 +506,27 @@ def configure_step(self, **kwargs):
             if '-DFFT_PACK=' not in self.cfg['configopts']:
                 self.cfg.update('configopts', '-DFFT_PACK=array')
 
-        # https://lammps.sandia.gov/doc/Build_extras.html
-        # KOKKOS
+        # detect the CPU and GPU architecture (used for Intel and Kokkos packages belos)
```
ocaisa:
```suggestion
        # detect the CPU and GPU architecture (used for Intel and Kokkos packages below)
```

easybuild/easyblocks/l/lammps.py
```diff
@@ -676,15 +713,53 @@ def sanity_check_step(self, *args, **kwargs):
...
         ]
 
+        # add accelerator-specific tests
+        if self.pkg_intel:  # INTEL package
+            custom_commands.append(
+                'from lammps import lammps; l=lammps(cmdargs=["-sf", "intel"]); l.file("%s")' %
+                os.path.join(self.installdir, "examples", "msst", "in.msst")
+            )
+        if self.cfg['kokkos']:  # KOKKOS package
+            if self.cuda:  # NOTE: requires a GPU to run
```
ocaisa: This is a difficult requirement to meet, we should be able to support a cross-compilation. In EESSI, for example, this is going to cause problems as we must cross-compile for a lot of architectures. You could make it conditional on the `nvidia-smi` command running successfully?

ocaisa: See for example 
https://github.com/easybuilders/easybuild-easyblocks/blob/6a0f356cda727feaaf79fdb72ae2cfbe48bbdf28/easybuild/easyblocks/t/tensorflow.py#L1048-L1054

ocaisa (dismissed):

easybuild/easyblocks/l/lammps.py
lorisercole: ok yes, good idea

laraPPr:
Ok I'm gonna start testing this one.

lorisercole:
Now it should be ready to work with https://github.com/easybuilders/easybuild-easyconfigs/pull/23719

laraPPr:
It was not failing all of the tests but it was failing those new msst tests. I believe the report should be coming in the next hour.

laraPPr:
Test report by @laraPPr
 * **FAIL (build issue)** _LAMMPS-28Oct2024-foss-2023a-kokkos-mace-CUDA-12.1.1.eb_ (partial log available at https://gist.github.com/laraPPr/ba316b797c362dbdac5e3b989882a10b)

node4014.donphan.os - Linux RHEL 9.6 (Plow), x86_64, Intel(R) Xeon(R) Gold 6240 CPU @ 2.60GHz (cascadelake), 1 x NVIDIA NVIDIA A2, 580.95.05, Python 3.11.3
See https://gist.github.com/laraPPr/433687a154089f9c86bc0575581df111 for a full test report.

laraPPr:
For the CUDA builds the new tests that you added seem to have some problems.

lorisercole:
For versions >= `29Aug2024` you must include the updated easyconfigs: https://github.com/easybuilders/easybuild-easyconfigs/pull/23719
Tests need PyYAML and mpi4py to work. (Perhaps you should run these builds from that PR, with the `--include-easyblocks-from-pr=3894` option.

For older versions, I've just added a condition, so that we won't need to update all these old easyconfigs.

laraPPr:
@boegelbot please test @ jsc-zen3 EB_ARGS="LAMMPS-2Aug2023_update2-foss-2023a-kokkos.eb LAMMPS-23Jun2022-foss-2021a-kokkos.eb LAMMPS-23Jun2022-foss-2021b-kokkos.eb LAMMPS-23Jun2022-foss-2022a-kokkos.eb"

laraPPr:
@lorisercole Can you help me test this easyconfig, `2Aug2023_update2-foss-2023a-kokkos-CUDA-12.1.1.eb`. The sanity check has been stuck on this test for over an hour and I'm testing on two different systems. 
```
 >> running command 'cd /tmp/eb-z8aput5f/tmpiscdmvrr && mpirun -n 4 python -c 'from lammps import lammps; l=lammps();

l.file("/data/gent/vo/000/gvo00002/vsc46128/easybuild/RHEL9/cascadelake-ampere-ib/software/LAMMPS/2Aug2023_update2-fos

s-2023a-kokkos-CUDA-12.1.1/examples/colloid/in.colloid"); l.finalize() if l else None'' ...
```

PR #3896, opened 2025-08-22
Remove osmesa part of the sanity check, since osmesa is removed from mesa 25.1 onwards

casparvl:
Untested for now, but as stated in https://github.com/easybuilders/easybuild-easyconfigs/pull/23144#issuecomment-3213676715 we should adapt the sanity check for Mesa 25.1 onwards, since that's when `osmesa` got removed (see https://docs.mesa3d.org/relnotes/25.1.0.html and https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/33836 ).

Tested to work with Mesa 25.1.3 from https://github.com/easybuilders/easybuild-easyconfigs/pull/23144 . Now also testing with older versions to confirm it didn't break anything.

### DIFF
--- a/easybuild/easyblocks/m/mesa.py
+++ b/easybuild/easyblocks/m/mesa.py
@@ -1,5 +1,5 @@
 ##
-# Copyright 2009-2024 Ghent University
+# Copyright 2009-2025 Ghent University
 #
 # This file is part of EasyBuild,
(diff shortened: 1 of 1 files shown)


Crivella:
EB_ARGS="Mesa-24.1.3-GCCcore-13.3.0.eb Mesa-23.1.9-GCCcore-13.2.0.eb"

PR #3899, opened 2025-08-26
Fix sanity check step of LLVM by loading `binutils` as extra module

### LINKED ISSUE #3891: [Regression]: LLVM sanity check may fail due to system `ld` being used
The sanity check introduced in https://github.com/easybuilders/easybuild-easyblocks/pull/3877 tries to build a test program to see if we're missing any dependencies.
While this works in most cases, having a toolchain too old may cause issues because the system linker is used. Neither GCCcore nor LLVM use `binutils` as a dependency (only a `builddependency`), hence trying to invoke the system linker may fail with:

https://github.com/easybuilders/easybuild-easyconfigs/pull/23603#issuecomment-3199592809

```
g++ minimal.cpp -o minimal_cpp $(llvm-config --link-static --system-libs all) failed with exit code 1 (output: /usr/bin/ld.gold: /tmp/boegelbot/ecpr-23603/software/GCCcore/10.3.0/lib64/libstdc++.so.6: version `GLIBCXX_3.4.29' not found (required by /usr/bin/ld.gold)
```

To fix this, we would either need to move this check to a place where `binutils` is loaded, load `binutils` for this sanity check only, or omit the check entirely...

Flamefire:
When the system ld is used the sanity check step fails
> g++ minimal.cpp -o minimal_cpp $(llvm-config --link-static --system-libs all) failed with exit code 1 (output: /usr/bin/ld.gold: /tmp/boegelbot/ecpr-23603/software/GCCcore/10.3.0/lib64/libstdc++.so.6: version `GLIBCXX_3.4.29' not found (required by /usr/bin/ld.gold)

Ensure binutils gets loaded similar to the GCC sanity check step

Fixes #3891

### DIFF
--- a/easybuild/easyblocks/l/llvm.py
+++ b/easybuild/easyblocks/l/llvm.py
@@ -1714,6 +1714,9 @@ class EB_LLVM(CMakeMake):
         # Here, we add the system libraries LLVM expects to find
         minimal_cpp_compiler_cmd += "$(llvm-config --link-static --system-libs all)"
         custom_commands.append(minimal_cpp_compiler_cmd)
+        # binutils is required for the linking step
+        kwargs.setdefault('extra_modules', []).extend(
+            d['short_mod_name'] for d in self.cfg.dependencies() if d['name'] == 'binutils')
 
         return super().sanity_check_step(custom_paths=custom_paths, custom_commands=custom_commands, *args, **kwargs)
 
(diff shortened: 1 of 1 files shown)


Thyre (approved):
Looks straight-forward and matches what we're doing in GCC.
Tested via https://github.com/easybuilders/easybuild-easyconfigs/pull/23603, which failed for boegelbot previously.

Lets do one more test for safety

Crivella:
Right missed the `g++` now I am wondering what was the reason we are using GCC instead of clang in the sanity check to check for the libs

PR #3902, opened 2025-08-27
Changes to LLVM easyblock required for LLVM 21

Crivella:
## Changes

- Filter unsupported `flang` args from `FFLAGS` and `F90FLAGS` as now flang is used to compile some `.mod` files
  Fixes error of the type `flang-21: error: unknown argument: '-fno-math-errno'` during the build step
- Added EC config parameter `test_suite_include_benchmarks` to allow including/excluding benchmarks in the test-suite
- Ensure `LLVM_INCLUDE_BENCHMARKS` is set both for the projects and runtimes builds

### DIFF
--- a/easybuild/easyblocks/l/llvm.py
+++ b/easybuild/easyblocks/l/llvm.py
@@ -1619,7 +1619,7 @@ class EB_LLVM(CMakeMake):
         shlib_ext = '.' + get_shared_lib_ext()
 
         if not hasattr(self, 'nvptx_target_cond'):
-            # Need to perform the target configuration if not done already eg whe running in `--module-only`
+            # Need to perform the target configuration if not done already e.g. when running in `--module-only`
             # or `--sanity-check-only` modes
             self._configure_build_targets()
 


bartoldeman (dismissed):
I also need this change:
```
--- llvm.py.1   2025-11-11 18:57:04.979489101 +0000
+++ llvm.py     2025-11-11 18:58:56.322052977 +0000
@@ -1103,7 +1103,7 @@
         opts = [f'--gcc-install-dir={self.gcc_prefix}']
 
         if self.dynamic_linker:
-            opts.append(f'-Wl,-dynamic-linker,{self.dynamic_linker}')
+            opts.append(f'-Wl,-dynamic-linker={self.dynamic_linker}')
             # The --dyld-prefix flag exists, but beside being poorly documented it is also not supported by flang
             # https://reviews.llvm.org/D851
             # prefix = self.sysroot.rstrip('/')
```
otherwise `clang-nvlink-wrapper` complains.

easybuild/easyblocks/l/llvm.py
```diff
@@ -594,6 +624,13 @@ def _configure_final_build(self):
             # OMPT based tools
             self._cmakeopts['OPENMP_ENABLE_OMPT_TOOLS'] = ompt_value
 
+        if 'flang' in self.final_projects:
+            self.remove_unsupported_flang_opts()
+
+        include_benchmarks = 'ON' if self.cfg['test_suite_include_benchmarks'] else 'OFF'
+        self._cmakeopts['LLVM_INCLUDE_BENCHMARKS'] = include_benchmarks
+        self.runtimes_cmake_args['LLVM_INCLUDE_BENCHMARKS'] = include_benchmarks
```
Thyre: We need these three here:
`-DLIBCXX_INCLUDE_BENCHMARKS=OFF`
`-DLIBC_INCLUDE_BENCHMARKS=OFF`
`-DLLVM_INCLUDE_BENCHMARKS=OFF`

See https://github.com/easybuilders/easybuild-easyconfigs/pull/25009#issuecomment-3714619644 for a successful test report if all three are set.

easybuild/easyblocks/l/llvm.py
```diff
@@ -125,7 +125,7 @@
 
 GENERAL_OPTS = {
     'CMAKE_VERBOSE_MAKEFILE': 'ON',
-    'LLVM_INCLUDE_BENCHMARKS': 'OFF',
+    # 'LLVM_INCLUDE_BENCHMARKS': 'OFF',
```
Thyre:
```suggestion
```

Lets clean this up entirely, since we have the benchmark option now

easybuild/easyblocks/l/llvm.py
```diff
@@ -231,6 +231,7 @@ def extra_options():
             'skip_sanitizer_tests': [True, "Do not run the sanitizer tests", CUSTOM],
             'test_suite_ignore_patterns': [None, "List of test to ignore (if the string matches)", CUSTOM],
             'test_suite_ignore_timeouts': [False, "Do not treat timedoud tests as failures", CUSTOM],
+            'test_suite_include_benchmarks': [False, "If False do not build the benchmarks tests", CUSTOM],
```
Thyre:
```suggestion
            'test_suite_include_benchmarks': [False, "Include benchmarks in the LLVM tests (default False)", CUSTOM],
```

easybuild/easyblocks/l/llvm.py
```diff
@@ -567,6 +569,34 @@ def _configure_intermediate_build(self):
...
+                unsupported_flang_opts.update([
+                    '-fmath-errno', '-fno-math-errno',
+                    # These are supported since 21.x
+                    # '-fslp-vectorize', '-fvectorize', '-fno-vectorize',
+                    '-fno-unsafe-math-optimizations',
+                ])
+
+        if unsupported_flang_opts:
+            self._remove_iterable_from_envvar('F90FLAGS', unsupported_flang_opts)
+            self._remove_iterable_from_envvar('FFLAGS', unsupported_flang_opts)
```
Thyre, line 595:
```suggestion
    def remove_unsupported_flang_opts(self):
        """For version of LLVM where `flang` is invoked at build time to build modules, ensure that unsupported
        options are removed from F90FLAGS and FFLAGS."""
        unsupported_flang_opts = set()
        if self.version >= '21':
            if self.version < '22':
                unsupported_flang_opts.update([
                    '-fmath-errno', '-fno-math-errno',
                    # These are supported since 21.x
                    # '-fslp-vectorize', '-fvectorize', '-fno-vectorize',
                    '-fno-unsafe-math-optimizations',
                ])

        if unsupported_flang_opts:
            self._remove_iterable_from_envvar('F90FLAGS', unsupported_flang_opts)
            self._remove_iterable_from_envvar('FFLAGS', unsupported_flang_opts)
```

Is this something we should move to Framework, or is this really only necessary here?
I guess the array of unsupported options could be interesting for other EasyBlocks to consume as well...

easybuild/easyblocks/l/llvm.py
```diff
@@ -567,6 +569,34 @@ def _configure_intermediate_build(self):
...
+    def remove_unsupported_flang_opts(self):
+        """For version of LLVM where `flang` is invoched at build time to build modules, ensure that unsupported
+        options are removed from F90FLAGS and FFLAGS."""
+        unsupported_flang_opts = set()
+        if self.version >= '21':
+            if self.version < '22':
+                unsupported_flang_opts.update([
+                    '-fmath-errno', '-fno-math-errno',
+                    # These are supported since 21.x
+                    # '-fslp-vectorize', '-fvectorize', '-fno-vectorize',
```
Thyre:
```suggestion
```

easybuild/easyblocks/l/llvm.py
```diff
@@ -567,6 +569,34 @@ def _configure_intermediate_build(self):
...
+                "Removing unsupported options from %s for flang %s: `%s`", varname, self.version, ', '.join(torm)
+            )
+            setvar(varname, sep.join(flags - torm))
+
+    def remove_unsupported_flang_opts(self):
+        """For version of LLVM where `flang` is invoched at build time to build modules, ensure that unsupported
+        options are removed from F90FLAGS and FFLAGS."""
+        unsupported_flang_opts = set()
+        if self.version >= '21':
+            if self.version < '22':
```
Thyre, line 587: What about LLVM 20?

easybuild/easyblocks/l/llvm.py
```diff
@@ -1582,6 +1621,11 @@ def sanity_check_step(self, custom_paths=None, custom_commands=None, *args, **kw
             lib_dir_runtime = self.get_runtime_lib_path(self.installdir)
         shlib_ext = '.' + get_shared_lib_ext()
 
+        if not hasattr(self, 'nvptx_target_cond'):
+            # Need to perform the target configuration if not done already eg whe running in `--module-only`
```
Thyre:
```suggestion
            # Need to perform the target configuration if not done already e.g. when running in `--module-only`
```

easybuild/easyblocks/l/llvm.py
Crivella: LLVM 20 does not need knowledge of the flags here since it does not compile fortran stuff at build time (atleast not in a way that it uses our flags from EB causing failures)

easybuild/easyblocks/l/llvm.py
Crivella: We do have https://github.com/easybuilders/easybuild-framework/blob/eebe50784546f8168779a665d1297585a66a2607/easybuild/toolchains/compiler/llvm_compilers.py#L57-L68

for the list of flags that could be re-used, but the logic to filter from the FFLAGS is different.

Agree we might benefit from generalizing this more

Thyre: We can follow up on this later, just crossed my mind 😄

PR #3903, opened 2025-08-27
add custom easyconfig parameter for LLVM to ignore timed out tests

Crivella:
See discussion in https://github.com/easybuilders/easybuild-easyconfigs/pull/23144

This would allow to not treat timed out tests as failures by defining `test_suite_ignore_timeouts = True` in the EC file.
The default is kept to False to be consistent with previous behavior


boegel (approved):
lgtm

@Crivella let's get a couple of test reports for this?

boegel (approved):

PR #3904, opened 2025-08-28
Accept arguments for sanity_check_step in HPL easyblock

### LINKED ISSUE #3889: HPCC easyblock is broken after changing to super() call, since it intentioanlly used to skip the HPL checks
In https://github.com/easybuilders/easybuild-easyblocks/commit/85e35d872867afd1a20f0667af4d791afde07cd1 I missed one change by @Flamefire on line
https://github.com/easybuilders/easybuild-easyblocks/blob/85e35d872867afd1a20f0667af4d791afde07cd1/easybuild/easyblocks/h/hpcc.py#L81

which used to say `super(EB_HPL, self)....` i.e. it skipped to the super-super class. 
This is just broken now, since the HPL overload of this method doesn't take any arguments.

Flamefire:
Only accept kwargs though such that custom_paths cannot be passed twice (via positional args)

Fixes #3889

cc @Micket

### DIFF
--- a/easybuild/easyblocks/h/hpl.py
+++ b/easybuild/easyblocks/h/hpl.py
@@ -119,14 +119,15 @@ class EB_HPL(ConfigureMake):
             srcfile = os.path.join(srcdir, filename)
             copy_file(srcfile, destdir)
 
-    def sanity_check_step(self):
+    def sanity_check_step(self, **kwargs):
         """
         Custom sanity check for HPL
         """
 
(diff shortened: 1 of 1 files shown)


boegel (approved):
lgtm

boegel:
EB_ARGS="HPCC-1.5.0-foss-2022a.eb HPL-2.3-foss-2025a.eb"

PR #3917, opened 2025-09-05
Make HPL easyblock intel-aware + allow running the `xhpl` benchmark as a test

Crivella:
Potential fix for 

- https://github.com/easybuilders/easybuild-easyconfigs/issues/19322

I think we could also not run the benchmark as a a test instead of pinning everything to 1 core (if we do not wanna deal with pinning with every different type of MPI)

### DIFF
--- a/easybuild/easyblocks/h/hpl.py
+++ b/easybuild/easyblocks/h/hpl.py
@@ -37,7 +37,6 @@ import re
 import os
 
 import easybuild.tools.toolchain as toolchain
-from easybuild.framework.easyconfig import CUSTOM
 from easybuild.easyblocks.generic.configuremake import ConfigureMake
 from easybuild.tools.build_log import EasyBuildError
 from easybuild.tools.config import build_option


Crivella:
EB_ARGS="--installpath /tmp/$USER/ebpr-3917 HPL-2.3-intel-2021a.eb HPL-2.3-intel-2022.00.eb HPL-2.3-intel-2022.11.eb HPL-2.3-intel-2022a.eb HPL-2.3-intel-2023a.eb HPL-2.3-intel-2024a.eb HPL-2.3-foss-2025a.eb HPL-2.3-foss-2024a.eb HPL-2.3-foss-2023a.eb HPL-2.3-foss-2022a.eb"

easybuild/easyblocks/h/hpl.py
```diff
@@ -104,10 +108,62 @@ def build_step(self, topdir=None):
...
+        comp_fam = self.toolchain.comp_family()
+        if not build_option('mpi_tests') or self.cfg.parallel < req_cpus:
+            self.log.info("MPI tests disabled or not enough cpus... Running tests with 1 oversubscribed process")
+            pin_str = ','.join(['0'] * req_cpus)
+            if comp_fam in [toolchain.INTELCOMP]:
+                pre_cmd = f"I_MPI_PIN_PROCESSOR_LIST=\"{pin_str}\" I_MPI_PIN=on "
+            elif comp_fam in [toolchain.GCC]:
+                post_cmd = f"--cpu-set {pin_str}"
+            else:
+                self.report_test_failure("Don't know how to oversubscribe for %s compiler family" % comp_fam)
```
Thyre: We should do this based on the MPI, not the compiler toolchain, right?
We just recently merged new Intel toolchains using OpenMPI for example.
If I understand correctly, those would incorrectly use `I_MPI_PIN_PROCESSOR_LIST`...

Does MPICH also use `--cpu-set`?

If we could generalize this per MPI instead of compiler toolchain, we would also be a bit more future proof for NVHPC & LLVM.

easybuild/easyblocks/h/hpl.py
```diff
@@ -104,10 +108,62 @@ def build_step(self, topdir=None):
...
+
+        cmd = f"{pre_cmd} mpirun {post_cmd} -np {req_cpus} ./xhpl"
```
Thyre: There might be cases where `mpirun` is not available.
On our JSC systems, we explicitly try to not provide `mpirun`/`mpiexec`, as users try to use that instead of using `srun`.
Maybe add a configuration option for this, defaulting to `mpiexec`/`mpirun`?

Crivella:
Test report by @Crivella
 * **SUCCESS** _HPL-2.3-intel-2021a.eb_ 
 * **SUCCESS** _HPL-2.3-intel-2022.00.eb_ 
 * **SUCCESS** _HPL-2.3-intel-2022.11.eb_ 
 * **SUCCESS** _HPL-2.3-intel-2022a.eb_ 
 * **SUCCESS** _HPL-2.3-intel-2023a.eb_ 
 * **SUCCESS** _HPL-2.3-intel-2024a.eb_ 
 * **SUCCESS** _HPL-2.3-intel-2025b.eb_ 
 * **SUCCESS** _HPL-2.3-foss-2025a.eb_ 
 * **SUCCESS** _HPL-2.3-foss-2024a.eb_ 
 * **SUCCESS** _HPL-2.3-foss-2023a.eb_ 
 * **SUCCESS** _HPL-2.3-foss-2022a.eb_ 

Build succeeded for 11 out of 11 (11 easyconfigs in total)
crivella-desktop - Linux Ubuntu 22.04.5 LTS (Jammy Jellyfish), x86_64, 13th Gen Intel(R) Core(TM) i9-13900K (skylake), Python 3.11.13
See https://gist.github.com/Crivella/a947c3fdf94fdfd1d73f5206012236f7 for a full test report.

easybuild/easyblocks/h/hpl.py
Crivella: Added options taken from the gromacs easyblock

Thyre (approved):
LGTM.
Let me throw together an Intel + OpenMPI build real quick to also test that, but I would expect to see this succeed as well.

I guess you've tested the residual test check before implementing the configure option? It's hard to test this when we have CCNOOPT set 😄

easybuild/easyblocks/h/hpl.py
Crivella: Yeah i had a feeling there should've been something similar to compiler family for mpi but had never used it.

Now of the known ones we are missing `MVAPICH2`  `MPICH2` and `MPITRAMPOLINE` .
I wold assume the first 2 would use the same flag for binding to a specific list of cpus as MPICH, but would need to be tested before adding them

Thyre: I think we can go ahead with the current ones, and expand this if necessary. For now, OpenMPI, Intel & MPICH should be the main MPI implementations...

Thyre:
Hm, we get this warning during the test step:

```
WARNING: Deprecated functionality, will no longer work in EasyBuild v6.0: Easyconfig parameter 'parallel' is deprecated, use 'max_parallel' or the parallel property instead.; see https://docs.easybuild.io/deprecated-functionality/ for more information
```

Should we use `max_parallel` instead?

Crivella:
EB_ARGS="--installpath /tmp/$USER/ebpr-3917 HPL-2.3-intel-2021a.eb HPL-2.3-intel-2022.00.eb HPL-2.3-intel-2022.11.eb HPL-2.3-intel-2022a.eb HPL-2.3-intel-2023a.eb HPL-2.3-intel-2024a.eb HPL-2.3-foss-2025a.eb HPL-2.3-foss-2024a.eb HPL-2.3-foss-2023a.eb HPL-2.3-foss-2022a.eb"

boegel (approved):
lgtm

PR #3943, opened 2025-09-25
enhance LAMMPS easyblock to make it possible to overide `kokkos_arch` for GENERIC ARM target

laraPPr:
Some context for this change: [there is a problem with CUDA<13.1 for Arm architectures which try to leverage NEON instructions](https://github.com/kokkos/kokkos/issues/7483). The workaround in https://github.com/EESSI/software-layer-scripts/pull/87 is to override the architecture detection so these are not included. This change allows you to do that even for the generic case.

### DIFF
--- a/easybuild/easyblocks/l/lammps.py
+++ b/easybuild/easyblocks/l/lammps.py
@@ -276,7 +276,16 @@ class EB_LAMMPS(CMakeMake):
             # For other architectures we set a custom/non-existent type, which will disable all optimizations,
             # and it should use the compiler (optimization) flags set by EasyBuild for this architecture.
             if get_cpu_architecture() == AARCH64:
-                processor_arch = 'ARMV80'
+                if kokkos_arch:
+                    # If someone is trying a manual override for this case, let them
+                    if kokkos_arch not in KOKKOS_CPU_ARCH_LIST:
+                        warning_msg = "Specified CPU ARCH (%s) " % kokkos_arch
+                        warning_msg += "was not found in listed options [%s]." % KOKKOS_CPU_ARCH_LIST
+                        warning_msg += "Still might work though."
+                        print_warning(warning_msg)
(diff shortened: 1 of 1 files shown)


ocaisa (approved):
Verified via https://github.com/EESSI/software-layer-scripts/pull/88#issuecomment-3333282269

easybuild/easyblocks/l/lammps.py
```diff
@@ -276,7 +276,16 @@ def get_kokkos_arch(self, cuda_cc, kokkos_arch):
...
+                if kokkos_arch:
+                    # If someone is trying a manual override for this case, let them
+                    if kokkos_arch not in KOKKOS_CPU_ARCH_LIST:
+                        warning_msg = "Specified CPU ARCH (%s) " % kokkos_arch
+                        warning_msg += "was not found in listed options [%s]." % KOKKOS_CPU_ARCH_LIST
+                        warning_msg += "Still might work though."
+                        print_warning(warning_msg)
+                    processor_arch = kokkos_arch
+                else:
+                    processor_arch = 'ARMV80'
```
Thyre, line 288: I understand the idea, but this looks weird nonetheless.
This will still print a warning because `kokkos_arch` is set, even though it is _not_ ignored but will be used. It's also confusing that this only applies for `aarch64`, even though one might want to use the same approach for x86 when the generic optarch is set.

Instead, we should probably check `kokkos_arch` first and the `optarch` if `kokkos_arch` is not set.
This would allow EESSI to use the overwrite, and increase flexibility in general.

boegel: @ocaisa @laraPPr Should we follow up on this?

@Thyre Maybe it's better to open an issue, following up on things in merged PRs is notoriously difficult

boegel: Ah, already done in:
* #3945

PR #3948, opened 2025-10-02
fix custom easyblock for bzip2 to ensure all versioned symlinks are created

jfgrimm:
got some software explicitly linking `libbz.so.1`, the symlink for which we don't create (so it ended up using the OS bzip2 instead)

(created using `eb --new-pr`)

### DIFF
--- a/easybuild/easyblocks/b/bzip2.py
+++ b/easybuild/easyblocks/b/bzip2.py
@@ -88,7 +88,7 @@ class EB_bzip2(ConfigureMake):
 
             # create (un)versioned symlinks for libbz2.so.X.Y.Z
             split_ver = self.version.split('.')
-            sym_exts = ['.' + '.'.join(split_ver[-3:x]) for x in range(1,3)]  # -> e.g. ['.1', '.1.0'] for version 1.0.8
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/b/bzip2.py
```diff
@@ -86,24 +86,28 @@ def install_step(self):
...
+                        os.chdir(libdir)
+                        os.symlink(f'libbz2.{shlib_ext}.{self.version}', sym)
+                        os.chdir(cwd)
+                    except OSError as err:
+                        raise EasyBuildError("Creating symlink for libbz2.so failed: %s", err)
 
     def sanity_check_step(self):
         """Custom sanity check for bzip2."""
         libs = ['lib/libbz2.a']
         if self.cfg['with_shared_libs']:
```
Thyre, line 105: Should we add a sanity check for the unversioned symlinks as well?

boegel: already done, we're already checking for `lib/libbz2.so`

boegel (approved):
lgtm

PR #3971, opened 2025-10-24
enhance LLVM easyblock to allow control over `$LLVM_PARALLEL_LINK_JOBS`

### LINKED ISSUE #3970: Building LLVM crashes on low-memory machines
Building LLVM crashes on low-memory machines when using a parallel build.  The only diagnostic available from the logs is :

```sh
g++: fatal error: Killed signal terminated program cc1plus.
```

This is roughly ~45 mins into a build on a 16-core machine with 32 GB RAM.

prateekchawla168:
Parallel builds of LLVM crash on low-memory machines. This fix adds an option for `max_link_jobs` (default value = 2) to prevent OOM kills. 

Source: [LLVM Docs](https://llvm.org/docs/CMake.html)

Fixes #3970.


Thyre:
We're not using Ninja, so I'm quite certain that the option will not change anything. 

From the doc for this option:
> These options, when you use the Ninja generator, allow you to restrict the parallelism.

Issue #3969, opened 2025-10-24
Some easyblocks hardwired to leverage `CPATH` but this may not be available under some configurations


ocaisa:
With recent EB releases, a configuration like
```
EASYBUILD_MODULE_SEARCH_PATH_HEADERS=include_paths
EASYBUILD_SEARCH_PATH_CPP_HEADERS=include_paths
```
means there is now a scenario where `CPATH` is never (or at least almost never) set. However there seem to be a few easyblocks that are still hard-coded to use/access `CPATH`:

* https://github.com/easybuilders/easybuild-easyblocks/blob/2a7aad714fef75e88297498dbb7cf040c8202667/easybuild/easyblocks/q/qt.py#L102
* https://github.com/easybuilders/easybuild-easyblocks/blob/2a7aad714fef75e88297498dbb7cf040c8202667/easybuild/easyblocks/r/root.py#L67
* https://github.com/easybuilders/easybuild-easyblocks/blob/2a7aad714fef75e88297498dbb7cf040c8202667/easybuild/easyblocks/x/xmipp.py#L230
* https://github.com/easybuilders/easybuild-easyblocks/blob/2a7aad714fef75e88297498dbb7cf040c8202667/easybuild/easyblocks/r/rosetta.py#L128
* https://github.com/easybuilders/easybuild-easyblocks/blob/2a7aad714fef75e88297498dbb7cf040c8202667/easybuild/easyblocks/c/cuda.py#L111
* https://github.com/easybuilders/easybuild-easyblocks/blob/2a7aad714fef75e88297498dbb7cf040c8202667/easybuild/easyblocks/t/tbb.py#L232
* https://github.com/easybuilders/easybuild-easyblocks/blob/2a7aad714fef75e88297498dbb7cf040c8202667/easybuild/easyblocks/c/clang_aomp.py#L229
* https://github.com/easybuilders/easybuild-easyblocks/blob/2a7aad714fef75e88297498dbb7cf040c8202667/easybuild/easyblocks/n/numpy.py#L464
* https://github.com/easybuilders/easybuild-easyblocks/blob/2a7aad714fef75e88297498dbb7cf040c8202667/easybuild/easyblocks/p/petsc.py#L97
* https://github.com/easybuilders/easybuild-easyblocks/blob/2a7aad714fef75e88297498dbb7cf040c8202667/easybuild/easyblocks/generic/cmakemake.py#L84
[...]

PR #3973, opened 2025-10-24
fix compatibility of ROOT easyblock with `search-path-cpp-headers` option

lexming:
fix for ROOT related to issue https://github.com/easybuilders/easybuild-easyblocks/issues/3969

edit: requires:
* https://github.com/easybuilders/easybuild-framework/pull/5031

### DIFF
--- a/easybuild/easyblocks/r/root.py
+++ b/easybuild/easyblocks/r/root.py
@@ -64,11 +64,15 @@ class EB_ROOT(CMakeMake):
         llvm_root = get_software_root('LLVM')
         if llvm_root:
             llvm_inc = os.path.join(llvm_root, 'include')
-            cpath = os.getenv('CPATH')
-            if cpath:
-                new_cpath = [p for p in cpath.split(os.pathsep) if p != llvm_inc]
-                self.log.info("Filtered %s from $CPATH to avoid using LLVM loaded as indirect dependency", llvm_inc)
-                setvar('CPATH', os.pathsep.join(new_cpath))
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/r/root.py
```diff
@@ -64,11 +64,15 @@ def configure_step(self):
...
-            cpath = os.getenv('CPATH')
-            if cpath:
-                new_cpath = [p for p in cpath.split(os.pathsep) if p != llvm_inc]
-                self.log.info("Filtered %s from $CPATH to avoid using LLVM loaded as indirect dependency", llvm_inc)
-                setvar('CPATH', os.pathsep.join(new_cpath))
+            header_vars = self.toolchain.search_path_vars_headers
+            for header_var in header_vars:
+                header_paths = os.getenv(header_var, "")
+                new_header_paths = [p for p in header_paths.split(os.pathsep) if p != llvm_inc]
+                if new_header_paths:
```
jfgrimm, line 71: I don't think this would ever evaluate to `False`?
If `header_paths` is an empty string, then `new_header_paths` will be `['']`, which is truthy

lexming: there is one case, if all paths in `header_paths` are already in `llvm_inc`, then `new_header_paths = []`

lexming: you are right than an empty string in `header_paths` will pass through though, that needs fixing

PR #3982, opened 2025-10-31
enhance GROMACS easyblock to provide more control over building `gmxapi` Python package

ostueker:
This can be helpful when Python dependency is pulled in by a different dependency but the Python bindings to gmxapi are not desired.

### DIFF
--- a/easybuild/easyblocks/g/gromacs.py
+++ b/easybuild/easyblocks/g/gromacs.py
@@ -336,7 +336,7 @@ class EB_GROMACS(CMakeMake):
                 if gromacs_version >= '2020':
                     # build Python bindings if Python is loaded as a dependency
                     python_root = get_software_root('Python')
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/g/gromacs.py
```diff
@@ -334,7 +336,14 @@ def configure_step(self):
                 if gromacs_version >= '2020':
                     # build Python bindings if Python is loaded as a dependency
                     python_root = get_software_root('Python')
-                    if python_root:
+                    if self.cfg['python_pkg'] is True and not python_root:
```
boegel:
```suggestion
                    if self.cfg['python_pkg'] and not python_root:

boegel (changes requested):

boegel (approved):
lgtm

PR #3892, opened 2025-08-19
Update CP2K easyblock for v2024 and v2025

pavelToman:
resolves https://github.com/vscentrum/vsc-software-stack/issues/297
This update to cp2k easyblock should merge other PRs:
- https://github.com/easybuilders/easybuild-easyblocks/pull/3888
- https://github.com/easybuilders/easybuild-easyblocks/pull/3862
- https://github.com/easybuilders/easybuild-easyblocks/pull/3623
- https://github.com/easybuilders/easybuild-easyblocks/pull/3433
- https://github.com/easybuilders/easybuild-easyblocks/pull/2984

PR #3983, opened 2025-11-04
Update CP2K easyblock for CUDA version

### LINKED ISSUE #297: CP2K
* link to support ticket: [#2024022760001074](https://otrsdict.ugent.be/otrs/index.pl?Action=AgentTicketZoom;TicketID=148802)
* website: https://www.cp2k.org/
* installation docs: https://github.com/cp2k/cp2k/blob/master/INSTALL.md
* toolchain: `?`
* easyblock to use: ``
* required dependencies:
  * [x] SIRIUS
    * https://github.com/easybuilders/easybuild-easyconfigs/pull/20989 
  * [x] Libint
  * [x] libxc
  * [x] libxsmm
  * [x] libvori
  * [x] FFTW
  * [x] PLUMED
* optional dependencies:
  * [x] ...
* notes:
  * ...
* effort: *(TBD)*
* other install methods
  * conda:
  * container image:
  * pre-built binaries (RHEL8 Linux x86_64): 
  * easyconfig outside EasyBuild:

pavelToman:
resolves https://github.com/vscentrum/vsc-software-stack/issues/297
A little bit updated PR: https://github.com/easybuilders/easybuild-easyblocks/pull/3892

For: https://github.com/easybuilders/easybuild-easyconfigs/pull/24454


easybuild/easyblocks/c/cp2k.py
```diff
@@ -214,26 +226,77 @@ def configure_step(self):
...
                 options['DFLAGS'] += ' -D__ELPA3'
 
         # CUDA
         cuda = get_software_root('CUDA')
         if cuda:
-            options['DFLAGS'] += ' -D__ACC -D__DBCSR_ACC'
-            options['LIBS'] += ' -lcudart -lcublas -lcufft -lrt'
-            options['NVCC'] = ' nvcc'
+            if cp2k_version >= LooseVersion('2024'):
+                gpuver = os.getenv('GPUVER')
```
boegel: Rather than requiring that `$GPUVER` is set, you can pass something like `ARCH_NUMBER=80` to the `make` command (via `buildopts`), based on output of `nvidia-smi --query-gpu=compute_cap`

boegel (changes requested):

easybuild/easyblocks/c/cp2k.py
```diff
@@ -214,26 +226,72 @@ def configure_step(self):
...
+            # includes for sirius.mod
+            include_dir = os.path.join(sirius, 'include', 'sirius')
+            if os.path.isdir(include_dir):
+                options['INCS'] += f' -I{include_dir}'
+            else:
+                options['INCS'] += f" -I{os.path.join(sirius, 'include')}"
+
+            # ensure C++ linker flags are OK
+            if not options.get('CXX'):
+                options['CXX'] = os.getenv('MPICXX') or 'mpicxx'
```
akesandgren, line 243: If this is required then it would be more correct to complain if the usempi toolchain opts is not set.
Just changing compiler like this won't really work unless the EC's toolchain is verified to be something that includes MPI in the first place.
I know that is a user error but this is not good practice.

easybuild/easyblocks/c/cp2k.py
```diff
@@ -214,26 +226,72 @@ def configure_step(self):
...
+
+            # library search paths
+            for libdir in (os.path.join(sirius, 'lib'), os.path.join(sirius, 'lib64')):
+                if os.path.isdir(libdir):
+                    options['LDFLAGS'] += f' -L{libdir}'
```
akesandgren, line 248: Do we really need this? Isn't this handled by our framework already?

akesandgren: Compare with the other options, like ELPA or CUDA, they aren't setting LDFLAGS

easybuild/easyblocks/c/cp2k.py
```diff
@@ -214,26 +226,72 @@ def configure_step(self):
...
         cuda = get_software_root('CUDA')
         if cuda:
-            options['DFLAGS'] += ' -D__ACC -D__DBCSR_ACC'
-            options['LIBS'] += ' -lcudart -lcublas -lcufft -lrt'
-            options['NVCC'] = ' nvcc'
+            if cp2k_version >= LooseVersion('2024'):
+                # find cuda compute capabilities and pass it to ARCH_NUMBER
+                # from https://github.com/cp2k/cp2k/blob/v2025.2/exts/build_dbcsr/Makefile#L78
+                get_gpu_cc = run_shell_cmd('nvidia-smi --query-gpu=compute_cap', hidden=True)
+                gpu_arch = get_gpu_cc.output.splitlines()[-1].strip().replace('.', '')
```
akesandgren, line 279: This is not correct, this sets ARCH_NUMBER to whatever compute_cap that the GPU the build is running on has instead of what the cuda-compute-capabilities configuration option specifies.

easybuild/easyblocks/c/cp2k.py
```diff
@@ -214,26 +226,72 @@ def configure_step(self):
...
+
+                options['DFLAGS'] += ' -D__OFFLOAD_CUDA -D__DBCSR_ACC '
+                options['LIBS'] += ' -lcufft -lcudart -lnvrtc -lcuda -lcublas'
+                options['OFFLOAD_CC'] = 'nvcc'
+                options['OFFLOAD_FLAGS'] = "-O3 -g -w --std=c++11 $(DFLAGS) -Xcompiler='-fopenmp -Wall -Wextra' "
```
akesandgren, line 284: This should take the EB settings of the flags into account. Hardcoding is bad practice.

easybuild/easyblocks/c/cp2k.py
```diff
@@ -214,26 +226,72 @@ def configure_step(self):
...
+                options['OFFLOAD_TARGET'] = 'cuda'
+                options['ARCH_NUMBER'] = gpu_arch
+                options['CXX'] = 'mpicxx'
```
akesandgren, line 287: Again, should not hardcode to mpicxx, must use the toolchains setting, i.e. toolchainopts usempi must be true and then CXX will already be correctly set.

easybuild/easyblocks/c/cp2k.py
```diff
@@ -214,26 +226,72 @@ def configure_step(self):
...
+                options['CXXFLAGS'] = '-O3 -fopenmp -g -w --std=c++14 -fPIC $(DFLAGS) $(INCS)'
```
akesandgren, line 288: Same here, don't hardcode options.

easybuild/easyblocks/c/cp2k.py
```diff
@@ -214,26 +226,72 @@ def configure_step(self):
...
-            options['NVCC'] = ' nvcc'
+            if cp2k_version >= LooseVersion('2024'):
+                # find cuda compute capabilities and pass it to ARCH_NUMBER
+                # from https://github.com/cp2k/cp2k/blob/v2025.2/exts/build_dbcsr/Makefile#L78
+                get_gpu_cc = run_shell_cmd('nvidia-smi --query-gpu=compute_cap', hidden=True)
+                gpu_arch = get_gpu_cc.output.splitlines()[-1].strip().replace('.', '')
+
+                options['DFLAGS'] += ' -D__OFFLOAD_CUDA -D__DBCSR_ACC '
+                options['LIBS'] += ' -lcufft -lcudart -lnvrtc -lcuda -lcublas'
+                options['OFFLOAD_CC'] = 'nvcc'
```
akesandgren, line 283: I think there is a framework setting for this too, @boegel ?
I.e. what nvcc is supposed to be instead of hardcoding

easybuild/easyblocks/c/cp2k.py
```diff
@@ -299,10 +359,12 @@ def prepmodinc(self):
     def configure_common(self):
         """Common configuration for all toolchains"""
 
+        cp2k_version = LooseVersion(self.version)
```
akesandgren, line 362: This should be set once as self.cp2k_version  probably in init and then used

easybuild/easyblocks/c/cp2k.py
akesandgren: and the fopenmp should not be hardcoded, it should only be used when toolchainopts openmp is true, see usage of
```
self.toolchain.get_flag('openmp')
```
below (~line 368)

easybuild/easyblocks/c/cp2k.py
```diff
@@ -316,8 +378,13 @@ def configure_common(self):
         # make sure a MPI-2 able MPI lib is used
         mpi2 = False
         if hasattr(self.toolchain, 'MPI_FAMILY') and self.toolchain.MPI_FAMILY is not None:
-            known_mpi2_fams = [toolchain.MPICH, toolchain.MPICH2, toolchain.MVAPICH2, toolchain.OPENMPI,
-                               toolchain.INTELMPI]
+            known_mpi2_fams = [
```
akesandgren, line 381: The check for mpi should probably be done much earlier in the code, then it should also check that openmpi toochainopts is true, and then there would be no reason to change CXX/CC

easybuild/easyblocks/c/cp2k.py
```diff
@@ -330,42 +397,50 @@ def configure_common(self):
...
+            'LDFLAGS': '$(FCFLAGS) %s ' % ldflags,
+            'LIBS': os.getenv('LIBS', ''),
         }
 
+        # Make LIBXSMM visible for all CP2K versions (2023.x and 2024+)
+        if self.libsmm_path:
+            # headers / .mod
+            options['INCS'] += f' -I{self.libsmm_path}/include'
+            # library search path (keep both lib and lib64 in case)
+            options['LDFLAGS'] += f' -L{self.libsmm_path}/lib'
```
akesandgren, line 438: Should consistently use os.path.join, not hardcoded "/"

akesandgren:
@pavelToman Have you had time to look at my comments?

pavelToman:
> @pavelToman Have you had time to look at my comments?

Not yet, sorry. There are many more prior tasks then CP2K. Also it is quite a long time ago I was working on cp2k and have no time to get back to it now.
Still, many thanks for your review! And if you could make a suggestion how to improve it I will commit it to the PR.

akesandgren:
Ok, I'll see if I can pick up the thread... I have user that needs 2025.1 :-)

PR #3984, opened 2025-11-04
enhance GROMACS easyblock to improve control over PLUMED support (native vs. patching)

ostueker:
Since [GROMACS 2025][1] one can compile GROMACS with a PLUMED interface without patching it by setting CMake option [`-DGMX_USE_PLUMED=ON`][2]. 
Applying [PLUMED patches to GROMACS 2025][3] is still possible in order to get the latest PLUMED functionality.

This PR adds a fourth option `'patch'` to the extra-option `plumed`.

* `plumed = None` (default) auto detects whether PLUMED is available and do the same as `plumed = True`
* ~`plumed = True` sets `-DGMX_USE_PLUMED=ON` for GROMACS >= 2025 or try to patch GROMACS with PLUMED's patches for older (<=2024.x) versions (depends whether the loaded version has patches for that version of GROMACS~
* `plumed = 'patch'` will try to patch GROMACS with PLUMED's patches regardless of the GROMACS version
* `plumed = False` will disable PLUMED functionality, even if it is found (neither apply patches nor set `-DGMX_USE_PLUMED=OFF` for GROMACS >=2025

`<edit>`_improved logic is [here](https://github.com/easybuilders/easybuild-easyblocks/pull/3984#issuecomment-3553835074)_`</edit>`

I think that's the best of both worlds. It behaves like before for older GROMACS releases (before 2025) and takes the minimal-invasive route (avoiding to patch) when set to `None` ~or `True`~.

[1]: https://manual.gromacs.org/2025.2/release-notes/2025/major/features.html#a-feature-limited-version-of-the-plumed-interface-is-available
[2]: https://manual.gromacs.org/2025.2/install-guide/index.html#building-with-plumed-support
[3]: https://www.plumed.org/doc-v2.10/user-doc/html/gromacs-2025-0.html

### DIFF
--- a/easybuild/easyblocks/g/gromacs.py
+++ b/easybuild/easyblocks/g/gromacs.py
@@ -72,7 +72,8 @@ class EB_GROMACS(CMakeMake):
             'mpiexec_numproc_flag': ['-np', "Flag to introduce the number of MPI tasks when running tests", CUSTOM],
             'mpi_numprocs': [0, "Number of MPI tasks to use when running tests", CUSTOM],
             'ignore_plumed_version_check': [False, "Ignore the version compatibility check for PLUMED", CUSTOM],
-            'plumed': [None, "Try to apply PLUMED patches. None (default) is auto-detect. " +
+            'plumed': [None, "Try to enable PLUMED support. None (default) is auto-detect. " +
+                       "'patch' apply PLUMED patches even for GROMACS 2025 and newer." +
                        "True or False forces behaviour.", CUSTOM],
         })
         return extra_vars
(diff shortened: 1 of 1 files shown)


jhmeinke:
Thanks for the update! 

I agree, the tricky part is `plumed = None`. I'm worried that somebody who has given `PLUMED` as a dependency  before and now switches to a version of GROMACS >= 2025 will not get the PLUMED functionality they are expecting if the default is to go with the native PLUMED support. 

If I hadn't come across this thread, I would have been wondering why some of our users are complaining about their simulations with GROMACS and PLUMED not working as expected anymore.

I would use patching as the default behavior when `PLUMED` is given as a dependency for all versions of GROMACS that don't have full native support for PLUMED and use `plumed = 'native'` to opt in for now. Once native support is complete, it should of course be the default.

pavelToman:
EB_ARGS="GROMACS-2026.2-foss-2025b.eb GROMACS-2025.2-foss-2025a.eb GROMACS-2024.4-foss-2024a.eb GROMACS-2024.4-foss-2023b-PLUMED-2.9.2.eb"
CORE_CNT=16

pavelToman:
EB_ARGS="GROMACS-2025.4-foss-2025b-CUDA-12.9.1.eb GROMACS-2024.4-foss-2023b-CUDA-12.4.0-PLUMED-2.9.2.eb GROMACS-2023.3-foss-2023a-CUDA-12.1.1-PLUMED-2.9.0.eb"
CORE_CNT=16

easybuild/easyblocks/g/gromacs.py
```diff
@@ -215,43 +217,81 @@ def configure_step(self):
                 self.cfg.update('configopts', "-DGMX_GPU=OFF")
 
         # PLUMED detection
-        # enable PLUMED support if PLUMED is listed as a dependency
-        # and PLUMED support is either explicitly enabled (plumed = True) or unspecified ('plumed' not defined)
+        # enable PLUMED support if PLUMED is listed as a dependency.
+        # plumed = 'native' will enable GROMACS' native PLUMED support ('-DGMX_USE_PLUMED=ON')
+        # for GROMACS 2025 and newer. plumed = 'patch' specifically requests PLUMED patches.
+        # In auto-detect ('plumed = None' or not defined) will prefer native support for 2026
+        # and newer. Older versions of GROMACS patches are applied to enable PLUMED support.
```
ocaisa:
```suggestion
        # and newer. For older versions of GROMACS, patches are applied to enable PLUMED support.
```

ocaisa (changes requested):
LGTM, just a trivial comment change, no need for further rebuilds

ocaisa:
Thanks for this @ostueker, this has also been tested in https://github.com/easybuilders/easybuild-easyconfigs/pull/25962

ocaisa (approved):

PR #3990, opened 2025-11-14
Update GAMESS easyblock for 20240715-R2

sven-hansen:
This PR enhances the GAMESS easyblock to install 20240715-R2 and hopefully upcoming versions as well. The README suggests using the Makefile rather than directly using the old comp*
scripts and lked. Building with the Makefile fixes some issues with compilation order for OpenMP-enabled builds and also significantly reduces compilation time at the cost of one extra patch. It also uses the rungms-dev script rather than the old rungms script which includes Slurm support. Because both rungms and the Makefile need some version-specific patches, the new build procedure is only used for 20240715-R2 onward to maintain backward compatibility with old ECs.


boegel (approved):
lgtm

PR #3995, opened 2025-11-21
Create lockfile for Cargo package if missing

Flamefire:
(created using `eb --new-pr`)

Requires/Includes for easier testing:
- [x] #4016
- [x] #3996
- [x] #3993

PR #3996, opened 2025-11-21
Explictly call `PythonPackage` and `Cargo` configure step in `CargoPythonPackage` easyblock

Flamefire:
That does additional configuration required for Rust crates with https://github.com/easybuilders/easybuild-easyblocks/pull/3995

### DIFF
--- a/RELEASE_NOTES
+++ b/RELEASE_NOTES
@@ -3,7 +3,72 @@ For more detailed information, please see the git log.
 
 These release notes can also be consulted at https://docs.easybuild.io/release-notes .
 
-The latest version of easybuild-easyblocks provides 201 software-specific easyblocks and 45 generic easyblocks.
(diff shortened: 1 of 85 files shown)


boegel:
> It was "just missing". So the question worth a comment would rather by why it _does not_ need to be this way.
> 
> If we don't do this we only call the configure_step of PythonPackage, not of Cargo which isn't what we want, do we?

Currently only `PythonPackage.configure_step` would be called, indeed.

```python
>>> class A():
...     def test(self):
...         print('A.test')
...
>>> class B():
...     def test(self):
...         print('B.test')
...
>>> class C(A,B):
...     pass
...
>>> c=C()
>>> c.test()
A.test
```

But note that `Cargo.configure_step` is *empty* currently, so I'm still not sure what the point is of this PR...

boegel:
@Flamefire You are changing `Cargo.configure_step` in:
* #3995

So then the changes being made here should be included in that PR. In isolation it just doesn't make much sense to me...

PR #3997, opened 2025-11-21
Allow FlexiBLAS to link to MKL with LLVM

bartoldeman:
LLVM flang follows gfortran for complex return values, but its OpenMP library libomp.so is compatible with Intel's.

requires:
- https://github.com/easybuilders/easybuild-framework/pull/4914


boegel (approved):
lgtm

PR #4007, opened 2025-12-04
Be more careful with `PATH`-like environment variables when using `buildenv` + use `pushenv` for other environment variables

### LINKED ISSUE #4004: `buildenv` block uses `setenv` and does not recognise `PATH`-like variables
This means that it blows away important environment variables like `CPLUS_INCLUDE_PATH`

https://github.com/easybuilders/easybuild-easyblocks/blob/7e8ed9ac42758d2a7f4b5ec3bca4561f8e4140a0/easybuild/easyblocks/generic/buildenv.py#L95-L98

ocaisa:
And also always use `pushenv` for everything else so users can recover environment variables they may set independently.

Fixes #4004

### DIFF
--- a/easybuild/easyblocks/generic/buildenv.py
+++ b/easybuild/easyblocks/generic/buildenv.py
@@ -50,12 +50,20 @@ class BuildEnv(Bundle):
     def extra_options():
         """Add extra easyconfig parameters for Boost."""
         extra_vars = {
-            'build_envvars': [True, "Export environment variables related to compilers, compilation flags, "
-                                    "optimisations, math libaries, etc. (overwriting any existing values).", CUSTOM],
+            'build_env_vars': [True, "Export environment variables related to compilers, compilation flags, "
+                                     "optimisations, math libaries, etc. (overwriting any existing values).", CUSTOM],
             'python_executable': ['python3', "Python executable to use for the wrappers (use None to use path to "
                                   "Python executable used by EasyBuild).", CUSTOM],
         }
         return Bundle.extra_options(extra_vars)
+
+    def __init__(self, *args, **kwargs):
+        """
+        Constructor for buildenv easyblock: initialise class variables
+        """
+        super().__init__(*args, **kwargs)
+
+        self.pushenv_env_vars = []
 
     def prepare_step(self, *args, **kwargs):
         """
@@ -97,7 +105,7 @@ class BuildEnv(Bundle):
         txt = super().make_module_extra()
 
         # include environment variables defined for (non-system) toolchain
-        for key, val in self.pushenv_envvars:
+        for key, val in self.pushenv_env_vars:
             txt += self.module_generator.set_environment(key, val)
 
         self.log.debug(f"make_module_extra added this: {txt}")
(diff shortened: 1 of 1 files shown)


ocaisa:
EB_ARGS="buildenv-default-foss-2023a.eb"

easybuild/easyblocks/generic/buildenv.py
```diff
@@ -93,17 +97,46 @@ def make_module_extra(self):
...
+            self.log.warning("")
+        else:
+            # Create a list of PATH-like environment variables that should always be safe to set.
+            # Be overly cautious and include LIBRARY_PATH and LD_LIBRARY_PATH, they are not
+            # strictly header search paths but they affect compile/runtime behaviour so explicitly
+            # include them just in case (does no harm).
+            path_like_vars = {key for key, _ in COMPILER_MAP_CLASS[SearchPaths]}
+            path_like_vars.add('LIBRARY_PATH')
+            path_like_vars.add('LD_LIBRARY_PATH')
+            for key, val in sorted(self.toolchain.vars.items()):
```
ocaisa, line 139: For the easybuild configuration of `EESSI/2025.06` I noted that header libraries are always set in these items (`C_INCLUDE_PATH` etc.). Since the easyblock used `setenv()`, the existing paths were being wiped away. This fixes that specific problem in a general way.

In `EESSI/2023.06` this does not seem to be a problem as `CPATH` was not being set. Regardless, all angles are covered now and it is also more flexible by the use of `pushenv()` in general.

easybuild/easyblocks/generic/buildenv.py
```diff
@@ -50,6 +52,8 @@ def extra_options():
         extra_vars = {
             'python_executable': ['python3', "Python executable to use for the wrappers (use None to use path to "
                                   "Python executable used by EasyBuild).", CUSTOM],
+            'build_envvars': [True, "Export environment variables related to compilers, compilation flags, "
+                                    "optimisations, math libaries, etc. (overwriting any existing values).", CUSTOM],
```
boegel: alphabetical sorting is preferred here, humans like that

```suggestion
            'build_envvars': [True, "Export environment variables related to compilers, compilation flags, "
                                    "optimisations, math libaries, etc. (overwriting any existing values).", CUSTOM],
            'python_executable': ['python3', "Python executable to use for the wrappers (use None to use path to "
                                  "Python executable used by EasyBuild).", CUSTOM],

File: easybuild/easyblocks/generic/buildenv.py [hunk 3]
```diff
@@ -93,17 +97,46 @@ def make_module_extra(self):
         txt = super().make_module_extra()
 
         # include environment variables defined for (non-system) toolchain
-        if not self.toolchain.is_system_toolchain():
-            for key, val in sorted(self.toolchain.vars.items()):
-                txt += self.module_generator.set_environment(key, val)
+        for key, val in self.pushenv_envvars:
```
boegel: I see this comes from `make_module_step`, so it should always be defined by the time `make_module_extra` is called, but I'll sleep better if this is properly initialised in the constructor...

boegel (changes requested):

easybuild/easyblocks/generic/buildenv.py
boegel: fixed in 3a2f8033d5

boegel:
EB_ARGS="buildenv-default-foss-2025a-CUDA-12.8.0.eb buildenv-default-intel-2025b.eb"

boegel (approved):
lgtm

PR #4014, opened 2025-12-09
extend ignore patterns for LLVM tests on aarch64

Crivella:
Ignore known/ignorable tests failure on `aarch64` for LLVM

See for more details
- https://github.com/EESSI/software-layer/pull/1318#issuecomment-3632720888

### DIFF
--- a/easybuild/easyblocks/l/llvm.py
+++ b/easybuild/easyblocks/l/llvm.py
@@ -723,7 +723,7 @@ class EB_LLVM(CMakeMake):
 
         # Some extra tests need to be ignored for aarch64
         if get_cpu_architecture() == AARCH64:
-            if LooseVersion(self.version) < '22':  # Force checking if this are resolved in newer LLVM versions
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/l/llvm.py
```diff
@@ -718,6 +718,41 @@ def _update_test_ignore_patterns(self):
         # Can give different behavior based on system Scrt1.o
         new_ignore_patterns.append('Flang :: Driver/missing-input.f90')
 
+        # We are not explicitly including Google Test
+        new_ignore_patterns.append('failed_to_discover_tests_from_gtest')
+
+        # Some extra tests need to be ignored for aarch64
+        if get_cpu_architecture() == AARCH64:
+            if LooseVersion(self.version) < '22':  # Force checking if this are resolved in newer LLVM versions
```
bedroge:
```suggestion
            if LooseVersion(self.version) < '22':  # Force checking if these are resolved in newer LLVM versions
```

PR #3988, opened 2025-11-13
Implement merging `Cargo.toml` file with workspace file

Flamefire:
Resolve backreferences like `authors.workspace = true`

https://doc.rust-lang.org/cargo/reference/workspaces.html#the-package-table
For ruff-0.14.3 the [salsa](https://github.com/salsa-rs/salsa) crate is using the[ `[workspace.package]` section](https://doc.rust-lang.org/cargo/reference/workspaces.html#the-package-table) where one can define values that can be inherited by members

So the members `Cargo.toml` file contains e.g.
```toml
[package]
name = "bar"
version.workspace = true
authors.workspace = true
description.workspace = true
documentation.workspace = true
[dependencies]
            leptos = { version = "0.6", features = ["csr"] }
```

For the offline builds we need to copy the members to standalone packages which has the side-effect that no "workspace" exists anymore causing `cargo` to fail with

>       error: failed to get `anyhow` as a dependency of package `ruff v0.14.3 (/dev/shm/ruff/0.14.3/GCCcore-14.3.0/ruff-0.14.3/crates/ruff)`
>       Caused by:
>         failed to load source for dependency `anyhow`
>       Caused by:
>         Unable to update registry `crates-io`
>       Caused by:
>         failed to update replaced source registry `crates-io`
>       Caused by:
>         failed to parse manifest at `/dev/shm/ruff/0.14.3/GCCcore-14.3.0/easybuild_vendor/salsa-macros/Cargo.toml`
>       Caused by:
>         error inheriting `edition` from workspace root manifest's `workspace.package.edition`
>       Caused by:
>         failed to find a workspace root

Checking `cargo vendor` output: It replaces `version.workspace = true` by `version = "1.2.3"`

Hence we need to parse both TOML files, "merge" them to replace those references and write the result to the member `Cargo.toml`

For that I enhanced the minimal TOML parser I implemented for the `"members"` key to create a `dict[str, dict[str, str]]` with sections as outer keys, keys as inner keys mapping the 1:1 copy of the value. I.e. I don't parse the value at all besides trying to find the end: TOML values support lists, strings (single & double quotes), multiline strings (triple quotes), bools, numbers

The above example results in Python:
```python
{
            'package': {
                "name": '"bar"',
                "version.workspace": 'true',
                "authors.workspace": 'true',
                "description.workspace": 'true',
                "documentation.workspace": 'true',
            },
            'dependencies': {
                "leptos": '{ version = "0.6", features = ["csr"] }',
            },
}
```

[...]

PR #4016, opened 2025-12-11
enhance `Cargo` easyblock to merge `Cargo.toml` file with workspace file

Flamefire:
Resolve backreferences like `authors.workspace = true`

https://doc.rust-lang.org/cargo/reference/workspaces.html#the-package-table
For ruff-0.14.3 the [salsa](https://github.com/salsa-rs/salsa) crate is using the[ `[workspace.package]` section](https://doc.rust-lang.org/cargo/reference/workspaces.html#the-package-table) where one can define values that can be inherited by members

So the members `Cargo.toml` file contains e.g.
```toml
[package]
name = "bar"
version.workspace = true
authors.workspace = true
description.workspace = true
documentation.workspace = true
[dependencies]
            leptos = { version = "0.6", features = ["csr"] }
```

For the offline builds we need to copy the members to standalone packages which has the side-effect that no "workspace" exists anymore causing `cargo` to fail with

>       error: failed to get `anyhow` as a dependency of package `ruff v0.14.3 (/dev/shm/ruff/0.14.3/GCCcore-14.3.0/ruff-0.14.3/crates/ruff)`
>       Caused by:
>         failed to load source for dependency `anyhow`
>       Caused by:
>         Unable to update registry `crates-io`
>       Caused by:
>         failed to update replaced source registry `crates-io`
>       Caused by:
>         failed to parse manifest at `/dev/shm/ruff/0.14.3/GCCcore-14.3.0/easybuild_vendor/salsa-macros/Cargo.toml`
>       Caused by:
>         error inheriting `edition` from workspace root manifest's `workspace.package.edition`
>       Caused by:
>         failed to find a workspace root

Checking `cargo vendor` output: It replaces `version.workspace = true` by `version = "1.2.3"`

Hence we need to parse both TOML files, "merge" them to replace those references and write the result to the member `Cargo.toml`

This PR supersedes #3988 by using:

-  `tomllib` from https://github.com/easybuilders/easybuild-framework/pull/5063
- [x] `dump_toml` from https://github.com/easybuilders/easybuild-framework/pull/5071

For future reference I kept that PR as a single, squashed commit.

With that it supports all examples from [the docs](https://doc.rust-lang.org/cargo/reference/workspaces.html#the-package-table) including merging dictionaries and sections which isn't feasible without a full parser and serializing to TOML

The tests check the parsed results, not their textual representation as e.g. comments and formatting get lost

### DIFF
--- a/easybuild/easyblocks/l/llvm.py
+++ b/easybuild/easyblocks/l/llvm.py
@@ -718,6 +718,41 @@ class EB_LLVM(CMakeMake):
         # Can give different behavior based on system Scrt1.o
         new_ignore_patterns.append('Flang :: Driver/missing-input.f90')
 
+        # We are not explicitly including Google Test
+        new_ignore_patterns.append('failed_to_discover_tests_from_gtest')
+
+        # Some extra tests need to be ignored for aarch64
+        if get_cpu_architecture() == AARCH64:
+            if LooseVersion(self.version) < '22':  # Force checking if these are resolved in newer LLVM versions
+                # Related to https://github.com/llvm/llvm-project/issues/100096 needs more investigation
+                new_ignore_patterns.append('BOLT :: AArch64/check-init-not-moved.s')
+                # BOLT-ERROR: cannot process binaries with unmarked object in code at address 0x10774 belonging
+                # to section .text in current mode
+                new_ignore_patterns.append('BOLT :: X86/dwarf5-dwarf4-types-backward-forward-cross-reference.test')
+                new_ignore_patterns.append('BOLT :: X86/dwarf5-locexpr-referrence.test')
+                new_ignore_patterns.append('BOLT :: eh-frame-hdr.test')
+                new_ignore_patterns.append('BOLT :: eh-frame-overwrite.test')
+                # Probably need https://github.com/llvm/llvm-project/pull/147589 to be fixed
+                new_ignore_patterns.append('BOLT :: runtime/AArch64/adrrelaxationpass.s')
+                new_ignore_patterns.append('BOLT :: runtime/AArch64/instrumentation-ind-call.c')
+                new_ignore_patterns.append('BOLT :: runtime/exceptions-instrumentation.test')
+                new_ignore_patterns.append('BOLT :: runtime/AArch64/hook-fini.test')
+
+                # Failing to trigger an expected stack overflow with ASAN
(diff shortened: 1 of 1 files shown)


boegel:
* Test failure for `helix-25.07.1-GCCcore-14.3.0.eb`:
   ```
   /usr/bin/bash: line 1: cd: helix-term: No such file or directory
   ```
    Also happens with current `develop`, so that problem is not introduced by the changes in this PR, so following up on that in a separate PR...

* Test failure for `torch-em-0.7.1-foss-2023a.eb`: `LibLZF/3.6-GCCcore-12.3.0` of `imagecodecs/2024.1.1-foss-2023a` was missing, fixing that...

boegel (approved):
lgtm

Thyre:
> Problem with `helix-25.07.1-GCCcore-14.3.0.eb` seems related to the changes made in [easybuilders/easybuild-framework#5057](https://github.com/easybuilders/easybuild-framework/pull/5057), if I roll that back it works fine.
> 
> I suspect that the changes there unmasked another bug, because the `start_dir` ends up being set to the path of a _file_ something now...
> 
> edit: Since the problem also pop up when _not_ using this updated `Cargo` easyblock, it's not introduced by these changes in this PR, so it shouldn't block this from getting merged either...

@boegel, for some reason, the current directory is set to `helix-stdx` during the install process, where it was the correct top level directory before. For installation, `helix` requires to run the install commands from `helix-term`.
A hacky solution would be to replace the `cd helix-term` in the EasyConfig with `cd ../helix-term`, but we should fix this properly...

```console
# Shell for the command: 'cd helix-term &&  cargo install --profile=release --root=/tmp/software/helix/25.07.1-GCCcore-14.3.0 --path=.'
# Use command history, exit to stop
eb-shell> pwd
/data/EasyBuild-develop/build/helix/25.07.1/GCCcore-14.3.0/helix-stdx
eb-shell> ls ../helix-term/
build.rs  Cargo.toml  src  tests
eb-shell> cd ../helix-term/
eb-shell> cargo install --profile=release --root=/tmp/software/helix/25.07.1-GCCcore-14.3.0 --path=.
  Installing helix-term v25.7.1 (/data/EasyBuild-develop/build/helix/25.07.1/GCCcore-14.3.0/helix-term)
[...]
```

PR #4018, opened 2025-12-12
Always setup config for vendored crates in Cargo easyblock

Flamefire:
Previously having `offline = False` skipped creating the checksum files for downloaded crates and adjusting the `$CARGO_HOME/config.toml` file to point to them causing them to be (almost silently) ignored.

Now always set up the config files for vendored crates and log information when `crates != not offline`, which looks suspicious at least.

This resolves the question raised in https://github.com/easybuilders/easybuild-easyconfigs/pull/24643 which fixes an EasyConfig that has (many) crates specified but with `offline = False` causing them to be downloaded, extracted and then ignored.   
As the result they will be downloaded which fails as the new versions available by now are not compatible with the used Rust version anymore

### DIFF
--- a/easybuild/easyblocks/generic/cargo.py
+++ b/easybuild/easyblocks/generic/cargo.py
@@ -291,7 +291,7 @@ class Cargo(ExtensionEasyBlock):
 
     def extract_step(self):
         """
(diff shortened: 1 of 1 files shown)


boegel:
@Flamefire Can you sync this with `develop` now #3995 is merged?

PR #4043, opened 2026-01-15
Set `$Python3_ROOT_DIR` in `PythonPackage` easyblock

JarneRenders:
Set the Python3_ROOT_DIR environment variable right before the configure command is run. In CMake Python3_ROOT_DIR defines the root directory of Python.

This solves, for example, issues with the GROMACS installations, since those pull in the gmxapi extension as a PythonPackage. The issue there occurs when easybuild has been configured with a virtual environment. Since the behaviour (as defined in gmxPythonDiscovery.cmake) is to use a venv Python if no hint for the Python installation was given, the venv Python installation rather than the loaded module Python installation will be chosen. This makes the build fail, because of an incompatibility between the dependencies (installed with the loaded module Python) and gmxapi which is trying to install using the venv Python.

Giving this Python3_ROOT_DIR hint fixes those issues.

### DIFF
--- a/easybuild/easyblocks/o/openfoam.py
+++ b/easybuild/easyblocks/o/openfoam.py
@@ -49,6 +49,7 @@ import easybuild.tools.toolchain as toolchain
 from easybuild.easyblocks.generic.cmakemake import setup_cmake_env
 from easybuild.framework.easyblock import EasyBlock
 from easybuild.framework.easyconfig import CUSTOM
+from easybuild.toolchains.compiler.fujitsu import TC_CONSTANT_FUJITSU
 from easybuild.tools.build_log import EasyBuildError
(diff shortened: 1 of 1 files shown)


lexming (approved):
LGTM

lexming:
Test report by @lexming
 * **SUCCESS** _GROMACS-2025.2-foss-2025a.eb_ 
  

Build succeeded for 1 out of 1 (total: 49 mins 28 secs) (1 easyconfigs in total)
node703.hydra.os - Linux Rocky Linux 9.6 (Blue Onyx), x86_64, AMD EPYC 9535 64-Core Processor, Python 3.9.21
See https://gist.github.com/lexming/73ed737c4c1e36bd966f26fc8a23db44 for a full test report.

lexming:
Merging, thanks @JarneRenders !

PR #4045, opened 2026-01-15
Ensure AmberTools will find the correct python library and include dir

Crivella:
Addresses failure of building `AmberTools` on EESSI

- https://github.com/EESSI/software-layer/pull/1332

AmberTools uses `find_package(PythonLibs)` which has been deprecated (require different hints than the one we currently set for python in CMake).

With this changes we provide the dedicated hints for it to find the correct library and include directory.

### DIFF
--- a/easybuild/easyblocks/a/amber.py
+++ b/easybuild/easyblocks/a/amber.py
@@ -356,7 +356,10 @@ class EB_Amber(CMakeMake):
                 testdir = self.builddir
                 testname_cs = 'test.cuda_serial'
                 testname_cp = 'test.cuda_parallel'
-            pretestcommands = 'source %s/amber.sh && cd %s' % (self.installdir, testdir)
+            pretestcommands = ' && '.join([
+                'export OMP_NUM_THREADS=1',  # avoid having as many threads as cores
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/a/amber.py
```diff
@@ -164,13 +165,25 @@ def configure_step(self):
 
         pythonroot = get_software_root('Python')
         if pythonroot:
+            version = get_software_version('Python')
+            shlib_ext = get_shared_lib_ext()
             self.cfg.update('configopts', '-DDOWNLOAD_MINICONDA=FALSE')
-            self.cfg.update('configopts', '-DPYTHON_EXECUTABLE=%s' % os.path.join(pythonroot, 'bin', 'python'))
```
Crivella, line 168: Removed as this is now set by the CMake `configure_step`

Crivella:
EB_ARGS="--installpath /tmp/$USER/ebpr-4045 AmberTools-23.6-foss-2023a.eb AmberTools-24.10-foss-2024a.eb AmberTools-25.2-foss-2025a.eb"

PR #4049, opened 2026-01-20
enhance `PythonPackage` and `PythonBundle` to verify Python package names and versions with `pip list` with `sanity_check_pip_list` parameter

smoors:
cfr. conf call discussion https://github.com/easybuilders/easybuild/wiki/Conference-call-notes-20260114

- check if the normalized Python package extension name and version in the easyconfig match the normalized `pip list` name and version.
- show close matches in `pip list` if there is no matching name.

implemented for `Python`, `PythonPackage`, `PythonBundle` easyblocks (+ derivatives).

### UPDATE 2026-03-22:
- by default, the extended `pip list` sanity check runs only if `--upload-test-report` is enabled. you can override this by setting easyconfig parameter `sanity_check_pip_list` to `True` or `False`
- the `pip list` check for `0.0.0` versions is still done by default as before this PR, but is now done in `run_pip_list()` instead of `run_pip_check()`.

### DIFF
--- a/easybuild/easyblocks/generic/pythonbundle.py
+++ b/easybuild/easyblocks/generic/pythonbundle.py
@@ -229,7 +229,7 @@ class PythonBundle(Bundle):
             self.log.deprecated(msg, '6.0')
 
         if sanity_pip_check:
-            run_pip_check(python_cmd=self.python_cmd, unversioned_packages=all_unversioned_packages)
+            run_pip_check(python_cmd=self.python_cmd)
             pkgs = [(x.name, x.version) for x in py_exts]
             run_pip_list(pkgs, python_cmd=self.python_cmd, unversioned_packages=all_unversioned_packages)
 
--- a/easybuild/easyblocks/generic/pythonbundle.py
+++ b/easybuild/easyblocks/generic/pythonbundle.py
@@ -29,7 +29,6 @@ EasyBuild support for installing a bundle of Python packages, implemented as a g
 """
 import os
 
-
 from easybuild.easyblocks.generic.bundle import Bundle
 from easybuild.easyblocks.generic.pythonpackage import EXTS_FILTER_PYTHON_PACKAGES, find_python_cmd_from_ec
 from easybuild.easyblocks.generic.pythonpackage import get_pylibdirs, PythonPackage, run_pip_check, run_pip_list
@@ -231,14 +230,15 @@ class PythonBundle(Bundle):
         for param in mismatched_params:
             msg = (f"For bundles of PythonPackage extensions the {param} parameter "
                    "must be set at the top level, outside of exts_list")
-            if param == 'sanity_pip_list':
-                raise EasyBuildError(msg)
-            else:
+            if param == 'sanity_pip_check':
                 self.log.deprecated(msg, '6.0')
+            else:
+                # no deprecation warning needed for sanity_pip_list
+                raise EasyBuildError(msg)
 
         if sanity_pip_check:
             run_pip_check(python_cmd=self.python_cmd, unversioned_packages=all_unversioned_packages)
 
         if sanity_pip_list:
-            eb_pkgs = [(x.name, x.version) for x in py_exts]
-            run_pip_list(eb_pkgs, python_cmd=self.python_cmd)
+            pkgs = [(x.name, x.version) for x in py_exts]
+            run_pip_list(pkgs, python_cmd=self.python_cmd)
--- a/easybuild/easyblocks/generic/pythonbundle.py
+++ b/easybuild/easyblocks/generic/pythonbundle.py
@@ -215,9 +215,10 @@ class PythonBundle(Bundle):
         for ext in py_exts:
             if ext.cfg['sanity_pip_check'] != sanity_pip_check:
                 mismatched_params.add('sanity_pip_check')
-                sanity_pip_check = True  # Either the main set it or any extension enabled it
-
             all_unversioned_packages.update(ext.cfg['unversioned_packages'])
+
+        if 'sanity_pip_check' in mismatched_params:
+            sanity_pip_check = True  # Either the main set it or any extension enabled it
 
         if all_unversioned_packages != unversioned_packages:
             mismatched_params.add('unversioned_packages')
--- a/easybuild/easyblocks/generic/pythonbundle.py
+++ b/easybuild/easyblocks/generic/pythonbundle.py
@@ -29,9 +29,11 @@ EasyBuild support for installing a bundle of Python packages, implemented as a g
 """
 import os
 
+
 from easybuild.easyblocks.generic.bundle import Bundle
-from easybuild.easyblocks.generic.pythonpackage import EXTS_FILTER_PYTHON_PACKAGES, run_pip_check, set_py_env_vars
-from easybuild.easyblocks.generic.pythonpackage import PythonPackage, get_pylibdirs, find_python_cmd_from_ec
+from easybuild.easyblocks.generic.pythonpackage import EXTS_FILTER_PYTHON_PACKAGES, find_python_cmd_from_ec
+from easybuild.easyblocks.generic.pythonpackage import get_pylibdirs, PythonPackage, run_pip_check, run_pip_list
+from easybuild.easyblocks.generic.pythonpackage import set_py_env_vars
 from easybuild.tools.build_log import EasyBuildError
 from easybuild.tools.config import build_option, PYTHONPATH, EBPYTHONPREFIXES
 from easybuild.tools.modules import get_software_root
@@ -201,26 +203,42 @@ class PythonBundle(Bundle):
         super()._sanity_check_step_extensions()
 
         sanity_pip_check = self.cfg['sanity_pip_check']
+        sanity_pip_list = self.cfg['sanity_pip_list']
         unversioned_packages = set(self.cfg['unversioned_packages'])
 
         # The options should be set in the main EC and cannot be different between extensions.
         # For backwards compatibility and to avoid surprises enable the pip-check if it is enabled
         # in the main EC or any extension and build the union of all unversioned_packages.
-        has_sanity_pip_check_mismatch = False
         all_unversioned_packages = unversioned_packages.copy()
-        for ext in self.ext_instances:
-            if isinstance(ext, PythonPackage):
-                if ext.cfg['sanity_pip_check'] != sanity_pip_check:
-                    has_sanity_pip_check_mismatch = True
-                all_unversioned_packages.update(ext.cfg['unversioned_packages'])
+        py_exts = [x for x in self.ext_instances if isinstance(x, PythonPackage)]
 
-        if has_sanity_pip_check_mismatch:
-            self.log.deprecated("For bundles of PythonPackage extensions the sanity_pip_check parameter "
-                                "must be set at the top level, outside of exts_list", '6.0')
-            sanity_pip_check = True  # Either the main set it or any extension enabled it
+        checks = {
+            'sanity_pip_check': sanity_pip_check,
+            'sanity_pip_list': sanity_pip_list,
+        }
+        mismatched_params = set()
+
+        for ext in py_exts:
+            for param, value in checks.items():
+                if ext.cfg[param] != value:
+                    mismatched_params.add(param)
+
+            all_unversioned_packages.update(ext.cfg['unversioned_packages'])
+
         if all_unversioned_packages != unversioned_packages:
-            self.log.deprecated("For bundles of PythonPackage extensions the unversioned_packages parameter "
-                                "must be set at the top level, outside of exts_list", '6.0')
+            mismatched_params.add('unversioned_packages')
+
+        for param in mismatched_params:
+            msg = (f"For bundles of PythonPackage extensions the {param} parameter "
+                   "must be set at the top level, outside of exts_list")
+            if param == 'sanity_pip_list':
+                raise EasyBuildError(msg)
+            else:
+                self.log.deprecated(msg, '6.0')
 
         if sanity_pip_check:
             run_pip_check(python_cmd=self.python_cmd, unversioned_packages=all_unversioned_packages)
+
+        if sanity_pip_list:
+            eb_pkgs = [(x.name, x.version) for x in py_exts]
+            run_pip_list(eb_pkgs, python_cmd=self.python_cmd)
--- a/easybuild/easyblocks/generic/pythonbundle.py
+++ b/easybuild/easyblocks/generic/pythonbundle.py
@@ -202,7 +202,6 @@ class PythonBundle(Bundle):
         super()._sanity_check_step_extensions()
 
         sanity_pip_check = self.cfg['sanity_pip_check']
-        sanity_pip_list = self.cfg['sanity_pip_list']
         unversioned_packages = set(self.cfg['unversioned_packages'])
 
         # The options should be set in the main EC and cannot be different between extensions.
@@ -211,16 +210,12 @@ class PythonBundle(Bundle):
         all_unversioned_packages = unversioned_packages.copy()
         py_exts = [x for x in self.ext_instances if isinstance(x, PythonPackage)]
 
-        checks = {
-            'sanity_pip_check': sanity_pip_check,
-            'sanity_pip_list': sanity_pip_list,
-        }
         mismatched_params = set()
 
         for ext in py_exts:
-            for param, value in checks.items():
-                if ext.cfg[param] != value:
-                    mismatched_params.add(param)
+            if ext.cfg['sanity_pip_check'] != sanity_pip_check:
+                mismatched_params.add('sanity_pip_check')
+                sanity_pip_check = True  # Either the main set it or any extension enabled it
 
             all_unversioned_packages.update(ext.cfg['unversioned_packages'])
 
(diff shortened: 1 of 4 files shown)


easybuild/easyblocks/generic/pythonbundle.py
```diff
@@ -29,9 +29,11 @@
 """
 import os
 
+
```
boegel: nitpicking: no need to introduce extra blank line?

easybuild/easyblocks/generic/pythonbundle.py
```diff
@@ -201,26 +203,42 @@ def _sanity_check_step_extensions(self):
...
         if all_unversioned_packages != unversioned_packages:
-            self.log.deprecated("For bundles of PythonPackage extensions the unversioned_packages parameter "
-                                "must be set at the top level, outside of exts_list", '6.0')
+            mismatched_params.add('unversioned_packages')
+
+        for param in mismatched_params:
+            msg = (f"For bundles of PythonPackage extensions the {param} parameter "
+                   "must be set at the top level, outside of exts_list")
+            if param == 'sanity_pip_list':
```
boegel: Question after quick review: what about the `sanity_pip_check` case?

easybuild/easyblocks/p/python.py
```diff
@@ -175,14 +175,68 @@ def det_installed_python_packages(names_only=True, python_cmd=None):
     return [pkg['name'] for pkg in pkgs] if names_only else pkgs
 
 
+def run_pip_list(eb_pkgs, python_cmd=None):
```
boegel: Nitpicking: why the `eb_`, just call it `pkgs` or `python_pkgs`?

easybuild/easyblocks/p/python.py
```diff
@@ -175,14 +175,68 @@ def det_installed_python_packages(names_only=True, python_cmd=None):
...
+    """
+    Run pip list to verify names and versions of installed Python packages
+
+    :param eb_pkgs: list of (name, version) tuples as specified in the easyconfig
+    """
+
+    log = fancylogger.getLogger('run_pip_list', fname=False)
+
+    if python_cmd is None:
+        python_cmd = 'python'
```
boegel, line 202: No need for this, since we just pass down `python_cmd` to `det_installed_python_packages`, which already does this

easybuild/easyblocks/p/python.py
```diff
@@ -175,14 +175,68 @@ def det_installed_python_packages(names_only=True, python_cmd=None):
...
+    msg = "Check on installed Python package names and versions with 'pip list': "
+
+    try:
+        pkgs = det_installed_python_packages(names_only=False, python_cmd=python_cmd)
+    except EasyBuildError as err:
+        trace_msg(msg + 'FAIL')
+        pip_list_errors.append(f"pip list cmd failed:\n{err}")
+
+    trace_msg(msg + 'OK')
+    log.info("pip list cmd passed successfully")
```
boegel: I would move these two lines, along with `msg = ` into the try, it's a bit confusing (to me) to have this here.

But again, nitpicking ^_^

boegel (changes requested):

easybuild/easyblocks/p/python.py
smoors: `eb_pkgs` is used to distinguish from `pip_pkgs`, but sure i can remove the prefix if you prefer

easybuild/easyblocks/generic/pythonbundle.py
smoors: the `sanity_pip_check` case is handled in the else block below, which emits a deprecation warning. i'll reverse the condition add a comment to make it clearer.

for `sanity_pip_list` we can use a hard error because there is no old behavior to deprecate.

easybuild/easyblocks/p/python.py
smoors: i guess that's a matter of preference, but i will not fight over it :)

smoors: done in https://github.com/easybuilders/easybuild-easyblocks/pull/4049/commits/4820b854b055d0ca9087836b3ce4282452fb00d3

easybuild/easyblocks/p/python.py
smoors: done in https://github.com/easybuilders/easybuild-easyblocks/pull/4049/commits/4820b854b055d0ca9087836b3ce4282452fb00d3

easybuild/easyblocks/p/python.py
smoors: done in https://github.com/easybuilders/easybuild-easyblocks/pull/4049/commits/4820b854b055d0ca9087836b3ce4282452fb00d3

easybuild/easyblocks/generic/pythonbundle.py
smoors: done in https://github.com/easybuilders/easybuild-easyblocks/pull/4049/commits/4820b854b055d0ca9087836b3ce4282452fb00d3

smoors:
new plan after short discussion with @boegel:
- normalize extension names and pip list names, and compare them as part of the `sanity_pip_list` check
- check the connection to pypi.org
- if the connection fails, print a warning and write the easyconfig extension name in the module file
- if the connection succeeds but querying the name on pypi.org fails, error (extension name is wrong)
- if querying the name on pypi.org succeeds, write that name in the module file.

smoors:
EB_ARGS="--sanity-check-only ASE-3.26.0-gfbf-2025b.eb ASE-3.26.0-iimkl-2025b.eb Cartopy-0.25.0-foss-2025b.eb Fiona-1.10.1-foss-2025b.eb MDAnalysis-2.10.0-foss-2025b.eb MDTraj-1.11.0-foss-2025b.eb MoviePy-2.2.1-foss-2025b.eb Optuna-4.6.0-foss-2025b.eb PyHMMER-0.11.3-gompi-2025b.eb PyOpenCL-2025.2.7-gfbf-2025b.eb PyTables-3.10.2-foss-2025b.eb Ray-project-2.47.1-foss-2025b.eb SMC++-1.15.4-gfbf-2025b.eb Scalene-1.5.55-gfbf-2025b.eb SciPy-bundle-2025.07-gfbf-2025b.eb SciPy-bundle-2025.07-iimkl-2025b.eb SciPy-bundle-2025.07-lfbf-2025b.eb Sniffles-2.7.1-gfbf-2025b.eb ViTables-3.1.0-foss-2025b.eb VisPy-0.15.2-gfbf-2025b.eb WebDataset-1.0.2-gfbf-2025b.eb Z3-4.15.4-GCCcore-15.2.0.eb abTEM-1.0.8-foss-2025b.eb astropy-7.1.0-gfbf-2025b.eb astropy-testing-7.1.0-gfbf-2025b.eb bokeh-3.7.3-gfbf-2025b.eb dask-2025.9.1-gfbf-2025b.eb dask-mpi-2025.10.0-foss-2025b.eb ipympl-0.9.8-gfbf-2025b.eb jax-0.8.1-gfbf-2025b.eb jupyterlab-h5web-12.6.1-foss-2025b.eb jupyterview-0.8.0-foss-2025b.eb librosa-0.11.0-foss-2025b.eb lit-18.1.8-GCCcore-15.2.0.eb matplotlib-3.10.5-gfbf-2025b.eb matplotlib-3.10.5-iimkl-2025b.eb meshio-5.3.5-foss-2025b.eb ml_dtypes-0.5.3-gfbf-2025b.eb modin-0.37.1-foss-2025b.eb morphosamplers-0.0.16-gfbf-2025b.eb mpi4py-4.1.0-gompi-2025b.eb mpi4py-4.1.0-iimpi-2025b.eb mpi4py-4.1.0-lmpich-2025b.eb mpi4py-4.1.0-lompi-2025b.eb napari-0.6.6-foss-2025b.eb netcdf4-python-1.7.2-foss-2025b.eb nglview-4.0-foss-2025b.eb numba-0.62.0-foss-2025b.eb phonopy-2.43.2-foss-2025b.eb psutil-7.2.1-GCCcore-15.2.0.eb pyiron-0.5.2-20251013-foss-2025b.eb pyiron-gpl-0.0.5-20251021-foss-2025b-mbocus.eb pymatgen-2025.10.7-foss-2025b.eb s3fs-2025.10.0-foss-2025b.eb scikit-image-0.25.0-foss-2025b.eb scikit-learn-1.7.1-gfbf-2025b.eb scikit-learn-1.7.1-iimkl-2025b.eb spglib-python-2.6.0-gfbf-2025b.eb spglib-python-2.6.0-iimkl-2025b.eb xarray-2025.10.0-gfbf-2025b.eb yt-4.4.1-foss-2025b.eb zarr-3.1.5-gfbf-2025b.eb"

smoors:
unrelated failures:
- mpi4py-4.1.0-lompi-2025b.eb -> module not available (mpi4py-4.1.0-gompi-2025b.eb succeeded)
- netcdf4-python-1.7.2-foss-2025b.eb -> `Sanity check failed: no (non-empty) directory found at 'UNKNOWN' in /project/def-maintainers/boegelbot/rocky9/zen3/software/netcdf4-python/1.7.2-foss-2025b` (`--sanity-check-only` not supported?)
- scikit-learn-1.7.1-iimkl-2025b.eb -> module not available (scikit-learn-1.7.1-gfbf-2025b.eb succeeded)

real issues
- SMC++-1.15.4-gfbf-2025b.eb -> `smc++ (close matches in pip list: ['smcpp'])`
- jax-0.8.1-gfbf-2025b.eb -> `jax 0.8.1 (pip list version: 0.8.1.dev20251214)`
- pyiron-0.5.2-20251013-foss-2025b.eb -> `pyiron 0.5.2-20251013 (pip list version: 0+unknown)`
- pyiron-gpl-0.0.5-20251021-foss-2025b-mbocus.eb -> `pyiron-gpl 0.0.5-20251021 (pip list version: 0.0.5.post20251021)`

easybuild/easyblocks/generic/pythonpackage.py
```diff
@@ -1222,7 +1223,8 @@ def sanity_check_step(self, *args, **kwargs):
                                          self.short_mod_name)
 
             unversioned_packages = self.cfg.get('unversioned_packages', [])
-            run_pip_check(python_cmd=python_cmd, unversioned_packages=unversioned_packages)
+            run_pip_check(python_cmd=python_cmd)
+            run_pip_list([(self.name, self.version)], python_cmd=python_cmd, unversioned_packages=unversioned_packages)
```
boegel: @smoors As discussed, let's make this opt-in via `sanity_check_pip_list`, and auto-enable it when `--upload-test-report` is used, to essentially opt-in to this for new easyconfig PRs going forward...

smoors: done

boegel:
CORE_CNT=16
EB_ARGS="--installpath /tmp/$USER/pr4049 Python-3.14.2-GCCcore-15.2.0.eb Python-bundle-PyPI-2023.06-GCCcore-12.3.0.eb yaff-1.6.0-foss-2023b.eb matplotlib-3.9.2-gfbf-2024a.eb"

boegel:
```
== 2026-04-07 19:01:57,291 build_log.py:233 ERROR EasyBuild encountered an error: 
The following Python packages were likely specified with a wrong version because they have another version in the 'pip list' output:
kiwisolver 1.4.5 (version in 'pip list' output: 1.4.4) (at easybuild/easyblocks/python.py:359 in run_pip_list)
```

This is expected, it's being fixed in:
- https://github.com/easybuilders/easybuild-easyconfigs/pull/25685

boegel:
CORE_CNT=16
EB_ARGS="--installpath /tmp/$USER/pr4049 matplotlib-3.9.2-gfbf-2024a.eb"

boegel (approved):
lgtm

boegel:
Test report by @boegel
 * **SUCCESS** _scikit-learn-1.7.1-iimkl-2025b.eb_ 
  
 * **FAIL** _netcdf4-python-1.7.2-foss-2025b.eb_ **(build issue)**
  (partial log available at https://gist.github.com/boegel/7577ab506751ae9fedf5bc0ed1842c9b)
 * **SUCCESS** _Python-3.11.3-GCCcore-12.3.0.eb_ 
  
 * **SUCCESS** _typing-extensions-4.9.0-GCCcore-12.3.0.eb_ 
  

Build succeeded for 3 out of 4 (total: 23 mins 36 secs) (4 easyconfigs in total)
node4232.shinx.os - Linux RHEL 9.6, x86_64, AMD EPYC 9654 96-Core Processor (zen4), Python 3.9.21
See https://gist.github.com/boegel/f78fef597bbfc4f6883059be018e47c1 for a full test report.

boegel:
@smoors Another actual problem uncovered:
```
The following Python packages were likely specified with a wrong name because they are missing in the 'pip list' output:
netcdf4-python (close matches in 'pip list' output: netcdf4 (at easybuild/easyblocks/python.py:360 in run_pip_list)
```

That shows that this all works as designed, so good!

boegel (approved):
lgtm

Flamefire:
This needs to also check for `modulename` as this causes failures for e.g. PyTorch -> torch, see https://github.com/easybuilders/easybuild-easyconfigs/pull/22357#issuecomment-4227506975

I'd block the release over that as it makes all installations of PyTorch and possibly similar ones fail

smoors:
> This needs to also check for `modulename` as this causes failures for e.g. PyTorch -> torch, see [easybuilders/easybuild-easyconfigs#22357 (comment)](https://github.com/easybuilders/easybuild-easyconfigs/pull/22357#issuecomment-4227506975)
> 
> I'd block the release over that as it makes all installations of PyTorch and possibly similar ones fail

see my comment here https://github.com/easybuilders/easybuild-easyconfigs/pull/22357#issuecomment-4230077929

Thyre:
> This needs to also check for `modulename` as this causes failures for e.g. PyTorch -> torch, see [easybuilders/easybuild-easyconfigs#22357 (comment)](https://github.com/easybuilders/easybuild-easyconfigs/pull/22357#issuecomment-4227506975)
> 
> I'd block the release over that as it makes all installations of PyTorch and possibly similar ones fail

Since this only occurs when uploading a test report and the option to overwrite this on an EasyConfig level, I don't consider this as problematic. However, some people might always use `--upload-test-report`, causing in build failures not seen previously.

Flamefire:
> Since this only occurs when uploading a test report and the option to overwrite this on an EasyConfig level, I don't consider this as problematic. However, some people might always use `--upload-test-report`, causing in build failures not seen previously.

I wasn't aware it only triggers for that, not on all sanity checks. Maybe should at least be a warning there?
Still not a good situation now given we know we have easyconfigs that will fail

> see my comment here https://github.com/easybuilders/easybuild-easyconfigs/pull/22357#issuecomment-4230077929
>
> > this is a perfect example why it's better to use PythonBundle over PythonPackage in general:

With that argument we could completely disallow using `PythonPackage` at the "top level" and I'm not sure this is helpful.
E.g. I've seen multiple easyconfigs put Python packages into `exts_list` which are also used by other easyconfigs, also in `exts_list`. So all our effort trying to avoid version clashes in a toolchain generation is undermined by this as we a) don't check extensions and b) it's harder to notice. And PythonBundle kind of encourages this practice over adding another easyconfig.
Especially when those packages are actually only builddependencies (e.g. used for testing) conflicts could be easily avoided.

I found https://github.com/easybuilders/easybuild-framework/issues/493 and https://github.com/easybuilders/easybuild-docs/issues/177 which touches another related point: We nowhere define what `modulename` actually is.
In my understanding, which motivated my comment above, this is the package name (or module, or whatever it is called for the specific extension type). So when we already use this to check if we need to install the extension under `--skip` why not for this check too?
Or have some special case when `PythonPackage` is used as a non-extension.

Coming back to

> it's better to use PythonBundle over PythonPackage in general:

So when would it be better to use PythonPackage?
I think we need to answer this in general, document this and possibly enforce, or at least strongly encourage, whatever we decide.
I find it a bit hard to understand (and document/motivate/explain) why we need a "bundle" for a single package. Just to pass this check would be a weak argument. Having it turn up in LMod as an "extension" would imply that we should always use PythonBundle even when name and python package name are the same. This causes the oddity that it turns up twice in `module spider` though, see TensorFlow

smoors:
just to clarify: the purpose of this PR is to ensure the correct (i.e. as listed by `pip list`) package names and versions can be found by the user when running `module spider`. sanity check will only fail if any python package name or version is not found in the `pip list` output, __and__ if `--upload-test-report` is used (unless `sanity_check_pip_list = False`). this limits the check to new PRs and open PRs for which a test report is being submitted.

to fix PythonPackage easyconfigs that fail the name check, one can either change the easyconfig `name`, or switch to PytonBundle and add an extension with the correct name. either will work, but currently the convention is to use the official software name rather than the `pip list` name as the easyconfig `name`.

i've made a follow-up PR that adds a warning, so users will see the issue before opening a PR: https://github.com/easybuilders/easybuild-easyblocks/pull/4109 . unfortunately, it was too late to be included in the release.

the `modulename` is the name used for importing the module as `import <modulename>`.  its only purpose is to check that the module can be imported. it doesn’t show up in any Lmod output.

the preference for PythonBundle over PythonPackage is just my personal opinion. i don’t see it as a problem that a name turns up twice in the `module spider` output, because extensions are clearly indicated with `(E)`, and better to see it twice than not at all :)

i share your concern over version clashes due to extension versions escaping the 'single dep version' rule. i think we should add a check for that too. now that i think of it, i this PR could make it much easier to implement such a check.

Flamefire:
Ok, I understand the motivation better now, thanks!

We definitely need some documentation somewhere that PythonBundle must be used when Software name and PYPI name are different and when to use PythonPackage.
I'd also put there that build-/test-only Python packages should go into separate easyconfigs, not exts_list.
Any idea where we can put that? I'd additionally put it into PythonBundle once we decided on the wording.

smoors:
not sure where to put that info. maybe add a section specific for Python packages in `docs/writing-easyconfig-files.md`?

PR #4056, opened 2026-02-02
enhance `PythonPackage` and `PythonBundle` to support updating module footers for click autocompletion for specified binaries

Crivella:
Currently, writing the module footers to automatically enable/disable `click` autocompletion on module load/unload is fairly cumbersome and can add significant chunks of repeated code to many easyconfigs.
See for example:
- https://github.com/easybuilders/easybuild-easyconfigs/pull/20459

This PR adds a new configuration `click_autocomplete_bins` option to `PythonPackage`, to allow passing a list of strings (click script/binaries names) to specify which scripts should have autocompletion enabled.

The module footer required for autocompletion will be added both in the case of isolated `PythonPackage`s or when `PythonPackage` is used through `PythonBundle`

## NOTE

I am not running `BINARY_NAME --help` as a sanity check automatically as, even if it is the default for `click` the normal name can be altered (see https://click.palletsprojects.com/en/stable/documentation/#help-parameter-customization)

The `_UPPERCASECOMMAND_COMPLETE=bash_source COMMAND` instead should always produce a `_UNDERSCORED_LOWERCASE_BINNAME_completion` function for the bash completion (since we manually unset this when unloading the module).
This will also catch eventual changes in the behavior of click in the future

### DIFF
--- a/easybuild/easyblocks/generic/pythonbundle.py
+++ b/easybuild/easyblocks/generic/pythonbundle.py
@@ -232,13 +232,9 @@ class PythonBundle(Bundle):
         footer = super().make_module_footer()
 
         click_autocomplete_bins = []
-        for extension in self.cfg['exts_list']:
-            try:
-                _, _, ext_data = extension
-            except ValueError:
-                self.log.warning("Could not unpack extension data for extension '%s'", str(extension))
-                ext_data = {}
-            click_autocomplete_bins += ext_data.get('click_autocomplete_bins') or []
+        for ext in self.cfg['exts_list']:
+            if isinstance(ext, tuple) and len(ext) == 3 and isinstance(ext[2], dict):
+                click_autocomplete_bins += ext[2].get('click_autocomplete_bins') or []
 
         extra_footer = []
         for click_bin in click_autocomplete_bins:


easybuild/easyblocks/generic/pythonpackage.py
```diff
@@ -74,6 +75,93 @@
...
+                    fish {{
+                        puts stdout {{functions -e _{_click_bin_nomin}_completion}}
+                        puts stdout {{complete -e -c {_click_bin}}}
+                    }}
+                }}
+            }}
+        }}
+    }}
+}} else {{
+    puts stderr "Autocompletion of `aiida-pseudo` cannot be setup automatically for shell: $shell"
```
lorisercole:
```suggestion
    puts stderr "Autocompletion of `{_click_bin}` cannot be setup automatically for shell: $shell"
```

easybuild/easyblocks/generic/pythonbundle.py
```diff
@@ -224,3 +224,25 @@ def _sanity_check_step_extensions(self):
 
         if sanity_pip_check:
             run_pip_check(python_cmd=self.python_cmd, unversioned_packages=all_unversioned_packages)
+
+    def make_module_footer(self):
```
boegel, line 228:
```suggestion
    def make_module_footer(self):
        """
        Extend module footer with statements to set up shell completion for Click-based Python tools.
        """

File: easybuild/easyblocks/generic/pythonpackage.py [hunk 3]
```diff
@@ -394,6 +482,9 @@ def extra_options(extra_vars=None):
                                "Otherwise it will be used as-is. A value of None then skips the build step. "
                                "The template %(python)s will be replace by the currently used Python binary.", CUSTOM],
             'check_ldshared': [None, 'Check Python value of $LDSHARED, correct if needed to "$CC -shared"', CUSTOM],
+            'click_autocomplete_bins': [None, "List of command line tools installed by the package that use "
+                                              "the 'click' package and for which autocompletion scripts should be "
+                                              "generated (default: None)", CUSTOM],
```
boegel: default value is automatically mentioned in `eb -a -e PythonPackage` output, no need to add it to help message

```suggestion
                                              "the 'click' package and for which autocompletion scripts "
                                              "should be generated", CUSTOM],
```

easybuild/easyblocks/generic/pythonpackage.py
```diff
@@ -460,6 +551,8 @@ def __init__(self, *args, **kwargs):
         self.pylibdir = UNKNOWN
         self.all_pylibdirs = [UNKNOWN]
 
+        self.click_autocomplete_bins = self.cfg['click_autocomplete_bins'] or []
```
boegel: safer (may matter for easyblocks that don't properly inherit custom easyconfig parater from parent easyblock):
```suggestion
        self.click_autocomplete_bins = self.cfg.get('click_autocomplete_bins', [])
```

easybuild/easyblocks/generic/pythonpackage.py
```diff
@@ -1155,3 +1257,30 @@ def make_module_extra(self, *args, **kwargs):
...
+            extra_footer.append(click_tcl_autocomplete_script(click_bin))
+        elif isinstance(self.module_generator, ModuleGeneratorLua):
+            self.log.debug("Adding Click autocomplete for '%s' in Lua module", click_bin)
+            extra_footer.append(click_lua_autocomplete_script(click_bin))
+        else:
+            self.log.warning("Not adding Click autocomplete for '%s' in unknown module syntax", click_bin)
+
+        return extra_footer
+
+    def make_module_footer(self):
```
boegel, line 1275:
```suggestion
    def make_module_footer(self):
        """
        Extend module footer with statements to set up shell completion for Click-based Python tools.
        """

File: easybuild/easyblocks/generic/pythonpackage.py [hunk 6]
```diff
@@ -460,6 +551,8 @@ def __init__(self, *args, **kwargs):
         self.pylibdir = UNKNOWN
         self.all_pylibdirs = [UNKNOWN]
 
+        self.click_autocomplete_bins = self.cfg.get('click_autocomplete_bins', [])
```
boegel:
```suggestion
        self.click_autocomplete_bins = self.cfg.get('click_autocomplete_bins') or []

File: easybuild/easyblocks/generic/pythonbundle.py [hunk 7]
```diff
@@ -224,3 +224,28 @@ def _sanity_check_step_extensions(self):
...
+        footer = super().make_module_footer()
+
+        click_autocomplete_bins = []
+        for extension in self.cfg['exts_list']:
+            try:
+                _, _, ext_data = extension
+            except ValueError:
+                self.log.warning("Could not unpack extension data for extension '%s'", str(extension))
+                ext_data = {}
+            click_autocomplete_bins += ext_data.get('click_autocomplete_bins', [])
```
boegel:
```suggestion
            click_autocomplete_bins += ext_data.get('click_autocomplete_bins') or []

PR #4037, opened 2026-01-06
transfer dependencies in Bundle easyblock and fix sanity check for module only

Thyre:
(created using `eb --new-pr`)

**Edit:**

`--sanity-check-only` is still broken, see https://github.com/easybuilders/easybuild-easyblocks/pull/4037#issuecomment-3739427251

**Edit 2:**

Maybe solved by https://github.com/easybuilders/easybuild-easyblocks/pull/4037/changes/5010c6f01aaf225daa823e80b1428fb1440ea576

----

During test builds of ROCm-LLVM, one sanity check was failing on `jsc-zen3` due to libraries not being found when trying to link a test program. Investigating further, I was able to create a reproducer EasyConfig and EasyBlock (see https://github.com/easybuilders/easybuild-easyblocks/issues/4009#issuecomment-3714393243), which mirrored the same issue.

When building a normal EasyConfig, `self.toolchain.dependencies` is set to include all the dependencies of the EasyConfig. For a Bundle however, this value is empty. I'm not sure if this is because we're setting most of the values in `__init__`. Could be.
In Framework, the dependencies are initialized during `prepare_step`. To correctly transfer the dependencies to components, add a `prepare_step` to the Bundle EasyBlock, simply transferring the dependencies as they are.

Additionally, `--module-only` and `--sanity-check-only` builds were not executing the sanity check of Bundle components. This is due to `install-step` being skipped in both cases. To fix this, move the logic for determining components, for which the sanity check should run, to the same `prepare_step`.

Resolves https://github.com/easybuilders/easybuild-easyblocks/issues/4008, Resolves https://github.com/easybuilders/easybuild-easyblocks/issues/4009

PR #4057, opened 2026-02-04
Fix duplicate `self` parameter in Bundle easyblock

Flamefire:
Fixes https://github.com/easybuilders/easybuild-easyblocks/pull/4037

`super().x()` already passes `self` so this was equivalent to `EasyBlock.prepare_step(self, self, *args, *kwargs)`

Also fix the name in `check_checksums` as that was very confusing, especially next to each other:
```
        for comp, _ in self.comp_instances:
        ...
        for _, comp in self.comp_instances:

```

### DIFF
--- a/easybuild/easyblocks/generic/bundle.py
+++ b/easybuild/easyblocks/generic/bundle.py
@@ -269,8 +269,8 @@ class Bundle(EasyBlock):
         """
         checksum_issues = super().check_checksums()
 
-        for comp, _ in self.comp_instances:
-            checksum_issues.extend(self.check_checksums_for(comp, sub="of component %s" % comp['name']))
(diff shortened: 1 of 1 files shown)


Thyre:
CORE_CNT=16
EB_ARGS="buildenv-default-foss-2023b.eb OpenSSL-3.eb Circuitscape-5.12.3-Julia-1.9.2.eb Perl-bundle-CPAN-5.38.2-GCCcore-13.3.0.eb gfbf-2024a.eb crypt4gh-1.7-GCC-12.3.0.eb jiter-0.4.1-GCCcore-12.3.0.eb X11-20240607-GCCcore-13.3.0.eb --installpath=/tmp/$USER/ebpr-4057"

PR #4062, opened 2026-02-11
fix call to parent for static method `src_parameter_names` in `Cargo` easyblock

Flamefire:
Required for https://github.com/easybuilders/easybuild-framework/pull/4661

### DIFF
--- a/easybuild/easyblocks/generic/cargo.py
+++ b/easybuild/easyblocks/generic/cargo.py
@@ -171,7 +171,7 @@ class Cargo(ExtensionEasyBlock):
 
     @staticmethod
     def src_parameter_names():
-        return super(Cargo, Cargo).src_parameter_names() + ['crates']
+        return ExtensionEasyBlock.src_parameter_names() + ['crates']
 
     @staticmethod
     def crate_src_filename(pkg_name, pkg_version, _url=None, rev=None):


easybuild/easyblocks/generic/cargo.py
```diff
@@ -171,7 +171,7 @@ def extra_options(extra_vars=None):
 
     @staticmethod
     def src_parameter_names():
-        return super().src_parameter_names() + ['crates']
+        return super(Cargo, Cargo).src_parameter_names() + ['crates']
```
boegel: Should this be:
```suggestion
        return ExtensionEasyBlock.src_parameter_names() + ['crates']
```

boegel: Both seem to work, but the proposed one makes more sense to me (maybe I'm just old).
The double `Cargo` also hurts my eyes a bit.

That's also what we do in other easyblocks, in `extra_options` static methods mainly, so consistency-wise it's also better to follow that pattern...

boegel:
Test report by @boegel
 * **SUCCESS** _skani-0.2.2-GCCcore-12.3.0.eb_ 
  

Build succeeded for 1 out of 1 (total: 1 min 53 secs) (1 easyconfigs in total)
node3902.accelgor.os - Linux RHEL 9.6, x86_64, AMD EPYC 7413 24-Core Processor (zen3), 1 x NVIDIA NVIDIA A100-SXM4-80GB, 590.48.01, Python 3.9.21
See https://gist.github.com/boegel/1ebfade2a05f6bfea733164affcc54fe for a full test report.

PR #4066, opened 2026-02-16
add custom easyblock for ESPResSo

jngrad:
Provide an easyblock to manage CMake flags for each minor release of ESPResSo and automatically download dependencies included via the FetchContent mechanism.

### DIFF
--- a/easybuild/easyblocks/e/espresso.py
+++ b/easybuild/easyblocks/e/espresso.py
@@ -1,0 +1,195 @@
+##
+# Copyright 2025-2026 Jean-Noël Grad
+#
+# EasyBuild is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation v2.
+#
+# EasyBuild is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with EasyBuild.  If not, see <http://www.gnu.org/licenses/>.
+##
+"""
+EasyBuild support for ESPResSo, implemented as an easyblock.
+
+@author: Jean-Noël Grad (University of Stuttgart)
+"""
+
+import os
+import re
+
+from easybuild.easyblocks.generic.cmakeninja import CMakeNinja
+from easybuild.tools.systemtools import get_cpu_architecture, get_cpu_features
+from easybuild.tools.systemtools import X86_64
+from easybuild.tools.utilities import trace_msg
+from easybuild.tools.build_log import print_error
+from easybuild.tools.filetools import remove_file, remove_dir
+from easybuild.tools import LooseVersion
+
+
+class EB_ESPResSo(CMakeNinja):
+    """Support for building and installing ESPResSo."""
+
+    def _get_extracted_tarball_paths(self):
+        """
+        Locate the source code of all dependencies.
+        """
+        extracted_paths = {}
+        for src in self.src:
+            name = src['name'].split('-', 1)[0]
+            # process main software
+            if name == 'espresso':
+                extracted_paths['espresso'] = src['finalpath']
+                continue
+            # process dependencies
+            tarball = src['name']
+            if not tarball.endswith('.tar.gz'):
+                raise ValueError(tarball + ' is not a tar.gz file')
+            prefix = tarball.rsplit('.', 2)[0]
+            matches = [x for x in os.listdir(src['finalpath']) if x.startswith(prefix)]
+            if len(matches) == 0:
+                raise RuntimeError(tarball + ' was not extracted')
+            if len(matches) > 1:
+                raise RuntimeError(tarball + ' matches multiple folders: ' + str(matches))
+            extracted_paths[name] = os.path.join(src['finalpath'], matches[0])
+        return extracted_paths
+
+    def _patch_fetchcontent(self):
+        """
+        Modify CMake ``FetchContent_Declare`` blocks to point to the folders
+        containing the already-downloaded dependencies rather than to URLs.
+        This avoids a download step during configuration.
+        """
+        extracted_paths = self._get_extracted_tarball_paths()
+        cmakelists_path = os.path.join(extracted_paths['espresso'], 'CMakeLists.txt')
+        with open(cmakelists_path, 'r') as f:
+            content = f.read()
+        for name, local_uri in extracted_paths.items():
+            if name == 'espresso':
+                continue
+            pattern = fr'FetchContent_Declare\(\s*{name}\s+GIT_REPOSITORY\s+\S+\s+GIT_TAG\s+\S+(?=\s|\))'
+            m = re.search(pattern, content, flags=re.IGNORECASE)
+            if m is None:
+                raise RuntimeError(f'{name} is not part of the ESPResSo FetchContent workflow')
+            content = re.sub(pattern, f'FetchContent_Declare({name} URL {local_uri}', content, flags=re.IGNORECASE)
+        with open(cmakelists_path, 'w') as f:
+            f.write(content)
+
+    def _get_version(self):
+        if '.' in self.version:
+            version = tuple(LooseVersion(self.version).version)
+        else:
+            version = 'commit'
+        return version
+
+    def _configure_step_release_420(self):
+        configopts = self.cfg.get('configopts', '')
+        dependencies = self.cfg.dependencies()
+        dependencies_names = {x.get('name', '') for x in dependencies}
+
+        def cmake_if_has(name):
+            return 'ON' if name in dependencies_names else 'OFF'
+
+        configopts += f' -DWITH_CUDA={cmake_if_has("CUDA")}'
+        configopts += f' -DWITH_GSL={cmake_if_has("GSL")}'
+        configopts += ' -DWITH_FFTW=ON'
+        configopts += ' -DWITH_PYTHON=ON'
+        configopts += ' -DWITH_SCAFACOS=OFF'
+        configopts += ' -DWITH_STOKESIAN_DYNAMICS=OFF'
+        configopts += ' -DWITH_TESTS=ON'
+
+        self.cfg['configopts'] = configopts
+
+    def _configure_step_release_500(self):
+        configopts = self.cfg.get('configopts', '')
+        dependencies = self.cfg.dependencies()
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/e/espresso.py
```diff
@@ -0,0 +1,195 @@
...
+            prefix = tarball.rsplit('.', 2)[0]
+            matches = [x for x in os.listdir(src['finalpath']) if x.startswith(prefix)]
+            if len(matches) == 0:
+                raise RuntimeError(tarball + ' was not extracted')
+            if len(matches) > 1:
+                raise RuntimeError(tarball + ' matches multiple folders: ' + str(matches))
+            extracted_paths[name] = os.path.join(src['finalpath'], matches[0])
+        return extracted_paths
+
+    def _patch_fetchcontent(self):
```
ocaisa, line 64: Can we not just provide the dependencies as we normally do in an easyconfig? Does ESPResSo not support using external packages?

jngrad: We support external packages, and prefer that over pulling the sources and building from scratch, except for waLBerla. We pull sources for waLBerla, which is highly configurable and is tricky to get right for ESPResSo. When ESPResSo builds a dependency from sources, it builds static libraries instead of shared objects, so that we don't end up with multiply defined shared objects.

ocaisa: Sorry, I should have paid more attention to what the easyblock is actually doing within the functions. At a first glance I thought it was trying to extract tarballs, but it is using the tarballs to try to figure out folders.

easybuild/easyblocks/e/espresso.py
```diff
@@ -0,0 +1,195 @@
...
+            # process main software
+            if name == 'espresso':
+                extracted_paths['espresso'] = src['finalpath']
+                continue
+            # process dependencies
+            tarball = src['name']
+            if not tarball.endswith('.tar.gz'):
+                raise ValueError(tarball + ' is not a tar.gz file')
+            prefix = tarball.rsplit('.', 2)[0]
+            matches = [x for x in os.listdir(src['finalpath']) if x.startswith(prefix)]
```
ocaisa: Isn't this assuming some very strict scenarios:
* All sources are `.tar.gz`. I guess this is true in the easyconfig PRs related to this but I wonder if you really _need_ to be so strict. Is there any scenario where you _don't_ want to process a source?
* All sources unpack to `<prefix>...`. Again this is likely true but you could be a little forgiving
```suggestion
            if not tarball.endswith('.tar.gz'):
                raise ValueError(tarball + ' is not a tar.gz file')
            prefix = tarball.rsplit('.', 2)[0]
            matches = [x for x in os.listdir(src['finalpath']) if x.lower().startswith(prefix.lower())]
```

easybuild/easyblocks/e/espresso.py
```diff
@@ -0,0 +1,195 @@
...
+"""
+
+import os
+import re
+
+from easybuild.easyblocks.generic.cmakeninja import CMakeNinja
+from easybuild.tools.systemtools import get_cpu_architecture, get_cpu_features
+from easybuild.tools.systemtools import X86_64
+from easybuild.tools.utilities import trace_msg
+from easybuild.tools.build_log import print_error
```
ocaisa:
```suggestion
from easybuild.tools.build_log import EasyBuildError, print_error
```
Use `EasyBuildError` so things end up in the logs

easybuild/easyblocks/e/espresso.py
```diff
@@ -0,0 +1,195 @@
...
+        return extracted_paths
+
+    def _patch_fetchcontent(self):
+        """
+        Modify CMake ``FetchContent_Declare`` blocks to point to the folders
+        containing the already-downloaded dependencies rather than to URLs.
+        This avoids a download step during configuration.
+        """
+        extracted_paths = self._get_extracted_tarball_paths()
+        cmakelists_path = os.path.join(extracted_paths['espresso'], 'CMakeLists.txt')
```
ocaisa: Need a sanity check here
```suggestion
        if 'espresso' in extracted_paths.keys():
            cmakelists_path = os.path.join(extracted_paths['espresso'], 'CMakeLists.txt')
        else:
            raise EasyBuildError(f"espresson not found in extracted_paths dict: {extracted_paths}")
```

easybuild/easyblocks/e/espresso.py
```diff
@@ -0,0 +1,195 @@
...
+
+        configopts += f' -DWITH_CUDA={cmake_if_has("CUDA")}'
+        configopts += f' -DWITH_GSL={cmake_if_has("GSL")}'
+        configopts += ' -DWITH_FFTW=ON'
+        configopts += ' -DWITH_PYTHON=ON'
+        configopts += ' -DWITH_SCAFACOS=OFF'
+        configopts += ' -DWITH_STOKESIAN_DYNAMICS=OFF'
+        configopts += ' -DWITH_TESTS=ON'
+
+        self.cfg['configopts'] = configopts
```
ocaisa: EasyBuild can help shorten this, see for example

https://github.com/easybuilders/easybuild-easyblocks/blob/0214837f651396dee3186487e6b3bc25835a1750/easybuild/easyblocks/r/r.py#L94-L101

so you could do this like
```suggestion
        for dep in ['CUDA', 'GSL,' 'FFTW', 'PYTHON', 'SCAFACOS']:
            dep_flag = "OFF"
            if get_software_root(dep):
                dep_flag = "ON"
            self.cfg.update('configopts', f"-DWITH_{dep.upper()}={dep_flag}")
        self.cfg.update('configopts', ' -DWITH_STOKESIAN_DYNAMICS=OFF')
        self.cfg.update('configopts', ' -DWITH_TESTS=ON')
```

easybuild/easyblocks/e/espresso.py
```diff
@@ -0,0 +1,195 @@
...
+
+    def _configure_step_release_500(self):
+        configopts = self.cfg.get('configopts', '')
+        dependencies = self.cfg.dependencies()
+        dependencies_names = {x.get('name', '') for x in dependencies}
+
+        def cmake_if_has(name):
+            return 'ON' if name in dependencies_names else 'OFF'
```
ocaisa: Similar to above, there is a good enough support in EasyBuild for a more direct approach for a lot of these optoons

ocaisa (changes requested):
Some suggestions to bring things more in line with other easyblocks

easybuild/easyblocks/e/espresso.py
```diff
@@ -0,0 +1,195 @@
...
+        for src in self.src:
+            name = src['name'].split('-', 1)[0]
+            # process main software
+            if name == 'espresso':
+                extracted_paths['espresso'] = src['finalpath']
+                continue
+            # process dependencies
+            tarball = src['name']
+            if not tarball.endswith('.tar.gz'):
+                raise ValueError(tarball + ' is not a tar.gz file')
```
ocaisa: Use `EasyBuildError` so things are tracked in the log

easybuild/easyblocks/e/espresso.py
```diff
@@ -0,0 +1,195 @@
...
+        cmakelists_path = os.path.join(extracted_paths['espresso'], 'CMakeLists.txt')
+        with open(cmakelists_path, 'r') as f:
+            content = f.read()
+        for name, local_uri in extracted_paths.items():
+            if name == 'espresso':
+                continue
+            pattern = fr'FetchContent_Declare\(\s*{name}\s+GIT_REPOSITORY\s+\S+\s+GIT_TAG\s+\S+(?=\s|\))'
+            m = re.search(pattern, content, flags=re.IGNORECASE)
+            if m is None:
+                raise RuntimeError(f'{name} is not part of the ESPResSo FetchContent workflow')
```
ocaisa: Use `EasyBuildError` so things are tracked in the log

easybuild/easyblocks/e/espresso.py
```diff
@@ -0,0 +1,195 @@
...
+            # process dependencies
+            tarball = src['name']
+            if not tarball.endswith('.tar.gz'):
+                raise ValueError(tarball + ' is not a tar.gz file')
+            prefix = tarball.rsplit('.', 2)[0]
+            matches = [x for x in os.listdir(src['finalpath']) if x.startswith(prefix)]
+            if len(matches) == 0:
+                raise RuntimeError(tarball + ' was not extracted')
+            if len(matches) > 1:
+                raise RuntimeError(tarball + ' matches multiple folders: ' + str(matches))
```
ocaisa: Use `EasyBuildError` so things are tracked in the log

easybuild/easyblocks/e/espresso.py
```diff
@@ -0,0 +1,195 @@
...
+
+        version = self._get_version()
+        if version == 'commit':
+            self._configure_step_release_500()
+        elif version[:2] >= (5, 0):
+            self._configure_step_release_500()
+        elif version[:2] >= (4, 2):
+            self._configure_step_release_420()
+        else:
+            raise NotImplementedError(
```
ocaisa: Use `EasyBuildError` so things are tracked in the log

ocaisa (commented):
I'm reasonably sure that we always use `EasyBuildError`

easybuild/easyblocks/e/espresso.py
jngrad: I've now expanded the code to check all commonly-used compression algorithms. This code finds the URI of the extracted tarball to overwrite the URL of the remote package (CMake will ignore the download and patch steps when an URI is provided). I would argue a mechanism like this one should be part of the `CMakeMake`/`CMakeNinja` easyblock, since we cannot download remote resources during the configure step.

The only use case for *not* processing a source is when it's a URI-based source. I'm not entirely sure how that would work, since paths to local files are platform-dependent. Pointing to an EESSI toolchain subdirectory wouldn't work, since that would prevent deploying ESPResSo to more than 1 toolchain. Pointing to a directory path passed as a CMake option could work, but I don't see any use case in ESPResSo.

easybuild/easyblocks/e/espresso.py
jngrad: Thanks, I've applied the changes in both methods.

boegel:
Quite a bit of what's going on in the easyconfigs for ESPResSo v5.0.0 can/should be moved into this easyblock, see:

- https://github.com/easybuilders/easybuild-easyconfigs/pull/25454
- https://github.com/easybuilders/easybuild-easyconfigs/pull/25455

boegel (approved):
lgtm

ocaisa (approved):
LGTM and lots of successful testing

PR #4070, opened 2026-02-18
[GROMACS] Support PLUMED for versions 2025 and newer

gkaf89:
This modification allows building versions of GROMACS newer than 2025.0 with support for PLUMED.

Any version of PLUMED is supported. From the [official documentation](https://manual.gromacs.org/documentation/2025.0/install-guide/index.html#building-with-plumed-support):

> GROMACS bundles the interface from version 2.10 of the [PLUMED library](https://www.plumed.org/) in its source distribution. The interface is compatible with any PLUMED version.

So in theory there is no need to check compatibility between versions of GROMACS and PLUMED in this new mechanism.

### DIFF
--- a/easybuild/easyblocks/g/gromacs.py
+++ b/easybuild/easyblocks/g/gromacs.py
@@ -155,6 +155,52 @@ class EB_GROMACS(CMakeMake):
         cuda = get_software_root('CUDA')
         return cuda and self.double_prec_pattern in self.cfg['configopts']
 
+    def plumed_through_patch(self):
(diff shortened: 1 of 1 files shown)


ostueker:
I just stumbled across this PR.  
Last fall I submitted #3984 to tackle GROMACS native PLUMED interface.

As far as I can tell, the native Plumed interface is not a full replacement for applying the patches authored by the Plumed team -- at least not yet.

Therefore being able to control whether GROMACS should use the native interface or patches should be applied is more versatile.

PR #4095, opened 2026-03-12
include `/usr/local` Python package directory for system-level Python package installations

### LINKED ISSUE #4100: `invalid command 'bdist_wheel'` error when installing SYSTEM Python packages, including EasyBuild on EL10 systems
Even `eb --install-latest-eb-release` may fail with 

```
Using pip 23.3.2 from /usr/lib/python3.12/site-packages/pip (python 3.12)
Processing /trinity/easybuild/build/EasyBuild/5.2.0/system-system/easybuild-framework-5.2.0
  Running command Preparing metadata (pyproject.toml)
  error: invalid command 'bdist_wheel'
  error: subprocess-exited-with-error
  
  × Preparing metadata (pyproject.toml) did not run successfully.
  │ exit code: 1
  ╰─> See above for output.
```

This can usually be resolved by installing the `wheel` Python package. However on EL10 ( e.g. Rocky 10) this still fails as there doesn't seem to be an OS wheel package. `pip install wheel` as `root` installs the wheel package to `/usr/local` where it isn't found when EasyBuild installs something.

Source of this issue is that there is a `/usr/lib64/python3.12/site.py` file which has this code:
```
if ENABLE_USER_SITE and 'RPM_BUILD_ROOT' not in os.environ:
    PREFIXES.insert(0, "/usr/local")
```

So `/usr/local` is **not** added when `PYTHONNOUSERSITE` is set, which we do set for all pip installations in EasyBuild. It only causes issues on such systems and when using `SYSTEM` python. The only source on where this originates is this SUSE CR: https://build.opensuse.org/projects/SUSE:SLE-15-SP6:Update/packages/python39.42007/files/F00251-change-user-install-location.patch?expand=0

`setuptools` >= 70.1 avoids the problem by "ingesting" the `bdist_wheel` functionality from `wheel`

It is also not possible to workaround this by setting `PYTHONPATH` because we remove those: https://github.com/easybuilders/easybuild-framework/blob/3c7b64d309cc3c87fe93edeacc9201025a2a1510/easybuild/tools/environment.py#L196-L203

A workaround from the user would be to install to /usr: `pip install --prefix=/usr wheel`

Flamefire:
There is a known situation on EL10 systems where the system admin used `pip install wheel` (as root) to install the `wheel` package required for most Python package installations.
However this was installed to `/usr/local/lib/python3.12/site-packages` but a site-config omits that directory when using `$PYTHONNOUSERSITE`. As wheel seems to be unavailable as a system package this is a burden and unexpected pitfall.

So just add this path (back) when using the SYSTEM toolchain.

Fixes https://github.com/easybuilders/easybuild-easyblocks/issues/4100

### DIFF
--- a/easybuild/easyblocks/generic/pythonpackage.py
+++ b/easybuild/easyblocks/generic/pythonpackage.py
@@ -714,9 +714,10 @@ class PythonPackage(ExtensionEasyBlock):
         This requires that 'pip install --prefix' is used, since the active Python installation scheme
         doesn't matter when using 'python setup.py install --prefix'.
         """
+        if not self.using_pip_install():
+            return False
         # see also  https://github.com/easybuilders/easybuild-easyblocks/issues/2976
-        py_install_scheme = det_py_install_scheme(python_cmd=self.python_cmd)
-        return py_install_scheme == PY_INSTALL_SCHEME_POSIX_LOCAL and self.using_pip_install()
+        return det_py_install_scheme(python_cmd=self.python_cmd) == PY_INSTALL_SCHEME_POSIX_LOCAL
 
     def should_use_ebpythonprefixes(self) -> bool:
         """
@@ -1054,13 +1055,20 @@ class PythonPackage(ExtensionEasyBlock):
         abs_bindir = os.path.join(actual_installdir, 'bin')
 
         # set PYTHONPATH and PATH as expected
+
+        # When using the system Python also include /usr/local as on some systems (e.g. EL 10 with Python 3.12)
+        # the site config at /usr/lib64/python3.12/site.py only adds the /usr/local folder when NOT using
+        # $PYTHONNOUSERSITE, which we set explicitely.
+        # This then causes packages installed by admins with `pip` to not be found.
+        if self.toolchain.is_system_toolchain():
+            abs_pylibdirs.append(os.path.join('/usr/local', self.pylibdir))
+
         old_values = {}
         for name, new_values in (('PYTHONPATH', abs_pylibdirs), ('PATH', [abs_bindir])):
             old_value = os.getenv(name)
             old_values[name] = old_value
             new_value = os.pathsep.join(new_values + ([old_value] if old_value else []))
-            if new_value:
-                env.setvar(name, new_value, verbose=False)
+            env.setvar(name, new_value, verbose=False)
 
         # actually install Python package
         cmd = self.compose_install_command(self.installdir)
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/generic/pythonpackage.py
```diff
@@ -1054,13 +1055,20 @@ def install_step(self):
         abs_bindir = os.path.join(actual_installdir, 'bin')
 
         # set PYTHONPATH and PATH as expected
+
+        # When using the system Python also include /usr/local as on some systems (e.g. EL 10 with Python 3.12)
+        # the site config at /usr/lib64/python3.12/site.py only adds the /usr/local folder when NOT using
+        # $PYTHONNOUSERSITE, which we set explicitely.
+        # This then causes packages installed by admins with `pip` to not be found.
```
boegel:
```suggestion
        # This then causes packages installed by admins with `pip` to not be found,
        # see also https://github.com/easybuilders/easybuild-easyblocks/issues/4100
```

easybuild/easyblocks/generic/pythonpackage.py
```diff
@@ -1054,13 +1055,21 @@ def install_step(self):
...
 
         # set PYTHONPATH and PATH as expected
+
+        # When using the system Python also include /usr/local as on some systems (e.g. EL 10 with Python 3.12)
+        # the site config at /usr/lib64/python3.12/site.py only adds the /usr/local folder when NOT using
+        # $PYTHONNOUSERSITE, which we set explicitely.
+        # This then causes packages installed by admins with `pip` to not be found,
+        # see also https://github.com/easybuilders/easybuild-easyblocks/issues/4100
+        if self.toolchain.is_system_toolchain():
+            abs_pylibdirs.append(os.path.join('/usr/local', self.pylibdir))
```
ocaisa: It may be overkill, but I'm worried about sysroot for this case, perhaps we can be a bit more careful if that is set?

Flamefire: Doesn't hurt I guess. Done

boegel (approved):
lgtm

boegel:
Test report by @boegel
 * **SUCCESS** _yaff-1.6.0-foss-2023b.eb_ 
  
 * **SUCCESS** _matplotlib-3.9.2-gfbf-2024a.eb_ 
  
 * **SUCCESS** _EasyBuild-5.2.1.eb_ 
  

Build succeeded for 3 out of 3 (total: 8 mins 55 secs) (3 easyconfigs in total)
node4232.shinx.os - Linux RHEL 9.6, x86_64, AMD EPYC 9654 96-Core Processor (zen4), Python 3.9.21
See https://gist.github.com/boegel/1ad0f1ad2d5fb1db57e0105b61e1596a for a full test report.

PR #4096, opened 2026-03-12
Limit number of parallel compile jobs in PyTorch tests by setting `$MAX_JOBS` + add support for using Arm Compute Library (ACL) as dependency

Flamefire:
PyTorch compiles extensions either JIT or AOT using Ninja. When `MAX_JOBS` isn't set, Ninja will determine the parallelism, likely using all available cores.
So limit them similar to the build/install step.

Also support the Arm Compute Library which may greatly enhance performance, see e.g. https://github.com/easybuilders/easybuild-easyconfigs/pull/21309

This doesn't change anything for now as we don't have easyconfigs yet

### DIFF
--- a/easybuild/easyblocks/p/pytorch.py
+++ b/easybuild/easyblocks/p/pytorch.py
@@ -534,7 +534,7 @@ class EB_PyTorch(PythonPackage):
 
         build_type = self.cfg.get('build_type')
         if build_type is None:
-            build_type = 'Debug' if self.toolchain.options.get('debug', None) else 'Release'
+            build_type = 'Debug' if self.toolchain.options.get('debug') else 'Release'
         else:
             for name in ('prebuildopts', 'preinstallopts', 'custom_opts'):
                 if '-DCMAKE_BUILD_TYPE=' in self.cfg[name]:
@@ -627,6 +627,7 @@ class EB_PyTorch(PythonPackage):
         env.setvar('SANDCASTLE', '1')
         # Skip this test(s) which is very flaky
         env.setvar('SKIP_TEST_BOTTLENECK', '1')
+        env.setvar('MAX_JOBS', str(self.cfg.parallel))
         if self.has_xml_test_reports:
             env.setvar(self.GENERATE_TEST_REPORT_VAR_NAME, '1')
         # Parse excluded_tests and flatten into space separated string
--- a/easybuild/easyblocks/p/pytorch.py
+++ b/easybuild/easyblocks/p/pytorch.py
@@ -448,6 +448,7 @@ class EB_PyTorch(PythonPackage):
         elif pytorch_version >= '1.9.0' and get_software_root('BLIS'):
             options.append('BLAS=BLIS')
             options.append('BLIS_HOME=' + get_software_root('BLIS'))
+            options.append('USE_MKLDNN=ON')
             options.append('USE_MKLDNN_CBLAS=ON')
         elif get_software_root('OpenBLAS'):
             # This is what PyTorch defaults to if no MKL is found.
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/p/pytorch.py
```diff
@@ -463,6 +464,13 @@ def add_enable_option(name, enabled):
         else:
             raise EasyBuildError("Did not find a supported BLAS in dependencies. Don't know which BLAS lib to use")
 
+        if pytorch_version >= '1.10':
+            acl_root = get_software_root('ArmComputeLibrary')
+            if acl_root:
+                options.append('USE_MKLDNN=ON')
```
ocaisa, line 470: Don't you also want `options.append('USE_MKLDNN_CBLAS=ON')` here (just from the comment in https://github.com/easybuilders/easybuild-easyconfigs/pull/21309#issue-2504670259 by @migueldiascosta )?

Flamefire: I don't see this being set in the PyTorch source. It defaults to OFF, and CI doesn't set it: 
https://github.com/pytorch/pytorch/blob/1d16a0978458457dc5c6b50bc19a37359a4bd822/.ci/pytorch/build.sh#L78-L82

Hence I left it disabled here too. It can be set using `custom_opts` e.g. via hooks

boegel:
CORE_CNT=16
EB_ARGS="--installpath /tmp/$USER/pr4096 PyTorch-2.9.1-foss-2024a.eb"

boegel:
Last test report for `PyTorch-2.9.1-foss-2024a.eb` took ~38h on `jsc-zen3` with 16 cores, let's see if there's a significant difference now (probably not, since 16 cores is all there is one those VMs, so I don't expect a positive impact by these changes in that type of setup), see https://github.com/easybuilders/easybuild-easyconfigs/pull/25240#issuecomment-3901730599

boegel:
On our A100 + AMD Zen3 system, an installation using 12 cores (out of the available 48) and 1 A100 GPU (out of 4 in the system) took ~46h. Let's see if it's any faster using this updated easyblock... Test install kicked off just now.

Flamefire:
> Build succeeded for 1 out of 1 (total: 36 hours 6 mins 14 secs) (1 easyconfigs in total)

Does this mean it was ~10hrs faster?

boegel:
Test report by @boegel
 * **SUCCESS** _PyTorch-2.7.1-foss-2024a-CUDA-12.6.0.eb_ 
  

Build succeeded for 1 out of 1 (total: 56 hours 24 mins 9 secs) (1 easyconfigs in total)
node3900.accelgor.os - Linux RHEL 9.6, x86_64, AMD EPYC 7413 24-Core Processor (zen3), 1 x NVIDIA NVIDIA A100-SXM4-80GB, 590.48.01, Python 3.9.21
See https://gist.github.com/boegel/7648f6ec2b40c5386c9e51eab4b4b254 for a full test report.

Flamefire:
You can i ignore the force-push, just undid a commit I just added to this PR.
Looks like we have imkl & imkl-FFTW, but the latter depends on the former so checking for imkl is enough.

boegel:
> On our A100 + AMD Zen3 system, an installation using 12 cores (out of the available 48) and 1 A100 GPU (out of 4 in the system) took ~46h. Let's see if it's any faster using this updated easyblock... Test install kicked off just now.

It actually took *56* hours now, instead of 46, see test report in https://github.com/easybuilders/easybuild-easyblocks/pull/4096#issuecomment-4145951050

@Flamefire How does that make any sense...

Flamefire:
I wish I knew. Could as well be the filesystem.
The logs contain the timings for individual suites. That might provide a hint which got much slower. But probably not worth spending much time on

boegel (approved):
lgtm

PR #4098, opened 2026-03-23
enhance custom easyblock for `intel-compilers` to support sysroot + dynamic linker

adammccartney:
Uppon passing sysroot, determine the path to dynamic linker and add both options to the compiler config files.


boegel:
@ocaisa Can you take a look at this (no rush, this will have to wait until after EasyBuild v5.3.0 release...)

adammccartney:
Maybe we should mark as a draft, I'm pretty sure it's not yet working for more complicated stuff. I tried to build vasp 6.6.0 using these compilers + config. It produced a working binary, but threw lots of errors at runtime. Needs further investigation.

PR #4105, opened 2026-04-08
add mapping to LAMMPS easyblock for `9.0a`, `10.0f` and `12.0f` CUDA compute capabilities

laraPPr:


boegel:
CORE_CNT=16
EB_ARGS="--installpath /tmp/$USER/pr4105 LAMMPS-22Jul2025-foss-2024a-kokkos.eb"

boegel (approved):

Thyre:
We should port these mapping changes to the Kokkos EasyBlock as well. Probably worth to remove the duplication in both EasyBlocks and just rely on the ones in the Kokkos EasyBlock instead.

PR #4120, opened 2026-04-27
adding the gcc buildenv easyblock

agimenog:
Regarding this ticket: https://gitlab.com/eessi/support/-/work_items/213

### DIFF
--- a/easybuild/easyblocks/g/gcc-buildenv.py
+++ b/easybuild/easyblocks/g/gcc-buildenv.py
@@ -1,0 +1,1268 @@
+##
+# Copyright 2009-2025 Ghent University
+#
+# This file is part of EasyBuild,
+# originally created by the HPC team of Ghent University (http://ugent.be/hpc/en),
+# with support of Ghent University (http://ugent.be/hpc),
+# the Flemish Supercomputer Centre (VSC) (https://www.vscentrum.be),
+# Flemish Research Foundation (FWO) (http://www.fwo.be/en)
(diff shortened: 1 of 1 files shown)


ocaisa:
This looks like a modified copy of the GCC easyblock, wouldn't it be better to add the capabilities to that easyblock instead?

easybuild/easyblocks/g/gcc-buildenv.py
```diff
@@ -0,0 +1,1268 @@
...
+            else:
+                raise EasyBuildError("Failed to isolate GCC build directory in %s", self.builddir)
+
+        else:
+            super().install_step(*args, **kwargs)
+
+            # Generate the rpath wrappers (if enabled)
+            if self.cfg['rpath_wrappers']: 
+                wrappers_dir = os.path.join(self.rpath_wrappers_dir, RPATH_WRAPPERS_SUBDIR)
+                if os.path.exists(wrappers_dir):
```
ocaisa, line 1009: This doesn't look right to me, you need to create the RPATH wrappers for the GCC compiler that you are installing (not the toolchain of the easyconfig)

ocaisa (changes requested):
It's hard to figure out here what the new code is. It looks like you are trying add some capabilities to the GCC easyblock, but all that code is duplicated so it's very hard to identify.

agimenog:
Hi @ocaisa, closing this PR and reopening as commented in the meeting

PR #4065, opened 2026-02-16
Fix setting of component directories when using buildininstalldir

Flamefire:
(created using `eb --new-pr`)

Fixes #4037 see https://github.com/easybuilders/easybuild-easyconfigs/pull/23852#issuecomment-3908856361

PR #4122, opened 2026-04-28
Fix for running `--sanity-check-only` for a `JuliaBundle`

Crivella:
Currently trying to run `--sanity-check-only` on a `JuliaBundle`-based EC fails due to the dependencies not being loaded when the main package is tested

```
  File "/home/crivella/Documents/GIT/easybuild-easyblocks/easybuild/easyblocks/generic/juliapackage.py", line 138, in julia_env_path
    julia_version = get_software_version('Julia').split('.')
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'split'
```

This in particular happens due to `get_software_version('Julia')` being called before the module is loaded

This PR adds same section as for `PythonPackage` to `JuliaPackage` to load module early

### DIFF
--- a/easybuild/easyblocks/generic/juliapackage.py
+++ b/easybuild/easyblocks/generic/juliapackage.py
@@ -38,7 +38,7 @@ import easybuild.tools.environment as env
 from easybuild.framework.easyconfig import CUSTOM
 from easybuild.framework.extensioneasyblock import ExtensionEasyBlock
 from easybuild.tools.build_log import EasyBuildError
-from easybuild.tools.modules import get_software_root, get_software_version
+from easybuild.tools.modules import get_software_root
 from easybuild.tools.filetools import copy_dir, mkdir
 from easybuild.tools.run import run_shell_cmd
 from easybuild.tools.utilities import trace_msg
--- a/easybuild/easyblocks/generic/juliapackage.py
+++ b/easybuild/easyblocks/generic/juliapackage.py
@@ -118,11 +118,31 @@ class JuliaPackage(ExtensionEasyBlock):
 
         return parsed_var
 
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._julia_version = None
+
+    @property
+    def julia_version(self) -> str:
+        """Needs to get Julia version from dependencies to allow --sanity-check-only to generate the fake module"""
+        if self._julia_version is None:
+            deps = self.cfg.dependencies(runtime_only=True)
+            for dep in deps:
+                if dep['name'] == 'Julia':
+                    self._julia_version = dep['version']
+                    break
+            else:
+                raise EasyBuildError(
+                    "Julia not included as dependency, cannot determine Julia version for installation of: %s",
+                    self.name
+                )
+        return self._julia_version
+
     def julia_env_path(self, absolute=True, base=True):
         """
         Return path to installation environment file.
         """
-        julia_version = get_software_version('Julia').split('.')
+        julia_version = self.julia_version.split('.')
         env_dir = "v{}.{}".format(*julia_version[:2])
         project_env = os.path.join("environments", env_dir, "Project.toml")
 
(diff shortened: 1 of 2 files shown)


Flamefire:
I'm not sure this is the right place: PythonPackage does this because it uses the module-dependent stuff directly afterwards. https://github.com/Crivella/easybuild-easyblocks/blob/0c62eb9f874ae91a4b9dce397c3b225838569f52/easybuild/easyblocks/generic/pythonpackage.py#L1218

This isn't the case here.

Can you post the full traceback?

Crivella:
It is the same as in https://github.com/easybuilders/easybuild-easyblocks/pull/4065#issuecomment-3915536689

The problem arise from

https://github.com/easybuilders/easybuild-easyblocks/blob/17d584071c73bbb51ce0ff9f58d17f201027e074/easybuild/easyblocks/generic/juliapackage.py#L324

where

https://github.com/easybuilders/easybuild-easyblocks/blob/17d584071c73bbb51ce0ff9f58d17f201027e074/easybuild/easyblocks/generic/juliapackage.py#L125

is called before the `Julia` module is loaded making `get_software_version('Julia')` result in `None` and `get_software_version('Julia').split('.')` hence giving that exception

Flamefire:
I see, but this reasoning is not correct:
>  since custom actions taken below require that environment is set up properly already

The problem we have here: In order to create the **module**  we need to load the module.
That is an entirely different issue and we should fix that instead.

The fix here works because for `sanity-check-only` the real module is loaded instead of creating one.
Another issue is that in `ExtensionEasyBlock` we use the fake module directly instead of `sanity_check_module`. The latter would load the real module as would the base easyblock

I have that fixed locally so I think I have an open PR for that. Looking for that or creating it.
Maybe we should try with that first as it is more general.

Crivella:
I guess if we want to be more specific we should load the only Julia module that should be available regardless beforehand

Flamefire:
What do you mean by that exactly?

I think we shouldn't do anything special. The current approach will work fine, if the sanity-check-only doesn't create a fake module.
And even with that we don't need to actually load the Julia module just to get the version: It should be a dependency of the easyconfig, so we can get it from there.

Crivella:
I'd be careful of touching the ExtensionEasyblock.

Indeed we can take the version of Julia from the deps, the weird thing happening after is that in

`--sanity-check-only`, `self.toolchain.deps` from

https://github.com/easybuilders/easybuild-framework/blob/632d4d24ca91f19077aaa52c52020791d82596a3/easybuild/framework/easyblock.py#L1464

is not set, while `self.cfg.dependencies()` still gives the correct result.

This results in `depends_on('Julia/XXX')` not being in the fake module causing the first check to fail finding the Julia executable... :thinking: 

EDIT: i guess it comes down to the fact that we are skipping the `prepare_step` where this is being set

https://github.com/easybuilders/easybuild-framework/blob/632d4d24ca91f19077aaa52c52020791d82596a3/easybuild/framework/easyblock.py#L3054

Crivella:
While

- https://github.com/easybuilders/easybuild-framework/pull/5182

is not required in general.
It is needed for an easyconfig that implements a `sanity_check_command` for the bundle such as

```
easyblock = 'JuliaBundle'
...

toolchain = SYSTEM

dependencies = [
    ('Julia', _julia_ver),
]

# Order is important as all dependencies of a package must be installed before the package itself
exts_list = [
    ....
]

sanity_check_commands = [
    f"julia -e 'using Pkg; Pkg.test(\"{name}\")'",
]
```

As `Julia`'s module would not be loaded otherwise by the fake_module when testing.

@Flamefire 
I think loading the existing module instead of the fake one could cause trouble in case of `--module-only --rebuild` in case changes to the Module are being made, but should be fine with `--sanity-check-only` (not sure if `--module-only` also sets the flags that would trigger the behavior of `--sanity-check-only`)

Flamefire (approved):
LGTM independent of other issues. Creating a module shouldn't depend on a module being loaded

Crivella:
Closing in favor of #4146 which includes these changes

PR #4124, opened 2026-05-01
`JuliaPackage`: Allow running tests in a dedicated test_step

Crivella:
Currently the only way to run the testsuite of a Julia package is to do so in the `sanity_check_commands`

This PR adds a way to use `runtest` and `testcmd` in the extension options to run the test suite of the package after installation.

## To discuss

- Currently the test is ran allowing installation of online packages. This is done because tests can require extra dependencies that are not installed with the package.
  The alternative currently would be to install the test dependencies as normal one so that they are available to the test step. Alternatively we would need to add a new `test_dep` flag to Julia package so that the extension can be downloaded and added to the test_dir during test but would require altering the current logic a lot
- Possibly add a `julia_cmd` detection similar to `python_cmd` for `PythonPackage`

### DIFF
--- a/easybuild/easyblocks/generic/juliapackage.py
+++ b/easybuild/easyblocks/generic/juliapackage.py
@@ -31,6 +31,7 @@ import ast
 import glob
 import os
 import re
+import tempfile
 
 from easybuild.tools import LooseVersion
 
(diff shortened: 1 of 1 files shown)


Crivella:
Closing in favor of https://github.com/easybuilders/easybuild-easyblocks/pull/4146 which includes these changes

PR #4126, opened 2026-05-03
Adapt easyblocks for HIP

zerefwayne:
- [UCX-ROCm](https://github.com/easybuilders/easybuild-easyconfigs/pull/25902/changes#diff-01fc14f3babdf3f264fae8f3fafd74e09368a0b013d546d649e83a15c144b2e5) can be configured for rocm with the options passed in the easyblock. This change in EB_UCX_Plugins makes the software_root check aligned with the checks used in [ROCmComponent](https://github.com/easybuilders/easybuild-easyblocks/blob/develop/easybuild/easyblocks/generic/rocmcomponent.py)


zerefwayne:
Not needed anymore, switched to ConfigureMake easyblock for UCX-ROCm.

PR #4129, opened 2026-05-05
Add PTX code for (highest) CUDA compute capability in PyTorch & torchvision

Flamefire:
E.g. "5.0+PTX" makes PyTorch add PTX code.
We can simply add it to the list in `TORCH_CUDA_ARCH_LIST` to add PTX code for the last architecture.

Uses parts of #4092. So with https://github.com/easybuilders/easybuild-framework/pull/5144 it will add PTX code for the highest arch

### DIFF
--- a/easybuild/easyblocks/p/pytorch.py
+++ b/easybuild/easyblocks/p/pytorch.py
@@ -503,20 +503,16 @@ class EB_PyTorch(PythonPackage):
                 options.append('USE_SYSTEM_NCCL=1')
                 options.append('NCCL_INCLUDE_DIR=' + os.path.join(nccl_root, 'include'))
 
-            # list of CUDA compute capabilities to use can be specifed in two ways (where (2) overrules (1)):
-            # (1) in the easyconfig file, via the custom cuda_compute_capabilities;
-            # (2) in the EasyBuild configuration, via --cuda-compute-capabilities configuration option;
-            cuda_cc = build_option('cuda_compute_capabilities') or self.cfg['cuda_compute_capabilities']
-            if not cuda_cc:
-                raise EasyBuildError('List of CUDA compute capabilities must be specified, either via '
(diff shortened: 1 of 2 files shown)


easybuild/easyblocks/p/pytorch.py
```diff
@@ -493,30 +493,26 @@ def add_enable_option(name, enabled):
...
-
-            self.log.info('Compiling with specified list of CUDA compute capabilities: %s', ', '.join(cuda_cc))
+                options.append('NCCL_INCLUDE_DIR=' + os.path.join(f'${nccl_var}', 'include'))
+
+            cuda_arch_list = self.cfg.get_cuda_cc_template_value('cuda_cc_semicolon_sep')
+            self.log.info('Compiling with specified list of CUDA compute capabilities: %s',
+                          ', '.join(cuda_arch_list.split(';')))
+            if cuda_arch_list:
+                self.log.info('Also creating PTX code for architecture ' + cuda_arch_list.split(';')[-1])
+                cuda_arch_list += '+PTX'
```
boegel, line 511: As currently proposed, we'll start doing this for *every* PyTorch version...

Maybe we should only do this for (very) recent PyTorch versions (say 2.9+), along with a way to explicitly opt-in to doing this?

boegel (changes requested):

easybuild/easyblocks/p/pytorch.py
Flamefire: IIRC we new cuda sanity checks check that the highest CC has PTX code and fails otherwise. So to me this seems like a bugfix

If you want I can run some builds with e.g. oldest and newest 1.x and 2.x version we have or select some randomly. I'd expect it to either work everywhere or fail everywhere as the biggest "risk" I see is if some PyTorch version didn't support it yet although it seems it's been always the case.

PR #4130, opened 2026-05-05
add `$CUDNN_HOME` and `$CUDNN_PATH` as extra environment variables to be set by cuDNN easyblock

ccoulombe:
Helps some software to build and find libcudnn at runtime.

Like `CUDA_HOME`, but for `cuDNN`.

### DIFF
--- a/easybuild/easyblocks/c/cudnn.py
+++ b/easybuild/easyblocks/c/cudnn.py
@@ -82,11 +82,6 @@ class EB_cuDNN(Tarball):
     def make_module_extra(self):
         """Set the install directory as CUDNN_HOME, CUDNN_PATH."""
 
-        # avoid adding of installation directory to $PATH (cfr. Binary easyblock) since that may cause trouble,
-        # for example when there's a clash between command name and a subdirectory in the installation directory
-        # (like compute-sanitizer)
-        self.cfg['prepend_to_path'] = False
-
         txt = super().make_module_extra()
         txt += self.module_generator.set_environment('CUDNN_HOME', self.installdir)
         txt += self.module_generator.set_environment('CUDNN_PATH', self.installdir)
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/c/cudnn.py
```diff
@@ -78,3 +78,17 @@ def fetch_step(self, *args, **kwargs):
...
         )
         return super().fetch_step(*args, **kwargs)
+
+    def make_module_extra(self):
+        """Set the install directory as CUDNN_HOME, CUDNN_PATH."""
+
+        # avoid adding of installation directory to $PATH (cfr. Binary easyblock) since that may cause trouble,
+        # for example when there's a clash between command name and a subdirectory in the installation directory
+        # (like compute-sanitizer)
+        self.cfg['prepend_to_path'] = False
```
bartoldeman: cudnn derives from Tarball instead of Binary, that's why CI didn't like this.

bartoldeman (changes requested):

easybuild/easyblocks/c/cudnn.py
ccoulombe: Then I should remove this ? Or revert back to the `modextravars` config previously used?

bartoldeman: You can remove that, it's needed to override the default behaviour of the Binary easyblock, but since cudnn doesn't derive from Binary there is nothing to override.

ccoulombe: done

bartoldeman:
EB_ARGS="cuDNN-8.9.7.29-CUDA-12.3.0.eb cuDNN-9.15.0.57-CUDA-12.9.1.eb --installpath /tmp/$USER/ebpr-4130"

bartoldeman (approved):
lgtm

PR #4132, opened 2026-05-11
enhance custom easyblock for OpenMPI to add ROCm support

zerefwayne:
This pull request adds rocm specific dependencies to known_dependencies and sanity check commands to ensure OpenMPI is properly linked to rocm libraries.

### DIFF
--- a/easybuild/easyblocks/o/openmpi.py
+++ b/easybuild/easyblocks/o/openmpi.py
@@ -76,6 +76,14 @@ class EB_OpenMPI(ConfigureMake):
         if LooseVersion(self.version) >= '5.0.0':
             known_dependencies.append('PRRTE')
 
+            # ROCm support added to OpenMPI after 5.0.x
+            rocmroot = get_software_root('ROCm-LLVM')
+            if rocmroot:
+                # remove plain UCC and UCX
+                known_dependencies = [d for d in known_dependencies if d not in ('UCX', 'UCC')]
+                # replace with rocm versions
+                known_dependencies.extend(['HIP', 'UCX-ROCm', 'UCC-ROCm'])
+
         # Value to use for `--with-<dep>=<value>` if the dependency is not specified in the easyconfig
         # No entry is interpreted as no option added at all
         # This is to make builds reproducible even when the system libraries are changed and avoids failures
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/o/openmpi.py
```diff
@@ -76,6 +76,14 @@ def config_opt_used(key, enable_opt=False):
         if LooseVersion(self.version) >= '5.0.0':
             known_dependencies.append('PRRTE')
 
+            # ROCm support added to OpenMPI after 5.0.x
+            rocmroot = get_software_root('ROCm-LLVM')
+            if rocmroot:
+                # remove plain UCC and UCX
+                known_dependencies = [d for d in known_dependencies if d not in ('UCX', 'UCC')]
+                # replace with rocm versions
+                known_dependencies.extend(['HIP', 'UCX-ROCm', 'UCC-ROCm'])
```
zerefwayne, line 85: UCX-ROCm and UCC-ROCm are built as complete standalone installations rather
than as additional component modules layered on top of the base UCX/UCC builds.

Two problems emerged with the layered approach:

1. When a plain UCX build and a UCX-ROCm component module were both present in the module environment, ucx_info reported the plain UCX configuration (without --with-rocm) because the plain UCX binary took precedence on PATH. The ROCm transport components were built, but the wrong ucx_info was being invoked, making it appear that ROCm support was absent even when it was not.

2. UCC-ROCm depends on UCX. When UCC-ROCm was built as a component module on top of a ROCm-aware UCX, loading it would also pull in the plain UCX module as a listed dependency, which then shadowed the ROCm-aware UCX on PATH and in LD_LIBRARY_PATH. The result was a UCC-ROCm build backed at runtime by a UCX with no ROCm transport support.

Building UCX-ROCm and UCC-ROCm as fully independent installations avoids both conflicts. The trade-off is a larger on-disk footprint, but I think it is acceptable given that these modules are only loaded in ROCm-aware toolchains.

Thyre (approved):
LGTM

Thyre:
EB_ARGS="--installpath /tmp/$USER/ebpr-4132 OpenMPI-4.1.4-GCC-11.3.0.eb OpenMPI-5.0.10-llvm-compilers-21.1.8.eb OpenMPI-5.0.3-intel-compilers-2024.2.0.eb OpenMPI-4.1.6-GCC-13.2.0.eb"

Thyre:
`FAIL: opal_path_nfs` is not unsurprising, so that's certainly unrelated to the changes here.

easybuild/easyblocks/o/openmpi.py
boegel: Disk space is definitely not a big problem here...

boegel (approved):
lgtm

boegel:
Test report by @boegel
 * **SUCCESS** _OpenMPI-4.1.4-GCC-11.3.0.eb_ 
  
 * **SUCCESS** _OpenMPI-4.1.5-GCC-12.3.0.eb_ 
  
 * **SUCCESS** _OpenMPI-5.0.3-GCC-13.3.0.eb_ 
  
 * **SUCCESS** _OpenMPI-5.0.10-GCC-15.2.0.eb_ 
  

Build succeeded for 4 out of 4 (total: 2 hours 23 mins 48 secs) (4 easyconfigs in total)
node4202.shinx.os - Linux RHEL 9.6, x86_64, AMD EPYC 9654 96-Core Processor (zen4), Python 3.9.21
See https://gist.github.com/boegel/9518fed6c4dd10459b663402601d7201 for a full test report.

PR #4136, opened 2026-05-13
adding the gcc buildenv easyblock

agimenog:
Regarding this ticket: https://gitlab.com/eessi/support/-/work_items/213

Hi @ocaisa, reopening it with the modifications on the gcc.py file

### DIFF
--- a/easybuild/easyblocks/g/gcc.py
+++ b/easybuild/easyblocks/g/gcc.py
@@ -1,5 +1,5 @@
 ##
-# Copyright 2009-2026 Ghent University
+# Copyright 2009-2025 Ghent University
 #
 # This file is part of EasyBuild,
 # originally created by the HPC team of Ghent University (http://ugent.be/hpc/en),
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/g/gcc.py
```diff
@@ -997,6 +1002,27 @@ def install_step(self, *args, **kwargs):
         else:
             super().install_step(*args, **kwargs)
 
+            # Generate the rpath wrappers (if enabled)
+            if self.cfg['rpath_wrappers']:
+                wrappers_dir = os.path.join(self.rpath_wrappers_dir, RPATH_WRAPPERS_SUBDIR)
+                if os.path.exists(wrappers_dir):
+                    self.rpath_wrappers_dir = os.path.join(self.installdir, 'bin')
+                    rpath_wrappers_path = os.path.join(self.rpath_wrappers_dir, RPATH_WRAPPERS_SUBDIR)
```
ocaisa, line 1010: This is (I think) installing the wrappers for the toolchain used, not the rpath wrappers for the GCC that was just installed.

easybuild/easyblocks/g/gcc.py
```diff
@@ -997,6 +1002,27 @@ def install_step(self, *args, **kwargs):
...
+                    copy_dir(wrappers_dir, rpath_wrappers_path)
+                    py_exe = None  # test to use the python from EB
+                    if py_exe is not None:
```
ocaisa, line 1013: I don't follow this, this conditions is never not None

ocaisa: If you use the framework function to create the wrappers, you would not need this anyway, the code below is part of that function

easybuild/easyblocks/g/gcc.py
```diff
@@ -1224,3 +1250,17 @@ def sanity_check_step(self):
...
+        Specify correct bin directories for buildenv installation and add environment variables
+        for the toolchain build environment.
+        """
+        wrappers_dir = os.path.join(self.rpath_wrappers_dir, RPATH_WRAPPERS_SUBDIR)
+        if os.path.exists(wrappers_dir):
+            self.module_load_environment.PATH = [
+                os.path.join(wrappers_dir, d)
+                for d in os.listdir(wrappers_dir)
+                if os.path.isdir(os.path.join(wrappers_dir, d))
+            ]
```
ocaisa, line 1265: The wrappers should only be exposed conditionally in the module file: an EasyBuild session should ignore them and create new ones, a user sessions should use them

ocaisa (changes requested):
This also requires additional sanity checks.

PR #4139, opened 2026-05-21
DRAFT: Enable new optional flags for petsc single double precisions

rreho:
With this commit I added few extra config options to enable real, complex, single and double precision compilation.
We would like to introduce a for loop that compiles in a single run the options alltogether and it's something to do for future work @nicspalla


ocaisa:
@rreho If you build with the various options, do you get different library names? It would be great if you do as otherwise there would be a name clash and it wouldn't be possible.

PR #4141, opened 2026-05-23
Add new generic EasyBlock HuggingFaceDataset

VRehnberg:
Extends the generic Dataset easyblock for any dataset from <https://huggingface.co>. Currently only handles datasets (not models etc), but I imagine it wouldn't be too hard to extend it.

### DIFF
--- a/easybuild/easyblocks/generic/huggingfacedataset.py
+++ b/easybuild/easyblocks/generic/huggingfacedataset.py
@@ -79,7 +79,7 @@ class HuggingFaceDataset(Dataset):
             ).output.strip()
 
         for src_spec in self.cfg['data_sources']:
-            _url = f"hf://datasets/{self.cfg['hf_name']}@{self.cfg['hf_revision']}/{src_spec['filename']}"
+            _url = f"hf://datasets/{self.cfg['hf_name']}@{self.cfg['hf_revision']}/{src_spec['download_filename']}"
             hash_filename = os.path.join(
                 _hf_cache_download_dir,
                 _hash_url_to_filename(_url)
--- a/easybuild/easyblocks/generic/huggingfacedataset.py
+++ b/easybuild/easyblocks/generic/huggingfacedataset.py
@@ -28,11 +28,13 @@ EasyBuild support for installing datasets
 @author: Viktor Rehnberg (Chalmers University of Technology)
 """
 import os
+from tempfile import TemporaryDirectory
 
 from easybuild.easyblocks.generic.dataset import Dataset
 from easybuild.framework.easyconfig import CUSTOM, MANDATORY
-from easybuild.tools.filetools import change_dir, clean_dir, expand_glob_paths, find_glob_pattern, mkdir, move_file
-from easybuild.tools.filetools import write_file
+from easybuild.tools.build_log import EasyBuildError
+from easybuild.tools.filetools import change_dir, clean_dir, dir_contains_files, expand_glob_paths, find_glob_pattern
+from easybuild.tools.filetools import mkdir, move_file, write_file
 from easybuild.tools.run import run_shell_cmd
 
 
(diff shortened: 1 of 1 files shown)


VRehnberg:
Test report by @VRehnberg
 * **SUCCESS** _CIFAR-10-data-20240104-hf-gfbf-2025a.eb_ 
  

Build succeeded for 1 out of 1 (total: 34 secs) (1 easyconfigs in total)
vera-icelake-build - Linux Rocky Linux 9.6, x86_64, Intel(R) Xeon(R) Silver 4316 CPU @ 2.30GHz, Python 3.9.21
See https://gist.github.com/VRehnberg/4d7f04b08044aecfe264819399ddbc75 for a full test report.

VRehnberg:
Running CIFAR-10 dataset with 2023a and 2025a leads to .arrow files which are not identical. I will have to investigate some more, might be that 2.18 to 4.5 is simply a too big a jump.

VRehnberg:
Is fine between 4.0.0 and 4.5.0 and 2.18.0 against a separate 2.18.0 build gives same results. Good enough for me.

VRehnberg:
There is some potential future work for this easyblock (not planned for this PR):

- ~adapting it for model weights~
- reducing disk footprint during build

## Adapting it for model weights
This is no longer especially feasible when I have switched to using save_to_disk. Model weights would have worked better with just running HF_MODEL="%(installdir)s".

On the other hand providing model weights might be simple enough to just use the base Dataset block.

## Reducing disk footprint during build
For large datasets this can currently be an issue. The disk footprint is:

- Data source directory:
    - Copy of dataset (usually in parquet format)
- Build directory:
    - Copy of source files
    - Processed dataset in arrow files in load cache
    - Processed dataset from save_to_disk
- Install directory:
    - Moving processed dataset result from save_to_disk

So at one point the dataset exists as four copies in different places in the filesystem. There are some possible options to mitigate this, but nothing I've thought seems as robust.

easybuild/easyblocks/generic/huggingfacedataset.py
```diff
@@ -0,0 +1,138 @@
...
+            return run_shell_cmd(
+                f'python -c "import datasets; print(datasets.utils.file_utils.hash_url_to_filename(\'{url}\'))"'
+            ).output.strip()
+
+        for src_spec in self.cfg['data_sources']:
+            _url = f"hf://datasets/{self.cfg['hf_name']}@{self.cfg['hf_revision']}/{src_spec['download_filename']}"
+            hash_filename = os.path.join(
+                _hf_cache_download_dir,
+                _hash_url_to_filename(_url)
+            )
```
VRehnberg: Doing this as a run_shell_cmd is quite slow. About 6 seconds per filename, which is fine for most things, but when there are 100s of files or more it adds a noteable overhead.

Should consider doing this in a batch with a single run_shell_cmd.

VRehnberg: This function is no longer needed and has been removed.

easybuild/easyblocks/generic/huggingfacedataset.py
```diff
@@ -0,0 +1,138 @@
...
+    '''Support for installing datasets from huggingface.co'''
+
+    @staticmethod
+    def extra_options(extra_vars=None):
+        """Extra easyconfig parameters specific to Data easyblock."""
+        extra_vars = Dataset.extra_options(extra_vars)
+        extra_vars.update({
+            'hf_name': [None, "Name of dataset on huggingface.co e.g. `uoft-cs/cifar10`", MANDATORY],
+            'hf_revision': [None, "Version tag or commit hash on huggingface.co", MANDATORY],
+            'extract_sources': [False, "Whether or not to extract data sources", CUSTOM],
```
VRehnberg, line 50:
```suggestion
            'extract_sources': [False, "Whether or not to extract data sources", CUSTOM],
            'object_storage_ignore_dirs': [["."], "List of directories (relative to installdir) to be excluded from "
                                               "object storage (use '.' for full installdir)", CUSTOM],

```

by disabling object storage, then the versionsuffix for datasets using this easyblock could be simplified. The argument for including HF-Datasets version/toolchain version is no longer as motivated.

boegel:
@smoors Can you pick up on this?

PR #4146, opened 2026-06-04
Rework of the `JuliaPackage` easyblock

### LINKED ISSUE #4123: `JuliaPackage` can't really work with depedencies
`JuliaPackage` uses 

https://github.com/easybuilders/easybuild-easyblocks/blob/17d584071c73bbb51ce0ff9f58d17f201027e074/easybuild/easyblocks/generic/juliapackage.py#L234

to bring in packages that were installed as an extension of a dependency.

The problem arise from the fact that the order of the dependencies matter.
Dependent packages must be installed after their dependencies or the package manager will raise errors such as

```
ERROR: Unsatisfiable requirements detected for package Preferences [21216c6a]:
 Preferences [21216c6a] log:
 ├─Preferences [21216c6a] has no known versions!
 └─restricted to versions 1 by PrecompileTools [aea7be01] — no versions left
   └─PrecompileTools [aea7be01] log:
     ├─possible versions are: 1.3.3 or uninstalled
     └─PrecompileTools [aea7be01] is fixed to version 1.3.3
```

this is happening because `PrecompileTools` is being installed  before `Preferences` which is one of its dependencies.

Right now `include_pkg_dependencies` includes and re-install the dependencies in an undefined order based on the results of 

https://github.com/easybuilders/easybuild-easyblocks/blob/17d584071c73bbb51ce0ff9f58d17f201027e074/easybuild/easyblocks/generic/juliapackage.py#L242

Crivella:
## Summary of changes

- Adds compile_cache check to sanity checks to ensure that the compiled version of the package can actually be found/re-used
- Allow specifying test-only dependencies with the `is_test_dependency` EC param
- Allow turning on Julia's verbose output with the `julia_debug` EC param
- Allow reusing packages from a dependency (fix #4123)
  - The environments of the dependencies are merged into the one of the package being installed
- Use `Pkg.instantiate()` to install all extensions instead of installing them one by one with `Pkg.develop`
  - Remove need for deps to be ordered in the EC files
  - Makes use of Julia's capability to install extension in parallel (still respecting the `max_parallel` option)
- Allows running package tests in a dedicated step (setting `runtest` to True for every extension that needs testing)
   - The test are both skippable (`--skip-test-step`) and ignorable (`--ignore-test-failure`)
   - Tests are ran in offline mode by default unless `test_online` is set to true
   - If both offline and online test are to be performed, the offline ones go first as they should not modify the test environment that we are setting up
- Switch to using an EB specific environment variable to keep track of the LOAD/DEPOT PATHS and use them to correctly set them for Julia in a site-specific startup script

## Maybe TODO

- [X] Improve Lua/Tcl syntax in module footer
- [X] Modify Julia itself similar to https://github.com/easybuilders/JSC/blob/6285a066b6debed28211730afb8ffc5eb2430c2f/Custom_EasyBlocks/julia.py to allow the interpreter to pick up the DEPOT/LOAD path directly from the new env variables
- [X] Test how reusable deps really are. The original approach aimed at copying all packages from the deps into the current package and re-install everything. Here we actually re-use them but the results seems to be dependent on the load order (see discussion in #4123) so if the dep order is switched w.r.t. the one at build time slow re-compilation might happen
  - PoC in https://github.com/easybuilders/easybuild-easyconfigs/pull/26214 at-least on package used at the same time should work, more investigation might be needed on loading multiple packages that re-compiles the same deps at the same time

## Notes

This PRs includes the changes from:

- #4122 
- #4124 

Fixes:

- #4123

### AI usage

- GitHub CoPilot FIM auto-completion is active in my IDE and has been used in this PR to sporadically auto-complete code/docstrings

### DIFF
--- a/easybuild/easyblocks/generic/juliabundle.py
+++ b/easybuild/easyblocks/generic/juliabundle.py
@@ -29,6 +29,7 @@ EasyBuild support for bundles of Julia packages, implemented as an easyblock
 """
 import os
 
+from easybuild.tools.build_log import EasyBuildError
 from easybuild.easyblocks.generic.bundle import Bundle
 from easybuild.easyblocks.generic.juliapackage import EXTS_FILTER_JULIA_PACKAGES, JuliaPackage
 
@@ -80,6 +81,9 @@ class JuliaBundle(Bundle, JuliaPackage):
                         'filename': '%(name)s-%(version)s.tar.gz',
                     }
                 ]
+
+        if self.cfg['is_test_dependency']:
+            raise EasyBuildError("Test dependencies can only be defined as an extension, not as a bundle itself")
 
         self.log.info("exts_default_options: %s", self.cfg['exts_default_options'])
 
--- a/easybuild/easyblocks/generic/juliabundle.py
+++ b/easybuild/easyblocks/generic/juliabundle.py
@@ -53,6 +53,12 @@ class JuliaBundle(Bundle, JuliaPackage):
         super().__init__(*args, **kwargs)
 
         self.cfg['exts_defaultclass'] = 'JuliaPackage'
+        # Ensure that Julia packages such as LLVM do not try to use the EB_LLVM easyblock
+        # many packages have names that overlap with EB easyblocks, and would end up using them as default class for extensions, which is not what we want here. We force to always use JuliaPackage unless explicitly specified
+        # otherwise in the easyconfig file.
+        self.cfg['exts_default_options'] = {
+            'easyblock': 'JuliaPackage',
+        }
         self.cfg['exts_filter'] = EXTS_FILTER_JULIA_PACKAGES
 
         # need to disable templating to ensure that actual value for exts_default_options is updated...
--- a/easybuild/easyblocks/generic/juliabundle.py
+++ b/easybuild/easyblocks/generic/juliabundle.py
@@ -79,10 +79,6 @@ class JuliaBundle(Bundle, JuliaPackage):
     def prepare_step(self, *args, **kwargs):
         """Prepare for installing bundle of Julia packages."""
         super().prepare_step(*args, **kwargs)
-
-    def install_step(self):
-        """Prepare installation environment and dd all dependencies to project environment."""
-        self.prepare_julia_env()
         self.include_pkg_dependencies()
 
     def sanity_check_step(self, *args, **kwargs):
--- a/easybuild/easyblocks/generic/juliabundle.py
+++ b/easybuild/easyblocks/generic/juliabundle.py
@@ -83,11 +83,6 @@ class JuliaBundle(Bundle, JuliaPackage):
 
         self.log.info("exts_default_options: %s", self.cfg['exts_default_options'])
 
-    def prepare_step(self, *args, **kwargs):
-        """Prepare for installing bundle of Julia packages."""
-        super().prepare_step(*args, **kwargs)
-        self.include_pkg_dependencies()
-
     def sanity_check_step(self, *args, **kwargs):
         """Custom sanity check for bundle of Julia packages"""
         custom_paths = {
--- a/easybuild/easyblocks/generic/juliabundle.py
+++ b/easybuild/easyblocks/generic/juliabundle.py
@@ -54,7 +54,8 @@ class JuliaBundle(Bundle, JuliaPackage):
 
         self.cfg['exts_defaultclass'] = 'JuliaPackage'
         # Ensure that Julia packages such as LLVM do not try to use the EB_LLVM easyblock
-        # many packages have names that overlap with EB easyblocks, and would end up using them as default class for extensions, which is not what we want here. We force to always use JuliaPackage unless explicitly specified
+        # many packages have names that overlap with EB easyblocks, and would end up using them as default class for
+        # extensions, which is not what we want here. We force to always use JuliaPackage unless explicitly specified
         # otherwise in the easyconfig file.
         self.cfg['exts_default_options'] = {
             'easyblock': 'JuliaPackage',
--- a/easybuild/easyblocks/generic/juliabundle.py
+++ b/easybuild/easyblocks/generic/juliabundle.py
@@ -26,6 +26,7 @@
 EasyBuild support for bundles of Julia packages, implemented as an easyblock
 
 @author: Alex Domingo (Vrije Universiteit Brussel)
+@author: Davide Grassano (CECAM, EPFL)
 """
 import os
 
--- a/easybuild/easyblocks/generic/juliabundle.py
+++ b/easybuild/easyblocks/generic/juliabundle.py
@@ -28,8 +28,6 @@ EasyBuild support for bundles of Julia packages, implemented as an easyblock
 @author: Alex Domingo (Vrije Universiteit Brussel)
 @author: Davide Grassano (CECAM, EPFL)
 """
-import os
-
 from easybuild.easyblocks.generic.bundle import Bundle
 from easybuild.easyblocks.generic.juliapackage import JuliaPackage
 
--- a/easybuild/easyblocks/generic/juliabundle.py
+++ b/easybuild/easyblocks/generic/juliabundle.py
@@ -31,7 +31,7 @@ import os
 
 from easybuild.tools.build_log import EasyBuildError
 from easybuild.easyblocks.generic.bundle import Bundle
-from easybuild.easyblocks.generic.juliapackage import EXTS_FILTER_JULIA_PACKAGES, JuliaPackage
+from easybuild.easyblocks.generic.juliapackage import JuliaPackage
 
 
 class JuliaBundle(Bundle, JuliaPackage):
@@ -61,7 +61,6 @@ class JuliaBundle(Bundle, JuliaPackage):
         self.cfg['exts_default_options'] = {
             'easyblock': 'JuliaPackage',
         }
-        self.cfg['exts_filter'] = EXTS_FILTER_JULIA_PACKAGES
 
         # need to disable templating to ensure that actual value for exts_default_options is updated...
         with self.cfg.disable_templating():
@@ -94,7 +93,3 @@ class JuliaBundle(Bundle, JuliaPackage):
             'dirs': [os.path.join('packages', self.name)],
         }
         super().sanity_check_step(custom_paths=custom_paths)
-
-    def make_module_extra(self, *args, **kwargs):
-        """Custom module environment from JuliaPackage"""
-        return super().make_module_extra(*args, **kwargs)
--- a/easybuild/easyblocks/generic/juliabundle.py
+++ b/easybuild/easyblocks/generic/juliabundle.py
@@ -81,12 +81,7 @@ class JuliaBundle(Bundle, JuliaPackage):
                     }
                 ]
 
-        self.log.info("exts_default_options: %s", self.cfg['exts_default_options'])
+        # The name of the bundle can be arbitrary and not necessarily a Julia package
+        self.cfg['exts_filter'] = None
 
-    def sanity_check_step(self, *args, **kwargs):
-        """Custom sanity check for bundle of Julia packages"""
-        custom_paths = {
-            'files': [],
-            'dirs': [os.path.join('packages', self.name)],
-        }
-        super().sanity_check_step(custom_paths=custom_paths)
+        self.log.info("exts_default_options: %s", self.cfg['exts_default_options'])
--- a/easybuild/easyblocks/generic/juliabundle.py
+++ b/easybuild/easyblocks/generic/juliabundle.py
@@ -30,7 +30,6 @@ EasyBuild support for bundles of Julia packages, implemented as an easyblock
 """
 import os
 
-from easybuild.tools.build_log import EasyBuildError
 from easybuild.easyblocks.generic.bundle import Bundle
 from easybuild.easyblocks.generic.juliapackage import JuliaPackage
 
@@ -81,9 +80,6 @@ class JuliaBundle(Bundle, JuliaPackage):
                         'filename': '%(name)s-%(version)s.tar.gz',
                     }
                 ]
-
-        if self.cfg['is_test_dependency']:
-            raise EasyBuildError("Test dependencies can only be defined as an extension, not as a bundle itself")
 
         self.log.info("exts_default_options: %s", self.cfg['exts_default_options'])
 
--- a/easybuild/easyblocks/generic/juliapackage.py
+++ b/easybuild/easyblocks/generic/juliapackage.py
@@ -33,6 +33,8 @@ import os
 import re
 import tempfile
 
+from typing import List, Dict, Tuple, Union
+
 from easybuild.tools import LooseVersion
 
 import easybuild.tools.environment as env
@@ -170,50 +172,64 @@ class JuliaPackage(ExtensionEasyBlock):
         return self._julia_version
 
     @property
-    def conflicts(self):
-        """List of conflicting software."""
+    def conflicts(self) -> List[Tuple[str, str, str, str]]:
+        """List of 4-tuple conflicting software (same package included from multiple locations) including:
+        - extension name where conflict is found
+        - package name with conflict
+        - previous source of the package that caused the conflict
+        - new source of the package that caused the conflict
+        """
         if self.is_extension:
             return self.master.conflicts
         return self._conflicts
 
     @property
-    def is_last_extension(self):
-        """Whether this extension is the last one to be installed in the installation."""
+    def is_last_extension(self) -> bool:
+        """Whether this extension is the last one to be installed."""
         if not self.is_extension:
             return True
 
         return self.master.ext_instances and self.master.ext_instances[-1] is self
 
     @property
-    def julia_deps(self):
-        """List of Julia dependencies found in this installation excluding test dependencies."""
+    def julia_deps(self) -> Dict[str, str]:
+        """List of Julia dependencies found in this installation excluding test dependencies
+        - key: package name
+        - value: package source
+        """
         if self.is_extension:
             return self.master.julia_deps
         return self._julia_deps
 
     @property
-    def julia_deps_test(self):
-        """List of Julia dependencies found in this installation including test dependencies."""
+    def julia_deps_test(self) -> Dict[str, str]:
+        """List of Julia dependencies found in this installation including test dependencies
+        - key: package name
+        - value: package source
+        """
         if self.is_extension:
             return self.master.julia_deps_test
         return self._julia_deps_test
 
     @property
-    def pkg_deps(self):
-        """List of easybuild runtime dependencies that will need to be sanity-checked."""
+    def pkg_deps(self) -> List[Tuple[str, str]]:
+        """List of easybuild runtime dependencies that will need to be sanity-checked. 2-tuple including:
+        - dependency name
+        - dependency root (equivalent to get_software_root)
+        """
         if self.is_extension:
             return self.master.pkg_deps
         return self._pkg_deps
 
     @property
-    def pkgs_to_test(self):
-        """List extension with `runtest` set to true that should be tested."""
+    def pkgs_to_test(self) -> List[str]:
+        """List extension names with `runtest` set to True that should be tested."""
         if self.is_extension:
             return self.master.pkgs_to_test
         return self._pkgs_to_test
 
     @property
-    def tmp_test_dir(self):
+    def tmp_test_dir(self) -> str:
         """Temporary path used for running the test step."""
         if self.is_extension:
             return self.master.tmp_test_dir
@@ -225,9 +241,14 @@ class JuliaPackage(ExtensionEasyBlock):
                 raise EasyBuildError("Failed to create temporary directory for Julia package testing: %s", str(e))
         return self._tmp_test_dir
 
-    def julia_env_path(self, absolute=True, base=True, basedir=None):
+    def julia_env_path(self, absolute: bool = True, base: bool = True, basedir: Union[str, None] = None) -> str:
         """
         Return path to installation environment file.
+
+        :param absolute: whether to return absolute path to environment file or relative path to `basedir`
+        :param base: whether to return path to environment file or its parent directory
+        :param basedir: base directory for the environment, as in env=BASEDIR/environments/v#.#/Project.toml.
+        Defaults to installation directory of this package if not specified.
         """
         basedir = basedir or self.installdir
         julia_version = self.julia_version.split('.')
@@ -241,7 +262,7 @@ class JuliaPackage(ExtensionEasyBlock):
 
         return project_env
 
-    def set_pkg_remote(self, online=False):
+    def set_pkg_remote(self, online: bool = False):
         """Enable online/offline mode of Julia Pkg"""
 
         julia_version = get_software_version('Julia')
@@ -257,7 +278,7 @@ class JuliaPackage(ExtensionEasyBlock):
             )
             raise EasyBuildError(errmsg, julia_version)
 
-    def prepare_julia_env(self, basedir=None, online=False):
+    def prepare_julia_env(self, basedir: Union[str, None] = None, online: bool = False):
         """
         1. Remove user depot and prepend installation directory to DEPOT_PATH.
         Top directory in Julia DEPOT_PATH is the target installation directory.
@@ -272,6 +293,12 @@ class JuliaPackage(ExtensionEasyBlock):
         3. Enable offline mode in Julia to avoid automatic downloads of packages.
 
         4. Enable automatic precompilation of packages after each build.
+
+        5. Enable/disable debug mode in Julia to get more verbose output during installation and testing.
+
+        :param basedir: base directory for the environment, as in env=BASEDIR/environments/v#.#/Project.toml.
+        Defaults to installation directory of this package if not specified.
+        :param online: whether to allow online operations of Julia Pkg
         """
         basedir = basedir or self.installdir
         # Grab both DEPOT_PATH and LOAD_PATH before any changes are made
@@ -304,8 +331,7 @@ class JuliaPackage(ExtensionEasyBlock):
         # Enable/disable offline mode
         self.set_pkg_remote(online=online)
 
-        if self.cfg['julia_debug']:
-            env.setvar('JULIA_DEBUG', 'all')
+        env.setvar('JULIA_DEBUG', 'all' if self.cfg['julia_debug'] else '')
 
         # Set the maximum number of concurrent package builds
         env.setvar('JULIA_NUM_PRECOMPILE_TASKS', str(self.cfg.parallel))
@@ -313,6 +339,12 @@ class JuliaPackage(ExtensionEasyBlock):
         env.setvar('JULIA_PKG_PRECOMPILE_AUTO', 'true')
 
     def add_package(self, pkg_source, test_only=False):
+        """Add a Julia package to the list of dependencies to be installed.
+
+        :param pkg_source: path to the package source to be added as dependency
+        :param test_only: whether this package is only needed for testing and should not be added to installation
+        environment
+        """
         pkg_name = os.path.basename(pkg_source)
         if pkg_name in self.julia_deps:
             prev_source = self.julia_deps[pkg_name]
@@ -334,8 +366,12 @@ class JuliaPackage(ExtensionEasyBlock):
                 )
             raise EasyBuildError("Conflicts detected in Julia dependencies:\n" + "\n".join(conflict_msgs))
 
-    def install_pkg(self, basedir=None):
-        """Execute Julia.Pkg command to install package from its sources"""
+    def install_pkg(self, basedir: Union[str, None] = None):
+        """Execute Julia.Pkg command to install all packages in `self.julia_deps` in the install environment.
+
+        :param basedir: base directory for the environment, as in env=BASEDIR/environments/v#.#/Project.toml.
+        Defaults to installation directory of this package if not specified
+        """
         self._check_conflicts()
         basedir = basedir or self.installdir
         env = self.julia_env_path(basedir=basedir)
@@ -373,8 +409,11 @@ class JuliaPackage(ExtensionEasyBlock):
 
         return res.output
 
-    def _deps_from_project(self, environment):
-        """Get list of dependencies from a Project.toml file"""
+    def _deps_from_project(self, environment: str) -> List[str]:
+        """Get list of dependencies from a Julia environment
+
+        :param environment: path to Julia environment to query for dependencies
+        """
         julia_root = get_software_root('Julia')
         cmd = "; ".join([
             'using Pkg',
@@ -454,8 +493,12 @@ class JuliaPackage(ExtensionEasyBlock):
         move_file(self.start_dir, package_dir)
         self.add_package(package_dir, test_only=self.cfg['is_test_dependency'])
 
-    def _build_install_step(self, basedir=None):
-        """Install step commons between single-package and extensions based installs."""
+    def _build_install_step(self, basedir: Union[str, None] = None):
+        """Install step commons between single-package and extensions based installs.
+
+        :param basedir: base directory for the environment, as in env=BASEDIR/environments/v#.#/Project.toml.
+        Defaults to installation directory of this package if not specified.
+        """
         self.install_source()
         if self.is_last_extension:
             self.prepare_julia_env(basedir=basedir, online=self.cfg['download_pkg_deps'])
@@ -579,13 +622,9 @@ class JuliaPackage(ExtensionEasyBlock):
 
     def make_module_extra(self, *args, **kwargs):
         """
-        Module load initializes JULIA_DEPOT_PATH and JULIA_LOAD_PATH with default values if they are not set.
-
-        Path to installation directory is appended to JULIA_DEPOT_PATH.
-        Path to the environment file of this installation is prepended to JULIA_LOAD_PATH.
-        This configuration fulfils the rule that user depot has to be the first path in JULIA_DEPOT_PATH,
-        allowing user to add custom Julia packages while having packages in this installation available.
-        See issue easybuilders/easybuild-easyconfigs#17455
+        Set EB-specific PATH like variables `EB_JULIA_DEPOT_PATH` and `EB_JULIA_LOAD_PATH`.
+        These are then used by Easybuild's custom Julia startup script (etc/julia/startup.jl) to set JULIA_DEPOT_PATH
+        and JULIA_LOAD_PATH correctly at runtime.
         """
         mod = super().make_module_extra()
 
--- a/easybuild/easyblocks/generic/juliapackage.py
+++ b/easybuild/easyblocks/generic/juliapackage.py
@@ -48,7 +48,7 @@ from easybuild.tools.run import run_shell_cmd
 from easybuild.tools.utilities import trace_msg, print_msg
 from easybuild.tools.hooks import BUILD_STEP, TEST_STEP
 
-from easybuild.easyblocks.j.julia import EB_JULIA_DEPOT_PATH_VAR, EB_JULIA_LOAD_PATH_VAR
+from easybuild.easyblocks.julia import EB_JULIA_DEPOT_PATH_VAR, EB_JULIA_LOAD_PATH_VAR
 
 _COMPILECACHE_CHECK = ' | '.join([
     # "julia -E 'Base.compilecache_path(Base.identify_package(\"%(ext_name)s\"), Base.get_world_counter())'",
--- a/easybuild/easyblocks/generic/juliapackage.py
+++ b/easybuild/easyblocks/generic/juliapackage.py
@@ -207,7 +207,7 @@ class JuliaPackage(ExtensionEasyBlock):
 
     @property
     def pkgs_to_test(self):
-        """List extension wiht `runtest` set to true that should be tested."""
+        """List extension with `runtest` set to true that should be tested."""
         if self.is_extension:
             return self.master.pkgs_to_test
         return self._pkgs_to_test
--- a/easybuild/easyblocks/generic/juliapackage.py
+++ b/easybuild/easyblocks/generic/juliapackage.py
@@ -81,8 +81,8 @@ setenv("JULIA_LOAD_PATH", os.getenv("EBJULIA_LOAD_PATH") .. ":")
     # This needs testing, the module generate seems to work fine, but the loading of the module within EB does not
     # properly resolve the $::env() and leaves it as a string
     "Tcl": """
-setenv JULIA_DEPOT_PATH ":\$::env(EBJULIA_DEPOT_PATH)"
-setenv JULIA_LOAD_PATH "\$::env(EBJULIA_LOAD_PATH):"
+setenv JULIA_DEPOT_PATH ":$::env(EBJULIA_DEPOT_PATH)"
+setenv JULIA_LOAD_PATH "$::env(EBJULIA_LOAD_PATH):"
 """,
 }
 
@@ -262,7 +262,7 @@ class JuliaPackage(ExtensionEasyBlock):
             pkg_dir = os.path.dirname(pkg_path)
 
         if not os.path.isfile(pkg_path):
-            raise EasyBuildError("Project.toml file not found for package %s in path: %s", pkg_name, pkg_dir)
+            raise EasyBuildError("Project.toml file not found in path: %s", pkg_dir)
 
         project_toml = cls.read_project_toml(pkg_path)
         pkg_name = project_toml['name']
@@ -554,7 +554,10 @@ class JuliaPackage(ExtensionEasyBlock):
         # mod += self.module_generator.append_paths('JULIA_DEPOT_PATH', [''])
 
         mod += self.module_generator.prepend_paths('EBJULIA_DEPOT_PATH', [''])
-        mod += self.module_generator.prepend_paths('EBJULIA_LOAD_PATH', [self.julia_env_path(absolute=False, base=False)])
+        mod += self.module_generator.prepend_paths(
+            'EBJULIA_LOAD_PATH',
+            [self.julia_env_path(absolute=False, base=False)]
+        )
 
         return mod
 
--- a/easybuild/easyblocks/generic/juliapackage.py
+++ b/easybuild/easyblocks/generic/juliapackage.py
@@ -39,17 +39,19 @@ import easybuild.tools.environment as env
 from easybuild.framework.easyconfig import CUSTOM
 from easybuild.framework.extensioneasyblock import ExtensionEasyBlock
 from easybuild.tools.build_log import EasyBuildError
+from easybuild.tools.config import build_option
 from easybuild.tools.modules import get_software_root, get_software_version
 from easybuild.tools.filetools import move_file
 from easybuild.tools.run import run_shell_cmd
-from easybuild.tools.utilities import trace_msg
+from easybuild.tools.utilities import trace_msg, print_msg
+from easybuild.tools.hooks import BUILD_STEP, TEST_STEP
 
 from easybuild.easyblocks.j.julia import EB_JULIA_DEPOT_PATH_VAR, EB_JULIA_LOAD_PATH_VAR
 
 _COMPILECACHE_CHECK = ' | '.join([
     # "julia -E 'Base.compilecache_path(Base.identify_package(\"%(ext_name)s\"), Base.get_world_counter())'",
     # compilecache_path is not part of the stable API and changes arguments across versions
-    # allow internal cache resolution and use debug statemnts to get the path of the loaded cache
+    # allow internal cache resolution and use debug statements to get the path of the loaded cache
     "JULIA_DEBUG=loading julia -e 'using %(ext_name)s' 2>&1 1>/dev/null",
     "grep -E 'Loading (object )?cache file .*%(grep_loc)s'"
 ])
@@ -149,6 +151,7 @@ class JuliaPackage(ExtensionEasyBlock):
         self._pkg_deps = []
         self._pkg_to_test = []
         self._tmp_test_dir = None
+        self._installdir_created = False
 
     @property
     def julia_version(self) -> str:
@@ -177,7 +180,7 @@ class JuliaPackage(ExtensionEasyBlock):
     def is_last_extension(self):
         """Whether this extension is the last one to be installed in the installation."""
         if not self.is_extension:
-            return False
+            return True
 
         return self.master.ext_instances and self.master.ext_instances[-1] is self
 
@@ -395,15 +398,14 @@ class JuliaPackage(ExtensionEasyBlock):
         build_only_deps = self.cfg.dependencies(build_only=True)
         # add packages found in dependencies to this installation environment
         for dep in self.cfg.dependencies():
-            test_only = dep in build_only_deps
             dep_name = dep['name']
             dep_root = get_software_root(dep_name)
             dep_env = self.julia_env_path(absolute=True, base=True, basedir=dep_root)
-            dep_manifest = os.path.join(dep_env, 'Manifest.toml')
-            if not os.path.isfile(dep_manifest):
-                self.log.warning("No Manifest.toml file found in dependency %s, skipping: %s", dep_name, dep_env)
+            if not os.path.isdir(dep_env):
+                self.log.warning("No environment directory found in dependency %s, skipping: %s", dep_name, dep_env)
                 continue
 
+            test_only = dep in build_only_deps
             if test_only:
                 trace_msg(
                     f"Incorporating Julia package in TEST    environment {dep_name}: {dep_env}"
@@ -415,11 +417,19 @@ class JuliaPackage(ExtensionEasyBlock):
                 )
 
             for pkg_path in self._deps_from_project(dep_env):
-                self.add_package(pkg_path, test_only=dep in build_only_deps)
+                self.add_package(pkg_path, test_only=test_only)
+
+    def make_installdir(self):
+        """Create new installation directory and ensure this is done only once"""
+        # Stop installdir from being re-created at the install step
+        if not self._installdir_created:
+            super().make_installdir()
+            self._installdir_created = True
 
     def prepare_step(self, *args, **kwargs):
         """Prepare for Julia package installation."""
         super().prepare_step(*args, **kwargs)
+        self.make_installdir()
 
         if get_software_root('Julia') is None:
             raise EasyBuildError("Julia not included as dependency!")
@@ -430,9 +440,32 @@ class JuliaPackage(ExtensionEasyBlock):
         """Configure step for Julia packages is a no-op, as there is no configuration needed before installation."""
         pass
 
+    def install_source(self):
+        """Add the Julia package source files in the installation directory."""
+        if self.cfg['is_test_dependency']:
+            self.log.debug(
+                f"Package {self.name} is only needed for testing, installing sources in {self.tmp_test_dir}"
+            )
+            trace_msg("Installing as a test dependency...")
+            package_dir = os.path.join(self.tmp_test_dir, 'packages', self.name)
+        else:
+            package_dir = os.path.join(self.installdir, 'packages', self.name)
+
+        move_file(self.start_dir, package_dir)
+        self.add_package(package_dir, test_only=self.cfg['is_test_dependency'])
+
+    def _build_install_step(self, basedir=None):
+        """Install step commons between single-package and extensions based installs."""
+        self.install_source()
+        if self.is_last_extension:
+            self.prepare_julia_env(basedir=basedir, online=self.cfg['download_pkg_deps'])
+            return self.install_pkg(basedir=basedir)
+        else:
+            trace_msg("Delegating build+install to last extension")
+
     def build_step(self):
         """Build step for Julia packages is a no-op, as there is no build needed before installation."""
-        pass
+        return self._build_install_step()
 
     def test_step(self):
         """
@@ -440,6 +473,13 @@ class JuliaPackage(ExtensionEasyBlock):
 
         :param return_output: return output and exit code of test command
         """
+        if self.cfg['runtest']:
+            self.pkg_to_test.append(self.name)
+
+        if not self.is_last_extension:
+            trace_msg("Delegating testing to last extension")
+            return
+
         if self.pkg_to_test:
             env = self.julia_env_path(basedir=self.tmp_test_dir)
             pkg_specs = ', '.join(f'PackageSpec(path="{path}")' for path in self.julia_deps_test.values())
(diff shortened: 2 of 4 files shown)


Crivella:
Moved the final setting of `LOAD_PATH` and `DEPOT_PATH` to Julia's site-specific startup script.
This resolves:
- having to write custom module code
- the *reversing setenv* problem 
- Lua-vs-Tcl syntax

easybuild/easyblocks/generic/juliapackage.py
```diff
@@ -272,33 +425,100 @@ def build_step(self):
...
+                self.cfg['testopts'],
+            ])
+
+            run_shell_cmd(cmd)
 
-        self.prepare_julia_env()
-        self.include_pkg_dependencies()
+    def install_source(self):
+        """Add the Julia package source files in the installation directory."""
+        if self.cfg['is_test_dependency']:
```
Flamefire: This looks odd. How does this work? It (may) designate a whole easyconfig as test-only making the files missing from the installed module.

We should avoid:
a) That it is set on anything but a (bundle-)extension  
b) That it turns up in `extensions` of the module.

Crivella: Yeah should probably add more checks around that.
Right now i am considering what i want to do with it, and if i should shift this from a flag to single extensions to just treating build-deps as *test-only* dependencies.

The main thing that i am focusing on now is how feasible it is is re-using deps. It is working but is fairly sensitive. If 2 ECs ships the same Julia package, it is possible that the compile cache will be broken based on load order.
The check i've added right now only works if the 2 ECs are being used as a deps of another.
The other point is that if we go this route every Bundle should ideally define only 1 exts (which begs the question if bundles are needed at all), otherwise we go back to the old intended (but broken way) of just using deps as a fancy way to tell Julia which packages to re-install but not really use the deps themselves (still think that solution might not really allow to load multiple packages at the same time)

easybuild/easyblocks/generic/juliapackage.py
```diff
@@ -180,6 +172,13 @@ def pkg_deps(self):
             return self.master.pkg_deps
         return self._pkg_deps
 
+    @property
+    def pkg_to_test(self):
```
Flamefire: I'd suggest to name it such that it is clear it returns a list and/or add type hints:

    def pkgs_for_test(self) -> List[str]:

easybuild/easyblocks/generic/juliabundle.py
```diff
@@ -81,6 +82,9 @@ def __init__(self, *args, **kwargs):
                     }
                 ]
 
+        if self.cfg['is_test_dependency']:
```
Flamefire: Wouldn't it work to check this in the `__init__` of `JuliaPackage` guarded by `self.is_extension`?
That's where the property is defined and used.

easybuild/easyblocks/generic/juliapackage.py
Crivella: Yeah forgot the plural

I am really waiting for us to drop support for Python <3.9 to start typehinting everything so we can use the base `list` instead of having to use the uppercase types from typehint :sweat_smile:

easybuild/easyblocks/generic/juliabundle.py
Crivella: It is already in the init, not sure if the extra guard is needed as the extensions should be initialized with `JuliaPackage` while only the top-level EC uses the init of `JuliaBundle`

easybuild/easyblocks/generic/juliapackage.py
Flamefire: Dito. But let's start using those now to make our lifes easier, the transition is trivial.

BTW: The s/to/for was deliberate: "package to test" sounds like this package should be tested while the description says it is required (only) for tests

easybuild/easyblocks/generic/juliabundle.py
Flamefire: JuliaBundle calls `__init__` of JuliaPackage, doesn't it? So moving it would work here, and additionally in every other easyblock deriving from JuliaPackage so it cannot be missed.

Crivella: Ah see what you meant, yeah not against it

easybuild/easyblocks/generic/juliapackage.py
Crivella: Ah MB there was a copy-paste left over in the description, will fix it tomorrow, that list flags all the packages with `runtest` and than launches all the tests at once at the end

easybuild/easyblocks/j/julia.py
```diff
@@ -0,0 +1,103 @@
...
+    pushfirst!(LOAD_PATH, EB_LOAD_PATH...)
+end
+# Prepend EB_DEPOT_PATH to the default DEPOT_PATH
+if !isempty(EB_DEPOT_PATH)
+    push!(DEPOT_PATH, EB_DEPOT_PATH...)
+end
+"""
+
+WRAPPER_CONTENT = r"""#!/bin/sh
+LD_LIBRARY_PATH="$EBROOTJULIA/lib:$EBROOTJULIA/lib/julia:$LD_LIBRARY_PATH" julia.bin "$@"
```
Flamefire, line 60: Why can this not be done in the module? Or is this only for rpath?

easybuild/easyblocks/generic/juliapackage.py
```diff
@@ -106,51 +128,136 @@ def get_julia_env(env_var):
...
+    @property
+    def pkg_deps(self):
+        """List of easybuild runtime dependencies that will need to be sanity-checked."""
+        if self.is_extension:
+            return self.master.pkg_deps
+        return self._pkg_deps
+
+    @property
+    def pkgs_to_test(self):
+        """List extension wiht `runtest` set to true that should be tested."""
```
Flamefire: Typo `wiht`, but sounds like there could be extension where runtest=True but should not be tested.
If that is not the case, maybe:
```suggestion
        """List extension that should be tested (`runtest` set to true)."""
```

BTW: It is never set

easybuild/easyblocks/j/julia.py
Crivella: This is in the original wrapper that is currently installed at the EC level. I actually encountered what this is safeguarding against while not including it: Julia does not RPATH against its own libraries (also they are not directly linked but dlopened when needed as far as i understand).
If by loading modules you end up having stuff like `libLLVM.so` or `libunwind.so` etc come before Julia's own dirs, it will actually try to use them and most likely crash. This guard ensure that for Julia, its own libraries are always preferred.
I guess it is putting an hard block to power-users that actually want to try and mix those libraries, so we might want to add an env variable that can turn this off...

easybuild/easyblocks/generic/juliabundle.py
Crivella: [482c8fb](https://github.com/easybuilders/easybuild-easyblocks/pull/4146/commits/482c8fb2da2e0d93e254c62e056ecabad690208f)

Crivella:
Concerning the points of "re-usability" of deps I am unsure what is the best way.
It is possible to do so, but if we wanted to do things properly, every bundle should never define more than 1 package (or 1 package + its optional subpackages if those are never used independently).

The reason being that if 2 packages define the same extension, the compile cache will could work or be broken depending on the load order of the packages from what i've seen (this is why i check for conflicts when rebuilding the manifest).
I think this mainly happens if one package could use the other as an optional dependency, eg: I am working on a package that depends on `JLD2` and `Flux` as deps.
`Flux` also has `JLD2` as an optional dependency. With this approach if `JLD2` is not built into `Flux` (what i am doing right now) if `JLD2` is loaded first everything works, but if not i got into an infinite compilation loop which exited after some max-try limit and ended up working in what i assume is an interpreted-only mode


The old attempted route of just using the dependencies just as a source of information for the packages to install (but build a completely new and isolated environment) also has downsides:

- Bundle should still be "1-package" only or anything using it as a dependency will also bring in potentially un-needed extra packages
- I think weird think could still happen also in this scenario with compile caches depending on load order

EDIT: added a PR for what i have concerning the aforementioned packages

- https://github.com/easybuilders/easybuild-easyconfigs/pull/26214

Pinging also @lexming since he was the original author of the easyblock to get feedback on these points and the new EB in general

Flamefire:
> every bundle should never define more than 1 package 

Why would we need a bundle then? Isn't this effectively saying: Use `JuliaPackage`?

Crivella:
> Why would we need a bundle then? Isn't this effectively saying: Use JuliaPackage?

There are some packages that have optional features that are enabled by installing a subpackage. In that case i expected them to always be bundable without problems.

Indeed i think the proper approach would be to only have Package, even in Python we sometime encounter trouble when extensions of one package shadow one of another which depending on version might or might not work

Flamefire:
> Indeed i think the proper approach would be to only have Package, even in Python we sometime encounter trouble when extensions of one package shadow one of another which depending on version might or might not work

I mentioned something like that in Slack recently as it seems like PythonBundle is/should be preferred but IMO it makes it too easy to add additional packages that might silently conflict with others. The suggestion was to enhance the CI checks to catch that.
So might be worth to discuss this in a larger group

ocaisa:
@lexming Your feedback on this would be very valuable

easybuild/easyblocks/generic/juliabundle.py
```diff
@@ -74,25 +79,7 @@ def __init__(self, *args, **kwargs):
                     }
                 ]
 
-        self.log.info("exts_default_options: %s", self.cfg['exts_default_options'])
-
-    def prepare_step(self, *args, **kwargs):
-        """Prepare for installing bundle of Julia packages."""
-        super().prepare_step(*args, **kwargs)
+        # The name of the bundle can be arbitrary and not necessarily a Julia package
+        self.cfg['exts_filter'] = None
```
Flamefire, line 83: I don't understand the comment: Why is the bundle name relevant and what is not a Julia package?
And `exts_filter` refers to extensions which (usually) are Julia packages as indicated by the name and description of this easyblock: "Bundle of JuliaPackages"

And `EXTS_FILTER_JULIA_PACKAGES` worked so far didn't it? 
What happens now? Doesn't this change mean that no extensions of JuliaBundles will be sanity-checked (via `julia -e 'using ...`) anymore at all? And you cannot even set it in the easyconfig of the bundle as it will be overwritten here.

Crivella: The extensions are still checked individually (even more than before since we also check that we are actually using the compile cache from EB and not rebuilding them in the user''s home depot)

https://github.com/Crivella/easybuild-easyblocks/blob/e100f571271aa2ed9ddad1c3e66b14fc3f37f164/easybuild/easyblocks/generic/juliapackage.py#L623

If i recall correctly this was to avoid using the name of the bundle itself with `julia -e BundleName` to avoid being forced to name the EC as the specific Julia package.

Flamefire: But `exts_filter` is evaluated before handling of extensions, i.e. in the Bundle. So removing it from the Bundle means it won't be handled at all and https://github.com/Crivella/easybuild-easyblocks/blob/e100f571271aa2ed9ddad1c3e66b14fc3f37f164/easybuild/easyblocks/generic/juliapackage.py#L657C23-L657C34 is actually wrong: `exts_filter` for an extension is ignored unless that has extensions itself, which IMO it cannot.

With this removal e.g. `--skip` won't work anymore.

> If i recall correctly this was to avoid using the name of the bundle itself with julia -e BundleName to avoid being forced to name the EC as the specific Julia package.

I don't see this happening as it isn't used for the main-Easyconfig, is it?

easybuild/easyblocks/generic/juliapackage.py
```diff
@@ -231,96 +421,269 @@ def install_pkg_source(self, pkg_source, environment, trace=True):
...
+
+                    )
+            kwargs.setdefault('custom_commands', []).extend(custom_commands)
+
+            custom_paths = {
+                'files': [],
+                'dirs': [pkg_dir],
+            }
+
+            exts_filter = " && ".join(exts_filter)
```
Flamefire, line 653: It seems odd that it is set outside the if, but only used in one branch.
Besides that this will likely be fully ignored. Maybe you meant to add this as a sanity check command only?
And why isn't it done for non-extensions, i.e. plain JuliaPackages?

easybuild/easyblocks/generic/juliabundle.py
Crivella: I do not understand what you mean.
`exts_filter` is used in the sanity check step of the extensions

https://github.com/easybuilders/easybuild-framework/blob/ccbc8ea852afb2a1ec83988eac07e7e9caa40212/easybuild/framework/extension.py#L336


In the details an example of a sanity check on top of https://github.com/easybuilders/easybuild-easyconfigs/pull/26214
<details><summary>Details</summary>
<p>

```
(easybuild-dev) crivella@crivella-desktop:~$ eb --from-pr 26214 --sanity-check-only Flux-0.16.10-Julia-1.12.2.eb
== Temporary log file in case of crash /home/crivella/.local/easybuild/logs/easybuild-hjdlyp2s.log

WARNING: Index for /home/crivella/Documents/GIT/easybuild-easyconfigs/easybuild/easyconfigs is no longer valid (too old), so ignoring it...

== processing EasyBuild easyconfig /tmp/eb-xufj5it_/files_pr26214/f/Flux/Flux-0.16.10-Julia-1.12.2.eb
== building and installing Flux/0.16.10-Julia-1.12.2...
  >> installation prefix: /home/crivella/.local/easybuild/software/Flux/0.16.10-Julia-1.12.2
== fetching files and verifying checksums [skipped]
== creating build dir, resetting environment [skipped]
== unpacking [skipped]
== patching [skipped]
== preparing [skipped]
== configuring [skipped]
== building [skipped]
== testing [skipped]
== installing [skipped]
== taking care of extensions [skipped]
== restore after iterating [skipped]
== postprocessing [skipped]
== sanity checking...
  >> (non-empty) directory 'packages' found: OK
  >> (non-empty) directory 'environments' found: OK
  >> loading modules: Flux/0.16.10-Julia-1.12.2...
  >> Extension sanity check command 'JULIA_DEBUG=loading julia -e 'using ADTypes' 2>&1 1>/dev/null | grep -E 'Loading (object )?cache file .*ADTypes' && julia -e 'using ADTypes'': OK
  >> (non-empty) directory 'packages/ADTypes' found: OK
  >> Extension sanity check command 'JULIA_DEBUG=loading julia -e 'using ArgCheck' 2>&1 1>/dev/null | grep -E 'Loading (object )?cache file .*ArgCheck' && julia -e 'using ArgCheck'': OK
  >> (non-empty) directory 'packages/ArgCheck' found: OK
  >> Extension sanity check command 'JULIA_DEBUG=loading julia -e 'using Baselet' 2>&1 1>/dev/null | grep -E 'Loading (object )?cache file .*Baselet' && julia -e 'using Baselet'': OK
  >> (non-empty) directory 'packages/Baselet' found: OK
  >> Extension sanity check command 'JULIA_DEBUG=loading julia -e 'using CompositionsBase' 2>&1 1>/dev/null | grep -E 'Loading (object )?cache file .*CompositionsBase' && julia -e 'using CompositionsBase'': OK
  >> (non-empty) directory 'packages/CompositionsBase' found: OK
  >> Extension sanity check command 'JULIA_DEBUG=loading julia -e 'using ConstructionBase' 2>&1 1>/dev/null | grep -E 'Loading (object )?cache file .*ConstructionBase' && julia -e 'using ConstructionBase'': OK
  >> (non-empty) directory 'packages/ConstructionBase' found: OK
  >> Extension sanity check command 'JULIA_DEBUG=loading julia -e 'using DataAPI' 2>&1 1>/dev/null | grep -E 'Loading (object )?cache file .*DataAPI' && julia -e 'using DataAPI'': OK
  >> (non-empty) directory 'packages/DataAPI' found: OK
  >> Extension sanity check command 'JULIA_DEBUG=loading julia -e 'using DataValueInterfaces' 2>&1 1>/dev/null | grep -E 'Loading (object )?cache file .*DataValueInterfaces' && julia -e 'using DataValueInterfaces'': OK
  >> (non-empty) directory 'packages/DataValueInterfaces' found: OK
  >> Extension sanity check command 'JULIA_DEBUG=loading julia -e 'using DefineSingletons' 2>&1 1>/dev/null | grep -E 'Loading (object )?cache file .*DefineSingletons' && julia -e 'using DefineSingletons'': OK
  >> (non-empty) directory 'packages/DefineSingletons' found: OK
  >> Extension sanity check command 'JULIA_DEBUG=loading julia -e 'using DocStringExtensions' 2>&1 1>/dev/null | grep -E 'Loading (object )?cache file .*DocStringExtensions' && julia -e 'using DocStringExtensions'': OK
  >> (non-empty) directory 'packages/DocStringExtensions' found: OK
  >> Extension sanity check command 'JULIA_DEBUG=loading julia -e 'using EnzymeCore' 2>&1 1>/dev/null | grep -E 'Loading (object )?cache file .*EnzymeCore' && julia -e 'using EnzymeCore'': OK
  >> (non-empty) directory 'packages/EnzymeCore' found: OK
  >> Extension sanity check command 'JULIA_DEBUG=loading julia -e 'using InitialValues' 2>&1 1>/dev/null | grep -E 'Loading (object )?cache file .*InitialValues' && julia -e 'using InitialValues'': OK
  >> (non-empty) directory 'packages/InitialValues' found: OK
  >> Extension sanity check command 'JULIA_DEBUG=loading julia -e 'using InverseFunctions' 2>&1 1>/dev/null | grep -E 'Loading (object )?cache file .*InverseFunctions' && julia -e 'using InverseFunctions'': OK
  >> (non-empty) directory 'packages/InverseFunctions' found: OK
  >> Extension sanity check command 'JULIA_DEBUG=loading julia -e 'using IrrationalConstants' 2>&1 1>/dev/null | grep -E 'Loading (object )?cache file .*IrrationalConstants' && julia -e 'using IrrationalConstants'': OK
  >> (non-empty) directory 'packages/IrrationalConstants' found: OK
  >> Extension sanity check command 'JULIA_DEBUG=loading julia -e 'using IteratorInterfaceExtensions' 2>&1 1>/dev/null | grep -E 'Loading (object )?cache file .*IteratorInterfaceExtensions' && julia -e 'using IteratorInterfaceExtensions'': OK
  >> (non-empty) directory 'packages/IteratorInterfaceExtensions' found: OK
  >> Extension sanity check command 'JULIA_DEBUG=loading julia -e 'using MLStyle' 2>&1 1>/dev/null | grep -E 'Loading (object )?cache file .*MLStyle' && julia -e 'using MLStyle'': OK
  >> (non-empty) directory 'packages/MLStyle' found: OK
  >> Extension sanity check command 'JULIA_DEBUG=loading julia -e 'using PrettyPrint' 2>&1 1>/dev/null | grep -E 'Loading (object )?cache file .*PrettyPrint' && julia -e 'using PrettyPrint'': OK
  >> (non-empty) directory 'packages/PrettyPrint' found: OK
  >> Extension sanity check command 'JULIA_DEBUG=loading julia -e 'using PtrArrays' 2>&1 1>/dev/null | grep -E 'Loading (object )?cache file .*PtrArrays' && julia -e 'using PtrArrays'': OK
  >> (non-empty) directory 'packages/PtrArrays' found: OK
  >> Extension sanity check command 'JULIA_DEBUG=loading julia -e 'using SciMLPublic' 2>&1 1>/dev/null | grep -E 'Loading (object )?cache file .*SciMLPublic' && julia -e 'using SciMLPublic'': OK
  >> (non-empty) directory 'packages/SciMLPublic' found: OK
  >> Extension sanity check command 'JULIA_DEBUG=loading julia -e 'using ShowCases' 2>&1 1>/dev/null | grep -E 'Loading (object )?cache file .*ShowCases' && julia -e 'using ShowCases'': OK
  >> (non-empty) directory 'packages/ShowCases' found: OK
  >> Extension sanity check command 'JULIA_DEBUG=loading julia -e 'using CommonSubexpressions' 2>&1 1>/dev/null | grep -E 'Loading (object )?cache file .*CommonSubexpressions' && julia -e 'using CommonSubexpressions'': OK
  >> (non-empty) directory 'packages/CommonSubexpressions' found: OK
  >> Extension sanity check command 'JULIA_DEBUG=loading julia -e 'using DataStructures' 2>&1 1>/dev/null | grep -E 'Loading (object )?cache file .*DataStructures' && julia -e 'using DataStructures'': OK
  >> (non-empty) directory 'packages/DataStructures' found: OK
  >> Extension sanity check command 'JULIA_DEBUG=loading julia -e 'using DelimitedFiles' 2>&1 1>/dev/null | grep -E 'Loading (object )?cache file .*DelimitedFiles' && julia -e 'using DelimitedFiles'': OK
  >> (non-empty) directory 'packages/DelimitedFiles' found: OK
  >> Extension sanity check command 'JULIA_DEBUG=loading julia -e 'using DiffResults' 2>&1 1>/dev/null | grep -E 'Loading (object )?cache file .*DiffResults' && julia -e 'using DiffResults'': OK
  >> (non-empty) directory 'packages/DiffResults' found: OK
  >> Extension sanity check command 'JULIA_DEBUG=loading julia -e 'using Missings' 2>&1 1>/dev/null | grep -E 'Loading (object )?cache file .*Missings' && julia -e 'using Missings'': OK
  >> (non-empty) directory 'packages/Missings' found: OK
  >> Extension sanity check command 'JULIA_DEBUG=loading julia -e 'using NameResolution' 2>&1 1>/dev/null | grep -E 'Loading (object )?cache file .*NameResolution' && julia -e 'using NameResolution'': OK
  >> (non-empty) directory 'packages/NameResolution' found: OK
  >> Extension sanity check command 'JULIA_DEBUG=loading julia -e 'using TableTraits' 2>&1 1>/dev/null | grep -E 'Loading (object )?cache file .*TableTraits' && julia -e 'using TableTraits'': OK
  >> (non-empty) directory 'packages/TableTraits' found: OK
  >> Extension sanity check command 'JULIA_DEBUG=loading julia -e 'using AliasTables' 2>&1 1>/dev/null | grep -E 'Loading (object )?cache file .*AliasTables' && julia -e 'using AliasTables'': OK
  >> (non-empty) directory 'packages/AliasTables' found: OK
  >> Extension sanity check command 'JULIA_DEBUG=loading julia -e 'using JuliaVariables' 2>&1 1>/dev/null | grep -E 'Loading (object )?cache file .*JuliaVariables' && julia -e 'using JuliaVariables'': OK
  >> (non-empty) directory 'packages/JuliaVariables' found: OK
  >> Extension sanity check command 'JULIA_DEBUG=loading julia -e 'using NaNMath' 2>&1 1>/dev/null | grep -E 'Loading (object )?cache file .*NaNMath' && julia -e 'using NaNMath'': OK
  >> (non-empty) directory 'packages/NaNMath' found: OK
  >> Extension sanity check command 'JULIA_DEBUG=loading julia -e 'using SortingAlgorithms' 2>&1 1>/dev/null | grep -E 'Loading (object )?cache file .*SortingAlgorithms' && julia -e 'using SortingAlgorithms'': OK
  >> (non-empty) directory 'packages/SortingAlgorithms' found: OK
  >> Extension sanity check command 'JULIA_DEBUG=loading julia -e 'using Tables' 2>&1 1>/dev/null | grep -E 'Loading (object )?cache file .*Tables' && julia -e 'using Tables'': OK
  >> (non-empty) directory 'packages/Tables' found: OK
  >> Extension sanity check command 'JULIA_DEBUG=loading julia -e 'using Accessors' 2>&1 1>/dev/null | grep -E 'Loading (object )?cache file .*Accessors' && julia -e 'using Accessors'': OK
  >> (non-empty) directory 'packages/Accessors' found: OK
  >> Extension sanity check command 'JULIA_DEBUG=loading julia -e 'using ProgressLogging' 2>&1 1>/dev/null | grep -E 'Loading (object )?cache file .*ProgressLogging' && julia -e 'using ProgressLogging'': OK
  >> (non-empty) directory 'packages/ProgressLogging' found: OK
  >> Extension sanity check command 'JULIA_DEBUG=loading julia -e 'using Setfield' 2>&1 1>/dev/null | grep -E 'Loading (object )?cache file .*Setfield' && julia -e 'using Setfield'': OK
  >> (non-empty) directory 'packages/Setfield' found: OK
  >> Extension sanity check command 'JULIA_DEBUG=loading julia -e 'using StructArrays' 2>&1 1>/dev/null | grep -E 'Loading (object )?cache file .*StructArrays' && julia -e 'using StructArrays'': OK
  >> (non-empty) directory 'packages/StructArrays' found: OK
  >> Extension sanity check command 'JULIA_DEBUG=loading julia -e 'using AbstractFFTs' 2>&1 1>/dev/null | grep -E 'Loading (object )?cache file .*AbstractFFTs' && julia -e 'using AbstractFFTs'': OK
  >> (non-empty) directory 'packages/AbstractFFTs' found: OK
  >> Extension sanity check command 'JULIA_DEBUG=loading julia -e 'using BangBang' 2>&1 1>/dev/null | grep -E 'Loading (object )?cache file .*BangBang' && julia -e 'using BangBang'': OK
  >> (non-empty) directory 'packages/BangBang' found: OK
  >> Extension sanity check command 'JULIA_DEBUG=loading julia -e 'using FillArrays' 2>&1 1>/dev/null | grep -E 'Loading (object )?cache file .*FillArrays' && julia -e 'using FillArrays'': OK
  >> (non-empty) directory 'packages/FillArrays' found: OK
  >> Extension sanity check command 'JULIA_DEBUG=loading julia -e 'using IRTools' 2>&1 1>/dev/null | grep -E 'Loading (object )?cache file .*IRTools' && julia -e 'using IRTools'': OK
  >> (non-empty) directory 'packages/IRTools' found: OK
  >> Extension sanity check command 'JULIA_DEBUG=loading julia -e 'using LogExpFunctions' 2>&1 1>/dev/null | grep -E 'Loading (object )?cache file .*LogExpFunctions' && julia -e 'using LogExpFunctions'': OK
  >> (non-empty) directory 'packages/LogExpFunctions' found: OK
  >> Extension sanity check command 'JULIA_DEBUG=loading julia -e 'using RealDot' 2>&1 1>/dev/null | grep -E 'Loading (object )?cache file .*RealDot' && julia -e 'using RealDot'': OK
  >> (non-empty) directory 'packages/RealDot' found: OK
  >> Extension sanity check command 'JULIA_DEBUG=loading julia -e 'using SimpleTraits' 2>&1 1>/dev/null | grep -E 'Loading (object )?cache file .*SimpleTraits' && julia -e 'using SimpleTraits'': OK
  >> (non-empty) directory 'packages/SimpleTraits' found: OK
  >> Extension sanity check command 'JULIA_DEBUG=loading julia -e 'using StatsAPI' 2>&1 1>/dev/null | grep -E 'Loading (object )?cache file .*StatsAPI' && julia -e 'using StatsAPI'': OK
  >> (non-empty) directory 'packages/StatsAPI' found: OK
  >> Extension sanity check command 'JULIA_DEBUG=loading julia -e 'using ChainRulesCore' 2>&1 1>/dev/null | grep -E 'Loading (object )?cache file .*ChainRulesCore' && julia -e 'using ChainRulesCore'': OK
  >> (non-empty) directory 'packages/ChainRulesCore' found: OK
  >> Extension sanity check command 'JULIA_DEBUG=loading julia -e 'using ContextVariablesX' 2>&1 1>/dev/null | grep -E 'Loading (object )?cache file .*ContextVariablesX' && julia -e 'using ContextVariablesX'': OK
  >> (non-empty) directory 'packages/ContextVariablesX' found: OK
  >> Extension sanity check command 'JULIA_DEBUG=loading julia -e 'using Functors' 2>&1 1>/dev/null | grep -E 'Loading (object )?cache file .*Functors' && julia -e 'using Functors'': OK
  >> (non-empty) directory 'packages/Functors' found: OK
  >> Extension sanity check command 'JULIA_DEBUG=loading julia -e 'using MLCore' 2>&1 1>/dev/null | grep -E 'Loading (object )?cache file .*MLCore' && julia -e 'using MLCore'': OK
  >> (non-empty) directory 'packages/MLCore' found: OK
  >> Extension sanity check command 'JULIA_DEBUG=loading julia -e 'using MicroCollections' 2>&1 1>/dev/null | grep -E 'Loading (object )?cache file .*MicroCollections' && julia -e 'using MicroCollections'': OK
  >> (non-empty) directory 'packages/MicroCollections' found: OK
  >> Extension sanity check command 'JULIA_DEBUG=loading julia -e 'using SplittablesBase' 2>&1 1>/dev/null | grep -E 'Loading (object )?cache file .*SplittablesBase' && julia -e 'using SplittablesBase'': OK
  >> (non-empty) directory 'packages/SplittablesBase' found: OK
  >> Extension sanity check command 'JULIA_DEBUG=loading julia -e 'using StatsBase' 2>&1 1>/dev/null | grep -E 'Loading (object )?cache file .*StatsBase' && julia -e 'using StatsBase'': OK
  >> (non-empty) directory 'packages/StatsBase' found: OK
  >> Extension sanity check command 'JULIA_DEBUG=loading julia -e 'using FLoopsBase' 2>&1 1>/dev/null | grep -E 'Loading (object )?cache file .*FLoopsBase' && julia -e 'using FLoopsBase'': OK
  >> (non-empty) directory 'packages/FLoopsBase' found: OK
  >> Extension sanity check command 'JULIA_DEBUG=loading julia -e 'using MLDataDevices' 2>&1 1>/dev/null | grep -E 'Loading (object )?cache file .*MLDataDevices' && julia -e 'using MLDataDevices'': OK
  >> (non-empty) directory 'packages/MLDataDevices' found: OK
  >> Extension sanity check command 'JULIA_DEBUG=loading julia -e 'using OpenSpecFun_jll' 2>&1 1>/dev/null | grep -E 'Loading (object )?cache file .*OpenSpecFun_jll' && julia -e 'using OpenSpecFun_jll'': OK
  >> (non-empty) directory 'packages/OpenSpecFun_jll' found: OK
  >> Extension sanity check command 'JULIA_DEBUG=loading julia -e 'using Optimisers' 2>&1 1>/dev/null | grep -E 'Loading (object )?cache file .*Optimisers' && julia -e 'using Optimisers'': OK
  >> (non-empty) directory 'packages/Optimisers' found: OK
  >> Extension sanity check command 'JULIA_DEBUG=loading julia -e 'using SparseInverseSubset' 2>&1 1>/dev/null | grep -E 'Loading (object )?cache file .*SparseInverseSubset' && julia -e 'using SparseInverseSubset'': OK
  >> (non-empty) directory 'packages/SparseInverseSubset' found: OK
  >> Extension sanity check command 'JULIA_DEBUG=loading julia -e 'using Transducers' 2>&1 1>/dev/null | grep -E 'Loading (object )?cache file .*Transducers' && julia -e 'using Transducers'': OK
  >> (non-empty) directory 'packages/Transducers' found: OK
  >> Extension sanity check command 'JULIA_DEBUG=loading julia -e 'using ZygoteRules' 2>&1 1>/dev/null | grep -E 'Loading (object )?cache file .*ZygoteRules' && julia -e 'using ZygoteRules'': OK
  >> (non-empty) directory 'packages/ZygoteRules' found: OK
  >> Extension sanity check command 'JULIA_DEBUG=loading julia -e 'using ChainRules' 2>&1 1>/dev/null | grep -E 'Loading (object )?cache file .*ChainRules' && julia -e 'using ChainRules'': OK
  >> (non-empty) directory 'packages/ChainRules' found: OK
  >> Extension sanity check command 'JULIA_DEBUG=loading julia -e 'using FLoops' 2>&1 1>/dev/null | grep -E 'Loading (object )?cache file .*FLoops' && julia -e 'using FLoops'': OK
  >> (non-empty) directory 'packages/FLoops' found: OK
  >> Extension sanity check command 'JULIA_DEBUG=loading julia -e 'using SpecialFunctions' 2>&1 1>/dev/null | grep -E 'Loading (object )?cache file .*SpecialFunctions' && julia -e 'using SpecialFunctions'': OK
  >> (non-empty) directory 'packages/SpecialFunctions' found: OK
  >> Extension sanity check command 'JULIA_DEBUG=loading julia -e 'using DiffRules' 2>&1 1>/dev/null | grep -E 'Loading (object )?cache file .*DiffRules' && julia -e 'using DiffRules'': OK
  >> (non-empty) directory 'packages/DiffRules' found: OK
  >> Extension sanity check command 'JULIA_DEBUG=loading julia -e 'using NNlib' 2>&1 1>/dev/null | grep -E 'Loading (object )?cache file .*NNlib' && julia -e 'using NNlib'': OK
  >> (non-empty) directory 'packages/NNlib' found: OK
  >> Extension sanity check command 'JULIA_DEBUG=loading julia -e 'using ForwardDiff' 2>&1 1>/dev/null | grep -E 'Loading (object )?cache file .*ForwardDiff' && julia -e 'using ForwardDiff'': OK
  >> (non-empty) directory 'packages/ForwardDiff' found: OK
  >> Extension sanity check command 'JULIA_DEBUG=loading julia -e 'using MLUtils' 2>&1 1>/dev/null | grep -E 'Loading (object )?cache file .*MLUtils' && julia -e 'using MLUtils'': OK
  >> (non-empty) directory 'packages/MLUtils' found: OK
  >> Extension sanity check command 'JULIA_DEBUG=loading julia -e 'using OneHotArrays' 2>&1 1>/dev/null | grep -E 'Loading (object )?cache file .*OneHotArrays' && julia -e 'using OneHotArrays'': OK
  >> (non-empty) directory 'packages/OneHotArrays' found: OK
  >> Extension sanity check command 'JULIA_DEBUG=loading julia -e 'using Zygote' 2>&1 1>/dev/null | grep -E 'Loading (object )?cache file .*Zygote' && julia -e 'using Zygote'': OK
  >> (non-empty) directory 'packages/Zygote' found: OK
  >> Extension sanity check command 'JULIA_DEBUG=loading julia -e 'using Flux' 2>&1 1>/dev/null | grep -E 'Loading (object )?cache file .*Flux' && julia -e 'using Flux'': OK
  >> (non-empty) directory 'packages/Flux' found: OK
  == ... (took 52 secs)
```

</p>
</details> 

> I don't see this happening as it isn't used for the main-Easyconfig, is it?

If you name your EC something like `Flux-julia` but the Julia package name is just `Flux` you will get an `julia -e 'using Flux-julia''`  that will fail if i recall correctly will test again but I am sure i put it there because something was failing without

Crivella: Also a test skipping sanity checks (not sure if this is what you were refering to)

<details><summary>Details</summary>
<p>

```
(easybuild-dev) crivella@crivella-desktop:~$ eb --installpath=$(eb_tmpdir2) --from-pr 26214 --rebuild --module-only --skip-sanity-check Flux-0.16.10-Julia-1.12.2.eb 
== Temporary log file in case of crash /home/crivella/.local/easybuild/logs/easybuild-2yay9kyx.log

WARNING: Index for /home/crivella/Documents/GIT/easybuild-easyconfigs/easybuild/easyconfigs is no longer valid (too old), so ignoring it...

== processing EasyBuild easyconfig /tmp/eb-nsa6gyjq/files_pr26214/f/Flux/Flux-0.16.10-Julia-1.12.2.eb
== building and installing Flux/0.16.10-Julia-1.12.2...
  >> installation prefix: /tmp/eb_tmpdir-W0Mrhw79/software/Flux/0.16.10-Julia-1.12.2
== fetching files and verifying checksums [skipped]
== creating build dir, resetting environment...
  >> build dir: /home/crivella/.local/easybuild/build/Flux/0.16.10/system-system-Julia-1.12.2                                                                                                                                                                                                   
== ... (took < 1 sec)                                                                                                                                                                                                                                                                           
== unpacking [skipped]                                                                                                                                                                                                                                                                          
== patching [skipped]                                                                                                                                                                                                                                                                           
== preparing...                                                                                                                                                                                                                                                                                 
  >> (no build dependencies specified)                                                                                                                                                                                                                                                          
  >> loading modules for (runtime) dependencies:                                                                                                                                                                                                                                                
  >>  * Julia/1.12.2                                                                                                                                                                                                                                                                            
  >>  * Compat/4.18.1-Julia-1.12.2                                                                                                                                                                                                                                                              
  >>  * GPUArrays/11.5.5-Julia-1.12.2                                                                                                                                                                                                                                                           
  >> Incorporating Julia package in INSTALL environment Compat: /home/crivella/.local/easybuild/software/Compat/4.18.1-Julia-1.12.2/environments/v1.12                                                                                                                                          
  >> Incorporating Julia package in INSTALL environment GPUArrays: /home/crivella/.local/easybuild/software/GPUArrays/11.5.5-Julia-1.12.2/environments/v1.12                                                                                                                                    
== ... (took 2 secs)                                                                                                                                                                                                                                                                            
== configuring [skipped]                                                                                                                                                                                                                                                                        
== building [skipped]                                                                                                                                                                                                                                                                           
== testing [skipped]                                                                                                                                                                                                                                                                            
== installing [skipped]                                                                                                                                                                                                                                                                         
== taking care of extensions [skipped]                                                                                                                                                                                                                                                          
== restore after iterating...                                                                                                                                                                                                                                                                   
== ... (took < 1 sec)                                                                                                                                                                                                                                                                           
== postprocessing [skipped]                                                                                                                                                                                                                                                                     
== sanity checking [skipped]                                                                                                                                                                                                                                                                    
== cleaning up [skipped]                                                                                                                                                                                                                                                                        
== creating module...                                                                                                                                                                                                                                                                           
  >> generating module file @ /tmp/eb_tmpdir-W0Mrhw79/modules/all/Flux/0.16.10-Julia-1.12.2.lua                                                                                                                                                                                                 
== ... (took < 1 sec)                                                                                                                                                                                                                                                                           
== permissions [skipped]                                                                                                                                                                                                                                                                        
== packaging [skipped]                                                                                                                                                                                                                                                                          
== COMPLETED: Installation ended successfully (took 3 secs)                                                                                                                                                                                                                                     
== Results of the build can be found in the log file(s) /tmp/eb_tmpdir-W0Mrhw79/software/Flux/0.16.10-Julia-1.12.2/easybuild/easybuild-Flux-0.16.10-20260716.115354.log                                                                                                                         
== Build succeeded for 1 out of 1 (total: 3 secs)                                                                                                                                                                                                                                               
== Summary:                                                                                                                                                                                                                                                                                     
   * [SUCCESS] Flux/0.16.10-Julia-1.12.2                                                                                                                                                                                                                                                        
== Temporary log file(s) /home/crivella/.local/easybuild/logs/easybuild-2yay9kyx.log* have been removed.                                                                                                                                                                                        
== Temporary directory /tmp/eb-nsa6gyjq has been removed.
```

</p>
</details>

easybuild/easyblocks/generic/juliapackage.py
Crivella: I do not understand what you mean, in the case of a Bundle the `sanity_check_step` is handling setting it's own `exts_filter` in case it is an extension or just checking for the expected dir if it is the main bundle.

Do not remember if i tested it for simple packages, but with this, in the same spirit as python we should always use bundles

easybuild/easyblocks/generic/juliabundle.py
Crivella: This is what happens as i mentioned if you want to name your EC with a name that does not match the Julia package itself while keeping `self.cfg['exts_filter'] = EXTS_FILTER_JULIA_PACKAGES` in the JuliaBundle

<details><summary>Details</summary>
<p>

```
(easybuild-dev) crivella@crivella-desktop:~$ eb --rebuild --module-only Flux-0.16.10-Julia-1.12.2.eb                                                                                                                                                                                            
== Temporary log file in case of crash /home/crivella/.local/easybuild/logs/easybuild-yrajq77e.log                                                                                                                                                                                              
                                                                                                                                                                                                                                                                                                
WARNING: Index for /home/crivella/Documents/GIT/easybuild-easyconfigs/easybuild/easyconfigs is no longer valid (too old), so ignoring it...                                                                                                                                                     
                                                                                                                                                                                                                                                                                                
== processing EasyBuild easyconfig /home/crivella/Documents/GIT/easybuild-easyconfigs/easybuild/easyconfigs/f/Flux/Flux-0.16.10-Julia-1.12.2.eb                                                                                                                                                 
== building and installing Flux-julia/0.16.10-Julia-1.12.2...                                                                                                                                                                                                                                   
  >> installation prefix: /home/crivella/.local/easybuild/software/Flux-julia/0.16.10-Julia-1.12.2                                                                                                                                                                                              
== fetching files and verifying checksums [skipped]                                                                                                                                                                                                                                             
== creating build dir, resetting environment...                                                                                                                                                                                                                                                 
== backup of existing module file stored at /home/crivella/.local/easybuild/modules/all/Flux-julia/0.16.10-Julia-1.12.2.bak_20260716121328_1891994                                                                                                                                              
  >> build dir: /home/crivella/.local/easybuild/build/Fluxjulia/0.16.10/system-system-Julia-1.12.2                                                                                                                                                                                              
== ... (took < 1 sec)                                                                                                                                                                                                                                                                           
== unpacking [skipped]                                                                                                                                                                                                                                                                          
== patching [skipped]                                                                                                                                                                                                                                                                           
== preparing...                                                                                                                                                                                                                                                                                 
  >> (no build dependencies specified)                                                                                                                                                                                                                                                          
  >> loading modules for (runtime) dependencies:                                                                                                                                                                                                                                                
  >>  * Julia/1.12.2                                                                                                                                                                                                                                                                            
  >>  * Compat/4.18.1-Julia-1.12.2                                                                                                                                                                                                                                                              
  >>  * GPUArrays/11.5.5-Julia-1.12.2                                                                                                                                                                                                                                                           
  >> Incorporating Julia package in INSTALL environment Compat: /home/crivella/.local/easybuild/software/Compat/4.18.1-Julia-1.12.2/environments/v1.12                                                                                                                                          
  >> Incorporating Julia package in INSTALL environment GPUArrays: /home/crivella/.local/easybuild/software/GPUArrays/11.5.5-Julia-1.12.2/environments/v1.12                                                                                                                                    
== ... (took 2 secs)                                                                                                                                                                                                                                                                            
== configuring [skipped]                                                                                                                                                                                                                                                                        
== building [skipped]                                                                                                                                                                                                                                                                           
== testing [skipped]                                                                                                                                                                                                                                                                            
== installing [skipped]                                                                                                                                                                                                                                                                         
== taking care of extensions [skipped]                                                                                                                                                                                                                                                          
== restore after iterating...                                                                                                                                                                                                                                                                   
== ... (took < 1 sec)                                                                                                                                                                                                                                                                           
== postprocessing [skipped]                                                                                                                                                                                                                                                                     
== sanity checking...                                                                                                                                                                                                                                                                           
  >> Extension sanity check command 'julia -e 'using Flux-julia'': FAIL                                                                                                                                                                                                                         
  >> (non-empty) directory 'packages' found: OK         
  >> (non-empty) directory 'environments' found: OK
  >> loading modules: Flux-julia/0.16.10-Julia-1.12.2.
```

</p>
</details> 

without it works (OFC not using `name` but hardcoding the name for the extension) eg

```
    ('Flux', version, {
```

easybuild/easyblocks/generic/juliapackage.py
Flamefire: You set `exts_filter` outside the `if` but only use it inside. So why not set it inside, closer to where it is used?

For non-extensions I'm quite sure `exts_filter` won't be used so you need the `custom_cmds` above. See comment there

> in the same spirit as python we should always use bundles

I still think this is counter-productive as bundles for a single package are awkward to write and invite adding more packages that would better be suited for own easyconfigs.
I'm referring to both syntax by having to put everything in a dict with lots of extra quoting and the restrictions on what extensions can do, e.g. single-source only

easybuild/easyblocks/generic/juliapackage.py
```diff
@@ -231,96 +421,269 @@ def install_pkg_source(self, pkg_source, environment, trace=True):
...
+        if LooseVersion(self.julia_version) >= LooseVersion('1.8'):
+            cc_check = True
+            exts_filter.insert(0, _COMPILECACHE_CHECK % {'ext_name': self.name, 'grep_loc': self.name})
+
+        if self.is_extension:
+            pkg_dir = os.path.join('packages', self.name)
+
+            custom_commands = []
+            # Check that the compile cache of the dependencies can still be loaded
+            if not self.is_extension and cc_check:
```
Flamefire: This is inside `self.is_extension` so will always be false.
Suggestion: Constructs exts_filter once and add it to `custom_commands` if not an extension.

Crivella: You are right there was no if here before and i ended up moving this block in the wrong branch.
Fixed by [42a23d1](https://github.com/easybuilders/easybuild-easyblocks/pull/4146/commits/42a23d1e00b99fb0ca0c1ab110c541bb04fd3af6)

This indeed was not there in [my previous logs](https://github.com/easybuilders/easybuild-easyblocks/pull/4146#discussion_r3594373761)
```
== sanity checking...                                                                                                                                                                                                                                                                                                                                                                                                                                                   
  >> (non-empty) directory 'packages' found: OK                                                                                                                                                                                                                                                 
  >> (non-empty) directory 'environments' found: OK                                                                                                                                                                                                                                             
  >> loading modules: Flux/0.16.10-Julia-1.12.2...                                                                                                                                                                                                                                              
  >> running command 'JULIA_DEBUG=loading julia -e 'using Compat' 2>&1 1>/dev/null | grep -E 'Loading (object )?cache file .*Compat'' ...                                                                                                                                                       
  >> result for command 'JULIA_DEBUG=loading julia -e 'using Compat' 2>&1 1>/dev/null | grep -E 'Loading (object )?cache file .*Compat'': OK                                                                                                                                                    
  >> running command 'JULIA_DEBUG=loading julia -e 'using GPUArrays' 2>&1 1>/dev/null | grep -E 'Loading (object )?cache file .*GPUArrays'' ...                                                                                                                                                 
  >> result for command 'JULIA_DEBUG=loading julia -e 'using GPUArrays' 2>&1 1>/dev/null | grep -E 'Loading (object )?cache file .*GPUArrays'': OK 
...
```

easybuild/easyblocks/generic/juliabundle.py
Flamefire: > I do not understand what you mean. exts_filter is used in the sanity check step of the extensions

I missed that. I thought it is only used to filter extensions. Badly named then.

> Also a test skipping sanity checks (not sure if this is what you were refering to)

I meant use of `--skip` to skip installing (i.e. filter, see above) already installed extensions, see https://github.com/easybuilders/easybuild-framework/blob/ccbc8ea852afb2a1ec83988eac07e7e9caa40212/easybuild/framework/easyblock.py#L2028

> If you name your EC something like Flux-julia but the Julia package name is just Flux you will get an julia -e 'using Flux-julia'' that will fail if i recall correctly will test again but I am sure i put it there because something was failing without

I see. Seems this simply wasn't the case in most ECs and the ~4 software we have use e.g. `options = {'modulename': 'onnxruntime'}`. IIRC there was talk about making that a top-level param to have `modulename = "onnxruntime"`

Maybe we shouldn't use the `exts_filter` of the parent for skipping, but that of each extension. Not sure if that works though but seems more correct.

easybuild/easyblocks/generic/juliapackage.py
```diff
@@ -231,96 +421,265 @@ def install_pkg_source(self, pkg_source, environment, trace=True):
...
+        if self.cfg['is_test_dependency']:
+            self.log.debug(f"Package {self.name} is only used for testing, skipping sanity check")
+            return (True, "Test dependency, skipping sanity check")
 
-        pkg_dir = os.path.join('packages', self.name)
+        exts_filter = ["julia -e 'using %(ext_name)s'"]
+        cc_check = False
+        if LooseVersion(self.julia_version) >= LooseVersion('1.8'):
+            cc_check = True
+            exts_filter.insert(0, _COMPILECACHE_CHECK % {'ext_name': self.name, 'grep_loc': self.name})
```
Flamefire: To be specific:

```suggestion
        cc_check = LooseVersion(self.julia_version) >= LooseVersion('1.8')
```
And inside the branch below:
```
        exts_filter = "julia -e 'using %(ext_name)s'"
        if cc_check:
            cc_check = True
            exts_filter = _COMPILECACHE_CHECK % {'ext_name': self.name, 'grep_loc': self.name} + f" && exts_filter"
        self.cfg['exts_filter'] = (exts_filter, "")
```

But given that the intention of `exts_filter` is (or at least was) to check if the extension is installed I'd say that the `_COMPILECACHE_CHECK` belongs to the `custom_commands`

Crivella: I guess your idea here is to move `exts_filter` i the only part of the branch that is using it? Will still keep using lists, i think it is more readable than concatenating strings with the risk of missing spaces and such

Crivella: [516e550](https://github.com/easybuilders/easybuild-easyblocks/pull/4146/commits/516e550960133f6dbb6c1d17f1113a5f0b0fd9b7)

easybuild/easyblocks/generic/juliapackage.py
Crivella: > You set exts_filter outside the if but only use it inside. So why not set it inside, closer to where it is used?

Can agree on this (same point as the other thread and solved)

Crivella: > I still think this is counter-productive as bundles for a single package are awkward to write and invite adding more packages that would better be suited for own easyconfigs.

I mean that is a decision that was taken more broadly, do not think this is the right venue to try and change this

easybuild/easyblocks/generic/juliapackage.py
```diff
@@ -231,96 +421,267 @@ def install_pkg_source(self, pkg_source, environment, trace=True):
...
+            self.log.debug(f"Package {self.name} is only used for testing, skipping sanity check")
+            return (True, "Test dependency, skipping sanity check")
+
+        cc_check = LooseVersion(self.julia_version) >= LooseVersion('1.8')
+
+        if self.is_extension:
+            exts_filter = []
+            if cc_check:
+                exts_filter.append(_COMPILECACHE_CHECK % {'ext_name': self.name, 'grep_loc': self.name})
+            exts_filter.append("julia -e 'using %(ext_name)s'")
```
Flamefire: Duplicate now

```suggestion
```

easybuild/easyblocks/generic/juliapackage.py
Flamefire: > Will still keep using lists, i think it is more readable than concatenating strings with the risk of missing spaces and such

What about using `custom_commands` for that? See my earlier comment on possibly using `exts_filter` from the extension instead of the parent for filtering. Then it would become an issue and I'd say it is cleaner anyway to separate the "does it exist" from the "does it work" check.

> I mean that is a decision that was taken more broadly, do not think this is the right venue to try and change this

Not sure where the right venue is but I'd like to bring that up before applying it to other Bundle-Package-changes than already is

easybuild/easyblocks/generic/juliabundle.py
Crivella: I think in general `--skip` should work.
The main problem that is making this useless is the fact that we are re-creating the install dir at the prepare step which basically makes it useless.
This was needed to avoid problems with `--rebuild` but would need to look back up what was happening exactly

Crivella: Also not sure it would be a good Idea in Julia packages as in general it is possible that an extension has to be re-installed as it might build additional Julia extensions (in the julia sense) when built together with another package

Flamefire: > I think in general `--skip` should work.

How can it if you removed the filter? It will just be a hard error if you try to: https://github.com/easybuilders/easybuild-framework/blob/ccbc8ea852afb2a1ec83988eac07e7e9caa40212/easybuild/framework/easyblock.py#L2031

> The main problem that is making this useless is the fact that we are re-creating the install dir at the prepare step which basically makes it useless.

That is new. So `--skip` basically doesn't work anymore. Since when? Especially for that huge Perl- or R-bundles it was really helpful

easybuild/easyblocks/generic/juliapackage.py
Crivella: Was not even that aware that `exts_filter` was called so because of the `--skip` (even then still bad name), but for what i can see it is in general used to define the sanity check to be performed on an extension.

> What about using custom_commands for that? 

I think twice you refered to a wrong quote so i am not sure i am getting what you mean.

I think the problem is that we do not want to run the commands themselves on the Bundle.

We either split the sanity check for bundles in JuliaBundle and for Package only in JuliaPackage or i do not see this working. That is not just trivial refactoring...

easybuild/easyblocks/generic/juliabundle.py
Crivella: > How can it if you removed the filter? It will just be a hard error if you try to

Umh i see what you mean, for the `--skip` the one at the bundle level is used not the one for the individual ext that we set in the sanity checks.
Still I really do not think this should be a thing for Julia Bundles, it has a huge probability of breaking stuff

> That is new. So --skip basically doesn't work anymore. Since when? Especially for that huge Perl- or R-bundles it was really helpful

I mean here for JuliaBundle, not in genera, due to

```
    def make_installdir(self):
        """Create new installation directory and ensure this is done only once"""
        # Stop installdir from being re-created at the install step
        if not self._installdir_created:
            super().make_installdir()
            self._installdir_created = True

    def prepare_step(self, *args, **kwargs):
        """Prepare for Julia package installation."""
        super().prepare_step(*args, **kwargs)
        self.make_installdir()
```

Crivella: Also if this PR goes through we really install the extension all at once already using Pkg.instantiate, so it is really fast even with a lot of them.
Basically every extension only does the unpacking and adds its info to the Manifest and than the last extension also does the instantiation which allows to install the packages in parallel instead of serially + removes the problem of needing the exts to be ordered

easybuild/easyblocks/generic/juliapackage.py
Flamefire: > but for what i can see it is in general used to define the sanity check to be performed on an extension.

I guess it was an after-thought to use it as part of the sanity check too, which made sense when the definition is/was(?) basically: Check if this package is installed

> I think twice you refered to a wrong quote so i am not sure i am getting what you mean.
 
What I meant: Have `exts_filter` be the simple existence check (`julia -e ...`), which (I think) it was supposed to be, and move the other command(s) that checks for correctness to `custom_commands`. This way you don't have to care about how separating and joining commands and it is cleaner, to me at least.

> I think the problem is that we do not want to run the commands themselves on the Bundle.

If I see this correctly for extensions you run `_COMPILECACHE_CHECK` once, otherwise for each Julia package in all dependencies.
That's what you do right now: The bundle runs it through `custom_commands` already, the package and extension through `exts_filter`
If all you change is moving `exts_filter.append(_COMPILECACHE_CHECK % {'ext_name': self.name, 'grep_loc': self.name})` to `custom_commands` instead and always add that, the behavior will still be the same, won't it?

I would even set `cfg[exts_filter]` in the constructor, so it can be changed by derived classes and later be used instead of the parent/bundle easyconfig for `--skip` (if we do that change at some point)

Crivella: I have a feeling that for this to work properly

https://github.com/easybuilders/easybuild-framework/blob/a464f4fa70b8df3bb4f54976d31f0aae199923cb/easybuild/framework/extensioneasyblock.py#L191
should really be a `elif not self.is_extension`.

Not sure why we would run the `Extension` sanity check on the bundle.
This combined with a

```python
        if not self.is_extension:
            self.sanity_check_load_module()
```

at the start of JuliaPackage sanity check would solve the issue.

Not sure though if the first part of the fix would have unforseen consequences, the way code for bundles is written is way too entangled for me to predict (and at the very least i would not block this PR for this since it can be added as a feature later).

easybuild/easyblocks/generic/juliapackage.py
```diff
@@ -26,40 +26,39 @@
...
-from easybuild.tools.filetools import copy_dir, mkdir
+from easybuild.tools.filetools import move_file
 from easybuild.tools.run import run_shell_cmd
-from easybuild.tools.utilities import trace_msg
+from easybuild.tools.utilities import trace_msg, print_msg
+from easybuild.tools.hooks import BUILD_STEP, TEST_STEP
 
-EXTS_FILTER_JULIA_PACKAGES = ("julia -e 'using %(ext_name)s'", "")
-USER_DEPOT_PATTERN = re.compile(r"\/\.julia\/?(.*\.toml)*$")
+from easybuild.easyblocks.j.julia import EB_JULIA_DEPOT_PATH_VAR, EB_JULIA_LOAD_PATH_VAR
```
Flamefire:
```suggestion
from easybuild.easyblocks.julia import EB_JULIA_DEPOT_PATH_VAR, EB_JULIA_LOAD_PATH_VAR
```

Otherwise e.g. `--include-easyblocks-from-pr` doesn't work which places `julia.py` in `easybuild/easyblocks`
The letter-folders are transparent anyway
See e.g. `from easybuild.easyblocks.perl import...` or `from easybuild.easyblocks.python import UNLIMITED`

easybuild/easyblocks/generic/juliapackage.py
```diff
@@ -231,96 +421,263 @@ def install_pkg_source(self, pkg_source, environment, trace=True):
...
+        if self.cfg['is_test_dependency']:
+            self.log.debug(
+                f"Package {self.name} is only needed for testing, installing sources in {self.tmp_test_dir}"
+            )
+            trace_msg("Installing as a test dependency...")
+            package_dir = os.path.join(self.tmp_test_dir, 'packages', self.name)
+        else:
+            package_dir = os.path.join(self.installdir, 'packages', self.name)
+
+        move_file(self.start_dir, package_dir)
```
Flamefire: This causes failures for me, e.g. in the Pluto easyconfig: When extracting sources we cd into `start_dir`. You move that here so we are now in a folder that is missing. The test then fails to run due to that.

Flamefire:
I now think this approach might not be the best: The major issue is that a JuliaBundle needs to be a Julia package or things fail at least in the sanity check. I'm not fully sure if this is really inherited but IIRC Bundles cannot have sources themselves, so that doesn't make sense to treat it as a package.
Hence I think the better approach is to make JuliaBundle inherit only from Bundle similar to the other bundles like PythonBundle.
To avoid duplicated work I imagine that it is possible to do it once in the Bundle and disable it in the extensions. 

Because unless we enforce that JuliaPackages are only ever part of a JuliaBundle then relying on `self.is_extension` to decide what to do is very fragile and may break at least standalone installations or when part of other easyconfigs.
E.g. I recently had a similar use case: Install some software with CMakeMake and then the Python wrappers via `exts_list`. I imagine this to be a use case for Julia too

Crivella:
> I now think this approach might not be the best:

So the main reason is that this was how the package was structured in the beginning so when i started making changes i tried to keep the original structure.

If the only problem is using `exts_filters` to enable `--skip` that is a non issue since we have to delete the installdir before we start an installation (the old way of installing them all 1by1 is tens of times slower than doing them all at once).
Also the mechanism of installing them all at once would become much more difficult to handle since that would need to be implemented twice (once for packages trough deps and once for bundles).

> E.g. I recently had a similar use case: Install some software with CMakeMake and then the Python wrappers via exts_list. I imagine this to be a use case for Julia too
Probably not impossible but much less likely for a Julia package to use external libraries (usually they compile everything within Julia)

I'd be more in favor of hiding/disallowing to make use of `JuliaPackage` directly than do double work on the easyblocks especially since it is policy in EB to prefer bundles over single packages.
Probably considering how little is happening in `JuliaBundle` itself we could just move that extra part in `JuliaPackage` and rename it to `JuliaBundle`

easybuild/easyblocks/generic/juliapackage.py
```diff
@@ -106,51 +130,168 @@ def get_julia_env(env_var):
...
+        """Enable online/offline mode of Julia Pkg"""
 
-    def prepare_julia_env(self):
+        julia_version = get_software_version('Julia')
+        if LooseVersion(julia_version) >= LooseVersion('1.5'):
+            # Enable offline mode of Julia Pkg
+            # https://pkgdocs.julialang.org/v1/api/#Pkg.offline
+            env.setvar('JULIA_PKG_OFFLINE', 'false' if online else 'true')
+        else:
+            errmsg = (
```
Flamefire, line 287: How does this not cause an error in Julia <1.5? Shouldn't you at least only error when `online == True`?
Also `download_pkg_deps` doesn't avoid this error: If set to True the error will still happen when passed to this function. And `pkgs_to_test_online` also triggers this error unconditionally and without need.
And maybe also mention `test_online`

Flamefire:
> if the only problem is using `exts_filters` to enable `--skip`

It is not only that. Also `'dirs': [os.path.join('packages', self.name)],` won't be checked for standalone packages anymore. Maybe there are other related changes I missed

> I'd be more in favor of hiding/disallowing to make use of JuliaPackage directly than do double work on the easyblocks 

It could be a shared method. Or if the main point is doing a single "install cmd" on the last extension that could be something done by the Bundle: Go over the extensions and on the last consecutive JuliaPackage issue an install command on it, e.g. by passing an appropriate parameter to a method of that extensions easyblock

> especially since it is policy in EB to prefer bundles over single packages.

Besides that I wanted to discuss that I would not go so far as assuming that every `JuliaPackage` is part of a `JuliaBundle` and assuming things will be done there. As Julia allows wrapping C libraries I do expect the use case of having them as extensions to other easyblocks.
Maybe this at least needs to check its assumptions: That the parent is a JuliaBundle and that all extensions are JuliaPackages. Otherwise e.g. the use of `is_last_extension` will be unreliable.
That would exclude installing Julia wrappers together with their main library in the same module but at least avoids silently wrong behavior.

PR #4153, opened 2026-06-11
disable `rust.lld` for Rust >= 1.90, use system linker instead

### LINKED ISSUE #26232: RPATH-ing broken for `Rust-1.91.1-GCCcore-14.3.0.eb`
With RPATH enabled I expected the build to work even if I unset `LD_LIBRARY_PATH`, but it doesn't.

```
[Test c3-builder@vera-r08-02(Icelake) c3se-test-easyconfigs]$ ml purge
[Test c3-builder@vera-r08-02(Icelake) c3se-test-easyconfigs]$ ml Rust/1.91.1-GCCcore-14.3.0 EasyBuild/5.3.0
[Test c3-builder@vera-r08-02(Icelake) c3se-test-easyconfigs]$ eb --show-full-config | grep '^rpath '
rpath                                     (D) = True
[Test c3-builder@vera-r08-02(Icelake) c3se-test-easyconfigs]$ rustc -vV
rustc 1.91.1 (ed61e7d7e 2025-11-07) (built from a source tarball)
binary: rustc
commit-hash: ed61e7d7e242494fb7057f2657300d9e77bb4fcb
commit-date: 2025-11-07
host: x86_64-unknown-linux-gnu
release: 1.91.1
LLVM version: 21.1.2
[Test c3-builder@vera-r08-02(Icelake) c3se-test-easyconfigs]$ unset LD_LIBRARY_PATH && rustc -vV
rustc: /lib64/libstdc++.so.6: version `GLIBCXX_3.4.30' not found (required by /apps/Test/software/Rust/1.91.1-GCCcore-14.3.0/bin/../lib/librustc_driver-6632e262b06c890f.so)
rustc: /lib64/libstdc++.so.6: version `GLIBCXX_3.4.32' not found (required by /apps/Test/software/Rust/1.91.1-GCCcore-14.3.0/bin/../lib/librustc_driver-6632e262b06c890f.so) 
```

Crivella:
Ensure `rust.lld` is not used by default with RPATH-ing on to avoid it bypassing Easybuild's wrappers.

An alternative solution would be to ship also wrappers that encapsulate `rust.lld`

- check whether normal LLVM wrappers would be enough (as simple LDD might be called under the hood) or if we need a dedicated one

Fixes
- Fix https://github.com/easybuilders/easybuild-easyconfigs/issues/26232

### DIFF
--- a/easybuild/easyblocks/r/rust.py
+++ b/easybuild/easyblocks/r/rust.py
@@ -114,6 +114,11 @@ class EB_Rust(ConfigureMake):
         # old llvm builds from CI get deleted after a certain time
         self.cfg.update('configopts', "--set=llvm.download-ci-llvm=false")
 
+        # Use system linker to avoid rust.ldd to bypass our rpath wrappers
+        # See https://github.com/easybuilders/easybuild-easyconfigs/issues/26232 for more details
+        if build_option('rpath') and LooseVersion(self.version) >= '1.9':
+            self.cfg.update('configopts', "--disable-lld")
+
         # set channel to "stable", otherwise Rust will be built with nightly channel,
         # see also https://rust-lang.github.io/rustup/concepts/channels.html;
         # for recent version of Rust, this would also result in using rust-lld instead of the default linker,
--- a/easybuild/easyblocks/r/rust.py
+++ b/easybuild/easyblocks/r/rust.py
@@ -116,7 +116,7 @@ class EB_Rust(ConfigureMake):
 
         # Use system linker to avoid rust.ldd to bypass our rpath wrappers
         # See https://github.com/easybuilders/easybuild-easyconfigs/issues/26232 for more details
-        if build_option('rpath') and LooseVersion(self.version) >= '1.9':
+        if build_option('rpath') and LooseVersion(self.version) >= '1.90':
             self.cfg.update('configopts', "--disable-lld")
 
         # set channel to "stable", otherwise Rust will be built with nightly channel,


easybuild/easyblocks/r/rust.py
```diff
@@ -114,6 +114,11 @@ def configure_step(self):
         # old llvm builds from CI get deleted after a certain time
         self.cfg.update('configopts', "--set=llvm.download-ci-llvm=false")
 
+        # Use system linker to avoid rust.ldd to bypass our rpath wrappers
+        # See https://github.com/easybuilders/easybuild-easyconfigs/issues/26232 for more details
+        if build_option('rpath') and LooseVersion(self.version) >= '1.90':
+            self.cfg.update('configopts', "--disable-lld")
```
boegel, line 120: Woudln't wrapping `rust.lld` be a better option?

easybuild/easyblocks/r/rust.py
```diff
@@ -114,6 +114,11 @@ def configure_step(self):
         # old llvm builds from CI get deleted after a certain time
         self.cfg.update('configopts', "--set=llvm.download-ci-llvm=false")
 
+        # Use system linker to avoid rust.ldd to bypass our rpath wrappers
```
boegel, line 117:
```suggestion
        # Use system linker to avoid rust.lld to bypass our rpath wrappers
```

boegel:
Alternative fix, along with sanity check, implemented in:
- https://github.com/easybuilders/easybuild-easyblocks/pull/4156

easybuild/easyblocks/r/rust.py
Crivella: Yeah, my idea here was to have a quick fix that should restore the behavior of Rust pre-1.90 while we work on a more proper solution.
As far as i understand they switched to the internal linker since they had been using it for a long time in CI, but the system one should still be expected to work

Crivella:
Closing in favor of 

- #4156

PR #4154, opened 2026-06-12
Add extra easyconfig parameter `object_storage_ignore_dirs` to generic `Dataset` easyblock

VRehnberg:
Some datasets can have a very large number of files. Symlinking each individual file can be costly during installation, but especially concerning is that it might be really tank the read performance on some file systems.

This commit introduces a variable for the dataset easyblock where you can specify which directories to ignore when creating the object storage files.

An alternative could be to put the entire directory into object storage.

I should also add that I haven't benchmarked how much of a performance hit this actually is, its just what I suspect at the moment.

### DIFF
--- a/easybuild/easyblocks/generic/dataset.py
+++ b/easybuild/easyblocks/generic/dataset.py
@@ -48,7 +48,8 @@ class Dataset(Binary):
         extra_vars.update({
             'extract_sources': [True, "Whether or not to extract data sources", CUSTOM],
             'data_install_path': [None, "Custom installation path for datasets", CUSTOM],
-            'cleanup_data_sources': [False, "Whether or not to delete the data sources after installation", CUSTOM]
+            'cleanup_data_sources': [False, "Whether or not to delete the data sources after installation", CUSTOM],
+            'object_storage_ignore_dirs': [None, "Directories to be excluded from object storage", CUSTOM],
         })
         return extra_vars
 
@@ -78,7 +79,7 @@ class Dataset(Binary):
         change_dir(self.installdir)
         object_storage = os.path.normpath(os.path.join(os.getcwd(), os.pardir, 'object_storage'))
 
-        datafiles = create_index(os.curdir)
+        datafiles = create_index(os.curdir, ignore_dirs=self.cfg['object_storage_ignore_dirs'])
 
         for datafile in datafiles:
             cks = compute_checksum(datafile, checksum_type='sha256')
--- a/easybuild/easyblocks/generic/dataset.py
+++ b/easybuild/easyblocks/generic/dataset.py
@@ -49,7 +49,7 @@ class Dataset(Binary):
             'extract_sources': [True, "Whether or not to extract data sources", CUSTOM],
             'data_install_path': [None, "Custom installation path for datasets", CUSTOM],
             'cleanup_data_sources': [False, "Whether or not to delete the data sources after installation", CUSTOM],
-            'object_storage_ignore_dirs': [None, "Directories to be excluded from object storage", CUSTOM],
+            'object_storage_ignore_dirs': [[], "List of directories to be excluded from object storage", CUSTOM],
         })
         return extra_vars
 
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/generic/dataset.py
```diff
@@ -78,7 +79,7 @@ def post_processing_step(self):
         change_dir(self.installdir)
         object_storage = os.path.normpath(os.path.join(os.getcwd(), os.pardir, 'object_storage'))
 
-        datafiles = create_index(os.curdir)
+        datafiles = create_index(os.curdir, ignore_dirs=self.cfg['object_storage_ignore_dirs'])
```
smoors: this doesn't work if the directory to ignore is the installdir itself (i.e. ignore everything)
we can special-case this case? something like this
```python
if object_storage_ignore_dirs == self.installdir:
    datafiles = []
else:
    datafiles = create_index(os.curdir, ignore_dirs=self.cfg['object_storage_ignore_dirs'])

**Comment by smoors** [comment]
> Some datasets can have a very large number of files. Symlinking each individual file can be costly during installation, but especially concerning is that it might be really tank the read performance on some file systems.
> 
> This commit introduces a variable for the dataset easyblock where you can specify which directories to ignore when creating the object storage files.
> 
> An alternative could be to put the entire directory into object storage.
> 
> I should also add that I haven't benchmarked how much of a performance hit this actually is, its just what I suspect at the moment.

we could also add an option to create hardlinks instead of symlinks?

**Comment by VRehnberg** [comment]
@smoors I added a suggestion for hardlinking instead, but would need to add hardlink capabilities to framework for that and don't have an easy way to test and see if it works.

**Comment by VRehnberg** [comment]
Åke had some [concerns](https://github.com/easybuilders/easybuild-framework/pull/5200#issuecomment-4730680859) about hardlinks so this might still be something to complement it when hardlinks aren't supported on the filesystem.

File: easybuild/easyblocks/generic/dataset.py [hunk 2]
```diff
@@ -48,7 +48,8 @@ def extra_options(extra_vars=None):
         extra_vars.update({
             'extract_sources': [True, "Whether or not to extract data sources", CUSTOM],
             'data_install_path': [None, "Custom installation path for datasets", CUSTOM],
-            'cleanup_data_sources': [False, "Whether or not to delete the data sources after installation", CUSTOM]
+            'cleanup_data_sources': [False, "Whether or not to delete the data sources after installation", CUSTOM],
+            'object_storage_ignore_dirs': [None, "Directories to be excluded from object storage", CUSTOM],
```
smoors:
```suggestion
            'object_storage_ignore_dirs': [[], "List of directories to be excluded from object storage", CUSTOM],
```

easybuild/easyblocks/generic/dataset.py
```diff
@@ -78,7 +79,10 @@ def post_processing_step(self):
         change_dir(self.installdir)
         object_storage = os.path.normpath(os.path.join(os.getcwd(), os.pardir, 'object_storage'))
 
-        datafiles = create_index(os.curdir)
+        if self.cfg['object_storage_ignore_dirs'] and self.installdir in self.cfg['object_storage_ignore_dirs']:
```
smoors: add a type check?

```suggestion
        ignore_dirs = self.cfg['object_storage_ignore_dirs']
        if not isinstance(ignore_dirs, (list, tuple)):
            raise EasyBuildError("Incorrect value type for object_storage_ignore_dirs, should be list or tuple: ",
                                 ignore_dirs)

        if ignore_dirs and self.installdir in ignore_dirs:
```

easybuild/easyblocks/generic/dataset.py
```diff
@@ -78,7 +79,10 @@ def post_processing_step(self):
...
+            datafiles = []
+        else:
+            datafiles = create_index(os.curdir, ignore_dirs=self.cfg['object_storage_ignore_dirs'])
```
smoors:
```suggestion
            datafiles = create_index(os.curdir, ignore_dirs=ignore_dirs)
```

smoors:
@VRehnberg i fixed a bug in the error msg.
also moved the type check to the init phase so we error out more quickly, and changed `installdir` to `'.'` to align with the relative directories that should be given to `object_storage_ignore_dirs`.

let me know if you agree with those changes?

VRehnberg:
Thanks. Looks good to me.

VRehnberg:
I was debating comparing resolved paths against installdir, but I think this is good enough.

smoors:
Test report by @smoors
 * **SUCCESS** _RFdiffusion-schedules-1.1.0.eb_ 
  

Build succeeded for 1 out of 1 (total: 4 secs) (1 easyconfigs in total)
node700.hydra.os - Linux Rocky Linux 9.7, x86_64, AMD EPYC 9535 64-Core Processor (zen5), Python 3.9.25
See https://gist.github.com/smoors/ccedcfcab00ad4a1195c1c6658bd5053 for a full test report.

smoors (approved):
lgtm

smoors:
Going in, thanks @VRehnberg!

PR #4162, opened 2026-06-17
Add extra variable hardlink_object_storage to generic `Dataset` easyblock

VRehnberg:
Alternative to https://github.com/easybuilders/easybuild-easyblocks/pull/4154


VRehnberg:
Successful build on a modified CIFAR-10-data easyconfig with this easyblock + the easyblock from #4141

Edit: i.e. modified to use hardlink_object_storage = True and I checked that symlinks weren't used

PR #4178, opened 2026-07-15
Add ignore patterns for tests on RISC-V to custom easyblock for LLVM

julianmorillo:


boegel:
@bedroge Can you take a look at this, you're probably more aware of the effort behind this than I am...

boegel:
@Crivella Or maybe you're a better fit to review this, since it's quite specific to LLVM...

easybuild/easyblocks/l/llvm.py
```diff
@@ -791,59 +792,137 @@ def _update_test_ignore_patterns(self):
 
         # Some extra tests need to be ignored for RISC-V
         if get_cpu_architecture() == RISCV64:
+            # This test assumes a native x86 runtime environment
```
Crivella: I am mainly wondering why we are removing some ignored tests from the version check block (instead of bumping up the version in which we know they are expected to fail to `< 24`)  and also adding the new ones outside of it.

Is there clearly no intention for these to ever be fixed upstream?

julianmorillo: I hope they will fix those tests upstream at some point. 
Anyway, you are right with your comment so I added a check block for version < 24.

Crivella:
I am in general fine with the current changes i guess we are going to need a test report on a RISCV system

boegel:
> I am in general fine with the current changes i guess we are going to need a test report on a RISCV system

@bedroge @julianmorillo Can you look into that?

julianmorillo:
This PR (together with https://github.com/easybuilders/easybuild-easyconfigs/pull/26078) has a SUCCESS build in https://github.com/EESSI/dev.eessi.io-riscv/pull/123

Crivella:
K gave it a more detail check comparing the output of the old and new function in case of `version < '22'`

The new one is missing
`llvm-libc++-shared.cfg.in :: std/utilities/variant/variant.visit/visit_return_type.pass.cpp`

would this cause trouble with building an older LLVM?

julianmorillo:
It has been moved to https://github.com/easybuilders/easybuild-easyblocks/pull/4178/changes#diff-9560297abd14bd1649f4ea97ffc545517160060a13db3b3194ac97a3c2541049R876, protected under `version < '24'` so it still applies for older versions.

Apart from that, @bedroge is doing a build of version 21.1.8 using this PR in https://github.com/EESSI/software-layer/pull/1593/changes that tests this new version with an older version of LLVM.

Crivella:
> Apart from that, @bedroge is doing a build of version 21.1.8 using this PR in https://github.com/EESSI/software-layer/pull/1593/changes that tests this new version with an older version of LLVM.

Once we have the report for that i think we are good to morge

PR #4181, opened 2026-07-15
fix indentation in sanity check of custom easyblock for AEDT

ryan-maher:
Fixes issue #4180

### DIFF
--- a/easybuild/easyblocks/a/aedt.py
+++ b/easybuild/easyblocks/a/aedt.py
@@ -160,4 +160,4 @@ class EB_AEDT(PackedBinary):
             inpfile = os.path.join(tempdir, 'sm-1.aedt')
             custom_commands = ['ansysedt -ng -batchsolve -Distributed -monitor %s' % inpfile]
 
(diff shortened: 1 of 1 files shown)


boegel:
Looks like this was accidentally introduced in EasyBuild v5.1.0 through 85e35d872867afd1a20f0667af4d791afde07cd1, which was part of a large (and thus hard to test) cleanup PR:
- https://github.com/easybuilders/easybuild-easyblocks/pull/3681

The fact that AEDT is commercial software (so installation files are not publicly available) also doesn't help with being able to test this.

Fix looks fine to me, but I can't verify it myself since I don't have access to AEDT.

However, since this essentially rolls back a change that was probably made by accident, I think it's fine to go ahead and merge this

cc @Micket @appolloford (who are probably able to verify the fix)

boegel (approved):
lgtm

appolloford:
A user asks for installing AEDT 2026R1 today so I also run the test. I can verify the fix is correct.

PR #4173, opened 2026-06-25
fix OpenMPI and UCX easyblocks w.r.t. ROCm support

zerefwayne:
For UCX, see the comment: https://github.com/easybuilders/easybuild-easyconfigs/pull/25902#discussion_r3473153782

For OpenMPI fixes, since UCC-ROCm and UCX-ROCm are now built as plugins to UCC and UCX, respectively, we can rely on the existing ucc and ucx in the known_dependencies. We only need to redirect HIP to --with-rocm flag.

PR #4187, opened 2026-07-27
Add UCX sanity checks only when building with UCX

zerefwayne:


zerefwayne:
Closing in favour of https://github.com/easybuilders/easybuild-easyblocks/pull/4173

PR #4193, opened 2026-08-06
Add TensorFlow easyblock options for Bazel startup flags and TF_SYSTEM_LIBS exclusions

bedroesb:
This PR moves certain patches from https://github.com/easybuilders/easybuild-easyconfigs/pull/26602 into the TensorFlow easyblock.

Changes included here:

- add `bazel_startup_opts`, so easyconfigs can pass Bazel startup options such as `--batch` without editing `.bazelrc`;
- add `tf_system_libs_exclude`, so easyconfigs can opt out of specific `TF_SYSTEM_LIBS` entries without patching `.tf_configure.bazelrc`;
- skip `absl_py` automatically when TensorFlow’s `tf_http_archive` entry has no `system_build_file`;
- pass the temporary TensorFlow home directory to Bazel tests via `--test_env=HOME=...`;
- adjust TensorBoard sanity checking so an external `tensorboard` dependency is accepted instead of requiring `bin/tensorboard` inside the TensorFlow installation prefix.

The `absl_py` handling is needed for TensorFlow 2.21.0 because `absl_py` is still listed in TensorFlow’s `VALID_LIBS`, but its `tf_http_archive` entry no longer has a `system_build_file`. Without this, Bazel fails with:

```text
ERROR: /tmp/bert.droesbeke/easybuild-tensorflow-build/TensorFlow/2.21.0/foss-2025b/TensorFlow/tensorflow-2.21.0/WORKSPACE:100:14: fetching _tf_http_archive rule //external:absl_py: java.io.IOException: _tf_http_archive rule //external:absl_py must create a directory
ERROR: no such package '@@absl_py//absl/flags': _tf_http_archive rule //external:absl_py must create a directory
ERROR: /tmp/bert.droesbeke/easybuild-tensorflow-build/TensorFlow/2.21.0/foss-2025b/TensorFlow/tensorflow-2.21.0/tensorflow/python/tools/api/generator2/generator/BUILD:10:22: //tensorflow/python/tools/api/generator2/generator:generator depends on @@absl_py//absl/flags:flags in repository @@absl_py which failed to fetch. no such package '@@absl_py//absl/flags': _tf_http_archive rule //external:absl_py must create a directory
```

The `tf_system_libs_exclude` option is needed because some TensorFlow system-library stubs are not sufficient for all Bazel compile actions. For TensorFlow 2.21.0, enabling system `boringssl` made gRPC pick up OpenSSL headers via Bazel actions, which failed with:

```text
ERROR: /tmp/bert.droesbeke/easybuild-tensorflow-build/TensorFlow/2.21.0/foss-2025b/TensorFlow/bazel-root/29cea7ecf482505435700a02a31d6d02/external/com_github_grpc_grpc/BUILD:4219:16: Compiling src/core/tsi/ssl/session_cache/ssl_session_boringssl.cc failed: (Exit 1): clang failed: error executing CppCompile command (from target @@com_github_grpc_grpc//:tsi_ssl_session_cache)
external/com_github_grpc_grpc/src/core/tsi/ssl/session_cache/ssl_session.h:24:10: error: module com_github_grpc_grpc//:tsi_ssl_session_cache does not depend on a module exporting 'openssl/ssl.h'
/mnt/modules/easybuild/software/OpenSSL/3/include/openssl/ssl.h:29:10: fatal error: 'openssl/e_ostime.h' file not found
```

While validating the cleanup, using system cURL showed the same general problem with exposing dependency headers globally through `CPATH`:

```text
ERROR: /tmp/bert.droesbeke/easybuild-tensorflow-build/TensorFlow/2.21.0/foss-2025b/TensorFlow/bazel-root/29cea7ecf482505435700a02a31d6d02/external/xla/xla/tsl/platform/cloud/BUILD:219:11: Compiling xla/tsl/platform/cloud/curl_http_request.cc failed: absolute path inclusion(s) found in rule '@@xla//xla/tsl/platform/cloud:curl_http_request':
the source file 'xla/tsl/platform/cloud/curl_http_request.cc' includes the following non-builtin files with absolute paths:
  '/data/groups/technologies/data.core/bert.droesbeke/easybuild/software/cURL/8.14.1-GCCcore-14.3.0/include/curl/curl.h'
```

With these easyblock changes, the TensorFlow 2.21.0 easyconfig no longer needs the previous sed/CPATH/TensorBoard-symlink workarounds. The corresponding TensorFlow/2.21.0-foss-2025b build completed successfully.
```

### DIFF
--- a/easybuild/easyblocks/t/tensorflow.py
+++ b/easybuild/easyblocks/t/tensorflow.py
@@ -106,6 +106,22 @@ def get_system_libs_from_tf(source_dir):
     return result
 
 
+def tf_http_archive_has_system_build_file(source_dir, repo_name):
+    """Return whether tf_http_archive definition for repo_name has a system_build_file."""
+    workspace_paths = [
+        os.path.join(source_dir, 'tensorflow', 'workspace2.bzl'),
+        os.path.join(source_dir, 'third_party', 'xla', 'workspace2.bzl'),
+    ]
+    regex = re.compile(r'tf_http_archive\(\s*name\s*=\s*"%s",(.*?)\n\s*\)' % re.escape(repo_name), re.DOTALL)
+    for workspace_path in workspace_paths:
+        if os.path.exists(workspace_path):
+            workspace_txt = read_file(workspace_path)
+            for match in regex.finditer(workspace_txt):
+                if 'system_build_file' in match.group(1):
+                    return True
+    return False
+
+
 def get_system_libs_for_version(tf_version, as_valid_libs=False):
     """
     Determine valid values for $TF_SYSTEM_LIBS for the given TF version
@@ -245,6 +261,12 @@ class EB_TensorFlow(PythonPackage):
                                   "the number of GPUs). Use None (default) to automatically determine a value", CUSTOM],
             'jvm_max_memory': [4096, "Maximum amount of memory in MB used for the JVM running Bazel." +
                                "Use None to not set a specific limit (uses a default value).", CUSTOM],
+            'bazel_startup_opts': [[], "List of extra startup options to pass to Bazel before the command", CUSTOM],
+            'tf_system_libs_exclude': [
+                [],
+                "List of TensorFlow system library names to omit from $TF_SYSTEM_LIBS",
+                CUSTOM,
+            ],
         }
 
         return PythonPackage.extra_options(extra_vars)
@@ -355,6 +377,21 @@ class EB_TensorFlow(PythonPackage):
         Returns a tuple of lists: $TF_SYSTEM_LIBS names, include paths, library paths
         """
         dependency_mapping, python_mapping = get_system_libs_for_version(self.version)
+
+        tf_system_libs_exclude = self.cfg['tf_system_libs_exclude']
+        if isinstance(tf_system_libs_exclude, str):
+            tf_system_libs_exclude = [x.strip() for x in tf_system_libs_exclude.split(',') if x.strip()]
+        tf_system_libs_exclude = set(tf_system_libs_exclude)
+
+        # TensorFlow 2.21 still lists absl_py as a valid TF_SYSTEM_LIBS entry, but its
+        # tf_http_archive definition no longer has a system_build_file. Enabling it makes Bazel create an empty
+        # @absl_py repository and fail with "_tf_http_archive rule //external:absl_py must create a directory".
+        if python_mapping.get('absl') == 'absl_py' and not tf_http_archive_has_system_build_file(
+            self.start_dir, 'absl_py'
+        ):
+            self.log.info("Not using system absl_py: TensorFlow's @absl_py repository has no system_build_file")
+            del python_mapping['absl']
+
         # Some TF dependencies require both a (usually C++) dependency and a Python package
         deps_with_python_pkg = {tf_name for tf_name in dependency_mapping.values()
                                 if tf_name in python_mapping.values()}
@@ -363,6 +400,15 @@ class EB_TensorFlow(PythonPackage):
         cpaths = []
         libpaths = []
         ignored_system_deps = []
+
+        def add_system_lib(tf_name):
+            """Add a TensorFlow system library unless it was explicitly excluded."""
+            if tf_name in tf_system_libs_exclude:
+                self.log.info("Not using system %s: listed in tf_system_libs_exclude", tf_name)
+                return False
+            if tf_name not in system_libs:
+                system_libs.append(tf_name)
+            return True
 
         # Check direct dependencies
         dep_names = self.cfg.dependency_names()
@@ -374,11 +420,12 @@ class EB_TensorFlow(PythonPackage):
                     # Simply ignore. Error reporting is done in the other loop
                     if not self.python_pkg_exists(pkg_name):
                         continue
-                system_libs.append(tf_name)
+                if not add_system_lib(tf_name):
+                    continue
                 # When using cURL (which uses the system OpenSSL), we also need to use "boringssl"
                 # which essentially resolves to using OpenSSL as the API and library names are compatible
                 if dep_name == 'cURL':
-                    system_libs.append('boringssl')
+                    add_system_lib('boringssl')
                 sw_root = get_software_root(dep_name)
                 # Dependency might be filtered via --filter-deps. In that case assume globally installed version
                 if not sw_root:
@@ -407,21 +454,20 @@ class EB_TensorFlow(PythonPackage):
             if self.python_pkg_exists(pkg_name):
                 # If it is in deps_with_python_pkg we already added it
                 if tf_name not in deps_with_python_pkg:
-                    system_libs.append(tf_name)
+                    add_system_lib(tf_name)
             else:
                 ignored_system_deps.append('%s (Python package %s)' % (tf_name, pkg_name))
 
         # If we use OpenSSL (potentially as a wrapper) somewhere in the chain we must tell TF to use it too
         openssl_root = get_software_root('OpenSSL')
         if openssl_root:
-            if 'boringssl' not in system_libs:
-                system_libs.append('boringssl')
-            incpath = os.path.join(openssl_root, 'include')
-            if os.path.exists(incpath):
-                cpaths.append(incpath)
-            libpath = get_software_libdir('OpenSSL')
-            if libpath:
-                libpaths.append(os.path.join(openssl_root, libpath))
+            if add_system_lib('boringssl'):
+                incpath = os.path.join(openssl_root, 'include')
+                if os.path.exists(incpath):
+                    cpaths.append(incpath)
+                libpath = get_software_libdir('OpenSSL')
+                if libpath:
+                    libpaths.append(os.path.join(openssl_root, libpath))
 
         if ignored_system_deps:
             print_warning('%d TensorFlow dependencies have not been resolved by EasyBuild. '
@@ -882,6 +928,11 @@ class EB_TensorFlow(PythonPackage):
         if bazel_version >= '4.0.0':
             self.bazel_opts.append('--local_startup_timeout_secs=300')  # 5min
 
+        bazel_startup_opts = self.cfg['bazel_startup_opts']
+        if isinstance(bazel_startup_opts, str):
+            bazel_startup_opts = bazel_startup_opts.split()
+        self.bazel_opts.extend(bazel_startup_opts)
+
         # Environment variables and values needed for Bazel actions.
         action_env = {}
         # A value of None is interpreted as using the invoking environments value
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by Flamefire** [comment]
Before I jump into the implementation (which looks ok) I'd like to discuss the changes

> * add `bazel_startup_opts`, so easyconfigs can pass Bazel startup options such as `--batch` without editing `.bazelrc`;

Good idea, I'd say as an easy way to avoid modifying the easyblock. I'd expect you'd want `--batch` at least starting at a given Bazel version. Does it has any downsides or shall we always use it? Then we should add it here instead of in each easyconfig.

> runs Bazel in batch mode to avoid lock/server startup issues

How does this look like and how does --batch avoids it?

> * add `tf_system_libs_exclude`, so easyconfigs can opt out of specific `TF_SYSTEM_LIBS` entries without patching `.tf_configure.bazelrc`;

I'd say this is not a good idea: Only direct dependencies are considered. So instead of excluding it you can simply **not add** it, can't you? Otherwise you would be mixing EasyBuild dependencies and downloaded/vendored software which likely leads to issues.


> * skip `absl_py` automatically when TensorFlow’s `tf_http_archive` entry has no `system_build_file`;

Similar to above: If it doesn't work (and we can't patch it) remove the dependency from the easyconfig to avoid conflicts.
 
> * adjust TensorBoard sanity checking so an external `tensorboard` dependency is accepted instead of requiring `bin/tensorboard` inside the TensorFlow installation prefix.

Good idea but I'd do it differently: It looks like TensorFlow did install tensorboard and since some version does it. So instead guard it on the version that introduced the change so you are not required to include tensorboard as a dependency of TensorFlow.

> enabling system boringssl made gRPC pick up OpenSSL headers via Bazel actions, which failed with:

Not enabling it mixes the SSL installations leading to issues again. It looks rather like a problem in OpenSSL:
> /mnt/modules/easybuild/software/OpenSSL/3/include/openssl/ssl.h:29:10: fatal error: 'openssl/e_ostime.h' file not found

--> OpenSSL header includes an OpenSSL header that doesn't exist?

> the source file 'xla/tsl/platform/cloud/curl_http_request.cc' includes the following non-builtin files with absolute paths:
  '/data/groups/technologies/data.core/bert.droesbeke/easybuild/software/cURL/8.14.1-GCCcore-14.3.0/include/curl/curl.h'

This looks like something we should fix too: We do want to use our cURL to avoid conflicts. Maybe this is caused by the filtering of CPATH you mentioned? Maybe there is some other option. ChatGPT mentions using `build --copt=-isystem=/data/cURL/8.14.1-GCCcore-14.3.0/include` etc to the bazelrc file instead of `--action-env=CPATH=...` that we currently do. Might be worth a try. Currently can't because our cluster is in maintenance

**Comment by bedroesb** [comment]
Thanks for the detailed feedback:

Based on the suggestions I could cleanup quite a bit:

- removed the generic `tf_system_libs_exclude` option;
- stopped auto-enabling `boringssl` merely because OpenSSL is present somewhere in the dependency tree;
- `boringssl` is now only added when `cURL` is a direct TensorFlow dependency, matching the existing cURL-specific rationale;
- made `--batch` automatic for Bazel >= 7.0.0, while keeping `bazel_startup_opts` only as an escape hatch for extra startup options;
- changed TensorBoard sanity checking to be version-based for TensorFlow >= 2.21, so TensorFlow is no longer expected to install `bin/tensorboard`;
- kept the `absl_py` check because `absl-py` is still a required Python dependency, but TensorFlow 2.21 cannot use it as a `TF_SYSTEM_LIBS` repository since its `tf_http_archive` entry has no `system_build_file`.

For `absl_py`, the exact failure was:

```text
ERROR: .../WORKSPACE:100:14: fetching _tf_http_archive rule //external:absl_py:
java.io.IOException: _tf_http_archive rule //external:absl_py must create a directory
ERROR: no such package '@@absl_py//absl/flags': _tf_http_archive rule //external:absl_py must create a directory

I will go on holiday today, so feel free to change things without waiting on a response ;)

**Comment by Flamefire** [comment]
>     * stopped auto-enabling `boringssl` merely because OpenSSL is present somewhere in the dependency tree;

Why? If we have both and not using rpath the 2 will conflict, won't they?
The failure you see is because of use of system OpenSSL (via wrappers) and can be fixed by rebuilding the OpenSSL module. I'll add an update to the easyblock to avoid that in the future.

>     * kept the `absl_py` check because `absl-py` is still a required Python dependency, but TensorFlow 2.21 cannot use it as a `TF_SYSTEM_LIBS` repository since its `tf_http_archive` entry has no `system_build_file`.

Why is it a required Python dependency?
I do think the check is useful but more generic: The easyblock contains a version check for system deps, so we could state it as supported only until 2.21.
The check for system build file could then be done in the general "is systemlibs up to date, or update easyblock" function.

> For `absl_py`, the exact failure was:
> 
> ```
> ERROR: .../WORKSPACE:100:14: fetching _tf_http_archive rule //external:absl_py:
> java.io.IOException: _tf_http_archive rule //external:absl_py must create a directory
> ERROR: no such package '@@absl_py//absl/flags': _tf_http_archive rule //external:absl_py must create a directory

Just reproduced that. I'm testing a patch as the removal seems accidental by upstream.

**Comment by eylenth** [comment]
I'm now also trying to build TensorFlow with the CUDA variant ( I will create a PR for the easyconfig once it is working).
I'm testing out this easyblock as well, but I'm getting the following error message in the build step of bazel:
```sh
ERROR: /tools/eb/build/TensorFlow/2.21.0/foss-2025b-CUDA-12.9.1/TensorFlow/bazel-root/c741a8d54995072d0506b6004cf6ee2e/external/curl/BUILD.bazel:26:11: Compiling lib/altsvc.c failed: absolute path inclusion(s) found in rule '@@curl//:curl':
the source file 'lib/altsvc.c' includes the following non-builtin files with absolute paths (if these are builtin files, make sure these paths are in your toolchain):
  '/cvmfs/biogrid-software.rnd.seeds.basf.net/tools/eb/x86_64/generic/software/OpenSSL/3/include/openssl/opensslv.h'
  '/cvmfs/biogrid-software.rnd.seeds.basf.net/tools/eb/x86_64/generic/software/OpenSSL/3/include/openssl/macros.h'
  '/cvmfs/biogrid-software.rnd.seeds.basf.net/tools/eb/x86_64/generic/software/OpenSSL/3/include/openssl/opensslconf.h'
  '/cvmfs/biogrid-software.rnd.seeds.basf.net/tools/eb/x86_64/generic/software/OpenSSL/3/include/openssl/configuration.h'
  '/cvmfs/biogrid-software.rnd.seeds.basf.net/tools/eb/x86_64/generic/software/OpenSSL/3/include/openssl/configuration-x86_64.h'
```
I tried to add the `('cURL', '8.14.1')` as dependency in the easyconfig recipe of TensorFlow, But then I'm getting another error message again.
I'm not using rpath btw.

Maybe this could be useful by mentioning this.

Flamefire:
> I'm testing out this easyblock as well, but I'm getting the following error message in the build step of bazel:

That's (kind of) expected. The change here to not use OpenSSL/BoringSSL creates this "conflict". Adding cURL is correct anyway. Working on a solution for related fixes.

Flamefire:
I made significant progress here after running into undeclared inclusions, (C++) module validation issues and more

A short test using system/our cURL can be done by passing the `@xla//xla/tsl/platform/cloud:curl_http_request` target instead of `//tensorflow/tools/pip_package:wheel` to `bazel`

This currently yields:
```
ERROR: /dev/shm/TensorFlow/2.21.0/foss-2025b/TensorFlow/bazel-root/38526f3dacc6677801c11293634cdb7f/external/xla/xla/tsl/platform/cloud/BUILD:219:11: Compiling xla/tsl/platform/cloud/curl_http_request.cc failed: absolute path inclusion(s) found in rule '@@xla//xla/tsl/platform/cloud:curl_http_request':
the source file 'xla/tsl/platform/cloud/curl_http_request.cc' includes the following non-builtin files with absolute paths (if these are builtin files, make sure these paths are in your toolchain):
  '/software/cURL/8.14.1-GCCcore-14.3.0/include/curl/curl.h'
...
Target @@xla//xla/tsl/platform/cloud:curl_http_request failed to build
```

After lots of debugging and chatting with AI agents I stumbled upon a suspicious toolchain being used and `git blame`d it to: https://github.com/tensorflow/tensorflow/commit/f7dfba3f742c (first released in 2.21)

It mentions the solution `--config=clang_local`

Another failure in the full build was:
```
In file included from external/com_github_grpc_grpc/src/core/tsi/ssl/session_cache/ssl_session_boringssl.cc:21:
external/com_github_grpc_grpc/src/core/tsi/ssl/session_cache/ssl_session.h:24:10: error: module com_github_grpc_grpc//:tsi_ssl_session_cache does not depend on a module exporting 'openssl/ssl.h'
   24 | #include <openssl/ssl.h>
```

This is that C++ module checker. I found that this can be disable by passing `--features=-layering_check` but still testing if we still need that after the above.

Currently that solution causes other issues: They assume their toolchain is used and pass flags accordingly causing e.g.
> gcc: error: unrecognized command-line option '-fno-experimental-sanitize-metadata=all'

I'm not sure if we could use their toolchain, although they pass our flags like `-march=native` to that compiler I see `--sysroot external/sysroot_linux_x86_64_glibc_2_27`
I'm afraid that will bite us at least with EESSI. And of course it circumvents our rpath wrappers

easybuild/easyblocks/t/tensorflow.py
```diff
@@ -355,6 +372,16 @@ def get_system_libs(self):
         Returns a tuple of lists: $TF_SYSTEM_LIBS names, include paths, library paths
         """
         dependency_mapping, python_mapping = get_system_libs_for_version(self.version)
+
+        # TensorFlow 2.21 still lists absl_py as a valid TF_SYSTEM_LIBS entry, but its
```
Flamefire, line 376: This can be fixed by a patch so does not need to be in the easyblock, see https://github.com/easybuilders/easybuild-easyconfigs/pull/26658/changes#diff-01cb94902867bc2e7ffd436faf712aca6604eb5ea903030bf45a6f5eb49f9792

Flamefire:
I created #4205 which does what is required for TF 2.19 and 2.21. I think here the parts to be kept are the bazel startup options with the `--batch` default. Not critical though

bedroesb:
@Flamefire I added you to my fork, but if you prefer I can update this PR or remove it if deemed redundant!

Flamefire:
You can close this one. What was left were  bazel startup options with the `--batch` default.
I added your commits for (only) that to my larger PR at https://github.com/easybuilders/easybuild-easyblocks/pull/4205

That should be the easiest way and avoids testing 2 PRs

Let me know if you see anything missing that should still be included

bedroesb:
Closing since change is part of https://github.com/easybuilders/easybuild-easyblocks/pull/4205

PR #4195, opened 2026-08-08
don't check package names and versions with pip list if `sanity_check_pip_list` is False

smoors:


smoors:
EB_ARGS="--sanity-check-only PyTorch-2.6.0-foss-2024a.eb"

easybuild/easyblocks/generic/pythonpackage.py
```diff
@@ -520,7 +520,8 @@ def extra_options(extra_vars=None):
                                        "installed and check for any package with an invalid (0.0.0) version.", CUSTOM],
             'sanity_check_pip_list': [None, "Fail if specified package names and versions do not match "
                                             "'python -m pip list' output. Defaults to True if --upload-test-report is "
-                                            "set. The check only runs if 'sanity_pip_check' is True.", CUSTOM],
+                                            "set. The check only runs if 'sanity_pip_check' is True and "
+                                            "'sanity_check_pip_list' is not False.", CUSTOM],
```
Flamefire: The last part seems odd as it refers to this very option.

I.e. why say that it gets disabled by setting it to False, when we have "Defaults to True"?

And don't we have:
- Default to True
- Run sanity_pip_check=True
- Fail build if failing with --upload-test-report, warn otherwise

The description doesn't reflect that, does it?

PR #4206, opened 2026-08-14
Add ICON easyblock

JarneRenders:
Move the configopts for ICON into an easyblock. In particular this is necessary for nicely including the CUDAFLAGS:

```
-gencode arch=compute_XX code=sm_XX
```
for each cuda compute capability. See easybuilders/easybuild-easyconfigs#26615.

Claude (Sonnet 5) was used to write and iterate on this easyblock under my direction and review; I verified claims against ICON's own configure.ac/autoconf macros and NVIDIA's nvcc documentation and ran all three (NVHPC/foss/intel) builds to completion to confirm correctness.

### DIFF
--- a/easybuild/easyblocks/i/icon.py
+++ b/easybuild/easyblocks/i/icon.py
@@ -6,31 +6,184 @@ EasyBuild support for ICON, implemented as an easyblock
 from easybuild.easyblocks.generic.configuremake import ConfigureMake
 from easybuild.tools import toolchain
 from easybuild.tools.build_log import EasyBuildError
+from easybuild.tools.modules import get_software_root
 
 
 class EB_ICON(ConfigureMake):
     """Support for building/installing ICON."""
 
-    def configure_step(self):
-        """Configure ICON build, setting toolchain-specific compiler flags."""
-        comp_fam = self.toolchain.comp_family()
+    # Dependencies providing the libraries referenced in LIBS, common to every toolchain
+    REQUIRED_DEPS = ['libxml2', 'libfyaml', 'ecCodes', 'netCDF', 'netCDF-Fortran']
 
-        self.cfg.update('configopts', 'MPI_LAUNCH=mpirun')
+    def _check_dependencies(self):
+        """Make sure the dependencies backing LIBS are actually loaded."""
+        for dep_name in self.REQUIRED_DEPS:
+            if not get_software_root(dep_name):
+                raise EasyBuildError("ICON requires %s as a dependency to link against, but it is not loaded.",
+                                     dep_name)
 
-        if comp_fam in (toolchain.GCC, toolchain.INTELCOMP):
-            # LIBS is identical for foss and intel (no CUDA libraries, both need
-            # -Wl,--disable-new-dtags); NVHPC sets its own LIBS below.
+    def _add_model_features(self, comp_fam):
+        """Model Features."""
+        self.cfg.update('configopts', (
+            '--enable-atmo '         # atmosphere component [default=yes]
+            '--enable-les '          # Large-Eddy Simulation component [default=yes]
+            '--enable-upatmo '       # upper atmosphere component [default=yes]
+            '--enable-ocean '        # ocean component [default=yes]
+            '--enable-jsbach '       # land component JSBACH [default=yes]
+            '--enable-jsbach-hd '    # hydrological discharge scheme of the JSBACH land component [default=auto]
+            '--enable-coupling '     # coupling [default=yes]
+            '--enable-aes '          # AES physics package [default=yes]
+            '--enable-nwp '          # NWP physics package [default=yes]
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/i/icon.py
```diff
@@ -0,0 +1,97 @@
...
+
+
+class EB_ICON(ConfigureMake):
+    """Support for building/installing ICON."""
+
+    def configure_step(self):
+        """Configure ICON build, setting toolchain-specific compiler flags."""
+        comp_fam = self.toolchain.comp_family()
+
+        self.cfg.update('configopts', 'MPI_LAUNCH=mpirun')
```
Crivella: we might want to have this in a check like
```suggestion
        if self.toolchain.mpi_family():
            self.cfg.update('configopts', 'MPI_LAUNCH=mpirun')
        else:
            raise EasyBuildError("ICON requires an mpi-enabled toolchain.")
```

easybuild/easyblocks/i/icon.py
```diff
@@ -0,0 +1,97 @@
...
+                'LIBS="'
+                '    -Wl,--disable-new-dtags -Wl,--as-needed'
+                '    -lxml2'
+                '    -lfyaml'
+                '    -leccodes_f90 -leccodes'
+                '    $LIBLAPACK'
+                '    -lnetcdff -lnetcdf'
+                '    -lstdc++'
+                '"'
+            ))
```
Crivella: I think this can be confusing for people touching the EB after.
Here i would do something like

```
gcc_libs = ...
intel_libs = gcc_libs
nvhpc_libs = ...
```

and then do a `self.cfg.update` in the respective if blocks so that we have only one `if/elseif/else` chahin

Crivella: Also since we are writing an EB we can also add checks so that the required dependencies for the libraries specified here are actually loaded

More generically we should bring all the options that are now in the EC here and enable them when the appropriate dependency or toolchain + toolchain options are set

easybuild/easyblocks/i/icon.py
```diff
@@ -0,0 +1,97 @@
...
+                f'    -acc=gpu,verystrict -gpu={cuda_sm}"'
+            ))
+
+            # nvcc's plain -arch= flag only accepts a single target (unlike NVHPC's -gpu=
+            # flag above), so build a real multi-arch fat binary via -gencode instead.
+            cuda_ccs = self.cfg.get_cuda_cc_template_value('cuda_int_comma_sep')
+            nvcc_gencode = ' '.join(
+                f'-gencode arch=compute_{cc},code=sm_{cc}' for cc in cuda_ccs.split(',')
+            )
+            self.cfg.update('configopts', f'CUDAFLAGS="-ccbin=$CXX -O3 {nvcc_gencode}"')
```
Crivella, line 203: Here and also for the other places i think we should try to respect the optimization flags passed by the TChain

Crivella (changes requested):

easybuild/easyblocks/i/icon.py
JarneRenders: It is now guarded with such a check

easybuild/easyblocks/i/icon.py
JarneRenders: Moved all configopts to the easyblock. Checking the necessary dependencies are present for the libraries. Also checking now for dependencies which would replace the installation via git submodules (one slightly strange thing here is that I had to infer the names of such a dependency even though no easyconfigs for those exist yet.) Is this what you had in mind?

easybuild/easyblocks/i/icon.py
JarneRenders: Passed them in the other places, but not here as I am not sure we have appropriate CUDAFLAGS for nvcc anywhere in easybuild. (CFLAGS is only compatible with nvc and not with nvcc.)

easybuild/easyblocks/i/icon.py
```diff
@@ -1,7 +1,7 @@
 """
 EasyBuild support for ICON, implemented as an easyblock
 
-@author: Jarne Renders (Vrije Universiteit Brussel)
+@author: Jarne Renders (Vrije Universiteit Brussel) with Sonnet 5
```
Crivella, line 4: Could you add a statement on AI usage to the PR description

easybuild/easyblocks/i/icon.py
Crivella: We do in https://github.com/easybuilders/easybuild-framework/blob/develop/easybuild/toolchains/compiler/cuda.py, but we might need to initialize a separate TC inside the EB to extract this info, i am wondering if this should not somehow be part of the `nvidia-compilers` TC as well

JarneRenders: Yes, that makes more sense to me to add it to nvidia-compilers.

PR #4204, opened 2026-08-12
enhance `binutils` easyblock to ensure system search path is used for `ld` instead of the default one by setting `$LIB_PATH` in build step

Crivella:
This should fix the problem observed in https://github.com/easybuilders/easybuild-easyconfigs/issues/20375

Currently the `binutils` EBlock determines the system search path through `gcc -print-search-dirs` and uses it to rpath builds in case of a system toolchain.

The problem is that the `SEARCH_DIRS` of `ld` are not properly set to the system ones and the default (on eg Debian/Ubuntu) misses `/usr/lib/x86_64-linux-gnu` and `/lib/x86_64-linux-gnu`.
For `nvidia-compilers` this causes failures when liking eg `libc` `libdl` and `libpthread` through `ld` directly.

This is not a problem in GCC based toolchains as gcc does its own logic to add this paths with `-L` statements
`nvc++` in `nvidia-compilers` instead both adds `-lc -ldl -lpthread` to every invocation and passes it directly to `ld` causing failures of the type

```
/home/crivella/.local/easybuild/software/binutils/2.42-GCCcore-14.2.0/bin/ld: cannot find -ldl: No such file or directory
/home/crivella/.local/easybuild/software/binutils/2.42-GCCcore-14.2.0/bin/ld: cannot find -lpthread: No such file or directory
/home/crivella/.local/easybuild/software/binutils/2.42-GCCcore-14.2.0/bin/ld: cannot find -lc: No such file or directory
```

<details><summary>Details</summary>
<p>

```
/tmp/eb-av4z3kjy/tmpf22k8are$ nvc++ minimal.cpp  -v
Export NVCOMPILER=/home/crivella/.local/easybuild/software/nvidia-compilers/25.3-CUDA-12.8.0/Linux_x86_64/25.3
Export PGI=/home/crivella/.local/easybuild/software/nvidia-compilers/25.3-CUDA-12.8.0

[...]

PR #4209, opened 2026-08-26
Ensure dependencies libpaths are not baked into `binutils`'s ld

Crivella:
Follow up to 

- https://github.com/easybuilders/easybuild-easyblocks/pull/4204

to make sure we are not adding unneeded search paths from the dependencies used to build `binutils`

Technically we could only filter build deps since runtime deps will still need to be accessible, but i think it is better to handle that through the environment variable and rpath logic only

### DIFF
--- a/easybuild/easyblocks/b/binutils.py
+++ b/easybuild/easyblocks/b/binutils.py
@@ -109,7 +109,18 @@ class EB_binutils(ConfigureMake):
         # this should 'harden' the resulting binutils to bootstrap GCC
         # (no trouble when other libstdc++ is build etc)
         lib_paths = self.determine_used_library_paths()
-        self.search_paths = lib_paths.copy()
+
+        # Filter out any search paths from dependencies, to avoid baking non-needed libraries into ld
+        dep_roots = []
+        for dep in self.cfg.dependencies():
+            dep_root = get_software_root(dep['name'])
+            if dep_root:
+                dep_roots.append(dep_root)
+
+        for spath in lib_paths:
+            if not any(spath.startswith(dep_root) for dep_root in dep_roots):
+                self.search_paths.append(spath)
+
         if self.toolchain.is_system_toolchain():
             # The installed lib dir must come first though to avoid taking system libs over installed ones, see:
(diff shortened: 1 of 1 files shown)


Crivella:
Test report by @Crivella
 * **SUCCESS** _binutils-2.42-GCCcore-14.2.0.eb_ 
  
 * **SUCCESS** _nvidia-compilers-25.3-CUDA-12.8.0.eb_ 
  

Build succeeded for 2 out of 2 (total: 8 mins 26 secs) (2 easyconfigs in total)
crivella-desktop - Linux Ubuntu 22.04.5 LTS (Jammy Jellyfish), x86_64, 13th Gen Intel(R) Core(TM) i9-13900K (skylake), 1 x NVIDIA NVIDIA GeForce RTX 5060, 580.173.02, Python 3.11.15
See https://gist.github.com/Crivella/3809a883db4ab27f83803ad12b1f5533 for a full test report.

Crivella:
Current SEARCH_DIRS

```
ld -verbose | grep SEARCH | tr ';' '\n'
SEARCH_DIR("=/home/crivella/.local/easybuild/software/binutils/2.42-GCCcore-14.2.0/x86_64-pc-linux-gnu/lib64")
 SEARCH_DIR("/home/crivella/.local/easybuild/software/GCCcore/14.2.0/lib64")
 SEARCH_DIR("/lib/x86_64-linux-gnu")
 SEARCH_DIR("/lib64")
 SEARCH_DIR("/lib")
 SEARCH_DIR("=/home/crivella/.local/easybuild/software/binutils/2.42-GCCcore-14.2.0/x86_64-pc-linux-gnu/lib")
```

we still get the GCCcore ones but i do not think that is a problem unless we want to use `ld` with a different GCC than the one it was built with

boegel:
@Crivella Do we also want to pull defining of `self.search_paths` out of `configure_step`, so `--sanity-check-only` works fully as expected?

Crivella:
IIRC we do not have the `get_software_root` for the deps available yet at the `__init__` so it would not be feasible if we also want the filtering?
Also not sure the GCC we get before `configure` would necessarily be the correct one (eg EB vs system)

PR #4210, opened 2026-08-27
document extra_option modulename of PythonPackage

Louwrensth:
I also was bothered by
- https://github.com/easybuilders/easybuild-docs/issues/177

### DIFF
--- a/easybuild/easyblocks/generic/pythonpackage.py
+++ b/easybuild/easyblocks/generic/pythonpackage.py
@@ -516,6 +516,9 @@ class PythonPackage(ExtensionEasyBlock):
             'req_py_minver': [None, "Required minor Python version (only relevant when using system Python)", CUSTOM],
             'max_py_majver': [None, "Maximum major Python version (only relevant when using system Python)", CUSTOM],
             'max_py_minver': [None, "Maximum minor Python version (only relevant when using system Python)", CUSTOM],
+            'modulename': [None, "Module name of the Python package, as used in import check. "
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/generic/pythonpackage.py
```diff
@@ -516,6 +516,9 @@ def extra_options(extra_vars=None):
             'req_py_minver': [None, "Required minor Python version (only relevant when using system Python)", CUSTOM],
             'max_py_majver': [None, "Maximum major Python version (only relevant when using system Python)", CUSTOM],
             'max_py_minver': [None, "Maximum minor Python version (only relevant when using system Python)", CUSTOM],
+            'modulename': [None, "Module name of the Python package, as used in import check. "
```
boegel, line 519: Including it here is not fully correct, because `modulename` can only be set for specific *extensions*.

To set it top-level, you currently still need to use:

```python
options = {'modulename': 'example'}
```

This is being changed/fixed in a recent framework PR:
- https://github.com/easybuilders/easybuild-framework/pull/5257

So the change you propose is strictly speaking not correct, and it will be irrelevant soon (since `modulename` is being deprecated, being replaced with `load_name`)

Louwrensth: Good that it is being tackled more properly! I'll simply close this PR then.

Louwrensth:
Closed in favor of:
 - https://github.com/easybuilders/easybuild-framework/pull/5257

PR #4213, opened 2026-09-03
Correct `buildopts` propagation across iterations for OpenBLAS

stamenminkov:
AI usage: `GPT-5` for the suggestion on modifying the loop

On one of the machines I use I had to specify `TARGET` in the `OpenBLAS` EC for the installation to be able to discover the correct architecture. However I noticed that `TARGET` doesn't propagate across all the iterations and the installation would die again on iteration 2/3.

In the easyblock there's a specific code block that defines that:
```
# ensure build/test/install options don't persist between iterations
if self.cfg['enable_ilp64']:
            if self.iter_idx > 0:
                # reset to original build/test/install options
                for key, opt_val in self.orig_opts.items():
                    self.cfg[key] = opt_val
            else:
                # store original options
                for key, opt_val in self.orig_opts.items():
                    self.orig_opts[key] = opt_val
```

However the last line indicates that this `else` is not really saving the current values from `self.cfg`, so the `buildopts` are reset to to just nothing.

### DIFF
--- a/easybuild/easyblocks/o/openblas.py
+++ b/easybuild/easyblocks/o/openblas.py
@@ -84,14 +84,14 @@ class EB_OpenBLAS(ConfigureMake):
 
         # ensure build/test/install options don't persist between iterations
         if self.cfg['enable_ilp64']:
-            if self.iter_idx > 0:
-                # reset to original build/test/install options
-                for key, opt_val in self.orig_opts.items():
-                    self.cfg[key] = opt_val
-            else:
+            if self.iter_idx == 0:
                 # store original options
                 for key in self.orig_opts:
                     self.orig_opts[key] = self.cfg[key]
+            else:
+                # reset to original build/test/install options
+                for key, opt_val in self.orig_opts.items():
+                    self.cfg[key] = opt_val
 
         if '%s=' % TARGET in self.cfg['buildopts']:
             # Add any TARGET in buildopts to default_opts, so it is passed to testopts and installopts
--- a/easybuild/easyblocks/o/openblas.py
+++ b/easybuild/easyblocks/o/openblas.py
@@ -90,7 +90,7 @@ class EB_OpenBLAS(ConfigureMake):
                     self.cfg[key] = opt_val
             else:
                 # store original options
-                for key in self.orig_opts.keys():
+                for key in self.orig_opts:
                     self.orig_opts[key] = self.cfg[key]
 
         if '%s=' % TARGET in self.cfg['buildopts']:
--- a/easybuild/easyblocks/o/openblas.py
+++ b/easybuild/easyblocks/o/openblas.py
@@ -91,7 +91,7 @@ class EB_OpenBLAS(ConfigureMake):
             else:
                 # store original options
                 for key, opt_val in self.orig_opts.items():
-                    self.orig_opts[key] = opt_val
+                    self.orig_opts[key] = self.cfg[key]
 
         if '%s=' % TARGET in self.cfg['buildopts']:
             # Add any TARGET in buildopts to default_opts, so it is passed to testopts and installopts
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/o/openblas.py
```diff
@@ -91,7 +91,7 @@ def configure_step(self):
             else:
                 # store original options
                 for key, opt_val in self.orig_opts.items():
```
Thyre:
```suggestion
                for key in self.orig_opts.keys():
```

easybuild/easyblocks/o/openblas.py
```diff
@@ -90,8 +90,8 @@ def configure_step(self):
                     self.cfg[key] = opt_val
             else:
                 # store original options
-                for key, opt_val in self.orig_opts.items():
-                    self.orig_opts[key] = opt_val
+                for key in self.orig_opts.keys():
```
Flamefire: For dict key iteration just use
```suggestion
                for key in self.orig_opts:
```

Flamefire (changes requested):
Looks correct except for the style issue noted.

The original is actually a no-op: assign orig_opts values to itself
```
for key, opt_val in self.orig_opts.items():
                    self.orig_opts[key] = opt_val
```

Micket (dismissed):
LGTM same conclusion at Flamefire

Micket:
I think the logic would have been more logical to follow if we swapped the order around

```python
            if self.iter_idx == 0:
                # store original options
                for key, opt_val in self.orig_opts.items():
                    self.orig_opts[key] = self.cfg[key]
            else:
                # reset to original build/test/install options
                for key, opt_val in self.orig_opts.items():
                    self.cfg[key] = opt_val
```

but i wasn't able to make a "suggestion" edit (it tried to edit the patch change somehow). I'm just going to ignore that cosmetic change.

Flamefire:
I think that change should be done while we are at it. Reads much clearer. And the new version of course
```python
            if self.iter_idx == 0:
                # store original options
                for key in self.orig_opts:
                    self.orig_opts[key] = self.cfg[key]
            else:
                # reset to original build/test/install options
                for key, opt_val in self.orig_opts.items():
                    self.cfg[key] = opt_val
```

boegel:
> I think that change should be done while we are at it. Reads much clearer. And the new version of course
> 
> ```python
>             if self.iter_idx == 0:
>                 # store original options
>                 for key in self.orig_opts:
>                     self.orig_opts[key] = self.cfg[key]
>             else:
>                 # reset to original build/test/install options
>                 for key, opt_val in self.orig_opts.items():
>                     self.cfg[key] = opt_val
> ```

Yeah, positive logic is always easier to follow.

@stamenminkov Can you update the PR accordingly?

stamenminkov:
Changes applied, thank you all!

Micket (approved):
lgtm

Micket:
Test report by @Micket
 * **SUCCESS** _OpenBLAS-0.3.32-GCC-15.2.0.eb_ 
  

Build succeeded for 1 out of 1 (total: 2 hours 46 mins 27 secs) (1 easyconfigs in total)
vera-icelake-build - Linux Rocky Linux 9.8, x86_64, Intel(R) Xeon(R) Silver 4316 CPU @ 2.30GHz, Python 3.9.25
See https://gist.github.com/Micket/8fc355d3123c3e747cf8e99eb2df1198 for a full test report.

boegel:
bot:build instance:easybuild-test-bot repository:eb-maintainer-stack for:arch=x86_64/amd/zen3 jobargs:EB_ARGS=OpenBLAS-0.3.23-GCC-12.3.0.eb

boegel (approved):
lgtm

PR #4216, opened 2026-09-08
Handle CCache in PyTorch easyblock + consistency fixes

Flamefire:
Avoid using ccache which is usually not desirable for one-off builds.
Allow it when it's a (build)dependency.

Also consistently use `add_enable_option` and ON/OFF
Simplified that to have a default so `add_enable_option('USE_MKLDNN')` adds `USE_MKLDNN=ON`

There should be no functional changes:
 - CCache is usually not installed
 - All others changes have equivalent  behavior

### DIFF
--- a/easybuild/easyblocks/p/pytorch.py
+++ b/easybuild/easyblocks/p/pytorch.py
@@ -254,8 +254,8 @@ class EB_PyTorch(PythonPackage):
     GENERATE_TEST_REPORT_VAR_NAME = 'EASYBUILD_WRITE_PYTORCH_TEST_REPORTS'
 
     @staticmethod
-    def extra_options():
-        extra_vars = PythonPackage.extra_options()
+    def extra_options(extra_vars=None):
+        extra_vars = PythonPackage.extra_options(extra_vars=extra_vars)
         extra_vars.update({
             'build_type': [None, "Build type for CMake, e.g. Release."
                                  "Defaults to 'Release' or 'Debug' depending on toolchainopts[debug]", CUSTOM],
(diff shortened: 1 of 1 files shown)


easybuild/easyblocks/p/pytorch.py
```diff
@@ -428,16 +428,18 @@ def configure_step(self):
...
+            """Add `name=OFF` or `name=ON` depending on enabled"""
+            options.append(f"{name}={'ON' if enabled else 'OFF'}")
 
         # enable verbose mode when --debug is used (to show compiler commands)
         add_enable_option('VERBOSE', build_option('debug'))
 
         # Restrict parallelism
         options.append(f'MAX_JOBS={self.cfg.parallel}')
 
+        add_enable_option('USE_CCACHE', 'ccache' in self.cfg.dependency_names())
```
Thyre, line 441: Should we also consider the EasyBuild option for `ccache`?
```
$ eb --help | grep -A2 ccache
    --use-ccache=PATH   Enable use of ccache to speed up compilation, with
                        specified cache dir (type <class 'str'>; default:
                        False)
```

Flamefire: Probably not. How does that option work exactly? Does it create custom wrappers for g++ etc?
Because the PyTorch option causes it to run `ccache g++` so we might end up with multiple levels of ccache

Thyre: It looks like it does. I haven't used this option before, but am aware of Score-P having issues when `ccache` is passed. However, a test build succeeded.

The logs contain:
```
== 2026-09-09 09:39:36,036 toolchain.py:987 INFO Using ccache for these compiler commands: gcc, g++
== 2026-09-09 09:39:36,036 environment.py:109 INFO Environment variable CCACHE_DIR set to /tmp/ccache (previously undefined)
== 2026-09-09 09:39:36,036 environment.py:109 INFO Environment variable CCACHE_TEMPDIR set to /tmp/eb-4fmc76yl/tmpvqt57az8 (previously undefined)
== 2026-09-09 09:39:36,041 filetools.py:587 INFO Command ccache found at /usr/bin/ccache
== 2026-09-09 09:39:36,041 environment.py:109 INFO Environment variable PATH set to /tmp/eb-4fmc76yl/tmpnim5xhch:/data/EasyBuild-develop/software/Python/3.14.2-GCCcore-15.2.0/bin:[...]')
```

Flamefire: It actually does: `self.symlink_commands({cache_tool: (cache_path, compilers)})`
So it basically adds symlinks to ccache for each compiler command.
So ignoring this here is fine. Maybe even disabling it here. but I'd say not worth the effort as it will only be slower and in a very rare condition
And the change here (usually) avoids the double-ccache call

PR #4228, opened 2026-09-16
use `modulename` for dependency lookup in RPackage easyblock

### LINKED ISSUE #4227: RPackage fails to detect dependencies when modulename differs from extension name
### Description

When installing R extensions in parallel, `RPackage.required_deps` may fail to detect dependencies when the EasyBuild extension name differs from the actual R package name specified in `DESCRIPTION`.

I encountered this while building:

```text
inferCNV-1.22.0-foss-2024a-R-4.4.2.eb
```

with parallel extension installation enabled.

The extension is defined as `inferCNV`, but specifies:

```python
'modulename': '%(namelower)s',
```

so the actual R package name is `infercnv`.

The package `DESCRIPTION` contains:

```text
Package: infercnv
...
Imports: ..., ape, phyclust, ...
```

However, EasyBuild reports:

```text
Required dependencies for inferCNV: []
Missing required dependencies for inferCNV:
```

and starts both `phyclust` and `inferCNV` concurrently:

```text
Started installation of extension phyclust in the background...
Started installation of extension inferCNV in the background...

Checking for completed extension installations (2 running)...
```

`inferCNV` then fails because `phyclust` has not completed installation yet:

```text
ERROR: dependency ‘phyclust’ is not available for package ‘infercnv’
```

### Cause

`RPackage.required_deps` builds the dependency map using the package names parsed from the R `DESCRIPTION` files.

The dependency map therefore contains an entry for:

```text
infercnv
```

but dependency lookup currently uses the EasyBuild extension name:

```python
self._required_deps = deps_map.get(self.name, [])
```

For this extension:

```text
self.name = inferCNV
modulename = infercnv
```

so the lookup does not find the `infercnv` entry.

This is not necessarily a case-sensitivity issue: more generally, the extension name and the actual R package name may differ, and `modulename` already provides the actual package name in that situation.

### Expected behavior

When `modulename` is specified for an R extension, it should be used to identify the package when resolving dependencies from `DESCRIPTION`.

If no `modulename` is specified, the existing behavior using `self.name` should remain unchanged.

For example:

```python
pkg_name = self.options.get('modulename') or self.name
self._required_deps = deps_map.get(pkg_name, [])
```

For `inferCNV`, this resolves the lookup using:

```text
infercnv
```

and correctly detects `phyclust` as a dependency.

The parallel extension installer can then wait for `phyclust` to complete before starting `inferCNV`.

### Workaround

Disabling parallel extension installation avoids the issue because the extensions are installed sequentially.

PhoenixEmik:
(created using `eb --new-pr`)
Fixes: #4227


Micket (commented):
logic looks correct to me, but i know there is ongoing work on modulename ("load_name") for general use (outside of somewhat outdated "options" block), so I suspect we want to revisit this when that dust has settled on chose changes

